PR #139, opened 2014-11-15
Verified file transfer

Taloth:
Episode moves now include filesize checks.
Basically the procedure is as follows:
- If a hardlink was requested, try the hardlink first revert to copy if that fails. (this is what was done already)
- If a copy, do the copy and compare the filesizes, retry 2 times if wrong, throws IOException if it still fails.
- If a move a couple of steps:

1. make a hardlink to '{filename}.movebackup' in same directory. If fails, skip to 3.
2. Move that backup file to the destination and compare the filesizes. If succeeds, delete original and exit.
3. Delete backup file if created. Then do a Copy like before, 2 retries at the most.
4. Delete original if succeeded, throw IOException if failed.

The goal of this procedure is to detect CIFS errors. Another method would be to revert to read and write the file streams from .net directly, to append missing bytes, but I didn't want to do that at this point.
So if a copy fails, delete the partial destination and retry, instead of appending the missing bytes.

I had to include 1 commit from the torrent branch which adds hardlink support.

I opted to create a separate service for all this instead of embedding it in the DiskProvider, mainly because this makes it possible to have unit tests for it.

### DIFF
--- a/src/NzbDrone.Common.Test/DiskTests/DiskTransferServiceFixture.cs
+++ b/src/NzbDrone.Common.Test/DiskTests/DiskTransferServiceFixture.cs
@@ -48,9 +48,23 @@ namespace NzbDrone.Common.Test.DiskTests
         }
 
         [Test]
+        public void should_not_use_verified_transfer_on_windows()
+        {
+            WindowsOnly();
+
+            var result = Subject.TransferFile(_sourcePath, _targetPath, TransferMode.Move);
+
+            Mocker.GetMock<IDiskProvider>()
+                .Verify(v => v.TryCreateHardLink(_sourcePath, _backupPath), Times.Never());
+
+            Mocker.GetMock<IDiskProvider>()
+                .Verify(v => v.MoveFile(_sourcePath, _targetPath, false), Times.Once());
+        }
+
+        [Test]
         public void should_retry_if_partial_copy()
         {
-            WithSuccessfulHardlink(_sourcePath, _backupPath);
+            MonoOnly();
 
             var retry = 0;
             Mocker.GetMock<IDiskProvider>()
@@ -69,6 +83,8 @@ namespace NzbDrone.Common.Test.DiskTests
         [Test]
         public void should_retry_twice_if_partial_copy()
         {
+            MonoOnly();
+
             var retry = 0;
             Mocker.GetMock<IDiskProvider>()
                 .Setup(v => v.CopyFile(_sourcePath, _tempTargetPath, false))
@@ -87,6 +103,8 @@ namespace NzbDrone.Common.Test.DiskTests
         [Test]
         public void should_hardlink_before_move()
         {
+            MonoOnly();
+
             WithSuccessfulHardlink(_sourcePath, _backupPath);
 
             var result = Subject.TransferFile(_sourcePath, _targetPath, TransferMode.Move);
@@ -98,6 +116,8 @@ namespace NzbDrone.Common.Test.DiskTests
         [Test]
         public void should_remove_source_after_move()
         {
+            MonoOnly();
+
             WithSuccessfulHardlink(_sourcePath, _backupPath);
 
             Mocker.GetMock<IDiskProvider>()
@@ -112,6 +132,8 @@ namespace NzbDrone.Common.Test.DiskTests
         [Test]
         public void should_remove_backup_if_move_throws()
         {
+            MonoOnly();
+
             WithSuccessfulHardlink(_sourcePath, _backupPath);
 
             Mocker.GetMock<IDiskProvider>()
@@ -126,6 +148,8 @@ namespace NzbDrone.Common.Test.DiskTests
         [Test]
         public void should_remove_partial_if_move_fails()
         {
+            MonoOnly();
+
             WithSuccessfulHardlink(_sourcePath, _backupPath);
 
             Mocker.GetMock<IDiskProvider>()
@@ -144,6 +168,8 @@ namespace NzbDrone.Common.Test.DiskTests
         [Test]
         public void should_fallback_to_copy_if_hardlink_failed()
         {
+            MonoOnly();
+
             WithFailedHardlink();
 
             var result = Subject.TransferFile(_sourcePath, _targetPath, TransferMode.Move);
--- a/src/NzbDrone.Common.Test/DiskTests/DiskTransferServiceFixture.cs
+++ b/src/NzbDrone.Common.Test/DiskTests/DiskTransferServiceFixture.cs
@@ -1,0 +1,390 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text;
+using Moq;
+using NUnit.Framework;
+using NzbDrone.Common.Disk;
+using NzbDrone.Test.Common;
+using FluentAssertions;
+
+namespace NzbDrone.Common.Test.DiskTests
+{
+    [TestFixture]
+    public class DiskTransferServiceFixture : TestBase<DiskTransferService>
+    {
+        private readonly String _sourcePath = @"C:\source\my.video.mkv".AsOsAgnostic();
+        private readonly String _targetPath = @"C:\target\my.video.mkv".AsOsAgnostic();
+        private readonly String _backupPath = @"C:\source\my.video.mkv.backup~".AsOsAgnostic();
+        private readonly String _tempTargetPath = @"C:\target\my.video.mkv.partial~".AsOsAgnostic();
+
+        [SetUp]
+        public void SetUp()
+        {
+            Mocker.GetMock<IDiskProvider>(MockBehavior.Strict);
+
+            WithEmulatedDiskProvider();
+
+            WithExistingFile(_sourcePath);
+        }
+
+        [Test]
+        public void should_hardlink_only()
+        {
+            WithSuccessfulHardlink(_sourcePath, _targetPath);
+
+            var result = Subject.TransferFile(_sourcePath, _targetPath, TransferMode.HardLink);
+
+            result.Should().Be(TransferMode.HardLink);
+        }
+
+        [Test]
+        public void should_throw_if_hardlink_only_failed()
+        {
+            WithFailedHardlink();
+
+            Assert.Throws<IOException>(() => Subject.TransferFile(_sourcePath, _targetPath, TransferMode.HardLink));
+        }
+
+        [Test]
+        public void should_retry_if_partial_copy()
+        {
+            WithSuccessfulHardlink(_sourcePath, _backupPath);
+
+            var retry = 0;
+            Mocker.GetMock<IDiskProvider>()
+                .Setup(v => v.CopyFile(_sourcePath, _tempTargetPath, false))
+                .Callback(() =>
+                    {
+                        WithExistingFile(_tempTargetPath, true, 900);
+                        if (retry++ == 1) WithExistingFile(_tempTargetPath, true, 1000);
+                    });
+
+            var result = Subject.TransferFile(_sourcePath, _targetPath, TransferMode.Copy);
+
+            ExceptionVerification.ExpectedWarns(1);
+        }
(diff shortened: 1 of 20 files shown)


Taloth:
@markus101 @kayone Finally took some time to move TransferFolder to the service as well.

src/NzbDrone.Common/Disk/DiskTransferService.cs
```diff
@@ -0,0 +1,267 @@
...
+{
+    public interface IDiskTransferService
+    {
+        TransferMode TransferFolder(String sourcePath, String targetPath, TransferMode mode);
+        TransferMode TransferFolderVerified(String sourcePath, String targetPath, TransferMode mode);
+        TransferMode TransferFile(String sourcePath, String targetPath, TransferMode mode, bool overwrite = false);
+        TransferMode TransferFileVerified(String sourcePath, String targetPath, TransferMode mode);
+    }
+
+    public class DiskTransferService : IDiskTransferService
```
kayone: There is a lot of shared code between the `Transfer` and `Verified` counterpart. we should move the shared logic to private methods.

kayone: It seems like putting the `Verification` behind a flag would help. 

`TransferMode TransferFile(String sourcePath, String targetPath, TransferMode mode, bool overwrite = false, bool verified = false);`

kayone: or we could even add it as a `TransferMode` flag.

src/NzbDrone.Common/Disk/DiskTransferService.cs
```diff
@@ -0,0 +1,267 @@
...
+        public DiskTransferService(IDiskProvider diskProvider, Logger logger)
+        {
+            _diskProvider = diskProvider;
+            _logger = logger;
+        }
+
+        public TransferMode TransferFolder(String sourcePath, String targetPath, TransferMode mode)
+        {
+            Ensure.That(sourcePath, () => sourcePath).IsValidPath();
+            Ensure.That(targetPath, () => targetPath).IsValidPath();
```
kayone: we should also check that target isn't a child of source.

src/NzbDrone.Common/Disk/DiskTransferService.cs
```diff
@@ -0,0 +1,267 @@
...
+        public TransferMode TransferFile(String sourcePath, String targetPath, TransferMode mode, bool overwrite = false)
+        {
+            Ensure.That(sourcePath, () => sourcePath).IsValidPath();
+            Ensure.That(targetPath, () => targetPath).IsValidPath();
+
+            _logger.Debug("{0} [{1}] > [{2}]", mode, sourcePath, targetPath);
+
+            if (sourcePath.PathEquals(targetPath))
+            {
+                _logger.Warn("Source and destination can't be the same {0}", sourcePath);
```
kayone: why is this not an error?

src/NzbDrone.Common/Disk/DiskTransferService.cs
kayone: or the same

src/NzbDrone.Common/Disk/DiskTransferService.cs
```diff
@@ -0,0 +1,267 @@
...
+
+            if (mode.HasFlag(TransferMode.Move))
+            {
+                if (TryMoveFile(sourcePath, targetPath))
+                {
+                    return TransferMode.Move;
+                }
+            }
+
+            throw new IOException(String.Format("Failed to completely transfer [{0}] to [{1}], aborting.", sourcePath, targetPath));
```
kayone: do we need `completely` in there?

src/NzbDrone.Common/Disk/DiskTransferService.cs
```diff
@@ -0,0 +1,267 @@
...
+                    return true;
+                }
+
+                _diskProvider.DeleteFile(targetPath);
+
+                if (i == RetryCount)
+                {
+                    _logger.Error("Failed to completely transfer [{0}] to [{1}], aborting.", sourcePath, targetPath, i + 1, RetryCount);
+                }
+                else if (i == 0)
```
kayone: I think it's safe to log both as warn.

src/NzbDrone.Common/Disk/DiskTransferService.cs
```diff
@@ -0,0 +1,267 @@
...
+            }
+
+            throw new IOException(String.Format("Failed to completely transfer [{0}] to [{1}], aborting.", sourcePath, targetPath));
+        }
+
+        private Boolean TryCopyFile(String sourcePath, String targetPath)
+        {
+            var originalSize = _diskProvider.GetFileSize(sourcePath);
+
+            for (var i = 0; i <= RetryCount; i++)
```
kayone: would this ever pass when tried right after? shouldn't we put some sleep in there?

src/NzbDrone.Common/Disk/DiskTransferService.cs
```diff
@@ -0,0 +1,267 @@
...
+            {
+                _diskProvider.CopySingleFile(sourcePath, targetPath);
+
+                var targetSize = _diskProvider.GetFileSize(targetPath);
```
kayone: what if the file doesn't exist?

src/NzbDrone.Common/Disk/DiskTransferService.cs
```diff
@@ -0,0 +1,267 @@
...
+
+            _logger.Trace("Hardlink move failed, reverting to copy.");
+            if (TryCopyFile(sourcePath, targetPath))
+            {
+                _logger.Trace("Copy succeeded, deleting source.");
+                _diskProvider.DeleteFile(sourcePath);
+                return true;
+            }
+
+            _logger.Trace("Copy failed.");
```
kayone: include source and destination.

src/NzbDrone.Common/Disk/IDiskProvider.cs
```diff
@@ -25,13 +25,9 @@ public interface IDiskProvider
...
         void CreateFolder(string path);
-        void CopyFolder(string source, string destination);
-        void MoveFolder(string source, string destination);
-        void TransferFolder(string source, string destination, TransferMode transferMode);
         void DeleteFile(string path);
-        void CopyFile(string source, string destination, bool overwrite = false);
-        void MoveFile(string source, string destination, bool overwrite = false);
-        TransferMode TransferFile(string source, string destination, TransferMode transferMode, bool overwrite = false);
+        void CopySingleFile(string source, string destination, bool overwrite = false);
+        void MoveSingleFile(string source, string destination, bool overwrite = false);
```
kayone: Why not `CopyFile`, `MoveFile`?

src/NzbDrone.Common/Disk/DiskProviderBase.cs
```diff
@@ -234,17 +178,21 @@ public void DeleteFile(string path)
...
-        }
+            Ensure.That(source, () => source).IsValidPath();
+            Ensure.That(destination, () => destination).IsValidPath();
 
-        public void MoveFile(string source, string destination, bool overwrite = false)
-        {
-            TransferFile(source, destination, TransferMode.Move, overwrite);
+            if (source.PathEquals(destination))
+            {
+                Logger.Warn("Source and destination can't be the same {0}", source);
```
kayone: I'm thinking this should throw.

src/NzbDrone.Common.Test/DiskTests/DiskTransferServiceFixture.cs
```diff
@@ -0,0 +1,314 @@
...
+using Moq;
+using NUnit.Framework;
+using NzbDrone.Common.Disk;
+using NzbDrone.Test.Common;
+using FluentAssertions;
+
+namespace NzbDrone.Common.Test.DiskTests
+{
+    [TestFixture]
+    public class DiskTransferServiceFixture : TestBase<DiskTransferService>
```
kayone: AFAICT, you don't have any tests for when copy/move fails because destination is readonly or doesn't exists.

kayone:
I really like the move to a separate service on top of diskprovider. :+1:

src/NzbDrone.Common/Disk/DiskTransferService.cs
Taloth: The purpose of this change is to detect and gracefully handle the problem of partial file copies/moves we've seen with CIFS share. the error indicates that we probably were able to copy it partially, but it failed to finish.

src/NzbDrone.Common/Disk/DiskTransferService.cs
Taloth: You mean the Debug one?

src/NzbDrone.Common/Disk/DiskTransferService.cs
Taloth: Are you sure? Most, if not all, paths are already logged by the caller of this method.

src/NzbDrone.Common/Disk/IDiskProvider.cs
Taloth: At first I just needed a separate name to refactor easily. Just kept it coz it would be a gentle push for other devs to use the TransferService. But I'm okay either way.

Taloth:
Every few months I work on this, LOL.

One concern I have with this is that it will try to do every Move via Hardlink first. Which means if the hardlink api is borked/hangs then we're in serious trouble.

Taloth:
After testing it locally I had to fix a few things. Seems to work as advertised.
I tested it on my server connected via CIFS to a bad (slow) drive on windows.
It indeed properly reverts to copy.

I changed the process a bit:

Move:
- Hardlink to file with ```{source}.backup~``` suffix.
- Move ```{source}.backup~``` to ```{target}.partial~```.
- If move succeeds and the target file is the same size. Then  ```{target}.partial~``` is renamed to ```{target}``` and the  ```{source}``` file deleted.
- Otherwise it reverts to Copy followed by a deletion of ```{source}```.

Copy:
- Copy ```{source}``` to ```{target}.partial~```.
- If copy fails, or target file isn't of the same size, then the partial is deleted, and the process retries once.
- Otherwise the  ```{target}.partial~``` is renamed to ```{target}```.

This process is almost the same as it was before, but I added ```.partial~``` to ensure no incomplete .mkv files would be in the target directory.

The idea of the hardlink backup is to preserve the original while the move is made.
Advantage is that moves can be instantaneous if local (hardlink->move->rename).
If remote then it transparently becomes a hardlink->copy->rename.

The problem with this is that the source filesystem must support hardlinks.

But ReFS doesn't.
This isn't an issue for the remote scenario, coz a failing hardlink will mean it reverts to a copy, which effectively happens anyway.
But for a local scenario, it would mean it always copies, and never moves.

@markus101  My question is how to deal with this?
I may have an idea for a plan B, which involves opening {source} for reading without filelock, which 'should' keep a line to the file contents even after it gets deleted. But I don't know if it works.
Allowing me to recreate it if the move fails.

But I could also just try detect on windows whether the volume FS is ReFS and use a simple move instead.

@markus101 Second question:

I wrote a commit that adds a config option, but it's ugly IMHO. It requires me to add configService to several places in Sonarr. I parked the commit here: https://github.com/Taloth/NzbDrone/commit/99c0c30460947f890552249917234c39687baf41
I would rather do without it, so I really could use a second opinion on this.

markus101:
> But I could also just try detect on windows whether the volume FS is ReFS and use a simple move instead.

Are we seeing failing transfers on Windows? I know *nix to SMB shares has been an issue for some people, but I don't recall any time its failed on Windows, maybe we could skip the verified transfer on Windows entirely?

Regarding the option, yeah that gets its hands all over the place, I think we can leave it out for now and add it later if we think we'll need it.

Taloth:
I recall one unconfirmed report on windows with remote smb share, but I'll have to check irc and the forum.

PR #156, opened 2014-12-10
Add Series Options

markus101:
@kayone @Taloth

This adds the backend support to let us add series and let users choose which episodes to monitor automatically (instead of at a season level) and let them perform a backlog search for missing episodes after adding.

This requires us to listen for two events, the series being added and the disk scan completing. I do feel its a bit fragile right now, since it will do nothing if Sonarr is restarted, the other option is a database table to store it.

No UI or tests in this one, wanted to make sure you guys are good with the approach. Also its specific to this scenario, maybe we need to make it more generic later on, but I didn't want to deal with that unless we require it (I couldn't think of a reason why we would right now).


src/NzbDrone.Core/Tv/SeriesAddedService.cs
```diff
@@ -0,0 +1,140 @@
...
+            _addSeriesOptionsCache.Set(message.Series.Id.ToString(), message.Options);
+        }
+
+        public void Handle(SeriesScannedEvent message)
+        {
+            var options = _addSeriesOptionsCache.Find(message.Series.Id.ToString());
+
+            if (options == null)
+            {
+                _logger.Debug("[{0}] Was not recently added, skipping post-add actions", message.Series.Title);
```
markus101: Rebased, so this is hidden, but not really sure how we want to handle this exactly, this is all events that need to get handled, not commands.

I'm not a fan of if Sonarr is restarted this logic of ignoring/searching is lost though.

markus101: @Taloth had some ideas to make this more generic (his original idea), maybe implementing that backed with a DB instead of a cache will help.

src/NzbDrone.Api/Series/SeriesModule.cs
```diff
@@ -109,7 +110,12 @@ private List<SeriesResource> AllSeries()
 
         private int AddSeries(SeriesResource seriesResource)
         {
-            return GetNewId<Core.Tv.Series>(_seriesService.AddSeries, seriesResource);
+            var options = new AddSeriesOptions();
+            options.InjectFrom(Request.Body);
```
Taloth: Won't SeriesResource and AddSeriesOptions both come from the same object? Probably better to add a property 'addOptions'.

src/NzbDrone.Core.Test/TvTests/SeriesAddedServiceTests/SetEpisodeMontitoredFixture.cs
```diff
@@ -0,0 +1,227 @@
...
+                                        .With(e => e.EpisodeFileId = 1)
+                                         //Future
+                                        .TheNext(1)
+                                        .With(e => e.EpisodeFileId = 0)
+                                        .With(e => e.AirDateUtc = DateTime.UtcNow.AddDays(7))
+                                        //Future/TBA
+                                        .TheNext(1)
+                                        .With(e => e.EpisodeFileId = 0)
+                                        .With(e => e.AirDateUtc = null)
+                                        //TODO: Need cases for specials (they should always be ignored)
```
Taloth: Yes, Specials definitely needs to be tested.

src/NzbDrone.Core.Test/TvTests/SeriesAddedServiceTests/SetEpisodeMontitoredFixture.cs
```diff
@@ -0,0 +1,227 @@
...
+
+            Mocker.GetMock<ISeriesService>()
+                  .Verify(v => v.UpdateSeries(It.Is<Series>(s => s.Seasons.Single(n => n.SeasonNumber == 1).Monitored)));
+
+            Mocker.GetMock<ISeriesService>()
+                  .Verify(v => v.UpdateSeries(It.Is<Series>(s => s.Seasons.Where(n => n.SeasonNumber != 1).All(n => !n.Monitored))));
+        }
+
+        [Test]
+        public void should_not_monitor_missing_speicals()
```
Taloth: speicals = specials

src/NzbDrone.Core.Test/TvTests/SeriesAddedServiceTests/SetEpisodeMontitoredFixture.cs
```diff
@@ -0,0 +1,227 @@
...
+
+            WithSeriesScannedEvent();
+
+            VerifyMonitored(e => e.AirDateUtc.HasValue && e.AirDateUtc.Value.After(DateTime.UtcNow));
+            VerifyMonitored(e => !e.AirDateUtc.HasValue);
+            VerifyNotMonitored(e => e.AirDateUtc.HasValue && e.AirDateUtc.Value.Before(DateTime.UtcNow));
+        }
+
+        [Test]
+        public void should_be_able_to_monitor_first_season_only()
```
Taloth: Is this the 'download pilot' kind of option?

src/NzbDrone.Core/Tv/AddSeriesOptions.cs
```diff
@@ -0,0 +1,18 @@
...
+namespace NzbDrone.Core.Tv
+{
+    public class AddSeriesOptions
+    {
+        public bool SearchForMissingEpisodes { get; set; }
+        public bool IgnoreEpisodesWithFiles { get; set; }
+        public bool IgnoreEpisodesWithoutFiles { get; set; }
+        public List<int> IgnoreSeasons { get; set; }
+
+        public AddSeriesOptions()
```
Taloth: Aren't we missing 'IgnoreAlreadyAired' or is that IgnoreEpisodesWithFiles+IgnoreEpisodesWithoutFiles?

Taloth:
Yes, make it db backed. But the proposed db migration needs work.
Also include a SeriesID column for example, for lookups.
I'm also wondering about housekeeping, when to remove stuff from the db, although the seriesID column would help with that.

My original solution was more a pipeline, coz I assumed we had to do MediaInfo on a background/task too to be able to import existing episodes without saturating cpu/io. But my original solutions are usually a bit convoluted, so nevermind me.

I looked through the remaining logic, couldn't find anything wrong.

src/NzbDrone.Core.Test/TvTests/SeriesAddedServiceTests/SetEpisodeMontitoredFixture.cs
markus101: Yeah, I'd personally use it to grab a series to see if its any good.

markus101:
@kayone @Taloth any more feedback for this one?

src/NzbDrone.Core/Datastore/Migration/073_task_queue.cs
```diff
@@ -0,0 +1,17 @@
...
+{
+    [Migration(73)]
+    public class task_queue : NzbDroneMigrationBase
+    {
+        protected override void MainDbUpgrade()
+        {
+            Create.TableForModel("TaskQueue")
+                  .WithColumn("TypeName").AsString().NotNullable()
+                  .WithColumn("Priority").AsInt32().NotNullable()
+                  .WithColumn("Parameters").AsString().NotNullable();
```
Taloth: TaskQueue doesn't exist does it? Probably a leftover?

src/NzbDrone.Core/Datastore/Migration/074_add_add_options_to_series.cs
```diff
@@ -0,0 +1,14 @@
...
+using NzbDrone.Core.Datastore.Migration.Framework;
+
+namespace NzbDrone.Core.Datastore.Migration
+{
+    [Migration(74)]
+    public class add_add_options_to_series : NzbDroneMigrationBase
+    {
+        protected override void MainDbUpgrade()
+        {
+            Alter.Table("Series").AddColumn("AddOptions").AsString().Nullable();
```
Taloth: So we're going to add a column in this PR and then remove it again once the TaskQueue is implemented?

kayone: no, It'll likely stay to avoid the need of chaining different commands, Can't ignore episodes until the series has been updated and AddOptions was part of the add command,

unless we add an overload to the UpdateSeries method.

Taloth: I thought markus was implementing that chaining (Deferred events handling) in a later PR.

markus101: I think thats a separate table/concept. I think we still need to be able to chain different events, like we're doing here.

The other case is when we do a series/episode refresh that adds a new already aired series (which I had to disable because of this case):

1. episode is missing from Sonarr
2. Gets added to TVDB, but Skyhook cache delays it getting into Sonarr
3. User grabs file manually and imports it
4. Sonarr picks up new episode
5. Event is triggered that contains newly added past episode
6. Sonarr searches for it because its missing
7. Sonarr queues episode in download client
8. Sonarr scans disk and find existing file
9. Queued download finishes and imports overwriting the existing file

Ultimately, the user downloaded the same file 2x for no reason, Sonarr should do the disk scan before the search, but that requires two events to be handled with some sort of storage in the middle.

1. Refresh event triggered
2. Sonarr stores event with newly added episodes in a "queue" (if no newly added episodes this step is skipped)
3. Disk scan complete event is triggered (New episode files from disk are now in DB)
4. Sonarr checks the "queue" if there is no stored event from the refresh, its done, otherwise start step 5
5. Sonarr searches for newly added + missing episode
6. Finds release
7. Queues release

The command queue PR still makes a lot of sense because it eliminates a few minor issues we have today and gives us better command tracking.

src/NzbDrone.Core/IndexerSearch/EpisodeSearchService.cs
```diff
@@ -66,6 +66,41 @@ public void MissingEpisodesAiredAfter(DateTime dateTime, IEnumerable<Int32> grab
             _logger.ProgressInfo("Completed search for {0} episodes. {1} reports downloaded.", missing.Count, downloadedCount);
         }
 
+        private void SearchForMissingEpisodes(List<Episode> episodes)
+        {
+            _logger.ProgressInfo("Performing missing search for {0} episodes", episodes.Count);
+            var downloadedCount = 0;
+
+            using (var rateGate = new RateGate(100, TimeSpan.FromSeconds(60)))
```
kayone: this assumes we can process ~2 items a second, thats going to all the indexers and processing the results in 600ms.

also, why would we rate block this? it's happening serially not in parallel.

src/NzbDrone.Core/IndexerSearch/EpisodeSearchService.cs
```diff
@@ -66,6 +66,41 @@ public void MissingEpisodesAiredAfter(DateTime dateTime, IEnumerable<Int32> grab
...
+            {
+                foreach (var series in episodes.GroupBy(e => e.SeriesId))
+                {
+                    foreach (var season in series.Select(e => e).GroupBy(e => e.SeasonNumber))
```
kayone: move the first select to the parent foreach

src/NzbDrone.Core/IndexerSearch/EpisodeSearchService.cs
```diff
@@ -66,6 +66,41 @@ public void MissingEpisodesAiredAfter(DateTime dateTime, IEnumerable<Int32> grab
...
+                            var decisions = _nzbSearchService.SeasonSearch(series.Key, season.Key);
+                            var processed = _processDownloadDecisions.ProcessDecisions(decisions);
+
+                            downloadedCount += processed.Grabbed.Count;
+                        }
+
+                        else
+                        {
+                            var decisions = _nzbSearchService.EpisodeSearch(season.First());
+                            var processed = _processDownloadDecisions.ProcessDecisions(decisions);
```
kayone: these two lines are repeated, move them outside of the if/else

src/NzbDrone.Core/IndexerSearch/EpisodeSearchService.cs
```diff
@@ -79,36 +114,33 @@ public void Execute(EpisodeSearchCommand message)
...
+                                                                        SortKey = "Id",
+                                                                        FilterExpression =
+                                                                            v =>
+                                                                            v.Monitored == true &&
+                                                                            v.Series.Monitored == true
+                                                                    }).Records.ToList();
             }
 
-            _logger.ProgressInfo("Completed missing search for {0} episodes. {1} reports downloaded.", missing.Count, downloadedCount);
+            var missing = episodes.Where(e => !_queueService.GetQueue().Select(q => q.Episode.Id).Contains(e.Id)).ToList();
```
kayone: get the queue outside and project it to episode Ids, otherwise we grab the queue once per item in the episode list and process the whole list multiple times.

src/NzbDrone.Core/IndexerSearch/EpisodeSearchService.cs
markus101: The idea is that we won't spam indexers looking for missing episodes in batch (namely when doing a search for all missing), but yes, the numbers don't match.

src/NzbDrone.Core.Test/TvTests/SeriesAddedHandlerTests/SetEpisodeMontitoredFixture.cs
```diff
@@ -0,0 +1,209 @@
...
+        private Series _series;
+        private List<Episode> _episodes;
+
+        [SetUp]
+        public void Setup()
+        {
+            var seasons = 4;
+
+            _series = Builder<Series>.CreateNew()
+                                     .With(s => s.Seasons = Builder<Season>.CreateListOfSize(seasons).All().With(n => n.Monitored = true).Build().ToList())
```
kayone: needs more line breaks ;)

src/NzbDrone.Core.Test/TvTests/SeriesAddedHandlerTests/SetEpisodeMontitoredFixture.cs
```diff
@@ -0,0 +1,209 @@
...
+                  .Setup(s => s.GetEpisodeBySeries(It.IsAny<int>()))
+                  .Returns(_episodes);
+        }
+
+        private void WithSeriesAddedEvent(AddSeriesOptions options)
+        {
+            _series.AddOptions = options;
+        }
+
+        private void WithSeriesScannedEvent()
```
kayone: `TriggerSeriesScannedEvent()`

src/NzbDrone.Core/IndexerSearch/EpisodeSearchService.cs
kayone: tbh, I think we should remove the rateGate here.

src/NzbDrone.Core/IndexerSearch/EpisodeSearchService.cs
```diff
@@ -79,36 +115,34 @@ public void Execute(EpisodeSearchCommand message)
...
-                                                         SortDirection = SortDirection.Ascending,
-                                                         SortKey = "Id",
-                                                         FilterExpression = v => v.Monitored == true && v.Series.Monitored == true
-                                                     }).Records.ToList();
-
-            var missing = episodes.Where(e => !_queueService.GetQueue().Select(q => q.Episode.Id).Contains(e.Id)).ToList();
-
-            _logger.ProgressInfo("Performing missing search for {0} episodes", missing.Count);
-            var downloadedCount = 0;
+            if (message.SeriesId > 0)
```
kayone: we should switch to `int?` instead of using 0.

src/NzbDrone.Core/IndexerSearch/EpisodeSearchService.cs
```diff
@@ -79,36 +115,34 @@ public void Execute(EpisodeSearchCommand message)
...
+                                                                        SortKey = "Id",
+                                                                        FilterExpression =
+                                                                            v =>
+                                                                            v.Monitored == true &&
+                                                                            v.Series.Monitored == true
+                                                                    }).Records.ToList();
             }
 
-            _logger.ProgressInfo("Completed missing search for {0} episodes. {1} reports downloaded.", missing.Count, downloadedCount);
+            var queue = _queueService.GetQueue();
```
kayone: `var queuedEpisodes = _queueService.GetQueue().Select(q => q.Episode.Id)`

src/NzbDrone.Core/IndexerSearch/EpisodeSearchService.cs
```diff
@@ -79,36 +115,34 @@ public void Execute(EpisodeSearchCommand message)
...
+                episodes = _episodeService.EpisodesWithoutFiles(new PagingSpec<Episode>
+                                                                    {
+                                                                        Page = 1,
+                                                                        PageSize = 100000,
+                                                                        SortDirection = SortDirection.Ascending,
+                                                                        SortKey = "Id",
+                                                                        FilterExpression =
+                                                                            v =>
+                                                                            v.Monitored == true &&
+                                                                            v.Series.Monitored == true
```
kayone: does this work? isn't this the same filter used for the missing page?

is this method ever used with unmonitored/series episdoes? if not we should make the filter built in and only pass in the other paging specs.

src/NzbDrone.Core/Tv/SeriesScannedHandler.cs
```diff
@@ -0,0 +1,106 @@
...
+        private void SearchForMissingEpisodes(Series series)
+        {
+            _commandExecutor.PublishCommand(new MissingEpisodeSearchCommand(series.Id));
+        }
+
+        private void HandleScanEvents(Series series)
+        {
+            if (series.AddOptions == null)
+            {
+                _logger.Debug("[{0}] Was not recently added, skipping post-add actions", series.Title);
```
kayone: prob should be trace.

kayone: or add one in info where it says new series applying the following rules. instead of logging one every other time,

src/NzbDrone.Core/Tv/SeriesScannedHandler.cs
```diff
@@ -0,0 +1,106 @@
...
+                                  ICommandExecutor commandExecutor,
+                                  Logger logger)
+        {
+            _seriesService = seriesService;
+            _episodeService = episodeService;
+            _commandExecutor = commandExecutor;
+            _logger = logger;
+        }
+
+        private void UnmonitorEpisodes(IEnumerable<Episode> episodes)
```
kayone: helper methods should be after the methods that use them,

src/NzbDrone.Core/Tv/SeriesScannedHandler.cs
```diff
@@ -0,0 +1,106 @@
...
+                _logger.Debug("[{0}] Was not recently added, skipping post-add actions", series.Title);
+                return;
+            }
+
+            var episodes = _episodeService.GetEpisodeBySeries(series.Id);
+            SetEpisodeMonitoredStatus(series, episodes);
+
+            if (series.AddOptions.SearchForMissingEpisodes)
+            {
+                SearchForMissingEpisodes(series);
```
kayone: inline method.

src/NzbDrone.Core/Tv/SeriesScannedHandler.cs
```diff
@@ -0,0 +1,106 @@
...
+            }
+
+            series.AddOptions = null;
+            _seriesService.UpdateSeries(series);
```
kayone: I'm thinking this should be a more targeted update, using the `SetFields` method in the repo.

src/NzbDrone.Core/Tv/SeriesScannedHandler.cs
```diff
@@ -0,0 +1,106 @@
...
+
+            if (series.AddOptions.IgnoreEpisodesWithoutFiles)
+            {
+                _logger.Debug("Ignoring Episodes without Files");
+                UnmonitorEpisodes(episodes.Where(e => !e.HasFile && e.AirDateUtc.HasValue && e.AirDateUtc.Value.Before(DateTime.UtcNow)));
+            }
+
+            foreach (var season in series.Seasons)
+            {
+                if (series.Seasons.Select(s => s.SeasonNumber).MaxOrDefault() != season.SeasonNumber &&
```
kayone: `var lastSeason = series.Seasons.Select(s => s.SeasonNumber).MaxOrDefault()`

src/UI/AddSeries/MonitoringTooltipTemplate.hbs
```diff
@@ -0,0 +1,14 @@
+﻿<dl class="monitor-tooltip-contents">
+    <dt>All</dt>
+    <dd>Monitor all episodes except for specials</dd>
```
kayone: `Monitor all episodes except specials`?

src/UI/AddSeries/MonitoringTooltipTemplate.hbs
```diff
@@ -0,0 +1,14 @@
...
+    <dt>Future</dt>
+    <dd>Only episodes that haven't aired yet</dd>
```
kayone: "Only monitor episodes that haven't aired yet"

src/UI/AddSeries/MonitoringTooltipTemplate.hbs
```diff
@@ -0,0 +1,14 @@
+﻿<dl class="monitor-tooltip-contents">
+    <dt>All</dt>
+    <dd>Monitor all episodes except for specials</dd>
+    <dt>Future</dt>
```
kayone: Not aired yet

src/UI/AddSeries/MonitoringTooltipTemplate.hbs
```diff
@@ -0,0 +1,14 @@
+﻿<dl class="monitor-tooltip-contents">
```
kayone: https://gist.github.com/kayone/99587521e5bbf724006e

kayone:
this concludes my review ;)

src/NzbDrone.Core/IndexerSearch/EpisodeSearchService.cs
markus101: It does work and yes its used with unmonitored series/episodes (the missing page has a filter).

Not a big fan of this, the other option it to have a separate method that gets missing.

PR #161, opened 2014-12-25
New: MediaBrowser notifications

markus101:
@kayone @Taloth 

Converted it from RestSharp to our own HttpClient and since MediaBrowser actually supports library updates now we can actually release it.

### DIFF
--- a/src/NzbDrone.Core/Notifications/MediaBrowser/MediaBrowser.cs
+++ b/src/NzbDrone.Core/Notifications/MediaBrowser/MediaBrowser.cs
@@ -1,0 +1,65 @@
+﻿using System;
+using System.Collections.Generic;
+using FluentValidation.Results;
+using NzbDrone.Common.Extensions;
+using NzbDrone.Core.Tv;
+
+namespace NzbDrone.Core.Notifications.MediaBrowser
+{
+    public class MediaBrowser : NotificationBase<MediaBrowserSettings>
+    {
+        private readonly IMediaBrowserService _mediaBrowserService;
+
+        public MediaBrowser(IMediaBrowserService mediaBrowserService)
+        {
+            _mediaBrowserService = mediaBrowserService;
+        }
+
+        public override string Link
+        {
+            get { return "http://mediabrowser.tv/"; }
+        }
+
+        public override void OnGrab(String message)
+        {
+            const string title = "Sonarr - Grabbed";
+
+            if (Settings.Notify)
+            {
+                _mediaBrowserService.Notify(Settings, title, message);
+            }
+        }
+
+        public override void OnDownload(DownloadMessage message)
+        {
+            const string title = "Sonarr - Downloaded";
+
+            if (Settings.Notify)
+            {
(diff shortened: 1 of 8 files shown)


src/NzbDrone.Core/Notifications/MediaBrowser/MediaBrowser.cs
```diff
@@ -0,0 +1,66 @@
+﻿using System;
+using System.Collections.Generic;
+using FluentValidation.Results;
+using NzbDrone.Common;
+using NzbDrone.Common.Extensions;
+using NzbDrone.Core.Tv;
+
+namespace NzbDrone.Core.Notifications.MediaBrowser
+{
+    public class MediaBrowser : NotificationBase<MediaBrowserSettings>
```
markus101: Any particular reason why? Currently we use the class name as the name when selection the type, so this would need to change or look bad, but I don't think re-factoring all that is warranted currently.

src/NzbDrone.Core/Notifications/MediaBrowser/MediaBrowserService.cs
```diff
@@ -0,0 +1,62 @@
...
+            {
+                if (ex.Response.StatusCode == HttpStatusCode.Unauthorized)
+                {
+                    return new ValidationFailure("ApiKey", "API Key is incorrect");
+                }
+            }
+            catch (Exception ex)
+            {
+                _logger.ErrorException("Unable to send test message: " + ex.Message, ex);
+                return new ValidationFailure("Host", "Unable to send test message");
```
markus101: Agreed, I'll send back the exception message.

src/NzbDrone.Core/Notifications/MediaBrowser/MediaBrowserProxy.cs
```diff
@@ -0,0 +1,71 @@
...
+            _logger.Trace("Response: {0}", response.Content);
+
+            CheckForError(response);
+
+            return response.Content;
+        }
+
+        private HttpRequest BuildRequest(string path, MediaBrowserSettings settings)
+        {
+            var url = String.Format(@"http://{0}/mediabrowser", settings.Address);
```
markus101: You need a valid URI to build the request, and need a request to build the segment. Chicken vs Egg.

src/NzbDrone.Core/Notifications/MediaBrowser/MediaBrowser.cs
kayone: It's just too generic. Honestly I don't like the `NzbGet`, `Sabnzbd` classes either.

markus101: Fair enough. This requires some other architectural changes, which I agree is a good idea (added to Trello: https://trello.com/c/2AW7y73n/912-be-more-explicit-with-class-names-for-3rd-party-integrations).

I'd like to get this merged in and we can make those other changes in another PR.

PR #162, opened 2014-12-26
Now generating unique ids for queue items sent to the api

Taloth:
To get rid of the id issues for the api.
The generator will create a new id for every key, and will preserve that id for 24h.

### DIFF
--- a/src/NzbDrone.Core/Download/Pending/PendingReleaseService.cs
+++ b/src/NzbDrone.Core/Download/Pending/PendingReleaseService.cs
@@ -2,6 +2,7 @@
 using System.Collections.Generic;
 using System.Linq;
 using NLog;
+using NzbDrone.Common.Crypto;
 using NzbDrone.Common.Extensions;
 using NzbDrone.Core.Configuration;
 using NzbDrone.Core.DecisionEngine;
@@ -101,7 +102,7 @@ namespace NzbDrone.Core.Download.Pending
 
                     var queue = new Queue.Queue
                                 {
-                                    Id = episode.Id ^ (pendingRelease.Id << 16),
+                                    Id = HashConverter.GetHashInt31(string.Format("pending-{0}-ep{1}", pendingRelease.Id, episode.Id)),
                                     Series = pendingRelease.RemoteEpisode.Series,
                                     Episode = episode,
                                     Quality = pendingRelease.RemoteEpisode.ParsedEpisodeInfo.Quality,
--- a/src/NzbDrone.Common/Crypto/HashConverter.cs
+++ b/src/NzbDrone.Common/Crypto/HashConverter.cs
@@ -1,0 +1,20 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Security.Cryptography;
+using System.Text;
+
+namespace NzbDrone.Common.Crypto
+{
+    public static class HashConverter
+    {
+        private static SHA1 HashAlgorithm = SHA1.Create();
+
+        public static int GetHashInt31(string target)
+        {
+            var hash = HashAlgorithm.ComputeHash(Encoding.Default.GetBytes(target));
+
+            return BitConverter.ToInt32(hash, 0) & 0x7fffffff;
+        }
+    }
+}
--- a/src/NzbDrone.Common/NzbDrone.Common.csproj
+++ b/src/NzbDrone.Common/NzbDrone.Common.csproj
@@ -132,6 +132,7 @@
     <Compile Include="Expansive\TreeNodeList.cs" />
     <Compile Include="Extensions\Base64Extentions.cs" />
     <Compile Include="Extensions\DateTimeExtensions.cs" />
+    <Compile Include="Crypto\HashConverter.cs" />
     <Compile Include="Extensions\Int64Extensions.cs" />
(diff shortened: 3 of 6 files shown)


src/NzbDrone.Common/UniqueIdentifier/UniqueIdentifierGenerator.cs
```diff
@@ -0,0 +1,50 @@
...
+        }
+
+        public int Get(string key)
+        {
+            return _cache.Get(key, GetNextIdentifier, TimeSpan.FromHours(24));
+        }
+
+        private int GetNextIdentifier()
+        {
+            var nextId = Interlocked.Increment(ref _lastIdentifier);
```
kayone: prob a stupid question, but why not just use a sequential integer? does this have to be unique across restarts? then can't we just append the process ID?

src/NzbDrone.Common/UniqueIdentifier/UniqueIdentifierGenerator.cs
```diff
@@ -0,0 +1,50 @@
...
+    public interface IUniqueIdentifierGenerator
+    {
+        int Get(string key);
+    }
+
+    public class UniqueIdentifierGenerator : IUniqueIdentifierGenerator
+    {
+        private static int _lastIdentifier;
+
+        private ICached<int> _cache;
```
kayone: readonly

src/NzbDrone.Core/Download/Pending/PendingReleaseService.cs
```diff
@@ -101,7 +105,7 @@ public List<Queue.Queue> GetPendingQueue()
 
                     var queue = new Queue.Queue
                                 {
-                                    Id = episode.Id ^ (pendingRelease.Id << 16),
+                                    Id = _uidProvider.Get(string.Format("pending-{0}-ep{1}", pendingRelease.Id , episode.Id)),
```
kayone: `p-{0}-{1}` just so the line is not as long.

/nitpicking.

src/NzbDrone.Common/UniqueIdentifier/UniqueIdentifierGenerator.cs
```diff
@@ -0,0 +1,50 @@
...
+
+        public UniqueIdentifierGenerator(ICacheManager cacheManager)
+        {
+            _cache = cacheManager.GetCache<int>(typeof(UniqueIdentifierGenerator));
+        }
+
+        public int Get(string key)
+        {
+            return _cache.Get(key, GetNextIdentifier, TimeSpan.FromHours(24));
```
kayone: wouldn't it be safer to convert the string to base 10? that way it'll always represent the same thing and we don't have to do this cache business.

src/NzbDrone.Core/Download/Pending/PendingReleaseService.cs
Taloth: I prefer the pending prefix, that way a future dev knows what it means, 'p' could be anything. It's just to avoid collisions between pending and CDH queue.

src/NzbDrone.Common/UniqueIdentifier/UniqueIdentifierGenerator.cs
Taloth: I don't think we have a big enough keyspace. The Downloadclient id stuff is a full string, convert that to base10 and you'll have a very big number, not valid JSON integer.
The UniqueIdentifierGenerator is intended to guarantee there won't be collisions on the UI.

kayone: what if you let it overflow? and then do `Math.Abs`

Taloth: An simple Math.Abs on the original code would've done that, but that doesn't prevent collisions.
The point of this PR was to end that once and for all. At least, not until 2 billion unique items have been shown on the UI.
I just wanted something that was guaranteed unique during the session.

kayone: if you do abs it still gives you  2,147,483,647 unique values without all this complication (no need to cache and they even stay the same across restarts)

kayone: Also, one thing to keep in mind, the lib we use for caching does support expiry but doesn't actually purge items (it uses the expiry to figure out if it should refetch the values). So setting the expiry to 24 hours in this case doesn't protect you against running out of memory.

Taloth:
Updated to hash.

src/NzbDrone.Common/UniqueIdentifier/UniqueIdentifierGenerator.cs
```diff
@@ -0,0 +1,50 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading;
+using NzbDrone.Common.Cache;
+
+namespace NzbDrone.Common.UniqueIdentifier
+{
+    public interface IUniqueIdentifierGenerator
```
kayone: Do we still need this?

kayone:
Looks good.

PR #167, opened 2015-01-06
fixed: changed Journaling for osx to prevent db corruption

kayone:
@Taloth @markus101


Taloth:
What about fullfsync: https://trello.com/c/ulbTdXu6/832-sqlite-on-osx-prone-to-corruption

kayone:
I would say we should do that as a last resort. My theory about WAL not working is because it uses some "fancy" (not really but see below)  file system features that might be miss behaving in osx, so using basic journaling should most likely fix the issue.

"WAL normally requires that the VFS support shared-memory primitives. (Exception: WAL without shared memory) The built-in unix and windows VFSes support this but third-party extension VFSes for custom operating systems might not."

PR #171, opened 2015-01-16
Remove animezb

markus101:
@kayone @Taloth 

Removing Animezb because its dead and has been for some time. Alternatively we can do this as a migration, I was torn so I went the non-migration route.


src/NzbDrone.Core/Indexers/IndexerFactory.cs
```diff
@@ -15,6 +15,8 @@ public interface IIndexerFactory : IProviderFactory<IIndexer, IndexerDefinition>
 
     public class IndexerFactory : ProviderFactory<IIndexer, IndexerDefinition>, IIndexerFactory
     {
+        private readonly IIndexerRepository _providerRepository;
```
Taloth: Why not just make the baseclass protected readonly?

markus101: We'd have to have the RemoveImplementations in the ProviderFactory, in this case its actually sufficient to remove it and `RemoveMissingImplementations()` will take care of it for us, but thats only because Animezb has a NullConfig, if it had a its own Config this method would fail since it couldn't deserialize from the DB. I'll clean this up and make a note to fix it up.

PR #172, opened 2015-01-16
Fixed: Better parsing of quality with no word boundaries.

kayone:
@markus101 @Taloth

### DIFF
--- a/src/NzbDrone.Core.Test/ParserTests/QualityParserFixture.cs
+++ b/src/NzbDrone.Core.Test/ParserTests/QualityParserFixture.cs
@@ -78,6 +78,7 @@ namespace NzbDrone.Core.Test.ParserTests
         [TestCase("Hannibal.S01E05.480p.BluRay.DD5.1.x264-HiSD", false)]
         [TestCase("Heidi Girl of the Alps (BD)(640x480(RAW) (BATCH 1) (1-13)", false)]
         [TestCase("[Doki] Clannad - 02 (848x480 XviD BD MP3) [95360783]", false)]
+        [TestCase("The.Simpsons.S25E21.Pay.Pal.1080p.WEB-DL.DD5.1.H.264-NTb-RP", false)]
         public void should_parse_dvd_quality(string title, bool proper)
         {
             ParseAndVerifyQuality(title, Quality.DVD, proper);
--- a/src/NzbDrone.Core.Test/ParserTests/QualityParserFixture.cs
+++ b/src/NzbDrone.Core.Test/ParserTests/QualityParserFixture.cs
@@ -184,6 +184,7 @@ namespace NzbDrone.Core.Test.ParserTests
         [TestCase("[Zurako] Log Horizon - 01 - The Apocalypse (BD 1080p AAC) [7AE12174].mkv", false)]
         [TestCase("WEEDS.S03E01-06.DUAL.1080p.Blu-ray.AC3.-HELLYWOOD.avi", false)]
         [TestCase("[Coalgirls]_Durarara!!_01_(1920x1080_Blu-ray_FLAC)_[8370CB8F].mkv", false)]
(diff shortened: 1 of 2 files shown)


src/NzbDrone.Core/Parser/QualityParser.cs
```diff
@@ -37,7 +37,7 @@ public class QualityParser
         private static readonly Regex RealRegex = new Regex(@"\b(?<real>)real\b",
                                                                 RegexOptions.Compiled | RegexOptions.IgnoreCase);
 
-        private static readonly Regex ResolutionRegex = new Regex(@"\b(?:(?<_480p>480p|640x480|848x480)|(?<_576p>576p)|(?<_720p>720p|1280x720)|(?<_1080p>1080p|1920x1080))\b",
+        private static readonly Regex ResolutionRegex = new Regex(@"(?:(?<_480p>480p|640x480|848x480|bdrip)|(?<_576p>576p)|(?<_720p>720p|1280x720)|(?<_1080p>1080p|1920x1080))\b",
```
Taloth: Already handled here: https://github.com/Sonarr/Sonarr/blob/quality-parsing-plus/src/NzbDrone.Core/Parser/QualityParser.cs#L143

kayone: the test was failing, I can take another look and see why

kayone: it's because before the source regex wasn't matching the `bd` in `bdrip`

I think it's more accurate now, since the source is actually Blu-Ray.

src/NzbDrone.Core/Parser/QualityParser.cs
```diff
@@ -18,11 +18,11 @@ public class QualityParser
                                                                 (?<hdtv>HDTV)|
                                                                 (?<bdrip>BDRiP)|
                                                                 (?<brrip>BRRip)|
-                                                                (?<dvd>DVD|DVDRip|NTSC|PAL|xvidvd)|
+                                                                (?<dvd>DVD|DVDRip|NTSC|\bPAL\b|xvidvd)|
```
Taloth: what about making it ```(?:(?-i)\bPAL\b)```, which enabled case-sensitivity.

kayone: That won't work. We to lower everything before parsing.

src/NzbDrone.Core.Test/ParserTests/QualityParserFixture.cs
```diff
@@ -78,6 +78,7 @@ public void should_parse_sdtv_quality(string title, bool proper)
         [TestCase("Hannibal.S01E05.480p.BluRay.DD5.1.x264-HiSD", false)]
         [TestCase("Heidi Girl of the Alps (BD)(640x480(RAW) (BATCH 1) (1-13)", false)]
         [TestCase("[Doki] Clannad - 02 (848x480 XviD BD MP3) [95360783]", false)]
+        [TestCase("The.Simpsons.S25E21.Pay.Pal.1080p.WEB-DL.DD5.1.H.264-NTb-RP", false)]
```
markus101: This is not DVD though, if anything this should be a failing WEB-DL test.

src/NzbDrone.Core/Parser/QualityParser.cs
markus101: These could all be word-bound.

src/NzbDrone.Core/Parser/QualityParser.cs
markus101: bdrip is not a resolution, it should not be here.

markus101:
The initial case for these changes was a badly named release and should be in the hall of shame, I don't think we should deal with it in this case. I think we should fix the glitch that parsed it at all, because we shouldn't match a number thats followed by more numbers, it leads to errors in most cases (and why we try to avoid it).

PR #174, opened 2015-01-17
Disable EZTV

markus101:
@kayone @Taloth I think this is the last thing we need to do before we merge.... finally.

### DIFF
--- a/src/NzbDrone.Core/Datastore/Migration/074_disable_eztv.cs
+++ b/src/NzbDrone.Core/Datastore/Migration/074_disable_eztv.cs
@@ -1,0 +1,14 @@
+﻿using FluentMigrator;
+using NzbDrone.Core.Datastore.Migration.Framework;
+
+namespace NzbDrone.Core.Datastore.Migration
+{
+    [Migration(74)]
+    public class disable_eztv : NzbDroneMigrationBase
+    {
+        protected override void MainDbUpgrade()
+        {
+            Execute.Sql("UPDATE Indexers SET EnableRss = 0, EnableSearch = 0 WHERE Implementation = 'Eztv' AND Settings LIKE '%ezrss.it%'");
+        }
+    }
+}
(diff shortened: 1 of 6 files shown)


src/NzbDrone.Core/Datastore/Migration/074_disable_eztv.cs
```diff
@@ -0,0 +1,14 @@
...
+using NzbDrone.Core.Datastore.Migration.Framework;
+
+namespace NzbDrone.Core.Datastore.Migration
+{
+    [Migration(74)]
+    public class disable_eztv : NzbDroneMigrationBase
+    {
+        protected override void MainDbUpgrade()
+        {
+            Execute.Sql("UPDATE Indexers SET EnableRss = 0, EnableSearch = 0 WHERE Implementation = 'Eztv' AND Settings LIKE '%ezrss.it/%'");
```
markus101: I went with handwritten inline SQL because the Update fluent syntax didn't appear to support the LIKE statement, plus we has tests.

kayone: `%ezrss.it%`?

kayone: also can we remove it instead of disabling it.

delphiactual:
http://rss.bt-chat.com/?group=3 is a mirror that appears to be working also they are still migrating to .ch from .it for ezrss

kayone:
@delphiactual the actual RSS url is 404ing,

http://rss.bt-chat.com/feed/

src/NzbDrone.Core/Datastore/Migration/074_disable_eztv.cs
markus101: Sure, to both, really doesn't matter, disabling seems less invasive

PR #175, opened 2015-01-18
Command queue

markus101:
@kayone @Taloth 

This will put queued jobs into a queue and execute them in order. Some cleanup is still required and the ability to mark commands as singular (no other commands of the same type can execute), so it can be utilized for an add series queue.

Bonus: https://trello.com/c/Qx8lCvNT/845-send-command-information-back-to-the-client comes free with this change since we store the full command, not just the base class.


src/NzbDrone.Api/Commands/CommandResource.cs
```diff
@@ -8,11 +8,73 @@ public class CommandResource : RestResource
...
+        public CommandStatus Status { get; set; }
+        public DateTime Queued { get; set; }
+        public DateTime? Started { get; set; }
+        public DateTime? Ended { get; set; }
+        public TimeSpan? Duration { get; set; }
+        public string Exception { get; set; }
+        public CommandTrigger Trigger { get; set; }
+
+        //Legacy
+        public CommandStatus State
```
markus101: Not sure if we can about keeping these, not sure if we've exposed this functionality to any 3rd party devs (nzbToMedia?)

src/NzbDrone.Core.Test/Messaging/Commands/CommandExecutorFixture.cs
```diff
@@ -1,121 +1,121 @@
...
-    {
-
-        public CommandB()
-        {
-        }
-    }
-
-}
\ No newline at end of file
+﻿//using System;
```
markus101: Need some direction on how to handle these tests, currently they expect to execute when called, but the behavior has changed completely.

src/NzbDrone.Core/Datastore/Converters/CommandConverter.cs
```diff
@@ -0,0 +1,39 @@
...
+            var stringValue = (string)context.DbValue;
+
+            if (string.IsNullOrWhiteSpace(stringValue))
+            {
+                return NullConfig.Instance;
+            }
+
+            var ordinal = context.DataRecord.GetOrdinal("Name");
+            var contract = context.DataRecord.GetString(ordinal);
+            var impType = typeof (Command).Assembly.FindTypeByName(contract + "Command");
```
markus101: This is because Command names are stored without Command (mainly so the UI doesn't see it when we send down  updated messages). We can handle it in the UI instead though, just say the word.

src/NzbDrone.Core/MediaFiles/MediaFileTableCleanupService.cs
```diff
@@ -3,19 +3,20 @@
...
 
 namespace NzbDrone.Core.MediaFiles
 {
+    public interface IMediaFileTableCleanupService
+    {
+        void Clean(Series series);
+    }
 
-    public class MediaFileTableCleanupService : IExecute<CleanMediaFileDb>
+    public class MediaFileTableCleanupService : IMediaFileTableCleanupService
```
markus101: This was the only command that wanted to be run Synchronous so I killed it.

src/NzbDrone.Core/Messaging/Commands/CommandExecutor.cs
```diff
@@ -1,117 +1,65 @@
...
         private readonly Logger _logger;
         private readonly IServiceFactory _serviceFactory;
-        private readonly ITrackCommands _trackCommands;
+        private readonly ICommandService _commandService;
         private readonly IEventAggregator _eventAggregator;
         private readonly TaskFactory _taskFactory;
 
-        public CommandExecutor(Logger logger, IServiceFactory serviceFactory, ITrackCommands trackCommands, IEventAggregator eventAggregator)
+        private static CancellationTokenSource _cancellationTokenSource;
+        private const int THREAD_LIMIT = 1;
```
markus101: At 1 for testing (debugging, otherwise its a PITA), previously we allowed 3, which seems like an okay number to stick with, thoughts?

src/NzbDrone.Core/Messaging/Commands/CommandService.cs
```diff
@@ -0,0 +1,189 @@
...
+                              IEventAggregator eventAggregator,
+                              ICacheManager cacheManager,
+                              Logger logger)
+        {
+            _repo = repo;
+            _serviceFactory = serviceFactory;
+            _eventAggregator = eventAggregator;
+            _logger = logger;
+
+            _messageCache = cacheManager.GetCache<string>(GetType());
```
markus101: Using a cache because storing it in the DB is a great way to DOS yourself when updating during an RSS Sync.

src/NzbDrone.Api/Commands/CommandResource.cs
Taloth: We should set up integration tests to check the api for known third-party usage.

src/NzbDrone.Core/Messaging/Commands/CommandExecutor.cs
Taloth: For disk IO you want one thread, for the individual indexer calls you want parallel, but we it's still one command.
Maybe make it a config option, or config file option? Default to 1.

src/NzbDrone.Core/Datastore/Converters/CommandConverter.cs
Taloth: no, leave it, from the UI perspective it's talking to a command endpoint, having a suffix there seems redundant.
Also protects against injection of stuff that isn't actually a Command.

src/NzbDrone.Core/Messaging/Commands/CommandService.cs
```diff
@@ -0,0 +1,189 @@
...
+        }
+
+        public CommandModel PublishCommand<TCommand>(TCommand command) where TCommand : Command
+        {
+            Ensure.That(command, () => command).IsNotNull();
+
+            _logger.Trace("Publishing {0}", command.GetType().Name);
+
+            var existingCommands = _repo.FindCommands(command.Name).Where(c => c.Status == CommandStatus.Queued ||
```
Taloth: ```lock (Mutex)```

src/NzbDrone.Core/Messaging/Commands/CommandExecutor.cs
markus101: Yeah, maybe we need to categorize them. I don't think 1 thread is good enough and making it customizable seems like option bloat.

Side effect of not running scheduled tasks is a long series refresh won't block imports and RSS syncs. :+1:

Taloth: Actually, CheckForFinishedDownload & Rss Sync is a command on it's own. Wouldn't that be run on 1 thread as well?

markus101: > Actually, CheckForFinishedDownload & Rss Sync is a command on it's own. Wouldn't that be run on 1 thread as well?

Not following exactly what you mean. Each command will be single threaded, but we need to avoid running multiple add series commands at the same time (maybe its a single job and has its own queue?)

src/NzbDrone.Core/Messaging/Commands/CommandService.cs
```diff
@@ -0,0 +1,193 @@
...
+        }
+
+        public CommandModel PublishCommand(string commandName)
+        {
+            var command = PublishCommand(commandName, null);
+
+            return command;
+        }
+
+        public void PublishScheduledTasks(List<ScheduledTask> scheduledTasks)
```
kayone: I don't think this belongs here. Should be part of the scheduler or whatever actually makes this call.

CommandService should not care where stuff comes from, it just adds stuff to the queue.

src/NzbDrone.Core/Messaging/Commands/CommandService.cs
```diff
@@ -0,0 +1,193 @@
...
+        {
+            _repo = repo;
+            _serviceFactory = serviceFactory;
+            _eventAggregator = eventAggregator;
+            _logger = logger;
+
+            _messageCache = cacheManager.GetCache<string>(GetType());
+        }
+
+        public CommandModel PublishCommand<TCommand>(TCommand command) where TCommand : Command
```
kayone: EnqueCommand

src/NzbDrone.Core/Messaging/Commands/CommandService.cs
```diff
@@ -0,0 +1,193 @@
...
+        {
+            Ensure.That(command, () => command).IsNotNull();
+
+            _logger.Trace("Publishing {0}", command.GetType().Name);
+
+            lock (Mutex)
+            {
+                var existingCommands = _repo.FindCommands(command.Name).Where(c => c.Status == CommandStatus.Queued ||
```
kayone: This should run in the DB, otherwise it'll get really heavy getting a list of all `CheckForCompletedDownloads` and filtering it in memory.

src/NzbDrone.Core/Messaging/Commands/CommandService.cs
```diff
@@ -0,0 +1,193 @@
...
+            {
+                PublishCommand(scheduledTask.TypeName, scheduledTask.LastExecution, CommandTrigger.Scheduled);
+            }
+        }
+
+        public CommandModel Pop()
+        {
+            lock (Mutex)
+            {
+                var nextCommand = _repo.Queued().OrderByDescending(c => c.Priority).ThenBy(c => c.Queued).FirstOrDefault();
```
kayone: should be done in repo. There is no point getting everything that is queue if we only need the first one.

src/NzbDrone.Core/Messaging/Commands/CommandRepository.cs
```diff
@@ -0,0 +1,56 @@
...
+            : base(database, eventAggregator)
+        {
+        }
+
+        public void Clean()
+        {
+            Delete(c => c.Ended < DateTime.UtcNow.AddDays(-1));
+        }
+
+        public void UpdateAborted()
```
kayone: `OrphanStarted`

src/NzbDrone.Core/Messaging/Commands/CommandRepository.cs
```diff
@@ -0,0 +1,56 @@
...
+        {
+            var aborted = Query.Where(c => c.Status == CommandStatus.Started).ToList();
+
+            foreach (var command in aborted)
+            {
+                command.Status = CommandStatus.Aborted;
```
kayone: I think we need a new orphan status.

src/NzbDrone.Core/Datastore/Migration/075_add_commands_table.cs
```diff
@@ -0,0 +1,24 @@
...
+    public class add_commands_table : NzbDroneMigrationBase
+    {
+        protected override void MainDbUpgrade()
+        {
+            Create.TableForModel("Commands")
+                  .WithColumn("Name").AsString().NotNullable()
+                  .WithColumn("Body").AsString().NotNullable()
+                  .WithColumn("Priority").AsInt32().NotNullable()
+                  .WithColumn("Status").AsInt32().NotNullable()
+                  .WithColumn("Queued").AsDateTime().NotNullable()
```
kayone: `QueuedOn` ?

src/NzbDrone.Core/Datastore/Migration/075_add_commands_table.cs
```diff
@@ -0,0 +1,24 @@
...
+                  .WithColumn("Started").AsDateTime().Nullable()
+                  .WithColumn("Ended").AsDateTime().Nullable()
```
kayone: `QueuedOn` `StartedOn` `EndedOn` ?

src/NzbDrone.Core/Download/RedownloadFailedDownloadService.cs
```diff
@@ -58,7 +61,7 @@ public void HandleAsync(DownloadFailedEvent message)
 
             _logger.Debug("Failed download contains multiple episodes, probably a double episode, searching again");
 
-            _commandExecutor.PublishCommandAsync(new EpisodeSearchCommand(message.EpisodeIds));
+            _commandService.PublishCommand(new EpisodeSearchCommand(message.EpisodeIds));
```
kayone: I think `CommandQueue` might be a better name or `CommandQueueManager`

src/NzbDrone.Core/Datastore/Migration/075_add_commands_table.cs
markus101: Any particular reason why? `QueuedAt` `StartedAt` `EndedAt` are more time-based, "On" signifies on a day to me.

src/NzbDrone.Core/Messaging/Commands/CommandRepository.cs
markus101: Whats the difference between Orphan and Aborted?

src/NzbDrone.Core/Messaging/Commands/CommandService.cs
markus101: Yeah, I had a reason why before, but I pulled out that code, will do.

src/NzbDrone.Core/Messaging/Commands/CommandExecutor.cs
kayone: I think we can leave it for future. also, 3 threads working on disk isn't that bad.

src/NzbDrone.Core/Messaging/Commands/CommandExecutor.cs
```diff
@@ -1,117 +1,65 @@
...
+                    {
+                        ExecuteCommand((dynamic)command.Body, command);
+                    }
+                    catch (Exception ex)
+                    {
+                        _logger.ErrorException("Error occurred while executing task " + command.Name, ex);
+                    }
+                }
+                
+                Thread.Sleep(50);
```
kayone: put this in the else.

src/NzbDrone.Core/Messaging/Commands/CommandExecutor.cs
```diff
@@ -120,43 +68,74 @@ private dynamic GetCommand(string commandTypeName)
...
+        //       Please note that the issue only shows itself on mono because since Microsoft .net implementation supports Task inlining on WaitAll.
+        public void Handle(ApplicationStartedEvent message)
+        {
+            _cancellationTokenSource = new CancellationTokenSource();
+
+            if (Enum.IsDefined(typeof(TaskCreationOptions), (TaskCreationOptions)0x10))
+            {
+                for (int i = 0; i < THREAD_LIMIT; i++)
+                {
+                    _taskFactory.StartNew(ExecuteCommands, TaskCreationOptions.PreferFairness | (TaskCreationOptions)0x10 | TaskCreationOptions.LongRunning)
```
kayone: should use actual threads instead of `Tasks`. Task has a mind of it's own and might not start the threads when you think it is.

src/NzbDrone.Core/Download/RedownloadFailedDownloadService.cs
markus101: Ewwww manager? I agree service isn't good here, I think Manager actually fits in this case though.

src/NzbDrone.Core/Messaging/Commands/CommandRepository.cs
kayone: aborted could be used when we support canceling an in-progress command.

src/NzbDrone.Core/Datastore/Migration/075_add_commands_table.cs
kayone: sure At works.

src/NzbDrone.Core/Messaging/Commands/CommandExecutor.cs
markus101: Ewww threads, I was somewhat basing my decision on: http://stackoverflow.com/questions/13429129/task-vs-thread-differencesn but also that I didn't need to change things too much. Re-reading that I see an advantage to Threads in that we can detect them being aborted and re-run them. That said, using the LongRunning flag means it gets executed on its own thread, outside of the thread pool.

kayone: Its not a grantee, its just a hint.
"It provides a hint to the TaskScheduler that oversubscription may be warranted."


https://msdn.microsoft.com/en-us/library/system.threading.tasks.taskcreationoptions(v=vs.110).aspx

also, them dying is a big issue. if one of the threads gets aborted because of an exception, you are out that thread for the life of the app, you kill 3 threads and sonarr pretty much becomes a zombie and does nothing.

src/NzbDrone.Core/Messaging/Commands/CommandRepository.cs
markus101: Wouldn't that be Cancelled?

markus101:
@kayone this one needs another set of eyes as well.

It occurred to me though, that we don't have a great way to ensure a command is run on its own, namely the update command. If that command runs when another has been running (like refresh series), its possible that we restart mid-operation to apply the update, the DB should be fine given the protection in SQLite (except for OS X :unamused:) but it also means you could be in the middle of a big refresh and it gets killed.

kayone:
That's not a new issue, it could happen right now as well, or anytime someone clicks on the restart button.

PR #177, opened 2015-01-18
Rename gulpFile.js to gulpfile.js

scottrobertson:
Otherwise you get:

```
gulp watch
Error: Cannot find module './gulp/gulpfile.js'
```


kayone:
Thanks!

PR #178, opened 2015-01-19
Fixed: Remove on Activity page should now work for Blackhole items.

Taloth:


markus101:
Word of warning, if your merge message and the commit message are the same, you will get double messages in the change log (we should fix that, but still :) )

markus101:
Looks good, but we should merge the two blackhole clients, only difference is the Protocol and the Download method.

Taloth:
I wanted to keep the two blackhole separate coz their behavior might be different when it comes to seeding and hardlinks.
You also would have to add new code to make the Protocol based on settings if you merged em.

PR #179, opened 2015-01-20
Title mismatches

markus101:
@kayone @Taloth 

This fixes a regression that came out of @kayone's refactoring of Download Tracking and a weird bug where we were clearing out completed downloads even when CDH was disabled and subsequently disabled, which affected some old users that haven't switched to CDH yet.


src/NzbDrone.Core/Download/CompletedDownloadService.cs
```diff
@@ -23,25 +24,37 @@ public class CompletedDownloadService : ICompletedDownloadService
...
         {
             if (trackedDownload.DownloadItem.Status != DownloadItemStatus.Completed)
             {
                 return;
             }
 
+            if (ignoreWarnings)
+            {
+                _logger.Info("Skipping ");
+                Import(trackedDownload);
```
kayone: doesn't this cause it to import twice, once here and once at the end?

src/NzbDrone.Core/Download/TrackedDownloads/DownloadMonitoringService.cs
```diff
@@ -78,7 +78,7 @@ private List<TrackedDownload> ProcessClientDownloads(IDownloadClient downloadCli
                 trackedDownloads.AddRange(newItems);
             }
 
-            if (_configService.RemoveCompletedDownloads)
+            if (_configService.EnableCompletedDownloadHandling && _configService.RemoveCompletedDownloads)
```
kayone: pff

src/NzbDrone.Core/Download/CompletedDownloadService.cs
markus101: Yeah, it would, I shall follow it with a return.

kayone: I think you should wrap all the check for warnings in this `if`, instead of short circuiting it,

src/NzbDrone.Core/MediaFiles/DownloadedEpisodesImportService.cs
```diff
@@ -31,23 +30,23 @@ public class DownloadedEpisodesImportService : IDownloadedEpisodesImportService
...
+                                            {
+                                                _diskProvider = diskProvider;
+                                                _diskScanService = diskScanService;
+                                                _seriesService = seriesService;
+                                                _parsingService = parsingService;
+                                                _importDecisionMaker = importDecisionMaker;
+                                                _importApprovedEpisodes = importApprovedEpisodes;
+                                                _sampleService = sampleService;
+                                                _logger = logger;
+                                            }
```
Taloth: Wrong indentation

markus101: WOAH! yeah... will fix.

PR #180, opened 2015-01-21
Updatecheck fixes

Taloth:
@markus101 Haven't actually tested this integrated yet, but just wanted put it in a PR for some feedback.

Started out with some minor changes, but a check for writable folders were sprinkled all over the place, so put that in DiskProvider.
On linux I could probably check for permissions, but not on windows and not on unc and not on.... so I decided to just keep the 'write a file' check.

The last commit is about making the user aware the manual triggered install failed. We still don't report if the Update app failed, but at least we can check if the folder is writable and such.

I also intend to move the branch switch to InstallUpdate, makes more sense there.


src/NzbDrone.Core/Update/InstallUpdateService.cs
```diff
@@ -170,7 +182,12 @@ public void Execute(ApplicationUpdateCommand message)
 
         public void Execute(InstallUpdateCommand message)
         {
-            InstallUpdate(message.UpdatePackage);
+            var success = InstallUpdate(message.UpdatePackage);
+
+            if (!success)
+            {
+                throw new NzbDroneClientException(System.Net.HttpStatusCode.Conflict, "Failed to install update");
```
markus101: Not sure if conflict is the right status code here, but offhand I can't think of a better one.

src/NzbDrone.Core/Update/UpdateCheckService.cs
```diff
@@ -30,13 +31,13 @@ public class CheckUpdateService : ICheckUpdateService
 
         public UpdatePackage AvailableUpdate()
         {
+            var latestAvailable = _updatePackageProvider.GetLatestUpdate(_configFileProvider.Branch, BuildInfo.Version);
+
             if (OsInfo.IsNotWindows && !_configFileProvider.UpdateAutomatically)
```
markus101: I was looking at this the other day and thought it was weird, this means people without auto update on don't get counted, doesn't it?

Taloth: Exactly, so I swapped it around.

src/NzbDrone.Core/Update/InstallUpdateService.cs
Taloth: It's a hack imo, we need a better way to report command failures to the UI.

src/NzbDrone.Core/Update/UpdateCheckService.cs
markus101: Okay good, I wasn't going crazy, which means we also might have a lot more users... on old versions.

src/NzbDrone.Common/Disk/DiskProviderBase.cs
```diff
@@ -109,6 +109,23 @@ public bool FileExists(string path, StringComparison stringComparison)
             }
         }
 
+        public bool FolderWritable(string path)
+        {
+            Ensure.That(path, () => path).IsValidPath();
+
+            try
+            {
+                var testPath = Path.Combine(path, "drone_test.txt");
```
kayone: `sonarr_write_test.txt`

src/NzbDrone.Common/Disk/DiskProviderBase.cs
```diff
@@ -109,6 +109,23 @@ public bool FileExists(string path, StringComparison stringComparison)
...
+                File.WriteAllText(testPath, DateTime.Now.ToString());
```
kayone: I think we should make the text:
`this file was created to verify if {path} is writable. It should've been automatically deleted. Feel free to delete it.`

src/NzbDrone.Core/Update/InstallUpdateService.cs
markus101: Right now we do it through command progress messaging, I'm not sure that this exception will even make it out of the command executor.

src/NzbDrone.Core/Update/InstallUpdateService.cs
```diff
@@ -61,12 +62,20 @@ public class InstallUpdateService : IExecute<ApplicationUpdateCommand>, IExecute
...
         {
             try
             {
                 EnsureAppDataSafety();
 
+                if (OsInfo.IsWindows || _configFileProvider.UpdateMechanism != UpdateMechanism.Script)
+                {
+                    if (!_diskProvider.FolderWritable(_appFolderInfo.StartUpFolder))
+                    {
+                        throw new ApplicationException("Cannot install update because startup folder is not writable.");
```
kayone: include the path in the exception.

src/NzbDrone.Core/Update/UpdateCheckService.cs
kayone: which we will never know about, since they won't get this update ;)

kayone: We shall call them _The Lost Generation!_

src/NzbDrone.Core/Update/UpdateCheckService.cs
```diff
@@ -46,7 +47,7 @@ public UpdatePackage AvailableUpdate()
                 try
                 {
                     _logger.Info("Branch [{0}] is being redirected to [{1}]]", _configFileProvider.Branch, latestAvailable.Branch);
-                    var config = _configFileProvider.GetConfigDictionary();
+                    var config = new Dictionary<string, object>();
                     config["Branch"] = latestAvailable.Branch;
```
kayone: we should write a unit-test for `ConfigFileProvider` to make sure it will do what this excepts it to do.

src/NzbDrone.Core/Update/UpdateCheckService.cs
markus101: maybe we will... if they manually update, one day.

Taloth:
Updated a few things.
Also moved the branch switch logic to InstallUpdate so the branch is only switched when a new version is installed instead only during the check.

src/NzbDrone.Common.Test/ConfigFileProviderTest.cs
```diff
@@ -141,8 +144,30 @@ public void SaveDictionary_should_save_proper_value()
...
+        }
+
+        [Test]
+        public void SaveDictionary_should_only_save_specified_values()
+        {
+            int port = 20555;
+            int origSslPort = 20551;
+            int sslPort = 20552;
+
+            WithPartialDiskProvider();
```
Taloth: A bit torn about this, I could also just setup the mock to return the correct Config xml code.

src/NzbDrone.Core/Update/InstallUpdateService.cs
```diff
@@ -108,13 +119,35 @@ private void InstallUpdate(UpdatePackage updatePackage)
...
             }
-            catch (Exception ex)
+            catch (UpdateFailedException ex)
             {
                 _logger.ErrorException("Update process failed", ex);
+                throw;
+            }
+        }
+
+        private void SwitchUpdatePackageBranch(UpdatePackage package)
```
kayone: The method name implies, switch is happening, we should either get rid of the if in this method, or call it `EnsureValidBranch` or something similar.

src/NzbDrone.Core/Update/InstallUpdateService.cs
```diff
@@ -108,13 +119,35 @@ private void InstallUpdate(UpdatePackage updatePackage)
...
                 _logger.Info("Starting update client {0}", _appFolderInfo.GetUpdateClientExePath());
-                _logger.ProgressInfo("NzbDrone will restart shortly.");
+                _logger.ProgressInfo("Sonarr will restart shortly.");
 
                 _processProvider.Start(_appFolderInfo.GetUpdateClientExePath(), GetUpdaterArgs(updateSandboxFolder));
+
+                return;
             }
-            catch (Exception ex)
+            catch (UpdateFailedException ex)
```
kayone: What if it fails with another exception? the error message below seems pretty generic so I think we should catch all exceptions.

src/NzbDrone.Core/Update/InstallUpdateService.cs
```diff
@@ -108,13 +119,35 @@ private void InstallUpdate(UpdatePackage updatePackage)
...
+                try
+                {
+                    _logger.Info("Branch [{0}] is being redirected to [{1}]]", _configFileProvider.Branch, package.Branch);
+                    var config = new Dictionary<string, object>();
+                    config["Branch"] = package.Branch;
+                    _configFileProvider.SaveConfigDictionary(config);
+                }
+                catch (Exception e)
+                {
+                    _logger.ErrorException("Couldn't save the branch redirect.", e);
```
kayone: "Couldn't save branch redirect. {from}>{to}"

src/NzbDrone.Core/Update/InstallUpdateService.cs
```diff
@@ -153,7 +186,29 @@ private void EnsureAppDataSafety()
             if (_appFolderInfo.StartUpFolder.IsParentPath(_appFolderInfo.AppDataFolder) ||
                 _appFolderInfo.StartUpFolder.PathEquals(_appFolderInfo.AppDataFolder))
             {
-                throw new NotSupportedException("Update will cause AppData to be deleted, correct you configuration before proceeding");
+                throw new UpdateFailedException("Update will cause AppData to be deleted, correct you configuration before proceeding");
```
kayone: "You sonarr configuration is being stored in Application folder which will cause data lost during the upgrade. Please remove any symlinks or redirects before trying again."

src/NzbDrone.Core/Update/InstallUpdateService.cs
```diff
@@ -164,13 +219,20 @@ public void Execute(ApplicationUpdateCommand message)
...
         }
 
         public void Execute(InstallUpdateCommand message)
         {
-            InstallUpdate(message.UpdatePackage);
+            var latestAvailable = _checkUpdateService.AvailableUpdate();
+
+            if (latestAvailable == null || latestAvailable.Hash != message.UpdatePackage.Hash)
+            {
+                throw new ApplicationException("Unknown or invalid update specified");
```
kayone: What does this mean? how can a user specify an update?

src/NzbDrone.Core/Update/InstallUpdateService.cs
```diff
@@ -153,7 +186,29 @@ private void EnsureAppDataSafety()
...
+
+                message.Completed("Restarting Sonarr to apply updates");
+            }
+            catch (UpdateFolderNotWritableException ex)
+            {
+                message.Failed(ex, string.Format("Startup folder not writable by user '{0}'", Environment.UserName));
+            }
+            catch (UpdateVerificationFailedException ex)
+            {
+                message.Failed(ex, "Update verification failed");
```
kayone: needs a better error message.

kayone:
pretty nice changes. 

I just feel like sometimes we try to be too brief in our error messages or tool-tips and we sacrifice them actually being useful. I would rather have a long error message that tells me why something happened and how I could fix it rather than a short one that only tells me something didn't work with no help on how to fix it.

src/NzbDrone.Core/Update/InstallUpdateService.cs
kayone: also, null update and invalid hash warrant different error messages.

src/NzbDrone.Core/Messaging/Commands/Command.cs
```diff
@@ -60,21 +60,21 @@ public void Start()
...
             _stopWatch.Stop();
             StateChangeTime = DateTime.UtcNow;
             State = CommandStatus.Failed;
             Exception = exception;
-            SetMessage("Failed");
+            SetMessage(message);
         }
 
-        public void Completed()
+        public void Completed(string message = "Completed")
```
markus101: :+1:

src/NzbDrone.Core/Update/InstallUpdateService.cs
markus101: Previously we accepted a POST of whatever was submitted, which could have resulted in a malicious package being installed.

If the update isn't the latest the messaging should say that, entirely possible that it happens due to the user sitting on the update page for a while and a new update is available (develop in particular).

src/NzbDrone.Core/Update/InstallUpdateService.cs
Taloth: Previously a generic Exception would be caught by InstallUpdate and logged as Error, and then it swallowed.
Right now I let it get thrown up the stack and caught in the CommandExecutor, which logs it as Error too. So I didn't see any use in catching Exception specifically.

The only difference is that I catch UpdateFailedException and log it as Error, then it gets thrown again, and caught by the Execute(Command command) which turns it into a specific Failure (expected failure) and doesn't actually throw anymore. So all UpdateFailedExceptions get swallowed, but any other are allowed to get thrown all the way up.

I could leave out the entire try catch here and log in ExecuteInstallUpdate instead, if that is preferable.

src/NzbDrone.Core/Update/InstallUpdateService.cs
Taloth: It can't be too long coz it's shown on the UI notification. "Downloaded update package is corrupt"?

src/NzbDrone.Core/Update/InstallUpdateService.cs
Taloth: Mentioned this on Hangouts. We took the package sent back from the UI, including url and hash, which would allow anyone with access to the UI to upload any application they want and run it by submitting a package of their own.
This check simple verifies if the latest available update is the one the user clicked on and panics otherwise. It shouldn't happen in normal circumstances.

src/NzbDrone.Core/Update/InstallUpdateService.cs
kayone: your second paragraph sounds like two component that shouldn't know about each other know why to much. this method is catching a specific exception because you know it get swallowed somewhere else.

Taloth:
If we want longer explanations in the notification we should allow someone to click on it (or a button in the notifaction) which opens up a popup with a full explanation. But to make that useful we need some kind of knowledge-base api to supply additional details.

src/NzbDrone.Core/Update/InstallUpdateService.cs
kayone: Why can't it be too long, messenger doesn't seem to truncate it.

image

kayone:
why can't we just show the full message in the pop-up?

src/NzbDrone.Core/Update/InstallUpdateService.cs
Taloth: Should I split up the message? or just return on latestAvailable == null, or just always take latest and ignore the message.UpdatePackage entirely?

kayone: We should update the code so it just uses the version sent from UI and ignores the hash and url.

src/NzbDrone.Core/Update/InstallUpdateService.cs
```diff
@@ -153,7 +176,32 @@ private void EnsureAppDataSafety()
             if (_appFolderInfo.StartUpFolder.IsParentPath(_appFolderInfo.AppDataFolder) ||
                 _appFolderInfo.StartUpFolder.PathEquals(_appFolderInfo.AppDataFolder))
             {
-                throw new NotSupportedException("Update will cause AppData to be deleted, correct you configuration before proceeding");
+                throw new UpdateFailedException("You Sonarr configuration ('{0}') is being stored in application folder ('{1}') which will cause data lost during the upgrade. Please remove any symlinks or redirects before trying again.", _appFolderInfo.AppDataFolder, _appFolderInfo.StartUpFolder);
+            }
```
kayone: we usually don't wrap values in logs with `()` we either use `[]` or `''`

src/NzbDrone.Core/Download/DownloadClientBase.cs
```diff
@@ -102,22 +102,13 @@ protected ValidationFailure TestFolder(String folder, String propertyName, Boole
...
-                }
-                catch (Exception ex)
-                {
-                    _logger.ErrorException(ex.Message, ex);
-                    return new NzbDroneValidationFailure(propertyName, "Unable to write to folder")
-                    {
-                        DetailedDescription = "The folder you specified is not writable. Please verify the folder permissions for the user account that is used to execute NzbDrone."
-                    };
-                }
+                    DetailedDescription = "The folder you specified is not writable. Please verify the folder permissions for the user account that is used to execute NzbDrone."
```
kayone: "Sonarr"
also should include the user name like we do on other message.

PR #181, opened 2015-01-22
Fix extern import

Taloth:
@markus101 @kayone Haven't yet tested it in intergrated environment so it can't be merged but all tests pass.

I also changed the trackedDownloadService again from TrackingId to DownloadId hopefully not overlooking a  ```Find``` call. Using the TrackingId was to make items, but in the db the DownloadId has to be unique anyway to be effective so it was giving more headaches than it solved.

I basically took the old implementation and refactored it to ```ProcessPath```.

### DIFF
--- a/src/NzbDrone.Core/MediaFiles/Commands/DownloadedEpisodesScanCommand.cs
+++ b/src/NzbDrone.Core/MediaFiles/Commands/DownloadedEpisodesScanCommand.cs
@@ -14,5 +14,9 @@ namespace NzbDrone.Core.MediaFiles.Commands
         }
 
         public Boolean SendUpdates { get; set; }
+
+        // Properties used by third-party apps, do not modify.
+        public String Path { get; set; }
+        public String DownloadClientId { get; set; }
     }
 }
--- a/src/NzbDrone.Core/MediaFiles/DownloadedEpisodesCommandService.cs
+++ b/src/NzbDrone.Core/MediaFiles/DownloadedEpisodesCommandService.cs
@@ -4,7 +4,10 @@ using System.IO;
 using System.Linq;
 using NLog;
 using NzbDrone.Common.Disk;
+using NzbDrone.Common.Extensions;
 using NzbDrone.Core.Configuration;
+using NzbDrone.Core.Download;
+using NzbDrone.Core.Download.TrackedDownloads;
 using NzbDrone.Core.MediaFiles.Commands;
 using NzbDrone.Core.MediaFiles.EpisodeImport;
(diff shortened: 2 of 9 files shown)


src/NzbDrone.Core/MediaFiles/DownloadedEpisodesCommandService.cs
```diff
@@ -48,9 +54,49 @@ private List<ImportResult> ProcessDroneFactoryFolder()
...
+                _logger.Warn("Folder specified for import scan [{0}] doesn't exist.", message.Path);
+                return new List<ImportResult>();
+            }
+
+            if (message.DownloadClientId.IsNotNullOrWhiteSpace())
+            {
+
+                var trackedDownload = _trackedDownloadService.Find(message.DownloadClientId);
+
+                if (trackedDownload == null)
```
kayone: I would invert this if since the not null is the common case and else is the edge case.

src/NzbDrone.Core/MediaFiles/DownloadedEpisodesCommandService.cs
```diff
@@ -48,9 +54,49 @@ private List<ImportResult> ProcessDroneFactoryFolder()
...
+            if (!_diskProvider.FolderExists(message.Path))
+            {
+                _logger.Warn("Folder specified for import scan [{0}] doesn't exist.", message.Path);
+                return new List<ImportResult>();
+            }
+
+            if (message.DownloadClientId.IsNotNullOrWhiteSpace())
+            {
+
+                var trackedDownload = _trackedDownloadService.Find(message.DownloadClientId);
```
markus101: The ID needed here is:
`downloadClient.Id + "-" + downloadItem.DownloadId,`

I ran into this issue fixing the failed download handling so it would actually delete from the download client.

TL/DR: its always going to be null

src/NzbDrone.Core/MediaFiles/Commands/DownloadedEpisodesScanCommand.cs
```diff
@@ -14,5 +14,9 @@ public override bool SendUpdatesToClient
         }
 
         public Boolean SendUpdates { get; set; }
+
+        // Properties used by third-party apps, do not modify.
```
markus101: Yay, good call.

src/NzbDrone.Core/MediaFiles/DownloadedEpisodesCommandService.cs
Taloth: I removed TrackingId entirely.

PR #184, opened 2015-01-27
Import using folder name

markus101:
@kayone @Taloth 

This will use the folder name when its:

- Parsable
- Contains only one non-sample file
- Not a full season

This means we can clean up a lot of hashed release logic and we'll properly handle DirFixes.

### DIFF
--- a/src/NzbDrone.Core.Test/MediaFiles/EpisodeImport/ImportDecisionMakerFixture.cs
+++ b/src/NzbDrone.Core.Test/MediaFiles/EpisodeImport/ImportDecisionMakerFixture.cs
@@ -45,31 +45,20 @@ namespace NzbDrone.Core.Test.MediaFiles.EpisodeImport
             _fail2 = new Mock<IImportDecisionEngineSpecification>();
             _fail3 = new Mock<IImportDecisionEngineSpecification>();
 
-            _pass1.Setup(c => c.IsSatisfiedBy(It.IsAny<LocalEpisode>())).Returns(true);
-            _pass1.Setup(c => c.RejectionReason).Returns("_pass1");
+            _pass1.Setup(c => c.IsSatisfiedBy(It.IsAny<LocalEpisode>())).Returns(Decision.Accept());
+            _pass2.Setup(c => c.IsSatisfiedBy(It.IsAny<LocalEpisode>())).Returns(Decision.Accept());
+            _pass3.Setup(c => c.IsSatisfiedBy(It.IsAny<LocalEpisode>())).Returns(Decision.Accept());
 
-            _pass2.Setup(c => c.IsSatisfiedBy(It.IsAny<LocalEpisode>())).Returns(true);
-            _pass2.Setup(c => c.RejectionReason).Returns("_pass2");
+            _fail1.Setup(c => c.IsSatisfiedBy(It.IsAny<LocalEpisode>())).Returns(Decision.Reject("_fail1"));
+            _fail2.Setup(c => c.IsSatisfiedBy(It.IsAny<LocalEpisode>())).Returns(Decision.Reject("_fail2"));
+            _fail3.Setup(c => c.IsSatisfiedBy(It.IsAny<LocalEpisode>())).Returns(Decision.Reject("_fail3"));
 
-            _pass3.Setup(c => c.IsSatisfiedBy(It.IsAny<LocalEpisode>())).Returns(true);
-            _pass3.Setup(c => c.RejectionReason).Returns("_pass3");
-
-
-            _fail1.Setup(c => c.IsSatisfiedBy(It.IsAny<LocalEpisode>())).Returns(false);
-            _fail1.Setup(c => c.RejectionReason).Returns("_fail1");
-
-            _fail2.Setup(c => c.IsSatisfiedBy(It.IsAny<LocalEpisode>())).Returns(false);
-            _fail2.Setup(c => c.RejectionReason).Returns("_fail2");
-
-            _fail3.Setup(c => c.IsSatisfiedBy(It.IsAny<LocalEpisode>())).Returns(false);
-            _fail3.Setup(c => c.RejectionReason).Returns("_fail3");
-
-            _videoFiles = new List<string> { @"C:\Test\Unsorted\The.Office.S03E115.DVDRip.XviD-OSiTV.avi" };
             _series = Builder<Series>.CreateNew()
                                      .With(e => e.Profile = new Profile { Items = Qualities.QualityFixture.GetDefaultQualities() })
                                      .Build();
 
             _quality = new QualityModel(Quality.DVD);
+
             _localEpisode = new LocalEpisode
             { 
                 Series = _series,
@@ -78,17 +67,24 @@ namespace NzbDrone.Core.Test.MediaFiles.EpisodeImport
             };
 
             Mocker.GetMock<IParsingService>()
-                  .Setup(c => c.GetLocalEpisode(It.IsAny<String>(), It.IsAny<Series>(), It.IsAny<Boolean>()))
+                  .Setup(c => c.GetLocalEpisode(It.IsAny<String>(), It.IsAny<Series>(), It.IsAny<ParsedEpisodeInfo>(), It.IsAny<Boolean>()))
                   .Returns(_localEpisode);
 
-            Mocker.GetMock<IMediaFileService>()
-                .Setup(c => c.FilterExistingFiles(_videoFiles, It.IsAny<Series>()))
-                .Returns(_videoFiles);
+            GivenVideoFiles(new List<string> { @"C:\Test\Unsorted\The.Office.S03E115.DVDRip.XviD-OSiTV.avi".AsOsAgnostic() });
         }
 
         private void GivenSpecifications(params Mock<IImportDecisionEngineSpecification>[] mocks)
         {
-            Mocker.SetConstant<IEnumerable<IRejectWithReason>>(mocks.Select(c => c.Object));
+            Mocker.SetConstant(mocks.Select(c => c.Object));
+        }
+
+        private void GivenVideoFiles(IEnumerable<string> videoFiles)
+        {
+            _videoFiles = videoFiles.ToList();
+
+            Mocker.GetMock<IMediaFileService>()
+                .Setup(c => c.FilterExistingFiles(_videoFiles, It.IsAny<Series>()))
+                .Returns(_videoFiles);
         }
 
         [Test]
@@ -96,7 +92,7 @@ namespace NzbDrone.Core.Test.MediaFiles.EpisodeImport
         {
             GivenSpecifications(_pass1, _pass2, _pass3, _fail1, _fail2, _fail3);
 
-            Subject.GetImportDecisions(_videoFiles, new Series(), false);
+            Subject.GetImportDecisions(_videoFiles, new Series(), null, false);
 
             _fail1.Verify(c => c.IsSatisfiedBy(_localEpisode), Times.Once());
             _fail2.Verify(c => c.IsSatisfiedBy(_localEpisode), Times.Once());
@@ -111,7 +107,7 @@ namespace NzbDrone.Core.Test.MediaFiles.EpisodeImport
         {
             GivenSpecifications(_fail1);
 
-            var result = Subject.GetImportDecisions(_videoFiles, new Series(), false);
+            var result = Subject.GetImportDecisions(_videoFiles, new Series());
 
             result.Single().Approved.Should().BeFalse();
         }
@@ -121,7 +117,7 @@ namespace NzbDrone.Core.Test.MediaFiles.EpisodeImport
         {
             GivenSpecifications(_pass1, _fail1, _pass2, _pass3);
 
-            var result = Subject.GetImportDecisions(_videoFiles, new Series(), false);
+            var result = Subject.GetImportDecisions(_videoFiles, new Series());
 
             result.Single().Approved.Should().BeFalse();
         }
@@ -131,7 +127,7 @@ namespace NzbDrone.Core.Test.MediaFiles.EpisodeImport
         {
             GivenSpecifications(_pass1, _pass2, _pass3);
 
-            var result = Subject.GetImportDecisions(_videoFiles, new Series(), false);
+            var result = Subject.GetImportDecisions(_videoFiles, new Series());
 
             result.Single().Approved.Should().BeTrue();
         }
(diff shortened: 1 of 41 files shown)


kayone:
I thought this wasn't always a good idea. (remember "The Americans" episode)

src/NzbDrone.Core.Test/MediaFiles/EpisodeImport/ImportDecisionMakerFixture.cs
```diff
@@ -226,10 +233,100 @@ public void should_not_throw_if_episodes_are_not_found()
...
+        {
+            var videoFiles = new[]
+                             {
+                                 @"C:\Test\Unsorted\Series.Title.S01\S01E01.mkv".AsOsAgnostic(),
+                                 @"C:\Test\Unsorted\Series.Title.S01\S01E02.mkv".AsOsAgnostic(),
+                                 @"C:\Test\Unsorted\Series.Title.S01\S01E03.mkv".AsOsAgnostic()
+                             };
+
+            GivenSpecifications(_pass1);
+            GivenVideoFiles(videoFiles.ToList());
```
kayone: make this method take `IEnumarble` and call `.ToList()` in it. (would look nicer, less needy)

src/NzbDrone.Core.Test/MediaFiles/EpisodeImport/ImportDecisionMakerFixture.cs
```diff
@@ -226,10 +233,100 @@ public void should_not_throw_if_episodes_are_not_found()
...
+
+            Mocker.GetMock<IParsingService>()
+                  .Verify(c => c.GetLocalEpisode(It.IsAny<string>(), It.IsAny<Series>(), true, It.IsAny<ParsedEpisodeInfo>()), Times.Exactly(1));
+
+            Mocker.GetMock<IParsingService>()
+                  .Verify(c => c.GetLocalEpisode(It.IsAny<string>(), It.IsAny<Series>(), true, null), Times.Never());
+        }
+
+        [Test]
+        public void should_use_folder_when_only_one_video_file_and_a_sample()
```
kayone: I think we should also use the file name in the folder if file name is parable. so only use folder name if file name is not parseable/hash

markus101:
That is why its a good idea. The release name was fixed, but the file inside is wrong (its really what ever the release/folder name is), so we import to a random episode and nuke another file. I think worst case scenario here is it gets grabbed, replaces the same episode number that was grabbed and if its wrong its pretty easy to fix, the other way around its hard to track down what actually happened.

This also fixes us playing cat and mouse with hashed releases (abc.mkv, xyz.mkv or anything new thats thrown at us).

src/NzbDrone.Core.Test/MediaFiles/EpisodeImport/ImportDecisionMakerFixture.cs
markus101: That means we're missing out on any dir fix releases, exactly the scenario you saw with The Americans.

src/NzbDrone.Core/MediaFiles/EpisodeImport/ImportDecisionMaker.cs
```diff
@@ -15,7 +17,8 @@ namespace NzbDrone.Core.MediaFiles.EpisodeImport
 {
     public interface IMakeImportDecision
     {
-        List<ImportDecision> GetImportDecisions(List<String> videoFiles, Series series, bool sceneSource, QualityModel quality = null);
+        List<ImportDecision> GetImportDecisions(List<String> videoFiles, Series series);
+        List<ImportDecision> GetImportDecisions(List<String> videoFiles, Series series, bool sceneSource, ParsedEpisodeInfo folderInfo);
```
kayone: `ParsedEpisodeInfo` `folderInfo` I notice you use folder info in other places too.

maybe a better name for the class would be `ParsedPathInfo`?

kayone: also, swap the position of `ParsedEpisodeInfo` and `SceneSource`, flag should always be last. (you can use re-sharper to change the order)

markus101: Yeah, I was thinking the name was bad (but didn't want to make 1000 changes and have a nasty PR).

src/NzbDrone.Core/MediaFiles/EpisodeImport/ImportDecisionMaker.cs
```diff
@@ -32,6 +36,7 @@ public class ImportDecisionMaker : IMakeImportDecision
                                    IMediaFileService mediaFileService,
                                    IDiskProvider diskProvider,
                                    IVideoFileInfoReader videoFileInfoReader,
+                                   ISampleService sampleService,
```
kayone: kind of unrelated, but `IDetectSampleService`

src/NzbDrone.Core/MediaFiles/EpisodeImport/ImportDecisionMaker.cs
```diff
@@ -39,39 +44,46 @@ public class ImportDecisionMaker : IMakeImportDecision
...
+        }
+
+        public List<ImportDecision> GetImportDecisions(List<string> videoFiles, Series series, bool sceneSource, ParsedEpisodeInfo folderInfo)
         {
             var newFiles = _mediaFileService.FilterExistingFiles(videoFiles.ToList(), series);
 
             _logger.Debug("Analyzing {0}/{1} files.", newFiles.Count, videoFiles.Count());
 
-            return GetDecisions(newFiles, series, sceneSource, quality).ToList();
+            return GetDecisions(newFiles, series, sceneSource, folderInfo).ToList();
```
kayone: inline this method.

markus101:
Here is another reason why we should use the folder name: `/path/Sons of Liberty 2015 S01E02 The Uprising 720p HDTV x264-DHD/Sons.of.Liberty.2015.E02.The.Uprising.720p.HDTV.x264-DHD.mkv` - no season number for the file, so it was not imported.

src/NzbDrone.Core/MediaFiles/EpisodeImport/ImportDecisionMaker.cs
```diff
@@ -138,10 +151,59 @@ private string EvaluateSpec(IRejectWithReason spec, LocalEpisode localEpisode)
...
+                if (parseResult == null)
+                {
+                    return false;
+                }
+
+                var sample = _sampleService.IsSample(series, GetQuality(folderInfo, fileQuality, series), file, size, parseResult.SeasonNumber);
+
+                if (!sample)
+                {
+                    files.Add(file);
```
kayone: can't you just return true here?

src/NzbDrone.Core/MediaFiles/EpisodeImport/ImportDecisionMaker.cs
```diff
@@ -138,10 +151,59 @@ private string EvaluateSpec(IRejectWithReason spec, LocalEpisode localEpisode)
...
+                }
+            }
+
+            return files.Count == 1;
```
kayone: and return false here.

src/NzbDrone.Core/MediaFiles/EpisodeImport/ImportDecisionMaker.cs
markus101: No, could be multiple files, won't know if they're samples or not until we loop through them all.

src/NzbDrone.Core/MediaFiles/EpisodeImport/ImportDecisionMaker.cs
```diff
@@ -138,10 +151,59 @@ private string EvaluateSpec(IRejectWithReason spec, LocalEpisode localEpisode)
...
                 //e.Data.Add("parsed", remoteEpisode.ParsedEpisodeInfo.ToJson());
                 _logger.ErrorException("Couldn't evaluate decision on " + localEpisode.Path, e);
-                return string.Format("{0}: {1}", spec.GetType().Name, e.Message);
+                return String.Format("{0}: {1}", spec.GetType().Name, e.Message);
             }
 
             return null;
         }
+
+        private bool ParseFolder(List<string> videoFiles, Series series, ParsedEpisodeInfo folderInfo)
```
kayone: This doesn't parse the folder, If I understand the purpose of it correctly then it should be called `ShouldUseFolderName`?

markus101: Yeah, missed that when I refactored.

src/NzbDrone.Core/MediaFiles/EpisodeImport/ImportDecisionMaker.cs
```diff
@@ -39,39 +44,46 @@ public class ImportDecisionMaker : IMakeImportDecision
...
+            }
+
             foreach (var file in videoFiles)
             {
                 ImportDecision decision = null;
 
                 try
                 {
-                    var localEpisode = _parsingService.GetLocalEpisode(file, series, sceneSource);
+                    var localEpisode = _parsingService.GetLocalEpisode(file, series, sceneSource, parseFolder ? folderInfo : null);
```
kayone: it's already set to null, so you don't need this `if` anymore.

src/NzbDrone.Core/MediaFiles/EpisodeImport/ImportDecisionMaker.cs
```diff
@@ -138,10 +151,59 @@ private string EvaluateSpec(IRejectWithReason spec, LocalEpisode localEpisode)
...
+                if (!sample)
+                {
+                    files.Add(file);
+                }
+            }
+
+            return files.Count == 1;
+        }
+
+        private QualityModel GetQuality(ParsedEpisodeInfo folderInfo, QualityModel fileQuality, Series series)
```
kayone: ` Series series, ParsedEpisodeInfo folderInfo, QualityModel fileQuality`

src/NzbDrone.Core/MediaFiles/EpisodeImport/ImportDecisionMaker.cs
markus101: Oops, the setting of `folderInfo` to null is wrong, we want to use it later to make sure we use the best quality (in the case the episode doesn't have the quality, like BR season packs).

src/NzbDrone.Core/Parser/ParsingService.cs
```diff
@@ -13,7 +13,8 @@ namespace NzbDrone.Core.Parser
 {
     public interface IParsingService
     {
-        LocalEpisode GetLocalEpisode(string filename, Series series, bool sceneSource);
+        LocalEpisode GetLocalEpisode(string filename, Series series);
+        LocalEpisode GetLocalEpisode(string filename, Series series, bool sceneSource, ParsedEpisodeInfo folderInfo);
```
kayone: flag at the end.

src/NzbDrone.Core.Test/MediaFiles/EpisodeImport/ImportDecisionMakerFixture.cs
kayone: What I meant to say:

I think we should also use the file name if it is parse-able. And only use folder name if file name is not parseable or is a hash

src/NzbDrone.Core/MediaFiles/EpisodeImport/ImportDecisionMaker.cs
kayone: actually `IDetectSample`

kayone: oh, oh, ISampleSpecification

markus101: We already have a sample spec, which uses this "service" I extracted this because of where its used.

src/NzbDrone.Core.Test/MediaFiles/EpisodeImport/ImportDecisionMakerFixture.cs
kayone: but my scenario sonarr actually did the write thing, the file was correct and the dir was wrong, if the same thing would happen after this fix, I would get an incorrect episode overriding a correct one.

src/NzbDrone.Core/MediaFiles/EpisodeImport/ImportDecisionMaker.cs
kayone: Still, Sample Service is not a good name, `IDetectSample`

src/NzbDrone.Core.Test/MediaFiles/EpisodeImport/ImportDecisionMakerFixture.cs
markus101: I think this particular case was an exception, not the rule, every other occurrence has been a dir fix that corrected a misnumbered episode, extra character in the name (S01E143 instead of S01E14) or in the case of Castle a dir fix could have nuked the S07E13 release and fixed it to S07E13.

If I grabbed the S07E03 release and ended up nuking my S07E13 file I'd be in the same boat.

I don't think rejecting them and letting the user choose is right either, in most cases they'd just want to hit import anyways, but which would we choose to follow automatically when that button is pressed, the folder name or the file name? Feels like a 99% feature, 99% of releases would benefit from this (or otherwise be unaffected), the 1% that has issues would need some manual intervention.

markus101:
@Taloth and I came to a compromise, if the episode is not an episode listed in the folder name (assuming its parseable), we'll mark it with an error and not import it, at a later time we'll offer a "manual" import that will let the user choose the series/episode(s)/quality of the episode file and effectively force the import.

thezoggy:
in the scene world.. the 'release name' (which ends up as the folder name due to it being the nzbname) is what is reviewed and corrected. Because each release name must be unique on sites.

another scenario, just a typo in the filename.. hence never going to get 'fixed' by a proper/real.
`Shark.Tank.S03E14.720p.HDTV.x264-2HD` has the filename `shark.tank.s03e143.720p.hdtv.x264-2hd` due to typo..

thezoggy:
this wont handle dirfixes as releases have samples.. youd have to nuke the sample via sab i guess for this to work right

markus101:
Samples are ignored when doing the check for a single video file.

PR #186, opened 2015-01-29
Fixed: Sonarr now installs as sonarr instead of NzbDrone

kayone:
This updates the application name (in uninstall list) and the shortcuts to Sonarr instead of NzbDrone.

I tried installing this over an existing install and it just renames the old instance to Sonarr. It keep any old icon the user has so they could still search for borth Sonarr or NzbDrone in start menu and it would show up.

shot_150129_111741


setup/build.bat
```diff
@@ -1 +1,3 @@
+#SET BUILD_NUMBER=1
+#SET branch=devlop
```
Taloth: devlop?

markus101:
Looks good™

PR #187, opened 2015-01-30
Webpack

kayone:
Switching to WebPack has a few advantages.

First and the main one is getting rid of requirejs.
The 2nd one is we can switch from the AMD module format to commonjs format,

```
define(
    [
        'marionette'
    ], function (Marionette) {

        return Marionette.CompositeView.extend({
            template: 'Series/Index/EmptyTemplate'
        });
    });
```

becomes

```
var marionette = require('marionette');
module.exports = marionette.CompositeView.extend({
    template : 'Series/Index/EmptyTemplate'
});;
```

specially helps with modules that have tons of dependencies, 

it also has support for source maps so we can debug a deployed build like we do in dev.

and generally a much more flexible loader, It'll allow us to do better caching (you can break the file into components (settings, system, series, history etc)) and they get their own hash breaker/file automatically.


kayone:
Here is the diffrence between r.js and webpack

**two main advantages right off the bat, no more version cache breaking, everything is done using hashes, so backend changes won't unnecessarily invalidate the cache.
Every file gets a cache breaker (even main.js), so it's much safer. 
We don't have to load r.js anymore. save 20K there.
And as I mentioned before we can even break the 2MB main js into separate packs so app is loaded much fast and cache is invalidated less often. so potential savings are huge**

### R.JS
image


### Webpack
image

gulp/webpack.js
```diff
@@ -0,0 +1,14 @@
+var gulp = require('gulp');
+
+var run = require('gulp-run');
+
+
+gulp.task('webpack', function (cb) {
+  run('webpack -d').exec()
```
Taloth: Shouldn't you use something like gulp-webpack here?

package.json
```diff
@@ -26,10 +26,9 @@
...
     "gulp-print": "1.1.0",
     "gulp-replace": "0.5.0",
+    "gulp-run": "1.6.6",
     "gulp-wrap": "0.5.0",
-    "gulp-wrap-amd": "0.3.1",
     "jshint-stylish": "1.0.0",
-    "requirejs": "2.1.15",
     "run-sequence": "1.0.2",
     "streamqueue": "0.1.1"
   }
```
Taloth: Missing webpack/gulp-webpack here, I had to install webpack manually from npm to make it compile.
If you use gulp-webpack you can probably remove the gulp-run again.

src/NzbDrone.Api/Frontend/Mappers/StaticResourceMapper.cs
```diff
@@ -27,6 +27,7 @@ public override bool CanHandle(string resourceUrl)
         {
             return resourceUrl.StartsWith("/Content") ||
                    resourceUrl.EndsWith(".js") ||
+                   resourceUrl.EndsWith(".map") ||
```
Taloth: .map is for the sourcemap right, or are we talking about a different kind of map? Isn't a sourcemap embedded in the .js better?

src/UI/.idea/encodings.xml
```diff
@@ -1,5 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project version="4">
-  <component name="Encoding" useUTFGuessing="true" native2AsciiForPropertiesFiles="false" />
-</project>
-
+  <component name="Encoding" useUTFGuessing="true" native2AsciiForPropertiesFiles="false">
+    <file url="file://$PROJECT_DIR$/System/Logs/Files/LogFileModel.js" charset="UTF-8" />
```
Taloth: Adding this encoding probably wasn't intentional.

src/UI/Shims/jquery.js
```diff
@@ -0,0 +1,8 @@
+var jquery = require('../JsLibraries/jquery');
+
+require('../Instrumentation/StringFormat');
+
+window.$= jquery;
```
Taloth: Nitpick: space before =

webpack.config.js
```diff
@@ -0,0 +1,42 @@
...
+      'backbone'                : 'Shims/backbone',
+      'moment'                  : 'JsLibraries/moment',
+      'filesize'                : 'JsLibraries/filesize',
+      'handlebars'              : 'Shared/Shims/handlebars',
+      'handlebars.helpers'      : 'JsLibraries/handlebars.helpers',
+      'bootstrap'               : 'JsLibraries/bootstrap',
+      'backbone.deepmodel'      : 'JsLibraries/backbone.deep.model',
+      'backbone.pageable'       : 'JsLibraries/backbone.pageable',
+      'backbone-pageable'       : 'JsLibraries/backbone.pageable',
+      'backbone-pageable'       : 'JsLibraries/backbone.pageable',
```
Taloth: Duplicate with '-', and another one with  '.', how about picking one?

Taloth:
I'm not experienced at all at this, so I just skimmed over it. 
Overall it looks like an excellent idea, afaik we don't have too many outstanding UI changes. But if we do this, we have to do this in one go, rebasing this stuff would be absolute hell.

gulp/webpack.js
kayone: the watcher in webpack is much more efficient that if we used gulp and re-ran the whole thing.

src/NzbDrone.Api/Frontend/Mappers/StaticResourceMapper.cs
kayone: embedding the source map would kind of defeat it's purpose of trying to minimize the amount of data that needs to be sent to the client.

webpack.config.js
kayone: so there was a reason for this, depending on modules (3rd party) that depend on backbone.pagable and how they reference it. sometime they require it by `-` sometime by `.`

we can re-evalute this, but this is straight from the require js config.

kayone:
Yeah, specially the change that I haven't pushed to this branch, it touches EVERY single file, in a big way.

gulp/webpack.js
Taloth: from what I can see gulp-webpack has a 'watch: true' option.

src/NzbDrone.Api/Frontend/Mappers/StaticResourceMapper.cs
Taloth: You would only include the sourcemap in debug builds right?

kayone: source maps are in a different file, so we can have them included in the release builds as well, they don't get downloaded unless dev tools is open.

webpack.config.js
Taloth: I can understand the . and -, but - is list twice.
I understand it's straight from the require file, doesn't mean the duplicate shouldn't be fixed.

Taloth:
One problem: I tried it out, but if I change for example Router.js, webpack doesn't refresh the main.js, could be something wrong with my system, so please verify and lemme know.

webpack.config.js
kayone: yeah, I fixed the duplicate, I was just explaining the `-` and `.`

kayone:
no, it's not you, It used to work, it doesn't work for me here either. :(

PR #188, opened 2015-02-03
Updated zero.clipboard so it works with requirejs

Taloth:


markus101:
@Taloth do you have an example where 'Wrap Always' would be worse for these short arrays?

I think we need to be able to reformat any file at any time and have an expected result, so things stay consistent over time. In practice that might not be viable, but its a good goal. I think I'd rather have short arrays wrap than deal with opening brackets on the wrong line.

Taloth:
https://github.com/Sonarr/Sonarr/commit/2c35110bb01ecd79e71ad700619fa3bfba45da44#diff-554208e3076bedc8524b1b89f206833bR31

would be

```
        filterModes : {
            "all"      : [
                null,
                null
            ],
            "grabbed"  : [
                'eventType',
                '1'
            ],
            "imported" : [
                'eventType',
                '3'
            ],
            "failed"   : [
                'eventType',
                '4'
            ],
            "deleted"  : [
                'eventType',
                '5'
            ]
        },
 ```

kayone:
I can live with this, also makes adding stuff to arrays, and then the diffs after the change easier.

markus101:
Yeah, I think thats a better option, inflates some smaller arrays instead of requiring manual control of every new array.

Taloth:
Merged zeroclipboard change, will do the UI cleanup piece by piece.

PR #189, opened 2015-02-06
Add support for ImmortalSeed.me private tracker with tests

MichelZ:


Taloth:
Thnx for going through the effort of adding a test.

One thing however, this indexer doesn't seem to require any special handling of the RSS, so I rather roll it into a generic TorrentRssIndexer.
The Generic [TorrentRss](https://trello.com/c/uJAZt2WW/861-torrent-support-part-2) is intended to pull in the rss and inspect it to see if it contains the fields what we need, allowing the user to add a myriad of rss indexer feeds without us having to add support for every indexer out there. (which we won't)

So it might be a while till I get to making TorrentRssIndexer and adding the ImmortalSeed RSS as testcase.
But if you're interested in developing that I would be happy to explain the idea in more detail.

MichelZ:
Let me know the requirements, and I might get on it when I get some time

src/NzbDrone.Core/Indexers/ImmortalSeed/ImmortalSeedRequestGenerator.cs
```diff
@@ -0,0 +1,52 @@
...
+        }
+
+        public virtual IList<IEnumerable<IndexerRequest>> GetSearchRequests(SpecialEpisodeSearchCriteria searchCriteria)
+        {
+            return new List<IEnumerable<IndexerRequest>>();
+        }
+
+        private IEnumerable<IndexerRequest> GetRssRequests(String searchParameters)
+        {
+            yield return new IndexerRequest(String.Format("{0}", Settings.BaseUrl.Trim().TrimEnd('/')), HttpAccept.Rss);
```
kayone: don't really need a `String.Format` here.

Taloth:
We want the TorrentRssIndexer to check the rss contents during [Test](https://github.com/Sonarr/Sonarr/blob/develop/src/NzbDrone.Core/Indexers/HttpIndexerBase.cs#L203) so the user cannot configure sites that we don't support.

For example, if you check for the ezrss namespace, you can use the EzrssTorrentRssParser.
Otherwise you have to check if it contains the mandatory properties. (Name, size, publish date, download url, etc)
Sometimes the size has a separate element, sometimes it's embedded in the description.

Ltek:
Is this merged into the main branch?

MichelZ:
No, it is not.
The new PR with this functionality is here:
https://github.com/Sonarr/Sonarr/pull/193

Ltek:
I installed the latest Develop branch and it is not in that? How can I integrate your code? (sorry for the noob question, I'm a noob)

thx

MichelZ:
You would need to download the TorrentRssIndexer branch and compile it yourself.

Ltek:
I don't even see that in the branch drop-down list?

Any ETA on when it might be merged with Develop?

MichelZ:
You would need to look in my repo where the PR is coming from: https://github.com/MichelZ/Sonarr/tree/TorrentRssIndexer

PR #190, opened 2015-02-07
Command queue

markus101:
@kayone @Taloth

Take 2 on this one, turns out hitting the DB 60 times per second is bad (when waiting for work), 50ms sleep * 3 workers = 60 times/second. I have added a command queue which will be instead hit 60 times a second when waiting for work.

The alternative would be to terminate threads when there isn't work for them and spin them up when there is (a bit like what we did before) or have them poll less often, I think a bit less often would be suitable, 100-200ms, but any longer it would be noticeable by users queuing up commands (searches, etc).


delphiactual:
@markus101 when scanning drone factory, the toast no longer appears. I don't know if that is intentional or not.

kayone:
so, I wanted to suggest this latter, since I figured it won't be an issue but it seems like we were both wrong and it screwed us, so here was my suggestion,

when a new job gets add by a scheduler or the user it's added to the db but it's also added to a blocking collection, all worker threads should block on that collection for tasks instead of DB (I hope blocking collections support ordering etc, if not another in-memory structure).

when a task is done/failed etc, it updated in the db like it is now
when we start the app we load our collection from the db.

this should reduce our calls to the db significantly since we won't be pooling the db for jobs anymore.

src/NzbDrone.Api/Commands/CommandModule.cs
```diff
@@ -40,31 +36,31 @@ public class CommandModule : NzbDroneRestModuleWithSignalR<CommandResource, Comm
...
+            var commandType = 
+              _serviceFactory.GetImplementations(typeof(Command))
                         .Single(c => c.Name.Replace("Command", "")
                         .Equals(commandResource.Name, StringComparison.InvariantCultureIgnoreCase));
 
             dynamic command = Request.Body.FromJson(commandType);
-            command.Manual = true;
+            command.Trigger = CommandTrigger.Manual;
 
-            var trackedCommand = (Command)_commandExecutor.PublishCommandAsync(command);
```
kayone: don't need the cast.

src/NzbDrone.Api/Commands/CommandModule.cs
```diff
@@ -40,31 +36,31 @@ public class CommandModule : NzbDroneRestModuleWithSignalR<CommandResource, Comm
...
+            var trackedCommand = (CommandModel)_commandQueueManager.Push(command, CommandPriority.Normal, CommandTrigger.Manual);
             return trackedCommand.Id;
         }
 
         private List<CommandResource> GetAllCommands()
```
kayone: Rename this to `GetRunning`?

src/NzbDrone.Common/Cache/Cached.cs
```diff
@@ -100,6 +100,14 @@ public void Clear()
             _store.Clear();
         }
 
+        public void ClearExpired()
```
kayone: This was added to purse recently,
https://github.com/kayone/Purse/commit/a12f51bea15ff4f3ca3a9373e28db5cfd32f3f3d

src/NzbDrone.Common/Cache/ICached.cs
```diff
@@ -6,6 +6,7 @@ namespace NzbDrone.Common.Cache
     public interface ICached
     {
         void Clear();
+        void ClearExpired();
```
kayone: Why do we not use purse?

src/NzbDrone.Core/Datastore/Converters/CommandConverter.cs
```diff
@@ -0,0 +1,39 @@
...
+
+namespace NzbDrone.Core.Datastore.Converters
+{
+    public class CommandConverter : EmbeddedDocumentConverter
+    {
+        public override object FromDB(ConverterContext context)
+        {
+            if (context.DbValue == DBNull.Value)
+            {
+                return NullConfig.Instance;
```
kayone: What is happening here? `IProviderConfig` isn't convertible to `Command`

copy/past fail?

src/NzbDrone.Core/Datastore/Converters/TimeSpanConverter.cs
```diff
@@ -0,0 +1,52 @@
...
+
+namespace NzbDrone.Core.Datastore.Converters
+{
+    public class TimeSpanConverter : IConverter
+    {
+        public object FromDB(ConverterContext context)
+        {
+            if (context.DbValue == DBNull.Value)
+            {
+                return DBNull.Value;
```
kayone: this is gonna blow up, 

prob want `new TimeSpan()`

src/NzbDrone.Core/Datastore/Converters/TimeSpanConverter.cs
```diff
@@ -0,0 +1,52 @@
...
+            }
+
+            if (context.DbValue is TimeSpan)
```
kayone: this can't be true.

src/NzbDrone.Core/Datastore/Converters/TimeSpanConverter.cs
```diff
@@ -0,0 +1,52 @@
...
+        }
+
+        public object ToDB(object clrValue)
+        {
+            if (clrValue.ToString().IsNullOrWhiteSpace())
+            {
+                return null;
+            }
+
+            return clrValue.ToString();
```
kayone: use invariant culture here and during parse, otherwise it could blow up if a user changes their culture.

src/NzbDrone.Core/Housekeeping/HousekeepingService.cs
```diff
@@ -40,7 +41,7 @@ private void Clean()
                 }
             }
 
-            // Vacuuming the log db isn't needed since that's done hourly at the TrimLogCommand.
+            // Vacuuming the log db isn't needed since that's done in a separate housekeeping task
```
kayone: this comment doesn't apply anymore, since you are calling vacuum

src/NzbDrone.Common/Cache/ICached.cs
markus101: Things were never converted over.

src/NzbDrone.Core/Housekeeping/HousekeepingService.cs
markus101: Sure it does, its being called for the Log DB via a separate house keeping task, so we don't need to call it again here.

src/NzbDrone.Core/Messaging/Commands/CommandExecutor.cs
```diff
@@ -120,47 +62,69 @@ private dynamic GetCommand(string commandTypeName)
...
-                if (command.State == CommandStatus.Running)
-                {
-                    _trackCommands.Completed(command);
-                }
+                _commandQueueManager.SetMessage(commandModel, command.CompletionMessage);
+                _commandQueueManager.Complete(commandModel);
+            }
+            catch (CommandFailedException ex)
+            {
+                _commandQueueManager.SetMessage(commandModel, ex.Message);
```
kayone: `SetMessage` should automatically happen as part of the `Fail`, `Complete`

src/NzbDrone.Core/Messaging/Commands/CommandExecutor.cs
```diff
@@ -120,47 +62,69 @@ private dynamic GetCommand(string commandTypeName)
 
             try
             {
-                _trackCommands.Start(command);
-                BroadcastCommandUpdate(command);
+                _commandQueueManager.Start(commandModel);
+                BroadcastCommandUpdate(commandModel);
 
                 if (!MappedDiagnosticsContext.Contains("CommandId") && command.SendUpdatesToClient)
```
kayone: we should set this regardless of `SendUpdateToClient` we can use it later for grouping log messages

PR #192, opened 2015-02-10
Removed redundant LINQ call

SchadLucas:


markus101:
This change would cause all models to be loaded from the DB and then processed in memory because MarrDataMapper (The ORM we use) doesn't support SingleOrDefault() as a means of filtering.

This commit was when we realized it was a huge performance hit and switched it over. https://github.com/Sonarr/Sonarr/commit/78dacf68503b87ca114aade8b4eb77f880bb6d63

ReSharper warns about these double LINQ calls, but anything in a Repository is expected and can be ignored.

PR #193, opened 2015-02-12
Add generic TorrentRssIndexer support

MichelZ:
Add support for generic RSS feeds. Parses the feed and tests if it is
EZTV compatible, or if it has all required fields for the generic
TorrentRssParser

Please keep the comments coming... I'm not yet experienced with Sonarr and I'm probably doing some stuff sub-optimal :) But it's a start.


src/NzbDrone.Core/Parser/Model/TorrentInfo.cs
```diff
@@ -2,6 +2,8 @@
 
 namespace NzbDrone.Core.Parser.Model
 {
+    using System.Text;
```
kayone: using statements should be at the root

src/NzbDrone.Core.Test/IndexerTests/TorrentRssIndexerTests/TestTorrentRssIndexer.cs
```diff
@@ -0,0 +1,44 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
```
kayone: no space between using statements.

src/NzbDrone.Core.Test/IndexerTests/TorrentRssIndexerTests/TorrentRssIndexerFixture.cs
```diff
@@ -0,0 +1,123 @@
...
+        public void Setup()
+        {
+            Subject.Definition = new IndexerDefinition()
+            {
+                Name = "TorrentRssIndexer",
+                Settings = new TorrentRssIndexerSettings() { }
+            };
+        }
+
+        [TestCase("https://www.ezrss.it/", @"Files/RSS/Eztv.xml")]
```
kayone: 2nd argument should just be the filename, path is implied.

src/NzbDrone.Core.Test/IndexerTests/TorrentRssIndexerTests/TestTorrentRssIndexer.cs
MichelZ: Are there any sonarr specific ReSharper settings file that I could use?

src/NzbDrone.Core.Test/IndexerTests/TorrentRssIndexerTests/TorrentRssIndexerFixture.cs
```diff
@@ -0,0 +1,123 @@
...
+        [Test]
+        public void should_return_empty_list_on_404()
+        {
+            Mocker.GetMock<IHttpClient>()
+                .Setup(o => o.Execute(It.Is<HttpRequest>(v => v.Method == HttpMethod.GET)))
+                .Returns<HttpRequest>(r => new HttpResponse(r, new HttpHeader(), new Byte[0], System.Net.HttpStatusCode.NotFound));
+
+            var releases = Subject.FetchRecent();
+
+            releases.Should().HaveCount(0);
```
kayone: `releases.Should().BeEmpty();` better error message in case of failure ;)

src/NzbDrone.Core/Parser/Model/TorrentInfo.cs
```diff
@@ -19,5 +21,21 @@ public class TorrentInfo : ReleaseInfo
             }
             return torrentInfo.Seeders;
         }
+
+        public override string ToString(string format)
```
markus101: If we only have one case the format isn't needed.

markus101: Never mind I see where its coming from now.

src/NzbDrone.Core/Indexers/TorrentRssIndexer/TorrentRssIndexer.cs
```diff
@@ -0,0 +1,293 @@
...
+                throw new InvalidOperationException("You have to initialize ParserSettings first.");
+            }
+
+            if (ParserSettings.UseEZTVFormat)
+            {
+                return new EzrssTorrentRssParser();
+            }
+            else
+            {
+                return new TorrentRssParser() { UseGuidInfoUrl = false, ParseSeedersInDescription = ParserSettings.ParseSeedersInDescription, ParseSizeInDescription = ParserSettings.ParseSizeInDescription, SizeElementName = ParserSettings.SizeElementName };
```
kayone: add a line break for every property.

src/NzbDrone.Core/Indexers/TorrentRssIndexer/TorrentRssIndexerParserSettings.cs
```diff
@@ -0,0 +1,18 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace NzbDrone.Core.Indexers.TorrentRssIndexer
+{
+    public class TorrentRssIndexerParserSettings
```
kayone: I don't think settings (as in data like this) should be provided to the generic indexer, it should follow other indexer where you tell it what type/instance of parser it should use. the parser could accept settings, but not the indexer client itself.

this makes indexer too involved in parsing, and we don't like overly involved components in sonnar ;) we try to keep thing as specialized and delegate unrelated tasks.

src/NzbDrone.Core/Indexers/TorrentRssParser.cs
```diff
@@ -88,6 +91,20 @@ protected virtual String GetMagnetUrl(XElement item)
             return null;
         }
 
+        protected override long GetSize(XElement item)
+        {
+            var size = base.GetSize(item);
+            if (size == 0 && (!string.IsNullOrEmpty(SizeElementName)))
```
kayone: `SizeElementName.IsNotNullOrEmpty()`

src/NzbDrone.Core/Parser/Model/TorrentInfo.cs
kayone: this is interesting...

src/NzbDrone.Core.Test/IndexerTests/TorrentRssIndexerTests/TestTorrentRssIndexer.cs
kayone: there should be resharper settings checked in and loaded automatically, but I'm not sure if stuff like this are gonna be handled automatically, I think they only run if you explicitly call clean up.

https://github.com/Sonarr/Sonarr/blob/develop/src/NzbDrone.sln.DotSettings

src/NzbDrone.Core/Indexers/TorrentRssIndexer/TorrentRssIndexerParserSettings.cs
MichelZ: That's just an internal setting, nothing external.
It has to store data somewhere :)

src/NzbDrone.Core/Parser/Model/TorrentInfo.cs
MichelZ: Why?

src/NzbDrone.Core.Test/IndexerTests/TorrentRssIndexerTests/TestTorrentRssIndexer.cs
MichelZ: I'll have a look, thx

src/NzbDrone.Core/Parser/Model/TorrentInfo.cs
kayone: sorry, didn't mean interesting in a bad way, I meant it in a good way.

MichelZ:
I have corrected some of the issues

Taloth:
I think this is an excellent start, but I haven't yet looked at everything in detail.
There are a few things that should be changed to better match the rest of the coding, but I first would like to get into the functional part.

In my opinion the auto-detection mechanism is important. So storing the detected options somewhere is required.
I rather not have the chosen parser selected by the user, but I don't know what markus and kayone think about it.

The validity check needs some work. The DownloadUrl for example could be a valid url, but it could also be some random guid/hash, or a link to a comments page. Depending on which element it comes from. By default it comes from the encapsulation, so it shouldn't be a big issue, but it's something that needs to be checked more thoroughly.
Also check all rss results instead of only the first. Run the titles through Parser.ParseTitle and see if it gets a valid result in >=75% cases (or another reasonable threshold).
The Size could be the actual download size, or could be the size of the torrent file, validate that. The majority of results will be bigger than a dozen mb, so if all of em are below that threshold, the size is that of the torrent file itself and thus wrong.
That's just a couple of things of the top of my head.

MichelZ:
I thought about parsing & saving the settings as well, but did not know where to save the settings (especially not without exposing them to the user). Any hints there?

MichelZ:
I would like to ask users to provide sample RSS XML files for testing purposes. What would be the best way to do this? Any ideas?

Taloth:
@markus101 @kayone We have to talk about the ParserSettings, how best to deal with it and where.
Coz I wouldn't want to do it on the fly every time we do a request, so it has to be stored somewhere. It probably belongs as a property of the IndexerSettings (not shown on the UI), meaning we have to add that data once the Test is run or when the indexer is saved.

As for asking for sample rss, dunno the best approach, maybe post on the trello card? Gotta be explicit though that ppl have to sanitize it first before posting it somewhere, coz private trackers contain sensitive information.

markus101:
@taloth Yeah, we could actually use these hidden settings in other places as well, like the token for Plex, since at the moment we just cache it.

I originally cached the Plex token since I didn't like the idea of a consumer rewriting its settings and didn't have a clean way to do it. I'm sure we can come up with something reasonable though.

kayone:
what if we cache it for some hours in memory? I would rather us do that than store it in the db. when we have to deal with situations where we update the logic and then have to run a migrations. why not do it on the fly and just cache it?

MichelZ:
Can I help here in any way?

Taloth:
@MichelZ We had a chat internally yesterday about this PR. Specifically the parser settings.
We won't store the Parser settings in the UI/database, but rather cache it.
We also think the ```TorrentRssIndexer``` class contains a bit too much code. The ```*Indexer``` classes are intended to tie together the various components, but do fairly little on it's own.
So the code needs to be refactored a bit.

The idea is as follows:
Extract the logic that instantiates and configures the Parser into a separate class ```TorrentRssParserFactory``` (with interface).
The primary task of this Factory is to instantiate the appropriate Parser based on the settings.
To get the Settings, we need some detection logic and a cache. The cache resides in the Factory while the detection logic goes into a separate class ```TorrentRssSettingsDetector```, the result of which would be cached by ```TorrentRssParserFactory``` based on url, the cache can be preserved for the lifetime of the application, or a long time (day/week).
That way we have a *Factory which is tasked to instantiate the Parser and cache settings, a *Detector tasked with the detection/configuration logic keeping everything nice and separate, easier to unit test. The Detector will slowly grow, probably eventually get split up a bit.

So ```Indexer``` calls ```Factory.GetParser(url)```, ```Factory``` calls ```Detector``` (if not cached), ```Detector``` does it's job and returns ```TorrentRssParserInfo```, ```Factory``` uses Info to instantiate the appropriate ```Parser```, configures it, and returns it to the ```Indexer```.

How does that sound?

descention:
Would something like the following be acceptable? I'm currently unable to test the changes. I'm fairly certain I've covered all the required parts as I just took the EZTV rss indexer and removed static strings.

https://github.com/descention/Sonarr/commit/43975e621bb199107c3aa72e9850bb50096c0171

Taloth:
@descention That's pretty much the opposite direction compared to what I explained above.

Please look at the work Michel has done, validation, unittests etc. and the design suggestions I posted last.

Picking up this PR to finish it up is on my todo list, but it's going to take a while before I get to it.

MichelZ:
I was pretty busy the last weeks. I hope that I may be able to work on this soon as well.

stack72:
any idea if this PR is going to be merged? It's functionality that i'd love to see supported

MichelZ:
I finally came around to work on this PR. Please review.

src/NzbDrone.Core/Indexers/TorrentRssIndexer/TorrentRssIndexer.cs
```diff
@@ -0,0 +1,129 @@
...
+            }
+            catch (Exception)
+            {
+                failures.Add(new ValidationFailure("", "Feed cannot be parsed or is invalid. See log for details."));
+            }
+            
+            base.Test(failures);
+        }
+
+        protected override IList<ReleaseInfo> FetchReleases(IList<IEnumerable<IndexerRequest>> pageableRequests)
```
Taloth: What's the difference with the base class implementation?

src/NzbDrone.Core/Indexers/TorrentRssIndexer/TorrentRssIndexer.cs
```diff
@@ -0,0 +1,129 @@
...
+        {
+            var factory = new TorrentRssParserFactory();
+            return factory.GetParser(Settings, FetchIndexerResponse);
+        }
+
+        protected override void Test(List<ValidationFailure> failures)
+        {
+            try
+            {
+                var parser = this.GetParser();
```
Taloth: In the base class ```TestConnection``` already calls ```GetParser```, so you shouldn't need to override ```Test``` to add a separate test.

src/NzbDrone.Core/Indexers/TorrentRssIndexer/TorrentRssIndexer.cs
```diff
@@ -0,0 +1,129 @@
...
+        }
+
+        public override IIndexerRequestGenerator GetRequestGenerator()
+        {
+            return new TorrentRssIndexerRequestGenerator() { Settings = Settings };
+        }
+
+        public override IParseIndexerResponse GetParser()
+        {
+            var factory = new TorrentRssParserFactory();
```
Taloth: Add this as ```ITorrentRssParserFactory``` to the constructor of ```TorrentRssIndexer```. That way it gets properly instantiated by tinyioc.

src/NzbDrone.Core/Indexers/TorrentRssIndexer/TorrentRssIndexer.cs
```diff
@@ -0,0 +1,129 @@
...
+            return factory.GetParser(Settings, FetchIndexerResponse);
```
Taloth: Maybe provide the request generator as well?

src/NzbDrone.Core/Indexers/TorrentRssIndexer/TorrentRssIndexerSettings.cs
```diff
@@ -0,0 +1,42 @@
...
+            BaseUrl = string.Empty;
+            ValidEntryPercentage = 75;
+            ValidSizeThresholdMegabytes = 12;
+        }
+
+        [FieldDefinition(0, Label = "Full RSS Feed URL")]
+        public String BaseUrl { get; set; }
+
+        [FieldDefinition(1, Advanced = true, Label = "Valid Entry Percentage", HelpText = "Percentage of entries in the RSS feed which need to be parseable")]
+        public int ValidEntryPercentage { get; set; }
```
Taloth: Shouldn't everything be parsable?

src/NzbDrone.Core/Indexers/TorrentRssIndexer/TorrentRssIndexerSettings.cs
```diff
@@ -0,0 +1,42 @@
...
+
+        [FieldDefinition(2, Advanced = true, Label = "Valid Size Threshold (MB)", HelpText = "Threshold of Parsed Torrent Size to consider it valid")]
+        public int ValidSizeThresholdMegabytes { get; set; }
```
Taloth: Do we really want to expose this to the user?
I assume this and the previous option is intended to filter between .torrent size and the actual data size. So I understand the conundrum, but we should pick a sane value, put that in the Detector and not allow users to tweak it. We try to keep the number of config options to a minimum.

src/NzbDrone.Core/Indexers/TorrentRssIndexer/TorrentRssParserFactory.cs
```diff
@@ -0,0 +1,47 @@
+﻿using System;
+using NLog;
+
+namespace NzbDrone.Core.Indexers.TorrentRssIndexer
+{
+    public class TorrentRssParserFactory : ITorrentRssParserFactory
+    {
+        private static readonly ITorrentRssParserSettingsCache SettingsCache = new TorrentRssParserSettingsCache();
+
+        protected readonly Logger _logger = LogManager.GetCurrentClassLogger();
```
Taloth: Add constructor and init via that constructor, let tinyioc instantiate the class.

src/NzbDrone.Core/Indexers/TorrentRssIndexer/TorrentRssParserFactory.cs
```diff
@@ -0,0 +1,47 @@
+﻿using System;
+using NLog;
+
+namespace NzbDrone.Core.Indexers.TorrentRssIndexer
+{
+    public class TorrentRssParserFactory : ITorrentRssParserFactory
+    {
+        private static readonly ITorrentRssParserSettingsCache SettingsCache = new TorrentRssParserSettingsCache();
```
Taloth: No static. Use ```ICacheManager``` to get a nice persistent cache.
You also won't need to write your own threadsafe cache, CacheManager already does it. We use it throughout sonarr.

src/NzbDrone.Core/Indexers/TorrentRssIndexer/TorrentRssParserFactory.cs
```diff
@@ -0,0 +1,47 @@
...
+
+        /// <summary>
+        /// Get configured Parser based on <paramref name="settings"/>
+        /// </summary>
+        /// <param name="settings">Indexer Settings to use for Parser</param>
+        /// <param name="fetchIndexerResponseFunc">Func to retrieve Feed</param>
+        /// <returns>Configured Parser</returns>
+        public TorrentRssParser GetParser(TorrentRssIndexerSettings settings, Func<IndexerRequest, IndexerResponse> fetchIndexerResponseFunc)
+        {
+            TorrentRssIndexerParserSettings parserSettings = SettingsCache.Get(settings.BaseUrl);
```
Taloth: ICached (via ICacheManager) supports ```Get(key, () => func, TimeSpan lifetime)```. So you can properly initialize it.

src/NzbDrone.Core/Indexers/TorrentRssIndexer/TorrentRssParserFactory.cs
```diff
@@ -0,0 +1,47 @@
...
+
+            if (parserSettings == null)
+            {
+                _logger.Debug("Parser Settings not in cache. Trying to Parse Feed.");
```
Taloth: Add url in there so we know what it's trying to parse.

src/NzbDrone.Core/Indexers/TorrentRssIndexer/TorrentRssParserFactory.cs
```diff
@@ -0,0 +1,47 @@
...
+                
+                var detector = new TorrentRssSettingsDetector();
+                parserSettings = detector.Detect(settings, fetchIndexerResponseFunc);
+
+                if (parserSettings == null)
+                {
+                    throw new Exception("Could not parse stream. See log for more information.");
```
Taloth: Don't think 'See log for more information' is needed. Goes without saying.
Also add the url to the exception.

src/NzbDrone.Core/Indexers/TorrentRssIndexer/TorrentRssParserFactory.cs
```diff
@@ -0,0 +1,47 @@
...
+        /// <returns>Configured Parser</returns>
+        public TorrentRssParser GetParser(TorrentRssIndexerSettings settings, Func<IndexerRequest, IndexerResponse> fetchIndexerResponseFunc)
+        {
+            TorrentRssIndexerParserSettings parserSettings = SettingsCache.Get(settings.BaseUrl);
+
+            if (parserSettings == null)
+            {
+                _logger.Debug("Parser Settings not in cache. Trying to Parse Feed.");
+                
+                var detector = new TorrentRssSettingsDetector();
```
Taloth: again, ```ITorrentRssSettingsDetector``` tinyioc via constructor.

src/NzbDrone.Core/Indexers/TorrentRssIndexer/TorrentRssParserSettingsCache.cs
```diff
@@ -0,0 +1,84 @@
+﻿using System.Collections.Generic;
+
+namespace NzbDrone.Core.Indexers.TorrentRssIndexer
+{
+    public class TorrentRssParserSettingsCache : ITorrentRssParserSettingsCache
```
Taloth: As mentioned, replace this with a call to ICacheManager.GetCache.

src/NzbDrone.Core/Indexers/TorrentRssIndexer/TorrentRssSettingsDetector.cs
```diff
@@ -0,0 +1,223 @@
...
+
+        protected Func<IndexerRequest, IndexerResponse> _fetchIndexerResponseFunc;
+
+        /// <summary>
+        /// Detect settings for Parser, based on URL
+        /// </summary>
+        /// <param name="settings">Indexer Settings to use for Parser</param>
+        /// <param name="fetchIndexerResponseFunc">Func to retrieve Feed</param>
+        /// <returns>Parsed Settings or <c>null</c></returns>
+        public TorrentRssIndexerParserSettings Detect(TorrentRssIndexerSettings settings, Func<IndexerRequest, IndexerResponse> fetchIndexerResponseFunc)
```
Taloth: Not sure about the Func<> here, it makes sense in a way but feels a bit awkward. @markus101 What do you think about this?

src/NzbDrone.Core/Indexers/TorrentRssIndexer/TorrentRssSettingsDetector.cs
```diff
@@ -0,0 +1,223 @@
...
+
+            IndexerResponse response;
+            var settings = new TorrentRssIndexerParserSettings();
+            try
+            {
+                response = _fetchIndexerResponseFunc(request);
+            }
+            catch (Exception ex)
+            {
+                _logger.WarnException("Unable to connect to indexer: " + ex.Message, ex);
```
Taloth: "which indexer?" is going to be the user's question.

src/NzbDrone.Core/Indexers/TorrentRssIndexer/TorrentRssSettingsDetector.cs
MichelZ: It does feel awkward, but looked like the easiest way to get what I need :)

src/NzbDrone.Core/Indexers/TorrentRssIndexer/TorrentRssIndexerSettings.cs
MichelZ: Don't know, up to you. I thought it might make sense as an advanced option.

src/NzbDrone.Core/Indexers/TorrentRssIndexer/TorrentRssIndexerSettings.cs
MichelZ: They are advanced options. Up to you, I can hardcode them.

MichelZ:
@Taloth I've updated according to your suggestions, however I seem to be missing something with TinyIoC, I feel like I have to register some stuff but I'm lost as to where... Any help you could offer? Thanks.

Taloth:
You don't have to register anything, Sonarr basically auto-registers everything in those assemblies. So if you ask for a specific interface, the ioc container will find and instantiate an appropriate implementation.

I left some comments in the commit itself.

Other than that it looks almost ready.

MichelZ:
Fixed these.
I just can't get the tests to work anymore... They worked before changing to TinyIoC.. I'll keep looking

Taloth:
The testcases are isolated in a mock environment. So everything in the 'Subject' constructor will get a Mock instance unless specified differently.
You can use the Mocker.SetConstant and Mocker.Resolve methods to set or get the desired instances, for anything else it will create a mock.

In your case it means the TestRssIndexer gets a mock ITorrentRssParserFactory  instance, returning null for those methods.
Essentially, your tests cover a bit too big a unit. 

You would, for example, test TestRssIndexer by mocking ITorrentRssParserFactory, doing a Mocker.GetMock<ITorrentRssParserFactory>().Setup(v => v.GetParser(.....)).Returns(...);
That way you isolate RssIndexer from other components.
Then in a different test you would test TorrentRssParserFactory, mocking it's dependencies. to see if it behaves as intended.

If you want, I can setup a couple of tests.

src/NzbDrone.Core/Indexers/TorrentRssIndexer/TorrentRssIndexerSettings.cs
markus101: What's the maximum size that a torrent would be? I think this should be handled automatically, its just going to be another burden trying to support users that set this randomly and we have to figure it out.

Taloth: Hardcode the limit at 2MB. Torrent files normally will never reach that. Only very badly created .torrent files (too small piece size) would even come close.

src/NzbDrone.Core/Indexers/TorrentRssIndexer/TorrentRssIndexerSettings.cs
Taloth: In all honestly, no. A feed that isn't 100% parsable isn't reliable imo. Unless you can give me a good reason why we should allow it.

src/NzbDrone.Core/Indexers/TorrentRssIndexer/TorrentRssSettingsDetector.cs
Taloth: Talked to markus about it. Better just have the Detector import IHttpClient via constructor (IoC as usual) and fetch the url directly like here: https://github.com/Sonarr/Sonarr/pull/193/files#diff-140395caeff547e2a3eed6e67b699e29R202
That way you don't need the Func in Detect(..) and GetParser(..) making those interfaces cleaner.

markus101: Can we just hit http client directly? We're using the same implementation here so a func seems like overkill. Alternatively we could pass the result to this function, but that adds another level to go through.

MichelZ:
I've done the requested changes. I will try to update/redo the tests later this week.

Taloth:
@MichelZ @markus101 I picked up these commits, squashed it and then did some work on the tests, some additional code and cleanup.

See https://github.com/Sonarr/Sonarr/tree/torrentrss

I moved around a lot of the test xml files, figured it's more manageable this way.

Michel, you'll notice I split up the tests in 3 separate fixtures. One to test the Detector, one to test the factory and 'integration'-like test ```TorrentRssIndexerFixture``` that checks the whole stack.
For that one I had to setup quite a few Mocker Constants to make sure the right mocks vs instances were created. 
It should give us an acceptable test coverage and good harness in case we add support for others.

I also added the config option AllowZeroSize. Some feeds, particularly ShowRSS, but also womble, Torrentleech and others have no size.
I wanted this to be explicit, coz having no Size means we can't properly do size checks.

Also had to refactor the Detector code. Tbh I still am not entirely happy with it, doesn't report a useful message back to the user, but it will do the job for now.

Taloth:
One more thing, the build server failed the ImmortalSeed tests because of a timezone difference.
That feed doesn't have a timezone in the pubdate. We're going to have to assume something, prolly UTC, but can you check what it actually is in reality?

MichelZ:
@Taloth thx for taking care of the tests. The timezone is actually settable when you create the RSS feed for immortalseed, so it *could* be anything... don't know how to handle that :)

Taloth:
Could you ask em to include the timezone chosen (just " +0100" appended)? If not then no worries, but we'll assume UTC.
Their answer will possibly be no, in fear of breaking existing clients, but any rss client out there should be able handle it.

Taloth:
Merged to develop by 39aef8bed5bdb175f947b9e8305de5b4d282f315

PR #194, opened 2015-02-12
fix tooltip for add series button

thezoggy:


src/UI/AddSeries/SearchResultViewTemplate.hbs
```diff
@@ -80,12 +80,12 @@
...
                         <div class="btn-group">
-                            <button class="btn btn-success add x-add">
-                                <i class="icon-plus" title="Add"></i>
+                            <button class="btn btn-success add x-add" title="Add">
+                                <i class="icon-plus"></i>
                             </button>
 
-                            <button class="btn btn-success add x-add-search">
-                                <i class="icon-search" title="Add and Search for missing episodes"></i>
+                            <button class="btn btn-success add x-add-search" title="Add and Search for missing episodes">
```
kayone: This could actually be misleading if the user has selected a different search option from the dropdown.

kayone: not an issue with this change, just about this area in general.

markus101:
This will actually cause the button styling to have a fit when the tooltip is injected. We can attach it to the body instead, but that can lead to stuck tooltips.

Tooltip on the button also means IOS users need to click twice (once to activate the tooltip, another to activate the button), due to an IOS bug (though its a problem elsewhere, so I wouldn't reject it for this reason).

PR #195, opened 2015-02-12
Modify Tooltip mixin to deal with the btn-group distortions

Taloth:
To avoid the button distortions we usually bind to body, but that messes up tooltips visible in dynamic (signalr) rows. coz they get stuck.
This offers best of both worlds, we automatically find the btn-group, input-group and attach as sibling there.

We do this by overriding the getOptions of the Tooltip object to determine the correct container. If we don't find a btn group we don't change the container, thus the tooltip gets added near the element.

Also removed the localized fixes.
Also had to remove the tooltip from Copy clipboard in the ical modal coz that gets copied to the zeroclipboard and gets stuck when the model closes.

### DIFF
--- a/src/UI/Activity/Queue/TimeleftCell.js
+++ b/src/UI/Activity/Queue/TimeleftCell.js
@@ -14,7 +14,6 @@ module.exports = NzbDroneCell.extend({
                 var time = '{0} at {1}'.format(FormatHelpers.relativeDate(ect), moment(ect).format(UiSettingsModel.time(true, false)));
                 this.$el.html('-');
                 this.$el.attr('title', 'Will be processed during the first RSS Sync after {0}'.format(time));
-                this.$el.attr('data-container', 'body');
                 return this;
             }
             var timeleft = this.cellValue.get('timeleft');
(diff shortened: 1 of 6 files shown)


Taloth:
Bit of an extension on https://github.com/Sonarr/Sonarr/pull/194

src/UI/Shared/Tooltip.js
```diff
@@ -1,8 +1,28 @@
 var $ = require('jquery');
 
+var Tooltip = $.fn.tooltip.Constructor;
```
kayone: require bootstrap as well.

src/UI/Activity/Queue/TimeleftCell.js
```diff
@@ -14,7 +14,6 @@ module.exports = NzbDroneCell.extend({
                 var time = '{0} at {1}'.format(FormatHelpers.relativeDate(ect), moment(ect).format(UiSettingsModel.time(true, false)));
                 this.$el.html('-');
                 this.$el.attr('title', 'Will be processed during the first RSS Sync after {0}'.format(time));
-                this.$el.attr('data-container', 'body');
```
markus101: This is going to cause the column to shift (similar to btn-group). Before seeing this I was thinking we should do the same for table cells though (can we?).

PR #196, opened 2015-02-13
Dedupe tags

markus101:
@kayone @Taloth 

Currently there are a number of users with duplicate tags (different tags with the same label), this will clean those up (and fix things that are referencing deleted tags) as well as prevent it from reoccurring.

### DIFF
--- a/src/NzbDrone.Core/Datastore/Migration/066_add_tags.cs
+++ b/src/NzbDrone.Core/Datastore/Migration/066_add_tags.cs
@@ -16,6 +16,9 @@ namespace NzbDrone.Core.Datastore.Migration
 
             Alter.Table("Notifications")
                  .AddColumn("Tags").AsString().Nullable();
+
+            Execute.Sql("UPDATE Series SET Tags = '[]'");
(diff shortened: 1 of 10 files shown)


src/NzbDrone.Core/Datastore/Migration/066_add_tags.cs
```diff
@@ -16,6 +16,9 @@ protected override void MainDbUpgrade()
 
             Alter.Table("Notifications")
                  .AddColumn("Tags").AsString().Nullable();
+
+            Execute.Sql("UPDATE Series SET Tags = '[]'");
+            Execute.Sql("UPDATE Notifications SET Tags = '[]'");
```
kayone: I'm not sure about these, we should update `EmbededDocumentConverter` to deal with these properly. null should be treated the same as empty list when target is an `IEnumarable`

markus101: Its other migrations that would be the issue, if the user is coming from a really old DB version the migrations don't null check and it fails.

This should have been done originally, but this will help anyone else that upgrades from an old version.

PR #197, opened 2015-02-15
Grab the most popular torrent from kickass, not the newest

grillba:
Apologies, new here. Let me know if I've done the wrong this with this pull request

### DIFF
--- a/src/NzbDrone.Core/Indexers/KickassTorrents/KickassTorrentsRequestGenerator.cs
+++ b/src/NzbDrone.Core/Indexers/KickassTorrents/KickassTorrentsRequestGenerator.cs
@@ -123,7 +123,7 @@ namespace NzbDrone.Core.Indexers.KickassTorrents
 
             if (PageSize == 0)
             {
-                var request = new IndexerRequest(String.Format("{0}/{1}{2}/?rss=1&field=time_add&sorder=desc", Settings.BaseUrl.TrimEnd('/'), rssType, searchUrl), HttpAccept.Rss);
+                var request = new IndexerRequest(String.Format("{0}/{1}{2}/?rss=1&field=users&sorder=desc", Settings.BaseUrl.TrimEnd('/'), rssType, searchUrl), HttpAccept.Rss);
                 request.HttpRequest.SuppressHttpError = true;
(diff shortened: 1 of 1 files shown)


MichelZ:
You may want to make this configurable in an advanced setting?

Taloth:
This will break the entire kickass rss feed and it won't fix what you want either. I'll explain why.

First, why it breaks: The same method is used for the recent rss feed, by modifying this function you've basically destroyed the entire recent feed capability because it no longer returns newly posted items, but only the most popular ones.

Secondly, prioritization isn't done at this level, it's done [here](https://github.com/Sonarr/Sonarr/blob/develop/src/NzbDrone.Core/DecisionEngine/DownloadDecisionPriorizationService.cs#L28-L44), so any sorting you've done at feed level will be ignored. Btw., the result from all indexers are combined and prioritized there.
Also, please note line 39, where we already prioritize by seeds.
@kayone had some ideas about how to improve the prioritization, mostly to fix problems related to the size and season packs, but we've not decided yet on a best approach.

The problem you're running into is likely that the size classification is more important than many seeds.

On a final note, the https://github.com/grillba/Sonarr/commit/7c5ffe3e0a0fd05b5591ad3df9a9816282cdf765 commit doesn't belong here, never use a PR to your own repository as a means to merge stuff, that's what git merge/rebase is for.

grillba:
Alrighty! Thanks for the feedback, as I said I'm rather new at this so I bound to make a few mistakes as I get up to scratch.

I'll try explain what I'm trying to get working better. I have mostly an out of the box setup, and only the KickassTorrents indexer in use. When viewing a missing Ep and performing an "Automatic Search", it is pulling down the most recent match with only a handful of seeds, rather than the most popular one with thousands of seeds.

Now that this all ballsed up, do I cancel it while I revisit my code, or do you guys push it to the junk pile?

kayone:
I think it makes sense to grab the most recent items when doing an RSS fetch and most popular when doing a manual search.

grillba:
Even when there is multiple matches in an RSS feed (excluding propers,
etc), don't you still want to grab the most popular torrent?

PR #198, opened 2015-02-15
Multilanguage support when searching for new series

vawen:
This changes uses the system locale information (language) when doing tvdb show searchs. Two queries are done if the system locale is not english, the second query is done in the machine locale language.

With the results of the second query the list returned includes the metadata information in the "locale language" if it exists or in english if not.

The "locale title" is stored in the Series DAO and used later on when parsing episodies names to ensure that the show is located. (IE: a file named "Juego de tronos" or "Game of Thrones" both will match for "Game of Thrones" serie if the locale of the machine is spanish).

The Series info is overwritten on the update method with the skyhook information, so all the metadata are still in english after adding the serie (the update method preserves locale title in the new attribute).


src/NzbDrone.Core/Parser/Parser.cs
```diff
@@ -182,7 +186,7 @@ public static class Parser
         private static readonly Regex AnimeReleaseGroupRegex = new Regex(@"^(?:\[(?<subgroup>(?!\s).+?(?<!\s))\](?:_|-|\s|\.)?)",
                                                                 RegexOptions.IgnoreCase | RegexOptions.Compiled);
 
-        private static readonly Regex LanguageRegex = new Regex(@"(?:\W|_)(?<italian>\b(?:ita|italian)\b)|(?<german>german\b|videomann)|(?<flemish>flemish)|(?<greek>greek)|(?<french>(?:\W|_)(?:FR|VOSTFR)(?:\W|_))|(?<russian>\brus\b)|(?<dutch>nl\W?subs?)",
+        private static readonly Regex LanguageRegex = new Regex(@"(?:\W|_)(?<spanish>\bSPA\b|castellano|español)|(?<italian>\bita\b|italian)|(?<german>german\b|videomann)|(?<flemish>flemish)|(?<greek>greek)|(?<french>(?:\W|_)(?:FR|VOSTFR)(?:\W|_))|(?<russian>\brus\b)|(?<dutch>nl\W?subs?)",
```
markus101: Looks like this might be changed off an old version of the code, since Italian shouldn't be changed this way.

vawen: True, will correct that.

Taloth:
Couple of things:
- migration 78 & 79 is already in use, use 80 instead.
- searching is going to move to skyhook, so all changes to the tvdbproxy are going to be useless.
- I'm not happy with using the system locale, most of our users would not even want that. If we're going to support this it will have to be tied to the series language specified by the user.

src/NzbDrone.Core/Parser/Parser.cs
```diff
@@ -22,6 +22,10 @@ public static class Parser
 //                new Regex(@"^(?:(?<absoluteepisode>\d{2,3})(?:_|-|\s|\.)+)+(?<title>.+?)(?:\W|_)+(?:S?(?<season>(?<!\d+)\d{1,2}(?!\d+))(?:(?:\-|[ex]|\W[ex]){1,2}(?<episode>\d{2}(?!\d+)))+)",
 //                          RegexOptions.IgnoreCase | RegexOptions.Compiled),
 
+                //Serie with Title - Some text and season as number, Single ([Cap.105]) AND Multi ([Caps.105_107])
+                new Regex(@"^(?<title>.+?)(\s*\-\s*)(\w+)(\s*|.)?(\s*|.)?(?<!\d)(?<season>\d+)?(.+?)(\[Cap(\w*?)(.))(?<!\d)(\k<season>(?<episode>\d+))((_\k<season>(?<episode>\d+)\])|(.*))",
```
markus101: This seems very specific, but I don't really see why its needed given the short example here.

A few things:

- Tests need to be added for this regex, for both single and multi-episode conventions.
- Season/episode numbers need to exclude other leading/trailing episode numbers
- I don't think this should be above all the anime regexes either, but tests will confirm

vawen:
Updates:
* Migration moved to 80.
* Added an option to choose aditional language at indexers settings.

markus101:
No the direction we want to take searching and parsing.

PR #199, opened 2015-02-16
English: Clarify some option explanations

larsjohnsen:


markus101:
Please re-open this PR against develop, we don't merge directly into master.

Personally I'm not a fan of "Activate to", its implied given its a yes/no option.

I'd prefer:
`Be notified when episodes are available for download and has been sent to a download client`

over:
`Activate to be notified when episodes are available for download and has been sent to a download client`

PR #207, opened 2015-02-18
Validation warnings for missing Category in sabnzbd/nzbget

Taloth:
Atm only for sabnzbd/nzbget coz most others need plugins or workarounds to have categories.

Fire away.

### DIFF
--- a/src/NzbDrone.Core/Download/Clients/Nzbget/NzbgetSettings.cs
+++ b/src/NzbDrone.Core/Download/Clients/Nzbget/NzbgetSettings.cs
@@ -3,6 +3,7 @@ using FluentValidation;
 using FluentValidation.Results;
 using NzbDrone.Core.Annotations;
 using NzbDrone.Core.ThingiProvider;
+using NzbDrone.Core.Validation;
 using NzbDrone.Core.Validation.Paths;
 
 namespace NzbDrone.Core.Download.Clients.Nzbget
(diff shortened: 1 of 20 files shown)


src/NzbDrone.Core/Download/Clients/Nzbget/NzbgetSettings.cs
```diff
@@ -18,6 +19,8 @@ public NzbgetSettingsValidator()
 
             RuleFor(c => c.TvCategory).NotEmpty().When(c => !String.IsNullOrWhiteSpace(c.TvCategoryLocalPath));
             RuleFor(c => c.TvCategoryLocalPath).IsValidPath().When(c => !String.IsNullOrWhiteSpace(c.TvCategoryLocalPath));
+
+            RuleFor(c => c.TvCategory).NotEmpty().WithMessage("Using a separate category for Sonarr is recommended").AsWarning();
```
markus101: Technically its required now, since CDH is broken without it. Ideally we require it when editing settings, but not when validating for using it (which was our issue with making it mandatory for existing users).

src/NzbDrone.Core/NzbDrone.Core.csproj
```diff
@@ -869,8 +869,10 @@
     <Compile Include="Update\UpdateVerificationFailedException.cs" />
     <Compile Include="Validation\FolderValidator.cs" />
     <Compile Include="Validation\IpValidation.cs" />
-    <Compile Include="Validation\LangaugeValidator.cs" />
+    <Compile Include="Validation\LanguageValidator.cs" />
```
markus101: Took me a few tries to figure out why this line changed, spelling FTL.

kayone:
I've brought this up before, I'm not sure why we need to make categories required. It's just an extra step for users to do. If a user wants sonnar to be scoped to a specific category for whatever reason, it's their choice and they can set the category, otherwise we shouldn't force everyone to have to go to sab/nzbget whatever and add a category there.

if right now CDH is broken without a category that's a different issue that needs to be fixed, I don't think making them required is the solution.

Taloth:
It's already fixed.

With category we accept imports that sonarr didn't grab, without we don't do that. Having a category is recommended but not required, which is why I went through all this to add warning logic to validation.

Taloth:
image

PR #208, opened 2015-02-20
Episode file editor

markus101:
@kayone @Taloth 

Not sure about the icon (we need to update to Font Awesome 4 to get some nicer ones):
https://www.dropbox.com/s/zrgousjbd6n7q8s/shot_150219_221448.png

Pretty simple editor:
https://www.dropbox.com/s/31sgxyi3xeywhra/shot_150219_221805.png?dl=0

Delete has a confirm, changing the quality does not since its reversible (without setting the recycling bin first).

### DIFF
--- a/src/UI/Cells/EpisodeFilePathCell.js
+++ b/src/UI/Cells/EpisodeFilePathCell.js
@@ -1,0 +1,19 @@
+var reqres = require('../reqres');
+var NzbDroneCell = require('./NzbDroneCell');
+
+module.exports = NzbDroneCell.extend({
+    className : 'episode-file-path-cell',
+
+    render : function() {
+        this.$el.empty();
+
+        if (reqres.hasHandler(reqres.Requests.GetEpisodeFileById)) {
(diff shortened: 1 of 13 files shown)


Taloth:
UI should be more like series editor, with a fixed footer.

markus101:
So not a modal? I went with modal to keep it within the context of the series, instead of a separate page, since thats the style we typically use.

Wondering if it makes more sense to make the existing season grids editable via a toggle, that way the footer idea fits and the page context stays the same.

Taloth:
Season grids don't have filenames and such. It's probably also complicated to make it work like that.
Personally I think this should be a full page with it's own route. Gives you more space to do exotic stuff in the future.

markus101:
I think we could make the same argument about renaming. I'm leaning towards leaving it as-is and revisiting it during the UI upgrades, especially since modals + mobile aren't great (even in BS3).

We need a tie breaker from @kayone

PR #202, opened 2015-02-17
Compiler: Support for XBuild

larsjohnsen:
I am trying to get Sonarr to compile on Linux using XBuild, and this commit represent what changes is required to have it running without loss of any functionality

Requirements to build and run:
- nodejs (which provides npm)
- mono (which provides xbuild).
- sqlite
- libmediainfo

Instructions:
1. Install certificates necessary for NuGet: `mozroots --import --machine --sync`
2. Install necessary npm packages by running `npm install`
3. To compile, run `MONO_IOMAP=case xbuild src/NzbDrone.sln` and `gulp build`.

None known issues when compared to using MSBuild.

Compiler warnings:
```
/tmp/yaourt-tmp-l/aur-sonarr-git/src/Sonarr/src/NzbDrone.sln:  warning : Don't know how to handle GlobalSection ExtensibilityGlobals, Ignoring.
/tmp/yaourt-tmp-l/aur-sonarr-git/src/Sonarr/src/NzbDrone.sln:  warning : Don't know how to handle GlobalSection JSLint, Ignoring.

/tmp/yaourt-tmp-l/aur-sonarr-git/src/Sonarr/src/NzbDrone.sln (default targets) ->
(Build target) ->
/tmp/yaourt-tmp-l/aur-sonarr-git/src/Sonarr/src/Marr.Data/Marr.Data.csproj (default targets) ->
/usr/lib/mono/4.5/Microsoft.CSharp.targets (CoreCompile target) ->

	DataMapper.cs(619,18): warning CS0219: The variable `parentType' is assigned but its value is never used
	DataMapper.cs(786,17): warning CS0219: The variable `mappingHelper' is assigned but its value is never used
	QGen/QueryBuilder.cs(226,17): warning CS0219: The variable `mappingHelper' is assigned but its value is never used

/tmp/yaourt-tmp-l/aur-sonarr-git/src/Sonarr/src/NzbDrone.sln (default targets) ->
(Build target) ->
/tmp/yaourt-tmp-l/aur-sonarr-git/src/Sonarr/src/NzbDrone.Common/NzbDrone.Common.csproj (default targets) ->
/usr/lib/mono/4.5/Microsoft.CSharp.targets (CoreCompile target) ->

	Exceptron/fastJSON/JSON.cs(625,18): warning CS0219: The variable `t1' is assigned but its value is never used
	Http/HttpClient.cs(214,25): warning CS0162: Unreachable code detected
[...]

PR #210, opened 2015-02-26
Compiler: Support for XBuild

larsjohnsen:
Continues #202 as it was merged then reverted.

I've fixed the issue that  `[...] the signatures are different` as mentioned by markus101 in https://github.com/Sonarr/Sonarr/pull/202#issuecomment-75893962.


kayone:
thanks for reopening it.

larsjohnsen:
I'd appreciate getting "Fix case inconsistencies" and "Fix case inconsistencies" merged, as they contain fairly benign changes.

markus101:
Which ones? You referenced the first one twice. Please open a new PR for "Fix case inconsistencies" so we can review before merging.

markus101:
Added changes that allow it to build under mono.

PR #211, opened 2015-02-26
Mono: Set process name

larsjohnsen:
As requested here: https://trello.com/c/XLNTIkb3/691-set-process-name-on-mono


markus101:
At this time this isn't something we can merge because we can't guarantee it won't cause issues for someone during an automatic update (and might not be caught in the catch block).

If anything this will be done in v3 of Sonarr.

PR #212, opened 2015-02-26
CSS: Use color red instead of grey when disabled

larsjohnsen:
This makes the off-state color consistent with the toggle switches, and easier to recognize.

### DIFF
--- a/src/UI/Settings/DownloadClient/DownloadClientItemViewTemplate.hbs
+++ b/src/UI/Settings/DownloadClient/DownloadClientItemViewTemplate.hbs
@@ -7,7 +7,7 @@
         {{#if enable}}
             <span class="label label-success">Enabled</span>
         {{else}}
-            <span class="label label-default">Not Enabled</span>
+            <span class="label label-danger">Not Enabled</span>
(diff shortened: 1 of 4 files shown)


kayone:
We generally try to include a screenshot for UI pull-request, perhaps a before and after.

That way we can actually see what they look like without having to checkout the branch/compile/run.

markus101:
Personally I don't like this it will make everything look like Christmas (Red and Green) and why we didn't do it originally.

markus101:
This is what it would look like: https://www.dropbox.com/s/aq1tsvuly05ljyx/shot_150225_231645.png?dl=0

larsjohnsen:
Perhaps a grayer red would be better? http://i.imgur.com/rJ19Qn2.png

markus101:
I'm still not a fan, its inconsistent with other usages of the red and to me implies something is wrong when really its just disabled. Perhaps we need to rethink the use of red for the off position of the toggles instead of altering these (not really something that I think we need to do right now though).

larsjohnsen:
Agreed.

PR #213, opened 2015-02-26
Delete obsolete file: handlebars.runtime.min.js

larsjohnsen:
handlebars.runtime.js is used instead


kayone:
thanks.

kayone:
in the future please name you branches related to change instead of generic `patch-#`

PR #214, opened 2015-03-01
UI: Separate setting groups & clarify tooltips

larsjohnsen:
New look: https://i.imgur.com/svnWK1i.png

### DIFF
--- a/src/UI/Settings/Notifications/Edit/NotificationEditViewTemplate.hbs
+++ b/src/UI/Settings/Notifications/Edit/NotificationEditViewTemplate.hbs
@@ -33,7 +33,7 @@
                         </label>
 
                         <span class="help-inline-checkbox">
-                            <i class="icon-sonarr-form-info" title="Do you want to get notifications when episodes are grabbed?"/>
+                            <i class="icon-sonarr-form-info" title="Be notified when episodes are available for download and has been sent to a download client"/>
                         </span>
                     </div>
                 </div>
@@ -55,7 +55,7 @@
                         </label>
 
                         <span class="help-inline-checkbox">
-                            <i class="icon-sonarr-form-info" title="Do you want to get notifications when episodes are downloaded?"/>
+                            <i class="icon-sonarr-form-info" title="Be notified when episodes are successfully downloaded"/>
                         </span>
                     </div>
                 </div>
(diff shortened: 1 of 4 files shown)


src/UI/Settings/Notifications/Edit/NotificationEditViewTemplate.hbs
```diff
@@ -77,24 +79,26 @@
                         </label>
 
                         <span class="help-inline-checkbox">
-                            <i class="icon-sonarr-form-info" title="Do you want to get notifications when episodes are upgraded to a better quality?"/>
+                            <i class="icon-sonarr-form-info" title="Be notifed when episodes are upgraded to a better quality"/>
```
kayone: spelling in `notifed`

larsjohnsen:
My reason for this change is that I was confused why I didn't get any notifications through Pushbullet. The reason was that I mistook "Tags" for how the Pushbullet notifications should be labeled.

"Filter Series Tags" could perhaps better be written as "For Tags".

Taloth:
I think I like how that looks, but the same lines should be added to Indexer, download clients and metadata view as well for consistency.

PS: if you copy/paste the image it gets uploaded to github.

larsjohnsen:
Amendment: Made equivalent changes on profiles, indexer, download clients, and metadata.

src/UI/Settings/Notifications/Edit/NotificationEditViewTemplate.hbs
```diff
@@ -77,24 +77,26 @@
...
-                            <i class="icon-sonarr-form-info" title="Do you want to get notifications when episodes are upgraded to a better quality?"/>
+                            <i class="icon-sonarr-form-info" title="Be notified when episodes are upgraded to a better quality"/>
                         </span>
                     </div>
                 </div>
             </div>
 
             <div class="form-group">
-                <label class="col-sm-3 control-label">Tags</label>
+                <label class="col-sm-3 control-label">Filter Series Tags</label>
```
markus101: I think for consistency this should remain Tags, we can use to the info tooltip to explain it.

src/UI/Settings/Notifications/Edit/NotificationEditViewTemplate.hbs
```diff
@@ -77,24 +77,26 @@
                         </label>
 
                         <span class="help-inline-checkbox">
-                            <i class="icon-sonarr-form-info" title="Do you want to get notifications when episodes are upgraded to a better quality?"/>
+                            <i class="icon-sonarr-form-info" title="Be notified when episodes are upgraded to a better quality"/>
```
markus101: I'm still struggling with the "Be notified" portion of this, thoughts on replacing it with "Notifications", Should it would becomes "Notifications when X".

PR #215, opened 2015-03-04
Manual Import

markus101:
@kayone @Taloth

This is mostly UI work for manual import, currently you have an extra button on the Wanted page that lets you choose a file to import (then select the series/season/episode(s)/quality) or a folder which lets you automatically import that folder (instead of using the drone factory).

I need to do the API still, but it feels a big ugly, I need to manually create a parsed episode result and localepisode - though I will extract what I can without making it all manual.


kayone:
screen shots?

markus101:
Here is the flow to import a file:

Select Series
shot_150304_073841

Select Season
shot_150304_074015

Select Episodes
shot_150304_074049

Select Quality
shot_150304_074108

Confirmation
shot_150304_074129

kayone:
cropped screenshots :disappointed:

kayone:
and I think it would be much nicer if the whole thing was doing in dropdowns in a single page. instead of different pages and cards.

markus101:
Its a modal.... You don't need to see a grayed out background behind it. The bottom buttons are cutoff because of the number of results.

That feels a lot more limiting IMO, especially for episodes where the title is important.

kayone:
I know, but cropped screenshots just look ugly, even if the ui is nice it'll look off (they are not very good to judge if it looks _good_)

I'm not sure how it would be more limiting, if anything the current UI would be more limiting.

markus101:
Why do you think the wizard style flow is limiting?

kayone:
it requires too much scrolling, eye scanning and clicking next/back. why would we want to show list of all seasons (think of some shows with 20/30 seasons) and having the user scroll throw a list, and finding the one and clicking on it, when they could just do a twitter search thing.

same with series and episode. I just don't think this UI fits.

kayone:
also, you need to make 4 choises, series, season, episode (season and episode could be merged) and quality. I don't think we need a wizard. you need the wizard now because the ui needs too much space and it wouldn't fit in one page. if you replaced with with dropdown/search boxes it would fit pretty nicely.

markus101:
So instead of scrolling a big page you would scroll through a tiny drop down list, in a bunch of cases since filtering on a number in a drop down sounds awkward to me or people would know to do it. Finding Canada in the list of countries still drives me nuts every time I need to do it.

How would you combine seasons and episodes? Not really viable for large shows like The Daily Show (DOM will be slow).

How would you select multiple episodes from a drop down?

I actually started with the navbar style search for series, it felt awkward.

If the scrolling is an issue we can do filtering on every page and I think that would be a better since would could show more details if we want/need to.

The next button is only required on the episodes page (multi-select), otherwise selecting an option advances it.

Even the import button might be better as a bigger button in the modal panel, more of a call to action.

kayone:
I still think dropdowns with search in a single panel would be better, but if you want to keep the list I would add search and get rid of the card style, it doesn't look good on small elements. even a simple table would probably be better, like the way episode is done. or buttons.

kayone:
your import and close buttons are in the wrong order, also, in other places we use cancel.

markus101:
New UI:

From common folder:
shot_150407_195334

From Activity: Queue
shot_150407_195457

I also removed Force Import as it was a hack and confusing to users, in favour of manual import.

PR #216, opened 2015-03-04
Sort icon

Taloth:
Sort icon now a bit more to the right and can overlap the header itself.
Also added tooltip support to the header.

Ow, and also fixed the sort direction cycle.

### DIFF
--- a/src/UI/Content/theme.less
+++ b/src/UI/Content/theme.less
@@ -292,3 +292,9 @@ dl.info {
     vertical-align : bottom;
   }
 }
+
+.sort-direction-icon {
+  .pull-right();
+  position : relative;
+  width    : 0px;
+}
--- a/src/UI/Shared/Grid/HeaderCell.js
+++ b/src/UI/Shared/Grid/HeaderCell.js
@@ -17,13 +17,19 @@ module.exports = function() {
         render : function() {
             this.$el.empty();
             this.$el.append(this.column.get('label'));
+            if (this.column.get('tooltip')) {
+                this.$el.attr({
+                    'title'          : this.column.get('tooltip'),
+                    'data-container' : '.table'
+                });
+            }
 
             var column = this.column;
             var sortable = Backgrid.callByNeed(column.sortable(), column, this.collection);
 
             if (sortable) {
                 this.$el.addClass('sortable');
-                this.$el.append(' <i class="pull-right"></i>');
+                this.$el.prepend(' <i class="sort-direction-icon"></i>');
             }
 
             //Do we need this?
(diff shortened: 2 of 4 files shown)


src/UI/Shared/Grid/HeaderCell.js
```diff
@@ -17,13 +17,17 @@ module.exports = function() {
         render : function() {
             this.$el.empty();
             this.$el.append(this.column.get('label'));
+            if (this.column.get('tooltip')) {
+                this.$el.attr('title', this.column.get('tooltip'));
+                this.$el.attr('data-container', '.table');
```
kayone: use an object to set both at the same time, less touching the DOM = more faster.

kayone:
can you post screenshots.

thanks.

src/UI/Content/theme.less
```diff
@@ -292,3 +292,9 @@ dl.info {
     vertical-align : bottom;
   }
 }
+
+.sort-direction-icon {
+  .pull-right();
+  position: relative;
+  width: 0px;
```
markus101: Are we aligning values in less files as well?

Taloth: originally the pull-right was in the js file. I can move it back there if you want.

Taloth:
Posted a couple of screenshots on hangout, so didn't think it was needed.
image

markus101:
In before @kayone "CROPPED?!?!?!" :smile:

It looks good to me.

src/UI/Content/theme.less
markus101: Sorry, I meant the keys/values

```
position : relative;
width     : 0px;
```

src/UI/Shared/Grid/HeaderCell.js
```diff
@@ -97,11 +101,16 @@ module.exports = function() {
             var column = this.column;
             var sortable = Backgrid.callByNeed(column.sortable(), column, collection);
             if (sortable) {
+                var isSorted = this.$el.children('.icon-sonarr-sort-asc,.icon-sonarr-sort-desc').length !== 0;
                 var direction = collection.state.order;
-                if (direction === 'ascending' || direction === -1) {
-                    direction = 'descending';
+                if (column.get('sortType') === 'fixed' || !isSorted) {
+                    direction = column.get('direction') || 'ascending';
                 } else {
```
markus101: New line the else (I'm royally confused with what our style should be at this point).

kayone:
the red icon on the header could make people think there is something wrong.

Taloth:
So you suggest making it black? add a icon to icons.less?

kayone:
honestly, I would suggest to not have the icon, even in black (though it is better) it still suggests there is something wrong. any reason why not just have the blank column header with the sort indicator? if it's a tool tip we could just attach it to an invisible icon.

Taloth:
We didn't have content in the header, I think we should. Having nothing in there caused the whole mess in the first place.

What about http://fortawesome.github.io/Font-Awesome/icon/exclamation/
Distinctly different than the http://fortawesome.github.io/Font-Awesome/icon/exclamation-circle/ we use in the rest of the application.

kayone:
how about this, we add a green checkmark icon for the rows that don't have an error, and make the header http://fortawesome.github.io/Font-Awesome/icon/flag/

PR #217, opened 2015-03-04
Removed hardcoded dot prefix from the transmission category [migration 81]

Taloth:
With migration to add the dot to existing configured Transmission Settings so we don't break existing users.

### DIFF
--- a/src/NzbDrone.Core/Datastore/Migration/081_move_dot_prefix_to_transmission_category.cs
+++ b/src/NzbDrone.Core/Datastore/Migration/081_move_dot_prefix_to_transmission_category.cs
@@ -1,0 +1,56 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Data;
+using System.Linq;
+using FluentMigrator;
+using NzbDrone.Common.Extensions;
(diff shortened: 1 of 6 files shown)


kayone:
needs migration tests

src/NzbDrone.Core/Datastore/Migration/081_move_dot_prefix_to_transmission_category.cs
```diff
@@ -0,0 +1,56 @@
...
+                {
+                    while (reader.Read())
+                    {
+                        var id = reader.GetInt32(0);
+                        var settingsJson = reader.GetString(1);
+
+                        var settings = Json.Deserialize<Dictionary<string, object>>(settingsJson);
+
+                        var tvCategory = settings.GetValueOrDefault("tvCategory") as string;
+                        if (!tvCategory.IsNullOrWhiteSpace())
```
kayone: `IsNotNullOrWhiteSpace `

Taloth:
There you go, migration test.

kayone:
are there any indication that someone might be launching one of these?

PR #218, opened 2015-03-04
Fanzub url can now be modified [migration 82]

Taloth:
Running a bit ahead to the possibility that ppl will use the fanzub sources to setup alternative sites. This allows em to configure the url for Sonarr.

### DIFF
--- a/src/NzbDrone.Core/Datastore/Migration/082_add_fanzub_settings.cs
+++ b/src/NzbDrone.Core/Datastore/Migration/082_add_fanzub_settings.cs
@@ -1,0 +1,20 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Data;
+using System.Linq;
+using FluentMigrator;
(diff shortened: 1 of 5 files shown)


kayone:
I think we should hold this off until it happens.

Taloth:
I've seen a couple of ppl working on it, not sure if they're going to be successful. But wouldn't it be nice to be a bit ahead?
But yes, PRed it just to be ready, and to reserve the migration ID.

markus101:
Whats the drop dead date? 31st of March?

I think this is likely the direction it will go, my only concern is there will be a bunch of them that don't work 100% correctly and it ends up being a pain to support, but thats not a big concern for me.

kayone:
are there any indication that someone might be launching one of these?

markus101:
Looks like this might be a clone: http://www.animenzb.com/

PR #219, opened 2015-03-05
UI: (Prototype) User definable link to file

larsjohnsen:
Example implementation of a direct link to the media file.

In most browsers, the link must be right-clicked then opened, due to security measures. There must also exist an application which can open it. In my instance `smb://` is launched by the media player Bomi, which has Samba support, so I can create a link to match the expected path.

### DIFF
--- a/src/UI/Cells/FileLinkCell.js
+++ b/src/UI/Cells/FileLinkCell.js
@@ -1,0 +1,40 @@
+var vent = require('vent');
+var Backgrid = require('backgrid');
+
+module.exports = Backgrid.Cell.extend({
+    className : 'file-link-cell',
+
+    render : function() {
+        function setCustomReplace() {
+            while (true) {
+                localStorage.fileLinkReplace = window.prompt("Set characters that shall be replaced:", "/mnt/");
+                localStorage.fileLinkReplaceWith = window.prompt("Set what aforeselected characters shall be replaced with", "smb://192.168.1.1/main/");
+
+                var link = makeLink();
+                if (window.confirm("Link will be: " + link + ". Click OK if this is correct or you give up. To later modify, clear localStorage.")) {
+                    return link;
+                }
+            }
+        }
+
+        function makeLink() {
+            if (localStorage.hasOwnProperty("fileLinkReplace") && localStorage.hasOwnProperty("fileLinkReplaceWith")) {
+                return path.replace(localStorage.fileLinkReplace, localStorage.fileLinkReplaceWith);
+            } else {
+                if (window.confirm("Pattern for defining link to " + path + " has not been created. Do it now?")) {
+                    return setCustomReplace();
+                }
+            }
+        }
+
+        path = this.model.get('path');
+        var link = makeLink();
+        if (link) {
+            this.$el.empty();
+            this.$el.html('<a><i class="icon-open-file x-open-file" title="Open"></a>');
+            this.$("a").prop("href", link);
+        }
+
+        return this;
+    },
+});
--- a/src/UI/Cells/cells.less
+++ b/src/UI/Cells/cells.less
@@ -170,7 +170,7 @@ td.episode-status-cell, td.quality-cell, td.history-quality-cell, td.progress-ce
   width : 80px;
 }
 
-td.delete-episode-file-cell {
+td.delete-episode-file-cell, td.file-link-cell {
   .clickable();
 
   text-align : center;
--- a/src/UI/Content/icons.less
+++ b/src/UI/Content/icons.less
@@ -236,6 +236,10 @@
   .fa-icon-color(@brand-danger);
 }
 
+.icon-open-file {
+  .fa-icon-content(@fa-var-external-link);
+}
+
 .icon-sonarr-search {
   .fa-icon-content(@fa-var-search);
 }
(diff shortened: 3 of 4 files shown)


src/UI/Cells/EpisodeStatusCell.js
```diff
@@ -53,11 +53,21 @@ module.exports = NzbDroneCell.extend({
...
                 } else {
-                    this.$el.html('<span class="badge" title="{0}">{1}</span>'.format(title, quality.quality.name));
+                    this.$el.html('<span class="badge"><span title="{0}">{1}</span></span>'.format(title, quality.quality.name));
                 }
 
+                var externalLink = this.episodeFile.get('path').split('/').map(encodeURIComponent).join('/');
+                var replace, replaceWith;
+                // Network accessible path to /mnt directory
+                replace = /^\/mnt\//;
+                replaceWith = "smb://192.168.1.1/main/";
```
larsjohnsen: These are of course dependent on the user's setup, and must be configured.

larsjohnsen:
This could be moved to `episode-file-info` since it is a feature which are dependent on external applications and must be configured.

src/UI/Cells/EpisodeStatusCell.js
```diff
@@ -53,11 +53,21 @@ module.exports = NzbDroneCell.extend({
...
 
                 if (this.episodeFile.get('qualityCutoffNotMet')) {
-                    this.$el.html('<span class="badge badge-inverse" title="{0}">{1}</span>'.format(title, quality.quality.name));
+                    this.$el.html('<span class="badge badge-inverse"><span title="{0}">{1}</span></span>'.format(title, quality.quality.name));
                 } else {
-                    this.$el.html('<span class="badge" title="{0}">{1}</span>'.format(title, quality.quality.name));
+                    this.$el.html('<span class="badge"><span title="{0}">{1}</span></span>'.format(title, quality.quality.name));
                 }
 
+                var externalLink = this.episodeFile.get('path').split('/').map(encodeURIComponent).join('/');
```
larsjohnsen: `encodeURIComponent` is likely not necessary.

markus101:
This highly depends on the users computer supporting opening links and the links being accessible from a remote location.

This type of thing is better suited as a plugin than built into Sonarr (its come up before and that was our stance). UI wise it also doesn't fit, it looks forced.

larsjohnsen:
I modified the placement of the link:
image

src/UI/Cells/FileLinkCell.js
```diff
@@ -0,0 +1,19 @@
+var vent = require('vent');
+var Backgrid = require('backgrid');
+
+module.exports = Backgrid.Cell.extend({
+    className : 'file-link-cell',
+
+    render : function() {
+        // Network accessible path to /mnt directory
+        var replace = /^\/mnt\//;
```
markus101: This is specific to your setup, how are we going to handle this for everyone?

markus101:
The location was the least of my concerns, this process relies heavily on the setup of the client computer, yours is setup to handle smb:// links, but the other thousands of users likely aren't.

Personally I think the support of this is going to far outweigh the usefulness of it, for every user that gets it working I suspect there will be hundreds that don't understand how to make it work.

I think this is a perfect candidate for a browser plugin, it would just need to insert the HTML into the DOM and could also take care of the path translations that will be inevitable.

src/UI/Cells/FileLinkCell.js
larsjohnsen: As I wrote in an outdated diff: `These variables are of course dependent on the user's setup, and must be configurable.`. So the user would have to discover the correct replacement themselves.

larsjohnsen:
I agree on all points.

larsjohnsen:
I amended the commit to make the link generation user-specifiable (saved in LocalStorage, as the pattern usually depends on the client environment). It should be noted that the transformation is naive; it uses a simple string search and replace. All in all, the method for making this work is obviously not user friendly.

I found a nice extension for Chrome which can be used to open the file directly in the browser by clicking on the link: https://chrome.google.com/webstore/detail/locallinks/jllpkdkcdjndhggodimiphkghogcpida

A few pointers on getting this working:
- If Sonarr is on local computer
 - replace: ""
 - replaceWith: "file://"
- If Sonarr is a networked device (accessible on IP 192.168.1.1, has a Samba share "tv" which translates to local path D:\TV\
 - replace: "D:\TV\"
 - replaceWith: "file://192.168.1.1/tv/"

Guruleenyc:
Thanks! Will give it a shot...

ciao

On Sat, Apr 11, 2015 at 8:36 PM, larsjohnsen <notifications@github.com>
wrote:

> I amended the commit to make the link generation user-specifiable (saved
> in LocalStorage, as the pattern usually depends on the client environment).
> It should be noted that the transformation is naive; it uses a simple
> string search and replace. All in all, the method for making this work is
> obviously not user friendly.
>
> I found a nice extension for Chrome which can be used to open the file
> directly in the browser by clicking on the link:
> https://chrome.google.com/webstore/detail/locallinks/jllpkdkcdjndhggodimiphkghogcpida
>
> A few pointers on getting this working:
>
>    - If Sonarr is on local computer
>       - replace: ""
>       - replaceWith: "file://"
>     - If Sonarr is a networked device (accessible on IP 192.168.1.1, has
>    a Samba share "tv" which translates to local path D:\TV\
>       - replace: "D:\TV\"
>       - replaceWith: "file://192.168.1.1/tv/"
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/Sonarr/Sonarr/pull/219#issuecomment-91954619>.
>



-- 
_____________________
Lee
MapMyFitness: http://tinyurl.com/k9mxhgw

markus101:
This is best implemented as a browser plugin, since one is required for links to work properly anyways, it would be trivial in inject the link into the DOM.

PR #220, opened 2015-03-05
 Pushbullet: Retrieval of device ids

larsjohnsen:
This populates a list with device ids, and automatically sets its value in the "device id" field.

image

### DIFF
--- a/src/UI/Settings/Notifications/Edit/NotificationEditView.js
+++ b/src/UI/Settings/Notifications/Edit/NotificationEditView.js
@@ -1,5 +1,7 @@
 var vent = require('vent');
 var Marionette = require('marionette');
+var $ = require('jquery');
+var Messenger = require('../../../Shared/Messenger');
 var DeleteView = require('../Delete/NotificationDeleteView');
 var AsModelBoundView = require('../../../Mixins/AsModelBoundView');
 var AsValidatedView = require('../../../Mixins/AsValidatedView');
@@ -32,6 +34,65 @@ var view = Marionette.ItemView.extend({
         this.ui.tags.tagInput({
             model    : this.model,
             property : 'tags'
+        });
+
+        if ('PushBullet' === this.model.get('implementation')) {
+            this._pushbullet();
+        }
+    },
+
+    _pushbullet : function() {
+        var $pushbullet = this.$('.x-pushbullet').show();
+        var $deviceIds = $pushbullet.find(".x-pushbullet-deviceIds");
+        var $button = $pushbullet.find("button");
+        var $deviceId = this.$("input[validation-name=DeviceId]");
+        var $accessToken = this.$("input[validation-name=ApiKey]");
+
+        function populateList(response) {
+            // 'All devices' in not included in the response, so prepopulate the array
+            var devices = [{name: 'All devices', id: ''}];
+
+            response.devices
+              .filter(function(v) { return v.pushable; })
+              .forEach(function(v) {
+                  devices.push({
+                    name: v.nickname + (v.active ? "" : " (Inactive}"),
+                    id: v.iden
+                  });
+              });
+
+            var $options = devices.map(function(v) {
+                return $("<option>").html(v.name).val(v.id);
+            });
+
+            $deviceIds.html($options).change();
+        }
+
+        function getDevices() {
+            // Use the entered API key as access token
+            var headers = { "Authorization": "Bearer " + $accessToken.val() };
+            $.ajax({
+                type: 'GET',
+                url: 'https://api.pushbullet.com/v2/devices',
+                headers: headers,
+                success: populateList,
+                error: function(xhr) {
+                    var error = 'Unknown.';
+                    if (xhr.responseJSON && xhr.responseJSON.error && xhr.responseJSON.error.message) {
+                        error = xhr.responseJSON.error.message;
+                    }
+
+                    Messenger.show({
+                        type    : 'error',
+                        message : 'Could not retrieve devices IDs: ' + error
+                    });
+                }
+            });
+        }
+
+        $button.click(getDevices);
+        $deviceIds.change(function() {
+            $deviceId.val(this.value);
         });
     },
 
--- a/src/UI/Settings/Notifications/Edit/NotificationEditViewTemplate.hbs
+++ b/src/UI/Settings/Notifications/Edit/NotificationEditViewTemplate.hbs
@@ -98,6 +98,14 @@
             <hr>
 
             {{formBuilder}}
+
+            <div class="x-pushbullet form-group" style="display: none">
+                  <label class="col-sm-3 control-label">Select device ID</label>
+                  <div class="col-sm-5 form-inline">
+                      <select class="x-pushbullet-deviceIds form-control"></select>
+                      <button class="btn">Populate list</button>
+                  </div>
+            </div>
         </div>
     </div>
     <div class="modal-footer">
--- a/src/UI/jQuery/jquery.ajax.js
+++ b/src/UI/jQuery/jquery.ajax.js
@@ -2,9 +2,7 @@ module.exports = function() {
 
     var $ = this;
 
-    var original = $.ajax;
-    $.ajax = function(xhr) {
-        'use strict';
+    function modifyLocalRequest(xhr) {
         if (xhr && xhr.data && xhr.type === 'DELETE') {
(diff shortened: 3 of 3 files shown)


src/UI/Settings/Notifications/Edit/NotificationEditView.js
```diff
@@ -13,7 +13,8 @@ var view = Marionette.ItemView.extend({
     ui : {
         onDownloadToggle : '.x-on-download',
         onUpgradeSection : '.x-on-upgrade',
-        tags             : '.x-tags'
+        tags             : '.x-tags',
+        pushbullet       : '#x-pushbullet'
```
markus101: Don't use an ID to select, follow the 'x-class' style we're already using.

src/UI/Settings/Notifications/Edit/NotificationEditView.js
```diff
@@ -33,6 +34,41 @@ var view = Marionette.ItemView.extend({
...
+
+    _ifPushbullet : function() {
+        var $deviceIds = this.ui.pushbullet.find("#x-pushbullet-deviceIds");
+        var $button = this.ui.pushbullet.find("button");
+        var $deviceId = this.$("input[validation-name=DeviceId]");
+        var $accessToken = this.$("input[validation-name=ApiKey]");
+
+        function getDevices() {
+            var accessToken = $accessToken.val();
+            $.ajax({
```
markus101: This will leak the Sonarr API key since its injected in all JSON requests.

src/UI/Settings/Notifications/Edit/NotificationEditView.js
```diff
@@ -33,6 +34,41 @@ var view = Marionette.ItemView.extend({
...
             property : 'tags'
         });
+
+        if ('PushBullet' === this.model.get('implementation')) {
+            this._ifPushbullet();
+        }
+    },
+
+    _ifPushbullet : function() {
+        var $deviceIds = this.ui.pushbullet.find("#x-pushbullet-deviceIds");
```
markus101: Use an 'x-class' selector.

src/UI/Settings/Notifications/Edit/NotificationEditView.js
```diff
@@ -33,6 +34,41 @@ var view = Marionette.ItemView.extend({
...
+        var $button = this.ui.pushbullet.find("button");
+        var $deviceId = this.$("input[validation-name=DeviceId]");
+        var $accessToken = this.$("input[validation-name=ApiKey]");
+
+        function getDevices() {
+            var accessToken = $accessToken.val();
+            $.ajax({
+                type: 'GET',
+                url: 'https://api.pushbullet.com/v2/devices',
+                beforeSend: function(xhr) { xhr.setRequestHeader("Authorization", "Bearer " + accessToken); }
```
markus101: Use the headers object.

```
headers : {
    Authorization : "Bearer " + accessToken
}
```

src/UI/Settings/Notifications/Edit/NotificationEditView.js
```diff
@@ -33,6 +34,41 @@ var view = Marionette.ItemView.extend({
...
+            }).success(function(r) {
+                var $options = r.devices
+                    .filter(function(v) { return v.pushable; })
+                    .concat([{iden: '', active: true, nickname: 'All devices' }]) // Add extra option
+                    .map(function(v) { return $("<option>").val(v.iden).text(v.nickname + (v.active ? "" : "(Inactive)")); });
+                $deviceIds.html($options);
+            }).error(function() {
+                alert("Could not retrieve device ids.");
```
markus101: Use Messenger instead of a browser alert window.

src/UI/Settings/Notifications/Edit/NotificationEditView.js
```diff
@@ -33,6 +34,41 @@ var view = Marionette.ItemView.extend({
...
+            $.ajax({
+                type: 'GET',
+                url: 'https://api.pushbullet.com/v2/devices',
+                beforeSend: function(xhr) { xhr.setRequestHeader("Authorization", "Bearer " + accessToken); }
+            }).success(function(r) {
+                var $options = r.devices
+                    .filter(function(v) { return v.pushable; })
+                    .concat([{iden: '', active: true, nickname: 'All devices' }]) // Add extra option
+                    .map(function(v) { return $("<option>").val(v.iden).text(v.nickname + (v.active ? "" : "(Inactive)")); });
+                $deviceIds.html($options);
```
markus101: Not a fan of this block, it mashes together processing and DOM manipulation and is awkward to follow/read.

src/UI/Settings/Notifications/Edit/NotificationEditView.js
```diff
@@ -33,6 +34,41 @@ var view = Marionette.ItemView.extend({
             model    : this.model,
             property : 'tags'
         });
+
+        if ('PushBullet' === this.model.get('implementation')) {
+            this._ifPushbullet();
+        }
+    },
+
+    _ifPushbullet : function() {
```
markus101: Method name should describe what it does, not when its triggered.

src/UI/Settings/Notifications/Edit/NotificationEditView.js
markus101: Also not sure if this belongs as a selector for the whole view since its very specific to PushBullet.

src/UI/Settings/Notifications/Edit/NotificationEditViewTemplate.hbs
```diff
@@ -95,6 +95,14 @@
                 </div>
             </div>
 
+            <div id="x-pushbullet" class="form-group" style="display: none">
+                <label class="col-sm-3 control-label">Select device ID</label>
+                <div class="col-sm-5 form-inline">
+                  <select id="x-pushbullet-deviceIds" class="form-control"></select>
+                  <button class="btn">Populate list</button>
```
markus101: Personally I think the button should be something like an icon with a refresh button, maybe even grouped with the input so they share the one edge (input-group) http://getbootstrap.com/components/#input-groups - not sure how this will look for a select though.

markus101: Also ATM this button does nothing.

markus101:
This is definitely something we wanted to add, but I wonder if the logic to fetch it is better suited to a sever side implementation, it would be more work, but doesn't require CORS and could be built similar to how we do the form builder (I think, haven't given it too much thought).

src/UI/Settings/Notifications/Edit/NotificationEditView.js
kayone: this is the base view, I'm not sure if we want to have pushbullet specific code in here. the alternative would be to just send all the devices back from the server.

larsjohnsen:
I've amended the commit based on your feedback. All diff comments have unfortunately been hidden, although some are still relevant.

src/UI/jQuery/jquery.ajax.js
```diff
@@ -14,7 +14,7 @@ module.exports = function() {
             xhr.url += $.param(xhr.data);
             delete xhr.data;
         }
-        if (xhr) {
+        if (xhr && !(xhr.headers instanceof Object && Object.isFrozen(xhr.headers))) {
```
larsjohnsen: It's likely better to create an alternative ajax function, such as $.ajaxOriginal(), for these cases.

larsjohnsen:
Further changes may be:
- Hiding the device ID text value field
- Creating a two step solution for retrieving device ids
-- Write API key -> click OK -> retrieve device ids -> (if valid access token && retrival sucessful) select device

I was not able to find any other example of a specific client / indexer / notifier modifying the behavior of the form, so that should probably be standardized.

thezoggy:
when you prevent it to the user, i would show `devices[i].manufacturer + ' ' + devices[i].nickname`
can play around with how i did it in sb if you want to mimic that

larsjohnsen:
@thezoggy I believe `manufacturer` is included in nickname by default.

src/UI/jQuery/jquery.ajax.js
larsjohnsen: Or, a simpler way: Only inject when url is the sonarr server.

markus101:
Closing this in favour of getting the device IDs through the server,

PR #221, opened 2015-03-09
selecting range with shift-key in Series Editor should now work

Taloth:
@markus101

A hack imo.

Problem is that this.model.collection points to the original unsorted collection and thus messes up the indexes.
Strangely enough I couldn't find a collection property on the cell, so this hack piggybacks it via column.

Any suggestions are welcome.

### DIFF
--- a/src/UI/Cells/SelectAllCell.js
+++ b/src/UI/Cells/SelectAllCell.js
@@ -4,14 +4,16 @@ var BackgridSelectAll = require('backgrid.selectall');
 
 module.exports = BackgridSelectAll.extend({
     enterEditMode : function(e) {
-        if (e.shiftKey && this.model.collection.lastToggled) {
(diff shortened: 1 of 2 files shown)


src/UI/Cells/SelectAllCell.js
```diff
@@ -38,7 +40,11 @@ module.exports = BackgridSelectAll.extend({
...
         });
 
-        this.model.collection.lastToggled = undefined;
-        this.model.collection.checked = undefined;
+        collection.lastToggled = undefined;
+        collection.checked = undefined;
+    },
+
+    _getCollection : function() {
+        return this.column.get('sortedCollection') || this.model.collection;
```
markus101: Inline this method and pass the collection to _selectRange instead of getting it again.

PR #222, opened 2015-03-10
Download clients: New client rTorrent

larsjohnsen:
The client must be accessible via HTTP, usually via a scgi proxy such as Nginx. For more info, see https://github.com/rakshasa/rtorrent/wiki/RPC-Setup-XMLRPC.

- Can be improved
 - Magnets must currently be resolved before adding priority / label (however: whether the torrent get resolved is often indicative of swarm health)
 - Absolute download path must be specified for post-processing to work

### DIFF
--- a/src/NzbDrone.Core/Download/Clients/rTorrent/RTorrent.cs
+++ b/src/NzbDrone.Core/Download/Clients/rTorrent/RTorrent.cs
@@ -1,0 +1,214 @@
+﻿using System;
+using System.Linq;
+using System.Collections.Generic;
+using System.Threading;
+using NzbDrone.Common.Disk;
+using NzbDrone.Common.Extensions;
+using NzbDrone.Common.Http;
+using NzbDrone.Core.Configuration;
+using NzbDrone.Core.MediaFiles.TorrentInfo;
+using NLog;
+using NzbDrone.Core.Validation;
+using FluentValidation.Results;
+using System.Net;
+using NzbDrone.Core.Parser.Model;
+using NzbDrone.Core.RemotePathMappings;
+
+namespace NzbDrone.Core.Download.Clients.RTorrent
+{
+    public class RTorrent : TorrentClientBase<RTorrentSettings>
+    {
+        private readonly IRTorrentProxy _proxy;
+
+        public RTorrent(IRTorrentProxy proxy,
+                        ITorrentFileInfoReader torrentFileInfoReader,
+                        IHttpClient httpClient,
+                        IConfigService configService,
+                        IDiskProvider diskProvider,
+                        IRemotePathMappingService remotePathMappingService,
+                        Logger logger)
+            : base(torrentFileInfoReader, httpClient, configService, diskProvider, remotePathMappingService, logger)
+        {
+            _proxy = proxy;
+        }
+
+        protected override string AddFromMagnetLink(RemoteEpisode remoteEpisode, string hash, string magnetLink)
+        {
+            _proxy.AddTorrentFromUrl(magnetLink, Settings);
+
+            // Wait until url has been resolved before returning
+            var TRIES = 5;
+            var RETRY_DELAY = 500; //ms
+            var ready = false;
+            for (var i = 0; i < TRIES; i++)
+            {
+                ready = _proxy.HasHashTorrent(hash, Settings);
+                if (ready)
+                {
+                    break;
+                }
+
+                Thread.Sleep(RETRY_DELAY);
+            }
+
+            if (ready)
+            {
+                _proxy.SetTorrentLabel(hash, Settings.TvCategory, Settings);
+
+                var priority = (RTorrentPriority)(remoteEpisode.IsRecentEpisode() ?
+                                        Settings.RecentTvPriority : Settings.OlderTvPriority);
+                _proxy.SetTorrentPriority(hash, Settings, priority);
+
+                return hash;
+            }
+            else
+            {
+                _logger.Debug("Magnet {0} could not be resolved in {1} tries at {2} ms intervals.", magnetLink, TRIES, RETRY_DELAY);
+                // Remove from client, since it is discarded
+                RemoveItem(hash, true);
+
+                return null;
+            }
+        }
+
+        protected override string AddFromTorrentFile(RemoteEpisode remoteEpisode, string hash, string filename, byte[] fileContent)
+        {
+            _proxy.AddTorrentFromFile(filename, fileContent, Settings);
+
+            _proxy.SetTorrentLabel(hash, Settings.TvCategory, Settings);
+
+            var priority = (RTorrentPriority)(remoteEpisode.IsRecentEpisode() ?
+                                    Settings.RecentTvPriority : Settings.OlderTvPriority);
+            _proxy.SetTorrentPriority(hash, Settings, priority);
+
+            return hash;
+        }
+
+        public override string Name
+        {
+            get
+            {
+                return "rTorrent";
+            }
+        }
+
+        public override IEnumerable<DownloadClientItem> GetItems()
+        {
+            try
+            {
+                var torrents = _proxy.GetTorrents(Settings);
+
+                _logger.Debug("Retrieved metadata of {0} torrents in client", torrents.Count);
+
+                var items = new List<DownloadClientItem>();
+                foreach (RTorrentTorrent torrent in torrents)
+                {
+                    // Don't concern ourselves with categories other than specified
+                    if (torrent.Category != Settings.TvCategory) continue;
+
+                    if (torrent.Path.StartsWith("."))
+                    {
+                        throw new DownloadClientException("Download paths paths must be absolute. Please specify variable \"directory\" in rTorrent.");
+                    }
+
+                    var item = new DownloadClientItem();
+                    item.DownloadClient = Definition.Name;
+                    item.Title = torrent.Name;
+                    item.DownloadId = torrent.Hash;
+                    item.OutputPath = _remotePathMappingService.RemapRemoteToLocal(Settings.Host, new OsPath(torrent.Path));
+                    item.TotalSize = torrent.TotalSize;
+                    item.RemainingSize = torrent.RemainingSize;
+                    item.Category = torrent.Category;
+
+                    if (torrent.DownRate > 0) {
+                        var secondsLeft = torrent.RemainingSize / torrent.DownRate;
+                        item.RemainingTime = TimeSpan.FromSeconds(secondsLeft);
+                    } else {
+                        item.RemainingTime = TimeSpan.Zero;
+                    }
+
+                    if (torrent.IsFinished) item.Status = DownloadItemStatus.Completed;
+                    else if (torrent.IsActive) item.Status = DownloadItemStatus.Downloading;
+                    else if (!torrent.IsActive) item.Status = DownloadItemStatus.Paused;
+
+                    // Since we do not know the user's intent, do not let Sonarr to remove the torrent
+                    item.IsReadOnly = true;
+
+                    items.Add(item);
+                }
+
+                return items;
+            }
+            catch (DownloadClientException ex)
+            {
+                _logger.ErrorException(ex.Message, ex);
+                return Enumerable.Empty<DownloadClientItem>();
+            }
+
+        }
+
+        public override void RemoveItem(string downloadId, bool deleteData)
+        {
+            if (deleteData)
+            {
+                DeleteItemData(downloadId);
+            }
+
+            _proxy.RemoveTorrent(downloadId, Settings);
+        }
+
+        public override DownloadClientStatus GetStatus()
+        {
+            // XXX: This function's correctness has not been considered
+
+            var status = new DownloadClientStatus
+            {
+                IsLocalhost = Settings.Host == "127.0.0.1" || Settings.Host == "localhost"
+            };
+
+            return status;
+        }
+
+        protected override void Test(List<ValidationFailure> failures)
+        {
+            failures.AddIfNotNull(TestConnection());
+            if (failures.Any()) return;
+            failures.AddIfNotNull(TestGetTorrents());
+        }
+
+        private ValidationFailure TestConnection()
+        {
+            try
+            {
+                var version = _proxy.GetVersion(Settings);
+
+                if (new Version(version) < new Version("0.9.0"))
+                {
+                    return new ValidationFailure(string.Empty, "rTorrent version should be at least 0.9.0. Version reported is {0}", version);
+                }
+            }
+            catch (Exception ex)
+            {
+                _logger.ErrorException(ex.Message, ex);
+                return new NzbDroneValidationFailure(string.Empty, "Unknown exception: " + ex.Message);
+            }
+
+            return null;
+        }
+
+        private ValidationFailure TestGetTorrents()
+        {
+            try
+            {
+                _proxy.GetTorrents(Settings);
+            }
+            catch (Exception ex)
+            {
+                _logger.ErrorException(ex.Message, ex);
+                return new NzbDroneValidationFailure(string.Empty, "Failed to get the list of torrents: " + ex.Message);
+            }
+
+            return null;
+        }
+    }
+}
--- a/src/NzbDrone.Core/Download/Clients/rTorrent/RTorrentPriority.cs
+++ b/src/NzbDrone.Core/Download/Clients/rTorrent/RTorrentPriority.cs
@@ -1,0 +1,10 @@
+namespace NzbDrone.Core.Download.Clients.RTorrent
+{
+    public enum RTorrentPriority
+    {
+        DoNotDownload = 0,
+        Low = 1,
+        Normal = 2,
+        High = 3
+    }
+}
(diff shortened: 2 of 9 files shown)


larsjohnsen:
For some reason is the torrent removed once the file is imported. I would not think that shall happen, given my settings:
```
Completed Download Handling
Enabled: true
Remove: false
```

Any thoughts on what is wrong?

src/NzbDrone.Core/Download/Clients/rTorrent/RTorrent.cs
```diff
@@ -0,0 +1,178 @@
...
+
+            // Wait until url has been resolved before returning
+            var remainingTries = 5;
+            var retryDelay = 500; //ms
+            var ready = false;
+            while (remainingTries-- > 0)
+            {
+                if (ready = _proxy.IsHashReadyTorrent(hash, Settings)) break;
+                else Thread.Sleep(retryDelay);
+            }
```
Taloth: Obviously not too fond if this delay loop. Am I correct to assume that you cannot set the label and priorities until it retrieved the torrent from DHT?

src/NzbDrone.Core/Download/Clients/rTorrent/RTorrent.cs
```diff
@@ -0,0 +1,178 @@
...
+        {
+            _proxy.AddTorrentFromUrl(magnetLink, Settings);
+
+            // Wait until url has been resolved before returning
+            var remainingTries = 5;
+            var retryDelay = 500; //ms
+            var ready = false;
+            while (remainingTries-- > 0)
+            {
+                if (ready = _proxy.IsHashReadyTorrent(hash, Settings)) break;
```
Taloth: if with parenthesis, same for else.

src/NzbDrone.Core/Download/Clients/rTorrent/RTorrent.cs
```diff
@@ -0,0 +1,178 @@
...
+                    if (torrent.RemainingTime != -1) item.RemainingTime = TimeSpan.FromSeconds(torrent.RemainingTime);
+                    item.Category = torrent.Category;
+
+                    if (torrent.IsFinished) item.Status = DownloadItemStatus.Completed;
+                    else if (torrent.IsActive) item.Status = DownloadItemStatus.Downloading;
+                    else if (!torrent.IsActive) item.Status = DownloadItemStatus.Paused;
+                    else if (!torrent.IsOpen) item.Status = DownloadItemStatus.Queued;
+
+                    if (item.Status != DownloadItemStatus.Completed) item.IsReadOnly = true;
+                    // FIXME Set conditions for  isreadonly = false
```
Taloth: IsReadOnly = finished seeding and okay to delete

src/NzbDrone.Core/Download/Clients/rTorrent/RTorrent.cs
```diff
@@ -0,0 +1,178 @@
...
+            var remainingTries = 5;
+            var retryDelay = 500; //ms
+            var ready = false;
+            while (remainingTries-- > 0)
+            {
+                if (ready = _proxy.IsHashReadyTorrent(hash, Settings)) break;
+                else Thread.Sleep(retryDelay);
+            }
+           
+            if (!ready)
```
Taloth: Log failure via _logger.Debug(...)

src/NzbDrone.Core/Download/Clients/rTorrent/RTorrent.cs
```diff
@@ -0,0 +1,178 @@
...
+        public override IEnumerable<DownloadClientItem> GetItems()
+        {
+            try
+            {
+                var torrents = _proxy.GetTorrents(Settings);
+
+                var items = new List<DownloadClientItem>();
+                foreach (RTorrentTorrent torrent in torrents)
+                {
+                    var item = new DownloadClientItem();
```
Taloth: Filter on category as early as possible.

src/NzbDrone.Core/Download/Clients/rTorrent/RTorrent.cs
```diff
@@ -0,0 +1,178 @@
...
+            {
+                _logger.ErrorException(ex.Message, ex);
+                return Enumerable.Empty<DownloadClientItem>();
+            }
+
+        }
+
+        public override void RemoveItem(string downloadId, bool deleteData)
+        {
+            _proxy.RemoveTorrent(downloadId, deleteData, Settings);
```
Taloth: Move deleteData logic out of the proxy, and handle the deletion yourself.

src/NzbDrone.Core/Download/Clients/rTorrent/RTorrent.cs
```diff
@@ -0,0 +1,178 @@
...
+
+        private ValidationFailure TestConnection()
+        {
+            try
+            {
+                var version = _proxy.GetVersion(Settings);
+
+                if (new Version(version) < new Version("0.9.0"))
+                {
+                    return new ValidationFailure(string.Empty, "RTorrent version should be at least 0.9.0");
```
Taloth: Debug log the returned version.

src/NzbDrone.Core/Download/Clients/rTorrent/RTorrentPriority.cs
```diff
@@ -0,0 +1,10 @@
+namespace NzbDrone.Core.Download.Clients.RTorrent
+{
+    public enum RTorrentPriority
+    {
+        High = 3,
+        Normal = 2,
+        Low = 1,
+        DoNotDownload = 0
```
Taloth: Reverse the enum values.

src/NzbDrone.Core/Download/Clients/rTorrent/RTorrentProxy.cs
```diff
@@ -0,0 +1,212 @@
...
+                item.BaseDir = (String)torrent[2];
+                item.TotalSize = (Int64)torrent[3];
+                item.RemainingSize = (Int64)torrent[4];
+                item.Category = category;
+                item.IsOpen = Convert.ToBoolean((Int64)torrent[7]);
+                item.IsActive = Convert.ToBoolean((Int64)torrent[8]);
+                item.IsFinished = Convert.ToBoolean((Int64)torrent[9]);
+
+                var down_rate = (Int64)torrent[5];
+                item.RemainingTime = (down_rate > 0 && item.IsActive) ? (item.RemainingSize / down_rate) : -1;
```
Taloth: proxy shouldn't do these calculations, it's a shallow interface. Return the down_rate in a property on RTorrentTorrent and do the conversion in RTorrent class.

src/NzbDrone.Core/Download/Clients/rTorrent/RTorrentProxy.cs
```diff
@@ -0,0 +1,212 @@
...
+            }
+            return items;
+        }
+
+        /*
+         * It is only ready if:
+         *  - It exists
+         *  - The name does not match "{$hash}.meta"
+         */
+        public bool IsHashReadyTorrent(String hash, RTorrentSettings settings)
```
Taloth: Remove comment and change method to something _like_ HasMagnetLinkResolved, not sure yet, but that way you don't need a comment.

src/NzbDrone.Core/Download/Clients/rTorrent/RTorrentProxy.cs
```diff
@@ -0,0 +1,212 @@
...
+        public RTorrentProxy(Logger logger)
+        {
+            _logger = logger;
+        }
+
+        public string GetVersion(RTorrentSettings settings)
+        {
+            var client = BuildClient(settings);
+            var version = client.GetVersion();
+            return version;
```
Taloth: Find a way to log the various api calls to Debug.

src/NzbDrone.Core/Download/Clients/rTorrent/RTorrentSettings.cs
```diff
@@ -0,0 +1,65 @@
...
+        }
+
+        [FieldDefinition(0, Label = "Host", Type = FieldType.Textbox)]
+        public String Host { get; set; }
+
+        [FieldDefinition(1, Label = "Port", Type = FieldType.Textbox)]
+        public Int32 Port { get; set; }
+
+        [FieldDefinition(2, Label = "Path", Type = FieldType.Textbox, HelpText = "Can be found in the URL: {protocol}://{host}:{port}/{path}")]
+        public String Path { get; set; }
```
Taloth: I believe we call this UrlBase elsewhere. also try mimic the helptext.

Taloth:
we had a short while that we used String,Int32 and such. But we're back to string,int,long etc, and will replace em as we go.
So just use string,int etc.

larsjohnsen:
Thanks for your feedback.

```Am I correct to assume that you cannot set the label and priorities until it retrieved the torrent from DHT?```
It is possible to set priority and label on a magnet, but they do not automatically transfer when the magnet has been resolved and the torrent is created.

File: src/NzbDrone.Core/Download/Clients/rTorrent/RTorrent.cs (hunk 3 above)
Comment (larsjohnsen) [inline src/NzbDrone.Core/Download/Clients/rTorrent/RTorrent.cs]: I'm uncertain what you mean by delete here. If it is queued to seed, wouldn't the user want to keep it in the client and not have it deleted?

Comment (Taloth) [inline src/NzbDrone.Core/Download/Clients/rTorrent/RTorrent.cs]: Yes, so then IsReadOnly=true

**Comment by markus101** [comment]
Whats left for this one?

**Comment by Taloth** [comment]
Quite a few things:
- IsReadOnly isn't set properly. Now it's always true so the torrent will never be deleted.
- deleteData needs to be implemented. Got a helper for that in the base class ```DeleteItemData``` since recently.
- Prolly gotta look over the code styling coz it doesn't match the rest of the code, not a big deal.
- No tests, that's the biggest deal I guess.

File: src/NzbDrone.Core/Download/Clients/rTorrent/RTorrent.cs [hunk 13]
```diff
@@ -0,0 +1,217 @@
...
+                        item.RemainingTime = TimeSpan.FromSeconds(secondsLeft);
+                    }
+
+                    if (torrent.IsFinished) item.Status = DownloadItemStatus.Completed;
+                    else if (torrent.IsActive) item.Status = DownloadItemStatus.Downloading;
+                    else if (!torrent.IsActive) item.Status = DownloadItemStatus.Paused;
+                    else if (!torrent.IsOpen) item.Status = DownloadItemStatus.Queued;
+
+                    // Since we do not know the user's intent, do not let the Sonarr to remove torrent
+                    item.IsReadOnly = true;
```
larsjohnsen: This variable is always declared `true` since I did not find any way to reliably detect when user want the torrent removed.

markus101: When the torrent is complete and has been stopped is when you want to toggle IsReadOnly to false, as-is this breaks automated removal when seeding goals are reached.

larsjohnsen: While true, there does not seem to be any globally defined seeding goals in rTorrent.

markus101: Whether they exist or not is irrelevant here, when the torrent is 100% downloaded and stopped we want read-only off. That behavior will then match all other torrent clients.

markus101: @Taloth disagrees with me here due to a couple compelling reasons, so ignore me, we can deal with this later.

src/NzbDrone.Core/Download/Clients/rTorrent/RTorrent.cs
```diff
@@ -0,0 +1,217 @@
...
+                        var secondsLeft = torrent.RemainingSize / torrent.DownRate;
+                        item.RemainingTime = TimeSpan.FromSeconds(secondsLeft);
+                    }
+
+                    if (torrent.IsFinished) item.Status = DownloadItemStatus.Completed;
+                    else if (torrent.IsActive) item.Status = DownloadItemStatus.Downloading;
+                    else if (!torrent.IsActive) item.Status = DownloadItemStatus.Paused;
+                    else if (!torrent.IsOpen) item.Status = DownloadItemStatus.Queued;
+
+                    // Since we do not know the user's intent, do not let the Sonarr to remove torrent
```
markus101: Can you elaborate on this comment? Their intent in which regard? And a small grammatical error, the second part should be `do not let the Sonarr remove the torrent`

src/NzbDrone.Core/Download/Clients/rTorrent/RTorrent.cs
```diff
@@ -0,0 +1,217 @@
...
+
+        private ValidationFailure TestConnection()
+        {
+            try
+            {
+                var version = _proxy.GetVersion(Settings);
+
+                if (new Version(version) < new Version("0.9.0"))
+                {
+                    return new ValidationFailure(string.Empty, "RTorrent version should be at least 0.9.0. Version reported is {0}", version);
```
markus101: is it `rTorrent` or `RTorrent`?

src/NzbDrone.Core/Download/Clients/rTorrent/RTorrent.cs
Taloth: Hm, I don't know if rTorrent pauses/stops torrents if seeding slots are fully used. So just deleting it isn't a good idea, with risk of hnr and all.
For Deluge I actually have some compare logic that compares the ratio with deluge own config. Mainly because Deluge would 'pause' torrents that finished seeding, and doesn't have a stopped state.

Dunno how that it for rTorrent. This is pretty much why we need tests. So we know how rTorrent behaves.

src/NzbDrone.Core/Download/Clients/rTorrent/RTorrent.cs
```diff
@@ -0,0 +1,217 @@
...
+        protected override string AddFromMagnetLink(RemoteEpisode remoteEpisode, string hash, string magnetLink)
+        {
+            _proxy.AddTorrentFromUrl(magnetLink, Settings);
+
+            // Wait until url has been resolved before returning
+            var TRIES = 5;
+            var remainingTries = TRIES;
+            var RETRY_DELAY = 500; //ms
+            var ready = false;
+            while (true)
```
markus101: Personally I like:

```
foreach (var i = 0; i < tries; i++)
{
  if (_proxy.HasHashTorrent(hash, Settings))
  {
    break;
  }

  Thread.Sleep(RETRY_DELAY);
}
```

Better, thoughts?

src/NzbDrone.Core/Download/Clients/rTorrent/RTorrent.cs
larsjohnsen: I deleted my earlier comment --

I am uncertain how wise it is to override the clients behavior, since if torrent ratios are defined, then the user would likely also choose what happens with the torrent on reaching the goal.

src/NzbDrone.Core/Download/Clients/rTorrent/RTorrent.cs
larsjohnsen: - If hash has torrent, it will sleep unnecessary
- Variable "ready" is used afterwards, to avoid making this loop too deep

src/NzbDrone.Core/Download/Clients/rTorrent/RTorrent.cs
larsjohnsen: You tell me!

Taloth: Mixed: http://rakshasa.github.io/rtorrent/

The official one seems to be rTorrent.

src/NzbDrone.Core/Download/Clients/rTorrent/RTorrent.cs
markus101: > If hash has torrent, it will sleep unnecessary

No, it will break immediately.

> Variable "ready" is used afterwards, to avoid making this loop too deep

Right, then a two liner for it makes sense.

jkaberg:
Just thought I'd clarify, 

rTorrent: https://github.com/rakshasa/rtorrent
lib(r)torrent: https://github.com/rakshasa/libtorrent (the lib on which rTorrent is built)
libtorrent: http://www.libtorrent.org/ (used by Deluge, qBitorrent and others. Not the same as above)

It's confusing, I know :smile: 

Looking forward to see this implemented

larsjohnsen:
Updated to include some tests.

src/NzbDrone.Core/Download/Clients/rTorrent/RTorrent.cs
```diff
@@ -0,0 +1,214 @@
...
+                return Enumerable.Empty<DownloadClientItem>();
+            }
+
+        }
+
+        public override void RemoveItem(string downloadId, bool deleteData)
+        {
+            _proxy.RemoveTorrent(downloadId, Settings);
+
+            if (deleteData)
```
Taloth: Turn this around, delete first. Otherwise sonarr won't know what to delete.

Taloth:
Other than that, I think we should get this into develop, have some of our users play with it.

binxycraft:
This is the only thing keeping me from using sonarr exclusively, can't wait for it.

PR #225, opened 2015-03-12
Torznab support

Taloth:
Just based on Newznab, schema for attributes included.

Not sure yet about all attributes so suggestions are welcome. If there's a particular field you've always wanted, here's your chance.

I've included seedtype, minimumratio and minimumseedtime as a mechanism to allow a tracker to specify which seed requirements apply. Specifying it per torrent allows the tracker to change behavior based on type or other criteria.

Also need a better namespace "torznab.local" isn't good. And we won't have much time to come up with a definitive one.


Taloth:
Merged into develop and master

PR #226, opened 2015-03-13
Removed VOSTFR from french parsing

bizen-ya:


markus101:
If I'm not mistaken, VOSTFR subs are hardcoded (or often are at least), given that, this is expected behavior and removing this would be a step backwards for a large number of users (thousands).

This is something we can look at improving later on, but without a clean migration it will cause too many issues.

thezoggy:
yes, please dont allow VOSTFR.

bizen-ya:
I dont think there is a lot of people who want a mix of french and hardsubbed show... You can reject if you want, i have compiled one version for myself.

thezoggy:
sure, your change completely goes against the purpose of its intended effect. classify non english release as such. feel free to close this.

bizen-ya:
I think that a hard subbed release should be categorized as english, but it's m'y own view... I close the request

PR #227, opened 2015-03-18
Allow startup on case sensitive file systems

c0unt0:
Fixed EXE_PATH to match NzbDrone.exe’s capitalisation, so that on mac case sensitive systems, Sonarr boot up without modifications.

### DIFF
--- a/osx/Sonarr
+++ b/osx/Sonarr
@@ -4,7 +4,7 @@
 DIR=$(cd "$(dirname "$0")"; pwd)
  
 #change these values to match your app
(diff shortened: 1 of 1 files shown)


markus101:
For future pull requests please create a feature branch instead of develop.

c0unt0:
Hi,




Thanks for the merge.

And apologies: I didn't read the md before diving in, it's my first step into this kind of things




Thanks,

Alex.

markus101:
Thanks for the contribution! (I should have led with that).

PR #228, opened 2015-03-21
Added support for alternate pid file location

k0ekk0ek:
I've added support for writing the pid file to an alternate location using a /pidfile=PATH command line switch because it's customary to store the pid file in /var/run/<proc>.pid whereas the other files Sonarr writes are better placed in /var/lib (or /var/cache depending on their purpose). The switch is only visible on non-Windows platforms.

### DIFF
--- a/src/NzbDrone.Common/Processes/PidFileProvider.cs
+++ b/src/NzbDrone.Common/Processes/PidFileProvider.cs
@@ -12,16 +12,16 @@ namespace NzbDrone.Common.Processes
 
     public class PidFileProvider : IProvidePidFile
     {
-        private readonly IStartupContext _startupContext;
         private readonly IAppFolderInfo _appFolderInfo;
         private readonly IProcessProvider _processProvider;
+        private readonly IStartupContext _startupContext;
         private readonly Logger _logger;
 
-        public PidFileProvider(IStartupContext startupContext, IAppFolderInfo appFolderInfo, IProcessProvider processProvider, Logger logger)
+        public PidFileProvider(IAppFolderInfo appFolderInfo, IProcessProvider processProvider, IStartupContext startupContext, Logger logger)
         {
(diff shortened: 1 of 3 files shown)


src/NzbDrone.Common/Processes/PidFileProvider.cs
```diff
@@ -12,12 +12,14 @@ public interface IProvidePidFile
 
     public class PidFileProvider : IProvidePidFile
     {
+        private readonly IStartupContext _startupContext;
         private readonly IAppFolderInfo _appFolderInfo;
         private readonly IProcessProvider _processProvider;
         private readonly Logger _logger;
 
-        public PidFileProvider(IAppFolderInfo appFolderInfo, IProcessProvider processProvider, Logger logger)
+        public PidFileProvider(IStartupContext startupContext, IAppFolderInfo appFolderInfo, IProcessProvider processProvider, Logger logger)
```
markus101: A bit pedantic, but the new class should be put at the end of the list (but before logger).

```
public PidFileProvider(IAppFolderInfo appFolderInfo, IProcessProvider processProvider, IStartupContext startupContext, Logger logger)
```

markus101:
Thanks for the PR.

One issue with this (not specific to your changes), is the switches are lost during a restart of Sonarr from the UI. Specific to your changes, you'll want to add this switch to the PreservedArguments in StartupContext: https://github.com/Sonarr/Sonarr/blob/develop/src/NzbDrone.Common/EnvironmentInfo/StartupContext.cs#L67

k0ekk0ek:
Processed the comments and added the /pidfile flag to the list of preserved arguments.

Taloth:
@markus101 Won't the pidpath be lost if restarted via NzbDrone.Update?

k0ekk0ek:
Support for this should be implemented in the configuration file.

PR #232, opened 2015-03-29
Twitter Notifications

halkeye:
Work in progress for Twitter notifications

https://trello.com/c/n80XsydJ/175-twitter-notifications
https://forums.sonarr.tv/t/dev-twitter-notification-connection/4907

This has not been cleaned up fully or rebased yet, its a work in progress but I want feedback as I go.

A couple fixmes:
* Using couchpotato's twitter keys right now, they really should be created by the sonarr group, so I didn't want to make my own
* I have no idea how to get the proper full return url 
* ProviderSettingsModelBase.js needs refractoring as it hard codes step1 and step2 for now (I was thinking pushbullet probably only has a step1 which would return a list of deviceids?)

Note tweetinvi has a large number of dependancies which would break my build because i have .NET 4.5 installed, apparently a BCL issue. I don't know enough about vs/msbuild to fix, tinytwitter and OAuth module seemed easy enough.

Am I on the right track? Crazy? Gotta move things around? Etc


thezoggy:
does it do direct message / mentions ? prob want to also try and save people from themselves on blasting to the world that they are obtaining pirated media....

halkeye:
Not as of right now, I can start adding that. Right now I personally use a private twitter account so its not a high use item for me but I can easily see how people could want it.

src/UI/Settings/Notifications/Edit/NotificationEditView.js
```diff
@@ -11,14 +11,17 @@ var view = Marionette.ItemView.extend({
...
 
     ui : {
-        onDownloadToggle : '.x-on-download',
-        onUpgradeSection : '.x-on-upgrade',
-        tags             : '.x-tags'
+        onDownloadToggle             : '.x-on-download',
+        onUpgradeSection             : '.x-on-upgrade',
+        tags                         : '.x-tags',
+        indicator                    : '.x-indicator',
+        authorizedNotificationButton : '.AuthorizeNotification'
```
markus101: Use `x` classes for events/UI referencing, following the same style, so `.x-authorize-notification`

src/UI/Settings/ProviderSettingsModelBase.js
```diff
@@ -3,6 +3,66 @@ var DeepModel = require('backbone.deepmodel');
 var Messenger = require('../Shared/Messenger');
 
 module.exports = DeepModel.DeepModel.extend({
+    connectData : function(action) {
+        var self = this;
+        
+        this.trigger('connect:sync');
+
+        var promise = $.Deferred();
```
markus101: $.ajax returns a promise, you don't need to create it like this.

src/UI/Settings/Notifications/Edit/NotificationEditView.js
halkeye: I was a little unsure how to handle that. Right now its a FieldAttribute of type button, so the class name is dependant on the field name. Should I just change it to be a boolean type field or something? Hard code the class?

src/UI/Settings/ProviderSettingsModelBase.js
halkeye: Yea, but i didn't want the promise to resolve until all the sub parts were done. For OAuth, you need to do a couple client/server calls.

src/UI/Settings/ProviderSettingsModelBase.js
```diff
@@ -3,6 +3,66 @@ var DeepModel = require('backbone.deepmodel');
...
+
+        var callAction = function(action) {
+            var params = {};
+            params.url = self.collection.url + '/connectData/' + action;
+            params.contentType = 'application/json';
+            params.data = JSON.stringify(self.toJSON());
+            params.type = 'POST';
+            params.isValidatedCall = true;
+
+            $.ajax(params).fail(promise.reject).success(function(response) {
```
markus101: `var promise = $.ajax(options);`

then use `promise.done()` and `promise.fail()` instead.

halkeye: instead of fail and success?

edraven:
Is there any way to use this yet, or is it still a work in progress?

markus101:
Did some cleanup and refactoring and opened a new PR: https://github.com/Sonarr/Sonarr/pull/644

PR #233, opened 2015-04-05
Linux build

giubil:
With those changes on the source code, I am able to build the project on linux.
The changes aren't huge, they are mostly renaming case sensitive files, and adding parameter to ambiguous methods.
Those shouldn't be able to stop windows users from building, although I am not sure.


kayone:
can you rebase this from head of develop, you have some commit in your PR that are already merged.

markus101:
Added changes that allow it to build under mono.

PR #238, opened 2015-04-20
Improved Blacklisting

markus101:
@Taloth @kayone 

- Blacklisting of torrents (manual only until stalled detection is a thing)
- Uses more information about the release to determine if it should be blacklisted (size and indexer in addition to date and name)
- Shows details in the UI as to why it was blacklisted (instead of having to look in history)

### DIFF
--- a/src/NzbDrone.Core/Blacklisting/BlacklistService.cs
+++ b/src/NzbDrone.Core/Blacklisting/BlacklistService.cs
@@ -3,20 +3,22 @@ using System.Linq;
 using NzbDrone.Common.Extensions;
 using NzbDrone.Core.Datastore;
 using NzbDrone.Core.Download;
+using NzbDrone.Core.Indexers;
 using NzbDrone.Core.Messaging.Commands;
 using NzbDrone.Core.Messaging.Events;
+using NzbDrone.Core.Parser.Model;
 using NzbDrone.Core.Tv.Events;
 
 namespace NzbDrone.Core.Blacklisting
 {
     public interface IBlacklistService
     {
-        bool Blacklisted(int seriesId, string sourceTitle, DateTime publishedDate);
+        bool Blacklisted(int seriesId, ReleaseInfo release);
         PagingSpec<Blacklist> Paged(PagingSpec<Blacklist> pagingSpec);
         void Delete(int id);
     }
-
     public class BlacklistService : IBlacklistService,
+
                                     IExecute<ClearBlacklistCommand>,
                                     IHandle<DownloadFailedEvent>,
                                     IHandleAsync<SeriesDeletedEvent>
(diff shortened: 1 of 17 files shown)


src/NzbDrone.Core/Blacklisting/BlacklistService.cs
```diff
@@ -28,11 +30,18 @@ public BlacklistService(IBlacklistRepository blacklistRepository)
...
-        public bool Blacklisted(int seriesId, string sourceTitle, DateTime publishedDate)
+        public bool Blacklisted(int seriesId, ReleaseInfo release)
         {
-            var blacklisted = _blacklistRepository.Blacklisted(seriesId, sourceTitle);
+            var blacklisted = _blacklistRepository.Blacklisted(seriesId, release.Title);
+
+            if (release.DownloadProtocol == DownloadProtocol.Torrent)
+            {
+                return blacklisted.Where(b => b.Protocol == DownloadProtocol.Torrent)
+                                  .Any(b => b.Indexer.Equals(release.Indexer, StringComparison.InvariantCultureIgnoreCase));
```
Taloth: Maybe using InfoHash is better. And fallback on title if missing.

src/NzbDrone.Core/Blacklisting/BlacklistService.cs
```diff
@@ -45,12 +54,48 @@ public void Delete(int id)
...
+        }
+
+        private bool HasSamePublishedDate(Blacklist item, DateTime publishedDate)
         {
             if (!item.PublishedDate.HasValue) return true;
 
-            return item.PublishedDate.Value.AddDays(-2) <= publishedDate &&
-                   item.PublishedDate.Value.AddDays(2) >= publishedDate;
+            return item.PublishedDate.Value.AddMinutes(-2) <= publishedDate &&
+                   item.PublishedDate.Value.AddMinutes(2) >= publishedDate;
```
Taloth: 2 minutes grace period? seems a bit low.

src/NzbDrone.Core/Blacklisting/BlacklistService.cs
markus101: Good idea, to clarify, it would be missing for magnets? or it might be missing depending on the indexer?

src/NzbDrone.Core/Blacklisting/BlacklistService.cs
markus101: I've seen releases from different indexers that are even within 2 minutes that are very likely different releases, whereas the same release is likely within 10 seconds of it.

This definitely goes the other direction from where we were, allowing downloads from the same indexer with the same name and a different post times, but still allowing some grace period on different indexers.

src/NzbDrone.Core/Blacklisting/BlacklistService.cs
Taloth: Actually, magnets contain the infohash.
But no, atm the torrentrss parser only fills in the TorrentReleaseInfo.InfoHash if explicitly provided by the feed. It doesn't atm extract the info has from magnets.

PR #239, opened 2015-04-21
Added Color-Impaired mode to UI settings

Taloth:
Really not happy with the way the setting is initialized and updated in the UI. So could use some advice there.

We prolly should look at some other color coded items as well.


thezoggy:
Destiny just recently did an update to tackle the different colorblindness.. may be able to pull some ideas from: https://www.bungie.net/7_Bungie-Weekly-Update---03192015/en/News/News?aid=12692

PR #240, opened 2015-04-22
Fixed: Pushbullet settings typo

Royal2000H:
Fixed a small typo in a tooltip for the Pushbullet notification settings.


markus101:
FYI, using `Fixed: blah blah blah` means its added to the changelog for that release, which isn't very important when fixing small cosmetic issues such as this.

PR #246, opened 2015-05-07
Prevent the premature deletion of a download when slowly importing season pack

Taloth:
@markus101 

Should fix the problem, but I didn't test it in integration environment.

Also changed IHandleAsync to IHandle, seemed to make sense to do it sync now.

### DIFF
--- a/src/NzbDrone.Core/Download/TrackedDownloads/DownloadMonitoringService.cs
+++ b/src/NzbDrone.Core/Download/TrackedDownloads/DownloadMonitoringService.cs
@@ -12,11 +12,12 @@ using NzbDrone.Core.Messaging.Events;
 namespace NzbDrone.Core.Download.TrackedDownloads
 {
     public class DownloadMonitoringService : IExecute<CheckForFinishedDownloadCommand>,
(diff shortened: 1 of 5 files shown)


src/NzbDrone.Core/Download/TrackedDownloads/DownloadMonitoringService.cs
```diff
@@ -128,12 +144,12 @@ public void Execute(CheckForFinishedDownloadCommand message)
             Refresh();
         }
 
-        public void HandleAsync(EpisodeGrabbedEvent message)
+        public void Handle(EpisodeGrabbedEvent message)
```
markus101: Won't this cause any callers to pause and wait since Handle executes in series, not parallel?

Taloth: No coz all it does it trigger the Debouncer, it won't wait for it, less than a msec work.

markus101:
Ahh, right, not sure what I was thinking... Look good :ship:

PR #248, opened 2015-05-09
Profile editor: Integrate cutoff

larsjohnsen:
New look:
untitled


src/UI/Settings/Profile/Edit/EditProfileItemView.js
```diff
@@ -1,5 +1,21 @@
 var Marionette = require('marionette');
 
 module.exports = Marionette.ItemView.extend({
-    template : 'Settings/Profile/Edit/EditProfileItemViewTemplate'
-});
\ No newline at end of file
+    template : 'Settings/Profile/Edit/EditProfileItemViewTemplate',
+    ui: {
+        cutoff: "input[name=cutoff]"
```
markus101: Use an x-class to select UI elements.

src/UI/Settings/Profile/profile.less
```diff
@@ -30,6 +30,33 @@
   }
 }
 
+ul.ui-sortable {
```
markus101: Follow the existing style, 2 spaces for indents.

src/UI/Settings/Profile/profile.less
```diff
@@ -30,6 +30,33 @@
...
+    }
+
+    .selected {
+        .cutoff {
+            display: inline-block;
+            margin-top: 2px;
+            margin-right: 5px;
+            width: 90px;
+
+            input[name=cutoff] {
```
markus101: Use a class for applying styles.

src/UI/Settings/Profile/profile.less
```diff
@@ -30,6 +30,33 @@
...
+                &:not(:checked)~[notSelected] {
```
markus101: Why do you need these two conditions, its always `display: block`

src/UI/Settings/Profile/profile.less
```diff
@@ -30,6 +30,33 @@
   }
 }
 
+ul.ui-sortable {
+    width: 100%;
+
+    .cutoff > * {
```
markus101: This is a pretty aggressive selector, something more specific should be used.

src/UI/Settings/Profile/Edit/EditProfileItemView.js
```diff
@@ -1,5 +1,21 @@
...
+    },
+    onRender: function() {
+        if (this.model.get("quality").id === this.collectionListView.model.get("cutoff").id) {
+            this.ui.cutoff.prop("checked", true);
+        }
+    },
+    events : {
+        'click .cutoff': function($event) {
```
markus101: Use a named function to handle the click. We use `e` as the event handler.

src/UI/Settings/Profile/Edit/EditProfileItemView.js
```diff
@@ -1,5 +1,21 @@
...
+            this.collectionListView.model.set("cutoff", this.model.get("quality"));
+
+            // Do not let event bubble, as the parent will register the click as select / deselect
```
markus101: This is one thing I don't like about this change, if you're off the mark you toggle the quality off completely.

markus101:
At a minimum this needs validation to ensure you can't have the cutoff set to a quality thats not wanted. Looks like we need had server-side validation for that, but should.

I'm not sure I like this change though, it makes the UI feel cluttered.

Taloth:
I have to agree with markus on the UI look.

When I thought about it earlier I considered creating a 'slider' drag-able control that would highlight the area in which we still perform upgrades.
Kinda like:
image

Just a mockup of course. Needs other colors, text is wrong too.

We wanted to get rid of the term cutoff coz it isn't particularly obvious to the user what it means. Visualizing it is probably better.

larsjohnsen:
Taltoth: I like your idea better. 

An alternative that would require adding a dummy item to the list:
image

thezoggy:
leave cutoff as a drop down, and upon selection visually show where the cutt off is by the bg color change... so they see where it would stop. i think the biggest confusion with the list of qualities is the unknown of which way (top to bottom or bottom to top) does the quality get replaced/matter.

PR #544, opened 2015-05-14
Integrated MediaInfo wrapper to be able to properly handle Unicode

Taloth:
The original fix was a hack, trying to trick the system in accepting UTF-8 but that doesn't work on all systems.

In this revision I integrate the MediaInfo wrapper and properly handle the unicode pinvokes by auto-detecting the wchar size (linux wchar = 4 bytes).
That worked, but it would still fail while running via upstart (no clue why).
So I now took the final step, and use the Buffer api to send the file blocks to MediaInfo instead of filenames. (just the blocks mediainfo requests, not the whole file).
This means that the filename is no longer relevant, it's never sent to MediaInfo anymore.

It also means the unicode changes aren't important anymore, coz all stuff we want from MediaInfo is Ascii, but I left it in for completeness.

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -114,8 +114,8 @@ Function PackageMono()
     get-childitem $outputFolderMono -File -Filter sqlite3.* -Recurse | foreach ($_) {remove-item $_.fullname}
     get-childitem $outputFolderMono -File -Filter MediaInfo.* -Recurse | foreach ($_) {remove-item $_.fullname}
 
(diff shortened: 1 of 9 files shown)


kayone:
I suggest first committing the original version of the wrapper, then making the changes on top, that way the history would actually show what were the changes that were made.

Taloth:
Euh, I picked code from various implementations, half of it is my own.

markus101:
Looks good, if its not coming from a single base implementation (the original code), then I don't see a reason to split it to two commits.

PR #546, opened 2015-05-16
Season pass supports multi-select and new toggling options

markus101:
@Taloth Pretty basic look now (still want to add the seasons with individual toggles, but details of files/episodes at a per season level is going to be expensive (series details does it by fetching all the episodes for that series).

image

I need to fix some tests still.


Taloth:
Thought I already commented on this.

It's possible to modify the SeriesStatistics query quite easily to return per season instead of per series, also pretty fast (~100ms).

Btw. this is one of those ideal spots for a filter box to filter out tags.

markus101:
I'm a bit stuck on what the styling should look like for toggling seasons. I've used labels and increased the font size a bit, but it looks pretty bland: https://www.dropbox.com/s/18hvr84xueelx8p/shot_150521_071116.png?dl=0

Taloth:
Maybe make the badge border without fill, or other color fill, like the cutoff-unmet quality badge.

Wouldn't  S 1 or S01 be better? That way it fits. and gives you some space to add the episodefile/episode statistics data.

Also Season 0 -> Specials.

markus101:
The cutoff-unmet styling does look a lot better:
image


But adding in the statistics makes it pretty messy I think:
image

Thoughts?

Taloth:
image

Taloth:
And last: 
image

markus101:
@Taloth not seeing your CSS locally and getting `.badge is undefined in file C:\Dropbox\Dev\Sonarr\src\UI\SeasonPass\seasonpass.less line no. 9`. I've fixed the tests, but I forget if there is anything else for this one.

PR #547, opened 2015-05-17
Hand-off update logic to handle upstart/systemd restarts

Taloth:
Yes, that easy.

Basically, with this change, we don't terminate the process on linux (linux doesn't lock open files like windows does).
- Then we perform the update.
- Afterwards we kill sonarr.
- Give upstart/systemd/custom 5 seconds to restart it.
- If didn't get restarted we then do it ourselves.

### DIFF
--- a/src/NzbDrone.Update/UpdateEngine/InstallUpdateService.cs
+++ b/src/NzbDrone.Update/UpdateEngine/InstallUpdateService.cs
@@ -79,7 +79,10 @@ namespace NzbDrone.Update.UpdateEngine
 
             try
             {
-                _terminateNzbDrone.Terminate(processId);
+                if (OsInfo.IsWindows)
+                {
+                    _terminateNzbDrone.Terminate(processId);
+                }
 
                 _backupAndRestore.Backup(installationFolder);
                 _backupAppData.Backup();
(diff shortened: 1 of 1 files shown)


src/NzbDrone.Update/UpdateEngine/InstallUpdateService.cs
```diff
@@ -106,7 +109,31 @@ public void Start(string installationFolder, int processId)
...
+                    _startNzbDrone.Start(appType, installationFolder);
+                }
+                else
+                {
+                    _terminateNzbDrone.Terminate(processId);
+
+                    _logger.Info("Waiting for external auto-restart.");
+                    for (int i = 0; i < 5; i++)
+                    {
+                        System.Threading.Thread.Sleep(i * 1000);
```
Taloth: ```i * 1000``` makes me an idiot.

Taloth: On the other hand, it's a nice exponential back-off, checking quick in the beginning with a total of 15 sec.
Maybe that's what I intended when I wrote this yesterday, was kinda late.

markus101: 5 seconds did seem a bit low, but we'd only be waiting 10 seconds, since its wait 0, 1, 2, 3, 4 seconds. Maybe you meant `(i + 1) * 1000` ?

markus101:
Awesome, this is definitely a step in the right direction.

PR #552, opened 2015-05-20
Custom Commands

### LINKED ISSUE #439: Custom Notifications/Scripts - Similar to SAB Post-Processing Scripts
In order to support things that we don't have access to (Popcorn Hour Jukebox) it could be beneficial to allow users to write their own custom scripts to update those platforms.

This would also give us a place to tie in something like convert to iPod/iPad (mp4) without implementing it ourselves.

We should pass parameters to the script via environment variables instead of ordered variables so we aren't stuck to a specific order forever and ever.

We also need a way to capture the output of the script that was run and possible parse it for "service messages" that instruct drone to do something else, like update the file name of refresh the series (bonus).

markus101:
@Taloth @kayone 

This does need some more work (as mentioned in the TODO) as well as refactoring the notifications for "On Grab" to be more useful for actually running these scripts. I wanted to get some feedback before doing the refactoring to avoid clouding over the feature in this PR, so let me know what you think.

Closes #439

### DIFF
--- a/src/NzbDrone.Common/Processes/ProcessOutput.cs
+++ b/src/NzbDrone.Common/Processes/ProcessOutput.cs
@@ -1,17 +1,32 @@
-﻿using System;
-using System.Collections.Generic;
+﻿using System.Collections.Generic;
+using System.Linq;
 
 namespace NzbDrone.Common.Processes
 {
     public class ProcessOutput
     {
-        public List<String> Standard { get; private set; }
-        public List<String> Error { get; private set; }
+        public int ExitCode { get; set; }
+        public List<ProcessOutputLine> Lines { get; set; }
 
         public ProcessOutput()
         {
-            Standard = new List<string>();
-            Error = new List<string>();
+            Lines = new List<ProcessOutputLine>();
+        }
+
+        public List<ProcessOutputLine> Standard
+        {
+            get
+            {
+                return Lines.Where(c => c.Level == ProcessOutputLevel.Standard).ToList();
+            }
+        }
+
+        public List<ProcessOutputLine> Error
+        {
+            get
+            {
+                return Lines.Where(c => c.Level == ProcessOutputLevel.Error).ToList();
+            }
         }
     }
 }
--- a/src/NzbDrone.Core/Notifications/CustomScript/CustomScript.cs
+++ b/src/NzbDrone.Core/Notifications/CustomScript/CustomScript.cs
@@ -1,22 +1,21 @@
 ﻿using System.Collections.Generic;
 using FluentValidation.Results;
-using NzbDrone.Common.Extensions;
 using NzbDrone.Core.Tv;
 
-namespace NzbDrone.Core.Notifications.Plex
+namespace NzbDrone.Core.Notifications.CustomScript
 {
-    public class PlexServer : NotificationBase<PlexServerSettings>
+    public class CustomScript : NotificationBase<CustomScriptSettings>
     {
-        private readonly IPlexServerService _plexServerService;
+        private readonly ICustomScriptService _customScriptService;
 
-        public PlexServer(IPlexServerService plexServerService)
+        public CustomScript(ICustomScriptService customScriptService)
         {
-            _plexServerService = plexServerService;
+            _customScriptService = customScriptService;
         }
 
         public override string Link
         {
-            get { return "http://www.plexapp.com/"; }
+            get { return "https://github.com/Sonarr/Sonarr/wiki/Custom-Post-Processing-Scripts"; }
         }
 
         public override void OnGrab(string message)
@@ -25,27 +24,19 @@ namespace NzbDrone.Core.Notifications.Plex
 
         public override void OnDownload(DownloadMessage message)
         {
-            UpdateIfEnabled(message.Series);
+            _customScriptService.OnDownload(message.Series, message.EpisodeFile, Settings);
         }
 
         public override void OnRename(Series series)
         {
-            UpdateIfEnabled(series);
-        }
-
-        private void UpdateIfEnabled(Series series)
-        {
-            if (Settings.UpdateLibrary)
-            {
-                _plexServerService.UpdateLibrary(series, Settings);
-            }
+            _customScriptService.OnRename(series, Settings);
         }
 
         public override string Name
         {
             get
             {
-                return "Plex Media Server";
+                return "Custom Script";
             }
         }
 
@@ -60,8 +51,6 @@ namespace NzbDrone.Core.Notifications.Plex
         public override ValidationResult Test()
         {
             var failures = new List<ValidationFailure>();
-
-            failures.AddIfNotNull(_plexServerService.Test(Settings));
 
             return new ValidationResult(failures);
         }
--- a/src/NzbDrone.Core/Notifications/CustomScript/CustomScriptService.cs
+++ b/src/NzbDrone.Core/Notifications/CustomScript/CustomScriptService.cs
@@ -1,0 +1,92 @@
+﻿using System;
+using System.Collections.Specialized;
+using System.IO;
+using System.Linq;
+using FluentValidation.Results;
+using NLog;
+using NzbDrone.Common.Disk;
+using NzbDrone.Common.Processes;
+using NzbDrone.Core.MediaFiles;
+using NzbDrone.Core.Tv;
+using NzbDrone.Core.Validation;
+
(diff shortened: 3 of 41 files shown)


Taloth:
The biggest concern I have with this is maintainability. If we do this you're never going to get rid of it.
Env params are better then args though.
Personally I think using python would be better, or even c# script (I admit, less useful for a lot of ppl, but more powerful)

src/NzbDrone.Core/Notifications/CustomScript/CustomScript.cs
```diff
@@ -0,0 +1,53 @@
...
+
+        public CustomScript(ICustomScriptService customScriptService)
+        {
+            _customScriptService = customScriptService;
+        }
+
+        public override string Link
+        {
+            //TODO: Provide a proper link on how to use the custom script
+            get { return "http://wiki.sonarr.tv/"; }
```
kayone: https

src/NzbDrone.Core/Notifications/CustomScript/CustomScript.cs
```diff
@@ -0,0 +1,53 @@
+﻿using System.Collections.Generic;
+using FluentValidation.Results;
+using NzbDrone.Core.Tv;
+
+namespace NzbDrone.Core.Notifications.CustomScript
+{
+    public class CustomScript : NotificationBase<CustomScriptSettings>
```
kayone: I think externalscript or external process makes more sense.

src/NzbDrone.Core/Notifications/CustomScript/CustomScriptService.cs
```diff
@@ -0,0 +1,82 @@
...
+        public ValidationFailure Test(CustomScriptSettings settings)
+        {
+            return null;
+        }
+
+        private void ExecuteScript(StringDictionary environmentVariables, CustomScriptSettings settings)
+        {
+            _logger.Debug("Executing external script: {0}", settings.Path);
+
+            var process = _processProvider.StartAndCapture(settings.Path, settings.Arguments, environmentVariables);
```
kayone: we should store the output somewhere. otherwise it would be really annoying for people to debug it.

src/NzbDrone.Common/Processes/ProcessOutput.cs
```diff
@@ -5,6 +5,7 @@ namespace NzbDrone.Common.Processes
 {
     public class ProcessOutput
     {
+        public int ExitCode { get; set; }
         public List<String> Standard { get; private set; }
```
kayone: I think we should merge this into a list of `OutputLine` where it has `Level`, `Content` and `Date`, right now there is no way to see all messages in order.

src/NzbDrone.Core/Notifications/CustomScript/CustomScript.cs
```diff
@@ -0,0 +1,53 @@
...
+        {
+            get
+            {
+                return "Custom Script";
+            }
+        }
+
+        public override ValidationResult Test()
+        {
+            var failures = new List<ValidationFailure>();
```
kayone: shouldn't it be using `CustomScriptSettingsValidator`

src/NzbDrone.Core/Notifications/CustomScript/CustomScriptService.cs
```diff
@@ -0,0 +1,82 @@
...
+        {
+            _processProvider = processProvider;
+            _logger = logger;
+        }
+
+        public void OnDownload(Series series, EpisodeFile episodeFile, CustomScriptSettings settings)
+        {
+            var environmentVariables = new StringDictionary();
+
+            environmentVariables.Add("Sonarr.EventType", "Download");
```
kayone: so, this won't allow you to run a 3rd party program directly? you would always have to write a wrapping script that uses the evn variables and passes them as args?

src/NzbDrone.Core/Notifications/CustomScript/CustomScriptSettings.cs
```diff
@@ -0,0 +1,33 @@
...
+
+    public class CustomScriptSettings : IProviderConfig
+    {
+        private static readonly CustomScriptSettingsValidator Validator = new CustomScriptSettingsValidator();
+
+        [FieldDefinition(0, Label = "Path", Type = FieldType.Path)]
+        public String Path { get; set; }
+
+        [FieldDefinition(0, Label = "Arguments", HelpText = "Arguments to pass to the script")]
+        public String Arguments { get; set; }
```
kayone: if what I said about wrapping scripts is correct, why would you need arguments?

Taloth: Ppl will share scripts, args allow the user to configure it without having to edit it.

kayone: did I hear anyone say Script-repository?

src/NzbDrone.Core/Notifications/CustomScript/CustomScript.cs
markus101: I had external script originally, I felt it was a bit awkward, but I'd be okay with it. @Taloth whats your opinion?

src/NzbDrone.Core/Notifications/CustomScript/CustomScriptService.cs
markus101: Yeah, any thoughts on how we should do that?

src/NzbDrone.Common/Processes/ProcessOutput.cs
markus101: Yeah, sounds good, was part of my plan.

src/NzbDrone.Core/Notifications/CustomScript/CustomScript.cs
markus101: No, this is testing whether the notification works (in this case, just checking the script exists), but normally it would send a test message.

src/NzbDrone.Core/Notifications/CustomScript/CustomScriptService.cs
markus101: Yeah, the only way you could call the program directly would be to be able to customize the arguments with variables, which I think complicates our implementation, versus a simple wrapper script.

src/NzbDrone.Core/Notifications/CustomScript/CustomScriptSettings.cs
markus101: My thought for args would be to support running a script that requires an interpreter (like a python script), so someone could call `python script.py`, but Taloth's idea is much more likely.

src/NzbDrone.Core/Notifications/CustomScript/CustomScript.cs
Taloth: ExternalScript vs CustomScript isn't that different imho.
ExternalScript makes it more clear that it's outside of Sonarr.
CustomScript makes it more clear that it's a user-created script.

But anyone not understanding what it means shouldn't fiddle with it anyway.

I honestly dunno what you should pick. So lets vote for the least work :wink:

src/NzbDrone.Core/Notifications/CustomScript/CustomScriptService.cs
markus101: I'm wondering if we should store the post processing from all notifications, we could have a new event type or even a new table and store everything there, this way you could easily track down failures. Or a better way to link logs to a specific action, though we'll still probably need special table for notifications anyways.

Taloth: I would just dump it in the log file.

markus101: Info (so it ends up in the DB) or debug?

src/NzbDrone.Core/Notifications/CustomScript/CustomScript.cs
markus101: I'll leave it and we can change the name in the UI if we want to later on.

src/NzbDrone.Core/Notifications/CustomScript/CustomScriptService.cs
Taloth: Debug

Taloth:
looks good.

PR #561, opened 2015-05-24
Series/season metadata also created on episode import.

### LINKED ISSUE #545: Series/Season metadata should be created on episodeimport
We should probably raise an event if the episode import process creates one of those folders.
Otherwise it won't happen till the series gets refreshed.

Taloth:
Fixes #545

@markus101 Just a quick look for sign-off.

The idea is that on folder creation an event is raised. Atm only used to update metadata.
That should only happen when the series/season folders didn't exist yet.


markus101:
:ship:

PR #562, opened 2015-05-25
Added support for Rarbg as replacement for Eztv.

### LINKED ISSUE #555: Rarbg indexer integration
It would be awesome if we could get Rarbg integrated to Sonarr :)

API doc (use json_extended): https://rarbg.com/pubapi/apidocs.txt

Forum discussion: https://forums.sonarr.tv/t/eztv-has-been-hacked-please-remove-that-from-the-sources-and-add-rarbg/5629/7

@rarbg

Taloth:
Fixes #555

Not sure yet if we should default to RankedOnly=true or false.


markus101:
:ship:

PR #563, opened 2015-05-25
Support for updating single series in Plex Library

markus101:
These changes will let us update only a single series instead of the entire library, making the rescan faster, especially for larger libraries.

It requires a 0.9.12.0 or later and partial scans must be enabled, so we need to check for those and cache (to prevent having to check every refresh). Right now eligibility is cached for 2 hours, but definitely open to other values.


thezoggy:
dont you need to add the `X-Plex-*` releated header to make this work?

markus101:
Yeah, we do: https://github.com/Sonarr/Sonarr/blob/develop/src/NzbDrone.Core/Notifications/Plex/PlexServerProxy.cs#L112

PR #573, opened 2015-05-27
Ratelimitservice

Taloth:
@markus101 Did it in two separate commits so we can cherrypick to master if needed.


src/NzbDrone.Common/TPL/RateLimitService.cs
```diff
@@ -0,0 +1,43 @@
...
+
+        public RateLimitService(ICacheManager cacheManager, Logger logger)
+        {
+            _rateLimitStore = cacheManager.GetCache<ConcurrentDictionary<string, DateTime>>(GetType(), "rateLimitStore").Get("rateLimitStore", () => new ConcurrentDictionary<string, DateTime>());
+            _logger = logger;
+        }
+
+        public void WaitAndPulse(string key, TimeSpan interval)
+        {
+            var waitTill = _rateLimitStore.AddOrUpdate(key,
```
markus101: `waitUntil`

markus101:
Just the one thing, otherwise, looks good.

PR #580, opened 2015-05-29
Show an infinity instead of negative time values

clee:
Seeing negative timeleft values made me sad, so I figured I should try to fix it.


Taloth:
Better deal with this in the Download Client,  -1 isn't a valid time.
Committed a fix for Transmission on develop.

Issue #419, opened 2015-04-27
Accept Pushed NZBs/Torrents


SonarrBot:
Some sites, have the notion of watch lists that let users define series to watch and then those NZBs are pushed to the download client. We could implement the same interface in Sonarr and have those sites push to Sonarr, then Sonarr would be tracking the download to completion and also apply filtering rules (release restrictions, size, etc) on those.

bitPhex:
I attempted to create an IRC bot which monitored an announce channel, and pushed EpisodeSearch commands to Sonarr when wanted episodes were announced; however, Sonarr would not find the newly released episodes since, as it turns out, there is a delay from when the episode is uploaded to when it is indexed. 

I am going to make an attempt at this feature, as it is the only way I can think of to avoid the 10 minute RSS sink delay, and I'm really impatient when it comes to my TV! (Especially when the show airs at 10 and I need to wake up for work in the morning.)

After a quick scan through the source, it looks like the best way to accomplish this is to create a new post command under the `api/command` endpoint that takes a list of release data, which I can then parse into a list of `ReleaseInfo` and send off to be processed. If anyone more familiar with the codebase has a different idea on how to go execute this, or could point me in the right direction in terms of where the heck to implement the command, your input would be appreciated.

SAKUJ0:
> there is a delay from when the episode is uploaded to when it is indexed.

That is weird, as when a channel announces something, it has been triggered by a user of a site uploading something. In return people obtain a link, that is searchable.

I suppose I understand you wrong and you are trying to search for NZBs, as they are available some time after a .torrent is uploaded? If so, you will not beat down the RSS time. Some indexers take an hour to catch up. Others are too fast (OMGWTFNZBs has a delay setting integrated) - if you grab them too quickly, they will not have been fully uploaded to your usenet servers, yet (or propagated there).

Back to the issue at hand:

Would it not be an idea, if Sonarr's torrent functionality would be extended almost exactly like suggested in this very issue for NZBs. So, when I throw in a .torrent into the Drone Factory Folder, it could treat the torrent exactly like an incoming NZB. Maybe that is just stupid and unnecessary bloat. I do not know, how Sonarr treats things internally in regards to Torrents.

Maybe I should open a new issue for this, let's see what markus thinks about this (he liked the NZB push idea). Likely, it is just easier to put a .torrent into a watch folder for the Torrent client instead of Drone Factory (however, implementation on Sonarr's side could be that simple).

[...]

PR #631, opened 2015-06-20
Release Pushing

bitPhex:
Push releases to Sonarr via `../api/release/push` to be processed and conditionally downloaded. 

Required parameters: Title, DownloadUrl, DownloadProtocol and PublishDate.

Feature Request: https://github.com/Sonarr/Sonarr/issues/419

### DIFF
--- a/src/NzbDrone.Api/Indexers/ReleasePushModule.cs
+++ b/src/NzbDrone.Api/Indexers/ReleasePushModule.cs
@@ -47,7 +47,10 @@ namespace NzbDrone.Api.Indexers
             var decisions = _downloadDecisionMaker.GetRssDecision(new List<ReleaseInfo> { info });
             var processed = _downloadDecisionProcessor.ProcessDecisions(decisions);
 
-            _logger.Info("Release " + (processed.Grabbed.Any() ? "grabbed" : processed.Rejected.Any() ? "rejected" : "pending") + ": '" + info.Title.ToString() + "'");
+            _logger.Info("Release " + (processed.Grabbed.Any() ? "grabbed" : 
+                                       processed.Rejected.Any() ? "rejected" : 
+                                       processed.Pending.Any() ? "pending" :
+                                       "error") + ": '" + info.Title.ToString() + "'");
 
             return processed.AsResponse();
         }
--- a/src/NzbDrone.Api/Indexers/ReleasePushModule.cs
+++ b/src/NzbDrone.Api/Indexers/ReleasePushModule.cs
@@ -41,20 +41,18 @@ namespace NzbDrone.Api.Indexers
         {
             _logger.Info("Release pushed: {0}", release.Title);
 
-            ReleaseInfo info = release.InjectTo<ReleaseInfo>();
+            var info = release.InjectTo<ReleaseInfo>();
             info.Guid = "PUSH-" + info.DownloadUrl;
 
             var decisions = _downloadDecisionMaker.GetRssDecision(new List<ReleaseInfo> { info });
             var processed = _downloadDecisionProcessor.ProcessDecisions(decisions);
 
-            string status = processed.Grabbed.Any()  ? "grabbed"  :
-                            processed.Rejected.Any() ? "rejected" :
-                            processed.Pending.Any()  ? "pending"  :
-                                                       "error"    ;
+            var status = processed.Grabbed.Any()  ? "grabbed"  :
+                         processed.Rejected.Any() ? "rejected" :
+                         processed.Pending.Any()  ? "pending"  :
+                                                    "error"    ;
 
-            _logger.Info("Release {0}: {1}", status, info.Title);
-
-            return processed.AsResponse();
+            return status.AsResponse();
         }
     }
 }
--- a/src/NzbDrone.Api/Indexers/ReleasePushModule.cs
+++ b/src/NzbDrone.Api/Indexers/ReleasePushModule.cs
@@ -34,23 +34,25 @@ namespace NzbDrone.Api.Indexers
             PostValidator.RuleFor(s => s.Title).NotEmpty();
             PostValidator.RuleFor(s => s.DownloadUrl).NotEmpty();
             PostValidator.RuleFor(s => s.DownloadProtocol).NotEmpty();
+            PostValidator.RuleFor(s => s.PublishDate).NotEmpty();
         }
 
         private Response ProcessRelease(ReleaseResource release)
         {
-            _logger.Info("Release pushed: '" + release.Title.ToString() + "'");
+            _logger.Info("Release pushed: {0}", release.Title);
 
             ReleaseInfo info = release.InjectTo<ReleaseInfo>();
             info.Guid = "PUSH-" + info.DownloadUrl;
-            if (info.PublishDate == default(DateTime)) info.PublishDate = DateTime.UtcNow;
 
             var decisions = _downloadDecisionMaker.GetRssDecision(new List<ReleaseInfo> { info });
             var processed = _downloadDecisionProcessor.ProcessDecisions(decisions);
 
-            _logger.Info("Release " + (processed.Grabbed.Any() ? "grabbed" : 
-                                       processed.Rejected.Any() ? "rejected" : 
-                                       processed.Pending.Any() ? "pending" :
-                                       "error") + ": '" + info.Title.ToString() + "'");
+            string status = processed.Grabbed.Any()  ? "grabbed"  :
+                            processed.Rejected.Any() ? "rejected" :
+                            processed.Pending.Any()  ? "pending"  :
+                                                       "error"    ;
+
+            _logger.Info("Release {0}: {1}", status, info.Title);
 
             return processed.AsResponse();
         }
--- a/src/NzbDrone.Api/Indexers/ReleasePushModule.cs
+++ b/src/NzbDrone.Api/Indexers/ReleasePushModule.cs
@@ -1,0 +1,55 @@
+﻿using Nancy;
+using Nancy.ModelBinding;
+using FluentValidation;
+using NzbDrone.Core.DecisionEngine;
+using NzbDrone.Core.Download;
+using NzbDrone.Core.Parser;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using NzbDrone.Core.Parser.Model;
+using NzbDrone.Api.Mapping;
+using NzbDrone.Api.Extensions;
+using NLog;
+
+namespace NzbDrone.Api.Indexers
+{
+    class ReleasePushModule : NzbDroneRestModule<ReleaseResource>
+    {
+        private readonly IMakeDownloadDecision _downloadDecisionMaker;
+        private readonly IProcessDownloadDecisions _downloadDecisionProcessor;
+        private readonly Logger _logger;
+
+        public ReleasePushModule(IMakeDownloadDecision downloadDecisionMaker,
+                                 IProcessDownloadDecisions downloadDecisionProcessor,
+                                 Logger logger)
+        {
+            _downloadDecisionMaker = downloadDecisionMaker;
+            _downloadDecisionProcessor = downloadDecisionProcessor;
+            _logger = logger;
+
+            Post["/push"] = x => ProcessRelease(this.Bind<ReleaseResource>());
+
+            PostValidator.RuleFor(s => s.Title).NotEmpty();
+            PostValidator.RuleFor(s => s.DownloadUrl).NotEmpty();
+            PostValidator.RuleFor(s => s.DownloadProtocol).NotEmpty();
+        }
+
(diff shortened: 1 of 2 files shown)


src/NzbDrone.Api/Indexers/ReleasePushModule.cs
```diff
@@ -0,0 +1,58 @@
...
+            Post["/push"] = x => ProcessRelease(this.Bind<ReleaseResource>());
+
+            PostValidator.RuleFor(s => s.Title).NotEmpty();
+            PostValidator.RuleFor(s => s.DownloadUrl).NotEmpty();
+            PostValidator.RuleFor(s => s.DownloadProtocol).NotEmpty();
+        }
+
+        private Response ProcessRelease(ReleaseResource release)
+        {
+            _logger.Info("Release pushed: '" + release.Title.ToString() + "'");
```
markus101: Use the built in formatting instead of string concatenation, you also don't need to call `.ToString()` on a string.

`_logger.Info("Release pushed: {0}", release.Title)`

src/NzbDrone.Api/Indexers/ReleasePushModule.cs
```diff
@@ -0,0 +1,58 @@
...
+
+            ReleaseInfo info = release.InjectTo<ReleaseInfo>();
+            info.Guid = "PUSH-" + info.DownloadUrl;
+            if (info.PublishDate == default(DateTime)) info.PublishDate = DateTime.UtcNow;
+
+            var decisions = _downloadDecisionMaker.GetRssDecision(new List<ReleaseInfo> { info });
+            var processed = _downloadDecisionProcessor.ProcessDecisions(decisions);
+
+            _logger.Info("Release " + (processed.Grabbed.Any() ? "grabbed" :
```
markus101: Build the message then call the logger, also use the formatting.

src/NzbDrone.Api/Indexers/ReleasePushModule.cs
```diff
@@ -0,0 +1,58 @@
...
+            PostValidator.RuleFor(s => s.DownloadProtocol).NotEmpty();
+        }
+
+        private Response ProcessRelease(ReleaseResource release)
+        {
+            _logger.Info("Release pushed: '" + release.Title.ToString() + "'");
+
+            ReleaseInfo info = release.InjectTo<ReleaseInfo>();
+            info.Guid = "PUSH-" + info.DownloadUrl;
+            if (info.PublishDate == default(DateTime)) info.PublishDate = DateTime.UtcNow;
```
markus101: I think setting this should be the responsibility of the caller.

bitPhex:
I have been running an IRC bot which pushes to `..api/release/push` for the past couple days, and it seems to be fairly stable, except for a couple of (possibly related) issues.

1. Occasionally I get an invalid response from a push. I don't have a log of exactly what response I get instead, but I do know it shouldn't happen. (There should always be `grabbed`, `rejected`, and `pending` keys in the json response, but every so often this isn't the case.)

2. Occasionally Sonarr crashes with a StackOverflowException, accompanied by a *very* large error message.  Here is a [link to my logs.](https://mega.nz/#F!yxdFVAwT!iQEOPr5a0jRArch9QuVmfA)

And the Call Stack at the time of the error, as reported by Visual Studio:

```
>	Marr.Data.dll!Marr.Data.Reflection.SimpleReflectionStrategy.GetPropertyGetter(System.Reflection.PropertyInfo propertyInfo) Line 127	C#
 	Marr.Data.dll!Marr.Data.Reflection.SimpleReflectionStrategy.BuildGetter(System.Type type, string memberName) Line 67	C#
 	Marr.Data.dll!Marr.Data.Reflection.SimpleReflectionStrategy.GetFieldValue(object entity, string fieldName) Line 42	C#
 	Marr.Data.dll!Marr.Data.QGen.WhereBuilder<NzbDrone.Core.Tv.Episode>.GetRightValue(System.Linq.Expressions.Expression rightExpression) Line 157	C#
 	Marr.Data.dll!Marr.Data.QGen.WhereBuilder<NzbDrone.Core.Tv.Episode>.VisitMemberAccess(System.Linq.Expressions.MemberExpression expression) Line 117	C#
 	Marr.Data.dll!Marr.Data.QGen.ExpressionVisitor.Visit(System.Linq.Expressions.Expression expression) Line 69	C#
 	Marr.Data.dll!Marr.Data.QGen.WhereBuilder<NzbDrone.Core.Tv.Episode>.VisitBinary(System.Linq.Expressions.BinaryExpression expression) Line 72	C#
 	Marr.Data.dll!Marr.Data.QGen.ExpressionVisitor.Visit(System.Linq.Expressions.Expression expression) Line 63	C#
 	Marr.Data.dll!Marr.Data.QGen.ExpressionVisitor.VisitLamda(System.Linq.Expressions.LambdaExpression lambdaExpression) Line 137	C#
 	Marr.Data.dll!Marr.Data.QGen.ExpressionVisitor.Visit(System.Linq.Expressions.Expression expression) Line 28	C#
 	Marr.Data.dll!Marr.Data.QGen.WhereBuilder<NzbDrone.Core.Tv.Episode>.WhereBuilder(System.Data.Common.DbCommand command, Marr.Data.QGen.Dialects.Dialect dialect, System.Linq.Expressions.Expression filter, Marr.Data.QGen.TableCollection tables, bool useAltName, bool tablePrefix) Line 50	C#
 	Marr.Data.dll!Marr.Data.QGen.QueryBuilder<NzbDrone.Core.Tv.Episode>.Where(System.Linq.Expressions.Expression<System.Func<NzbDrone.Core.Tv.Episode,bool>> filterExpression) Line 415	C#
 	NzbDrone.Core.dll!NzbDrone.Core.Datastore.TableMapping.Map.AnonymousMethod__2(Marr.Data.IDataMapper db, NzbDrone.Core.MediaFiles.EpisodeFile parent) Line 77	C#
 	Marr.Data.dll!Marr.Data.LazyLoaded<NzbDrone.Core.MediaFiles.EpisodeFile,System.Collections.Generic.List<NzbDrone.Core.Tv.Episode>>.LazyLoad() Line 108	C#
 	Marr.Data.dll!Marr.Data.LazyLoaded<System.Collections.Generic.List<NzbDrone.Core.Tv.Episode>>.Value.get() Line 34	C#
```

It looks like two instances of this crash were caught in the logs, and it may be worth noting that both times, it occurred after an episode of `Jimmy Fallon` was pushed, a series I monitor, but in both cases, had already grabbed the episode. 

I'm not sure if both the bugs I'm reporting are related, I'll have to do some more careful logging, but if anyone has any ideas please let me know.

markus101:
The exception looks to be caused by a recursive serialization, likely due to one of the models in the processed object. I'd say the errors are likely related because its trying to lazy load an associated model.

You should be able to test this manually by pushing the same info to Sonarr, maybe the JSON posted to this endpoint should be logged at trace (like we do for responses from other applications).

One concern that came to me is if indexers support this endpoint we would be opening the doors due to requiring them to have the API key to POST to this endpoint, it needs some sort of security, but maybe the API key is the wrong approach. @Taloth would like your eyes on this one as well.

src/NzbDrone.Api/Indexers/ReleasePushModule.cs
```diff
@@ -0,0 +1,60 @@
...
+
+            PostValidator.RuleFor(s => s.Title).NotEmpty();
+            PostValidator.RuleFor(s => s.DownloadUrl).NotEmpty();
+            PostValidator.RuleFor(s => s.DownloadProtocol).NotEmpty();
+            PostValidator.RuleFor(s => s.PublishDate).NotEmpty();
+        }
+
+        private Response ProcessRelease(ReleaseResource release)
+        {
+            _logger.Info("Release pushed: {0}", release.Title);
```
Taloth: Probably should be Debug instead of Info.

Taloth:
I would assume this is used in combination with for example autodl-irssi, in which case it's local, so not really a concern. But if indexers start support it, yes, that could be a potential security risk, a rather small one.

markus101:
Yeah, you'd need to trust your indexer in that case, I know SAB added a separate key that allows for pushing nzbs only. If we went that route I think we could reasonably support both or make a breaking change in major release.

Taloth:
I wouldn't do two apikeys, it's horribly confusing. "My api doesn't work?" "Did you use the full api key, or accidentally the push api key?" ** shivers

markus101:
Thats my concern as well, I don't think this is something we need to consider right now, but was a concern about the possible future of it, either way, no action required for it right now.

src/NzbDrone.Api/Indexers/ReleasePushModule.cs
```diff
@@ -0,0 +1,60 @@
...
+
+            ReleaseInfo info = release.InjectTo<ReleaseInfo>();
+            info.Guid = "PUSH-" + info.DownloadUrl;
+
+            var decisions = _downloadDecisionMaker.GetRssDecision(new List<ReleaseInfo> { info });
```
kayone: this is too much work for the API level to do, should be moved to it's own service.

kayone: line 47 to line 55.

src/NzbDrone.Api/Indexers/ReleasePushModule.cs
```diff
@@ -0,0 +1,60 @@
...
+            PostValidator.RuleFor(s => s.DownloadUrl).NotEmpty();
+            PostValidator.RuleFor(s => s.DownloadProtocol).NotEmpty();
+            PostValidator.RuleFor(s => s.PublishDate).NotEmpty();
+        }
+
+        private Response ProcessRelease(ReleaseResource release)
+        {
+            _logger.Info("Release pushed: {0}", release.Title);
+
+            ReleaseInfo info = release.InjectTo<ReleaseInfo>();
```
markus101: `var info = release.InjectTo<ReleaseInfo>();`

src/NzbDrone.Api/Indexers/ReleasePushModule.cs
kayone: I give the security concerns you guys had I think it makes sense to keep it as info.

also, maybe update the message to be a bit more descriptive, ("Release push request received. {title} - {url}")

src/NzbDrone.Api/Indexers/ReleasePushModule.cs
```diff
@@ -0,0 +1,60 @@
...
+
+            var decisions = _downloadDecisionMaker.GetRssDecision(new List<ReleaseInfo> { info });
+            var processed = _downloadDecisionProcessor.ProcessDecisions(decisions);
+
+            string status = processed.Grabbed.Any()  ? "grabbed"  :
+                            processed.Rejected.Any() ? "rejected" :
+                            processed.Pending.Any()  ? "pending"  :
+                                                       "error"    ;
+
+            _logger.Info("Release {0}: {1}", status, info.Title);
```
kayone: tbh, I'm not sure this is needed, `DownloadDecisionProcessor ` should be doing whatever logging is required.

markus101: In that case, a new service required for 2 lines of logic? The `ReleaseModule` has a bit of brains in it too.

kayone: maybe it's time for a release service, probably not as part of this PR.

TylerHorth:
I couldn't figure out what exactly was causing the serialization error, so I implemented a rough workaround by just returning the status of the pushed episode rather than the full ProcessedDecision.

bitPhex:
I couldn't figure out what exactly was causing the serialization error, so I implemented a rough workaround by simply returning the status of the pushed episode instead of the entire ProcessedDecision.

src/NzbDrone.Api/Indexers/ReleasePushModule.cs
```diff
@@ -0,0 +1,58 @@
...
+        {
+            _logger.Info("Release pushed: {0}", release.Title);
+
+            var info = release.InjectTo<ReleaseInfo>();
+            info.Guid = "PUSH-" + info.DownloadUrl;
+
+            var decisions = _downloadDecisionMaker.GetRssDecision(new List<ReleaseInfo> { info });
+            var processed = _downloadDecisionProcessor.ProcessDecisions(decisions);
+
+            var status = processed.Grabbed.Any()  ? "grabbed"  :
```
kayone: I think we should be returning the download decision here, that way the client can know why the release they tried to push failed to download, instead of just getting back `rejected`. 

you would need to create a `DownloadDecisionResource` and map the internal model to it.

kayone: your serialization issue was most likely because you tried to serialize the internal model,

markus101:
Closing in favour of https://github.com/Sonarr/Sonarr/pull/747

PR #632, opened 2015-06-20
Curl fallback

Taloth:
@markus101 Request for comments.

Basically, when there's a tls error, I try performing the https request via libcurl.
Obviously this is only done on mono + https.

The wrapper isn't perfect, I believe there are plenty of scenarios which won't be handle properly. That's why I always try the WebClient implementation first, if that fails I mark the domain as such and subsequently try curl.

### DIFF
--- a/src/NzbDrone.Common/Http/CurlHttpClient.cs
+++ b/src/NzbDrone.Common/Http/CurlHttpClient.cs
@@ -1,0 +1,233 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.IO.Compression;
+using System.Linq;
+using System.Net;
+using System.Runtime.InteropServices;
+using System.Text;
+using CurlSharp;
+using NLog;
+using NzbDrone.Common.Extensions;
+using NzbDrone.Common.Instrumentation;
+
+namespace NzbDrone.Common.Http
+{
+    public class CurlHttpClient
+    {
+        private static Logger Logger = NzbDroneLogger.GetLogger(typeof(CurlHttpClient));
+
+        public CurlHttpClient()
+        {
+            if (!CheckAvailability())
+            {
+                throw new ApplicationException("Curl failed to initialize.");
+            }
+        }
+
+        public static bool CheckAvailability()
+        {
+            try
+            {
+                return CurlGlobalHandle.Instance.Initialize();
+            }
+            catch (Exception ex)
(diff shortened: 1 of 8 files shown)


src/NzbDrone.Common/Http/HttpClient.cs
```diff
@@ -94,6 +96,74 @@ public HttpResponse Execute(HttpRequest request)
...
+            stopWatch.Stop();
+
+            _logger.Trace("{0} ({1:n0} ms)", response, stopWatch.ElapsedMilliseconds);
+
+            if (request.AllowAutoRedirect && !RuntimeInfoBase.IsProduction &&
+                (response.StatusCode == HttpStatusCode.Moved ||
+                response.StatusCode == HttpStatusCode.MovedPermanently ||
+                response.StatusCode == HttpStatusCode.Found))
+            {
+                _logger.Error("Server requested a redirect to [" + response.Headers["Location"] + "]. Update the request URL to avoid this redirect.");
```
markus101: Format the log message.

src/NzbDrone.Common/Http/CurlHttpClient.cs
```diff
@@ -0,0 +1,233 @@
...
+{
+    public class CurlHttpClient
+    {
+        private static Logger Logger = NzbDroneLogger.GetLogger(typeof(CurlHttpClient));
+
+        public CurlHttpClient()
+        {
+            if (!CheckAvailability())
+            {
+                throw new ApplicationException("Curl failed to initialize.");
```
markus101: If curl doesn't exist I assume it would fail, do we want it to fail in that case?

src/NzbDrone.Common/Http/CurlHttpClient.cs
```diff
@@ -0,0 +1,233 @@
...
+            curlEasy.HeaderFunction = (b, s, n, o) =>
+            {
+                headerStream.Write(b, 0, s * n);
+                return s * n;
+            };
+
+            curlEasy.UserAgent = webRequest.UserAgent;
+            curlEasy.FollowLocation = webRequest.AllowAutoRedirect;
+            curlEasy.Post = webRequest.Method == "POST";
+            curlEasy.HttpGet = webRequest.Method == "GET";
```
markus101: Put as well?

PR #637, opened 2015-06-24
Adding a new Protocol for Filehosters

Phmager:
Hi,

I wanted to get some Feedback on integrating filehosters and downloadmanagers into Sonarr. I think it might need some more work done, especially with test cases, refactorings, stability and with the UI of the Delay Profiles.

Main changes:
- Added Primewire.ag as an indexer (searches for streams)
- Added JDownloader 2 as an downloader
- Added link checking possibility between the indexer and the downloader.
- Reduced task interval for CheckForFinishedDownloadCommand (I thought it might be way to slow, and would like to even more increase it to be "real-time".

To integrate JDownloader, one has to adjust the advanced settings:
image


markus101:
Unfortunately this change is not something that we will be able to accept.

In general we don't accept pull requests that mix unrelated changes, in this case the changes to the scheduler, both the interval and the change from `int` to `double`. Changing the scheduler interval to 1 second adds extra hits to to the database that aren't required and changing the interval of the check for finished downloads is a bit overkill, our intention wasn't to make the queue realtime in Sonarr.

The ultimate issue with this change is we do not support screen scraping of websites for automation, either they have an RSS feed for processing or a proper API that allows searching or they are not accepted into Sonarr, this applies to any integration and is the reason why a number of torrent trackers have been left out of Sonarr.

For future features it would be beneficial to all if they were discussed with us before work is started, especially when there is such an impact, like there is here. Let us know if you have any questions, we'll be happy to answer them.

Phmager:
No problem. I reused much of the code from another project of mine.

PR #648, opened 2015-06-28
Back-off logic for indexers

Taloth:
<s>Work in progress, but a request for comments.
Won't actually run, didn't actually add a db migration yet.

I'm not happy with how this is done architecturally.</s>

The IndexerStatusService seems okay. We could also do this in IndexerBase directly but the service seems a better option.
The IndexerStatus contains both back-off logic and last recent-feed tracking. Two separate features, but wanted to look at both coz it affects the architecture.

The back-off logic should be self-explanatory. Upon failure the indexer will not be used for a while, the duration will exponentially increase after every failure up to 24h. Every success will reduce the duration again.
So if you get Failure, Failure, Failure, Success, Failure. Then the Success won't actually completely reset the escalation.

The recent-feed tracking is intended to be able to grab multiple pages of the recent feed to catch up in case of a server or client down-time (non-24/7 operation). The status is used to determine how far back it needs to go and whether there's still a gap left.
The tracking should be okay (no tests yet), but the gaps aren't reported yet.
That should probably be done by event.
I also didn't increase the number of allowed pages for the recent feed. But once we do we might be able to run Sonarr in non-24/7 mode, should make more than a few ppl happy.

Several problems exist:
1) <s>Since the nzb/torrent is grabbed in the DownloadClient instance, it's more difficult to differentiate between exceptions thrown by the nzb/torrent download and exceptions thrown by the download client. Atm I catch the exceptions in the caller (DownloadService).</s> done
2) <s>HttpIndexerBase currently gets a reference to IIndexerStatusService, I'm thinking we might want to use events instead.</s> nope
3) <s>Recent feed gaps aren't yet reported, we could raise an event, but we first have to run through all the indexers. So not sure how to do that properly. Could also just store the gap in the IndexerStatus and reset it when the MissingEpisode search finishes.</s> done
4) <s>We need to add a HealthCheck to warn about it.</s> done


markus101:
1. Exceptions in DownloadClient exception is something I wanted to handle so we could also deal with the download client being down (and storing the nzb) - perhaps we can raise different types of exceptions?

2. That would cut down on the indexer dependencies

3. I think this depends on whether we want to check via RSS or if we are okay with relying on a search. I think I'm leaning towards the former since there are non-search indexers.

4. Do we want these to be dismissable? Maybe we should track them in the DB?

Taloth:
1) Yes, but atm we differentiate between DownloadClientException (something the client reports to us) and other exceptions (for example HttpException/WebException when the connection fails). Atm there's no way of differentiating between WebException by indexer or WebException by downloadclient, that's fixable of course, with a well placed try catch. We should probably throw any nzb/torrent download failure as IndexerException or something with the actual exception as InnerException.

2) Not really, we would be replacing IIndexerStatusService with IEventAggregator. It will be more loosely coupled with events.

3) I'm not okay with relying on a search. That's why I'm doing this, to be able to fill in the blanks for a longer period and only resort to search if that fails. I'm leaning towards storing the gap in the IndexerStatus as well.

4) No, you can dismiss it by disabling the indexer. We already track the back-off in the database, so it will be easy possible to reraise the healthcheck on restart.

Taloth:
@markus101 
Okay, rewritten most of this. I didn't use eventAggregrator coz in most cases the caller service had to retrieve the IndexerStatus as well.

Also added HealthCheck to get a more prominent warning out there in case of failures.

Once concern is the current escalation speed.
5 sec, 10 sec, 20 sec, 40 sec, 80 sec, 160 sec, 320 sec, 640 sec etc. This means it takes 9 failures in a row to reach a 30min back-off.
I could start at 30 sec, in which case it will reach it in 6.
or start at 5 minutes, in which case it will reach it in 4.
The starting point isn't that important, coz it will generally take 15 min anyway to get to the next rss sync, but on one end you want it to get disabled fast enough, but not escalate to 24h delay too fast.

If I start at 5 minutes, it will start skipping rss syncs after 3 failures, and reach the 24h back-off after 9 failures (roughly 40h after the first failure).

So feel free to shoot at it now. Atm I'm running it live to see if the escalation works as intended.

src/NzbDrone.Core/Datastore/Migration/091_added_indexerstatus.cs
```diff
@@ -0,0 +1,26 @@
...
+    {
+        protected override void MainDbUpgrade()
+        {
+            Create.TableForModel("IndexerStatus")
+                  .WithColumn("IndexerId").AsInt32().NotNullable().Unique()
+                  .WithColumn("FirstFailure").AsDateTime().Nullable()
+                  .WithColumn("LastFailure").AsDateTime().Nullable()
+                  .WithColumn("FailureEscalation").AsInt32().NotNullable()
+                  .WithColumn("BackOffDate").AsDateTime().Nullable()
+                  .WithColumn("LastRecentSearch").AsDateTime().Nullable()
```
kayone: `MostRecentSearch`
`MostRecentReleaseInfo`

Taloth: It's actually MostRecent-RecentSearch. Recent = RssSync. It's not actually a Search.

src/NzbDrone.Core/Indexers/IndexerStatus.cs
```diff
@@ -0,0 +1,21 @@
...
+using NzbDrone.Core.Parser.Model;
+
+namespace NzbDrone.Core.Indexers
+{
+    public class IndexerStatus : ModelBase
+    {
+        public int IndexerId { get; set; }
+
+        public DateTime? FirstFailure { get; set; }
+        public DateTime? LastFailure { get; set; }
```
kayone: `MostRecentFailure`

src/NzbDrone.Core/Indexers/IndexerStatus.cs
```diff
@@ -0,0 +1,21 @@
...
+        public int FailureEscalation { get; set; }
```
kayone: `EscalationLevel` or `EscalationCount`

src/NzbDrone.Core/Indexers/IndexerStatusService.cs
```diff
@@ -0,0 +1,117 @@
...
+                    }
+                }
+
+                status.DisabledTill = now + CalculateBackOffPeriod(status);
+
+                _indexerStatusRepository.Upsert(status);
+            }
+        }
+
+        public void UpdateRecentSearchStatus(int indexerId, ReleaseInfo releaseInfo, bool fullyUpdated)
```
kayone: shouldn't this be `UpdateRssStatus`, since you are updating the rss sync flag.

src/NzbDrone.Core/Indexers/RssSyncService.cs
```diff
@@ -44,7 +51,7 @@ private ProcessedDecisions Sync()
         {
             _logger.ProgressInfo("Starting RSS Sync");
 
-            var reports = _rssFetcherAndParser.Fetch().Concat(_pendingReleaseService.GetPending()).ToList();
+            var reports = _rssFetcherAndParser.Fetch().Concat(FilterByIndexer(_pendingReleaseService.GetPending())).ToList();
```
kayone: there is way too many method calls in one line, I would break them into different calls with proper variable name for what they are returning.

src/NzbDrone.Core/Indexers/RssSyncService.cs
```diff
@@ -60,15 +67,70 @@ private ProcessedDecisions Sync()
...
         }
 
+        private IEnumerable<ReleaseInfo> FilterByIndexer(IEnumerable<ReleaseInfo> releases)
+        {
+            var indexerBackOff = new Dictionary<int, bool>();
+
+            foreach (var release in releases)
+            {
+                bool ignore;
+                if (!indexerBackOff.TryGetValue(release.IndexerId, out ignore))
```
kayone: you should use `Cached`instead of doing your own cache. that's exactly why it's there.

src/NzbDrone.Core/Indexers/RssSyncService.cs
```diff
@@ -60,15 +67,70 @@ private ProcessedDecisions Sync()
             return processed;
         }
 
+        private IEnumerable<ReleaseInfo> FilterByIndexer(IEnumerable<ReleaseInfo> releases)
```
kayone: `FilterBlockedIndexers`

src/NzbDrone.Core/Indexers/RssSyncService.cs
kayone: I would suggest `var pendingReleases` and `var liveReports` and concating them into `reports`.

also it's a bit strange that  `RssFetcherAndParser` is dealing with indexer backoff, but `PendingReleaseService` doesn't. it means `RssSyncService` knows a bit too much about their internals.

src/NzbDrone.Core/Indexers/RssSyncService.cs
```diff
@@ -60,15 +67,70 @@ private ProcessedDecisions Sync()
...
+            var indexerBackOff = new Dictionary<int, bool>();
+
+            foreach (var release in releases)
+            {
+                bool ignore;
+                if (!indexerBackOff.TryGetValue(release.IndexerId, out ignore))
+                {
+                    var indexerStatus = _indexerStatusService.GetIndexerStatus(release.IndexerId);
+
+                    if (indexerStatus == null || !indexerStatus.DisabledTill.HasValue || indexerStatus.DisabledTill.Value < DateTime.UtcNow)
```
kayone: should never be null is you are defaulting in `IndexerStatusService`

src/NzbDrone.Core/Indexers/RssSyncService.cs
```diff
@@ -60,15 +67,70 @@ private ProcessedDecisions Sync()
...
+            var rssStarted = DateTime.UtcNow;
+
             var processed = Sync();
             var grabbedOrPending = processed.Grabbed.Concat(processed.Pending).ToList();
 
             if (message.LastExecutionTime.HasValue && DateTime.UtcNow.Subtract(message.LastExecutionTime.Value).TotalHours > 3)
             {
-                _logger.Info("RSS Sync hasn't run since: {0}. Searching for any missing episodes since then.", message.LastExecutionTime.Value);
-                _episodeSearchService.MissingEpisodesAiredAfter(message.LastExecutionTime.Value.AddDays(-1), grabbedOrPending.SelectMany(d => d.RemoteEpisode.Episodes).Select(e => e.Id));
+                var lastGap = rssStarted;
```
kayone: to be honest, I don't think this should run as part of RSS sync. it should be it's own command.

src/NzbDrone.Core/Download/TorrentClientBase.cs
```diff
@@ -126,7 +126,13 @@ private string DownloadFromWebUrl(RemoteEpisode remoteEpisode, String torrentUrl
 
                 torrentFile = response.ResponseData;
             }
+            catch (HttpException ex)
+            {
+                _logger.ErrorException(String.Format("Downloading torrentfile for episode '{0}' failed ({1})",
```
kayone: `torrent file` or `.torrent`

src/NzbDrone.Core/Download/UsenetClientBase.cs
```diff
@@ -49,6 +49,13 @@ public override String Download(RemoteEpisode remoteEpisode)
             {
                 nzbData = _httpClient.Get(new HttpRequest(url)).ResponseData;
             }
+            catch (HttpException ex)
+            {
+                _logger.ErrorException(String.Format("Downloading nzb for episode '{0}' failed ({1})",
```
kayone: can't we move this to download service since both `torrentClient` and `usenet` client do it?

src/NzbDrone.Core/HealthCheck/Checks/IndexerStatusCheck.cs
```diff
@@ -0,0 +1,47 @@
...
+        {
+            _indexerFactory = indexerFactory;
+            _indexerStatusService = indexerStatusService;
+        }
+
+        public override HealthCheck Check()
+        {
+            var enabledIndexers = _indexerFactory.GetAvailableProviders();
+            var backOffIndexers = enabledIndexers
+                .Select(v => Tuple.Create(v, _indexerStatusService.GetIndexerStatus(v.Definition.Id)))
```
kayone: use an anonymous type instead of a tuple. makes things easier to read. also break this in multiple stages with proper variable names.

src/NzbDrone.Core/HealthCheck/Checks/IndexerStatusCheck.cs
```diff
@@ -0,0 +1,47 @@
...
+            }
+
+            if (backOffIndexers.Count == enabledIndexers.Count)
+            {
+                return new HealthCheck(GetType(), HealthCheckResult.Error, "All indexers unusable due to failures", "#indexer-unusable-due-to-failures");
+            }
+
+            if (backOffIndexers.Count > 1)
+            {
+                return new HealthCheck(GetType(), HealthCheckResult.Warning, "Multiple indexers unusable due to failures", "#indexer-unusable-due-to-failures");
```
kayone: use actual count instead of "multiple"

src/NzbDrone.Core/Housekeeping/Housekeepers/CleanupOrphanedIndexerStatus.cs
```diff
@@ -0,0 +1,26 @@
...
+        public CleanupOrphanedIndexerStatus(IMainDatabase database)
+        {
+            _database = database;
+        }
+
+        public void Clean()
+        {
+            var mapper = _database.GetDataMapper();
+
+            mapper.ExecuteNonQuery(@"DELETE FROM IndexerStatus
```
kayone: can we not just delete the status when someone deletes an indexer instead of relying on a housekeeping?

src/NzbDrone.Core/IndexerSearch/NzbSearchService.cs
```diff
@@ -257,6 +260,13 @@ private List<DownloadDecision> Dispatch(Func<IIndexer, IEnumerable<ReleaseInfo>>
 
             foreach (var indexer in indexers)
             {
+                var indexerStatus = _indexerStatusService.GetIndexerStatus(indexer.Definition.Id);
+                if (indexerStatus != null && indexerStatus.DisabledTill.HasValue && indexerStatus.DisabledTill.Value > DateTime.UtcNow)
+                {
+                    _logger.Debug("Temporarily ignoring indexer {0} till {1} due to recent failures.", indexer.Definition.Name, indexerStatus.DisabledTill.Value);
```
kayone: maybe as info? users should be able to tell why an indexer isn't being used.

src/NzbDrone.Core/IndexerSearch/NzbSearchService.cs
```diff
@@ -257,6 +260,13 @@ private List<DownloadDecision> Dispatch(Func<IIndexer, IEnumerable<ReleaseInfo>>
 
             foreach (var indexer in indexers)
             {
+                var indexerStatus = _indexerStatusService.GetIndexerStatus(indexer.Definition.Id);
+                if (indexerStatus != null && indexerStatus.DisabledTill.HasValue && indexerStatus.DisabledTill.Value > DateTime.UtcNow)
```
kayone: there is too much logic around whether or not an indexer is usable due to it's status, I think we should move this logic to `IndexerStatus` class as methods. `indexerStatus.IsDisabled()`

src/NzbDrone.Core/Indexers/FetchAndParseRssService.cs
```diff
@@ -42,6 +45,13 @@ public List<ReleaseInfo> Fetch()
 
             foreach (var indexer in indexers)
             {
+                var indexerStatus = _indexerStatusService.GetIndexerStatus(indexer.Definition.Id);
+                if (indexerStatus != null && indexerStatus.DisabledTill.HasValue && indexerStatus.DisabledTill.Value > DateTime.UtcNow)
```
kayone: maybe this should be moved to `indexerFactory.RssEnabled()` with a boolean flag `bool includedBlocked`

src/NzbDrone.Core/Indexers/HttpIndexerBase.cs
```diff
@@ -132,7 +143,38 @@ protected virtual IList<ReleaseInfo> FetchReleases(IList<IEnumerable<IndexerRequ
...
+                                if (oldestDate < DateTime.UtcNow - TimeSpan.FromHours(24))
+                                {
+                                    fullyUpdated = false;
+                                    break;
+                                }
+                            }
+                        }
+                        else
+                        {
+                            if (pagedReleases.Count >= MaxNumResultsPerQuery)
```
kayone: `else if`

src/NzbDrone.Core/Indexers/HttpIndexerBase.cs
```diff
@@ -132,7 +143,38 @@ protected virtual IList<ReleaseInfo> FetchReleases(IList<IEnumerable<IndexerRequ
...
+                            var oldestDate = page.Select(v => v.PublishDate).Min();
+                            if (oldestDate < lastReleaseInfo.PublishDate || page.Any(v => v.DownloadUrl == lastReleaseInfo.DownloadUrl))
+                            {
+                                fullyUpdated = true;
+                                break;
+                            }
+
+                            if (pagedReleases.Count >= MaxNumResultsPerQuery)
+                            {
+                                if (oldestDate < DateTime.UtcNow - TimeSpan.FromHours(24))
```
kayone: merge these two ifs,

src/NzbDrone.Core/Indexers/HttpIndexerBase.cs
```diff
@@ -120,6 +120,17 @@ protected virtual IList<ReleaseInfo> FetchReleases(IList<IEnumerable<IndexerRequ
 
             try
             {
+                var fullyUpdated = false;
+                ReleaseInfo lastReleaseInfo = null;
+                if (isRecent)
+                {
+                    var indexerStatus = _indexerStatusService.GetIndexerStatus(Definition.Id);
+                    if (indexerStatus != null)
```
kayone: don't need null gaurd.

src/NzbDrone.Core/Indexers/HttpIndexerBase.cs
```diff
@@ -140,44 +182,61 @@ protected virtual IList<ReleaseInfo> FetchReleases(IList<IEnumerable<IndexerRequ
...
+                }
+
+                _indexerStatusService.ReportSuccess(Definition.Id);
             }
             catch (WebException webException)
             {
                 if (webException.Message.Contains("502") || webException.Message.Contains("503") ||
                     webException.Message.Contains("timed out"))
                 {
+                    _indexerStatusService.ReportFailure(Definition.Id);
```
kayone: why not just do it outside of the if instead of in both if and the else?

src/NzbDrone.Core/Housekeeping/Housekeepers/CleanupOrphanedIndexerStatus.cs
Taloth: That involves modifying and overriding Delete on IndexerFactory and have IndexerFactory IIndexerStatusRepository as well. Fine by me though.

src/NzbDrone.Core/IndexerSearch/NzbSearchService.cs
Taloth: They already have warnings of the earlier errors. If you do this as Info you get alot of entries saying the same thing over and over again.

src/NzbDrone.Core/IndexerSearch/NzbSearchService.cs
Taloth: Agreed

src/NzbDrone.Core/Indexers/IndexerStatusService.cs
Taloth: Yes, renamed the property earlier, forgot the method.

src/NzbDrone.Core/Indexers/RssSyncService.cs
Taloth: It's a very short-lived cache. I could instantiate Cached, but I shouldn't use CacheManager, if that's what you wanted.

kayone: Yeah. Just create a cached everytime

src/NzbDrone.Core/Housekeeping/Housekeepers/CleanupOrphanedIndexerStatus.cs
kayone: Event?

src/NzbDrone.Core/Indexers/IndexerStatus.cs
Taloth: FirstFailure = first failure in the current escalation cycle..
LastFailure = last failure in the current escalation cycle.

I can change LastFailure to MostRecentFailure, but then what should FirstFailure be renamed to?

src/NzbDrone.Core/Download/UsenetClientBase.cs
Taloth: No, coz torrentclient will use magnet.

src/NzbDrone.Common/Http/TooManyRequestsException.cs
```diff
@@ -0,0 +1,21 @@
...
+namespace NzbDrone.Common.Http
+{
+    public class TooManyRequestsException : HttpException
+    {
+        public TimeSpan RetryAfter { get; private set; }
+
+        public TooManyRequestsException(HttpRequest request, HttpResponse response)
+            : base(request, response)
+        {
+            if ((response.Headers.ContainsKey("Retry-After")))
```
kayone: extra `()`

src/NzbDrone.Core/Download/DownloadService.cs
```diff
@@ -54,7 +60,26 @@ public void DownloadReport(RemoteEpisode remoteEpisode)
                 _rateLimitService.WaitAndPulse(uri.Host, TimeSpan.FromSeconds(2));
             }
 
-            var downloadClientId = downloadClient.Download(remoteEpisode);
+            string downloadClientId;
+            try
+            {
+                downloadClientId = downloadClient.Download(remoteEpisode);
+                _indexerStatusService.ReportSuccess(remoteEpisode.Release.IndexerId);
```
kayone: I think `RecordSuccess` might be a more accurate name, not a big deal.

src/NzbDrone.Core/Download/DownloadService.cs
```diff
@@ -54,7 +60,26 @@ public void DownloadReport(RemoteEpisode remoteEpisode)
...
+            }
+            catch (ReleaseDownloadException ex)
+            {
+                var http429 = ex.InnerException as TooManyRequestsException;
+                if (http429 != null)
+                {
+                    _indexerStatusService.ReportFailure(remoteEpisode.Release.IndexerId, http429.RetryAfter);
```
kayone: same here,

src/NzbDrone.Core/Download/DownloadService.cs
```diff
@@ -54,7 +60,26 @@ public void DownloadReport(RemoteEpisode remoteEpisode)
...
-            var downloadClientId = downloadClient.Download(remoteEpisode);
+            string downloadClientId;
+            try
+            {
+                downloadClientId = downloadClient.Download(remoteEpisode);
+                _indexerStatusService.ReportSuccess(remoteEpisode.Release.IndexerId);
+            }
+            catch (ReleaseDownloadException ex)
+            {
+                var http429 = ex.InnerException as TooManyRequestsException;
```
kayone: I think having a specific catch is cleaner than relaying on casting and `if`.

src/NzbDrone.Core/IndexerSearch/NzbSearchService.cs
```diff
@@ -25,6 +25,7 @@ public interface ISearchForNzb
     public class NzbSearchService : ISearchForNzb
     {
         private readonly IIndexerFactory _indexerFactory;
+        private readonly IIndexerStatusService _indexerStatusService;
```
kayone: I don't think it's being used anymore

src/NzbDrone.Core/Indexers/HttpIndexerBase.cs
```diff
@@ -132,7 +143,33 @@ protected virtual IList<ReleaseInfo> FetchReleases(IList<IEnumerable<IndexerRequ
...
 
-                        if (!IsFullPage(page) || pagedReleases.Count >= MaxNumResultsPerQuery)
+                        if (isRecent)
+                        {
+                            if (lastReleaseInfo == null)
+                            {
+                                fullyUpdated = true;
+                                break;
+                            }
+                            var oldestDate = page.Select(v => v.PublishDate).Min();
```
kayone: `oldestReleaseDate`

src/NzbDrone.Core/Download/DownloadService.cs
Taloth: The original TooManyRequests is caught (all exceptions are caught) and thrown as ReleaseDownloadException. That's why I'm checking the InnerException.

src/NzbDrone.Core/Indexers/IndexerFactory.cs
```diff
@@ -50,12 +50,35 @@ public override IndexerDefinition GetProviderCharacteristics(IIndexer provider,
...
+            var indexers = FilterBlockedIndexers(enabledIndexers);
+
+            return indexers.ToList();
+        }
+
+        private IEnumerable<IIndexer> FilterBlockedIndexers(IEnumerable<IIndexer> indexers)
+        {
+            foreach (var indexer in indexers)
+            {
+                var indexerStatus = _indexerStatusService.GetIndexerStatus(indexer.Definition.Id);
```
kayone: use `GetBlockedIndexers()` single call to DB instead of one call per indexer.

src/NzbDrone.Core/Indexers/IndexerStatusService.cs
```diff
@@ -0,0 +1,135 @@
...
+        }
+
+        public List<IndexerStatus> GetBlockedIndexers()
+        {
+            return _indexerStatusRepository.All().Where(v => v.IsDisabled()).ToList();
+        }
+
+        public IndexerStatus GetIndexerStatus(int indexerId)
+        {
+            return _indexerStatusRepository.FindByIndexerId(indexerId) ?? new IndexerStatus { IndexerId = indexerId };
```
kayone: you have a null guard here, but you still check for null all over the place.

src/NzbDrone.Core/Indexers/IndexerFactory.cs
```diff
@@ -50,12 +50,35 @@ public override IndexerDefinition GetProviderCharacteristics(IIndexer provider,
...
+                if (indexerStatus != null && indexerStatus.IsDisabled())
```
kayone: don't need the null guard.

src/NzbDrone.Core/Indexers/RssSyncService.cs
```diff
@@ -62,15 +69,11 @@ private ProcessedDecisions Sync()
 
         public void Execute(RssSyncCommand message)
         {
+            var rssStarted = DateTime.UtcNow;
```
kayone: not used?

src/NzbDrone.Core/ThingiProvider/Events/ProviderUpdatedEvent.cs
```diff
@@ -4,5 +4,11 @@ namespace NzbDrone.Core.ThingiProvider.Events
 {
     public class ProviderUpdatedEvent<TProvider> : IEvent
     {
+        public int ProviderId { get; private set; }
+
+        public ProviderUpdatedEvent(int id)
```
kayone: I think this should be the whole definition object rather than just the ID.

src/NzbDrone.Core/Indexers/IndexerStatusService.cs
Taloth: I think I mentioned this on hangouts. These null checks are needed to make the tests easier.

kayone: core code to make tests easier. _judging_

why not just mock it correctly?

Taloth: That would mean adding it to dozens of testfixtures that shouldn't concern themselves with the indexerstatusservice.
So if you really think it's ugly I'll move the null-coalescing to the places in the service where it's actually used, making the null guards actually required in production code as well.

kayone: I still think it should be responsibility of the test to simulate the context as it would be in production.

PR #650, opened 2015-06-29
Allow downloads to be tracked by the source name in addition to the download name

Kayomani:
I am working getting an anime indexer working and it has the issue that the torrent name isn't in a format that Sonarr supports so it doesn't track when added to the download client.  

I changed it so that if its unable to match using the download name then it looks up the history item using the download id and uses the source title which works as the name supplied by the indexer rather than inside the torrent file is in the correct format.

### DIFF
--- a/src/NzbDrone.Core.Test/Download/CompletedDownloadServiceFixture.cs
+++ b/src/NzbDrone.Core.Test/Download/CompletedDownloadServiceFixture.cs
@@ -34,11 +34,7 @@ namespace NzbDrone.Core.Test.Download
                                                     .With(h => h.Title = "Drone.S01E01.HDTV")
                                                     .Build();
 
-            var remoteEpisode = new RemoteEpisode
-                                {
-                                    Series = new Series(),
-                                    Episodes = new List<Episode> { new Episode { Id = 1 } }
-                                };
+            var remoteEpisode = BuildRemoteEpisode();
 
             _trackedDownload = Builder<TrackedDownload>.CreateNew()
                     .With(c => c.State = TrackedDownloadStage.Downloading)
@@ -65,6 +61,16 @@ namespace NzbDrone.Core.Test.Download
 
         }
 
+        private RemoteEpisode BuildRemoteEpisode()
+        {
+            return new RemoteEpisode
+            {
+                Series = new Series(),
+                Episodes = new List<Episode> { new Episode { Id = 1 } }
+            }; 
+        }
+
+
         private void GivenNoGrabbedHistory()
         {
             Mocker.GetMock<IHistoryService>()
@@ -80,6 +86,24 @@ namespace NzbDrone.Core.Test.Download
                     {
                         new ImportResult(new ImportDecision(new LocalEpisode() { Path = @"C:\TestPath\Droned.S01E01.mkv" }))
                     });
+        }
+
+
+        private void GivenABadlyNamedDownload()
+        {
+            _trackedDownload.DownloadItem.DownloadId = "1234";
+            _trackedDownload.DownloadItem.Title = "Droned Pilot"; // Set a badly named download
+            Mocker.GetMock<IHistoryService>()
+               .Setup(s => s.MostRecentForDownloadId(It.Is<string>(i => i == "1234")))
+               .Returns(new History.History() { SourceTitle = "Droned S01E01" });
+
+            Mocker.GetMock<IParsingService>()
+               .Setup(s => s.GetSeries(It.IsAny<string>()))
+               .Returns((Series)null);
+
+            Mocker.GetMock<IParsingService>()
+                .Setup(s => s.GetSeries("Droned S01E01"))
+                .Returns(BuildRemoteEpisode().Series);
         }
 
         private void GivenSeriesMatch()
@@ -278,6 +302,47 @@ namespace NzbDrone.Core.Test.Download
                                new ImportResult(new ImportDecision(new LocalEpisode{Path = @"C:\TestPath\Droned.S01E01.mkv"}),"Test Failure")
                            });
 
+
+            Subject.Process(_trackedDownload);
+
+            AssertNoCompletedDownload();
+        }
+
+        [Test]
+        public void should_mark_as_imported_if_the_download_can_be_tracked_using_the_source_seriesid()
+        {
+            GivenABadlyNamedDownload();
+
+            Mocker.GetMock<IDownloadedEpisodesImportService>()
+                  .Setup(v => v.ProcessPath(It.IsAny<string>(), It.IsAny<Series>(), It.IsAny<DownloadClientItem>()))
+                  .Returns(new List<ImportResult>
+                           {
+                               new ImportResult(new ImportDecision(new LocalEpisode {Path = @"C:\TestPath\Droned.S01E01.mkv"}))
+                           });
+
+            Mocker.GetMock<ISeriesService>()
+                  .Setup(v => v.GetSeries(It.IsAny<int>()))
+                  .Returns(BuildRemoteEpisode().Series);
+           
+            Subject.Process(_trackedDownload);
+
+            AssertCompletedDownload();
+        }
+
+        [Test]
+        public void should_not_mark_as_imported_if_the_download_cannot_be_tracked_using_the_source_title_as_it_was_initiated_externally()
+        {
+            GivenABadlyNamedDownload();
+
+            Mocker.GetMock<IDownloadedEpisodesImportService>()
+                  .Setup(v => v.ProcessPath(It.IsAny<string>(), It.IsAny<Series>(), It.IsAny<DownloadClientItem>()))
+                  .Returns(new List<ImportResult>
+                           {
+                               new ImportResult(new ImportDecision(new LocalEpisode {Path = @"C:\TestPath\Droned.S01E01.mkv"}))
+                           });
+
+            Mocker.GetMock<IHistoryService>()
+            .Setup(s => s.MostRecentForDownloadId(It.Is<string>(i => i == "1234")));
 
             Subject.Process(_trackedDownload);
 
--- a/src/NzbDrone.Core.Test/Download/TrackedDownloads/TrackedDownloadServiceFixture.cs
+++ b/src/NzbDrone.Core.Test/Download/TrackedDownloads/TrackedDownloadServiceFixture.cs
@@ -1,0 +1,82 @@
+﻿using System.Collections.Generic;
+using FizzWare.NBuilder;
+using FluentAssertions;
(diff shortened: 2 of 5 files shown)


src/NzbDrone.Core.Test/Download/CompletedDownloadServiceFixture.cs
```diff
@@ -65,6 +61,16 @@ public void Setup()
 
         }
 
+        private RemoteEpisode GetRemoteEpisode()
```
kayone: This should probably be `CreateRemoteEpisode` or `BuildRemoteEpisode`

src/NzbDrone.Core/Download/CompletedDownloadService.cs
```diff
@@ -80,8 +80,12 @@ public void Process(TrackedDownload trackedDownload, bool ignoreWarnings = false
 
                 if (series == null)
                 {
-                    trackedDownload.Warn("Series title mismatch, automatic import is not possible.");
-                    return;
+                    series = _parsingService.GetSeries(historyItem.SourceTitle);
```
kayone: `historyItem` could be null here if the download wasn't initiated by sonnar, you need to do a check before trying to get it's `SourceTitle`

src/NzbDrone.Core/Download/TrackedDownloads/TrackedDownloadService.cs
```diff
@@ -70,6 +75,11 @@ public TrackedDownload TrackDownload(DownloadClientDefinition downloadClient, Do
                         return null;
                     }
 
+                    if (parsedEpisodeInfo == null)
```
kayone: this `if` is a bit out of place, I think you should check for a `null` `parsedEpisodeInfo`as an `else` on line 67, you would have to re-arrange the method a bit but I think it would end up make more logical sense.

Kayomani: I don't think it should be an else so in the small case we parsed the wrong thing the map returned nothing.  I've refactored it to make more sense nonetheless.

It is my opinion that we shouldn't even parse the download name and always use the source name but I kept it close to what the original code does.

zone117x:
:+1:

src/NzbDrone.Core/Download/CompletedDownloadService.cs
```diff
@@ -80,8 +80,16 @@ public void Process(TrackedDownload trackedDownload, bool ignoreWarnings = false
 
                 if (series == null)
                 {
-                    trackedDownload.Warn("Series title mismatch, automatic import is not possible.");
-                    return;
+                    if (historyItem != null)
+                    {
+                        series = _parsingService.GetSeries(historyItem.SourceTitle);
```
markus101: `historyItem` should have the `SeriesId` so we can just use that.

src/NzbDrone.Core/Download/TrackedDownloads/TrackedDownloadService.cs
```diff
@@ -57,23 +58,29 @@ public TrackedDownload TrackDownload(DownloadClientDefinition downloadClient, Do
             try
             {
                 var parsedEpisodeInfo = Parser.Parser.ParseTitle(trackedDownload.DownloadItem.Title);
-                if (parsedEpisodeInfo == null) return null;
+                var historyItems = _historyService.FindByDownloadId(downloadItem.DownloadId);
 
-                var remoteEpisode = _parsingService.Map(parsedEpisodeInfo);
+                if (parsedEpisodeInfo != null)
+                    trackedDownload.RemoteEpisode = _parsingService.Map(parsedEpisodeInfo);
```
markus101: Wrap if block in `{}`

src/NzbDrone.Core/Download/TrackedDownloads/TrackedDownloadService.cs
```diff
@@ -57,23 +58,29 @@ public TrackedDownload TrackDownload(DownloadClientDefinition downloadClient, Do
...
+                        if (parsedEpisodeInfo != null)
+                            trackedDownload.RemoteEpisode = _parsingService.Map(parsedEpisodeInfo, firstHistoryItem.SeriesId, historyItems.Select(h => h.EpisodeId));
                     }
-
-                    remoteEpisode = _parsingService.Map(parsedEpisodeInfo, historyItems.First().SeriesId, historyItems.Select(h => h.EpisodeId));
                 }
 
-                trackedDownload.RemoteEpisode = remoteEpisode;
+                if (trackedDownload.RemoteEpisode == null)
+                    return null;
```
markus101: same here

src/NzbDrone.Core.Test/Download/TrackedDownloads/TrackedDownloadServiceFixture.cs
```diff
@@ -0,0 +1,87 @@
...
+
+namespace NzbDrone.Core.Test.Download
+{
+    [TestFixture]
+    public class TrackedDownloadServiceFixture : CoreTest<TrackedDownloadService>
+    {
+        private void GivenAHistoryWithADownload()
+        {
+           
+
```
markus101: OCD: random new lines.

Kayomani:
I have dealt with all the requests, are you happy with this now?

src/NzbDrone.Core/Download/CompletedDownloadService.cs
```diff
@@ -27,13 +28,15 @@ public class CompletedDownloadService : ICompletedDownloadService
...
+        private readonly ISeriesService _seriesService;
 
         public CompletedDownloadService(IConfigService configService,
                                         IEventAggregator eventAggregator,
                                         IHistoryService historyService,
                                         IDownloadedEpisodesImportService downloadedEpisodesImportService,
                                         IParsingService parsingService,
-                                        Logger logger)
+                                        Logger logger,
+                                        ISeriesService seriesService)
```
kayone: pretty minor, but we usually keep logger as the last arguments, (or at least try, you don't have to fix it but just for future)

src/NzbDrone.Core/Download/TrackedDownloads/TrackedDownloadService.cs
```diff
@@ -57,23 +58,33 @@ public TrackedDownload TrackDownload(DownloadClientDefinition downloadClient, Do
...
+                var historyItems = _historyService.FindByDownloadId(downloadItem.DownloadId);
 
-                var remoteEpisode = _parsingService.Map(parsedEpisodeInfo);
+                if (parsedEpisodeInfo != null)
+                {
+                    trackedDownload.RemoteEpisode = _parsingService.Map(parsedEpisodeInfo);
+                }
 
-                if (remoteEpisode.Series == null || !remoteEpisode.Episodes.Any())
+                if (historyItems != null && !historyItems.Empty())
```
kayone: `historyItems.Any()` instead of `!historyItems.Empty()`

kayone: also, `_historyService.FindByDownloadId` should never return null, if it does then it's a bug that should be fixed. our rule is to always return empty lists instead of null where the return type is enumerable.

src/NzbDrone.Core/Download/TrackedDownloads/TrackedDownloadService.cs
```diff
@@ -57,23 +58,33 @@ public TrackedDownload TrackDownload(DownloadClientDefinition downloadClient, Do
...
                 {
-                    var historyItems = _historyService.FindByDownloadId(downloadItem.DownloadId);
+                    var firstHistoryItem = historyItems.OrderByDescending(h => h.Date).First();
+                    trackedDownload.State = GetStateFromHistory(firstHistoryItem.EventType);
 
-                    if (historyItems.Empty())
+                    if (parsedEpisodeInfo == null ||
+                    trackedDownload.RemoteEpisode == null ||
+                    trackedDownload.RemoteEpisode.Series == null ||
+                    !trackedDownload.RemoteEpisode.Episodes.Any())
```
kayone: `Empty()` instead of `!.Any()` :)

kayone: general rules is to try to avoid ! if you can since it's easier to miss.

src/NzbDrone.Core/Download/TrackedDownloads/TrackedDownloadService.cs
```diff
@@ -57,23 +58,33 @@ public TrackedDownload TrackDownload(DownloadClientDefinition downloadClient, Do
...
                     {
-                        return null;
+                        parsedEpisodeInfo = Parser.Parser.ParseTitle(firstHistoryItem.SourceTitle);
+                        if (parsedEpisodeInfo != null)
```
kayone: always wrap your `if` with `{}`

Kayomani:
Sorted @kayone,  I do quite agree with the empty/any rather than inversion - much easier to read.

Kayomani:
Let me know if your wanting any further work on this as I believe its compete.

My only remaining thoughts are perhaps going a step further and doing away with parsing the torrent name entirely in the tracker service and flagging a warning if it does not parse in the completed download handler.  I presume it already marks it as not automatically importable if it does parse and doesn't match what was expected?

src/NzbDrone.Core/Download/TrackedDownloads/TrackedDownloadService.cs
```diff
@@ -57,23 +58,35 @@ public TrackedDownload TrackDownload(DownloadClientDefinition downloadClient, Do
...
-                if (remoteEpisode.Series == null || !remoteEpisode.Episodes.Any())
+                if (historyItems.Any())
                 {
-                    var historyItems = _historyService.FindByDownloadId(downloadItem.DownloadId);
+                    var firstHistoryItem = historyItems.OrderByDescending(h => h.Date).First();
+                    trackedDownload.State = GetStateFromHistory(firstHistoryItem.EventType);
 
-                    if (historyItems.Empty())
+                    if (parsedEpisodeInfo == null ||
+                    trackedDownload.RemoteEpisode == null ||
```
markus101: Align all the clauses of the if statement (makes it a bit more clear they're part of the same conditional):

```
if (statement1 = 1 ||
    statment2 = 2 ||
    etc...)
{
// Do something
}
```

kayone:
also, one last thing before the merge, please squash the commits into 1, with a descriptive title, like `new: Downloads can be tracked by the source name in addition to the download name `

Kayomani:
Merged as requested

PR #653, opened 2015-07-01
Update the kickass url to https://kat.cr

kayone:


Taloth:
Migration is already used. And why force ppl to https?

kayone:
Kickass redirects to https, migration to https just saves a request and redirect.

Taloth:
Ah, nevermind then... Except for the migration number of course.

kayone:
I thought i rebased, I even had to change the number once. I'll check again.

kayone:
I don't see another 89, is that what you mean?

Taloth:
Markus has 89 for custom scripts. 
https://github.com/Sonarr/Sonarr/pull/552/files#diff-c18c50967e1a00514a4e616a1a6c3382R6

kayone:
So what will happen when I merege this before custom scripts merge?

Taloth:
Then your 90 migration will be executed first, and when markus merges in his stuff 89 will be executed. So the only requirement is that 90 doesn't affect 89 and vice versa.

And since it doesn't, feel free to merge it in.

kayone:
I don't think it'll run 89 if db version is at 90. does it?

Taloth:
it does, tried that out a year ago when we had like 4 internal PRs at the same time. But don't take my word for it, it takes 1 merge to test it locally.

kayone:
yeah, just tested it and works.

PR #654, opened 2015-07-01
Fix: When removing pending items from the queue remove all releases for that episode

kayone:
Right now, we only delete the current release being shown, which is annoying. lets say 2x720p releases and 2x1080p releases have been tracked by sonarr, and I have a delay for the episode. I decide that I'm behind on the show and I just don't want it to download that episode, I would have to remove it from the queue four times. it could also cause the user to think the remove button is broken since the episode they deleted keeps reappearing, not knowing that each time they are deleting the next release.

### DIFF
--- a/src/NzbDrone.Api/Queue/QueueActionModule.cs
+++ b/src/NzbDrone.Api/Queue/QueueActionModule.cs
@@ -55,7 +55,7 @@ namespace NzbDrone.Api.Queue
 
             if (pendingRelease != null)
             {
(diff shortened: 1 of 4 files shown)


markus101:
Who is this guy? :ship:

kayone:
The prince that was promised.

Taloth:
Hold on. Don't ship it yet.

This might mess up Delay Profile stuff, coz that's not implemented the way we would want it to.

PR #655, opened 2015-07-01
Dots should not be removed from indexer queries

Kayomani:
I don't know of any reason why dots would want to be stripped from queries, whilst that class is used for parsing and indexers the replacement only affects queries.

### DIFF
--- a/src/NzbDrone.Core/IndexerSearch/Definitions/SearchCriteriaBase.cs
+++ b/src/NzbDrone.Core/IndexerSearch/Definitions/SearchCriteriaBase.cs
@@ -23,13 +23,12 @@ namespace NzbDrone.Core.IndexerSearch.Definitions
         {
             get
             {
-                return SceneTitles.Select(GetQueryTitle).ToList();
+                return SceneTitles.SelectMany(GetQueryTitle).ToList();
             }
         }
 
-        public static string GetQueryTitle(string title)
+        public static IEnumerable<string> GetQueryTitle(string title)
         {
-            Ensure.That(title,() => title).IsNotNullOrWhiteSpace();
 
             var cleanTitle = BeginningThe.Replace(title, String.Empty);
 
@@ -37,10 +36,26 @@ namespace NzbDrone.Core.IndexerSearch.Definitions
             cleanTitle = SpecialCharacter.Replace(cleanTitle, "");
             cleanTitle = NonWordLessDot.Replace(cleanTitle, "+");
 
-            //remove any repeating +s
+            // Remove any repeating +s
             cleanTitle = Regex.Replace(cleanTitle, @"\+{2,}", "+");
             cleanTitle = cleanTitle.RemoveAccent();
-            return cleanTitle.Trim('+', ' ');
+            cleanTitle = cleanTitle.Trim('+', ' ');
+
+            if (!string.IsNullOrWhiteSpace(cleanTitle))
+            {
+                yield return cleanTitle;
+            }
+
+            // If the title contains a dot then add an additional title without it
+            // as some series drop them.
+            if (cleanTitle.IndexOf('.') > -1)
+            {
+                cleanTitle = cleanTitle.Replace(".", string.Empty);
+                if (!string.IsNullOrWhiteSpace(cleanTitle))
+                {
+                    yield return cleanTitle;
+                }
+            }
         }
     }
 }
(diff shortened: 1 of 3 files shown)


markus101:
Is this an issue with a particular indexer? Because at least for Chicago P.D. this appears to be how its listed.

Kayomani:
I have run across several instances where the dots should not be removed, you are right that those releases were done without the the dots in the name.  When I tested it BTN was able to find the chicago releases using the dot in the name but on further inspection some other trackers are unable to be that clever.  

Perhaps a better answer should be for the series to auto alias itself so it does the search/match on both?

I'll add some more examples when I can remember them.

http://predb.me/?search=Aldnoah.Zero

Taloth:
BTN is searched by tvdb id, so not using the title.
Newznab searched by tvrage id, if available. fallback to title.

Kayomani:
I tested the above with a couple of indexers that do not use the rage id and it now returns for all cases.

markus101:
What indexers did you test with? I have a feel this will cause issues with Fanzub based indexers (as we rely on multiple titles when searching). Do newznab indexers support `|` as an OR clause (which would be used whenever the it falls back to title searching).

src/NzbDrone.Core/IndexerSearch/Definitions/SearchCriteriaBase.cs
```diff
@@ -23,24 +23,39 @@ public List<String> QueryTitles
...
+            cleanTitle = NonWordLessDot.Replace(cleanTitle, "+");
 
-            //remove any repeating +s
+            // Remove any repeating +s
             cleanTitle = Regex.Replace(cleanTitle, @"\+{2,}", "+");
             cleanTitle = cleanTitle.RemoveAccent();
-            return cleanTitle.Trim('+', ' ');
+            cleanTitle = cleanTitle.Trim('+', ' ');
+
+            if (!string.IsNullOrWhiteSpace(cleanTitle))
```
markus101: `cleanTitle.IsNotNullOrWhiteSpace()`

src/NzbDrone.Core/IndexerSearch/Definitions/SearchCriteriaBase.cs
```diff
@@ -23,24 +23,39 @@ public List<String> QueryTitles
...
+            {
+                yield return cleanTitle;
+            }
+
+            // If the title contains a dot then add an additional title without it
+            // as some series drop them.
+            if (cleanTitle.IndexOf('.') > -1)
+            {
+                cleanTitle = cleanTitle.Replace(".", string.Empty);
+                if (!string.IsNullOrWhiteSpace(cleanTitle))
```
markus101: Same here.

src/NzbDrone.Core/IndexerSearch/Definitions/SearchCriteriaBase.cs
```diff
@@ -23,24 +23,39 @@ public List<String> QueryTitles
...
+            cleanTitle = cleanTitle.Trim('+', ' ');
+
+            if (!string.IsNullOrWhiteSpace(cleanTitle))
+            {
+                yield return cleanTitle;
+            }
+
+            // If the title contains a dot then add an additional title without it
+            // as some series drop them.
+            if (cleanTitle.IndexOf('.') > -1)
```
kayone: use `Contains(".")` it's more clear.

src/NzbDrone.Core/IndexerSearch/NzbSearchService.cs
```diff
@@ -210,9 +210,10 @@ private List<DownloadDecision> SearchSpecial(Series series, List<Episode> episod
             var searchSpec = Get<SpecialEpisodeSearchCriteria>(series, episodes);
             // build list of queries for each episode in the form: "<series> <episode-title>"
             searchSpec.EpisodeQueryTitles = episodes.Where(e => !String.IsNullOrWhiteSpace(e.Title))
-                                                    .SelectMany(e => searchSpec.QueryTitles.Select(title => title + " " + SearchCriteriaBase.GetQueryTitle(e.Title)))
+                                                    .SelectMany(e => searchSpec.QueryTitles.Select(title => title + " " + SearchCriteriaBase.GetQueryTitle(e.Title).LastOrDefault()))
```
kayone: why pick the last one?

PR #660, opened 2015-07-02
Update RemotePathMappingEditViewTemplate.hbs

DaveIsMe:
You need the IP of the Host and it seemed like you needed the name of the host with the original wording so I've suggested the world's smallest change.


Taloth:
Actually, no, you need to fill in the value used in the Host field when configuring the Download Client in Sonarr. which could be IP or dns name.

DaveIsMe:
Oh, my bad, I read it as wanting the name of the host you gave but I see what you mean now.

Taloth:
It does mean it could use better clarification.

DaveIsMe:
Yeah possibly. I was thinking could you have a dropdown box with the various options rather than having users type it themselves, which is prone to misunderstanding and typos. If I can teach myself some of this Handlebars.js stuff, I'll see if I can submit a better patch.

Issue #506, opened 2015-02-24
Add TeamCity build for mono compile


SonarrBot:
We need a separate TC build queue to compile and test using the mono compiler (but not as deliverable artifact). So we can verify the build keeps working.

Checklist
- [x] Create Build.sh
- [ ] Add Build queue
- [ ] Add VM to run build and tests (archlinux, freebsd, or something other than debian)

kayone:
https://github.com/Sonarr/Sonarr/pull/661

markus101:
No real driving factor for this, closing.

PR #661, opened 2015-07-04
Fix linux build and add build.sh

ta264:
This fixes the build on linux (xbuild complains about a few ambiguous calls) and adds build.sh, the first item in #506 

_output_mono seems to work for me.  _output_osx and _output_osx_app look ok compared to running build.ps1 on windows but I don't have a mac to test it on.  _output is missing all the pdbs - it might be possible to generate them but I'm not sure if there's any need.
http://stackoverflow.com/questions/13499384/is-it-possible-to-debug-assemblies-compiled-with-mono-xbuild-with-visual-studi

_tests looks ok and mono_tests.sh runs with almost all the tests passing.  However the integration tests  from NzbDrone.Integration.Test.dll all seem to fail.  I've put a log file [here](https://gist.github.com/ta264/e37f0db2e5b49014e02b) in case you can see what's going wrong.  I tried running them in VS from the delevop branch and they still fail, with a similar file not found error, so I don't think it's anything I've done.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,0 +1,257 @@
+#! /bin/bash
+msBuild='/c/Windows/Microsoft.NET/Framework64/v4.0.30319/'
+outputFolder='./_output'
+outputFolderMono='./_output_mono'
+outputFolderOsx='./_output_osx'
+outputFolderOsxApp='./_output_osx_app'
+testPackageFolder='./_tests/'
+testSearchPattern='*.Test/bin/x86/Release/*'
+sourceFolder='./src'
+updateFolder=$outputFolder/NzbDrone.Update
+updateFolderMono=$outputFolderMono/NzbDrone.Update
+
+CheckExitCode()
+{
+    "$@"
+    local status=$?
+    if [ $status -ne 0 ]; then
+        echo "error with $1" >&2
+        exit 1
+    fi
+    return $status
+}
+
+CleanFolder()
+{
+    local path=$1
+    local keepConfigFiles=$2
+
+    echo "Removing XMLDoc files"
+    local xmlfiles=( $(find $path -name "*.xml") )
+    for filename in "${xmlfiles[@]}"
+    do
+        if [ -e ${filename%.xml}.dll ] || [ -e ${filename%.xml}.exe ]  ; then
+            rm $filename
+        fi
+    done
+
+    find $path -name "*.transform" -exec rm "{}" \;
+
+    if [ $keepConfigFiles != true ] ; then
+        find $path -name "*.dll.config" -exec rm "{}" \;
+    fi
+
+    echo "Removing FluentValidation.Resources files"
+    find $path -name "FluentValidation.resources.dll" -exec rm "{}" \;
+    find $path -name "App.config" -exec rm "{}" \;
+
+    echo "Removing .less files"
+    find $path -name "*.less" -exec rm "{}" \;
+
+    echo "Removing vshost files"
+    find $path -name "*.vshost.exe" -exec rm "{}" \;
+
+    if [ -d $path/NuGet ] ; then
+        echo "Removing NuGet"
+        rm -rf $path/NuGet
+    fi
+
+    echo "Removing Empty folders"
+    find $path -depth -empty -type d -exec rm -r "{}" \;
+}
+
+
+
+AddJsonNet()
+{
+    rm $outputFolder/Newtonsoft.Json.*
+    cp $sourceFolder/packages/Newtonsoft.Json.*/lib/net35/*.dll $outputFolder
+    cp $sourceFolder/packages/Newtonsoft.Json.*/lib/net35/*.dll $outputFolder/NzbDrone.Update
+}
+
+BuildWithMSBuild()
+{
+    export PATH=$msBuild:$PATH
+    CheckExitCode MSBuild.exe $sourceFolder/NzbDrone.sln //t:Clean //m
+    CheckExitCode MSBuild.exe $sourceFolder/NzbDrone.sln //p:Configuration=Release //p:Platform=x86 //t:Build //m
+}
+
+BuildWithXbuild()
+{
+    export MONO_IOMAP=case
+    CheckExitCode xbuild /t:Clean $sourceFolder/NzbDrone.sln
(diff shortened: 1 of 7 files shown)


ta264:
Just noticed that aaff2965454e58c46fdb84149092a64eeb728587 breaks the windows build...

ta264:
I think this should work on windows and linux now.

ta264:
The integration tests are fixed with #663

build.sh
```diff
@@ -0,0 +1,179 @@
+#! /bin/bash
```
kayone: I would love to replace our powershell script with this, much rather maintain a single platform agnostic one vs having two different ones for windows and *nix.

kayone: I'm gussing you used the `build.ps` as base and just translated it over?

src/NzbDrone.Api/Authentication/EnableAuthInNancy.cs
```diff
@@ -39,8 +39,8 @@ public void Register(IPipelines pipelines)
                 pipelines.EnableBasicAuthentication(new BasicAuthenticationConfiguration(_authenticationService, "Sonarr"));                
             }
 
-            pipelines.BeforeRequest.AddItemToEndOfPipeline(RequiresAuthentication);
-            pipelines.AfterRequest.AddItemToEndOfPipeline(RemoveLoginHooksForApiCalls);
+            pipelines.BeforeRequest.AddItemToEndOfPipeline((Func<NancyContext, Response>) RequiresAuthentication);
```
kayone: I think someone did this before and it broke something in production, @markus101 knows the details.

kayone: https://github.com/Sonarr/Sonarr/pull/202#issuecomment-75893962

markus101: I believe this is safe (and @ta264 already realized the other way was broken), running the PR build to confirm though.

build.sh
ta264: Yup, just translated it function by function

ta264: I had a look and I think I should be able to make this work for the windows build, using git bash which should have been installed with git. Is that what you had in mind?

kayone: yes, we already have a dependency on git, and the build script is not really needed for day to day dev, it mostly just runs on the build servers.

ta264:
I've updated it so that it works on windows too.  As far as I can tell it gives the same output as running build.ps1 on windows.

build.sh
```diff
@@ -0,0 +1,239 @@
...
+}
+
+CleanFolder()
+{
+    local path=$1
+    local keepConfigFiles=$2
+
+    echo "Removing XMLDoc files"
+    find $path -name "*.xml" -delete
+    find $path -name "*.transform" -delete
```
kayone: I'm getting an error here in windows, 

```
Removing XMLDoc files
find: invalid predicate `-delete'
find: invalid predicate `-delete'
find: invalid predicate `-delete'
Removing FluentValidation.Resources files
find: invalid predicate `-delete'
find: invalid predicate `-delete'
Removing Empty folders
find: invalid predicate `-delete'
```

ta264: Maybe that was only added in the new msys with git2... I can switch it to
use -exec rm

kayone: let me update my msys.

ta264: I'm using RC 5 from here https://github.com/git-for-windows/git/releases

kayone: update to latest version, still same issue.

kayone: I think it would be nice if we could keep it compatible with latest stable, some people might not want to run pre-release software.

markus101: The `-delete` flag on find isn't supported by all systems, most do, but I don't think we can expect it to work.

ta264: I've swapped it to use exec rm. I'm getting some warnings when it deletes
empty directories - I'll have a look tomorrow

ta264: I think i've fixed the problem with CleanFolder now.  I also double checked
the files produced and, for me at least, I get identical output when I use
this or build.ps1 on windows.  The output on linux is slightly different -
a few mdbs for the NuGet packages are missing and it adds an extra
dll, Interop.NetFwTypeLib.dll

kayone:
@ta264 can you squash this so we can merge it in,

Thanks.

TomAndrews:
@kayone I've pushed a squashed commit.  Thanks.

kayone:
I tried running the build in OSX and got, 

```
/Users/kayone/git/Sonarr/src/NzbDrone.sln (Build) ->
(Build target) ->
/Users/kayone/git/Sonarr/src/NzbDrone.Common/NzbDrone.Common.csproj (default targets) ->
/Library/Frameworks/Mono.framework/Versions/4.0.2/lib/mono/4.5/Microsoft.CSharp.targets (CoreCompile target) ->

	Cache/Cached.cs(19,17): error CS0118: `object' is a `type' but a `variable' was expected
	Cache/Cached.cs(19,17): error CS0131: The left-hand side of an assignment must be a variable, a property or an indexer
	Composition/ContainerBuilderBase.cs(27,13): error CS0118: `NzbDrone.Common.Composition.Container' is a `type' but a `variable' was expected
	Composition/ContainerBuilderBase.cs(27,29): error CS0131: The left-hand side of an assignment must be a variable, a property or an indexer
	Exceptron/ExceptronClient.cs(71,13): error CS0118: `NzbDrone.Common.Exceptron.Configuration' is a `namespace' but a `variable' was expected
	Exceptron/ExceptronClient.cs(71,13): error CS0131: The left-hand side of an assignment must be a variable, a property or an indexer
	Http/HttpRequest.cs(14,13): error CS0118: `System.UriBuilder' is a `type' but a `variable' was expected
	Http/HttpRequest.cs(14,30): error CS0131: The left-hand side of an assignment must be a variable, a property or an indexer
	TinyIoC.cs(3007,17): error CS0118: `System.Type' is a `type' but a `variable' was expected
	TinyIoC.cs(3005,37): error CS0131: The left-hand side of an assignment must be a variable, a property or an indexer

```

kayone:
merged

PR #662, opened 2015-07-04
Extrapolate scene numbering but won't auto import

Taloth:
Work in progress.

I ran into the problem that thexem had mappings for early episodes in a season, where tvdb had higher numbers. A new episode got released and overwrote one of the older ones coz Sonarr blindly assumed those newer episodes were mapped 1 to 1.

As discussed there are two options: We can extrapolate numbering ourselves, or prevent import.

This PR fills in numbering for old (unmapped) seasons, as well as future seasons. Unless there's a season number mismatch.
The idea was to create a ImportSpec that would block the import if series.(UseSeasonNumbering = true and episode.SceneEpisodeNumber = null.) But after thinking about it a bit more it seems ugly to do it that way.

So instead I think we should add an extra field to Episode. 'UnknownSceneNumbering' = true.
We could set that flag for every episode in a season if there is at least one episode in that season with mappings. That gives us a clean ImportSpec to warn on.
Additionally, it allows us to be more aggressive with extrapolation. Extrapolating shifted episode numbering, and shifted season number, as long as all extrapolated episodes get that flag set.
The added bonus is that users maintaining thexem can be alerted more easily.
We could add an Advanced config option that disables the ImportSpec.

Lemme know what you think about that idea.

### DIFF
--- a/src/NzbDrone.Api/Episodes/EpisodeResource.cs
+++ b/src/NzbDrone.Api/Episodes/EpisodeResource.cs
@@ -24,6 +24,7 @@ namespace NzbDrone.Api.Episodes
         public Nullable<Int32> SceneAbsoluteEpisodeNumber { get; set; }
         public Nullable<Int32> SceneEpisodeNumber { get; set; }
         public Nullable<Int32> SceneSeasonNumber { get; set; }
+        public Boolean UnverifiedSceneNumbering { get; set; }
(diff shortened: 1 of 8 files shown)


Taloth:
Okay, updated the code. Seems to work. 

Using importspec to block import. But maybe I should have a decisionspec as well.

PR #663, opened 2015-07-05
Fix integration tests on mono

ta264:
When running the integration tests on mono from _tests, it looks for bin/NzbDrone.exe which doesn't exist.  This changes the path to ../_output_mono/NzbDrone.exe so that the tests work.

I'm not sure if the path should ever be "bin" but I left it in so it doesn't break anything.

I also changed the default path to use / rather than \\ since that seems to be compatible with both windows and linux

### DIFF
--- a/src/NzbDrone.Test.Common/NzbDroneRunner.cs
+++ b/src/NzbDrone.Test.Common/NzbDroneRunner.cs
@@ -33,19 +33,22 @@ namespace NzbDrone.Test.Common
             AppData = Path.Combine(Directory.GetCurrentDirectory(), "_intg_" + DateTime.Now.Ticks);
 
             var nzbdroneConsoleExe = "NzbDrone.Console.exe";
+            var nzbdroneDebugPath = "../../../../../_output/";
+            var nzbdroneReleasePath = "bin";
 
             if (OsInfo.IsNotWindows)
             {
                 nzbdroneConsoleExe = "NzbDrone.exe";
+                nzbdroneReleasePath = "../_output_mono/";
             }
 
             if (BuildInfo.IsDebug)
             {
-                Start("..\\..\\..\\..\\..\\_output\\NzbDrone.Console.exe");
+                Start(Path.Combine(nzbdroneDebugPath, nzbdroneConsoleExe));
             }
             else
             {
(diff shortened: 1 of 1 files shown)


src/NzbDrone.Test.Common/NzbDroneRunner.cs
```diff
@@ -33,19 +33,22 @@ public void Start()
             AppData = Path.Combine(Directory.GetCurrentDirectory(), "_intg_" + DateTime.Now.Ticks);
 
             var nzbdroneConsoleExe = "NzbDrone.Console.exe";
+            var nzbdroneDebugPath = "../../../../../_output/";
```
kayone: these also need to use path.combine, otherwise it breaks windows.

src/NzbDrone.Test.Common/NzbDroneRunner.cs
```diff
@@ -33,19 +33,22 @@ public void Start()
...
+            var nzbdroneReleasePath = "bin";
 
             if (OsInfo.IsNotWindows)
             {
                 nzbdroneConsoleExe = "NzbDrone.exe";
+                nzbdroneReleasePath = "../_output_mono/";
```
kayone: also here.

ta264:
Actually, I had a look at the build server and noticed that it manages to run the integration tests successfully on Ubuntu.

This line in the log file looks like it's copying some files around;
[Resolving artifact dependencies] 517 files retrieved from <Sonarr :: v2 - Pull Requests :: 1. Build, build #2.0.0.466 [id 85078]> for pattern [tests.zip!**
NzbDrone.**.tar.gz!NzbDrone/** => bin/]

So maybe it's the build script that needs fixing instead

kayone:
I was gonna point you to https://github.com/Sonarr/Sonarr/pull/661 and the realized that is also your PR. :smile:

kayone:
@ta264 do you still think these changes are valid?

ta264:
I suspect they're probably not valid.  There is certainly an issue, in that if you run the build script and then try to run `integration_mono.sh` a lot of the tests will fail.  They do work on the build server but I can't see how the TeamCity jobs are configured so I'm reluctant to suggest changing anything that might break them.  Shall I open an issue instead?

PR #664, opened 2015-07-06
Plex Server Authentication Improvements

markus101:
Fixes:

- Plex.tv password not taking effect immediately
- An incorrect plex.tv password won't lock out plex.tv account

@Sonarr/team-sonarr

### DIFF
--- a/src/NzbDrone.Core/Notifications/Plex/PlexServerProxy.cs
+++ b/src/NzbDrone.Core/Notifications/Plex/PlexServerProxy.cs
@@ -42,7 +42,7 @@ namespace NzbDrone.Core.Notifications.Plex
             var response = client.Execute(request);
 
             _logger.Trace("Sections response: {0}", response.Content);
-            CheckForError(response);
+            CheckForError(response, settings);
 
             return Json.Deserialize<PlexMediaContainer>(response.Content)
                        .Directories
@@ -58,7 +58,7 @@ namespace NzbDrone.Core.Notifications.Plex
             var response = client.Execute(request);
 
             _logger.Trace("Update response: {0}", response.Content);
-            CheckForError(response);
+            CheckForError(response, settings);
         }
 
         public void UpdateSeries(int metadataId, PlexServerSettings settings)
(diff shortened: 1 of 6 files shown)


src/NzbDrone.Core/Notifications/Plex/PlexServerProxy.cs
```diff
@@ -119,19 +119,19 @@ public List<PlexPreference> Preferences(PlexServerSettings settings)
...
             var response = client.Execute(request);
 
             _logger.Debug("Authentication Response: {0}", response.Content);
-            CheckForError(response);
+            CheckForError(response, settings);
 
             var user = Json.Deserialize<PlexUser>(JObject.Parse(response.Content).SelectToken("user").ToString());
 
-            _authCache.Set(username, user.AuthenticationToken);
+            _authCache.Set(settings.Username+settings.Password, user.AuthenticationToken);
```
kayone: don't need to do this here, since the get is going through the cache

src/NzbDrone.Core/Notifications/Plex/PlexServerProxy.cs
```diff
@@ -170,26 +170,40 @@ private RestRequest GetPlexServerRequest(string resource, Method method, PlexSer
...
 
-        private void CheckForError(IRestResponse response)
+        private void CheckForError(IRestResponse response, PlexServerSettings settings)
         {
             _logger.Trace("Checking for error");
 
             if (response.StatusCode == HttpStatusCode.Unauthorized)
             {
-                throw new PlexException("Unauthorized");
+                if (settings.Username.IsNullOrWhiteSpace())
```
kayone: we should be checking this before even attempting to authenticate, rather than after.

markus101: We're not trying to authenticate here (we don't try to authenticate if the username hasn't been set), but we can still get an unauthorized response and this lets us tailor the message in that case.

kayone: but how would you ever get here if we don't try to authenticate if there is no username?

markus101: Make a call to update the library and you'll get an unauthenticated response if authentication is required and either you didn't set an username/password or the username/password is wrong.

PR #665, opened 2015-07-06
Matching anime using full series title instead of cleaned title

Taloth:
@markus101 

Basically to be able to detect the differences between Working!! Working`!! Working!!! (there are some more shows with the same issue)

scenenaming table is no longer Distinct on ParseTerm (wasn't anyway, coz thexem results were distinct but then had our own table added separately)

The SceneMappingService will now look up using the cleantitle (like before) but have a list of results.
It will then try a full title match to get a specific result.
This should even handle slight naming differences with multiple series.
If the full match doesn't get a result then I do a Levensthein comparison to at least get one result to return.

Overall a fairly small change in the SceneMappingService, but had to adjust the Parser to do it.

Parser now returns title without cleaning. Afaik I changed all places where it's used to clean it as needed.
Had to adjust a whole lot of tests of course, but also change the parser a bit coz it got some !, ) and - mismatches in the parsed titles.

Personally I think I should just ditch levensthein there, and only use a lower-case compare as fallback and return null otherwise.


src/NzbDrone.Core/DataAugmentation/Scene/SceneMappingService.cs
```diff
@@ -154,7 +154,33 @@ private SceneMapping FindTvdbId(string title)
...
+            {
+                return null;
+            }
+
+            if (candidates.Count == 1)
+            {
+                return candidates.First();
+            }
+
+            var exactMatch = candidates
```
markus101: Only thing was a nitpick here to move the orderby to the first line then align on the `.` (Pedantic I know).

src/NzbDrone.Core/DataAugmentation/Scene/SceneMappingService.cs
```diff
@@ -154,7 +154,33 @@ private SceneMapping FindTvdbId(string title)
...
+                .OrderByDescending(v => v.SeasonNumber)
+                .FirstOrDefault(v => v.Title == title);
+
+            if (exactMatch != null)
+            {
+                return exactMatch;
+            }
+
+            var closestMatch = candidates
```
markus101: Samsies.

PR #674, opened 2015-07-09
Choose download folder for rTorrent

### LINKED ISSUE #626: Ability to choose download folder for rTorrent
rTorrent accepts a category (or at least we ask for one), but that doesn't control where the downloads go, we should have an option to choose the path.

Since rTorrent may not be on the same server autocompletion might be useless, but we should include it so people can use it if it is, worst case ontario is it won't work. We can do validation on the path if its on the same server, IE the host is `localhost` or `127.0.0.1`.

Leave the path blank to use default.

markus101:
@Sonarr/team-sonarr 

Closes #626

### DIFF
--- a/src/NzbDrone.Core/Download/Clients/rTorrent/RTorrentProxy.cs
+++ b/src/NzbDrone.Core/Download/Clients/rTorrent/RTorrentProxy.cs
@@ -16,9 +16,11 @@ namespace NzbDrone.Core.Download.Clients.RTorrent
         void AddTorrentFromUrl(string torrentUrl, RTorrentSettings settings);
         void AddTorrentFromFile(string fileName, byte[] fileContent, RTorrentSettings settings);
         void RemoveTorrent(string hash, RTorrentSettings settings);
-        void SetTorrentPriority(string hash, RTorrentSettings settings, RTorrentPriority priority);
+        void SetTorrentPriority(string hash, RTorrentPriority priority, RTorrentSettings settings);
         void SetTorrentLabel(string hash, string label, RTorrentSettings settings);
+        void SetTorrentDownloadDirectory(string hash, string directory, RTorrentSettings settings);
         bool HasHashTorrent(string hash, RTorrentSettings settings);
+        void StartTorrent(string hash, RTorrentSettings settings);
     }
 
     public interface IRTorrent : IXmlRpcProxy
@@ -26,10 +28,10 @@ namespace NzbDrone.Core.Download.Clients.RTorrent
         [XmlRpcMethod("d.multicall2")]
         object[] TorrentMulticall(params string[] parameters);
 
-        [XmlRpcMethod("load.start")]
-        int LoadURL(string target, string data);
+        [XmlRpcMethod("load.normal")]
+        int LoadUrl(string target, string data);
 
-        [XmlRpcMethod("load.raw_start")]
+        [XmlRpcMethod("load.raw")]
         int LoadBinary(string target, byte[] data);
 
         [XmlRpcMethod("d.erase")]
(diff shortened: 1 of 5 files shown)


src/NzbDrone.Core/Download/Clients/rTorrent/RTorrentProxy.cs
```diff
@@ -185,13 +174,71 @@ public void SetTorrentLabel(string hash, string label, RTorrentSettings settings
...
+            }
+            catch (Exception)
+            {
+                return false;
+            }
+        }
+
+        public void StartTorrent(string hash, RTorrentSettings settings)
+        {
+            _logger.Debug("Executing remote method: d.directory.set");
```
Taloth: Copy-paste messup. wrong log message

src/NzbDrone.Core/Download/Clients/rTorrent/RTorrentProxy.cs
```diff
@@ -185,13 +174,71 @@ public void SetTorrentLabel(string hash, string label, RTorrentSettings settings
...
+                                                      new
+                                                      {
+                                                          methodName = "d.start",
+                                                          @params = new[] { hash }
+                                                      },
+                                                  }).SelectMany(c => ((IEnumerable<int>)c));
+
+            if (muilticallResponse.Any(r => r != 0))
+            {
+                throw new DownloadClientException("Could not set download directory on torrent: {0}.", hash);
```
Taloth: Same

src/NzbDrone.Core/Download/Clients/rTorrent/RTorrentProxy.cs
```diff
@@ -185,13 +174,71 @@ public void SetTorrentLabel(string hash, string label, RTorrentSettings settings
...
+            }
+        }
+
+        public void StartTorrent(string hash, RTorrentSettings settings)
+        {
+            _logger.Debug("Executing remote method: d.directory.set");
+
+            var client = BuildClient(settings);
+
+            var muilticallResponse = client.SystemMulticall(new[]
```
Taloth: spelling

src/NzbDrone.Core/Download/Clients/rTorrent/RTorrentSettings.cs
```diff
@@ -52,10 +50,13 @@ public RTorrentSettings()
         [FieldDefinition(6, Label = "Category", Type = FieldType.Textbox, HelpText = "Adding a category specific to Sonarr avoids conflicts with unrelated downloads, but it's optional.")]
         public string TvCategory { get; set; }
 
-        [FieldDefinition(7, Label = "Recent Priority", Type = FieldType.Select, SelectOptions = typeof(RTorrentPriority), HelpText = "Priority to use when grabbing episodes that aired within the last 14 days")]
+        [FieldDefinition(7, Label = "Directory", Type = FieldType.Textbox, HelpText = "Optional location to put downloads in, leave blank to use the default rTorrent location")]
+        public string TvDirectory { get; set; }
```
Taloth: Think we should make this Advanced?

Taloth: Also, maybe we need to validate that this isn't anywhere in the series dir.

markus101: Advanced makes sense. I was thinking for validation we could check if its localhost/127.0.0.1 and validate then, otherwise don't check it. I like the idea of making sure its not a series folder/root folder (or drone factory?), though its possible they use the same path on different systems (seems weird, but you never know..), in that case just block it?

Taloth: Yes, you can validate only if local/127. It's just to prevent the biggest of messups.

markus101: My problem is we don't have a way to inject into the validator and I can't think of a clean way to do it, any thoughts?

Taloth: Do it in the test instead?

markus101:
@Sonarr/team-sonarr updated PR with validation, take a look so we can get it merged, please.

Taloth:
Ship it

Issue #606, opened 2015-06-07
Manual Import Option to Copy or Hardlink


TheDeathly:
I noticed when i had to manually import a few episodes it didn't follow that option, it seems to always copy which isnt exactly ideal sometimes. Would it be possible to have it follow that option, or add another set of options for the manual import?

markus101:
If you use manual import from the queue for a torrent it should respect the Hardlinking option and will either copy or hardlink if the torrent is still seeding. In all other cases it will import by moving the file(s) (Importing from usenet or by picking a folder manually).

I was thinking it would be good to have the option to copy instead of move though.

TheDeathly:
Oh you know what, it had moved the seeding files not copied which messed my torrents up. Thats why I didn't like manual import. I wouldn't have cared too much if it had copied, but hard linking should be the highest priority if its enabled i feel.

Taloth:
@markus101 Are you sure it respects the IsReadOnly flag on Manual Import via Queue?

markus101:
Without diving in too much, it should as the tracked download is passed.

https://github.com/Sonarr/Sonarr/blob/develop/src/NzbDrone.Core/MediaFiles/EpisodeImport/Manual/ManualImportService.cs#L228

waldnercharles:
I've run into this same issue. It would be nice to have an option to always copy / hardlink regardless of seeding status. I'm not sure how accurate Sonarr's detection of whether a file is being seeded or not is.

TheDeathly:
The issue is that when you do a Manual Import it isn't looking at any download clients, so it just always moves. It would be really nice to have a checkbox or something that lets you choose different moving methods; copy, hard/soft link, move.

markus101:
Manual Import has two entry points, one from the wanted page and another from the queue, if you start manual import from the queue the seeding status is captured and if its still seeding then its hardlinked or copied.

We should probably add a message in the UI that says whether or not it will copy/hardlink or move though.

NMe84:
If you're making a message out of it, might as well add a checkbox and make it customizable. :) I've run into this problem a few times today, where I have downloaded some season manually because Sonarr couldn't understand the release name, but when I wanted to keep seeding, I noticed all the files were gone. It would be very helpful to have the option to hardlink rather than move.

Blackkatt:
"Manual Import has two entry points, one from the wanted page and another from the queue, if you start manual import from the queue the seeding status is captured and if its still seeding then its hardlinked or copied."

[...]

PR #682, opened 2015-07-12
UI styling for Manual Import

Taloth:
Just a PR to start collecting some UI changes.

Change 1: Hover border highlight on editable columns: (also in episode details 'quality' cell)

image

Had to jump through quite a few hoops to get the div to behave, seems to work on Chrome, Firefox and IE, but dunno if there are side-effects.

Checklist
- [x] Editable cell border
- [ ] Better selection of season.   
Atm you get to select the series in this bulky dialog. Doesn't even show the currently selected season when opening.
- [ ] Set same season for multiple items
- [ ] sort on episode
- [ ] Once a series is selected try use scene numbering to auto match season/episode.

### DIFF
--- a/src/UI/Cells/NzbDroneCell.js
+++ b/src/UI/Cells/NzbDroneCell.js
@@ -1,3 +1,4 @@
+var _ = require('underscore');
 var Backgrid = require('backgrid');
 var Backbone = require('backbone');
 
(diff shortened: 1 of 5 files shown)


src/UI/Cells/NzbDroneCell.js
```diff
@@ -18,6 +19,25 @@ module.exports = Backgrid.Cell.extend({
...
             });
         }
+
+        this.render = _.wrap(this.render, this._renderDecorator);
+    },
+
+    _renderDecorator : function(renderInner) {
+        var el = this.$el;
+
+        if (this.$el.hasClass('editable')) {
```
markus101: I think it would be better to pass a property for the cell that renders the cell and adds the class instead of relying on the DOM to control the flow.

Taloth: one problem is that editability in backgrid is determined by the column, not the cell.

Kayomani:
Could I drop in a request for a way to import files and copy rather than move?  Occasionally I have to download things manually / have downloaded things manually using utorrent and i'd rather not remove the seeding copy.

kayone:
@Kayomani there is already a propsal for it https://github.com/Sonarr/Sonarr/issues/606

PR #690, opened 2015-07-16
Alter backbone.pageable to stay on same page during fetch

### LINKED ISSUE #670: Queue Page number resets to 1, content of second page is still displayed
This picture explains it best:

image

markus101:
@Sonarr/team-sonarr Not a fan of editing the this external library, but neither am I with copying over the entire method.

Closes #670

### DIFF
--- a/src/UI/Activity/Queue/QueueCollection.js
+++ b/src/UI/Activity/Queue/QueueCollection.js
@@ -1,8 +1,10 @@
 var _ = require('underscore');
 var PageableCollection = require('backbone.pageable');
+//var PageableCollection = require('../../Shared/Grid/SonarrPageableCollection');
 var QueueModel = require('./QueueModel');
 var FormatHelpers = require('../../Shared/FormatHelpers');
 var AsSortedCollection = require('../../Mixins/AsSortedCollection');
+var AsPageableCollection = require('../../Mixins/AsPageableCollection');
 var moment = require('moment');
 
 require('../../Mixins/backbone.signalr.mixin');
@@ -64,9 +66,9 @@ var QueueCollection = PageableCollection.extend({
 });
 
 QueueCollection = AsSortedCollection.call(QueueCollection);
+QueueCollection = AsPageableCollection.call(QueueCollection);
 
 var collection = new QueueCollection().bindSignalR();
 collection.fetch();
-
 
 module.exports = collection;
--- a/src/UI/Mixins/AsPageableCollection.js
+++ b/src/UI/Mixins/AsPageableCollection.js
@@ -1,0 +1,45 @@
+var _ = require('underscore');
+
+module.exports = function() {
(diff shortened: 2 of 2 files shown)


DaddyCaleb:
Although I am not that familiar with this JS library, there must be a way to resolve this without editing the external library (although not as elegant as this approach).   Modifying solid, well-tested code external to the project is just inviting stability and maintenance headaches.

Perhaps there is a way to call a method after this executes to reset the page location (ideally before the UI is updated)?  Or, maybe a listener that can be added when the event occurs that could alter its behavior?

markus101:
@Caleb-C felt terrible doing it, so I rewrote it, not the cleanest in the world, but it fixes the underlying issue and keeps the page number properly.

The bug has to do with the page number being reset, if you're on the 2nd page and suddenly there isn't a second page you'd be left with a grid that shows page 1, but really is on page 2 (with no items).

Caleb-C:
Bravo!  Good solution - clean, well-written, and does not touch the original library.

src/UI/Shared/Grid/SonarrPageableCollection.js
```diff
@@ -0,0 +1,42 @@
...
+            var currentPage = this.state.currentPage;
+            var pageSize = this.state.pageSize;
+            var pageStart = currentPage * pageSize;
+            var pageEnd = pageStart + pageSize;
+
+            originalEventHandler.apply(this, [].slice.call(arguments, 1));
+
+            var totalPages = Math.ceil(this.state.totalRecords / pageSize);
+
+            if (currentPage > totalPages) {
```
Taloth: why reset to 0? why not snap to last page?

And isn't it possible to simplify this whole piece of code to:
```
var newPage = Math.min(currentPage, totalPages - 1);
if (newPage !== this.state.currentPage) {
   this.getPage(newPage);
}
```

src/UI/Shared/Grid/SonarrPageableCollection.js
```diff
@@ -0,0 +1,42 @@
...
+        var eventHandler = PageableCollection.prototype._makeCollectionEventHandler.apply(this, arguments);
+
+        return _.wrap(eventHandler, _.bind(this._eventHandler, this));
+    },
+
+    _eventHandler : function (originalEventHandler, event, model, collection, options) {
+        if (event === 'reset') {
+            var currentPage = this.state.currentPage;
+            var pageSize = this.state.pageSize;
+            var pageStart = currentPage * pageSize;
```
Taloth: pageStart & End should be commented out too. Or just calculated when required.

markus101:
@Taloth not exactly, but I think this is cleaner now without a bunch of commented out/duplicated code, and also stays on the last available page. We don't want to change the page with getPage otherwise a reset is fired.

PR #721, opened 2015-08-01
Add tests for CurlHttpClient and fix some bugs they uncovered

ta264:
This updates HttpTestFixture.cs so that all the tests are run in the normal way and using curl.

I've added a couple of tests (for https and malformed cookie expiry dates) and made a few fixes so that all the tests pass.

### DIFF
--- a/src/NzbDrone.Common/Http/CurlHttpClient.cs
+++ b/src/NzbDrone.Common/Http/CurlHttpClient.cs
@@ -6,8 +6,10 @@ using System.Linq;
 using System.Net;
 using System.Runtime.InteropServices;
 using System.Text;
+using System.Text.RegularExpressions;
 using CurlSharp;
 using NLog;
+using NzbDrone.Common.EnvironmentInfo;
 using NzbDrone.Common.Extensions;
 using NzbDrone.Common.Instrumentation;
 
@@ -16,6 +18,7 @@ namespace NzbDrone.Common.Http
     public class CurlHttpClient
     {
         private static Logger Logger = NzbDroneLogger.GetLogger(typeof(CurlHttpClient));
+        private static readonly Regex ExpiryDate = new Regex(@"(expires=)([^;]+)", RegexOptions.IgnoreCase | RegexOptions.Compiled);
 
         public CurlHttpClient()
         {
@@ -64,7 +67,12 @@ namespace NzbDrone.Common.Http
                     curlEasy.HttpGet = webRequest.Method == "GET";
                     curlEasy.Post = webRequest.Method == "POST";
                     curlEasy.Put = webRequest.Method == "PUT";
-                    curlEasy.Url = webRequest.RequestUri.ToString();
+                    curlEasy.Url = webRequest.RequestUri.AbsoluteUri;
+
+                    if (OsInfo.IsWindows)
+                    {
+                        curlEasy.CaInfo = "curl-ca-bundle.crt";
+                    }
 
                     if (webRequest.CookieContainer != null)
                     {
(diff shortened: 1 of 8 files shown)


src/NzbDrone.Common/Http/HttpClient.cs
```diff
@@ -28,11 +28,14 @@ public class HttpClient : IHttpClient
         private readonly ICached<CookieContainer> _cookieContainerCache;
         private readonly ICached<bool> _curlTLSFallbackCache;
 
+        public bool forceCurl { get; set; }
```
markus101: Properties should be PascalCase (start with a capital).

src/NzbDrone.Common/Http/HttpClient.cs
```diff
@@ -120,7 +131,7 @@ private void PrepareRequestCookies(HttpRequest request, HttpWebRequest webReques
                     {
                         persistentCookieContainer.Add(new Cookie(pair.Key, pair.Value, "/", request.Url.Host)
                         {
-                            Expires = DateTime.UtcNow.AddHours(1)
+                            Expires = DateTime.Now.AddHours(1)
```
markus101: Why change to local time? UTC is preferred so its consistent.

src/NzbDrone.Common/Http/HttpClient.cs
```diff
@@ -196,6 +207,7 @@ private HttpResponse ExecuteRequest(HttpRequest request, HttpWebRequest webReque
 
         private HttpResponse ExecuteCurlRequest(HttpRequest request, HttpWebRequest webRequest)
         {
+            _logger.Trace("Curl request");
```
markus101: I don't see a lot of value with this logging message, something more verbose would be preferred.

src/NzbDrone.Common/Http/HttpClient.cs
```diff
@@ -203,6 +215,7 @@ private HttpResponse ExecuteCurlRequest(HttpRequest request, HttpWebRequest webR
 
         private HttpResponse ExecuteWebRequest(HttpRequest request, HttpWebRequest webRequest)
         {
+            _logger.Trace("HttpWebRequest");
```
markus101: Same for this logging message.

src/NzbDrone.Common/Http/HttpClient.cs
markus101: Is this purely for testing? Can we do something else rather than setting a property thats never otherwise used?

src/NzbDrone.Common/Http/HttpClient.cs
ta264: I'm in England (so UTC+1 at the moment) and when I run the tests on linux, a cookie with expiry UtcNow.AddHours(1) shows as expired as soon as it's created so it never gets sent.  I've put a test case here showing the problem:
https://gist.github.com/ta264/7822b1424f72e5b4c961

If you prefer, switching to DateTime.UtcNow.AddHours(24) should also work around the problem.

src/NzbDrone.Common/Http/HttpClient.cs
ta264: Yes, it's just for testing.  It seemed like the easiest way to be able to re-use all the existing tests without having to modify them.  I guess I could make forcing curl a property of the request rather than the client?  I'm no expert so definitely open to suggestions!

kayone: proper way to do this would be to separate out the dispatcher (Curl or WebClinet) into two different classes `IHttpDistapcher``CurlHttpDispatcher` and `ManagedHttpDispatcher`, then have a `DispatcherFactory` that you would take a `HttpReqeust` and it would return you an `IHttpDisptacher`.

src/NzbDrone.Common/Http/CurlHttpClient.cs
```diff
@@ -160,12 +167,26 @@ private WebHeaderCollection ProcessHeaderStream(HttpWebRequest webRequest, Strea
...
             }
 
             return webHeaderCollection;
         }
 
+        private string FixSetCookieHeader(string setCookie)
+        {
+            // fix up the date if it was malformed
+            string pattern = @"(expires=)([^;]+)";
+            Regex expiryMatch = new Regex(pattern, RegexOptions.IgnoreCase | RegexOptions.Compiled);
```
kayone: please move the regex to a field like we do with the rest of regexes. eg. https://github.com/Sonarr/Sonarr/blob/develop/src/NzbDrone.Core/IndexerSearch/Definitions/SearchCriteriaBase.cs

ta264:
I've had a go at separating out the dispatcher into it's own class.  I didn't use a factory because I wasn't sure how to make the curlTLSFallbackCache persist if you keep creating a new dispatcher.  What do you think?

kayone:
Can you squash your changes, there are some minor clean up but I can do that after the it's merged in.

ta264:
@kayone Thanks, I've squashed the commits.  Note that I've added a couple of lines to build.ps1 - something similar will need adding to build.sh.

kayone:
@ta264 I merged your `build.sh` branch can you rebase this from develop and add what is missing to `build.sh` 

Thanks,

ta264:
I've rebased from develop and added the lines to `build.sh`.  I also fixed the `should_follow_redirects()` test when using curl - you can end up with two sets of headers which throws an error.  Finally, I removed the unused `IHttpDispatcherFactory` declaration.

src/NzbDrone.Common/Http/CurlHttpClient.cs
```diff
@@ -152,20 +160,34 @@ private WebHeaderCollection ProcessHeaderStream(HttpWebRequest webRequest, Strea
 
             var webHeaderCollection = new WebHeaderCollection();
 
-            foreach (var header in headerString.Split(new[] { '\r', '\n' }, StringSplitOptions.RemoveEmptyEntries).Skip(1))
+            // following a redirect we could have two sets of headers, so only process the last one
+            foreach (var header in headerString.Split(new[] { '\r', '\n' }, StringSplitOptions.RemoveEmptyEntries).Reverse())
             {
+                if (!header.Contains(":")) break;
                 webHeaderCollection.Add(header);
             }
```
ta264: This is the fix for the `should_follow_redirects()` test.

kayone:
@ta264 thanks, I'll take it into a branch and do some clean up, I'll let you know so you can review the changes.

ta264:
I've rebased on develop to fix the conflicts with the HttpRequestInterceptor changes

PR #731, opened 2015-08-06
Add Hungarian Language

Mirx:
Based on discussion https://forums.sonarr.tv/t/custom-quality-definition-for-filtering-possible/6552/4

Please note, I'm not a programmer and haven't tested the changes. Just did a search on Dutch and tried to add similar behaviour.

### DIFF
--- a/src/NzbDrone.Core/Parser/Parser.cs
+++ b/src/NzbDrone.Core/Parser/Parser.cs
@@ -205,7 +205,7 @@ namespace NzbDrone.Core.Parser
         private static readonly Regex AnimeReleaseGroupRegex = new Regex(@"^(?:\[(?<subgroup>(?!\s).+?(?<!\s))\](?:_|-|\s|\.)?)",
                                                                 RegexOptions.IgnoreCase | RegexOptions.Compiled);
 
-        private static readonly Regex LanguageRegex = new Regex(@"(?:\W|_)(?<italian>\b(?:ita|italian)\b)|(?<german>german\b|videomann)|(?<flemish>flemish)|(?<greek>greek)|(?<french>(?:\W|_)(?:FR|VOSTFR)(?:\W|_))|(?<russian>\brus\b)|(?<dutch>nl\W?subs?)",
+        private static readonly Regex LanguageRegex = new Regex(@"(?:\W|_)(?<italian>\b(?:ita|italian)\b)|(?<german>german\b|videomann)|(?<flemish>flemish)|(?<greek>greek)|(?<french>(?:\W|_)(?:FR|VOSTFR)(?:\W|_))|(?<russian>\brus\b)|(?<dutch>nl\W?subs?)|(?<hungarian>\b(?:HUNDUB|HUN)\b)",
                                                                 RegexOptions.IgnoreCase | RegexOptions.Compiled);
 
         private static readonly Regex YearInTitleRegex = new Regex(@"^(?<title>.+?)(?:\W|_)?(?<year>\d{4})",
@@ -498,6 +498,9 @@ namespace NzbDrone.Core.Parser
             if (lowerTitle.Contains("portuguese"))
                 return Language.Portuguese;
 
+            if (lowerTitle.Contains("hungarian"))
+                return Language.Hungarian;
+                
             var match = LanguageRegex.Match(title);
 
             if (match.Groups["italian"].Captures.Cast<Capture>().Any())
@@ -520,6 +523,9 @@ namespace NzbDrone.Core.Parser
 
             if (match.Groups["dutch"].Success)
                 return Language.Dutch;
+
+            if (match.Groups["hungarian"].Success)
+                return Language.Hungarian;
 
             return Language.English;
         }
--- a/src/NzbDrone.Core.Test/ParserTests/LanguageParserFixture.cs
+++ b/src/NzbDrone.Core.Test/ParserTests/LanguageParserFixture.cs
@@ -43,6 +43,9 @@ namespace NzbDrone.Core.Test.ParserTests
         [TestCase("Constantine.2014.S01E01.WEBRiP.H264.AAC.5.1-NL.SUBS", Language.Dutch)]
         [TestCase("Elementary - S02E16 - Kampfhaehne - mkv - by Videomann", Language.German)]
         [TestCase("Two.Greedy.Italians.S01E01.The.Family.720p.HDTV.x264-FTP", Language.English)]
+        [TestCase("Castle.2009.S01E14.HDTV.XviD.HUNDUB-LOL", Language.Hungarian)]
+        [TestCase("Castle.2009.S01E14.HDTV.XviD.ENG.HUN-LOL", Language.Hungarian)]
+        [TestCase("Castle.2009.S01E14.HDTV.XviD.HUN-LOL", Language.Hungarian)]
         public void should_parse_language(string postTitle, Language language)
         {
             var result = Parser.Parser.ParseTitle(postTitle);
--- a/src/NzbDrone.Core/Parser/Language.cs
+++ b/src/NzbDrone.Core/Parser/Language.cs
@@ -23,6 +23,7 @@
         Portuguese = 18,
         Flemish = 19,
         Greek = 20,
-        Korean = 21
+        Korean = 21,
+        Hungarian = 22
     }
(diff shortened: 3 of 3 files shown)


gurabli:
Thanks @Mirx appreciated!
Can you check with this https://github.com/Sonarr/Sonarr/pull/731
It looks to me there is some kind of confusion:)

Mirx:
I changed the regex to include HUN and ENG.HUN.
(?<hungarian>HUNDUB|ENG.DUB|\bHUN_|\bHUN\b)

tested with below strings to see if it wasn't getting false positives:
[TestCase("Castle.2009.S01E14.HDTV.XviD.HUNDUB-LOL", Language.Hungarian)]
[TestCase("Castle.2009.S01E14.HDTV.XviD.ENG.HUN-LOL", Language.Hungarian)]
[TestCase("Castle.2009.S01E14.HDTV.XviD.HUN-LOL", Language.Hungarian)]
[TestCase("Castle.2009.S01E14.HDTV.XviD.HUN_LOL", Language.Hungarian)
[TestCase("Castle.2009.S01E14.HDTV.XviD.HUN.LOL", Language.Hungarian)]
[TestCase("Castle.2009.S01E14.HDTV.XviD.HUN_LOL", Language.Hungarian)]
HUNG
PHUN_

Mirx:
Would [_\.-]HUN[_\.-] or [_\W]HUN[_\W] be a better regex?

src/NzbDrone.Core/Parser/Parser.cs
```diff
@@ -205,7 +205,7 @@ public static class Parser
         private static readonly Regex AnimeReleaseGroupRegex = new Regex(@"^(?:\[(?<subgroup>(?!\s).+?(?<!\s))\](?:_|-|\s|\.)?)",
                                                                 RegexOptions.IgnoreCase | RegexOptions.Compiled);
 
-        private static readonly Regex LanguageRegex = new Regex(@"(?:\W|_)(?<italian>\b(?:ita|italian)\b)|(?<german>german\b|videomann)|(?<flemish>flemish)|(?<greek>greek)|(?<french>(?:\W|_)(?:FR|VOSTFR)(?:\W|_))|(?<russian>\brus\b)|(?<dutch>nl\W?subs?)",
+        private static readonly Regex LanguageRegex = new Regex(@"(?:\W|_)(?<italian>\b(?:ita|italian)\b)|(?<german>german\b|videomann)|(?<flemish>flemish)|(?<greek>greek)|(?<french>(?:\W|_)(?:FR|VOSTFR)(?:\W|_))|(?<russian>\brus\b)|(?<dutch>nl\W?subs?)|(?<hungarian>HUNDUB|ENG.HUN|[_\.-]HUN[_\.-])",
```
markus101: No need to check for `ENG.HUN` and `HUN` the latter captures the former.

`(?<hungarian>\b(?:HUNDUB|HUN)\b)`

Looks like we're not super consistent in our REGEX here, but the above should do it.

src/NzbDrone.Core/Parser/Parser.cs
```diff
@@ -498,6 +498,8 @@ public static Language ParseLanguage(string title)
             if (lowerTitle.Contains("portuguese"))
                 return Language.Portuguese;
 
+            if (lowerTitle.Contains("hungarian"))
+                return Language.Hungarian;
             var match = LanguageRegex.Match(title);
```
markus101: Needs a new line here to keep the styling consistent.

Mirx: Added the new line, and changed the regex.

Just for some additional info:
The new regex would not parse anything starting or ending with an _ correctly, as it is a word-character, and therefore not seeing it as the boundary.
As a result, _HUN would fail, as well as HUN_ No idea if that schema is utilized much though.

Also, as I'm Dutch, "hun" is a pretty common word in our language, so I wanted to limit [space]HUN[space] hits.

That's why I ended up enclosing HUN with the _ . or - character.

src/NzbDrone.Core/Parser/Parser.cs
```diff
@@ -205,7 +205,7 @@ public static class Parser
         private static readonly Regex AnimeReleaseGroupRegex = new Regex(@"^(?:\[(?<subgroup>(?!\s).+?(?<!\s))\](?:_|-|\s|\.)?)",
                                                                 RegexOptions.IgnoreCase | RegexOptions.Compiled);
 
-        private static readonly Regex LanguageRegex = new Regex(@"(?:\W|_)(?<italian>\b(?:ita|italian)\b)|(?<german>german\b|videomann)|(?<flemish>flemish)|(?<greek>greek)|(?<french>(?:\W|_)(?:FR|VOSTFR)(?:\W|_))|(?<russian>\brus\b)|(?<dutch>nl\W?subs?)",
+        private static readonly Regex LanguageRegex = new Regex(@"(?:\W|_)(?<italian>\b(?:ita|italian)\b)|(?<german>german\b|videomann)|(?<flemish>flemish)|(?<greek>greek)|(?<french>(?:\W|_)(?:FR|VOSTFR)(?:\W|_))|(?<russian>\brus\b)|(?<dutch>nl\W?subs?)|(?<hungarian>\b(?:HUNDUB|HUN)\b)",
```
markus101: If the concern is releases that contain underscores you could use `(?:\W|_)` or `[-_. ]` instead of `\b`, the latter option being more explicit.

Mirx: The last one was the one I had in there :p (except the space, which was on purpose)
I did a quick NZB indexer search and didn't see an _ results, so I think the suggested regex by you will be fine.

I was just trying to make it as fail-safe as possible, trying to learn something, and overshot a bit.

markus101: Why exclude the space? Space is probably more common than underscore.

markus101:
This looks good, the commits need to be squashed into one commit: http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html

The commit message should be:

```
New: Hungarian language support

Closes #729
```

Mirx:
You had me stumbled there for a minute. All my changes until now where on the website itself. The Windows editor apparently doesn't use force push when using the sync button, so all the squashes were undone. Finally managed to succeed through my NAS. Guess in the end, the push --force command in the powershell of the windows editor would also have done the trick.

Mirx:
You had me stumbled there for a minute. All my changes until now where on the website itself. The Windows editor apparently doesn't use force push when using the sync button, so all the squashes were undone. Finally managed to succeed through my NAS. Guess in the end, the push --force command in the powershell of the windows editor would also have done the trick.

gurabli:
Great work Mirx, and really appreciated!

One thing that I'm not sure now: since many episodes get released as dualdub, mostly English as original language and Hungarian as the dub in this case, the release format is most of the times: ***.ENG.HUN.***
Unfortunately, no mater how the scene tries to create strict general rules that everybody should follow, there are exceptions, like:
***.ENG.HUN-***

I also found: ***.HuN.EnG-*** While I'm not sure if search is case sensitive or not?

I think I have found most of these, but luckily things get better now in naming.

Mirx:
Not sure about the case, because I figured the search itself would be preformed case-insensitive.

You can easily test the current regex on various webpages like https://regex101.com/

Just copy the regex (?<hungarian>\b(?:HUNDUB|HUN)\b) into the top and add gi in the right field. G is for all hits, i is for case-insensitive.

Then in the bottom field, you can paste the text you want validated. 

All the options in the above post will be detected as Hungarian, when perfoming an case-insensitive search.

gurabli:
@Mirx Many thanks, this site is great, one can always learn new things!

I did the test with (\b(?:HUNDUB|HUN)\b) adding /gi and everything I could throw at it was detected, except: .HUN_??? So the underscore is not detected. Note, I was not able to find any release that used _ therefore I think it is not a problem at all. 

I'm not sure where does the search is performed, what I mean, the torrent file name is looked at and not the content? If the torrent was not named properly, it will not be detected, no matter if the content of a season pack is properly named?

Mirx:
Not accepting the underscore is already above in the regex discussion.
For the other question, I think it would be better to continue that on the sonarr forum. But to my knowledge, it is used to validate against the indexers results, which don't provide information of the content within.

PR #747, opened 2015-08-17
Release pushing

markus101:
A new PR to finish off the changes from https://github.com/Sonarr/Sonarr/pull/631

I extracted ReleaseModule to a base class so I could use MapDecisions in both modules (instead of making an almost identical resource to map a DownloadDecision to a resource).

@Taloth @kayone Please take a quick look so we can close this one up.


Taloth:
Looks good

PR #754, opened 2015-08-21
Support for REAL releases

### LINKED ISSUE #453: REAL releases should replace existing episodes of the same quality
Issue is a fake gets released, followed by the actual episode with REAL in the name, but its not grabbed because its not considered better than the original.

How should these be named on disk?
How should these be handled internally - bitwise!

markus101:
Made it case sensitive to `REAL` only.

Closes #453


thezoggy:
so `REAL` is the correct tag and probably will work as intended during rss/search... does that 'tag' hold true during import when the filename is probably all lower case (since its the release folder that has the actual tag)... and what about during renaming.. is REAL a new variable/part of the PROPER variable/or does it get gobbled up?

markus101:
Sonarr uses the quality of the folder when it is a better quality than the file (which is part of the issue with the 480p mkvs). At the moment file names wouldn't include REAL, I'll get that added.

Issue #348, opened 2014-04-30
WebHook Notifications


SonarrBot:
Send notification via web hook to the specified endpoint. The format will need to be documented, but this could be much more powerful than an external script.

halkeye:
Is there any discussion on how this should be done? I think I'd rather implement a generic webhook support than specifically try to integrate with slack but i'm not sure how generic it could be done... Maybe some sort of dynamic build your own payload system?

 https://api.slack.com/incoming-webhooks

markus101:
Not sure how generic you can make it without it being an epic pain to configure and hell to troubleshoot. Using TC webhooks (TeamCity plugin), they support a few different body types (JSON, key/value pairs) and offer the ability to add a custom properties.

I could see a key/value entry system, but then there is a lot of work to do make sure the user is entering valid data (similar to naming), but that might come with: https://github.com/Sonarr/Sonarr/issues/328

For the system I'm thinking something where you would enter the key name and then enter the value using variables (same as renaming), but this would be better suited for v3 and a better editing UI for notifications.

PR #759, opened 2015-08-23
Add Webhook support to sonarr

halkeye:
Add webhook support to sonarr

Add Form type url (type=url input field)
Add isValidUrl input type validation

### DIFF
--- a/src/NzbDrone.Common/Extensions/UrlExtensions.cs
+++ b/src/NzbDrone.Common/Extensions/UrlExtensions.cs
@@ -1,0 +1,32 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace NzbDrone.Common.Extensions
+{
+    public static class UrlExtensions
+    {
+        public static bool IsValidUrl(this string path)
+        {
+            if (string.IsNullOrWhiteSpace(path))
+            {
+                return false;
+            }
+
+            Uri uri;
+            if (!Uri.TryCreate(path, UriKind.Absolute, out uri))
+            {
+                return false;
+            }
+
+            if (!uri.IsWellFormedOriginalString())
+            {
+                return false;
+            }
+
+            return true;
+
+        }
+    }
+}
--- a/src/NzbDrone.Core/Notifications/Webhook/Webhook.cs
+++ b/src/NzbDrone.Core/Notifications/Webhook/Webhook.cs
@@ -1,0 +1,55 @@
+﻿
+using System.Collections.Generic;
+using FluentValidation.Results;
+using NzbDrone.Core.Tv;
+using NzbDrone.Common.Extensions;
+
+namespace NzbDrone.Core.Notifications.Webhook
+{
+    public class Webhook : NotificationBase<WebhookSettings>
+    {
+        private readonly IWebhookService _service;
+
+        public Webhook(IWebhookService service)
+        {
+            _service = service;
+        }
+
+        public override string Link
+        {
+            get { return "https://github.com/Sonarr/Sonarr/wiki/Webhook"; }
+        }
+
+        public override void OnGrab(GrabMessage message)
+        {
+            _service.OnGrab(message.Series, message.Episode, message.Quality, Settings);
+        }
+
+        public override void OnDownload(DownloadMessage message)
+        {
+            _service.OnDownload(message.Series, message.EpisodeFile, Settings);
+        }
+
+        public override void OnRename(Series series)
+        {
+            _service.OnRename(series, Settings);
+        }
+
+        public override string Name
+        {
+            get
+            {
+                return "Webhook";
+            }
+        }
+
+        public override ValidationResult Test()
+        {
+            var failures = new List<ValidationFailure>();
+
+            failures.AddIfNotNull(_service.Test(Settings));
+
+            return new ValidationResult(failures);
+        }
+    }
+}
--- a/src/NzbDrone.Core/Notifications/Webhook/WebhookEpisode.cs
+++ b/src/NzbDrone.Core/Notifications/Webhook/WebhookEpisode.cs
@@ -1,0 +1,32 @@
+﻿using NzbDrone.Core.Tv;
+using System;
+
+namespace NzbDrone.Core.Notifications.Webhook
+{
+    public class WebhookEpisode
+    {
+        public WebhookEpisode() { }
+
+        public WebhookEpisode(Episode episode)
+        {
+            Id = episode.Id;
+            SeasonNumber = episode.SeasonNumber;
+            EpisodeNumber = episode.EpisodeNumber;
+            Title = episode.Title;
+            AirDate = episode.AirDate;
+            AirDateUtc = episode.AirDateUtc;
+        }
+
+        public int Id { get; set; }
+        public int EpisodeNumber { get; set; }
+        public int SeasonNumber { get; set; }
+        public string Title { get; set; }
+        public string AirDate { get; set; }
+        public DateTime? AirDateUtc { get; set; }
+
+        public string Quality { get; set; }
+        public int QualityVersion { get; set; }
+        public string ReleaseGroup { get; set; }
+        public string SceneName { get; set; }
+    }
+}
--- a/src/NzbDrone.Core/Notifications/Webhook/WebhookPayload.cs
+++ b/src/NzbDrone.Core/Notifications/Webhook/WebhookPayload.cs
@@ -1,0 +1,11 @@
+﻿using System.Collections.Generic;
+
+namespace NzbDrone.Core.Notifications.Webhook
+{
+    public class WebhookPayload
+    {
+        public string EventType { get; set; }
+        public WebhookSeries Series { get; set; }
+        public List<WebhookEpisode> Episodes { get; set; }
+    }
+}
--- a/src/NzbDrone.Core/Notifications/Webhook/WebhookSeries.cs
+++ b/src/NzbDrone.Core/Notifications/Webhook/WebhookSeries.cs
@@ -1,0 +1,22 @@
+﻿using NzbDrone.Core.Tv;
+
+namespace NzbDrone.Core.Notifications.Webhook
+{
+    public class WebhookSeries
+    {
+        public int Id { get; set; }
+        public string Title { get; set; }
+        public string Path { get; set; }
+        public int TvdbId { get; set; }
+
+        public WebhookSeries() { }
+
+        public WebhookSeries(Series series)
+        {
+            Id = series.Id;
+            Title = series.Title;
+            Path = series.Path;
+            TvdbId = series.TvdbId;
+        }
+    }
+}
--- a/src/NzbDrone.Core/Notifications/Webhook/WebhookService.cs
+++ b/src/NzbDrone.Core/Notifications/Webhook/WebhookService.cs
@@ -1,0 +1,120 @@
+﻿using FluentValidation.Results;
+using NLog;
+using NzbDrone.Common.Processes;
+using NzbDrone.Core.MediaFiles;
+using NzbDrone.Core.Tv;
+using NzbDrone.Core.Validation;
+using NzbDrone.Core.Rest;
+using RestSharp;
+using NzbDrone.Core.Qualities;
+using NzbDrone.Core.Parser.Model;
+using System.Collections.Generic;
+
+namespace NzbDrone.Core.Notifications.Webhook
+{
+    public interface IWebhookService
+    {
+        void OnDownload(Series series, EpisodeFile episodeFile, WebhookSettings settings);
+        void OnRename(Series series, WebhookSettings settings);
+        void OnGrab(Series series, RemoteEpisode episode, QualityModel quality, WebhookSettings settings);
+        ValidationFailure Test(WebhookSettings settings);
+    }
+
+    public class WebhookService : IWebhookService
+    {
+        public void OnDownload(Series series, EpisodeFile episodeFile, WebhookSettings settings)
+        {
+            var payload = new WebhookPayload
+            {
+                EventType = "Download",
+                Series = new WebhookSeries(series),
+                Episodes = episodeFile.Episodes.Value.ConvertAll(x => new WebhookEpisode(x) {
+                    Quality = episodeFile.Quality.Quality.Name,
+                    QualityVersion = episodeFile.Quality.Revision.Version,
+                    ReleaseGroup = episodeFile.ReleaseGroup,
+                    SceneName = episodeFile.SceneName
+                })
+            };
+
+            NotifyWebhook(payload, settings);
+        }
+
+        public void OnRename(Series series, WebhookSettings settings)
+        {
+            var payload = new WebhookPayload
+            {
+                EventType = "Rename",
+                Series = new WebhookSeries(series)
+            };
+
+            NotifyWebhook(payload, settings);
+        }
+
+        public void OnGrab(Series series, RemoteEpisode episode, QualityModel quality, WebhookSettings settings)
+        {
+            var payload = new WebhookPayload
+            {
+                EventType = "Grab",
+                Series = new WebhookSeries(series),
+                Episodes = episode.Episodes.ConvertAll(x => new WebhookEpisode(x)
+                {
+                    Quality = quality.Quality.Name,
+                    QualityVersion = quality.Revision.Version,
+                    ReleaseGroup = episode.ParsedEpisodeInfo.ReleaseGroup
+                })
+            };
+            NotifyWebhook(payload, settings);
+        }
+
+        public void NotifyWebhook(WebhookPayload body, WebhookSettings settings)
+        {
+            try {
+                var client = RestClientFactory.BuildClient(settings.Url);
+                var request = new RestRequest((Method) settings.Method);
+                request.RequestFormat = DataFormat.Json;
+                request.AddBody(body);
+                client.ExecuteAndValidate(request);
+            }
+            catch (RestException ex)
+            {
+                throw new WebhookException("Unable to post to webhook: {0}", ex, ex.Message);
+            }
+        }
+
+        public ValidationFailure Test(WebhookSettings settings)
+        {
+            try
+            {
+                NotifyWebhook(
+                    new WebhookPayload
+                    {
+                        EventType = "Test",
+                        Series = new WebhookSeries()
+                        {
+                            Id = 1,
+                            Title = "Test Title",
+                            Path = "C:\\testpath",
+                            TvdbId = 1234
+                        },
+                        Episodes = new List<WebhookEpisode>() {
+                            new WebhookEpisode()
+                            {
+                                Id = 123,
+                                EpisodeNumber = 1,
+                                SeasonNumber = 1,
+                                Title = "Test title"
+                            }
+                        }
+                    },
+                    settings
+                );
+            }
+            catch (WebhookException ex)
+            {
+                return new NzbDroneValidationFailure("Url", ex.Message);
+            }
+
+            return null;
+        }
+    }
+}
--- a/src/NzbDrone.Core/Notifications/Webhook/WebhookSettings.cs
+++ b/src/NzbDrone.Core/Notifications/Webhook/WebhookSettings.cs
@@ -1,0 +1,38 @@
+﻿using System;
+using FluentValidation;
+using NzbDrone.Core.Annotations;
+using NzbDrone.Core.ThingiProvider;
+using NzbDrone.Core.Validation;
+using NzbDrone.Core.Validation.Paths;
+
+namespace NzbDrone.Core.Notifications.Webhook
+{
+    public class WebhookSettingsValidator : AbstractValidator<WebhookSettings>
+    {
+        public WebhookSettingsValidator()
+        {
+            RuleFor(c => c.Url).IsValidUrl();
+        }
+    }
+
+    public class WebhookSettings : IProviderConfig
+    {
+        private static readonly WebhookSettingsValidator Validator = new WebhookSettingsValidator();
+
+        public WebhookSettings()
+        {
+            Method = Convert.ToInt32(WebhookMethod.POST);
+        }
+
+        [FieldDefinition(0, Label = "URL", Type = FieldType.Url)]
+        public string Url { get; set; }
+
+        [FieldDefinition(1, Label = "Method", Type = FieldType.Select, SelectOptions = typeof(WebhookMethod), HelpText = "Which HTTP method to use submit to the Webservice")]
+        public Int32 Method { get; set; }
+
+        public NzbDroneValidationResult Validate()
+        {
+            return new NzbDroneValidationResult(Validator.Validate(this));
+        }
+    }
+}
--- a/src/NzbDrone.Common/NzbDrone.Common.csproj
+++ b/src/NzbDrone.Common/NzbDrone.Common.csproj
@@ -138,6 +138,7 @@
     <Compile Include="Extensions\Int64Extensions.cs" />
     <Compile Include="Extensions\ObjectExtensions.cs" />
     <Compile Include="Extensions\StreamExtensions.cs" />
+    <Compile Include="Extensions\UrlExtensions.cs" />
     <Compile Include="Extensions\XmlExtentions.cs" />
     <Compile Include="HashUtil.cs" />
     <Compile Include="Http\CurlHttpClient.cs" />
--- a/src/NzbDrone.Core/Annotations/FieldDefinitionAttribute.cs
+++ b/src/NzbDrone.Core/Annotations/FieldDefinitionAttribute.cs
@@ -28,6 +28,7 @@ namespace NzbDrone.Core.Annotations
         Path,
         Hidden,
         Tag,
-        Action
+        Action,
+        Url
     }
 }
--- a/src/NzbDrone.Core/Notifications/Webhook/WebhookException.cs
+++ b/src/NzbDrone.Core/Notifications/Webhook/WebhookException.cs
@@ -1,0 +1,16 @@
+﻿using System;
+using NzbDrone.Common.Exceptions;
+
+namespace NzbDrone.Core.Notifications.Webhook
+{
+    public class WebhookException : NzbDroneException
+    {
+        public WebhookException(string message) : base(message)
+        {
+        }
+
+        public WebhookException(string message, Exception innerException, params object[] args) : base(message, innerException, args)
+        {
+        }
+    }
+}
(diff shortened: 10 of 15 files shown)


halkeye:
its a start for #348

I'm working on a sonarr plugin for hubot - https://github.com/halkeye/hubot-sonarr

src/NzbDrone.Core/Notifications/Webhook/Webhook.cs
```diff
@@ -0,0 +1,62 @@
...
+        public override void OnRename(Series series)
+        {
+            _service.OnRename(series, Settings);
+        }
+
+        public override bool SupportsOnGrab
+        {
+            get
+            {
+                return false;
```
kayone: any reasons why this was left out?

src/NzbDrone.Common/Extensions/UrlExtensions.cs
```diff
@@ -0,0 +1,30 @@
...
+{
+    public static class UrlExtensions
+    {
+        public static bool IsValidUrl(this string path)
+        {
+            if (string.IsNullOrWhiteSpace(path))
+                return false;
+
+            Uri uri;
+            if (!Uri.TryCreate(path, UriKind.RelativeOrAbsolute, out uri))
```
kayone: I don't think it makes sense to allow relative URLs, unless sonnar would want to send notifications to itself. :smile:

src/NzbDrone.Core/Notifications/Webhook/Webhook.cs
halkeye: only because it only has a string message and no actual data? no real reason though.

halkeye: and "CustomScript" skips it as well

src/NzbDrone.Common/Extensions/UrlExtensions.cs
halkeye: Very good point, fixing

src/NzbDrone.Core/Notifications/Webhook/Webhook.cs
kayone: I think it would make sense to leave it in, for example if connecting it to slack or an IRC bot you might want to be notified of new grabs.

kayone: custom script skips it generally because custom script is there to process the downloaded file, in case of grab there is nothing there to process.

halkeye: How about we leave this one disabled, and I create a new patch that turns onGrab from being a string to something like DownloadMessage object?

Least then the webhook could post more than just a string?

I'm in no way against it, it just seems like a kinda hack.
I guess I could throw in the kinda hack until the next patch is done.

halkeye:
@kayone I'll update this OnGrab if #760 gets accepted

src/NzbDrone.Core/Notifications/Webhook/Webhook.cs
markus101: I think it makes sense to allow on grab, same as other "text" notifiers do.

src/NzbDrone.Core/Notifications/Webhook/WebhookService.cs
```diff
@@ -0,0 +1,92 @@
...
+
+        public WebhookService(IProcessProvider processProvider, Logger logger)
+        {
+            _processProvider = processProvider;
+            _logger = logger;
+        }
+
+        public void OnDownload(Series series, EpisodeFile episodeFile, WebhookSettings settings)
+        {
+            this.NotifyWebhook("Download", series, episodeFile, settings);
```
markus101: `this.` is not required (and we leave it out).

src/NzbDrone.Core/Notifications/Webhook/WebhookSettings.cs
```diff
@@ -0,0 +1,30 @@
...
+            RuleFor(c => c.Url).IsValidUrl();
+        }
+    }
+
+    public class WebhookSettings : IProviderConfig
+    {
+        private static readonly WebhookSettingsValidator Validator = new WebhookSettingsValidator();
+
+        [FieldDefinition(0, Label = "URL", Type = FieldType.Url)]
+        public String Url { get; set; }
```
markus101: nit: `string` instead of `String` (things are a bit all over with this one, but we've landed on using `string`).

halkeye: Okay I've already shutdown for the night so I'll push a new patch in the
morning with all the stylistic changes in the morning.

If I can I'll throw in the text only on grab too but that might not be till
evening.

halkeye:
I think i've fixed the outstanding issues @markus101 @kenyonj

src/NzbDrone.Core/Notifications/Webhook/WebhookService.cs
```diff
@@ -0,0 +1,97 @@
...
+
+        public WebhookService(IProcessProvider processProvider, Logger logger)
+        {
+            _processProvider = processProvider;
+            _logger = logger;
+        }
+
+        public void OnDownload(Series series, EpisodeFile episodeFile, WebhookSettings settings)
+        {
+            NotifyWebhook(new { EventType = "Download", Series = series, EpisodeFile = episodeFile }, settings);
```
kayone: I would much prefer us having a separate data model we use for webhook payload instead of relying on `Series`, `EpisodeFile` which are considered internal contracts. that way if we add/remove/rename stuff from our internal models the webhook implementations are protected by the mapping.

it would be great if you could create a new class, `WebHookSeries`, `WebHookEpisode` and add whatever fields from `Series`, `Episodes` that you think makes sense to an external application. we want to be selective instead of copying the whole thing over since we can add stuff later but once something is added it can't be removed since it would be considered a breaking change.

kayone:
sorry if we are being a bit picky, but features like this are extremely hard to change once they are merged since they would be considered part of our API and changing them could break things, that's why it's very important to get them done as well as we possibly can on the initial run.

halkeye:
Its all good.

Should I just base it on the same fields that customscript uses?

kayone:
yes, that would be great, but break the fields into Series, Episode objects,

@markus101 anything we would change about custom scripts here (thinking of it as a second change)?

markus101:
Nothing comes to mind, webhooks should be a bit cleaner since we can serialize to JSON though :+1:

halkeye:
Who creates the new object? Webhook.cs? Or WebhookService.cs?

kayone:
I think it would be the service,

halkeye:
I wasn't sure if a helper function should be created, or it made sense to put in the constructor. I can code in C# but don't use it on a day to day basis.

halkeye:
@kayone @markus101 Any more feedback?
(not trying to be a pest, but I like to finish coding before i forget too much about what I was working on)

src/NzbDrone.Common/Extensions/UrlExtensions.cs
```diff
@@ -0,0 +1,30 @@
...
+using System.Linq;
+using System.Text;
+
+namespace NzbDrone.Common.Extensions
+{
+    public static class UrlExtensions
+    {
+        public static bool IsValidUrl(this string path)
+        {
+            if (string.IsNullOrWhiteSpace(path))
```
kayone: please add `{}`

halkeye: just a code style thing? I'll update my patch tongiht

src/NzbDrone.Core/Notifications/Webhook/Webhook.cs
```diff
@@ -0,0 +1,55 @@
...
+        private readonly IWebhookService _service;
+
+        public Webhook(IWebhookService service)
+        {
+            _service = service;
+        }
+
+        public override string Link
+        {
+            get { return "https://github.com/Sonarr/Sonarr/wiki/Webhook"; }
```
kayone: can you add a wiki page with sample requests that are sent by the webhook, example of the body and headers would be great, maybe mention that it's limited to `POST` and `JSON` for now,

src/NzbDrone.Core/Notifications/Webhook/WebhookSeries.cs
```diff
@@ -0,0 +1,21 @@
+﻿using NzbDrone.Core.Tv;
+
+namespace NzbDrone.Core.Notifications.Webhook
+{
+    internal class WebhookSeries
+    {
+
```
kayone: extra blank space,

src/NzbDrone.Core/Notifications/Webhook/WebhookPayload.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace NzbDrone.Core.Notifications.Webhook
+{
+    internal class WebhookPayload
+    {
+        public WebhookEpisodeFile EpisodeFile { get; internal set; }
```
kayone: can you move this to the end,

src/NzbDrone.Core/Notifications/Webhook/WebhookSeries.cs
```diff
@@ -0,0 +1,21 @@
...
+        public string Id { get; set; }
```
kayone: any reason you didn't use it as `int`?

src/NzbDrone.Core/Notifications/Webhook/WebhookSeries.cs
```diff
@@ -0,0 +1,21 @@
...
+        public string Title { get; set; }
+        public string Path { get; set; }
+        public string TvdbId { get; set; }
```
kayone: same here,

src/NzbDrone.Core/Notifications/Webhook/WebhookSeries.cs
halkeye: Nope, nice catch, was probably because I was autogenerating them and it still had the .ToString().

I'll update

src/NzbDrone.Core/Notifications/Webhook/WebhookService.cs
```diff
@@ -0,0 +1,101 @@
...
+        {
+            NotifyWebhook(new { EventType = "Grab", Message = grabMessage }, settings);
+        }
+
+        public void NotifyWebhook(object body, WebhookSettings settings)
+        {
+            try {
+                var client = RestClientFactory.BuildClient(settings.Url);
+                var request = new RestRequest(Method.POST);
+                request.RequestFormat = DataFormat.Json;
```
kayone: I'm thinking we might want to make the `ContentType` and `Method` configurable as part of settings, but not something we need right now,

halkeye: I don't think GET supports content type right? The rest would. If I did GET I'd have to append everything onto the URL, which would get pretty messy unless it just appended a giant json blob?

Could easily support PUT & PATCH as long as restsharp supports it

kayone:
once we merge https://github.com/Sonarr/Sonarr/pull/760 this should be able to get in with no issues.

Thanks for the great patch!

src/NzbDrone.Common/Extensions/UrlExtensions.cs
kayone: yeah, mainly, it's more explicit about where the if boundaries start and end, leaves less room for mistakes.

kayone:
I merged your other branch, you can rebase from develop to pickup the changes,

src/NzbDrone.Core/Notifications/Webhook/WebhookService.cs
kayone: I don't think `PATCH` makes sense either, it's too semantically wrong. I would say `PUT` and `POST` are the ones that make sense.

halkeye:
@kayone 
Rebased from develop. Added Method Selection and a few other cleanups

Why does EpisodeFile have a list of episodes? Is it for like a seasonpack? I think its safe for removal in this stuff?

Added wiki page 
https://github.com/Sonarr/Sonarr/wiki/Webhook

markus101:
> Why does EpisodeFile have a list of episodes? Is it for like a seasonpack? I think its safe for removal in this stuff?

Files with more than one episode, double episodes usually, not ever likely to see a full season in a single file (which would be rejected on import anyways).

Currently each notification is for a single episode so there is no need to have a list of them, though that will likely change later (reduce notifications when importing a season pack). Personally I'd expect the episode to be a separate subobject in the JSON.

src/NzbDrone.Core/Notifications/Webhook/WebhookEpisodeFile.cs
```diff
@@ -0,0 +1,33 @@
...
+{
+    internal class WebhookEpisodeFile
+    {
+        public WebhookEpisodeFile(Series series, EpisodeFile episodeFile)
+        {
+            Id = episodeFile.Id;
+            RelativePath = episodeFile.RelativePath.ToString();
+            FullPath = Path.Combine(series.Path, episodeFile.RelativePath).ToString();
+            SeasonNumber = episodeFile.SeasonNumber;
+            Episodes = episodeFile.Episodes.Value.ConvertAll(x => new WebhookEpisode(x));
```
markus101: This seems redundant when Episode is already on the main object.

src/NzbDrone.Core/Notifications/Webhook/WebhookEpisode.cs
```diff
@@ -0,0 +1,20 @@
...
+        public WebhookEpisode(Episode episode)
+        {
+            EpisodeNumber = episode.EpisodeNumber;
+            AirDate = episode.AirDate;
+            AirDateUtc = episode.AirDateUtc;
+        }
+
+        public string AirDate { get; internal set; }
+        public DateTime? AirDateUtc { get; internal set; }
+        public int EpisodeNumber { get; internal set; }
```
markus101: How about the episode title?

src/NzbDrone.Core/Notifications/Webhook/WebhookEpisodeFile.cs
```diff
@@ -0,0 +1,33 @@
...
+            SceneName = episodeFile.SceneName ?? string.Empty;
+        }
+
+        public int Id { get; set; }
+        public string FullPath { get; set; }
+        public string RelativePath { get; set; }
+        public int SeasonNumber { get; set; }
+        public List<WebhookEpisode> Episodes { get; set; }
+        public string Quality { get; set; }
+        public string QualityVersion { get; set; }
```
markus101: Why convert the int to a string?

halkeye:
> Currently each notification is for a single episode so there is no need to have a list of them, though that will likely change later (reduce notifications when importing a season pack). Personally I'd expect the episode to be a separate subobject in the JSON.

Me too, I'll remove the list from the EpisodeFile item

src/NzbDrone.Core/Notifications/Webhook/WebhookEpisodeFile.cs
halkeye: Good catch. I missed one from my copy and paste job from customscript

src/NzbDrone.Core/Notifications/Webhook/WebhookEpisode.cs
halkeye: Updated

src/NzbDrone.Core/Notifications/Webhook/Webhook.cs
halkeye: Updated

src/NzbDrone.Core/Notifications/Webhook/WebhookPayload.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace NzbDrone.Core.Notifications.Webhook
+{
+    internal class WebhookPayload
+    {
+        public string EventType { get; internal set; }
+        public WebhookSeries Series { get; internal set; }
+        public WebhookEpisodeFile EpisodeFile { get; internal set; }
+        public WebhookEpisode Episode { get; internal set; }
```
kayone: can we make this a list, that way in the future when we add support for bulk notifications (you get notification for a season pack) we don't have to break the API.

halkeye: 2 lists:
List<EpisodeFile> + List<Episode>
or 1 list:
List<Episode,EpisodeFile>

markus101: If we make episodes a list we'd need to make episode files a list as well since they could both be plural.

kayone: do we really need two separate items? couldn't we just have a list of episodes that contains the file info as well, right now it appears that random fields are spread across the two models,

kayone: just to clarify, I'm not saying it's random, I'm saying to a typical user of this API it would appear that it's random.

src/NzbDrone.Core/Notifications/Webhook/WebhookEpisode.cs
kayone: still missing the title?

src/NzbDrone.Core/Notifications/Webhook/WebhookPayload.cs
halkeye: > couldn't we just have a list of episodes that contains the file info as well

yea, that seems a lot cleaner to me. 

I need to figure out how to get restsharp to exclude nulls, but ongrab won't have any file information which kinda makes sense to me? I think i can massage the data to get the quality and release/scene group

kayone: so here is what I'm thinkings, we break what is currently `EpisodeFile` into two parts, the release info and file info,

we keep the release info separate as single item (since even a season pack would only have 1 release info), release info would be `sceneName`, `quality`, `releaseGroup` etc.

and we embed the file info into each episode, some will have duplicated info but if a consumer cares they can match items using the file path.

src/NzbDrone.Core/Notifications/Webhook/WebhookEpisodeFile.cs
```diff
@@ -0,0 +1,31 @@
...
+        public WebhookEpisodeFile(Series series, EpisodeFile episodeFile)
+        {
+            Id = episodeFile.Id;
+            RelativePath = episodeFile.RelativePath.ToString();
+            FullPath = Path.Combine(series.Path, episodeFile.RelativePath).ToString();
+            SeasonNumber = episodeFile.SeasonNumber;
+            Quality = episodeFile.Quality.Quality.Name;
+            QualityVersion = episodeFile.Quality.Revision.Version;
+            ReleaseGroup = episodeFile.ReleaseGroup ?? string.Empty;
+            SceneName = episodeFile.SceneName ?? string.Empty;
```
kayone: I'm a bit torn on this, @markus101 what do you think? should we send `''`back when these are empty or keep them `null`?

src/NzbDrone.Core/Notifications/Webhook/WebhookPayload.cs
halkeye: so like:

```json
Episode: {
Airdate: blah,
Airdateutc
EpisodeNumber
SeasonNumber,
Title
},
ReleaseInfo: {
sceneName,
quality,
releaseGroup
}
```

I think sometimes, especially with torrents, season packs can have different scene names and quality?

kayone: if I remember correctly, as far as sonarr is concerned scene name is the name of the torrent/nzb, we don't actually use the internal filename of episodes inside of a pack, so `SceneName` should be unique. as for quality, that is a good point,

we could always move the quality from `ReleaseInfo` to `Episode`

markus101: Scene name is usually the release (folder) name, but in a season pack it could be the file name. Its very unlikely that a season pack would have multiple qualities though.

src/NzbDrone.Core/Notifications/Webhook/WebhookEpisodeFile.cs
markus101: I think null makes the most sense.

src/NzbDrone.Core/Notifications/Webhook/WebhookPayload.cs
kayone: @markus101 so what do you think, include everything in the `episode` object? is there a point in having release and episdoe separated?

halkeye: So i'll have time to work on it again tonight. I'm going to move forward with making a list of episodes and throwing in the data in there.

halkeye:
@kayone @markus101 Updated the code as per suggestions (I think, this pull request is getting messy).

I've also updated the wiki page with the the sample outputs. https://github.com/Sonarr/Sonarr/wiki/Webhook

src/NzbDrone.Core/Notifications/Webhook/WebhookService.cs
```diff
@@ -0,0 +1,113 @@
...
+
+        public WebhookService(IProcessProvider processProvider, Logger logger)
+        {
+            _processProvider = processProvider;
+            _logger = logger;
+        }
+
+        public void OnDownload(Series series, EpisodeFile episodeFile, WebhookSettings settings)
+        {
+            WebhookPayload payload = new WebhookPayload
```
markus101: `var` instead of `WebhookPayload` would be more consistent with what we normally use.

src/NzbDrone.Core/Notifications/Webhook/WebhookService.cs
```diff
@@ -0,0 +1,113 @@
...
+                {
+                    Quality = quality.Quality.Name,
+                    QualityVersion = quality.Revision.Version,
+                    ReleaseGroup = episode.ParsedEpisodeInfo.ReleaseGroup
+                })
+            };
+            NotifyWebhook(payload, settings);
+        }
+
+        public void NotifyWebhook(object body, WebhookSettings settings)
```
markus101: `public void NotifyWebhook(WebhookPayload payload, WebhookSettings settings)`

halkeye: How should I handle test then? Make a fake series? or just null both?

src/NzbDrone.Core/Notifications/Webhook/WebhookService.cs
halkeye: @markus101 replaced WebhookPayload with var

src/NzbDrone.Core/Notifications/Webhook/WebhookService.cs
kayone: I think that would make more sense, the other option which I think would be ok is to add a default constructor to WebhookPayload` and all it's subclasses, that way you can create a custom one rather than always having to create them using `Series` etc.

src/NzbDrone.Core/Notifications/Webhook/WebhookPayload.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System.Collections.Generic;
+
+namespace NzbDrone.Core.Notifications.Webhook
+{
+    internal class WebhookPayload
+    {
+        public string EventType { get; internal set; }
```
kayone: I think it's ok to make these `public` we generally try to avoid using `internal`, 

internal is normally used for libraries where you share your code and you don't want 3rd parties to access certains parts of it, which isn't the case with us.

halkeye: it was autogenerated :) Just make the set normal then?

kayone: yeah, I would say make everything that is internal either private or public.

halkeye:
@kayone okay, everything is using the webhook object. Also no more internal

src/NzbDrone.Core/Notifications/Webhook/WebhookService.cs
```diff
@@ -0,0 +1,129 @@
...
+        void OnDownload(Series series, EpisodeFile episodeFile, WebhookSettings settings);
+        void OnRename(Series series, WebhookSettings settings);
+        void OnGrab(Series series, RemoteEpisode episode, QualityModel quality, WebhookSettings settings);
+        ValidationFailure Test(WebhookSettings settings);
+    }
+
+    public class WebhookService : IWebhookService
+    {
+        private readonly IProcessProvider _processProvider;
+        private readonly Logger _logger;
```
kayone: I don't think these two are being used.

halkeye: Yep yep. Removed

kayone:
@halkeye Again, thanks for contributing, sorry for making you run around a bit :smile: 

@markus101 any blockers before we merge this in?

src/NzbDrone.Core/Notifications/Webhook/WebhookService.cs
```diff
@@ -0,0 +1,124 @@
...
+    {
+        void OnDownload(Series series, EpisodeFile episodeFile, WebhookSettings settings);
+        void OnRename(Series series, WebhookSettings settings);
+        void OnGrab(Series series, RemoteEpisode episode, QualityModel quality, WebhookSettings settings);
+        ValidationFailure Test(WebhookSettings settings);
+    }
+
+    public class WebhookService : IWebhookService
+    {
+        public WebhookService(IProcessProvider processProvider, Logger logger)
```
kayone: you don't actually need the constructor anymore since it doesn't actually do anything. You can just kill it.

halkeye:
No problem, Its a feature I want, so I want it to be right.

Looks like my apt's power is flickering, so I might not be able to do much more today

kayone:
It is pretty windy out there :D

halkeye:
@kayone remove constructor

Its kinda fun watching the various lose stuff flying around out there

markus101:
Nope, it was all good from my end. Power is out though, ugh.

kayone:
@halkeye I can merge this in once you squash and rebase it.

halkeye:
yay awesome~

Squashed!

I kept all the commit messages as one, for future reference, do you want me to delete all the extra commit messages?

halkeye:
Can I add https://github.com/halkeye/hubot-sonarr/ to https://github.com/Sonarr/Sonarr/wiki/Webhook as an example app that supports sonarr webhooks?

markus101:
Please do, hopefully it can help someone else get started.

kayone:
@halkeye we generally don't really worry about different commit messages, if the changes are isolated we keep the commits separated, but a feature like this fits well with a single commit. 

also, please make the commit message something that starts with "New:" like "New: Added support for webhook notifications" that way it'll show up in the change log.

halkeye:
@kayone I've updated the commit message

kayone:
Merged, Thank you!

PR #774, opened 2015-09-02
Preserve startup arguments during restart

### LINKED ISSUE #325: Portable Mode/Change AppData Location
We either need to support portable mode and have the AppData stored in the same folder ./appdata or allow users to specify its path (or both).

This will enable better support for unRAID-like systems, where a flash drive is used to store some data and everything else lives in RAM.

markus101:
@kayone @Taloth 

Same as during an update we will preserve the arguments passed in when Sonarr was opened.

Closes #325

### DIFF
--- a/src/NzbDrone.Host/SpinService.cs
+++ b/src/NzbDrone.Host/SpinService.cs
@@ -14,12 +14,14 @@ namespace NzbDrone.Host
     {
         private readonly IRuntimeInfo _runtimeInfo;
         private readonly IProcessProvider _processProvider;
+        private readonly IStartupContext _startupContext;
         private readonly Logger _logger;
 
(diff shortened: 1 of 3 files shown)


src/NzbDrone.Host/SpinService.cs
```diff
@@ -30,13 +32,27 @@ public void Spin()
...
             }
 
-            _logger.Debug("wait loop was terminated.");
+            _logger.Debug("Wait loop was terminated.");
 
             if (_runtimeInfo.RestartPending)
             {
-                _logger.Info("attemptig restart.");
-                _processProvider.SpawnNewProcess(_runtimeInfo.ExecutingApplication, "--restart --nobrowser");
+                _logger.Info("Attempting restart.");
```
kayone: might be useful to log the restart args.

markus101: Good call.

PR #778, opened 2015-09-03
Relax restriction on a pubDate element being provided in torrent RSS streams

PsychoLogicAu:
I have a couple of private torrent trackers which don't strictly conform to the RSS format Sonarr expects. They provide a release date within the description element, and may specify "Today, {time}", "Yesterday, {time}" or "{date}". This change allow parsing of the pubDate from this field if provided when missing the pubDate element.

This will extract a valid pubDate when specified in the following formats:
... Added: Added: Today, 12:24 AM <br/> ...
... Added: Yesterday, 06:43 PM <br/> ...
... Added: 30 Aug 2015 <br/> ...


src/NzbDrone.Core/Indexers/XElementExtensions.cs
```diff
@@ -36,6 +36,34 @@ public static XElement StripNameSpace(this XElement root)
             return res;
         }
 
+        public static DateTime ParseTodayDate(string timeString)
+        {
+            try
+            {
+                DateTime result = DateTime.Today.Date + DateTime.Parse(timeString).TimeOfDay;
```
markus101: `var` instead of `DateTime`

markus101:
Please add unit tests to support the new parsing. `BasicRssParserFixture` is probably a good location for them.

PsychoLogicAu:
I've added unit tests for both of the RSS feeds which prompted this. Change to .csproj file is the addition of .xml files to the project and setting their properties to 'Copy always'. It might be better to ignore that file and re-do the project modification.

markus101:
It looks like the line endings in the .csprog file changed. You'll need to make sure the line endings are committed as LF (linux style). For the tests adding a couple test RSS elements should be sufficient instead of the full feed.

src/NzbDrone.Core.Test/IndexerTests/TorrentRssIndexerTests/TorrentRssIndexerFixture.cs
```diff
@@ -151,5 +151,106 @@ public void should_parse_recent_feed_from_Doki()
...
+
+        [Test]
+        public void should_parse_recent_feed_from_HD4Free()
+        {
+            GivenRecentFeedResponse("TorrentRss/HD4Free.xml");
+
+            var releases = Subject.FetchRecent();
+
+            releases.Should().HaveCount(50);
+            releases.First().Should().BeOfType<TorrentInfo>();
```
markus101: You can check if all the releases are TorrentInfo with this (instead of for each element): `releases.ForEach(s => s.Should().BeOfType<TorrentInfo>());`

PsychoLogicAu: ForEach is not a method of IList<T> but I can use a foreach loop.

Taloth:
Oops, I missed this PR.

One, hd4free support torznab, use that instead of the rss.

Two, it would be much easier if BeyondHD added the pubdate element, it isn't that hard (https://github.com/Bigjoos/U-232-V4/blob/master/rss.php#L43, not saying they're using u-232-v4, most are on v3 or one of the original projects, but it's the same story mostly).
Or alternatively implement torznab, that ain't that hard either, but requires a few db changes so obviously more work than the above.

Fixing it at the source is infinitely better than trying to hack our way around it. I'm not opposed to relaxing parsing a bit, but please, first try to get the site to add pubdate properly.

PR #779, opened 2015-09-03
Download client for qBittorrent

cbodley:
I had experimented with a qBittorrent plugin a few months ago, but I wasn't able to get it working well because their Web UI API (https://github.com/qbittorrent/qBittorrent/wiki/WebUI-API-Documentation) didn't have a way to set labels. I've been tracking their "[Web UI] Labels implementation" pull request at https://github.com/qbittorrent/qBittorrent/pull/3444, and it looks like it may merge in the near future.

I updated my prototype to make use of the setLabel command it introduces, so I wanted to share it with you ahead of time for feedback.

### DIFF
--- a/src/NzbDrone.Core.Test/Download/DownloadClientTests/QBittorrentTests/QBittorrentFixture.cs
+++ b/src/NzbDrone.Core.Test/Download/DownloadClientTests/QBittorrentTests/QBittorrentFixture.cs
@@ -1,0 +1,297 @@
+using System;
+using System.Linq;
+using System.Collections.Generic;
+using FluentAssertions;
+using Moq;
+using NUnit.Framework;
+using NzbDrone.Common.Http;
+using NzbDrone.Core.MediaFiles.TorrentInfo;
+using NzbDrone.Core.Download;
+using NzbDrone.Core.Download.Clients.QBittorrent;
+using NzbDrone.Test.Common;
+
+namespace NzbDrone.Core.Test.Download.DownloadClientTests.QBittorrentTests
+{
+    [TestFixture]
+    public class QBittorrentFixture : DownloadClientFixtureBase<QBittorrent>
+    {
+        [SetUp]
+        public void Setup()
+        {
+            Subject.Definition = new DownloadClientDefinition();
+            Subject.Definition.Settings = new QBittorrentSettings
+                                          {
+                                              Host = "127.0.0.1",
+                                              Port = 2222,
+                                              Username = "admin",
+                                              Password = "pass",
+                                              TvCategory = "tv"
+                                          };
+
+            Mocker.GetMock<ITorrentFileInfoReader>()
+                  .Setup(s => s.GetHashFromTorrentFile(It.IsAny<Byte[]>()))
+                  .Returns("CBC2F069FE8BB2F544EAE707D75BCD3DE9DCF951");
+
+            Mocker.GetMock<IHttpClient>()
+                  .Setup(s => s.Get(It.IsAny<HttpRequest>()))
+                  .Returns<HttpRequest>(r => new HttpResponse(r, new HttpHeader(), new Byte[0]));
+        }
+
+        protected void GivenRedirectToMagnet()
+        {
+            var httpHeader = new HttpHeader();
+            httpHeader["Location"] = "magnet:?xt=urn:btih:ZPBPA2P6ROZPKRHK44D5OW6NHXU5Z6KR&tr=udp";
+
+            Mocker.GetMock<IHttpClient>()
+                  .Setup(s => s.Get(It.IsAny<HttpRequest>()))
+                  .Returns<HttpRequest>(r => new HttpResponse(r, httpHeader, new Byte[0], System.Net.HttpStatusCode.SeeOther));
+        }
+
+        protected void GivenRedirectToTorrent()
+        {
+            var httpHeader = new HttpHeader();
+            httpHeader["Location"] = "http://test.sonarr.tv/not-a-real-torrent.torrent";
+
+            Mocker.GetMock<IHttpClient>()
+                  .Setup(s => s.Get(It.Is<HttpRequest>(h => h.Url.AbsoluteUri == _downloadUrl)))
+                  .Returns<HttpRequest>(r => new HttpResponse(r, httpHeader, new Byte[0], System.Net.HttpStatusCode.Found));
+        }
+
+        protected void GivenFailedDownload()
+        {
+            Mocker.GetMock<IQBittorrentProxy>()
+                .Setup(s => s.AddTorrentFromUrl(It.IsAny<string>(), It.IsAny<QBittorrentSettings>()))
+                .Throws<InvalidOperationException>();
+        }
+
+        protected void GivenSuccessfulDownload()
+        {
+            Mocker.GetMock<IQBittorrentProxy>()
+                .Setup(s => s.AddTorrentFromUrl(It.IsAny<string>(), It.IsAny<QBittorrentSettings>()))
+                .Callback(() =>
+                {
+                    var torrent = new QBittorrentTorrent
+                    {
+                        Hash = "HASH",
+                        Name = _title,
+                        Size = 1000,
+                        Progress = 1.0,
+                        Eta = 8640000,
+                        State = "queuedUP",
+                        Label = "",
+                        SavePath = ""
+                    };
+                    GivenTorrents(new List<QBittorrentTorrent> { torrent });
+                });
+        }
+
+        protected virtual void GivenTorrents(List<QBittorrentTorrent> torrents)
+        {
+            if (torrents == null)
+                torrents = new List<QBittorrentTorrent>();
+
+            Mocker.GetMock<IQBittorrentProxy>()
+                .Setup(s => s.GetTorrents(It.IsAny<QBittorrentSettings>()))
+                .Returns(torrents);
+        }
+
+        [Test]
+        public void error_item_should_have_required_properties()
+        {
+            var torrent = new QBittorrentTorrent
+            {
+                Hash = "HASH",
+                Name = _title,
+                Size = 1000,
+                Progress = 0.7,
+                Eta = 8640000,
+                State = "error",
+                Label = "",
+                SavePath = ""
+            };
+            GivenTorrents(new List<QBittorrentTorrent> { torrent });
+
+            var item = Subject.GetItems().Single();
+            VerifyFailed(item);
+        }
+
+        [Test]
+        public void paused_item_should_have_required_properties()
+        {
+            var torrent = new QBittorrentTorrent
+            {
+                Hash = "HASH",
+                Name = _title,
+                Size = 1000,
+                Progress = 0.7,
+                Eta = 8640000,
+                State = "pausedDL",
+                Label = "",
+                SavePath = ""
+            };
+            GivenTorrents(new List<QBittorrentTorrent> { torrent });
+
+            var item = Subject.GetItems().Single();
+            VerifyPaused(item);
+            item.RemainingTime.Should().NotBe(TimeSpan.Zero);
+        }
+
+        [TestCase("pausedUP")]
+        [TestCase("queuedUP")]
+        [TestCase("uploading")]
+        [TestCase("stalledUP")]
+        [TestCase("checkingUP")]
+        public void completed_item_should_have_required_properties(string state)
+        {
+            var torrent = new QBittorrentTorrent
+            {
+                Hash = "HASH",
+                Name = _title,
+                Size = 1000,
+                Progress = 1.0,
+                Eta = 8640000,
+                State = state,
+                Label = "",
+                SavePath = ""
+            };
+            GivenTorrents(new List<QBittorrentTorrent> { torrent });
+
+            var item = Subject.GetItems().Single();
+            VerifyCompleted(item);
+            item.RemainingTime.Should().Be(TimeSpan.Zero);
+        }
+
+        [TestCase("queuedDL")]
+        [TestCase("checkingDL")]
+        public void queued_item_should_have_required_properties(string state)
+        {
+            var torrent = new QBittorrentTorrent
+            {
+                Hash = "HASH",
+                Name = _title,
+                Size = 1000,
+                Progress = 0.7,
+                Eta = 8640000,
+                State = state,
+                Label = "",
+                SavePath = ""
+            };
+            GivenTorrents(new List<QBittorrentTorrent> { torrent });
+
+            var item = Subject.GetItems().Single();
+            VerifyQueued(item);
+            item.RemainingTime.Should().NotBe(TimeSpan.Zero);
+        }
+
+        [Test]
+        public void downloading_item_should_have_required_properties()
+        {
+            var torrent = new QBittorrentTorrent
+            {
+                Hash = "HASH",
+                Name = _title,
+                Size = 1000,
+                Progress = 0.7,
+                Eta = 60,
+                State = "downloading",
+                Label = "",
+                SavePath = ""
+            };
+            GivenTorrents(new List<QBittorrentTorrent> { torrent });
+
+            var item = Subject.GetItems().Single();
+            VerifyDownloading(item);
+            item.RemainingTime.Should().NotBe(TimeSpan.Zero);
+        }
+
+        [Test]
+        public void stalledDL_item_should_have_required_properties()
+        {
+            var torrent = new QBittorrentTorrent
+            {
+                Hash = "HASH",
+                Name = _title,
+                Size = 1000,
+                Progress = 0.7,
+                Eta = 8640000,
+                State = "stalledDL",
+                Label = "",
+                SavePath = ""
+            };
+            GivenTorrents(new List<QBittorrentTorrent> { torrent });
+
+            var item = Subject.GetItems().Single();
+            VerifyWarning(item);
+            item.RemainingTime.Should().NotBe(TimeSpan.Zero);
+        }
+
+        [Test]
+        public void Download_should_return_unique_id()
+        {
+            GivenSuccessfulDownload();
+
+            var remoteEpisode = CreateRemoteEpisode();
+
+            var id = Subject.Download(remoteEpisode);
+
+            id.Should().NotBeNullOrEmpty();
+        }
+
+        [TestCase("magnet:?xt=urn:btih:ZPBPA2P6ROZPKRHK44D5OW6NHXU5Z6KR&tr=udp", "CBC2F069FE8BB2F544EAE707D75BCD3DE9DCF951")]
+        public void Download_should_get_hash_from_magnet_url(string magnetUrl, string expectedHash)
+        {
+            GivenSuccessfulDownload();
+
+            var remoteEpisode = CreateRemoteEpisode();
+            remoteEpisode.Release.DownloadUrl = magnetUrl;
+
+            var id = Subject.Download(remoteEpisode);
+
+            id.Should().Be(expectedHash);
+        }
+
+        [Test]
+        public void should_return_status_with_outputdirs()
+        {
+            var configItems = new Dictionary<string, Object>();
+
+            configItems.Add("save_path", @"C:\Downloads\Finished\QBittorrent".AsOsAgnostic());
+
+            Mocker.GetMock<IQBittorrentProxy>()
+                .Setup(v => v.GetConfig(It.IsAny<QBittorrentSettings>()))
+                .Returns(configItems);
+
+            var result = Subject.GetStatus();
+
+            result.IsLocalhost.Should().BeTrue();
+            result.OutputRootFolders.Should().NotBeNull();
+            result.OutputRootFolders.First().Should().Be(@"C:\Downloads\Finished\QBittorrent".AsOsAgnostic());
+        }
+
+        [Test]
+        public void Download_should_handle_http_redirect_to_magnet()
+        {
+            GivenRedirectToMagnet();
+            GivenSuccessfulDownload();
+
+            var remoteEpisode = CreateRemoteEpisode();
+
+            var id = Subject.Download(remoteEpisode);
+
+            id.Should().NotBeNullOrEmpty();
+        }
+
+        [Test]
+        public void Download_should_handle_http_redirect_to_torrent()
+        {
+            GivenRedirectToTorrent();
+            GivenSuccessfulDownload();
+
+            var remoteEpisode = CreateRemoteEpisode();
+
+            var id = Subject.Download(remoteEpisode);
+
+            id.Should().NotBeNullOrEmpty();
+        }
+    }
+}
--- a/src/NzbDrone.Core/Download/Clients/qBittorrent/DigestAuthenticator.cs
+++ b/src/NzbDrone.Core/Download/Clients/qBittorrent/DigestAuthenticator.cs
@@ -1,0 +1,22 @@
+﻿using RestSharp;
+using System.Net;
+
+namespace NzbDrone.Core.Download.Clients.QBittorrent
+{
+    public class DigestAuthenticator : IAuthenticator
+    {
+        private readonly string _user;
+        private readonly string _pass;
+
+        public DigestAuthenticator(string user, string pass)
+        {
+            _user = user;
+            _pass = pass;
+        }
+
+        public void Authenticate(IRestClient client, IRestRequest request)
+        {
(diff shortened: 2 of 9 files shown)


src/NzbDrone.Core/Download/Clients/qBittorrent/QBittorrent.cs
```diff
@@ -0,0 +1,228 @@
...
+                        item.Status = DownloadItemStatus.Warning;
+                        item.Message = "QBittorrent is reporting an error";
+                        break;
+
+                    case "pausedUP": // torrent is paused and has finished downloading
+                    case "pausedDL": // torrent is paused and has NOT finished downloading
+                        item.Status = DownloadItemStatus.Paused;
+                        break;
+
+                    case "queuedUP": // queuing is enabled and torrent is queued for upload
```
markus101: This should be completed.

src/NzbDrone.Core/Download/Clients/qBittorrent/QBittorrent.cs
```diff
@@ -0,0 +1,228 @@
...
+                    case "queuedDL": // queuing is enabled and torrent is queued for download
+                    case "checkingUP": // torrent has finished downloading and is being checked
```
markus101: Completed as well.

src/NzbDrone.Core/Download/Clients/qBittorrent/QBittorrentProxy.cs
```diff
@@ -0,0 +1,187 @@
...
+            var request = new RestRequest("/command/upload", Method.POST);
+            request.AddFile("torrents", fileContent, fileName);
+
+            var client = BuildClient(settings);
+            ProcessRequest(client, request, settings);
+        }
+
+        public void RemoveTorrent(String hash, Boolean removeData, QBittorrentSettings settings)
+        {
+            String cmd = removeData ? "/command/deletePerm" : "/command/delete";
```
markus101: `var` instead of `String`

src/NzbDrone.Core/Download/Clients/qBittorrent/QBittorrentTorrent.cs
```diff
@@ -0,0 +1,24 @@
+﻿using System;
+
+namespace NzbDrone.Core.Download.Clients.QBittorrent
+{
+    public class QBittorrentTorrent
+    {
+        public String Hash { get; set; } // Torrent hash
```
markus101: Things aren't currently strict, but we prefer, string, int, long and bool instead of String, Int32, Int64 and Boolean.

src/NzbDrone.Core/Download/Clients/qBittorrent/QBittorrentSettings.cs
```diff
@@ -0,0 +1,56 @@
...
+
+        public QBittorrentSettings()
+        {
+            Host = "localhost";
+            Port = 9091;
+            TvCategory = "tv-sonarr";
+        }
+
+        [FieldDefinition(0, Label = "Host", Type = FieldType.Textbox)]
+        public String Host { get; set; }
```
markus101: Same here, `string` instead of `String`.

src/NzbDrone.Core/Download/Clients/qBittorrent/QBittorrentProxy.cs
```diff
@@ -0,0 +1,187 @@
...
+            var restClient = RestClientFactory.BuildClient(url);
+
+            restClient.Authenticator = new DigestAuthenticator(settings.Username, settings.Password);
+            restClient.CookieContainer = _cookieContainer;
+
+            return restClient;
+        }
+    }
+
+    public class DigestAuthenticator : IAuthenticator
```
markus101: This should be in its own file.

src/NzbDrone.Core/Download/Clients/qBittorrent/QBittorrentProxy.cs
```diff
@@ -0,0 +1,187 @@
...
+        private IRestResponse ProcessRequest(IRestClient client, IRestRequest request, QBittorrentSettings settings)
+        {
+            _logger.Debug("Url: {0}", client.BuildUri(request));
+            var clientResponse = client.Execute(request);
+
+            if (clientResponse.StatusCode == HttpStatusCode.Forbidden)
+            {
+                _logger.Debug("QBittorrent authentication error.");
+
+                Login(client, settings);
```
markus101: This will trigger a login every request, instead we can cache that we authenticated correctly (same as we do for Plex Server).

src/NzbDrone.Core/Download/Clients/qBittorrent/QBittorrentTorrent.cs
markus101: I won't comment everywhere since there are a number of cases, but please update those as well.

src/NzbDrone.Core/Download/Clients/qBittorrent/QBittorrent.cs
```diff
@@ -0,0 +1,228 @@
...
+        public override void RemoveItem(string downloadId, bool deleteData)
+        {
+            _proxy.RemoveTorrent(downloadId, deleteData, Settings);
+        }
+
+        public override DownloadClientStatus GetStatus()
+        {
+            var config = _proxy.GetConfig(Settings);
+
+            OsPath destDir = new OsPath((String)config.GetValueOrDefault("save_path"));
```
markus101: `var` instead of `OsPath`

src/NzbDrone.Core/Download/Clients/qBittorrent/QBittorrent.cs
```diff
@@ -0,0 +1,228 @@
...
+            foreach (var torrent in torrents)
+            {
+                var item = new DownloadClientItem();
+                item.DownloadId = torrent.Hash;
+                item.Title = torrent.Name;
+                item.TotalSize = torrent.Size;
+                item.DownloadClient = Definition.Name;
+                item.RemainingSize = (long)(torrent.Size * (1.0 - torrent.Progress));
+                if (torrent.Eta != -1)
+                    item.RemainingTime = TimeSpan.FromSeconds(torrent.Eta);
```
markus101:
```
if (torrent.Eta != -1)
{
    item.RemainingTime = TimeSpan.FromSeconds(torrent.Eta);
}
```

and a line break before the if starts.

src/NzbDrone.Core/Download/Clients/qBittorrent/QBittorrent.cs
```diff
@@ -0,0 +1,228 @@
...
+
+                var outputPath = _remotePathMappingService.RemapRemoteToLocal(Settings.Host, status.OutputRootFolders[0]);
+
+                if (outputPath == null || outputPath.FileName == torrent.Name)
+                    item.OutputPath = outputPath;
```
markus101: Braces around the if and else.

src/NzbDrone.Core/Download/Clients/qBittorrent/QBittorrent.cs
```diff
@@ -0,0 +1,228 @@
...
+                var item = new DownloadClientItem();
+                item.DownloadId = torrent.Hash;
+                item.Title = torrent.Name;
+                item.TotalSize = torrent.Size;
+                item.DownloadClient = Definition.Name;
+                item.RemainingSize = (long)(torrent.Size * (1.0 - torrent.Progress));
+                if (torrent.Eta != -1)
+                    item.RemainingTime = TimeSpan.FromSeconds(torrent.Eta);
+
+                var outputPath = _remotePathMappingService.RemapRemoteToLocal(Settings.Host, status.OutputRootFolders[0]);
```
markus101: What if the download is under a different root folder?

src/NzbDrone.Core/Download/Clients/qBittorrent/QBittorrentPriority.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace NzbDrone.Core.Download.Clients.QBittorrent
+{
+    public enum QBittorrentPriority
+    {
+        Last = 0,
+        First = 1
+    }
+}
```
markus101: nit: newline at the end of the file.

src/NzbDrone.Core/Download/Clients/qBittorrent/QBittorrent.cs
```diff
@@ -0,0 +1,228 @@
...
+using NzbDrone.Core.RemotePathMappings;
+
+namespace NzbDrone.Core.Download.Clients.QBittorrent
+{
+    public class QBittorrent : TorrentClientBase<QBittorrentSettings>
+    {
+        private readonly IQBittorrentProxy _proxy;
+
+        public QBittorrent(IQBittorrentProxy proxy,
+                        ITorrentFileInfoReader torrentFileInfoReader,
```
markus101: Align classes to the first line.

src/NzbDrone.Core/Download/Clients/qBittorrent/QBittorrent.cs
```diff
@@ -0,0 +1,228 @@
...
+                        item.Status = DownloadItemStatus.Downloading;
+                        break;
+
+                    case "stalledDL": // torrent is being downloaded, but no connection were made
+                        item.Status = DownloadItemStatus.Warning;
+                        item.Message = "The download is stalled with no connections";
+                        break;
+
+                    default:
+                        throw new Exception("Invalid torrent state " + torrent.State);
```
markus101: If a new state is added this will break the integration completely since each torrent isn't handled in a try/catch block. We either need to wrap each torrent processing in a try/catch and log the error or log an error here and default the state to something, downloading maybe? Its kind of a toss up which one is best, I'll let the other devs weigh in here.

cbodley:
@markus101, thank you for all the great feedback! I don't have a lot of experience with the language, so I'm going to blame visual studio for most of the formatting stuff :) I'll gladly fix those, and put some more thought into the other issues as well.

markus101:
Thanks for contributing! I know we're a bit picky with formatting, so thanks for addressing those.

I took a look at implementing qBittorrent a couple weeks back and ran into similar issues, the blocker being the lack of label changing in the API, but I didn't like how the paths are returned and ended up hitting the API again to get the General Properties for each torrent, which is ugly and expensive. Perhaps we should ask about having it returned with the initial GetTorrents request?

cbodley:
@markus101 Okay, now I'm remembering the issues I had with OutputPath. I had ended up just defaulting everything to the global save_path from /query/preferences.

I went ahead and implemented the call to /query/propertiesGeneral for every torrent in GetItems() to check for a save_path. And I agree that we should work with the qBittorrent devs to get the save_path included with /query/torrents, so I added a patch that checks for that before falling back to /query/propertiesGeneral.

JustMaier:
Hey @markus101 just thought I'd reach out to check the status on reviewing this branch.

src/NzbDrone.Core/Download/Clients/qBittorrent/DigestAuthenticator.cs
```diff
@@ -0,0 +1,22 @@
+﻿using RestSharp;
+using System.Net;
+
+namespace NzbDrone.Core.Download.Clients.QBittorrent
+{
+public class DigestAuthenticator : IAuthenticator
```
markus101: Indentation is wrong.

src/NzbDrone.Core.Test/Download/DownloadClientTests/QBittorrentTests/QBittorrentFixture.cs
```diff
@@ -0,0 +1,291 @@
...
+            GivenSuccessfulDownload();
+
+            var remoteEpisode = CreateRemoteEpisode();
+
+            var id = Subject.Download(remoteEpisode);
+
+            id.Should().NotBeNullOrEmpty();
+        }
+
+        // Proxy.GetTorrents does not return original url. So item has to be found via magnet url.
```
markus101: This comment seems out of place since this isn't testing GetTorrents.

src/NzbDrone.Core/Download/Clients/qBittorrent/QBittorrent.cs
```diff
@@ -0,0 +1,252 @@
...
+            {
+                IsLocalhost = Settings.Host == "127.0.0.1" || Settings.Host == "localhost",
+                OutputRootFolders = new List<OsPath> { _remotePathMappingService.RemapRemoteToLocal(Settings.Host, destDir) }
+            };
+        }
+
+        private OsPath GetOutputPath(QBittorrentTorrent torrent)
+        {
+            // is save_path supported by /query/torrents?
+            if (torrent.SavePath != null)
```
markus101: Is this the new property from your PR? Is there a new minimum API version that we can enforce and forego the block below? I'm quite happy forcing a minimum version for a new release.

markus101:
@cbodley very sorry about missing your comment, apparently I read the notification when travelling and didn't look at the PR again. Thanks @JustMaier for bumping this.

From the QB side it looks like they still haven't merged the label PR, which is unfortunate, but I don't think its mandatory, more a nice to have to prevent Sonarr from trying to process something obscure. Has the SavePath change been published to a new build?

From my POV this looks pretty solid, let me know on the above and we'll look at getting it merged. Again, sorry for the super long delay.

src/NzbDrone.Core/Download/Clients/qBittorrent/QBittorrent.cs
cbodley: Yes, they bumped to the API version to 5. I'll add a check for that in QBittorrent::TestConnection.

cbodley:
@markus101 The SavePath property did make it into the 3.2.4 release. The labels PR is part of the 3.3.0 milestone.

Note that `QBittorrentSettingsValidator` currently requires a non-empty TvCategory. So if we want to support pre-3.3.0 clients, I'll have to:

1. allow an empty TvCategory setting,
2. provide a useful error message if they set TvCategory with a pre-3.3.0 client (API version < 6), and
3. make the label-related bits conditional on having a TvCategory

I'll look into that this weekend.

cbodley:
@markus101 Okay, we should do the right thing now for 3.2.4 and 3.3.0+ clients.

cbodley:
There are check failures, but I don't see any evidence that they were caused by this pull request. If there's any action needed on my part, please let me know!

markus101:
A good chunk of them are due to changes we made to fix the tests on our new build agents, could you rebase the PR off of latest develop please.

cbodley:
@markus101 Okay, I rebased and the green light is back!

cbodley:
According to https://github.com/qbittorrent/qBittorrent/issues/3800#issuecomment-158794604, the qBittorrent v3.3.0 release is planned for this coming weekend.

cbodley:
@markus101 v3.3.0 is out, and I've finally tested the new setLabel command. It didn't work at first because we give it a hash in uppercase. I added a patch for that, and it's working nicely.

src/NzbDrone.Core/Download/Clients/qBittorrent/QBittorrentProxy.cs
```diff
@@ -0,0 +1,210 @@
...
+            if (response.ResponseStatus != ResponseStatus.Completed)
+            {
+                throw new DownloadClientException("Transport error: {0}", response.ResponseStatus);
+            }
+
+            if (response.StatusCode == HttpStatusCode.Forbidden)
+            {
+                _logger.Info("Authentication required, logging in.");
+
+                var logged_in = _logins.Get(settings.Username + settings.Password, () => Login(client, settings), _loginTimeout);
```
markus101: `loggedIn` instead of `logged_in`

src/NzbDrone.Core/Download/Clients/qBittorrent/QBittorrent.cs
```diff
@@ -0,0 +1,248 @@
...
+                _logger.ErrorException(ex.Message, ex);
+                return Enumerable.Empty<DownloadClientItem>();
+            }
+
+            var queueItems = new List<DownloadClientItem>();
+
+            foreach (var torrent in torrents)
+            {
+                var item = new DownloadClientItem();
+                item.DownloadId = torrent.Hash;
```
markus101: We need to convert the Hash to upper case so Sonarr can track it properly.

src/NzbDrone.Core/Download/Clients/qBittorrent/QBittorrent.cs
```diff
@@ -0,0 +1,248 @@
...
+                item.Title = torrent.Name;
+                item.TotalSize = torrent.Size;
+                item.DownloadClient = Definition.Name;
+                item.RemainingSize = (long)(torrent.Size * (1.0 - torrent.Progress));
+                item.RemainingTime = TimeSpan.FromSeconds(torrent.Eta);
```
markus101: The item need a category set so Sonarr can track things it didn't grab

src/NzbDrone.Core/Download/Clients/qBittorrent/QBittorrent.cs
```diff
@@ -0,0 +1,248 @@
...
+                    {
+                        DetailedDescription = "Please upgrade to qBittorrent version 3.2.4 or higher."
+                    };
+                }
+                else if (version < 6 && Settings.TvCategory.Length > 0)
+                {
+                    // API version 6 introduced support for labels
+                    return new NzbDroneValidationFailure("Category", "Category is not supported")
+                    {
+                        DetailedDescription = "Labels are not supported until qBittorrent version 3.3.0. Please try again with an empty Category."
```
markus101: Maybe: `Labels are not supported until qBittorrent version 3.3.0. Upgrade or try again with an empty Category.`?

src/NzbDrone.Core/Download/Clients/qBittorrent/QBittorrentProxy.cs
```diff
@@ -0,0 +1,210 @@
...
+                throw new DownloadClientException("Transport error: {0}", response.ResponseStatus);
+            }
+
+            if (response.StatusCode != HttpStatusCode.OK)
+            {
+                _logger.Warn("Login failed with {0}.", response.StatusCode);
+                return false;
+            }
+
+            if (response.Content != "Ok.") // returns "Fails." on bad login
```
markus101: Is this localized at all?

src/NzbDrone.Core/Download/Clients/qBittorrent/QBittorrentSettings.cs
```diff
@@ -0,0 +1,55 @@
...
+    public class QBittorrentSettingsValidator : AbstractValidator<QBittorrentSettings>
+    {
+        public QBittorrentSettingsValidator()
+        {
+            RuleFor(c => c.Host).ValidHost();
+            RuleFor(c => c.Port).InclusiveBetween(0, 65535);
+        }
+    }
+
+    public class QBittorrentSettings : IProviderConfig
```
markus101: Support for HTTPS should be added, since thats available in the client.

src/NzbDrone.Core/Download/Clients/qBittorrent/QBittorrentProxy.cs
cbodley: No, it's either 'Ok.' or 'Fails.' according to https://github.com/qbittorrent/qBittorrent/blob/master/src/webui/webapplication.cpp#L180

src/NzbDrone.Core/Download/Clients/qBittorrent/QBittorrentSettings.cs
cbodley: Okay, I'll look into that.

cbodley: @markus101 I pushed a patch that adds a UseSsl setting as the other clients do. I was able to generate a certificate/key, add them to qBittorrent's webui, and test the client to verify that it works.

src/NzbDrone.Core/Download/Clients/qBittorrent/QBittorrent.cs
cbodley: @markus101 I did make this change. But on second thought, do you think it's worth keeping the compatibility code at all?

src/NzbDrone.Core/Download/Clients/qBittorrent/QBittorrent.cs
```diff
@@ -0,0 +1,249 @@
...
+            : base(torrentFileInfoReader, httpClient, configService, diskProvider, remotePathMappingService, logger)
+        {
+            _proxy = proxy;
+        }
+
+        protected override string AddFromMagnetLink(RemoteEpisode remoteEpisode, string hash, string magnetLink)
+        {
+            _proxy.AddTorrentFromUrl(magnetLink, Settings);
+
+            if (Settings.TvCategory.Length > 0)
```
markus101: `if (Settings.TvCategory.IsNotNullOrEmpty())`

src/NzbDrone.Core/Download/Clients/qBittorrent/QBittorrent.cs
```diff
@@ -0,0 +1,249 @@
...
+            }
+
+            return hash;
+        }
+
+        protected override string AddFromTorrentFile(RemoteEpisode remoteEpisode, string hash, string filename, Byte[] fileContent)
+        {
+            _proxy.AddTorrentFromFile(filename, fileContent, Settings);
+
+            if (Settings.TvCategory.Length > 0)
```
markus101: Same here.

markus101:
I think there are just a couple more things to fix up before we get it squashed and merged. :+1:

cbodley:
@markus101 excellent, thanks for all your help with review. I'll take care of those last comments this evening.

I did have a question for you regarding support for pre-3.3.0 clients, but it was hidden under an outdated diff:

Do you think it's worth keeping the compatibility code at all? At this point, it's probably more of a maintenance burden than it's worth, but that's ultimately up to you folks.

src/NzbDrone.Core/Download/Clients/qBittorrent/QBittorrent.cs
```diff
@@ -0,0 +1,249 @@
...
+                var version = _proxy.GetVersion(Settings);
+                if (version < 5)
+                {
+                    // API version 5 introduced the "save_path" property in /query/torrents
+                    return new NzbDroneValidationFailure("Hostname", "Unsupported client version")
+                    {
+                        DetailedDescription = "Please upgrade to qBittorrent version 3.2.4 or higher."
+                    };
+                }
+                else if (version < 6 && Settings.TvCategory.Length > 0)
```
markus101: Use IsNotNullOrEmpty() here as well.

markus101:
I'm totally okay with making the minimum 3.3.0, I saw that we effectively make the minimum 3.2.4 because of `save_path` and then 3.3.0 if they set a category.

We don't req

cbodley:
@markus101 Okay, I added that warning to recommend setting the category when it's supported. I also did some squashing and cleanup to remove history related to the save_path from /query/propertiesGeneral.

markus101:
Awesome @cbodley I'll try to test it tonight and get back to you ASAP. Thanks for implementing this.

src/NzbDrone.Core/Download/Clients/qBittorrent/QBittorrent.cs
```diff
@@ -0,0 +1,263 @@
...
+                return Enumerable.Empty<DownloadClientItem>();
+            }
+
+            var queueItems = new List<DownloadClientItem>();
+
+            foreach (var torrent in torrents)
+            {
+                var item = new DownloadClientItem();
+                item.DownloadId = torrent.Hash.ToUpper();
+                item.Category = Settings.TvCategory;
```
markus101: This needs to use the label from the response.

markus101:
The unit tests are failing now and the one issue with the label, but everything else looks okay. Once those are fixed up please squash the commits into one and we can get it merged.

cbodley:
@markus101 the label issue is fixed, and everything is squashed and rebased on develop

markus101:
Thanks @cbodley

JustMaier:
Thanks @cbodley - I really look forward to having this in Sonarr.

rterblanche:
When can we expect this in the develop branch or is there a beta branch. Will be happy to test this.

PR #783, opened 2015-09-03
Mono: Set process name

larsjohnsen:


markus101:
Not sure why you opened and closed 3 of these, but we have no plans to support this before version 3 incase it blows up on some OS.

PR #788, opened 2015-09-06
Add update notification

halkeye:
Adding basic support for notifications being told when system updates are available.

### DIFF
--- a/src/NzbDrone.Core/Notifications/Boxcar/Boxcar.cs
+++ b/src/NzbDrone.Core/Notifications/Boxcar/Boxcar.cs
@@ -39,7 +39,7 @@ namespace NzbDrone.Core.Notifications.Boxcar
         {
         }
 
-        public override void OnSystemUpdateAvailable(UpdatePackage package)
+        public override void OnUpdateAvailable(UpdatePackage package)
         {
             const string title = "New update is available";
             var body = String.Format("New update is available - {0}", package.Version.ToString());
--- a/src/NzbDrone.Core/Notifications/Boxcar/Boxcar.cs
+++ b/src/NzbDrone.Core/Notifications/Boxcar/Boxcar.cs
@@ -1,7 +1,9 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
 using FluentValidation.Results;
 using NzbDrone.Common.Extensions;
 using NzbDrone.Core.Tv;
+using NzbDrone.Core.Update;
 
 namespace NzbDrone.Core.Notifications.Boxcar
 {
@@ -37,6 +39,14 @@ namespace NzbDrone.Core.Notifications.Boxcar
         {
         }
 
+        public override void OnSystemUpdateAvailable(UpdatePackage package)
+        {
+            const string title = "New update is available";
+            var body = String.Format("New update is available - {0} - {1}", package.Version.ToString(), package.Url.ToString());
+
+            _proxy.SendNotification(title, body, Settings);
+        }
+
         public override string Name
         {
             get
@@ -61,5 +71,6 @@ namespace NzbDrone.Core.Notifications.Boxcar
 
             return new ValidationResult(failures);
         }
+
     }
 }
--- a/src/NzbDrone.Core/Notifications/Boxcar/Boxcar.cs
+++ b/src/NzbDrone.Core/Notifications/Boxcar/Boxcar.cs
@@ -42,7 +42,7 @@ namespace NzbDrone.Core.Notifications.Boxcar
         public override void OnSystemUpdateAvailable(UpdatePackage package)
         {
             const string title = "New update is available";
-            var body = String.Format("New update is available - {0} - {1}", package.Version.ToString(), package.Url.ToString());
+            var body = String.Format("New update is available - {0}", package.Version.ToString());
 
             _proxy.SendNotification(title, body, Settings);
         }
--- a/src/NzbDrone.Core/Notifications/CustomScript/CustomScript.cs
+++ b/src/NzbDrone.Core/Notifications/CustomScript/CustomScript.cs
@@ -1,6 +1,7 @@
 ﻿using System.Collections.Generic;
 using FluentValidation.Results;
 using NzbDrone.Core.Tv;
+using NzbDrone.Core.Update;
 
 namespace NzbDrone.Core.Notifications.CustomScript
 {
@@ -32,6 +33,10 @@ namespace NzbDrone.Core.Notifications.CustomScript
             _customScriptService.OnRename(series, Settings);
         }
 
+        public override void OnSystemUpdateAvailable(UpdatePackage package)
+        {
+        }
+
         public override string Name
         {
             get
@@ -41,6 +46,14 @@ namespace NzbDrone.Core.Notifications.CustomScript
         }
 
         public override bool SupportsOnGrab
+        {
+            get
+            {
+                return false;
+            }
+        }
+
+        public override bool SupportsOnSystemUpdateAvailable
         {
             get
             {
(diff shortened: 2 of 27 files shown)


halkeye:
Should notifications have this feature turned off by default?

halkeye:
I have another personal todo item to create a PR to update the notifications system so all the apis provide the settings as a whole object for consistency cause i noticed some older ones like XBMC not only pass each setting individually, but also in different orders.

markus101:
> Should notifications have this feature turned off by default?

Yes. If people want it they can opt in.

src/NzbDrone.Core/Notifications/Boxcar/Boxcar.cs
```diff
@@ -37,6 +39,14 @@ public override void OnRename(Series series)
         {
         }
 
+        public override void OnSystemUpdateAvailable(UpdatePackage package)
+        {
+            const string title = "New update is available";
+            var body = String.Format("New update is available - {0} - {1}", package.Version.ToString(), package.Url.ToString());
```
markus101: I don't see the URL being very useful for most notifiers, if any.

markus101: If anything they may want to a link to access the Sonarr UI, but thats not really something we can provide (multiple IPs, not always accessible, etc).

src/NzbDrone.Core/Notifications/NotificationService.cs
```diff
@@ -158,5 +169,31 @@ public void Handle(SeriesRenamedEvent message)
                 }
             }
         }
+
+        public void HandleAsync(HealthCheckCompleteEvent message)
```
markus101: The actual update check should publish an event when an update is available which can be used to trigger the notification.

src/NzbDrone.Core/Notifications/NotificationService.cs
```diff
@@ -9,21 +9,32 @@
...
+          IHandle<SeriesRenamedEvent>,
+          IHandleAsync<HealthCheckCompleteEvent>
     {
         private readonly INotificationFactory _notificationFactory;
         private readonly Logger _logger;
+        private readonly IHealthCheckService _healthCheckService;
+        private readonly ICheckUpdateService _checkUpdateService;
 
-        public NotificationService(INotificationFactory notificationFactory, Logger logger)
+        private UpdatePackage _lastUpdate;
```
markus101: This is only going to be useful for the current lifecycle of the app, if Sonarr was restarted to install an update and it failed then they would get another notification the next time it ran.

src/NzbDrone.Core/Notifications/CustomScript/CustomScript.cs
```diff
@@ -48,6 +53,14 @@ public override bool SupportsOnGrab
             }
         }
 
+        public override bool SupportsOnSystemUpdateAvailable
```
markus101: `SupportsOnSystemUpdateAvailable` makes it sound like an OS update, `SupportsOnUpdateAvailable` would make more sense.

src/NzbDrone.Core/Update/UpdateCheckService.cs
```diff
@@ -31,7 +33,13 @@ public class CheckUpdateService : ICheckUpdateService
 
         public UpdatePackage AvailableUpdate()
         {
-            return _updatePackageProvider.GetLatestUpdate(_configFileProvider.Branch, BuildInfo.Version);
+            _latestUpdate = _updatePackageProvider.GetLatestUpdate(_configFileProvider.Branch, BuildInfo.Version);
+            return _latestUpdate;
+        }
+
+        public UpdatePackage LatestUpdate()
```
markus101: This is only required because of the approach being used, I don't think its needed.

markus101:
I think there are 3 update specific notifications that a user may want to subscribe to:

1. Automatic updates are disabled, but they want to know when one is available
2. Automatic updates are enabled and an update is installed successfully
3. Automatic updates are enabled and the update failed (this would be once per version, if the reason it failed is a problem on our end users don't need to get 100 notifications)

I think we'll need a database table to track this properly with at least these columns:

- Current version (would be used to trigger a notification after a successful update)
- Version of available update
- Version of update that is going to be installed (so we can trigger failed updates)

src/NzbDrone.Core/Notifications/NotificationService.cs
halkeye: Thats kinda how couch and sickbeard worked, so I decided to go with that behavior. It was either an oversight, or designed so you got notifications again if you missed them the first time.

What would be better?

halkeye:
Okay, i'm going to need to rethink this a bit.

Is there a way to disable automatic updates in windows? My real server is linux and was shocked to see it auto-enabled in windows.

So on startup, we check to see if build.version > db.currentVersion, then publish an upgraded event
On version check, if new version != db.nextVersion, publish event, update db.nextVersion

Whats the last one for? if build.version == db.currentVersion && build.version != db.attemptedVErsion then publish error upgrading event?

halkeye:
Also, mostly as a note to myself, add a migration for t0.OnSystemUpdateAvailable (or whatever it ends up calling)

src/NzbDrone.Core/Notifications/Webhook/WebhookService.cs
```diff
@@ -66,7 +70,21 @@ public void OnGrab(Series series, RemoteEpisode episode, QualityModel quality, W
             NotifyWebhook(payload, settings);
         }
 
-        public void NotifyWebhook(WebhookPayload body, WebhookSettings settings)
+
+        public void OnSystemUpdateAvailable(WebhookSettings settings, UpdatePackage package)
+        {
+            var payload = new 
+            {
+                EventType = "Grab",
```
kayone: Shouldn't `EventType` be `UpdateAvilable`?

kayone:
I'm a bit confused, we send users notifications when a system update is available, right before we automatically update them?

maybe we should send a notification when the app is updated? or only send update available if automatic update is off?

halkeye:
I'll admit, I didn't know auto updating was a thing before I started working on this. My install is on linux, I don't think auto update is an option on there (also, since couch once had a bad update and started to automatically redownload my entire collection, i've never been fond of auto update).

> maybe we should send a notification when the app is updated? or only send update available if automatic update is off?

Agreed. I don't think there's any need to notify if auto update is enabled. I like @markus101's list

> Automatic updates are disabled, but they want to know when one is available
> Automatic updates are enabled and an update is installed successfully
> Automatic updates are enabled and the update failed (this would be once per version, if the reason it failed is a problem on our end users don't need to get 100 notifications)

kayone:
since we are admitting things, I have to admin I completely missed @markus101's comments,

I think he explained it pretty well.

markus101:
@halkeye not exactly sure where this was left off, could you advise?

halkeye:
@markus101 oh wow, I totally dropped the ball on this one (I blame a new job).

I didn't get very far, it has a new event type, but I needed to create the new db schemas to store the various states.

I don't really know how to answer where its left off.

kayone:
@halkeye haha, I do that everytime I switch jobs :D

markus101:
And then again every few months to make sure we're on our toes :smile: 

No worries @halkeye I'll try to spend some time on it and see what I can do to revive it.

halkeye:
I know one thing that sorta blocked me was coming up with how to do the migrations and connecting to the database.

I know its pretty simple, but it wasn't int he stuff I had done previously, so was looking into doing it and got distracted.

src/NzbDrone.Core/Notifications/MediaBrowser/MediaBrowser.cs
```diff
@@ -53,6 +54,16 @@ public override void OnRename(Series series)
             }
         }
 
+        public override void OnUpdateAvailable(UpdatePackage package)
+        {
+            if (Settings.Notify)
+            {
+                const string title = "Sonarr [TV] - New System Update";
```
kayone: I don't think we need the `[TV]`

Issue #298, opened 2013-07-24
Movie Support


SonarrBot:
We would all like some kind of Movie support, however it is a very complicated matter to properly implement it in Sonarr. We rather not have it at all, than have it hacked in and barely working.
This feature will need to be designed with its own architecture, hundreds if not thousands of unit/integration tests, with hundreds of man hours involved.
We would need at least 2 additional developers to even consider this feature.

Sadly this means that, despite the number of votes, this feature is pretty much the last item on our todo-list.

scutzi128:
I'm in the CP sucks camp aswell. I'vs spent countless hours tweaking the setup but it still randomly fails to work sometimes for seemingly no reason. Sonarr on the other hand is a gem and works 99% of the time.

i-c:
A "nice to have" secondary feature for this could be an option to also track "TV Movie" specials as movies using the same file. I'm not sure how practical it is, though, since thetvdb doesn't have a "this is a movie" metadata flag.

For example - Babylon 5 had a set of TV movies that have their own IMDB entries, but are also tracked by thetvdb as specials for the series.

CatmanIX:
Yeah, i've been trying to think of a way for TV movies specials to be managed properly by anything. It would definitely require a 3rd party service to map specials on tvdb, trakt, and the like with IMDB pages (and maybe trakt movie pages too). 

Sonarr already uses xem though, right? I wonder if xem could be expanded to do this as well since it already covers most of the functions necessary.

DevanWright:
I would like to add a vote to have movie support added to Sonarr. The TV handling is amazing, so no doubt you guys would be able to come up with something amazing for movies. With that said, It's completely understandable for this to not be a priority request and that implementing this feature will take countless hours.

I'm still looking forward to it though. Television and Movies all handled by Sonarr would be fantastic. Bittorrent to Sonarr to Plex is a marvelous creation.

JigSawFr:
Yes, it will be a fantastic alternative to couchpotato :+1:

tidusjar:
+1 for Movie support from me. If I can help out with the implementation in anyway please let me know.

MrsAngelD:
I would really love to see Movie support on Sonarr

rje6459:
Movie support on Sonarr would be amazing! Not a fan of couchpotato.

rcketscientist:
[...]

PR #800, opened 2015-09-14
Add movies support.

vawen:
This PR is not intended to be merged, but to share some ongoing work for #298.
This branch has functional movie support, parse is very simple and uses same qualities as Series.
Search and RSS supports is working, only KAT and RARBG indexers have the searchs implemented for movies.

TODO: MoviesEditor (bulk), Wanted tab (Movies), TestCases (a lot of them!), Decisions (some decisions are common and those ones are in use).

Things I don't like how are done:
- History, Blacklist, PendingRelease: I don't know if it would be better to have separate classes for Movies and Series.
- RSS: At this moment they are together, so you just get 1 RSS from the indexer and then try to parse as series and as movies. I don't know if separate it would be better.
- Manual import is disabled.


pmwoodward3:
Really excited to see this, if there's anything I can do to help just walk me through it.

onedr0p:
+1 Sonarr handling TV is really stable, would like Movie support.

feld:
+1 as well, this would be a fantastic addition

Beatusvir:
+1 !! nice nice

PR #806, opened 2015-09-21
Mono/.NET45

GuillaumeLebeau:
Migrate all projects to .NET4.5 with Xamarin Studio.
Also updated NuGet packages and some node packages.


markus101:
This is the big thing that we're planning for v3, we haven't yet started on it so it won't be able to be merged until we do.

PR #817, opened 2015-09-26
New: Added support for newznab indexers using tvdbid for searching.

Taloth:
This code assumes that nn+ will adopt our supportedParams proposal.

Also updated the ParserService to use tvdbid in ReleaseInfo.
I didn't remove tvrageid, there will be plenty of services still using it, especially for old stuff.

Ended up being a bit bigger than I intended, but when you add ReleaseInfo.TvdbId you get a lot of tests where it 'should' be added.

I didn't want to drop TorznabSettings due to the validator and the fact that we may add params in the future, but I decided to base it on NewznabSettings.


src/NzbDrone.Core/Indexers/Newznab/NewznabRequestGenerator.cs
```diff
@@ -9,21 +9,79 @@ namespace NzbDrone.Core.Indexers.Newznab
...
+        }
+
+        private bool SupportsTvdbSearch
+        {
+            get
+            {
+                var capabilities = _capabilitiesProvider.GetCapabilities(Settings);
+
+                return capabilities.SupportedTvSearchParameters != null
+                    && capabilities.SupportedTvSearchParameters.Contains("tvdbid")
```
markus101: Trailing `&&` instead? (I think thats what we typically do).

src/NzbDrone.Core/Indexers/Torznab/TorznabRssParser.cs
```diff
@@ -99,6 +100,19 @@ protected override string GetDownloadUrl(XElement item)
             return url;
         }
 
+        protected virtual Int32 GetTvdbId(XElement item)
```
markus101: Saw some other changes to `int` from `Int32` so here too :smile:

markus101:
Didn't notice anything blocking, but I'm curious what the plan is if supportedParams isn't added (on all sites) will we use title searches on the sites that don't add it?

Taloth:
We would have no way to identify em without a manual blacklist.

src/NzbDrone.Core/Indexers/Torznab/TorznabRssParser.cs
Taloth: Not in this file, I tend to keep a file consistent until we can get Roslyn to replace em all.

src/NzbDrone.Core/Indexers/Newznab/NewznabRequestGenerator.cs
Taloth: It's what we typically do, but in this case leading && seems clearer.
(Btw this was copied from TorznabRequestGenerator, not new code).

markus101:
Nothing jumped out at me, good call on the new classes, cleans thing sup nicely.

PR #844, opened 2015-10-07
Improvements to History Specification

### LINKED ISSUE #625: Improve RSS Sync History Specification
Right now we only check if the download client is SAB (old logic from before nzbget support I think), otherwise we fall back to a naive BestQualityInHistory check that doesn't take failed downloads into account (it will count them in the check).

If the download client is one that we can't track (blackhole/pneumatic) we should rely on the grabbed state (assuming a failed event doesn't exist), otherwise rely on the download client's queue/history. I don't think we need to go nuclear on blocking downloads because we previously tried the same quality as it would block us from getting a repost hours/days/weeks/etc later.

### LINKED ISSUE #837: Blackhole doesn't treat previous grab as meeting cutoff
Normally Sonarr relies on the queue to determine if the cutoff has been met by an item in the queue, but with blackhole it doesn't have a queue to track, in this scenario Sonarr should use its history to see if the history if the cutoff has been met, until it knows otherwise (failed or completed).

markus101:
Prevents grabbing if recent grab is less than an hour old and fixes a bug with invalid season passes and SAB.

Closes #837
Fixes #625

### DIFF
--- a/src/NzbDrone.Core.Test/DecisionEngineTests/HistorySpecificationFixture.cs
+++ b/src/NzbDrone.Core.Test/DecisionEngineTests/HistorySpecificationFixture.cs
@@ -1,11 +1,11 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
 using FizzWare.NBuilder;
 using FluentAssertions;
 using Moq;
 using NUnit.Framework;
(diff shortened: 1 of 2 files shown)


Taloth:
Gave it a quick glance, looks good, except that you don't need the `_downloadClientProvider` anymore.
Also `String.Empty` -> `string.Empty` please.

PR #886, opened 2015-10-21
Added verified file transfer mode that doesn't revert to copy.

Taloth:
@markus101 

The original VerificationModes previously implemented:
None: Blindly copy/move without verifying filesizes.
VerifyOnly: Copy/move directly to target path, but check filesize afterwards, produces warning if incorrect. (This was default before the master merge)
Transactional: Uses hardlinks to perform a source -> source.backup~ -> target.partial~ -> target transactional transfer. Reverts to copy -> target.partial~ -> target if hardlink fails.

I added TryTransactional and made it default on mono. In this mode it will try to hardlink, otherwise it will revert to the same behavior as VerifyOnly instead of doing Copy.
This means that if someone is affected by the CIFS+mono bug AND does not support hardlinks, it will still potentially fail and lead to data loss, but in pretty much all other cases it will remain fast even in the absence of hardlink support.

TODO: Consider adding ConfigService option for VerificationMode (will need to be set in bootstrap coz TranferService is in Common).
We could default to TryTransactional, but switch to Transactional the first time we run into a partial. Which would make it adaptive, but less predictable.

I'm okay with it as is, but it means Transactional isn't used at all. But adding the config option neatly might be tricky. Lemme know what you think.

### DIFF
--- a/src/NzbDrone.Common/Disk/DiskTransferService.cs
+++ b/src/NzbDrone.Common/Disk/DiskTransferService.cs
@@ -21,6 +21,7 @@ namespace NzbDrone.Common.Disk
     {
         None,
         VerifyOnly,
+        TryTransactional,
         Transactional
     }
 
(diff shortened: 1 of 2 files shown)


src/NzbDrone.Common/Disk/DiskTransferService.cs
```diff
@@ -423,16 +394,63 @@ private bool TryMoveFile(string sourcePath, string targetPath)
...
+        private void TryMoveFileVerified(string sourcePath, string targetPath, long originalSize)
+        {
+            try
+            {
+                _diskProvider.MoveFile(sourcePath, targetPath);
+
+                var targetSize = _diskProvider.GetFileSize(targetPath);
+                if (targetSize != originalSize)
+                {
+                    throw new IOException(string.Format("File copy incomplete, data loss may have occured. [{0}] was {1} bytes long instead of the expected {2}.", targetPath, targetSize, originalSize));
```
markus101: move instead of copy?

markus101:
Config + common isn't something we've had to do previously, but I see why we would here. I think TryTransactional makes sense, we do what we can without impacting it too much.

Taloth:
Merged in by https://github.com/Sonarr/Sonarr/commit/cc72699b8a384cc95cc30c5188575954dee2d606

PR #901, opened 2015-10-29
UI: Update jQuery to 1.11.3

brgaulin:
Just a simple version bump to jQuery. Fixes an IOS bug in 8.2 and 8.3

### DIFF
--- a/src/UI/JsLibraries/jquery.js
+++ b/src/UI/JsLibraries/jquery.js
@@ -1,5 +1,5 @@
 /*!
- * jQuery JavaScript Library v1.11.2
+ * jQuery JavaScript Library v1.11.3
  * http://jquery.com/
  *
  * Includes Sizzle.js
@@ -9,7 +9,7 @@
  * Released under the MIT license
  * http://jquery.org/license
  *
- * Date: 2014-12-17T15:27Z
+ * Date: 2015-04-28T16:19Z
  */
 
 (function( global, factory ) {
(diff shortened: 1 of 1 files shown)


brgaulin:
Hello Sonarr Team!

As part of Hacktober, I was poking around repos of OS software I have seen before, and noticed your UI might want a little love =) I am a javascript developer in the US and would love to help out(much more than this little change). I was wondering if there was a UI punchlist?

Things I see so far:
 - Move from RequireJS to Webpack/CommonJS completely
 - Upgrade all the vendor libraries, integrate bower(was wondering if there was a reason for the old versions, or just they were new at the time?)
 - Minify assets for deployment, and use the already uglified vendor files
 - Probably many more.

But I have some bandwidth and over the next few months I would love to help out in any way =) Let me know

markus101:
Thanks for reaching out.

> Move from RequireJS to Webpack/CommonJS completely

Is there a particular part where this wasn't done? We definitely intended to overhaul everything when it was converted and I haven't seen any cases where it was missed that I recall.

> Upgrade all the vendor libraries, integrate bower(was wondering if there was a reason for the old versions, or just they were new at the time?)

We've updated a number of the libraries in some other UI changes we've been working on (much larger than dependency updates). We've looked at bower before (and used it professionally), but we haven't really considered adding it to Sonarr, as we tend to version lock things to make sure we're not breaking something, though automatically fetching libraries is certainly helpful.

> Minify assets for deployment, and use the already uglified vendor files

This is already done as part of the build process.

brgaulin:
Sounds good! As far as the minify, when o start Sonarr, at least the vendor files are all unminified. So there may be a bug there. 



--Brendan Gaulin

(508) 728-3582

markus101:
At the moment we don't minify it, I was mistaken before, I'll let @kayone comment, but we may want to minify vendor.js, but we previously didn't see massive transfer gains since its gzipped from the server.

markus101:
Sorry for the delay, if you're interested in helping out on the UI side send us an email to hello at sonarr.tv and we can give you some ideas on where we're going with things.

PR #915, opened 2015-11-06
Fix z-index on scroll to top button

uzegonemad:
Near the bottom of the page, there's a small area where the scroll to top button can no longer hover. This is because part of the footer has a "higher" z-index.

Confirmed in Chrome and Firefox. IE doesn't appear to be affected. Here's some screenshots with an added black bar to explain the overlay issue.

Before:
screen shot 2015-11-05 at 7 58 19 pm

After:
screen shot 2015-11-05 at 7 58 26 pm

Just an FYI, I tried to sign the CLA but was met with a server error.

### DIFF
--- a/src/UI/Content/theme.less
+++ b/src/UI/Content/theme.less
@@ -81,6 +81,7 @@
 
   .opacity   (0.2);
   position  : fixed;
+  z-index   : 9999;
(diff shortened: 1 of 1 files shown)


markus101:
Thanks! Sorry for the delay.

PR #924, opened 2015-11-09
Run gulp using npm

Dahlgren:
Simplifies usage of gulp and makes sure everyone is using same version

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -18,7 +18,9 @@ Setup guides, FAQ, the more information we have on the wiki the better.
 1.  Fork Sonarr
 2.  Clone (develop branch) *you may need pull in submodules separately if you client doesn't clone them automatically (CurlSharp)*
 3.  Run `npm install`
-4.  Run `gulp watch` - Used to compile the UI components and copy them (leave this window open)
+4.  Run `npm start` - Used to compile the UI components and copy them.
+    Leave this window open.
+    If you have gulp globally installed you can use `gulp watch` instead
 5.  Compile in Visual Studio
 
 ### Contributing Code ###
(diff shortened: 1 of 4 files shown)


build.sh
```diff
@@ -106,7 +106,7 @@ RunGulp()
     echo "##teamcity[progressFinish 'npm install']"
 
     echo "##teamcity[progressStart 'Running Gulp']"
-    CheckExitCode gulp build
+    CheckExitCode npm run gulp build
```
kayone: I really like this, since it means we won't need to install gulp on the build servers.

CONTRIBUTING.md
```diff
@@ -18,7 +18,7 @@ Setup guides, FAQ, the more information we have on the wiki the better.
 1.  Fork Sonarr
 2.  Clone (develop branch) *you may need pull in submodules separately if you client doesn't clone them automatically (CurlSharp)*
 3.  Run `npm install`
-4.  Run `gulp watch` - Used to compile the UI components and copy them (leave this window open)
+4.  Run `npm run gulp watch` - Used to compile the UI components and copy them (leave this window open)
```
kayone: tbh, I think most people that would be contributing might be confused by this, specially if they are already familiar with gulp, since most people would know gulp as using it by calling the globally installed version.

kayone:
I'm really sorry it took this long, I would be more than happy to merge this in, however I think we should keep the _global method_ in the `.md` files, we could however add a _if you don't want to install gulp globally you could run this command_. so an alternative not a replacement.

Dahlgren:
I'll try to take a look at it later today

Dahlgren:
Simplified it a bit so that `npm start` will run `gulp watch` directly since that will be the most common use case for developers.

`npm run build` will trigger a `gulp build`. By having a more generic name additional build steps could be added in the future or gulp could be changed to something else.

kayone:
Merged into develop, Thanks!

PR #928, opened 2015-11-11
Fixed tests for Mono

Dahlgren:


markus101:
The tests that were changed to be windows only, aren't windows only tests, they should apply to all platforms.

Dahlgren:
The ProcessProvider tests have problems while running on linux/osx due to the process name being mono, https://github.com/Sonarr/Sonarr/blob/45cdb98c582164eb7b34dcd70c4f491c81a7a92b/src/NzbDrone.Common/Processes/ProcessProvider.cs#L111

src/NzbDrone.Core.Test/TvTests/ShouldRefreshSeriesFixture.cs
```diff
@@ -66,6 +71,8 @@ private void GivenRecentlyAired()
         [Test]
         public void should_return_true_if_running_series_last_refreshed_more_than_6_hours_ago()
         {
+            GivenSeriesLastRefreshedHalfADayAgo();
```
Taloth: already set https://github.com/Dahlgren/Sonarr/blob/mono-tests/src/NzbDrone.Core.Test/TvTests/ShouldRefreshSeriesFixture.cs#L21

Dahlgren: Yes, but it fails for some reason on Mono

src/NzbDrone.Common.Test/ProcessProviderTests.cs
```diff
@@ -54,8 +54,9 @@ public void GetProcessById_should_return_null_for_invalid_process(int processId)
         }
 
         [Test]
-        public void Should_be_able_to_start_process()
+        public void Should_be_able_to_start_process_windows()
```
markus101: This one fails if run in docker, but works fine in Ubuntu on a VM, we don't want to make it Windows only, but we currently ignore it on our docker test agents.

src/NzbDrone.Common.Test/ProcessProviderTests.cs
```diff
@@ -69,8 +70,9 @@ public void Should_be_able_to_start_process()
 
 
         [Test]
-        public void kill_all_should_kill_all_process_with_name()
+        public void kill_all_should_kill_all_process_with_name_windows()
         {
+            WindowsOnly();
```
markus101: Same issue for docker, works for Ubuntu though, we'll ignore as required.

src/NzbDrone.Common.Test/WebClientTests.cs
```diff
@@ -23,7 +23,7 @@ public void DownloadString_should_be_able_to_dowload_text_file()
         [TestCase("http://")]
         public void DownloadString_should_throw_on_error(string url)
         {
-            Assert.Throws<ArgumentException>(() => Subject.DownloadString(url));
+            Assert.Catch(() => Subject.DownloadString(url));
```
markus101: Does this fail? We want to assert the correct exception is thrown, not catch all exceptions.

src/NzbDrone.Core.Test/IndexerSearchTests/SearchDefinitionFixture.cs
```diff
@@ -13,7 +13,7 @@ public class SearchDefinitionFixture : CoreTest<SingleEpisodeSearchCriteria>
         [TestCase("Hawaii Five-0", Result = "Hawaii+Five+0")]
         [TestCase("Franklin & Bash", Result = "Franklin+and+Bash")]
         [TestCase("Chicago P.D.", Result = "Chicago+PD")]
-        [TestCase("Kourtney And Khlo� Take The Hamptons", Result = "Kourtney+And+Khloe+Take+The+Hamptons")]
+        [TestCase("Kourtney And Khlo\u00E9 Take The Hamptons", Result = "Kourtney+And+Khloe+Take+The+Hamptons")]
```
markus101: Where does this fail? Currently it runs fine on our test platforms.

src/NzbDrone.Core.Test/TvTests/ShouldRefreshSeriesFixture.cs
markus101: This test passed as-is for our main builds, but fails for the PR builds (same tests are run, so its rather bizarre), can we drop the initial set of the last refreshed time/set it to UtcNow instead, that will make this method make more sense and I don't see a reason to nit at it since we do it for the other tests.

src/NzbDrone.Update.Test/ProgramFixture.cs
```diff
@@ -32,8 +32,9 @@ public void should_throw_if_first_arg_isnt_an_int(string arg1, string arg2)
         }
 
         [Test]
-        public void should_call_update_with_corret_path()
+        public void should_call_update_with_correct_path_windows()
```
markus101: Is this failing? Passing on our build server at the moment, updater should be called on all OSes to make sure it launches.

src/NzbDrone.Core.Test/IndexerSearchTests/SearchDefinitionFixture.cs
Dahlgren: Depending on which IDE is used. Xamarin Studio and MonoDevelop freaks out due to the unicode character. Using the escaped format works fine for both OSX and Windows development

src/NzbDrone.Common.Test/WebClientTests.cs
Dahlgren: I agree. On OSX, at least, `System.Net.WebException` is thrown for `http://`

src/NzbDrone.Update.Test/ProgramFixture.cs
Dahlgren: The test fails in `UpdateApp.cs` due to Start being called with two arguments. When parsing args there is a check if the OS is not Windows (https://github.com/Sonarr/Sonarr/blob/develop/src/NzbDrone.Update/UpdateApp.cs#L74) and it requires either 1 or at least 3 arguments. Otherwise an `IndexOutOfRangeException` is thrown.

src/NzbDrone.Common.Test/ProcessProviderTests.cs
Dahlgren: So the WindowsOnly() should be removed for these for now then?

Dahlgren:
`WebClientTests` `DownloadString_should_throw_on_error`fails on OSX due to `System.Net.WebException` is thrown for http:// which seems a bit odd

src/NzbDrone.Update.Test/ProgramFixture.cs
markus101: We should update the test, the test wasn't updated when the code was updated to support additional parameters and allows Windows to only pass a single argument for backwards compatibility.

`const string ProcessPath = @"C:\NzbDrone\nzbdrone.exe";` should be `var processPath = @"C:\NzbDrone\nzbdrone.exe".AsOsAgnostic();` and pass in `processPath` as the 3rd argument (executing assembly).

Dahlgren: Worked great with agnostic path as expected result too, thanks!

Dahlgren:
Anything else that needs to be fixed?

markus101:
@Dahlgren just needs to get squashed and rebased, please. Thanks for following up.

Dahlgren:
Done!

PR #941, opened 2015-11-16
Give calendar legend ul max width of 100%. Fixes #922

### LINKED ISSUE #922: Calendar looks too small on iOS webapp
On iPhone 5 I created a Sonarr web app on my home screen. When I open it it looks good and this is as far as it will zoom out.

I click on the calendar and it looks too small and zoomed out. See all the extra space on the right.

uzegonemad:
`.legend-labels` has set widths of 500px and 400px depending on the breakpoint. This adds a max-width of 100% so that it won't be larger than the container on screens smaller than 400px wide.

(On a side, completely unrelated note, I majorly screwed up some rebasing earlier... my git skills could be better.)

### DIFF
--- a/src/UI/Calendar/calendar.less
+++ b/src/UI/Calendar/calendar.less
@@ -212,7 +212,8 @@
   }
 
   .legend-labels {
-    width : 500px;
+    max-width : 100%;
+    width     : 500px;
 
     @media (max-width: @screen-xs-min) {
       width : 400px;


src/UI/Calendar/calendar.less
```diff
@@ -212,6 +212,7 @@
   }
 
   .legend-labels {
+    max-width: 100%;
```
markus101: Align properties on `:` and a space after the property name please.

uzegonemad:
I fixed the alignment and rebased! Should be good to go.

markus101:
Thanks!

PR #968, opened 2015-11-30
Release scoring

markus101:
@Taloth @kayone 

Started on this a few months ago and finally got a chance to finish it off. One thing I'd like to do is move the scoring elsewhere so we can track the score on the decision and use it for ordering.

### DIFF
--- a/src/NzbDrone.Core/DecisionEngine/DownloadDecisionComparer.cs
+++ b/src/NzbDrone.Core/DecisionEngine/DownloadDecisionComparer.cs
@@ -1,0 +1,232 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using NzbDrone.Core.Indexers;
+using NzbDrone.Core.Parser.Model;
+using NzbDrone.Core.Profiles.Delay;
+
+namespace NzbDrone.Core.DecisionEngine
+{
+    public class DownloadDecisionComparer : IComparer<DownloadDecision>
+    {
+        private readonly IDelayProfileService _delayProfileService;
+        public delegate int CompareDelegate(DownloadDecision x, DownloadDecision y);
+
+        public DownloadDecisionComparer(IDelayProfileService delayProfileService)
+        {
+            _delayProfileService = delayProfileService;
+        }
+
+        public int Compare(DownloadDecision x, DownloadDecision y)
+        {
+            var comparers = new List<CompareDelegate>
+            {
+                CompareQuality,
+                CompareProtocol,
+                CompareEpisodeCount,
+                CompareEpisodeNumber,
+                ComparePeers,
+                CompareAge,
+                CompareSize
+            };
+
+            return comparers.Select(comparer => comparer(x, y)).FirstOrDefault(result => result != 0);
+        }
+
+        private int CompareQuality(DownloadDecision x, DownloadDecision y)
+        {
+            var qualityX = x.RemoteEpisode.ParsedEpisodeInfo.Quality;
+            var qualityY = y.RemoteEpisode.ParsedEpisodeInfo.Quality;
+            var indexX = x.RemoteEpisode.Series.Profile.Value.Items.FindIndex(v => v.Quality == qualityX.Quality);
+            var indexY = y.RemoteEpisode.Series.Profile.Value.Items.FindIndex(v => v.Quality == qualityY.Quality);
+
+            if (indexX > indexY)
+            {
+                return 1;
+            }
+
+            if (indexX < indexY)
+            {
+                return -1;
+            }
+
+            if (qualityX.Revision.Real > qualityY.Revision.Real)
+            {
+                return 1;
+            }
+
+            if (qualityX.Revision.Real < qualityY.Revision.Real)
+            {
+                return -1;
+            }
+
+            if (qualityX.Revision.Version > qualityY.Revision.Version)
+            {
+                return 1;
+            }
+
+            if (qualityX.Revision.Version < qualityY.Revision.Version)
+            {
+                return -1;
+            }
+
+            return 0;
+        }
+
+        private int CompareProtocol(DownloadDecision x, DownloadDecision y)
+        {
+            var delayProfileX = _delayProfileService.BestForTags(x.RemoteEpisode.Series.Tags);
+            var delayProfileY = _delayProfileService.BestForTags(y.RemoteEpisode.Series.Tags);
+            var downloadProtocolX = x.RemoteEpisode.Release.DownloadProtocol;
+            var downloadProtocolY = y.RemoteEpisode.Release.DownloadProtocol;
+            var preferredX = downloadProtocolX == delayProfileX.PreferredProtocol;
+            var preferredY = downloadProtocolY == delayProfileY.PreferredProtocol;
+
+            if (preferredX && !preferredY)
+            {
+                return 1;
+            }
+
+            if (!preferredX && preferredY)
+            {
+                return -1;
+            }
+
+            return 0;
+        }
+
+        private int CompareEpisodeCount(DownloadDecision x, DownloadDecision y)
+        {
+            // Prefer season packs, otherwise prefer less episodes
+
+            var fullSeasonX = x.RemoteEpisode.ParsedEpisodeInfo.FullSeason;
+            var fullSeasonY = y.RemoteEpisode.ParsedEpisodeInfo.FullSeason;
+            var episodesX = x.RemoteEpisode.Episodes.Count;
+            var episodesY = y.RemoteEpisode.Episodes.Count;
+
+            if (fullSeasonX && !fullSeasonY)
+            {
+                return 1;
+            }
+
+            if (!fullSeasonX && fullSeasonY)
+            {
+                return -1;
+            }
+
+            if (episodesX < episodesY)
+            {
+                return 1;
+            }
+
+            if (episodesX > episodesY)
+            {
+                return -1;
+            }
+
+            return 0;
+        }
+
+        private int CompareEpisodeNumber(DownloadDecision x, DownloadDecision y)
+        {
+            // Prefer lower episode numbers
+
+            var episodeNumberX = x.RemoteEpisode.Episodes.Select(e => e.EpisodeNumber).MinOrDefault();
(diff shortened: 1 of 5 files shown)


Dahlgren:
Would be nice to be able to boost the score for indexers/trackers and keywords (such as release group) like in CouchPotato

markus101:
@Taloth Now uses a comparer instead of scoring, making it more concrete.

Indexer prioritization/preferred words should still be possible with direct comparisons.

src/NzbDrone.Core/DecisionEngine/DownloadDecisionComparer.cs
```diff
@@ -0,0 +1,129 @@
...
+                return downloadProtocol == delayProfile.PreferredProtocol;
+            });
+
+            return result;
+        }
+
+        private int CompareEpisodeCount(DownloadDecision x, DownloadDecision y)
+        {
+            return Compare(Compare(x.RemoteEpisode, y.RemoteEpisode, remoteEpisode => remoteEpisode.ParsedEpisodeInfo.FullSeason),
+                           CompareReverse(x.RemoteEpisode, y.RemoteEpisode, remoteEpisode => remoteEpisode.Episodes.Count));
```
Taloth: In case of a pilot double, this would mean it prefers a single episode over a double. The same rules as season pack applies, it has to try the double first, coz otherwise it may succeed at E01 but fail at E02 and never get it again coz the double is rejected for not being an upgrade foe E01.

Taloth:
Okay, lets see about our comparison rules:

Primary rules: (these are always to be honored, no fuzzy logic here, except for rejections)
- CompareQuality
- CompareProtocol

Season pack logic:
- CompareEpisodeCount

Then:
- CompareEpisodeNumber (Not really needed, mostly cosmetic.)

Release Scoring:
- ComparePeers
- CompareAge
- CompareSize

The last part is what ought to be fuzzy, I thought that was the goal of this whole exercise.
Especially 'size' isn't very useful and atm prevents our 'higher' peers rule from being effective.

I briefly mentioned somewhere to introduce a 'preferred' Size per quality. (so: min, preferred, max)
That way we can calculate how far off a release is compared to the preference. Which is far more valuable than using 'smaller' or 'larger' releases.
I would also multiply usenet sizes by 0.9 to compensate for par files.

the SizeScore should be calculated by:

```
\\ negative, coz deviations are a penalty, value itself is arbitrary. A 10% deviation = -1 score.
sizeScoreFactor = -10

preferredReleaseSize = preferredSize[quality] * runtime * episodeCount

              Math.Abs(preferredReleaseSize - (size * usenetCompensationFactor))
sizeScore = ----------------------------------------------------------------------- * sizeScoreFactor
                                       preferredReleaseSize
```

AgeScore only applies to usenet, and should be an as small as possible penalty. ```ageScore = 1 - age / 365.0``` should be fine. If we weren't using fuzzy logic then I would've put it in 3 categories: 1) a hour, 2) < 24h 3) more than 24h.

SeederScore is the most important one, I would use a logarithmic scale for this:

```
seederScore = seeders == 0 ? 0 : Math.Log10(seeders)
```

Which means if a 400MB release has 1 seed and a 440MB release 10 seeds, with a preferred size of 400MB:
score 1: -10 * abs(400 - 400) / 400 + Log(1) = 0 + 0 = 0
score 2: -10 * abs(400 - 440) / 400 + Log(10) = -1 + 1 = 0
So they are considered equal.
You can play around with the sizeScoreFactor to handle it better.

Age (penalty) vs seeds (bonus) is a bit tricky, coz it would mean torrents are always preferred over usenet. So we should balance that out so that usenet is preferred over 1 seed. Hence the offset of 1 to the ageScore.

Downside of all this is that balancing seeds vs age means that reasonably seeded torrents will often be preferred over usenet. I'm not sure how to compensate for that, other than increasing the offset of 1 to 2 or 3.

markus101:
Age vs seeders doesn't matter, that's handled by the preferred protocol. Since those are constant the preferred protocol always wins.

Regarding size, do we really care? Should we care? If its within the user's size limits I'm not sure it really matters, we put emphasis on smaller releases right now, but does it make much of a difference? If we don't know if the release has par2 files its a crap shoot, we ignore that and it can punish users in some cases. I still like the idea of preferring smaller releases, but having it below age and seeders assuming they are both fuzzy and at some point don't care, 10000 vs 1000 seeds is a bit irrelevant as is 10 vs 11 days and 45 minutes vs 90 minutes. For newznab do we have a way to get the files via the API or even to know if par2 files are included? That would be a very handy piece to have, just a Boolean, `hasPar2Files`, it would bring a lot of power to Sonarr, both in terms of preferring releases and staying within size limits, even if we're assuming that it's a certain percentage.

I think fuzzy for seeders/age/size is definitely what we want, the recent changes were more about how we do it, less about complete logic as there was no feedback on the purely fuzzy solution other than its not good for quality/protocol.

Taloth:
Age vs seeders, I forgot we don't have a 'doesn't matter' preferred protocol. So indeed, that point is irrelevant.
That also removes the torrent size vs usenet size issue, so the par2 compensation factor is less relevant.

Whether you like smaller releases vs bigger ones is entirely down to preference. min/max is a hard restriction.

hasPar2Files would be nice, but then it's better to just ask for the non-par total size. We could suggest it to be added to nn+/nzedb.

But yes, it comes down to age+size for usenet and seeds+size for torrents.
You could try categorize age and seeds, but the boundaries of the categories would be fairly arbitrary and you get the same issues as you had with size.
Is something of the perfect size but 3days old, a worse choice than a 3gb oversized hdtv at 1h?

The logarithm for seeds mostly takes care of 10000 vs 1000 seeds. coz on a logarithmic scale those two don't differ that much (4 vs 3 in log10).

markus101:
> ageScore = 1 - age / 365.0 should be fine. 

That felt too broad.

> If we weren't using fuzzy logic then I would've put it in 3 categories: 1) a hour, 2) < 24h 3) more than 24h.

We can be fuzzy and give it a score, so I did, let me know what you think.

> `seederScore = seeders == 0 ? 0 : Math.Log10(seeders)`

This feels too broad, since it still gives weight to more seeds, I went with `seeders.HasValue ? Math.Round(Math.Log10(seeders.Value)) : 0;`, which means 999 and 1000 seeds are the same as well as 10 and 50. This seems to be a good balance of favouring more seeds without caring too much about it.

src/NzbDrone.Core/DecisionEngine/DownloadDecisionComparer.cs
```diff
@@ -0,0 +1,151 @@
...
+            return leftValue.CompareTo(rightValue);
+        }
+
+        private int CompareReverse<TSubject, TValue>(TSubject left, TSubject right, Func<TSubject, TValue> funcValue)
+            where TValue : IComparable<TValue>
+        {
+            return Compare(left, right, funcValue)*-1;
+        }
+
+        private int Compare(params int[] comparers)
```
Taloth: I would call this 'CompareAll', to avoid the Compare(Compare(... you see below.

src/NzbDrone.Core/DecisionEngine/DownloadDecisionComparer.cs
```diff
@@ -0,0 +1,151 @@
...
+                CompareEpisodeNumber,
+                ComparePeers,
+                CompareAge,
+                CompareSize
+            };
+
+            return comparers.Select(comparer => comparer(x, y)).FirstOrDefault(result => result != 0);
+        }
+
+        private int Compare<TSubject, TValue>(TSubject left, TSubject right, Func<TSubject, TValue> funcValue)
```
Taloth: Idea to rename this to 'CompareBy' kinda like the DistinctBy we created so it's clear we're selecting a value on both left and right and comparing the result?

src/NzbDrone.Core/DecisionEngine/DownloadDecisionComparer.cs
```diff
@@ -0,0 +1,151 @@
...
+
+        public int Compare(DownloadDecision x, DownloadDecision y)
+        {
+            var comparers = new List<CompareDelegate>
+            {
+                CompareQuality,
+                CompareProtocol,
+                CompareEpisodeCount,
+                CompareEpisodeNumber,
+                ComparePeers,
```
Taloth: How about ComparePeersIfTorrent and CompareAgeIfUsenet?

Taloth:
I'm missing a couple of red tests, that check whether the two release are 'too close' to order by peers, age, size.

src/NzbDrone.Core/DecisionEngine/DownloadDecisionComparer.cs
```diff
@@ -0,0 +1,151 @@
...
+            {
+                return 0;
+            }
+
+            return Compare(
+                Compare(x.RemoteEpisode, y.RemoteEpisode, remoteEpisode =>
+                {
+                    var seeders = TorrentInfo.GetSeeders(remoteEpisode.Release);
+
+                    return seeders.HasValue ? Math.Round(Math.Log10(seeders.Value)) : 0;
```
Taloth: Add a test for 0 seeds. and watch it blow up. (Log10(0) = negInfinity)

src/NzbDrone.Core/DecisionEngine/DownloadDecisionComparer.cs
```diff
@@ -0,0 +1,151 @@
...
+                {
+                    return 100;
+                }
+
+                if (age <= 7)
+                {
+                    return 10;
+                }
+
+                return 1 - age/365;
```
Taloth: This means stuff older than 7 days rarely gets compared by size. Personally I would throw them all one one pile.

src/NzbDrone.Core/DecisionEngine/DownloadDecisionComparer.cs
markus101: CompareBy feels a bit awkward to me, but I don't have a strong objection to it and https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.comparisons/compare-by.html, so done.

markus101:
Added a couple tests to ensure we're doing a stable sort when peers + size or age + size are too similar to compare and we take the first release.

Taloth:
squash and ship it

PR #973, opened 2015-12-01
Store and forward releases when download client is unavailable

### LINKED ISSUE #949: Store and Forward if Download client is offline
I thought this was already filed, but I can't find it.

When Sonarr is able to download the NZB/Torrent, but fails to send it to the download client we should consider storing it, like we would do for a delayed grab so we don't miss out on that release. Most notably this would be great with the RSS back filling that we now do, if Sonarr can't connect tot he download client because Sonarr started first those results are lost and the user misses things they have been expecting.

markus101:
@kayone @Taloth 

Currently only tested with SAB (haven't looked what work other clients will require).
Closes #949


src/UI/Content/icons.less
```diff
@@ -151,6 +151,11 @@
   .fa-icon-content(@fa-var-clock-o);
 }
 
+.icon-sonarr-client-unavailable {
+  .fa-icon-content(@fa-var-clock-o);
```
markus101: Needs a better icon, suggestions?

src/NzbDrone.Api/Queue/QueueResource.cs
```diff
@@ -17,7 +19,10 @@ public class QueueResource : RestResource
         public decimal Size { get; set; }
         public string Title { get; set; }
         public decimal Sizeleft { get; set; }
+
+        [JsonConverter(typeof(TimeSpanConverter))]
         public TimeSpan? Timeleft { get; set; }
```
kayone: I thought we could register convertors to a specific type that way we wouldn't have to annotate every usage.

markus101: Yeah we can, I wanted to look at the usages first, in some cases having milliseconds might be useful (run time of a task perhaps).

kayone: any reason not to include milliseconds here?

markus101: It looks pretty bad in the UI, but this is pretty ugly, I'm going to see what I can do to improve the formatting of this in the UI, then this isn't required at all.

kayone: yeah, UI seems like the right place to do it.

markus101: moment.js FTW:

`As of 2.1.0, moment supports parsing ASP.NET style time spans.` We'll send it down and let the UI figure it out.

thezoggy:
can the user control how long before it stops retrying / does it work if sonarr is reloaded (because of update or manually) -- this feature also makes me think of another feature, the user have the ability to click an item in the history to resend?

Taloth:
DownloadClientUnavailableException  needs to be thrown in other download clients too.

Also, how do you deal with the situation where there is both a usenet and torrent client, and releases available for both... but one client is unavailable?

You must also put any other release into 'pending' to avoid releases getting lost.

May I suggest the following:
- maintain 3 local lists: newDelayed, newStoredUsenet and newStoredTorrent.
- if TemporarilyRejected, put in newDelayed list and continue.
- if usenet-dl and newStoredUsenet already contains anything, put in newStoredUsenet and continue. (prevents hitting the same download client multiple times if failing)
- if torrent-dl and newStoredTorrent already contains anything, put in newStoredTorrent and continue.
- try download with specified client.
- if failed due to unavailable, put in either newStoredUsenet or newStoredTorrent and continue.
After entire download loop finishes do some housekeeping:
- iterate over newDelayed, newStoredUsenet, newStoredTorrent. Removing any report for episodes that's already in 'grabbed'. the remaining should be added to _pendingReleaseService.

This ensures that if one client fails the other protocol could be used instead. the housekeeping at the end is needed to ensure we only pend releases that weren't grabbed later on.

Alternatively, we could select all releases that weren't grabbed into _pendingReleaseService too as 'Fallback', which would enable us to download alternative releases without backlog searching on failed downloads.
But that's another feature that can be added later on.

markus101:
> Also, how do you deal with the situation where there is both a usenet and torrent client, and releases available for both... but one client is unavailable?

Not sure this is significantly different from the case where there is a release for the unpreferred protocol has an available release and the preferred protocol doesn't, if there is a delay on the unpreferred then it will wait, otherwise it will be grabbed immediately.

> try download with specified client.

This would retry the grab once again and if it fails we'd store everything, but if it succeeds we'd try processing until it fails again? A best effort system to attempt again if the connection is restored?

Taloth:
> Not sure this is significantly different...

It is significant, coz if the most preferred release fails due to downloadclientunavailable, you [put it in the pending list](https://github.com/Sonarr/Sonarr/pull/973/files#diff-10af590b68a5c0d11a8ee23801cd6f9cR82), and thus any other release, for any client, [gets ignored](https://github.com/Sonarr/Sonarr/pull/973/files#diff-10af590b68a5c0d11a8ee23801cd6f9cR64).

> A best effort system to attempt again if the connection is restored?

Isn't that what this whole feature is about? To retry on every RssSync?

markus101:
> It is significant, coz if the most preferred release fails due to downloadclientunavailable, you put it in the pending list, and thus any other release, for any client, gets ignored.

Ahh yes, unless we apply the changes you suggested (which make sense) and each release would either be added to the pending list for that protocol or attempted to be grabbed (the other protocol), if both clients are offline then we'll have a list for both.

> Isn't that what this whole feature is about? To retry on every RssSync?

I meant retrying within that same RSS Sync, basically it was down X seconds ago, but lets try again to see if we can avoid waiting for the next RSS Sync. Or did you mean something else?

Taloth:
For the first, yes, exactly.

Second, no. we wouldn't retry in the same RssSync. step 3 & 4 are to prevent exactly that, don't try talk to the same download client if it failed.
The housekeeping is just to determine what needs to be added to the pendingReleaseService.
So if the first release was usenet, and failed, it end up in newStoreUsenet.
Then if the second release is usenet, it's added to newStoreUsenet without even trying to grab it.
Then if the third release is torrent, and it's grabbed succesfully, it gets added to the 'grabbed' list.
Finally the housekeeping will see that the first and second release don't need to be added to the pendingReleaseService, coz the third release was successfully grabbed.

markus101:
It was step 5 that threw me off, but 3/4/5 are all related (if, else if, else), that makes sense.

src/NzbDrone.Core/Download/Pending/PendingReleaseReason.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace NzbDrone.Core.Download.Pending
+{
+    public enum PendingReleaseReason
+    {
+        Delay = 0,
+        StoreAndForward = 1
```
Taloth: I don't think the PendingReleaseReason has any effect on what happens with it. Everything is re-evaluated on RssSync afaik.
`Delay` is caused by a DecisionSpec.
`StoreAndForward` is if sending to Downloader Failed or downloader Disabled.
`Fallback` is any release that didn't get rejected but wasn't the best available.

I don't think there's a difference between StoreAndForward (horrible name) and Fallback.

Delay & StoreAndForward items should be shown on the UI, dunno about Fallback, I think we should filter those.

src/NzbDrone.Core/Download/Pending/PendingReleaseService.cs
```diff
@@ -142,11 +148,12 @@ public List<Queue.Queue> GetPendingQueue()
                                     Size = pendingRelease.RemoteEpisode.Release.Size,
                                     Sizeleft = pendingRelease.RemoteEpisode.Release.Size,
                                     RemoteEpisode = pendingRelease.RemoteEpisode,
-                                    Timeleft = ect.Subtract(DateTime.UtcNow),
+                                    Timeleft = timeleft,
                                     EstimatedCompletionTime = ect,
-                                    Status = "Pending",
+                                    Status = pendingRelease.Reason == PendingReleaseReason.Delay ? "Pending" : "DownloadClientUnavailable",
```
Taloth: Move tertiary expression to a function!

src/UI/Activity/Queue/TimeleftCell.js
```diff
@@ -10,24 +10,39 @@ module.exports = NzbDroneCell.extend({
...
+                    var seconds = FormatHelpers.pad(duration.get('seconds'), 2);
 
-            if (timeleft === undefined) {
-                this.$el.html('-');
-            } else {
-                this.$el.html('<span title="{1} / {2}">{0}</span>'.format(timeleft, remainingSize, totalSize));
+                    var formattedTime = '{0}:{1}:{2}'.format(hours, minutes, seconds);
+
+                    if (days > 0) {
+                        formattedTime = days + '.' + formattedTime;
```
Taloth: '.' -> 'd ' is more human readable. But can't moment format it for us?

src/NzbDrone.Core/Download/ProcessDownloadDecisions.cs
```diff
@@ -36,26 +38,24 @@ public ProcessedDecisions ProcessDecisions(List<DownloadDecision> decisions)
...
-                //Skip if already grabbed
-                if (grabbed.SelectMany(r => r.RemoteEpisode.Episodes)
-                                .Select(e => e.Id)
-                                .ToList()
-                                .Intersect(episodeIds)
-                                .Any())
+                // Skip if already grabbed
+                if (IsGrabbed(grabbed, report))
                 {
                     continue;
```
Taloth: In the future I'd like to get these added as 'Fallback' to the pending list, and either hidden or greyed out on the UI.

src/UI/Activity/Queue/TimeleftCell.js
```diff
@@ -10,24 +10,39 @@ module.exports = NzbDroneCell.extend({
...
+            var status = this.cellValue.get('status').toLowerCase();
+            var ect = this.cellValue.get('estimatedCompletionTime');
+            var time = '{0} at {1}'.format(FormatHelpers.relativeDate(ect), moment(ect).format(UiSettingsModel.time(true, false)));
+
+            if (status === 'pending') {
                 this.$el.html('<div title="Delaying download till {0}">-</div>'.format(time));
-                return this;
-            }
+            } else if (status === 'downloadclientunavailable') {
+                this.$el.html('<div title="Retrying download at {0}">-</div>'.format(time));
```
Taloth: We're throwing every download on the pile, might confuse the user since it's not really gonna download all 20 releases.

Tbh adding Fallback immediately might be preferable, since then you can just mark the first as downloadclientunavailable and the rest as Fallback.

src/NzbDrone.Core/Download/Pending/PendingReleaseReason.cs
markus101: > StoreAndForward (horrible name)

If you have something better let's go for it. 😄 

> Delay & StoreAndForward items should be shown on the UI, dunno about Fallback, I think we should filter those.

Good point, I like that 👍

src/UI/Activity/Queue/TimeleftCell.js
markus101: I don't think moment wouldn't give us much here, good call on the change though.

src/UI/Activity/Queue/TimeleftCell.js
markus101: Yeah, good call, fallback is being added and for the time being at least they will not be displayed in the UI (filtered when building the queue), I like the idea of being able to hide them, so some filtering would be in order, but I don't think it's required.

src/NzbDrone.Core/Download/Pending/PendingReleaseService.cs
```diff
@@ -67,7 +66,7 @@ public class PendingReleaseService : IPendingReleaseService,
         }
 
 
-        public void Add(DownloadDecision decision)
+        public void Add(DownloadDecision decision, PendingReleaseReason reason)
         {
             var alreadyPending = GetPendingReleases();
```
Taloth: What if the release was first Delayed, but then hit a DownloadClientUnavailable?
I think it should update the ECT and Reason. (Can't use Upsert, since that would replace the Added date.)

I'd also give the Debug log entries a bit more info on the type of Pend.

src/NzbDrone.Core/Download/ProcessDownloadDecisions.cs
```diff
@@ -36,52 +38,91 @@ public ProcessedDecisions ProcessDecisions(List<DownloadDecision> decisions)
             var prioritizedDecisions = _prioritizeDownloadDecision.PrioritizeDecisions(qualifiedReports);
             var grabbed = new List<DownloadDecision>();
             var pending = new List<DownloadDecision>();
+            var storedUsenet = new List<DownloadDecision>();
+            var fallbackUsenet = new List<DownloadDecision>();
+            var storedTorrent = new List<DownloadDecision>();
+            var fallbackTorrent = new List<DownloadDecision>();
```
Taloth: A single 'var fallback = ' would do. Then you can also add items to it in 'skip if already grabbed'.

src/NzbDrone.Core/Download/ProcessDownloadDecisions.cs
```diff
@@ -36,52 +38,91 @@ public ProcessedDecisions ProcessDecisions(List<DownloadDecision> decisions)
...
-                        .Select(e => e.Id)
-                        .ToList()
-                        .Intersect(episodeIds)
-                        .Any())
+                if (IsEpisodeProcessed(pending, report))
                 {
                     continue;
                 }
 
+                if (downloadProtocol == DownloadProtocol.Usenet)
```
Taloth: I'm wondering if we're making this code too complex. It should be possible to simply this.
What if we have the 'grabbed, pending, stored, fallback' lists. And two flags, usenetClientFailed & torrentClientFailed, to substitute the storedUsenet.Any(), then we could write:

```
if (downloadProtocol == DownloadProtocol.Usenet && usenetClientFailed ||
    downloadProtocol == DownloadProtocol.Torrent && torrentClientFailed)
{
    if (IsEpisodeProcessed(stored, report))
    {
        fallback.Add(report);
    }
    else
    {
        stored.Add(report);
    }
}
```

Which is far less verbose, has less lists.

Obviously the usenetClientFailed & torrentClientFailed flags would have to be set appropriately, but that's basically already there.
It still maintains a separate 'path' for usenet vs torrent, as intended originally, and has no behavioral changes afaik.

src/NzbDrone.Core/Download/ProcessDownloadDecisions.cs
```diff
@@ -36,52 +38,91 @@ public ProcessedDecisions ProcessDecisions(List<DownloadDecision> decisions)
...
+                    else if (storedTorrent.Any())
+                    {
+                        storedTorrent.Add(report);
+                        continue;
+                    }
+                }
+
                 try
                 {
                     _downloadService.DownloadReport(remoteEpisode);
```
Taloth: I failed to realize, but I don't think we actually throw DownloadClientUnavailableException and simply return in side DownloadReport if no download client is enabled/available.

src/NzbDrone.Core/Download/ProcessDownloadDecisions.cs
```diff
@@ -36,52 +38,91 @@ public ProcessedDecisions ProcessDecisions(List<DownloadDecision> decisions)
...
                     pending.Add(report);
                     continue;
                 }
 
-                if (pending.SelectMany(r => r.RemoteEpisode.Episodes)
-                        .Select(e => e.Id)
-                        .ToList()
-                        .Intersect(episodeIds)
-                        .Any())
+                if (IsEpisodeProcessed(pending, report))
```
Taloth: It already was this way, but I wonder, wont this actually cause undesired behavior?
If one download is Delayed, for example because the user specified a Minimum Age for Usenet, it would cause the Torrent downloads to be delayed as well.

markus101: Hmm yeah, I can see two cases where that would happen, though there may be more:

- The preferred protocol has a delay, but the non-preferred one doesn't
- They both have delays, but the non-preferred delay has already expired

Do we want to filter on the protocol as well?

Taloth: If the non-preferred protocol has no delay then that's intentional or a configuration error. I used minimum age as an example, because it's not affected by delay profiles and thus isn't a configuration error.
We shouldn't filter at all, any release that merits a delay should be delayed by it's own specifications, not by a more preferred release. So I would just remove the entire if statement.

markus101:
Updated with the changes, though I changed things up a bit, instead of deciding whether it should be `DownloadClientUnavailable` of `Fallback` when processing it decides when processing the failed grabs so we don't end up with multiple items for the same episode as `DownloadClientUnavailable` because they are for different protocols. It will also store items that were grabbed but failed as fallback items so they can be retried later since it could be higher quality. A rather long comment explains the logic inline.

src/NzbDrone.Core/Download/ProcessDownloadDecisions.cs
```diff
@@ -36,52 +39,65 @@ public ProcessedDecisions ProcessDecisions(List<DownloadDecision> decisions)
...
 
                 try
                 {
                     _downloadService.DownloadReport(remoteEpisode);
                     grabbed.Add(report);
                 }
-                catch (Exception e)
+                catch (DownloadClientException ex)
+                {
+                    if (ex is DownloadClientUnavailableException || ex is DownloadClientAuthenticationException)
```
markus101: Since an `DownloadClientAuthenticationException` means we failed to connect properly we should also store those releases I think. Both `DownloadClientUnavailableException` and `DownloadClientAuthenticationException`  inherit `DownloadClientException`.

Taloth: So... where is the `else` in this if statement?

PR #981, opened 2015-12-06
iCal parameters

### LINKED ISSUE #974: iCal - Specify Number of Days in the future to request
Right now we let the user specify the start and end dates, but its not obvious and a hard end date sucks for subscribing in another application.

Instead we should let the user specify how many days into the future to show and expose that option in the UI to make it discoverable.

markus101:
@kayone rather simple, but the parameter names could use some work, thoughts?

Closes #974

### DIFF
--- a/src/NzbDrone.Api/Calendar/CalendarFeedModule.cs
+++ b/src/NzbDrone.Api/Calendar/CalendarFeedModule.cs
@@ -22,14 +22,31 @@ namespace NzbDrone.Api.Calendar
 
         private Response GetCalendarFeed()
         {
-            var start = DateTime.Today.AddDays(-7);
-            var end = DateTime.Today.AddDays(28);
+            var pastDays = 7;
+            var futureDays = 28;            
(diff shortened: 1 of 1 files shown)


src/NzbDrone.Api/Calendar/CalendarFeedModule.cs
```diff
@@ -22,15 +22,32 @@ public CalendarFeedModule(IEpisodeService episodeService)
...
+            var pastDays = 7;
+            var futureDays = 28;            
+            var start = DateTime.Today.AddDays(-pastDays);
+            var end = DateTime.Today.AddDays(futureDays);
 
+            // TODO: Remove start/end parameters in v3, they don't work well for iCal
             var queryStart = Request.Query.Start;
             var queryEnd = Request.Query.End;
+            var queryPastDays = Request.Query.pastDays;
+            var queryFutureDays = Request.Query.FutureDays;
```
Taloth: FutureDays -> futureDays

Taloth:
Looks good

PR #999, opened 2015-12-15
Updated OS X startup script to work with macports mono

GavinJS:


markus101:
@GavinJS could you clarify what the advantage of this change is? Saves the user from installing mono a second time if it already exists via another installation method?

GavinJS:
@markus101 yes that's correct...  If you install Mono via the MacPorts package manager it installs everything into the default location of /opt so the Sonarr startup script always fails because it's hard coded to use the libs in the Mono framework downloadable package.

ugokanain:
Was this change incorporated and then dropped? I also use macports, and Gavin's change is working fine

Please added to the mac version

kayone:
Merged into develop. 
Thanks!

Issue #969, opened 2015-11-30
Prevent Grabbing if Episode Hasn't aired


markus101:
Fakes are a problem on public sites that we support, if the episode hasn't aired yet preventing the download would prevent the fake.

If we add it it should be opt-in as there are many legitimate releases before the air date and this also adds yet another thing that will prevent an episode from being grabbed (support hell).

kayone:
how about this case, today is 20th, episode aired on 15th and we find a release today (20th) from 10th? would we still not download it?

markus101:
Correct, it would not be downloaded because it was released before the air date.

thezoggy:
maybe offer the toggle per indexer? that way you can turn it on for those crappy public torrent sites..

bjeanes:
Can I propose at least doing this for shows marked "daily" (which I also use for some weekly shows like Last Week Tonight). Those shows have enough regularity and are recorded mostly live that a pre-release is always going to be a fake. The occasional fake Elementary or similar doesn't bother me but with Last Week Tonight I keep getting episode for the *next* week but it's really some random episode from Season 1, for example.

bjeanes:
If you do have an option to prevent pre-release fetching, it could be per-show and opt-in. That way certain problem shows can be excluded.

magne4000:
Any news on this? Lately I had many fakes downloaded because this feature doesn't exists

Devian50:
I would like an update as well. I personally have had Sonarr somehow deciding a file it downloaded while looking for a special episode aired Friday was actually a main season episode that airs in over a month, which then causes it to download the other two special episodes again as main season episodes. The show's main season hasn't even started airing yet and won't for another month.

killerz298:
I have this issue with the new Ducktales. Sonarr keeps grabbing a scene release which is 1 episode number off for some reason and it imports it as the next episode which has not aired. I have to constantly fix this.

dejected1:
I too have this issue with the new ducktales. Came here to post for this reason actually.  Enabling this per show or per indexer seems ideal. i wouldn't mind a global option either.

aeonsablaze:
I would also quite like this feature. Particularly on a per show basis.

kelemvor33:
Came here about this specific issue.  Got a download for an episode that Sonarr shows doesn't air for another 12 days.  Since it knows it hasn't aired yet, it should be smart enough to know that the files isn't legit and skip it.

thezoggy:
[...]

PR #1018, opened 2015-12-21
Fixed: Prevent grabbing if episode hasn't aired

vawen:
Issue #969

### DIFF
--- a/src/NzbDrone.Api/Config/IndexerConfigResource.cs
+++ b/src/NzbDrone.Api/Config/IndexerConfigResource.cs
@@ -8,5 +8,6 @@ namespace NzbDrone.Api.Config
         public int MinimumAge { get; set; }
         public int Retention { get; set; }
         public int RssSyncInterval { get; set; }
+        public bool GrabUnairedReleases { get; set; }
     }
 }
--- a/src/NzbDrone.Core.Test/DecisionEngineTests/UnairedReleaseSpecificationFixture.cs
+++ b/src/NzbDrone.Core.Test/DecisionEngineTests/UnairedReleaseSpecificationFixture.cs
@@ -1,0 +1,82 @@
+﻿
+using FizzWare.NBuilder;
+using FluentAssertions;
+using NUnit.Framework;
+using NzbDrone.Core.Configuration;
+using NzbDrone.Core.DecisionEngine.Specifications;
+using NzbDrone.Core.Parser.Model;
+using NzbDrone.Core.Test.Framework;
+using NzbDrone.Core.Tv;
+
+namespace NzbDrone.Core.Test.DecisionEngineTests
+{
+    [TestFixture]
+    public class UnairedReleaseSpecificationFixture : CoreTest<UnairedReleaseSpecification>
+    {
+        private RemoteEpisode _goodRemoteEpisode;
+        private RemoteEpisode _fakeRemoteEpisode;
+
+        [SetUp]
+        public void Setup()
+        {
+            var show = Builder<Series>.CreateNew().With(s => s.Id = 1234).Build();
+
+            _goodRemoteEpisode = new RemoteEpisode
+            {
+                Episodes = Builder<Episode>.CreateListOfSize(2)
+                                           .All()
+                                           .With(e => e.AirDateUtc = System.DateTime.UtcNow.AddDays(-8))
+                                           .With(s => s.SeriesId = show.Id)
+                                           .BuildList(),
+                Series = show,
+                Release = new ReleaseInfo
+                {
+                    PublishDate = System.DateTime.Now
+                }
+            };
+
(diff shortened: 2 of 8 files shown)


thezoggy:
does this allow to be series specific (using tags? edit show / can be bulk set in series editor)

vawen:
At this moment it is a global setting. It can be changed easily to be defined per indexer,
I would need to check about tags (have not played with it).

markus101:
I don't think per series makes a lot of sense (support hell). Torn on per indexer, most indexers that actually have this problem also support verified only releases which avoids this issue at the cost of missing out of some less common series.

thezoggy:
the only time ive ever seen this feature to be a benefit is due to spammers trying to post fake GoT
lots of shows have ep released early.. which is why id want to not turn it on globally but only for a very specific show...
if i was to use it (i wouldnt, unless the need arises) 

ideally people should just use better indexers... but what if its an indexer thats normally good.. 
and the admin/mods are just away.. hence the GoT spam that happend each week for a month straight

globally setting it on, i'd think would do more harm than good.. as lots of people may be mift at the fact that the release they want is now too old to grab because of this cock blocking it
example, first ep of supergirl came out 6months early
thus if you waited for it to 'air' then well.. your going to try and grab an ep 6 months old

if you offer the series wide on/off.. you really should extend that to series specific/tags
and its no more 'support hell' than require/ignore words

Serneum:
Some more examples of where this might not be a good idea globally:
1. The Magicians. The first episode aired early but the official premiere is a two-parter this week.
2. The Shannara Chronicles. The first four episodes were made available online and then also aired weekly. 
3. DaVinci's Demons. The entire 3rd season was put online and then one episode was aired per week for 10 weeks.

That being said, I like the idea. I'd probably leave it off unless a series was having some issues. It might help with Saturday Night Live where sometimes the promos for the episodes get grabbed instead of the actual episodes.

markus101:
Good point about tags, I'd be for a tag based approach with the ability to apply all series by leaving it untagged. One option to enable and another to set the tags.

vawen:
Sound good, will implement the tag approach.

markus101:
@vawen sorry, I didn't mean to say get started immediately, I was hoping to get @Taloth or @kayone to weigh in first as it really changes the behaviour of the feature and may impact other decisions we make about settings.

Taloth:
We would be moving from an initial "no, too many good releases get released early, fix your indexer, and remove TPB" stance to "yeah, sure, we'll make it fully customizable".
If we do something like this we might as well make most configuration options 'tagable'. (mind you, that was one of the original concepts from when we started with tags, but considered too complex)
So, no, I can't agree with this.

Personally I would say: Keep just one global flag. If we ever decide to embrace tags, we can always add it. But for now, if ppl want to take the shortcut and turn the flag on, they can. Yes, they might have issues with episodes that are always released early, bad luck, it's them with crappy indexers, not us.
(btw. Looking at the code there's still some remnants from a per-indexer option, which should be removed)

But really, what are we trying to solve here? The question isn't about 'early' releases, it's about early releases occasionally being fake.
Why not add a config option that delays anything released before airdate until x hours after airdate? Preferring those released after?
With usenet you get the chance that the download is no longer available, but that's the tradeoff, it's a bit more softlogic compared to a hard rejection.
Easily done with ```public RejectionType Type { get { return RejectionType.Temporary; } }```

But why stop there? you can add the config option to Delay Profile, after all, 'delaying' is what we're talking about now. (You'll even get your tags for free... dammit why did I suggest this?!)

Only issue is how to reliably reduce the priority of such releases, without risking valid early 1080p releases from losing to some later SD release.

Anyway, think about it.

Sry, for the wall of text.

vawen:
If you really think about the problem it is crap indexers allowing fake uploads so it would have sense to have this linked per indexer. Just my 2 cents. (I had this problem only 1 time with kat)

markus101:
@vawen thats definitely a big source of it, but in a lot of cases the users are using only crappy indexers, blocking at purely at the indexer level still doesn't solve things that legitimately come out before the release date (airs a day earlier in another country) and blocking it per series would help there.

vawen:
Hum, maybe per indexer and be able to disable on individual shows? Probably it is too complex, for that route I think it would have more sense a tag approach.

markus101:
I think for now lets do it with tags.

markus101:
To clarify, I was referring to a delay profile with tags, which has its own complexities and a lot of logic that needs to support that (the more I think about it the more complicated it gets).

I think we're starting to over think this and trying to solve a bunch of potential uses when really we've been asked for a super simple toggle and we're getting away from that. A small percentage of the total user base has asked for this, so lets not over think this and go with a simple toggle, which applies to all indexers and all series.

wejendorp:
What prevents this from getting merged?

PR #1029, opened 2015-12-27
Provide 2x srcset for posters

### LINKED ISSUE #804: Posters need to be shown at 500px resized on retina screens
Webui needs to switch to 500px posters instead of 250px on retina screens

jakepusateri:
Fixes #804


jakepusateri:
Wrong commit author, closing

Issue #954, opened 2015-11-23
Paths with colons prevent Sonarr from communicating with Transmission


uzegonemad:
Based on [this forum thread](https://forums.sonarr.tv/t/cannot-combine-ospaths-of-different-platforms/8207), if a path in Transmission contains a colon, Sonarr will throw the OsPaths exception and will not be able to communicate with Transmission.

markus101:
Already resolved.

PR #1032, opened 2015-12-29
Fixed: Paths with colons prevent Sonarr from communicating with Transmission

vawen:
Issue #954


markus101:
I'm not seeing how this solves the issue, OsPath would still throw in that case.

vawen:
This problem only appear when a magnet is added and is downloading the metadata. Once the metadata is downloaded, the name of the torrent change and removes forbidden characters.

markus101:
The path is the issue, not the name, the forum post linked in #954 shows the full path Sonarr had issues with.

Taloth:
Yes, but the name isn't a path, it's the display name provided by the magnet url. Once the magnet finishes downloading and the torrent becomes available, the release name of either the dir or file is used instead.
I still agree we ought to fix the colon, but that's just a contributing factor, not the original issue. (I intended to fix the colon when merging in this PR, so don't worry about it :D)

PR #1033, opened 2015-12-29
RTorrent magnets fixed.

gpapaz:
* Magnets added will start downloading instead of waiting and
deleted after timeout.
	* Timeout increased to give more time for magnets to download,
 especially for older releases.

### DIFF
--- a/src/NzbDrone.Core/Download/Clients/rTorrent/RTorrent.cs
+++ b/src/NzbDrone.Core/Download/Clients/rTorrent/RTorrent.cs
@@ -41,22 +41,32 @@ namespace NzbDrone.Core.Download.Clients.RTorrent
         {
             _proxy.AddTorrentFromUrl(magnetLink, Settings);
 
+            //Magnet also download to the specified location.
+            SetDownloadDirectory(hash);
+
             //Wait for 5mins - some old magnets need more time to be downloaded!
             var tries = 30;
             var retryDelay = 10000;
             if (WaitForTorrent(hash, tries, retryDelay))
             {
+                            
                 _proxy.SetTorrentLabel(hash, Settings.TvCategory, Settings);
 
                 SetPriority(remoteEpisode, hash);
                 SetDownloadDirectory(hash);
 
                 _proxy.StartTorrent(hash, Settings);
+
+                //If download directory is specified, we need to force recheck to start torrent.
+                if (!Settings.TvDirectory.IsNullOrWhiteSpace()) {
+                    _proxy.RecheckTorrent(hash, Settings);
+                }
+                
                 return hash;
             }
             else
             {
-                _logger.Debug("rTorrent could not resolve magnet {0}. Removing", magnetLink);
+                _logger.Error("rTorrent could not resolve magnet {0}. Removing", magnetLink);
 
                 RemoveItem(hash, true);
 
--- a/src/NzbDrone.Core/Download/Clients/rTorrent/RTorrent.cs
+++ b/src/NzbDrone.Core/Download/Clients/rTorrent/RTorrent.cs
@@ -41,8 +41,9 @@ namespace NzbDrone.Core.Download.Clients.RTorrent
         {
             _proxy.AddTorrentFromUrl(magnetLink, Settings);
 
-            var tries = 10;
-            var retryDelay = 500;
+            //Wait for 5mins - some old magnets need more time to be downloaded!
+            var tries = 30;
+            var retryDelay = 10000;
             if (WaitForTorrent(hash, tries, retryDelay))
             {
                 _proxy.SetTorrentLabel(hash, Settings.TvCategory, Settings);
@@ -51,7 +52,6 @@ namespace NzbDrone.Core.Download.Clients.RTorrent
                 SetDownloadDirectory(hash);
 
                 _proxy.StartTorrent(hash, Settings);
-
                 return hash;
             }
             else
--- a/src/NzbDrone.Core/Download/Clients/rTorrent/RTorrentProxy.cs
+++ b/src/NzbDrone.Core/Download/Clients/rTorrent/RTorrentProxy.cs
@@ -28,7 +28,7 @@ namespace NzbDrone.Core.Download.Clients.RTorrent
         [XmlRpcMethod("d.multicall2")]
         object[] TorrentMulticall(params string[] parameters);
 
-        [XmlRpcMethod("load.normal")]
+        [XmlRpcMethod("load.start")]
         int LoadUrl(string target, string data);
 
         [XmlRpcMethod("load.raw")]
--- a/src/NzbDrone.Core/Download/Clients/rTorrent/RTorrentProxy.cs
+++ b/src/NzbDrone.Core/Download/Clients/rTorrent/RTorrentProxy.cs
@@ -21,6 +21,7 @@ namespace NzbDrone.Core.Download.Clients.RTorrent
         void SetTorrentDownloadDirectory(string hash, string directory, RTorrentSettings settings);
         bool HasHashTorrent(string hash, RTorrentSettings settings);
         void StartTorrent(string hash, RTorrentSettings settings);
+        void RecheckTorrent(string hash, RTorrentSettings settings);
     }
 
     public interface IRTorrent : IXmlRpcProxy
@@ -36,6 +37,9 @@ namespace NzbDrone.Core.Download.Clients.RTorrent
 
         [XmlRpcMethod("d.erase")]
         int Remove(string hash);
+
+        [XmlRpcMethod("d.check_hash")]
+        int Recheck(string hash);
 
         [XmlRpcMethod("d.custom1.set")]
         string SetLabel(string hash, string label);
@@ -240,6 +244,19 @@ namespace NzbDrone.Core.Download.Clients.RTorrent
             if (multicallResponse.Any(r => r != 0))
             {
                 throw new DownloadClientException("Could not start torrent: {0}.", hash);
+            }
+        }
+
+        public void RecheckTorrent(string hash, RTorrentSettings settings)
+        {
+            _logger.Debug("Executing remote method: d.check_hash");
+
+            var client = BuildClient(settings);
+
+            var response = client.Recheck(hash);
+            if (response != 0)
+            {
+                throw new DownloadClientException("Could not recheck torrent: {0}.", hash);
             }
         }
 


gpapaz:
Related to the problem reported here: https://forums.sonarr.tv/t/rtorrent-magnet-not-resolved/8608/3.
Review and let me know what you think of it.

Taloth:
This would hang up the thread for way too long. The current delay of 5 sec is already pushing it.

Might I suggest an alternative approach:
Write a new service ```MagnetResolverService``` with appropriate interface. Afaik the infohash is already extracted from the magnet. Then hit up torcache to download the torrent. Use that as primary means to fetch the torrent in case of a magnet.
Obviously, if torcache fails or doesn't return a result, fallback to the normal Add by magnet logic.
That should avoid the whole problem in 99% of the cases.

Keep the 5 sec delay as is was.
Replace the Debug log entry with Error, ppl ought to know when something fails.
Optionally: If the magnet failed within the specified time, don't delete it. This might be a bad idea, coz it probably clutters rTorrent needlessly.

gpapaz:
The most important issue is that the magnet files do not initialize their downloading. The load.start makes sure they do, and in most cases they are resolved before the timeout. However for some older releases they do take  a lot more time, that's why I opted to raise the wait time. I believe the best approach, time wise, would be to at least keep the same timeout and modify the load operation.

IMO the wait time could be an advanced option. I, for example, do not care if the thread is locked for 5 mins. I definitely prefer it instead of losing episodes.

Also I agree on the option to not delete the magnet files, if they haven't been resolved. That too, could be an advanced option.

Regarding the service approach, it certainly makes sense, but it would take a lot more time to write and develop and we still have to solve the timeout and non resolution issue, even if it is for a mere 1% of the cases. Will look into it, but I am not sure I will have the time/will to do it.

gpapaz:
OK I shouldn't have closed it ... not yet at least :)

Taloth:
Question, does using ```load.start``` allow us to set the category immediately? before the magnet finished fetching the torrent? coz that has been the biggest problem, without setting the category we can't track the download.

If you want I can make that service if you don't have the time.

gpapaz:
No., it doesn't and neither does the current implementation with the load.normal. But, the later loads the magnet and leaves it in stopped state. No matter how long the timeout is, it will never get the torrent file.

If you make the service, I do not see where you need me :).  I made this pull request to fix the initialization bug that the current implementation has. I went a step ahead to give to those magnets that required it, a bit more time to download. I understand why my approach is problematic, but at least the initialization part has to be included, if you want magnets to start the torrent downloading.

Taloth:
So you're saying that ```load.normal``` never even worked, ever? Wow.

Well, I wanted to make the service for you if you didn't have the time. Then I would shove it your way so you can test the whole thing :)
I was actually doing other stuff, you know.

gpapaz:
Yes that is exactly what I am saying, at least for the httpcrpc through the uTorrent case.

I can test it for my test environment. Not sure that it would be a good enough testing to accept it for other environments as well - e.g RPC2 end points and other mechanisms to do xmlrpc for the rtorrent which I am not aware of.

Taloth:
Joy... wrote all of it, then went to test it. Couldn't find anything on rarbg that had the torrent cached on torcache, and we knew kat already has torcache links. So that's going to be pretty useless.

gpapaz:
Lets, at least add the part that starts the magnets. Leave the timeout or increase it slightly. Another approach for a future update could be to create a thread for checking magnets only. So you could increase the timeout waiting the torrent to be downloaded without locking the main thread.

Taloth:
Yup, Atm my idea is indeed to include load.start and 15sec or something. But wanted to ask @markus101 about the timeout. And the possibility to add a switch to prefer either magnet or url. (another config option, yuck)
Problem with deferring the check to another thread is that we're in the middle of adding a download, if the download fails to add we pick another. So we simply have to wait for it to either succeed or fail.

gpapaz:
Why not to prefer the URL and switch to magnet if the indexer doesn't support it? No need to burden the user with that selection.

Taloth:
It's not that easy, some ppl prefer magnet (and it's the default atm). So swapping it around could make some users unhappy.

rarbg:
@Taloth 
Slow response from my side , but we do post everything to torcache.
Can you tell me 1-2 torrents that are not there ?

gpapaz:
@Taloth any chance we will include the basic fix into Sonarr's release? It gets a bit old to have to run from source to get the magnets work - even the latest ones.

Taloth:
@rarbg released yesterday ```The.Blacklist.S03E10.720p.HDTV.x264-FLEET[rartv]```
```HTTP request failed: [404:NotFound] [GET] at [http://torcache.net/torrent/5371776E1C8C5A7AFB681F1BBB3170E4E796E955.torrent]```

@gpapaz Sry, thought I already did pulled in that specific change to develop. I'll see if I can get it into develop this weekend.

**Comment by rarbg** [comment]
@Taloth 
works fine for me ?

**Comment by Taloth** [comment]
Aw fuck, it's that again:
Sonarr fails, browser works, wget fails. Probably User-Agent or referrer checks server side.

Set Referer to http://torcache.net and now it works in Sonarr.

Tnx @rarbg, for making me look again. :smile:

(Ugh, Referer hacks. :sob:)

**Comment by rarbg** [comment]
@Taloth 
np man :P also switch user agent just in case they decide to block that too

**Comment by Taloth** [comment]
@gpapaz I was prepping to merge stuff in, but something curious.
You changed load.normal to load.start, we discussed why before. But the same method is used for non-magnet downloads.
There's one issue with that, coz atm we load the torrent, set category etc, AND the download folder. Before finally starting the torrent.
I'm afraid that we have an issue of we if we don't keep the torrent/magnet suspended before setting the download folder that.
Any input on that?

**Comment by gpapaz** [comment]
@Taloth you do not use the LoadUrl method anywhere else but in the AddTorrentFromUrl, which is used in the AddFromMagnetLink. So, in that context, you need to start the magnet and then, when the resolution to a torrent file has occurred, proceed with the rest of the actions - this hasn't been changed by me by the way.

**Comment by Taloth** [comment]
My point is [this line](https://github.com/Sonarr/Sonarr/pull/1033/files#diff-74dcee7ccb9efdd8844b5a16a6c90f7dR31) changes the LoadUrl call to load.start instead of load.normal.
My question is, how does RTorrent behave to [this](https://github.com/gpapaz/Sonarr/blob/rTorrentMagnets/src/NzbDrone.Core/Download/Clients/rTorrent/RTorrent.cs#L54). We set the download directory the line above, while the torrent was already started (and thus downloading) earlier by load.start.

Guess my question is, does RTorrent properly move the already downloading files to the new specified location?

**Comment by gpapaz** [comment]
I only tested it with default download directory - blank setting option that is. I checked with a specific directory, and the torrent, after magnet being resolved, is paused, but set to download to the specified directory. I thought your code would work for that part, but it seems it did not :). Anyways, one can manually start the download - by forcing recheck on it. I am looking into doing it from within the code, of course ...

**Comment by PartTimeLegend** [comment]
This is a fix desperately needed.

**Comment by Taloth** [comment]
@gpapaz 
Is that latest test with 'load.normal' or 'load.start'?
Please forgive my ignorance, I'm not sure whether 'load.start' was needed to begin resolving the magnet, or only to auto-start the torrent after the magnet was resolved.

What we could do is have the RTorrent class keep a shadow list of 'pending magnets' together with the required settings to be applied after being resolved. That way, if the 5 sec wait time expires, we could still keep an eye on the magnet for a longer period without blocking the search/rss function.
And after 1-2 minutes mark the download as 'failed', which triggers the normal failed download handling in Sonarr, deleting the magnet and searching for an alternative release.
During the 1-2 minutes we could then also keep showing the download in the queue via the use of the shadow list.
Of course, restarting Sonarr would mean the shadow list becomes lost, and any magnet would become orphaned since it doesn't reside in the required category.

Just something I thought about yesterday, so there might be something I overlooked with this idea. But it might give us the best of both worlds.

**Comment by gpapaz** [comment]
@Taloth, first of all, I do not know how the behavior is for the torrent file case. I assume everything works as expected, so I didn't modify anything related to it. I couldn't check either, since all the results I got from my indexers where for magnet files :(

"load.start" is absolutely needed, otherwise the magnet doesn't start at all.

The initial code that I committed was working as expected for default download directories. For the case of a specified directory, the torrent was paused when it was moved to that directory. A force check starts it again. The latest commit fixes this. Priority and labels were and still are correctly applied and handled.

Only issue remaining is the time limit. You proposed solution might be the best, but I lack the deep understanding of your code base to implement it. Not to mention that these are the absolutely first lines in C# I have ever written. I would suggest a 1.5min limit for now and have someone implement the solution you described.

**Comment by Taloth** [comment]
Okay, tnx, that's useful.

One more question, does changing the Download folder leave any junk in the old one? Or does it get moved over and properly picked up in a Recheck?

I prefer pushing a fix for the load bug first, so I'll look into the 1.5min logic/shadow list later.
I may have to install rTorrent somewhere...

**Comment by gpapaz** [comment]
No junk, apart from the magnet files. But those are small and can be manually deleted, if required.

**Comment by rarbg** [comment]
@Taloth 
Rtorrent is known to have a lot of magnet issues , thats a known fact.
DHT is slow and broken. PEX is not working properly.
Out of curiosity why dont you just make an internal magnet2torrent in sonarr and just send the .torrent file to rtorrent instead of dealing of rtorrent's broken implemention of __anything__ ?

**Comment by gpapaz** [comment]
Because it was easier for me to find out how the rpc works in ruTorrent and add the few lines missing to fix the bug :)

**Comment by rarbg** [comment]
@gpapaz 
ruTorrent or rtorrent :P

**Comment by duramato** [comment]
A torrent file can be made from just the hash on torrent caches like http://torcache.net/ or http://itorrents.org/ 
from a link like http://torcache.net/torrent/640FE84C613C17F663551D218689A64E8AEBEABE.torrent

**Comment by rarbg** [comment]
@duramato NO SHIT dude ? how did you sleep ?

**Comment by duramato** [comment]
@rarbg lol , just pointing out :P

**Comment by gpapaz** [comment]
@rarbg ruTorrent, since I was having the issue for the httprpc plugin of it. And used that to undestand how it should work for my test case. But, I assume the commands are specified at the rTorrent's rpc implementation so you can interchange them :P

**Comment by rarbg** [comment]
@duramato this issue is started because of those mirror sites missing torrents or are changing stuff to block apps from downloading files
@gpapaz :) ah ok

**Comment by Taloth** [comment]
No worries, I already had torcache working. But not all torrents are available, and it isn't smart to depend solely on some site that can be taken down.
So I still want rtorrent magnets to work even if it's a little bit.

Atm I simply don't want to run our own Dht node, in case that's what you were suggesting.

**Comment by rarbg** [comment]
You dont need to run a DHT node in order to do magnet2torrent....
Sample code https://github.com/danfolkes/Magnet2Torrent/blob/master/Magnet_To_Torrent2.py
It just uses libtorrent( http://www.libtorrent.org/ ) to load the magnet,  check it every 1 second .has_metadata() and as soon as the .torrent file is downloaded from peers it saves the .torrent file and destroys the instance

c++ code samples http://www.libtorrent.org/examples.html
i cant find any c# bindings but im sure https://github.com/mono/monotorrent this will have/emulate the same functionality

**Comment by PartTimeLegend** [comment]
I see this pull request as fixing what I see as a critical bug as my instance just adds a magnet and then deletes it. Would it be acceptable to merge this into the developing branch and let those of us who like to live on the edge see if it resolves our problem?

**Comment by Taloth** [comment]
I actually installed rtorrent on my server to check this out. Eventually was like 2am so bailed out.

Here's the status:
rtorrent automatically creates a new download based on the meta download, but doesn't transfer any of it's properties, there has been a rtorrent bug report on that since 2013.
In the .rtorrent.rc file the task ```schedule = tied_directory,6,5,start_tied=``` causes that new download to be started automatically
I just commented out the line and the magnet remains paused, so far theory confirmed.

The current workaround is to set the download path and then recheck, but this potentially leaves junk files in the other directory, so far from optimal.
Then an idea: rtorrent scripting. Initially I regretted opening that can of worms, but at 2am I got something that worked, so happy that I did. (But man... undocumented shit)
Basically I do this:
```
var key = "event.download.inserted";
var cmd_key = "schedule_ignore_" + hash;

// rtorrent-scripting for:
// if (d.hash()==hash)
// {
//    d.ignore_commands.set(1)
//    system.method.set_key("event.download.inserted","schedule_ignore_" + hash)
// }
var value = "branch=\"equal=d.get_hash=,cat=" + hash + "\",d.set_ignore_commands=1";
value += ";branch=\"equal=d.get_hash=,cat=" + hash + "\",\"system.method.set_key=" + key + "," + cmd_key + "\"";
value += ";branch=\"equal=d.get_hash=,cat=" + hash + "\",print=SonarrSetIgnoreCommands";

RtorrentApi::system.method.set_key(key, cmd_key, value)
```

What this does is schedule an eventhandler to catch when the new download (resulting from the meta download) is inserted, and then immediately set the ignore_commands flag.
This prevents start_tied from starting the torrent.
The script also removes itself after it's first use.
And finally printing a line for debugging purposes :)

Two things:
1) I couldn't get multiple commands in one branch instruction, so I have 3 atm. That's something I wanna improve, but I'm glad I got this far already.
2) If we can do this... then who needs ignore_commands? Why not set the category (custom1) and download folder via this script? We could potentially even forgo waiting for it.

rarbg:
waaait wait waait wait.
are you trying to load a magnet with a custom download dir ? is that the issue ?

gpapaz:
@Taloth what version? I am testing against a rather old one 0.9.3, but I assume a lot of seedboxes are stuck with that. I had also initially tested with the latest one, but since I didn't see any major changes for the parts we care for, I reverted to use the seedbox as it was easier for my set up (no need for virtual machines etc.). Do you use xmlrpc, httprcp or any other plugin - I assume at the end it shouldn't matter, since all use the rpc specification of the rTorrent.

The .rtorrent.rc line is ```schedule = watch_directory,5,5,load_start= ```. Since it is set by me seedbox provider I wouldn't want to mess with it. I assume that the ```load_start``` would be safer, since it would start the magnet in your case and in mine ... 

You insist that the set_directory  command will leave potential junk in the old directory. I don't believe this is the case. The ruTorrent GUI allows you to move a torrent to a new directory. I didn't see any junk left and it uses the same commands. I remind you that in our case, the torrent hasn't yet started - we have to issue the StartTorrent command, **after** the magnet has been resolved! Otherwise the torrent is stopped, and no data have been downloaded. The flow I have proposed is, magnet add, magnet start and set new directory for magnet - this can also be left to default to be honest. When magnet is resolved to a torrent, set label, set directory, set priority, recheck torrent to make sure that the new directory is actually used by rTorrent and then start torrent. I can not see how junk can be left somewhere - unless by junk you mean the magnet and the torrent files. I didn't see any left in the default directory, though I can't guarantee that it won't.

I see where you go with the script. You want to set all the torrent properties from the initial magnet add command. I tried that too, but sadly, due to my old rTorrent version (?), they weren't picked from the magnet. As you said this would solve the issue with the wait time, but that might mean that newer version of rTorrent are only supported. So my seedbox would not :(

Bottom line IMO is , we currently can't add episodes to download by magnet files in the rTorrent due to them never started and deleted by the expiration timer. I have made some changes that will get the most of them - at least the latest and easily resolved magnets - to start and pull their episodes. The later will be downloaded with the Sonnar label, the defined priority and the specified directory and allow people to use Sonarr with the rTorrent based seedboxes. If they need to clean up some left junk now or then (magnets and torrent files, no actual episode files), I am sure they wouldn't mind the annoyance, until at least we come up with a better solution for the expiration timer and the junk removal.

Taloth:
@gpapaz Well, I'm on 0.9.2 which is even lower.
The script doesn't cause the properties of the magnet download to be copied to the torrent download, it can't, it's been a known limitation in rtorrent. What the script does is immediately setting some properties when the torrent download gets inserted into the download list by rtorrent. Initially it was intended to prevent start_tied from starting the torrent, so Sonarr can set the dl dir. But nothing prevents us from using the same mechanism to set the dl dir on that event.
I'm at work atm, so I can't test the 'junk file' theory. But pretty much all scripts i've seen separately trigger a 'mv' and 'set download_dir' command. The point remains that, however briefly, a sonarr torrent is downloaded in the wrong location.

@rarbg I don't care where the magnet meta gets downloaded, but I do care were the torrent gets downloaded, these are separate downloads. (you possibly meant the same, but terminology with magnet vs meta vs torrent is ambiguous)
With a default rtorrent config:
- Sonarr pushes magnet to rtorrent.
- Sonarr starts magnet.
- Magnet gets downloaded as meta file.
- Rtorrent deletes magnet download, replaces it with a torrent download attached to the meta file. But doesn't transfer any other property. So the torrent gets the default download dir set.
- Rtorrent start_tied causes the torrent to be started within 5 sec.

gpapaz:
@Taloth, the way I do it, and since the torrent isn't started after magnet download, there is no download happening, when I issue the set directory command. The force recheck is needed so that the torrent correctly registers the new directory, otherwise when I issue the start it starts and soon enough becomes and remains paused - with some content been downloaded in the meantime. But as I do it, when the force recheck is issued, no content has been downloaded, the directory is correctly set and the torrent starts downloading at the specified directory. No junk can have been left behind in this way, because none has ever been downloaded!.

rarbg:
@Taloth 
by
waaait wait waait wait.
are you trying to load a magnet with a custom download dir ? is that the issue ?
i meant exactly that
are you trying to load a magnet url into rtorrent and set a custom download location for the data of the torrent ( NOT .torrent file itself or the magnet temp data ) ?

Taloth:
Yes, but also the category (custom1).

rarbg:
custom1 is called label , not category but whatever

wont the following work ?
load(not load_start) with the magnet link
d.set_custom1
d.set_directory_base
d.set_directory

Taloth:
No, it won't work. The magnet download (with those properties set) gets deleted by rtorrent upon completion, and replaced with a torrent download (without those properties set). That's the whole problem.

rarbg:
does it loose both custom1 and directory ?

Taloth:
Yes, everything. (and yes, I checked the rtorrent sourcecode)

rarbg:
I get it now.
https://github.com/rakshasa/rtorrent/issues/376
Lets see if @rakshasa can help

Taloth:
@rarbg I appreciate the help of course, but growing a bit frustrated now.

1) I already posted earlier that "rtorrent automatically creates a new download based on the meta download, but doesn't transfer any of it's properties, **there has been a rtorrent bug report on that since 2013**."
Specifically https://github.com/rakshasa/rtorrent/issues/107, but that won't get resolved without someone PRing a fix. (larsjohnsen is the guy who implemented rtorrent support in Sonarr, so we've been aware of this limitation for a while)
2) Even if it gets fixed it won't help anyone with an older rtorrent version. And that's where my focus has been.

rarbg:
@Taloth 
for 1. All i can do for you was open a ticket from my account in rtorrent and maybe get some response.
for 2. Check my new project i just notified you about

Taloth:
@gpapaz This is what I pushed to develop. https://github.com/Sonarr/Sonarr/commit/cf25097cd137c2f409f8a2b816bd89a9d6dc6563 (version 2.0.0.3762)

If we can get a fix into rTorrent itself we'll need a version check, but so far so good.

This seems to work properly on 0.9.2, can you do an independent test?

Please note that if the magnet can't be resolved, it will remain stuck (I removed the RemoveTorrent step), but we can implement some Failed Download Handling to tackle that later asynchronously since the magnet has the correct hash &  category.

gpapaz:
@Taloth I tested your commit. The good news are that the meta now starts downloading and that the properties (label, directory, priority) are correctly set and applied to the torrent - when and if the meta resolution succeeds.

The bad news. Time limit remains too small. Older meta have no chance to be resolved. This as you said keeps them stuck. They either constantly download or, if they are resolved, their torrent stays stopped and requires user intervention. What is more annoying, is that even if they are resolved, the torrent remains stopped. Shouldn't you at least try to start the torrent, when the meta resolution succeeds? Do I need to add the schedule line you included in your comment https://github.com/Sonarr/Sonarr/pull/1033#issuecomment-173121777 to have the torrent auto-started? Shouldn't you somehow instruct users, they need to have a specific set up for their rTorrent client? IMO, I find this a bit awkward, if indeed a specific set up instruction is needed.

Taloth:
Add commands.Add("d.try_start=") just before the print= one should autostart it if you don't have start_tied configured. I haven't tested that though.

gpapaz:
Hmm, nope. It adds the meta/torrent alright and the later remains stopped - never starts in fact. Had to manually start it. Did you test without the schedule commands in the rtorrent.rc file? I haven't added anything in mine, yet.

Also the way you handle the failed meta resolution creates a "strange" effect. The episode is listed as added and queued for downloading, even when the meta hasn't yet been resolved. If the null is returned, but since meta is not deleted, when and if at some point it is resolved, wouldn't the torrent been added at the sonarr queue, since it is under the tv-sonarr label? Wouldn't this be better for the user's perspective of the state of her searches/downloads?

Taloth:
Yes, I tested with the schedule start_tied removed... twice.
Does the torrent get stuck as closed or open?
I assume you see the print message output in the rtorrent console? 

Secondly, I return the hash instead of null even after timeout. So sonarr knows it's been added. That way sonarr won't try another immediately. As I said, failed download handling will have to implement the logic to handle the magnet never resolving.

gpapaz:
@Taloth the torrent's status is stopped, if this is what you ask.
I do not have console access, only ruTorrent web GUI is given to me. 
I do not see any print messages output on the ruTorrent from your commands.

Taloth:
** sigh ** guess I'll be installing rutorrent too.

gpapaz:
You can use a docker to not have to tinker with basic set up. Btw I am using httprpc plugin to communicate with the ruTorrent.

IMO the problem is not with the ruTorrent. I say this, because our calls are handled as raw, thus they are transferred to rTorrent as we send them. I will try to test with my local set up as well, to see if I can help with debug info etc.

Taloth:
I compiled rtorrent 0.9.3, rutorrent, used the httprpc. same results all across the board. Only thing I can think of is that the ```d.try_start=``` should be ```d.start=``` depending on your rtorrent configuration. (I was being cautious using ```try_start``` since it honors the ```ignore_commands``` property)
The fact that the label and download dir get set properly means that the script is definitely running.

gpapaz:
Tested with d.start. It worked and torrent started!

Taloth:
w00t... :smiley: :tada: Tnx for testing that.

See [this](https://github.com/rakshasa/rtorrent/blob/62cb5a4605c0664bc522e0e0da9c72f09cf643a9/src/command_download.cc#L669). Could be it was hashing related too, or something else setting ignore_commands.
But I'm happy. I'll push the change.

Issue #423, opened 2013-10-28
Free space on Ubuntu doesn't include mounted shares or ZFS


SonarrBot:
Mounted tv share to /mnt/tv - free space is not shown on System Info

rfgamaral:
I just want to report that I've also had this problem but in my particular case, the problem was a symlinked folder. It seems Mono has problems with those too. I used the real absolute path instead of the symlinked one and this problem was gone.

c0unt0:
Hey all,
I’m seeing the same issue on an autofs mounted share on El Capitan: free space report tells me about “/“, but doesn’t include my mounted share, and import fails.

markus101:
Please open a thread on the forums with the requested info. You can also disable free space checking to allow the importing.

c0unt0:
Done, sorry for the misses of comments.

PR #1037, opened 2015-12-30
Parse /proc/mounts to get additional mountpoints on linux

Taloth:
This gives us a few neat features.
* More drives in System->Disk Space, including zfs. (fixes https://github.com/Sonarr/Sonarr/issues/423)
* Detection capability for CIFS mounts. (can't do a healthcheck without false positives)
* Force full transaction operation when move/copy to/from CIFS mounts, but not when moving inside it. (Better workaround for the partial-file issues)

Some concerns:
* Needs try catch, to prevent breaking way more important code such as updates.
* Half of the DiskTransferService becomes a candidate to refactor. Using the .partial~ is still nice, but strictly not needed.
* Won't work on FreeBSD (doesn't always have /proc/) but that should be okay.
* Can't always discern between system fs mounts (proc/tmp/run... etc) and real filesystems. stuff without nodev is easy, that's just a drive, /dev/ mounts too, but zfs is a nodev, so atm I keep a list of those special cases.

### DIFF
--- a/src/NzbDrone.Common/Disk/DiskProviderBase.cs
+++ b/src/NzbDrone.Common/Disk/DiskProviderBase.cs
@@ -346,12 +346,12 @@ namespace NzbDrone.Common.Disk
 
         public string[] GetFixedDrives()
         {
-            return (DriveInfo.GetDrives().Where(x => x.DriveType == DriveType.Fixed).Select(x => x.Name)).ToArray();
+            return GetMounts().Where(x => x.DriveType == DriveType.Fixed).Select(x => x.RootDirectory).ToArray();
         }
 
         public string GetVolumeLabel(string path)
         {
-            var driveInfo = DriveInfo.GetDrives().SingleOrDefault(d => d.Name == path);
+            var driveInfo = GetMounts().SingleOrDefault(d => d.RootDirectory.PathEquals(path));
 
             if (driveInfo == null)
             {
@@ -376,11 +376,28 @@ namespace NzbDrone.Common.Disk
             return new FileStream(path, FileMode.Create);
         }
 
-        public List<DriveInfo> GetDrives()
+        public virtual List<IMount> GetMounts()
+        {
+            return GetDriveInfoMounts();
+        }
+
+        public virtual IMount GetMount(string path)
+        {
+            var mounts = GetMounts();
+
+            return mounts.Where(drive => drive.RootDirectory.PathEquals(path) ||
+                                         drive.RootDirectory.IsParentPath(path))
+                      .OrderByDescending(drive => drive.RootDirectory.Length)
+                      .FirstOrDefault();
+        }
+
+        protected List<IMount> GetDriveInfoMounts()
         {
             return DriveInfo.GetDrives()
-                            .Where(d => d.DriveType == DriveType.Fixed || d.DriveType == DriveType.Network)
+                            .Where(d => d.DriveType == DriveType.Fixed || d.DriveType == DriveType.Network || d.DriveType == DriveType.Removable)
                             .Where(d => d.IsReady)
+                            .Select(d => new DriveInfoMount(d))
+                            .Cast<IMount>()
                             .ToList();
         }
 
--- a/src/NzbDrone.Common/Disk/DriveInfoMount.cs
+++ b/src/NzbDrone.Common/Disk/DriveInfoMount.cs
@@ -1,0 +1,76 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using NzbDrone.Common.Extensions;
+
+namespace NzbDrone.Common.Disk
+{
+    public class DriveInfoMount : IMount
+    {
+        private readonly DriveInfo _driveInfo;
+
+        public DriveInfoMount(DriveInfo driveInfo)
+        {
+            _driveInfo = driveInfo;
+        }
+
+        public long AvailableFreeSpace
+        {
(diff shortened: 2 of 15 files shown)


src/NzbDrone.Common/Disk/DiskProviderBase.cs
```diff
@@ -376,11 +376,28 @@ public FileStream OpenWriteStream(string path)
...
+                                         drive.RootDirectory.IsParentPath(path))
+                      .OrderByDescending(drive => drive.RootDirectory.Length)
+                      .FirstOrDefault();
+        }
+
+        protected List<IMount> GetDriveInfoMounts()
         {
             return DriveInfo.GetDrives()
-                            .Where(d => d.DriveType == DriveType.Fixed || d.DriveType == DriveType.Network)
+                            .Where(d => d.DriveType == DriveType.Fixed || d.DriveType == DriveType.Network || d.DriveType == DriveType.Removable)
```
markus101: Is this so we can see removal hard drives?

Taloth: Yup, wasn't sure if all USB drives always get listed as Fixed or Removable.

src/NzbDrone.Common/Disk/DriveInfoMount.cs
```diff
@@ -0,0 +1,76 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using NzbDrone.Common.Extensions;
+
+namespace NzbDrone.Common.Disk
+{
+    public class DriveInfoMount : IMount
```
kayone: don't you mean `DriveMountInfo`?

src/NzbDrone.Mono/ProcMountProvider.cs
```diff
@@ -0,0 +1,171 @@
...
+                var split = option.Split(new[] { '=' }, 1);
+
+                result.Add(split[0], split.Length == 2 ? split[1] : string.Empty);
+            }
+
+            return result;
+        }
+    }
+
+    public class ProcMount : IMount
```
kayone: different file :D

src/NzbDrone.Common/Disk/DriveInfoMount.cs
Taloth: No it's a wrapper around DriveInfo, implementing IMount. Like ProcMount wraps /proc.

src/NzbDrone.Core/DiskSpace/DiskSpaceService.cs
```diff
@@ -58,7 +58,7 @@ private IEnumerable<DiskSpace> GetDroneFactoryFreeSpace()
 
         private IEnumerable<DiskSpace> GetFixedDisksFreeSpace()
         {
-            return GetDiskSpace(_diskProvider.GetFixedDrives(), true);
+            return GetDiskSpace(_diskProvider.GetMounts().Where(d => d.DriveType == System.IO.DriveType.Fixed).Select(d => d.RootDirectory), true);
```
markus101: include `System.IO` and use `DriveType.Fixed`.

src/NzbDrone.Mono/ProcMountProvider.cs
```diff
@@ -0,0 +1,138 @@
...
+namespace NzbDrone.Mono
+{
+    public interface IProcMountProvider
+    {
+        List<IMount> GetMounts();
+    }
+
+    public class ProcMountProvider : IProcMountProvider
+    {
+        private static string[] _fixedTypes = "ext3,ext2,ext4,vfat,fuseblk,xfs,jfs,msdos,ntfs,minix,hfs,hfsplus,qnx4,ufs,btrfs".Split(',');
```
markus101: Torn on this, bit shorter than building it manually and less work, but it feels a bit odd still.

src/NzbDrone.Mono/ProcMountProvider.cs
```diff
@@ -0,0 +1,138 @@
...
+            return new ProcMount(driveType, name, mount, type, options);
+        }
+
+        private Dictionary<string, string> ParseOptions(string options)
+        {
+            var result = new Dictionary<string, string>();
+
+            foreach (var option in options.Split(','))
+            {
+                var split = option.Split(new[] { '=' }, 1);
```
markus101: This doesn't do anything, you'll always get the full string back.

src/NzbDrone.Mono/ProcMountProvider.cs
```diff
@@ -0,0 +1,138 @@
...
+
+                result.Add(split[0], split.Length == 2 ? split[1] : string.Empty);
```
markus101: And this always be `string.Empty` given we're not splitting anything above.

src/NzbDrone.Mono/DiskProvider.cs
```diff
@@ -116,22 +118,25 @@ public override void SetPermissions(string path, string mask, string user, strin
             }
         }
 
+        public override System.Collections.Generic.List<IMount> GetMounts()
+        {
+            return base.GetMounts()
+                .Concat(_procMountProvider.GetMounts())
```
markus101: nit: align on `.`

src/NzbDrone.Mono/ProcMountProvider.cs
Taloth: lol, dûh. Yeah.

PR #1046, opened 2016-01-05
Import/Rename extra files (Subtitles and other crap)

### LINKED ISSUE #459: Move Additional Files
Opt-in to move nfo, sfv, etc - we will need to keep the files named properly after a rename.

markus101:
I can't believe its 2016 and we're finally adding this. /s

I *think* I this is what we want, I opted for a giant table of all the things, since they are so closely related and only a couple columns are different. Some basic import testing has been done, but we'll need some eyes on testing it because its quite substantial, (you know, for a trivial feature /s).

New: Import subtitles and extra files when importing media files
New: Rename subtitles and extra files when renaming files
Closes #459

@kayone / @Taloth


duramato:
Are the file extensions hardcoded or an option to the user? If not it might move some funky files like html,exe or extra videos(like RARBG.mp4) etc

markus101:
@duramato its customizable, it looks at the episode filename (minus extension) and the extension of the extra file to determine if it should import.

`.tar` would be fine, but `.tar.gz` wouldn't work as the extension would be `.gz` and the filename wouldn't match. I'm not sure if that should actually be supported as it complicates things supporting double extensions.

src/NzbDrone.Core/Extras/ExistingExtraFileService.cs
```diff
@@ -0,0 +1,129 @@
...
+        {
+            var extraFiles = new List<ExtraFile>();
+
+            foreach (var possibleExtraFile in filesOnDisk)
+            {
+                var localEpisode = _parsingService.GetLocalEpisode(possibleExtraFile, series);
+
+                if (localEpisode == null)
+                {
+                    _logger.Debug("Unable to parse subtitle file: {0}", possibleExtraFile);
```
Caleb-C: How do you know that this file is a "subtitle" file?  Perhaps the message should read "extra" instead of "subtitle".

src/NzbDrone.Core/Extras/ExistingExtraFileService.cs
```diff
@@ -0,0 +1,129 @@
...
+
+                if (localEpisode.Episodes.Empty())
+                {
+                    _logger.Debug("Cannot find related episodes for: {0}", possibleExtraFile);
+                    continue;
+                }
+
+                if (localEpisode.Episodes.DistinctBy(e => e.EpisodeFileId).Count() > 1)
+                {
+                    _logger.Debug("Subtitle file: {0} does not match existing files.", possibleExtraFile);
```
Caleb-C: Also, may not be a "subtitle" file.

src/NzbDrone.Core/Extras/ExistingExtraFileService.cs
markus101: Good catch, pushing a fix now. Thanks.

src/NzbDrone.Core/Extras/ExtraFiles/ExtraFile.cs
```diff
@@ -0,0 +1,26 @@
+﻿using System;
+using NzbDrone.Core.Datastore;
+using NzbDrone.Core.Extras.Metadata;
+using NzbDrone.Core.Parser;
+
+namespace NzbDrone.Core.Extras.ExtraFiles
+{
+    public class ExtraFile : ModelBase
```
kayone: TBH, I think `MetaFile` or `MetaDataFile` is a better name that `ExtraFile`,

markus101: I'm not a huge fan of `ExtraFile`, but I wanted to separate Metadata from this file type as Metadata has been claimed by the metadata for other clients (Kodi, WDTV, etc).

src/NzbDrone.Core/Extras/ExistingExtraFileService.cs
```diff
@@ -0,0 +1,129 @@
...
+using NzbDrone.Core.Extras.Subtitles;
+using NzbDrone.Core.MediaFiles;
+using NzbDrone.Core.MediaFiles.Events;
+using NzbDrone.Core.Messaging.Events;
+using NzbDrone.Core.Parser;
+using NzbDrone.Core.Tv;
+
+namespace NzbDrone.Core.Extras
+{
+    public class ExistingExtraFileService : IHandle<SeriesScannedEvent>
```
kayone: This is really not a service, maybe 'ExtraFileImporter'

Ticstyle:
Awesome, really missed this from my manual DropIt setup that i used before i built my NAS/Media server

thezoggy:
on similar named files you should match name from left to right not right to left. that way you support extensions with multiple `.`. like `.en.srt` vs just doing `.srt`

thezoggy:
For associated files use length of the media file basename to determine the extension (approach from left).

This is needed for example on metadata like Sony PS3: filename.ext.cover.jpg
Also useful for multi-language subtitles: filename.en.srt, filename.nl.srt

to note in sb we allowed users to control over what associated files they did want to move (if at all)

`Move srr/srt/sfv/etc files with the episode when processed?`

`Filter associated files by specific non-media file extensions (separated by commas). (eg. .idx, .sub, .en.srt)`

markus101:
@kayone @Taloth refactored per our conversation.

markus101:
Some more thoughts came to mind.

Should the match be filename + extension exactly or would .sub include en.sub or gibberish.sub?
Use *.sub to match en.sub and fr.sub?
Should en.* allow all sub types for english?
Can I use * for everything?

thezoggy:
i say do it like we did in sb.. `.sub` matches *.sub so everything. if you want to limit you do `.en.sub`

markus101:
My one concern with that is you cant specify you want' `.en.sub` or `.sub` (unknown language), but not `fr.sub`, as always flexibility vs complexity.

thezoggy:
if you want french only subs you do fr.sub, if you want english you do en.sub, if you want all subs.. you do .sub?

Taloth:
Some questions and thoughts:
- Does it ever happen that one of these 'extra' files aren't prefixed with the same name as a video file?
- As mentioned, I don't like this being done on EpisodeImportEvent, since it requires the event to be processed synchronously to prevent the extras to be deleted too early.
- In my [original effort](https://github.com/Taloth/nzbdrone/tree/additional-files) I tried to define a 'FileSet' which contained a video filename and additional filenames. The idea was to pull the entire set through the import process and decide along the way which files we wanted. Eventually I stranded in the import process. Unfortunately back then CDH & torrent support were moving to the front of the queue, so I never got to look at it again.
- Just a thought, of what we might get in the future: Transfer Providers will require us to wrap the import process into an async multi-staged process, after all, the transfer could take a while, which means the import process should be async as well, and both visible on Activity Queue. Rar support would even generate something of a 'virtual file system' wrapper, to enable access to the rar content without copying it all over first. Both requires abstraction and the concept that analysis vs execution are two different things and potentially async.
That's not something we want to design yet, but it means that at one point in the future we can't just call ```diskScanService.GetVideoFiles```, coz the results will be wrapped in some logic. I'm not sure whether this affects how we can deal with extra/additional files, but it might.

The importer wants to perform a sequence of actions:
- Scan
- Decide
- Import
- Delete/cleanup

So everything fancy has to happen during those stages, particularly Import. Updating History could be async, since we provide all required data in the event. Of course we still want it executed synchronously to ensure some level of atomicity/consistency after the import finishes, but it isn't required.
Please note that some other 'async' stuff happens as well, such as updating the episode mediacenter metadata.

I like the separation in Metadata, Other, and Subtitle. Also like *Importer. What I don't like is how these are called.
It shouldn't be an event, it should be an integral part of the Import process.
```ExistingExtraFileService``` handling 'SeriesScannedEvent' is therefore wrong, you don't do 'extras' as an afterthought, coz series folder vs scan folder imports are basically the same, except that nothing is moved around in the series folder.
What you also need is a Scene/External*Importer. That way you can use IoC to get the appropriate Importers.
Of course you'll have a service somewhere that handles SeriesScannedEvent and refreshes the metadata. I imagine we would also have one that looks at which subs are required/monitored in the future.

I would expect the import to be handled as follows:
- Scan
- Decisions
- Delete existing (for upgrade only) (and also delete extra's? could be tricky with the nfo with the watched flag)
- Import new episode
- Import new extra's (for a series folder scan this includes Metadata, not just subs & 'other')
- raise EpisodeImportedEvent
- Delete src

Basically ```ImportApprovedEpisodes``` calling ```IImportAdditionalFiles.Something(ImportDecision, newDownload)``` (or IAdditionalFileImporter)
IImportAdditionalFiles would then invoke the appropriate importers using IoC depending on the newDownload flag.
I do wonder what we should do with the EpisodeImported/DownloadedEvent, should that include a list of 'extraFiles'? possibly.

I'll continue to look at the PR, but I had to write this down.

**Comment by markus101** [comment]
> Does it ever happen that one of these 'extra' files aren't prefixed with the same name as a video file?

Possibly, but I don't think we need to worry about this.

> As mentioned, I don't like this being done on EpisodeImportEvent, since it requires the event to be processed synchronously to prevent the extras to be deleted too early.

That can be moved, pretty simple to call this after importing the file.

Overall I think what you're thinking and what the code is doing (save for the event handling vs calling it explicitly) matches up pretty close. Currently Extra file importing of existing files is handled via the `SeriesScannedEvent`, which means we import all the files at once (same as we were doing for existing metadata files), though we can make importing new/existing extra files similar to episode file importing and use the same process (which makes the most sense).

File: src/NzbDrone.Core.Test/Housekeeping/Housekeepers/CleanupOrphanedMetadataFilesFixture.cs [hunk 5]
```diff
@@ -113,10 +113,10 @@ public void should_delete_episode_image_files_that_have_episodefileid_of_zero()
             Db.Insert(series);
 
             var metadataFile = Builder<MetadataFile>.CreateNew()
-                                                    .With(m => m.SeriesId = series.Id)
-                                                    .With(m => m.Type = MetadataType.EpisodeImage)
-                                                    .With(m => m.EpisodeFileId = 0)
-                                                    .BuildNew();
+                                                 .With(m => m.SeriesId = series.Id)
```
Taloth: Nitpick, but what's with these changes?

src/NzbDrone.Core.Test/ParserTests/LanguageParserFixture.cs
```diff
@@ -48,8 +48,17 @@ public class LanguageParserFixture : CoreTest
...
         {
-            var result = Parser.Parser.ParseTitle(postTitle);
-            result.Language.Should().Be(language);
+            var result = LanguageParser.ParseLanguage(postTitle);
+            result.Should().Be(language);
+        }
+
+        [TestCase("2 Broke Girls - S01E01 - Pilot.en.sub", Language.English)]
+        [TestCase("2 Broke Girls - S01E01 - Pilot.eng.sub", Language.English)]
+        [TestCase("2 Broke Girls - S01E01 - Pilot.sub", Language.Unknown)]
```
Taloth: FYI, .sub could be multiple sub streams. .idx file will tell. I do have a .idx .sub binary stream parser somewhere if it's required.

src/NzbDrone.Core/Extras/ExistingExtraFileService.cs
```diff
@@ -0,0 +1,70 @@
...
+            var extraFiles = new List<ExtraFile>();
+
+            if (!_diskProvider.FolderExists(series.Path))
+            {
+                return;
+            }
+
+            _logger.Debug("Looking for existing extra files in {0}", series.Path);
+
+            var filesOnDisk = _diskProvider.GetFiles(series.Path, SearchOption.AllDirectories);
```
Taloth: DiskScanService gets the whole list during it's scan. Somehow splitting that up is probably better coz it hits the filesystem only once. Although the impact should be minimal considering filesystem caches.

src/NzbDrone.Core/Extras/ExistingExtraFileService.cs
```diff
@@ -0,0 +1,70 @@
...
+            var possibleExtraFiles = filesOnDisk.Where(c => !MediaFileExtensions.Extensions.Contains(Path.GetExtension(c).ToLower()) && !c.StartsWith(Path.Combine(series.Path, "EXTRAS"))).ToList();
+
+            var filteredFiles = possibleExtraFiles;
+
+            foreach (var extraFileManager in _extraFileManagers)
+            {
+                filteredFiles = extraFileManager.FilterExistingFiles(series, possibleExtraFiles);
```
Taloth: Filtering Existing is not a very good idea, we also do it for importing the videos. I rather have a flag or an ImportDecision sorting that out so further along you can do associations more easily (and handle dupes).

markus101: Why is it not a good idea?

We do it for importing the episode files because it makes a significant difference in processing time, instead of re-parsing everything again we skip things we know about and sped up the subsequent scanning greatly.

src/NzbDrone.Core.Test/Housekeeping/Housekeepers/CleanupOrphanedMetadataFilesFixture.cs
markus101: Unintentional from previous changes.

markus101:
The more I think about it the more I like the idea of using a FileSet, it simplifies the disk scan and even the import process by passing a list of files back that already match cutting down matching files.

I think I'll end up making another PR for that to not bloat this one, unless there is any reason to hold off from doing so?

Taloth:
Well, it might conflict with series folder scans coz we filter out existing videos, preventing it from handling newly added extras. But it should be interesting to at least explore the option.

At one point I want to look over the entire import process. Coz atm, I feel like we have a lot of duplicate code. The recent parsing issue for Manual Import is a testament to that.

markus101:
Yeah, definitely something to look at. Good point about filtering existing files, its something I've wanted to improve on, but the speed improvements from skipping known files was massive.

Dediqated:
I would like to have this functionality. Sometimes something is downloaded with filenames that are hashed. There is a par2 file included with the original files. Sonarr deletes all files but the video file and therefore the par2 file gets deleted too. I wouldnt really mind if my Kodi install would index these hashed file names correctly but it doesn't. The video content doesn't show up. Also the case could be that subtitles are included externally as a .srt file. Then it gets deleted too. So please add the functionality, just as Sickbeard has, to keep all other files and folders of the download

valantislevas:
do we have any news about it? to moving subtitles (.srt) with mp4 to correct folder?

lukeroge:
Any chance we can get this one merged any time soon?

Taloth:
No, it isn't finished yet.

Prime-Omega:
Would love to see this feature included in Sonarr! Any updates on this commit?

src/NzbDrone.Core/Extras/Files/ExtraFileService.cs
```diff
@@ -0,0 +1,118 @@
...
+            {
+                var path = Path.Combine(series.Path, metadata.RelativePath);
+
+                if (_diskProvider.FileExists(path))
+                {
+                    _diskProvider.DeleteFile(path);
+                }
+            }
+
+            _logger.Debug("Deleting Metadata from database for episode file: {0}", episodeFile);
```
Taloth: Metadata => Extra

francisrath:
Is there anything that can be contributed to get this into the next version? I'm not into the codebase, but I'm handy with C#.

markus101:
We've decided to approach this a little different to avoid any issues. I'll be merging in the parsing aspect of this PR shortly, which will track the files in Sonarr's library, but will not import them for new downloads or rename them after existing media files are renamed. We want to ensure the parsing is working properly before adding importing and renaming.

valantislevas:
in which version is that option to see about the parsing the extra and srf files?
i have upgraded to  2.0.0.4409 i dont see something about it..

spooky69:
+1

Stibbons:
+1

PR #1067, opened 2016-01-17
Added theme-color meta for mobile chrome.

russjr08:
This will cause the Chrome header / status bar color for Android 5.0+ and Chrome 39+.

Also self-closed other meta tags for consistency (was already there for the http-equiv).

### DIFF
--- a/src/UI/index.html
+++ b/src/UI/index.html
@@ -6,7 +6,12 @@
     <meta name="viewport" content="width=device-width, initial-scale=1.0"/>
     <meta name="mobile-web-app-capable" content="yes"/>
     <meta name="apple-mobile-web-app-capable" content="yes"/>
+    
+    <!-- Used for Chrome, Opera, and Firefox OS -->
     <meta name="theme-color" content="#f5f5f5"/>
+    <!-- Windows Phone -->
+    <meta name="msapplication-navbutton-color" content="#f5f5f5">
+    
     <meta content="text/html;charset=utf-8" http-equiv="Content-Type"/>
 
     <link href="/Content/bootstrap.css" rel='stylesheet' type='text/css'/>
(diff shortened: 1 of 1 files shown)


thezoggy:
for those curious what this does: http://stackoverflow.com/questions/26960703/how-to-change-the-color-of-header-bar-and-address-bar-in-newest-android-chrome-v

and note that page has the solution if you want to do this for more than just mobile chrome.. 

```
<!-- Chrome, Firefox OS and Opera -->
<meta name="theme-color" content="#4285f4">
<!-- Windows Phone -->
<meta name="msapplication-navbutton-color" content="#4285f4">
<!-- iOS Safari -->
<meta name="apple-mobile-web-app-status-bar-style" content="#4285f4">
```

russjr08:
The safari tag doesn't actually support custom colors like that. 

https://developer.apple.com/library/iad/documentation/AppleApplications/Reference/SafariHTMLRef/Articles/MetaTags.html



I'll update for Windows phone when I get to my computer, wasn't aware they had that!

thezoggy:
about the meta tag not working for you:
`This meta tag has no effect unless you first specify full-screen mode as described in apple-apple-mobile-web-app-capable.`

markus101:
Is there a particular reason why we want to do this? It feels a bit like setting a custom cursor from the early 2000's, we can do it, but we probably shouldn't. Maybe I'm over thinking it.

Also, why #f5f5f5 as the colour?

russjr08:
I wouldn't compare it to a custom cursor. But rather it helps establish branding. Kind of like setting a favicon. 

Also, I picked white because it was the best I could think of. But obviously, that can be anything.

markus101:
Was taking a look at this and tested out the same blue we use for the logo:

image

Thoughts?

russjr08:
That looks great!

markus101:
I lost the vote we went for:

image

Turned out pretty good and less jarring than bright blue and dark grey. Merged it in with the new colour, thanks!

PR #1079, opened 2016-01-20
Download magnets from torcache when available

Taloth:
Basically just a placeholder for voting. There are other sites available to grab torrents from and even more options on the horizon.

Questions:
- Do we want this?
- Do we want to make it configurable by a dropdown?

If we implement more than just torcache, then we need a Factory pattern.


PartTimeLegend:
Is that a slackware torrent I see hiding in there?

Taloth:
We all download Linux distros right? So gotta use it as testcase two, for consistency sake.

PartTimeLegend:
@Taloth I thought the internet was just a lot of people downloading copies of Linux.
That said, it should be Arch.

src/NzbDrone.Core/Download/MagnetResolverService.cs
```diff
@@ -0,0 +1,76 @@
...
+using NzbDrone.Common.Http;
+
+namespace NzbDrone.Core.Download
+{
+    public interface IResolveMagnetLink
+    {
+        byte[] DownloadTorrentFromMagnet(string magnetUrl);
+    }
+
+    public class MagnetResolverService : IResolveMagnetLink
```
kayone: I think we are overusing the "service" postfix, something like this could just be called `MagnetLinkResolver`, `service` post fix should be used for entrypoint to main parts of the app, eg, `SeriesService`, `QualityService` etc, not helper components like this,

src/NzbDrone.Core/Download/MagnetResolverService.cs
```diff
@@ -0,0 +1,76 @@
...
+            _logger.Debug("Fetching torrent from torcache {0}", infoHash);
+
+            try
+            {
+                var response = _httpClient.Get(request);
+
+                if (response.Headers.ContentType != "application/x-bittorrent")
+                {
+                    _logger.Debug("Torcache returned unexpected ContentType: {0}", response.Headers.ContentType);
+                    return null;
```
kayone: `new byte[]()` instead of `null`?

src/NzbDrone.Core/Download/MagnetResolverService.cs
```diff
@@ -0,0 +1,76 @@
...
+                    tiers.Add(magnet.AnnounceUrls);
+
+                    data = dictionary.Encode();
+                }
+
+                return data;
+            }
+            catch (Exception ex)
+            {
+                _logger.Debug("Failed to fetch torrent from torcache, using magnet instead. {0}", ex.Message);
```
kayone: `using magnet instead.` I don't think it's responsibility of this class to make a assumption like this.

neox387:
What's going on with this, would like to have to torcache support or any method of saving torrent file in black hole :)

neox387:
still nothing ? :(

neox387:
Any solution yet ? now that kat is down no way to get rarbg .torrent files :(

PR #1095, opened 2016-01-27
Fix: xbuild doesn't support /m parameter

ta264:
Passing the /m parameter to xbuild causes an error.  See:
http://stackoverflow.com/questions/32220678/xbuild-reporting-too-many-project-files-but-works-with-msbuild


markus101:
Thanks!

Issue #641, opened 2015-06-25
Pushbullet Channel Support


M0UL:
add support for pushbullet channels, for easy sort of notications like coachpato does.

markus101:
What benefit would this add to Sonarr? What do you mean by "easy soft of notifications"?

M0UL:
the way in currently is that sonarr notification get placed in the everything section, if it supports channels then it gets its own section.  see screenshot
http://prntscr.com/7limbn

halkeye:
are pushbullet channels not super public? IS there a way to lock them down at all now?

markus101:
If its not featured you need to know the tag for the channel, not secure, but also not public. It'd be awesome to be able to make them private, but I'm not sure if they are planning to do that (I will ask though).

markus101:
Looks like the channel tag will accept dashes, so its pretty painless to create a unique channel using a GUID.

PR #1100, opened 2016-01-28
Pushbullet: Source device support

IvanBrazza:
Pushbullet allows you to create a new device via the API. This new field allows the user to specify an ID of a device to send from. Similar to #641, but instead of public channels, you get a private device to organise pushes from Sonarr.

image

image

### DIFF
--- a/src/NzbDrone.Core/Notifications/PushBullet/PushBulletProxy.cs
+++ b/src/NzbDrone.Core/Notifications/PushBullet/PushBulletProxy.cs
@@ -5,6 +5,7 @@ using FluentValidation.Results;
 using NLog;
 using RestSharp;
 using NzbDrone.Core.Rest;
+using NzbDrone.Common.Extensions;
 
 namespace NzbDrone.Core.Notifications.PushBullet
 {
@@ -150,6 +151,11 @@ namespace NzbDrone.Core.Notifications.PushBullet
                 request.AddParameter("type", "note");
                 request.AddParameter("title", title);
                 request.AddParameter("body", message);
+
+                if (settings.SenderId.IsNotNullOrWhiteSpace())
+                {
+                    request.AddParameter("source_device_iden", settings.SenderId);
+                }
 
                 client.Authenticator = new HttpBasicAuthenticator(settings.ApiKey, string.Empty);
                 client.ExecuteAndValidate(request);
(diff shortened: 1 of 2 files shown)


src/NzbDrone.Core/Notifications/PushBullet/PushBulletProxy.cs
```diff
@@ -151,6 +151,11 @@ private void SendNotification(string title, string message, RestRequest request,
                 request.AddParameter("title", title);
                 request.AddParameter("body", message);
 
+                if (!String.IsNullOrEmpty(settings.SenderId))
```
markus101: Use `if (settings.SenderId.IsNotNullOrEmpty())`

IvanBrazza: That method doesn't exist

image

markus101: Its a string extension we've added, coming from https://github.com/Sonarr/Sonarr/blob/develop/src/NzbDrone.Common/Extensions/StringExtensions.cs#L76

Include `using NzbDrone.Common.Extensions` at the top if it doesn't automatically suggest it.

IvanBrazza: Ah cool! That's handy. Done :)

markus101:
Awesome, forgot to mention it before, but if you can squash the two commits into one we can get this merged.

A good guide on this is:

http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html

`git rebase -i HEAD~2` well get the last two commits.

Set the commit message to `New: Setting Pushbullet source device` and it will get mentioned in the changelog.

IvanBrazza:
Hmm, followed through that and I can't push back now because the local branch isn't in sync with the remote (the guide also says not to do it on already pushed commits). I'm no expert on git... Should I make a new PR or is there other git wizardry that you could suggest?

markus101:
You need to force push the changes, `git push --force` will do it.

IvanBrazza:
Oh fab. Done. Thanks :)

markus101:
Thanks @IvanBrazza

Issue #732, opened 2015-08-06
Override global proxy to anonymize outbound connections to trackers/usenet-indexers.


mescon:
#### The goal
1. Anonymize downloads of .torrent files and .nzb files
2. Retain communication outside of the proxy.

#### The problem
* Currently, the only way to make Sonarr contact external resources (trackers and usenet sites) anonymously is to set a global proxy on your system - which defeats goal 2 (all traffic originating on the host where Sonarr is installed will use the global proxy).

* I learned that PAC files (https://en.wikipedia.org/wiki/Proxy_auto-config) can be configured to return a proxy to the requesting client, depending on the requested url and/or domain. However, for PAC files to function, the requesting program need to be able to parse the javascript instructions, i.e. not feasible for Sonarr.

* Another (possible) alternative is to go upstream to your router and configure fairly complex routing, using a VPN link to an offsite gateway which you trust (a dedicated seedbox for example). A user might not want/need to route ALL traffic from the host Sonarr is installed on via the VPN, which then complicates things further. Doing multi-homed WAN setup might be possible provided that .NET respects the default gateway of the secondary adapter it is bound to. In my own setup, I have done this by configuring a secondary network adapter on the host where Sonarr is installed and then mitigating "hot potato routing"-problems as described here: http://www.rjsystems.nl/en/2100-adv-routing.php.
This is probably not feasible for the vast majority of Sonarr users, and I'm not sure it fully works.

#### The (proposed) solution
I was thinking it should be easy enough to enable WebRequest (used by Sonarr afaik) to use a proxy to communicate when it initiates an outbound request to the Internet.
See: https://msdn.microsoft.com/en-us/library/fze2ytx2(v=vs.110).aspx.

Also, see:
[How to: Enable a WebRequest to Use a Proxy to Communicate With the Internet](https://msdn.microsoft.com/en-us/library/2s8th0s1(v=vs.110).aspx)
[How to: Override a Global Proxy Selection](https://msdn.microsoft.com/en-us/library/8zxx9wxb(v=vs.110).aspx)

Overriding global proxy selection is important in order to not route the whole system via a proxy.
The Sonarr UI/configuration/general (advanced) would need another input box where the user could enter proxy server and port. Some SOCKS-proxies might require username/password authentication but other proxies such as Polipo (http proxy) are authenticated via IP-number.

#### Finally...
[...]

PR #1115, opened 2016-02-01
Proxy Support #732

lloydsparkes:
As per the discussions on #732 attached is my pull request.

Still to do:

- [x] CurlSharp Integration
- [x] HealthCheck


markus101:
Haven't looked at the code, but the merge commits from `Sonarr:develop` will need to be removed, rebasing off develop instead of merging will prevent these. Probably a variety of ways to fix this, a simple one I've used before is:

- Create a new branch based off the `proxy-support` branch
- reset `proxy-support` to match Sonarr:develop
- Cherry pick the changes
- Force push the changes back to github

lloydsparkes:
It automatically closed it when there was no commits. Thanks for the instructions on how to do it. 

Also caught a few unneeded changes

.gitmodules
```diff
@@ -2,3 +2,6 @@
 	path = src/ExternalModules/CurlSharp
 	url = https://github.com/Sonarr/CurlSharp.git
 	branch = master
+[submodule "src/ExternalModules/SocksWebProxy"]
+	path = src/ExternalModules/SocksWebProxy
```
kayone: please avoid submodules, we will be removing other submodules as well, 

is there a reason we can't use https://www.nuget.org/packages/SocksWebProxy/?

src/NzbDrone.sln
```diff
@@ -90,192 +90,361 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "LogentriesNLog", "Logentrie
...
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
+		Debug|Any CPU = Debug|Any CPU
 		Debug|x86 = Debug|x86
+		Mono|Any CPU = Mono|Any CPU
 		Mono|x86 = Mono|x86
+		Release|Any CPU = Release|Any CPU
 		Release|x86 = Release|x86
 	EndGlobalSection
 	GlobalSection(ProjectConfigurationPlatforms) = postSolution
+		{FAFB5948-A222-4CF6-AD14-026BE7564802}.Debug|Any CPU.ActiveCfg = Debug|x86
```
kayone: We aren't planning to support Any CPU since it makes shipping native binaries for x64, more complicated/error prone.

src/UI/Settings/General/GeneralViewTemplate.hbs
```diff
@@ -162,6 +162,70 @@
             </div>
         </div>
     </fieldset>
+
```
kayone: please follow the same indentations,

.gitmodules
markus101: We should probably discuss this more.

src/NzbDrone.Common/Http/Dispatchers/ManagedHttpDispatcher.cs
```diff
@@ -23,6 +25,45 @@ public HttpResponse GetResponse(HttpRequest request, CookieContainer cookies)
             webRequest.ContentLength = 0;
             webRequest.CookieContainer = cookies;
 
+            if (request.Proxy != null)
+            {
+                IPAddress[] addresses = Dns.GetHostAddresses(request.Proxy.Host);
+                switch (request.Proxy.Type)
+                {
+                    case ProxyType.Http:
+                        if(request.Proxy.Username != null && request.Proxy.Password != null)
```
kayone: please extract `request.Proxy.Username != null && request.Proxy.Password != null` into a local varible something like `authenticated` to remove the duplicated checks.

also check for null as well as empty strings.

src/NzbDrone.Core/Configuration/ConfigFileProvider.cs
```diff
@@ -216,6 +223,36 @@ public string UrlBase
...
+        }
+
+        public ProxyType ProxyType
+        {
+            get { return GetValueEnum<ProxyType>("ProxyType", ProxyType.Http); }
+        }
+
+        public string ProxyHostname
+        {
+            get { return GetValue("ProxyHostname", "localhost"); }
```
kayone: I think `127.0.0.1` is considered safer and more reliable.

src/NzbDrone.Core/Http/ProxyHttpInterceptor.cs
```diff
@@ -0,0 +1,39 @@
...
+
+        public HttpRequest PreRequest(HttpRequest request)
+        {
+            if(_configService.ProxyEnabled == UseProxy.Global ||
+                (_configService.ProxyEnabled == UseProxy.IndexersOnly && request.IsIndexerRequest))
+            {
+                request.Proxy = new HttpRequestProxySettings(_configService.ProxyType,
+                    _configService.ProxyHostname,
+                    _configService.ProxyPort,
+                    _configService.ProxyUsername == string.Empty ? null : _configService.ProxyUsername,
```
kayone: there are times that you check against `string.empty` and times you check against `null` please use `String.IsNullOrWhiteSpace`

src/NzbDrone.Core/Configuration/ConfigFileProvider.cs
markus101: We use `localhost` in most other cases for a default, but in this case I don't think we should specify a default at all.

src/UI/Settings/General/GeneralViewTemplate.hbs
```diff
@@ -162,6 +162,70 @@
...
     </fieldset>
+	
+	<fieldset>
+        <legend>Proxy Settings</legend>
+		
+		<div class="form-group">
+            <label class="col-sm-3 control-label">Use Proxy</label>
+
+			<div class="col-sm-1 col-sm-push-4 help-inline">
+                <i class="icon-sonarr-form-warning" title="Requires restart to take effect"/>
```
kayone: I don't see anything in the code that make it seem it actually requires a restart, unless I'm missing something.

src/UI/Settings/SettingsLayout.js
```diff
@@ -240,6 +243,7 @@ module.exports = Marionette.Layout.extend({
     },
 
     _toggleAdvancedSettings : function() {
+		console.log("Advance Toggled")
```
markus101: Indentation here as well, use spaces, not tabs.

src/UI/Settings/SettingsLayout.js
```diff
@@ -248,5 +252,24 @@ module.exports = Marionette.Layout.extend({
...
+		console.log("Proxy Toggled")
+        var value = $('.x-proxy-settings').val();
+		Config.setValue(Config.Keys.ProxySettings, value);
+        if (value === 'off') {
+            $('body').removeClass('show-proxy-settings');
+        } else {
+            $('body').addClass('show-proxy-settings');
+        }
     }
 });
```
markus101: New line at end of file.

src/NzbDrone.Core/Http/UseProxyEnum.cs
```diff
@@ -0,0 +1,15 @@
...
+
+namespace NzbDrone.Core.Http
+{
+    public enum UseProxy
+    {
+        Off,
+        Global,
+        IndexersOnly
+    }
+
```
markus101: Extra new line.

src/NzbDrone.Core/Configuration/ConfigFileProvider.cs
markus101: Do we want to store this in the config file or config DB? If we plan to use it in Common I think it needs to live in the file, but the file is less flexible.

src/NzbDrone.Common/Http/HttpRequest.cs
```diff
@@ -16,6 +16,8 @@ public HttpRequest(string url, HttpAccept httpAccept = null)
             _segments = new Dictionary<string, string>();
             AllowAutoRedirect = true;
             Cookies = new Dictionary<string, string>();
+            IsIndexerRequest = false;
```
markus101: HttpRequest should not know about something that uses it, `IsIndexerRequest` should be `UseProxyForRequest` (or something more generic).

src/NzbDrone.Common/Http/Dispatchers/ManagedHttpDispatcher.cs
```diff
@@ -23,6 +25,45 @@ public HttpResponse GetResponse(HttpRequest request, CookieContainer cookies)
             webRequest.ContentLength = 0;
             webRequest.CookieContainer = cookies;
 
+            if (request.Proxy != null)
+            {
+                IPAddress[] addresses = Dns.GetHostAddresses(request.Proxy.Host);
```
markus101: `var address`, new line after assignment.

.gitmodules
lloydsparkes: https://www.nuget.org/packages/SocksWebProxy/ is .Net 4.5. 

I forked it to make a .Net 4 version (it doesn't have any 4.5 code, but it breaks, as Sonarr is 4.0)

src/UI/Settings/settings.less
```diff
@@ -100,6 +100,24 @@ li.save-and-add:hover {
   }
 }
 
+.proxy-setting {
+  display: none;
+
+  .control-label {
+    color: @brand-warning;
```
markus101: No need to set the color, just need to set them to advanced and it happens automatically.

src/UI/Settings/SettingsLayout.js
```diff
@@ -248,5 +252,24 @@ module.exports = Marionette.Layout.extend({
...
+
+        if (value !== 'off') {
+            $('body').addClass('show-proxy-settings');
+        }
+    },
+	
+	_toggleProxySettings : function() {
+		console.log("Proxy Toggled")
+        var value = $('.x-proxy-settings').val();
+		Config.setValue(Config.Keys.ProxySettings, value);
```
markus101: Indentation.

src/NzbDrone.Core/Configuration/ConfigFileProvider.cs
lloydsparkes: We are not getting at the config in Common (its in Core via ProxyHttpInterceptor.cs). 

I did originally do it all via the ConfigService/the db. But I moved it to the ConfigFileService because everything else on that  generals settings screen used that one, and I couldn't work out how to split the settings into two so they get persisted to different configuration services.

markus101: It would just need a call to `ConfigService.SaveConfigDictionary), as we do for the ConfigFileService: https://github.com/Sonarr/Sonarr/blob/develop/src/NzbDrone.Api/Config/HostConfigModule.cs#L77

If the service doesn't handle that config value it just ignores it.

Probably best to move it to the ConfigService and keep as little as possible (just what we need during startup/to let a user disable auth) in the config file.

lloydsparkes: Sure will move it

src/NzbDrone.sln
```diff
@@ -270,12 +274,12 @@ Global
...
-		{74420A79-CC16-442C-8B1E-7C1B913844F0}.Mono|x86.ActiveCfg = Release|Any CPU
-		{74420A79-CC16-442C-8B1E-7C1B913844F0}.Mono|x86.Build.0 = Release|Any CPU
-		{74420A79-CC16-442C-8B1E-7C1B913844F0}.Release|x86.ActiveCfg = Release|Any CPU
-		{74420A79-CC16-442C-8B1E-7C1B913844F0}.Release|x86.Build.0 = Release|Any CPU
+        {74420A79-CC16-442C-8B1E-7C1B913844F0}.Debug|x86.ActiveCfg = Debug|Any CPU
+        {74420A79-CC16-442C-8B1E-7C1B913844F0}.Debug|x86.Build.0 = Debug|Any CPU
+        {74420A79-CC16-442C-8B1E-7C1B913844F0}.Mono|x86.ActiveCfg = Release|Any CPU
+        {74420A79-CC16-442C-8B1E-7C1B913844F0}.Mono|x86.Build.0 = Release|Any CPU
+        {74420A79-CC16-442C-8B1E-7C1B913844F0}.Release|x86.ActiveCfg = Release|Any CPU
+        {74420A79-CC16-442C-8B1E-7C1B913844F0}.Release|x86.Build.0 = Release|Any CPU
```
kayone: this seems like indentation change only, can you please just revert them.

src/UI/Settings/SettingsLayout.js
```diff
@@ -248,5 +252,24 @@ module.exports = Marionette.Layout.extend({
...
+            $('body').addClass('show-proxy-settings');
+        }
+    },
+    
+    _toggleProxySettings : function() {
+        console.log("Proxy Toggled")
+        var value = $('.x-proxy-settings').val();
+        Config.setValue(Config.Keys.ProxySettings, value);
+        if (value === 'off') {
+            $('body').removeClass('show-proxy-settings');
```
kayone: I don't think you actually need to all the way up in the root of the tree, you could just add it to the current view. `this.$el.removeClass`

src/UI/Settings/SettingsLayout.js
```diff
@@ -240,6 +243,7 @@ module.exports = Marionette.Layout.extend({
     },
 
     _toggleAdvancedSettings : function() {
+        console.log("Advance Toggled")
```
kayone: I would suggest moving this down after you have the value and log the value as well,

`console.log("Advance Toggled. ", checked)`

src/UI/Settings/SettingsLayout.js
```diff
@@ -248,5 +252,24 @@ module.exports = Marionette.Layout.extend({
...
+    _setProxySettingsState : function() {
+        var value = Config.getValue(Config.Keys.ProxySettings);
+
+        if (value !== 'off') {
+            $('body').addClass('show-proxy-settings');
+        }
+    },
+    
+    _toggleProxySettings : function() {
+        console.log("Proxy Toggled")
```
kayone: same as above.

src/UI/Settings/SettingsLayout.js
```diff
@@ -49,7 +49,8 @@ module.exports = Marionette.Layout.extend({
         metadataTab        : '.x-metadata-tab',
         generalTab         : '.x-general-tab',
         uiTab              : '.x-ui-tab',
-        advancedSettings   : '.x-advanced-settings'
+        advancedSettings   : '.x-advanced-settings',
+        proxySettings       : '.x-proxy-settings'
```
kayone: this should be in the general view not in the main layout.

src/UI/Settings/SettingsLayout.js
markus101: Looks like this was for debugging more than anything.

src/UI/Settings/SettingsLayout.js
markus101: Don't need to set a class at all, take a look at how we're doing it for auth:

https://github.com/Sonarr/Sonarr/blob/develop/src/UI/Settings/General/GeneralView.js#L60

Advanced settings applies to all settings pages (and edit modals) so we used a class.

src/UI/Settings/settings.less
```diff
@@ -100,6 +100,20 @@ li.save-and-add:hover {
   }
 }
 
+.proxy-setting {
```
markus101: I don't think any of this CSS will be needed.

lloydsparkes:
I will look at publishing my own nuget for that / talking to the owner to see if he has any interest in such a thing, to resolve the submodule issue. 

And I still need to do the healthcheck and finish the testing against various configurations.

lloydsparkes:
I've tested as many configurations as possible, basic HTTP, and Socks. They all seem fine.

lloydsparkes:
So I've moved the Submodule to Nuget Packages and updated everything from that. Hopefully that resolves everything outstanding

src/NzbDrone.Common/Http/Dispatchers/ManagedHttpDispatcher.cs
```diff
@@ -1,11 +1,29 @@
...
 namespace NzbDrone.Common.Http.Dispatchers
 {
     public class ManagedHttpDispatcher : IHttpDispatcher
     {
+        /// <summary>
+        /// This method returns a free port. Now of course after we find it, it could be opened by something else. But it seems in practice this is not the case.
+        /// 
+        /// The OS will cycle through all free ports randomly in response to this, so its unlikely to happen.
+        /// </summary>
+        /// <returns></returns>
```
markus101: These XML comments aren't required.

src/NzbDrone.Common/Http/Dispatchers/ManagedHttpDispatcher.cs
```diff
@@ -1,11 +1,29 @@
...
+        private static int GetNextFreePort()
```
markus101: private methods should go at the bottom of the file.

src/NzbDrone.Common/Http/Dispatchers/ManagedHttpDispatcher.cs
```diff
@@ -1,11 +1,29 @@
...
+        {
+            TcpListener l = new TcpListener(IPAddress.Loopback, 0);
```
markus101: `var` instead of `TcpListener`. Also use a better variable name than `l`, such as `listener`.

src/NzbDrone.Common/Http/Dispatchers/ManagedHttpDispatcher.cs
```diff
@@ -1,11 +1,29 @@
...
+            l.Start();
+            int port = ((IPEndPoint)l.LocalEndpoint).Port;
```
markus101: `var` instead of `int`

src/NzbDrone.Common/Http/Dispatchers/ManagedHttpDispatcher.cs
```diff
@@ -1,11 +1,29 @@
...
+            l.Stop();
+            return port;
```
markus101: New line before the return to space it out a bit.

src/NzbDrone.Common/Http/Dispatchers/ManagedHttpDispatcher.cs
```diff
@@ -23,6 +41,34 @@ public HttpResponse GetResponse(HttpRequest request, CookieContainer cookies)
...
+                        }
+                        break;
+                    case ProxyType.Socks4:
+                        webRequest.Proxy = new SocksWebProxy(new ProxyConfig(IPAddress.Parse("127.0.0.1"), GetNextFreePort(), addresses[0], request.Proxy.Port, ProxyConfig.SocksVersion.Four, socksUsername, socksPassword), false);
+                        break;
+                    case ProxyType.Socks5:
+                        webRequest.Proxy = new SocksWebProxy(new ProxyConfig(IPAddress.Parse("127.0.0.1"), GetNextFreePort(), addresses[0], request.Proxy.Port, ProxyConfig.SocksVersion.Five, socksUsername, socksPassword), false);
+                        break;
+                }
+
```
markus101: Extra new line not required.

src/NzbDrone.Common/Http/Dispatchers/ManagedHttpDispatcher.cs
```diff
@@ -23,6 +41,34 @@ public HttpResponse GetResponse(HttpRequest request, CookieContainer cookies)
             webRequest.ContentLength = 0;
             webRequest.CookieContainer = cookies;
 
+            if (request.Proxy != null)
+            {
+                var addresses = Dns.GetHostAddresses(request.Proxy.Host);
+                string socksUsername = request.Proxy.Username == null ? "" : request.Proxy.Username;
```
markus101: `varsocksUsername = request.Proxy.Username.IsNotNullOrEmpty() ? string.Empty : request.Proxy.Username;`

src/NzbDrone.Common/Http/Dispatchers/ManagedHttpDispatcher.cs
```diff
@@ -23,6 +41,34 @@ public HttpResponse GetResponse(HttpRequest request, CookieContainer cookies)
...
+                string socksPassword = request.Proxy.Password == null ? "" : request.Proxy.Password;
```
markus101: Same as line above

src/NzbDrone.Common/Http/Dispatchers/ManagedHttpDispatcher.cs
```diff
@@ -23,6 +41,34 @@ public HttpResponse GetResponse(HttpRequest request, CookieContainer cookies)
...
+
+                switch (request.Proxy.Type)
+                {
+                    case ProxyType.Http:
+                        if(!string.IsNullOrWhiteSpace(request.Proxy.Username) && !string.IsNullOrWhiteSpace(request.Proxy.Password))
```
markus101: `if (request.Proxy.Username.IsNotNullOrWhiteSpace() && request.Proxy.Password.IsNotNullOrWhiteSpace())`

src/NzbDrone.Common/Http/HttpRequest.cs
```diff
@@ -16,6 +16,10 @@ public HttpRequest(string url, HttpAccept httpAccept = null)
             _segments = new Dictionary<string, string>();
             AllowAutoRedirect = true;
             Cookies = new Dictionary<string, string>();
+            
+            //For when the Proxy is set, but it is not a global proxy, this request to use it (I.e Indexers)
+            UseProxyIfAvailable = false;
```
markus101: False is the default for a boolean, you don't need to explicitly set it.

src/NzbDrone.Common/Http/HttpRequest.cs
```diff
@@ -16,6 +16,10 @@ public HttpRequest(string url, HttpAccept httpAccept = null)
...
+            Proxy = null;
```
markus101: Same here, null is the default, you don't need to set it.

src/NzbDrone.Common/Http/HttpRequestProxySettings.cs
```diff
@@ -0,0 +1,32 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace NzbDrone.Common.Http
+{
+    public enum ProxyType
```
markus101: Move to its own file.

src/NzbDrone.Core/HealthCheck/Checks/ProxyCheck.cs
```diff
@@ -0,0 +1,59 @@
...
+
+                var request = new HttpRequestBuilder("https://sonarr.tv").Build("/");
+                request.UseProxyIfAvailable = true;
+
+                try
+                {
+                    var response = _client.Execute(request);
+
+                    //400 Is bad Request so something is wrong. But CloudFlare causes 403 etc which are fine. The Server could have a Programming issue, so 500. So we only really want to catch the 400
+                    if ((int)response.StatusCode == 400)
```
markus101: `if (response.StatusCode == HttpStatusCode.BadRequest)`

src/NzbDrone.Core/HealthCheck/Checks/ProxyCheck.cs
```diff
@@ -0,0 +1,59 @@
...
+                }
+
+                var request = new HttpRequestBuilder("https://sonarr.tv").Build("/");
+                request.UseProxyIfAvailable = true;
+
+                try
+                {
+                    var response = _client.Execute(request);
+
+                    //400 Is bad Request so something is wrong. But CloudFlare causes 403 etc which are fine. The Server could have a Programming issue, so 500. So we only really want to catch the 400
```
markus101: `// We only care about 400 responses, other error codes can be ignored`

src/NzbDrone.Core/HealthCheck/Checks/ProxyCheck.cs
```diff
@@ -0,0 +1,59 @@
...
+                    if ((int)response.StatusCode == 400)
+                    {
+                        _logger.Error("Proxy Health Check failed because we got a 400 Response Code: " + response.StatusCode.ToString() + " Response Data: " + response.ResponseData);
```
markus101: `_logger.Error("Proxy Health Check Failed: {0}. Response data: {1}", response.StatusCode.ToString(), response.ResponseData)`

src/NzbDrone.Core/HealthCheck/Checks/ProxyCheck.cs
```diff
@@ -0,0 +1,59 @@
...
+                        return new HealthCheck(GetType(), HealthCheckResult.Error, "Failed to load https://sonarr.tv/, got HTTP " + response.StatusCode.ToString());
+                    }
+                }
+                catch (Exception ex)
+                {
+                    _logger.ErrorException("Proxy Health Check failed due to the following exception", ex);
```
markus101: `Proxy health check failed.`

src/NzbDrone.Core/HealthCheck/Checks/ProxyCheck.cs
```diff
@@ -0,0 +1,59 @@
...
+                    return new HealthCheck(GetType(), HealthCheckResult.Error, "An exception was recieved while trying to load https://sonarr.tv/: " + ex.Message);
```
markus101: "occurred" instead of "was received".

src/NzbDrone.Core/Http/ProxyHttpInterceptor.cs
```diff
@@ -0,0 +1,39 @@
...
+        public HttpRequest PreRequest(HttpRequest request)
+        {
+            if(_configService.ProxyEnabled == UseProxy.Global ||
+                (_configService.ProxyEnabled == UseProxy.IndexersOnly && request.UseProxyIfAvailable))
+            {
+                request.Proxy = new HttpRequestProxySettings(_configService.ProxyType,
+                    _configService.ProxyHostname,
+                    _configService.ProxyPort,
+                    string.IsNullOrWhiteSpace(_configService.ProxyUsername) ? null : _configService.ProxyUsername,
+                    string.IsNullOrWhiteSpace(_configService.ProxyPassword) ? null : _configService.ProxyPassword);
```
markus101: Same here.

src/UI/Settings/General/GeneralViewTemplate.hbs
```diff
@@ -162,6 +162,68 @@
...
+                <div class="col-sm-4">
+                    <input type="text" placeholder="localhost" name="proxyHostname" class="form-control"/>
+                </div>
+            </div>
+
+            <div class="form-group">
+                <label class="col-sm-3 control-label">Port</label>
+
+                <div class="col-sm-4">
+                    <input type="number" placeholder="9150" name="proxyPort" class="form-control"/>
```
markus101: The default port is 8080, but the placeholder is 9150.

src/UI/Settings/General/GeneralViewTemplate.hbs
```diff
@@ -162,6 +162,68 @@
...
+                </div>
+            </div>
+
+            <div class="form-group">
+                <label class="col-sm-3 control-label">Username</label>
+
+                <div class="col-sm-4">
+                    <input type="text" placeholder="optional" name="proxyUsername" class="form-control"/>
```
markus101: Use help text instead of a placeholder.

https://github.com/Sonarr/Sonarr/blob/develop/src/UI/Settings/General/GeneralViewTemplate.hbs#L102-L104

src/UI/Settings/General/GeneralViewTemplate.hbs
```diff
@@ -162,6 +162,68 @@
...
+                </div>
+            </div>
+
+            <div class="form-group">
+                <label class="col-sm-3 control-label">Password</label>
+
+                <div class="col-sm-4">
+                    <input type="password" placeholder="optional" name="proxyPassword" class="form-control"/>
```
markus101: Same here

Taloth:
It's been a while, but I've been planning to look over and merge in as much PRs after the master release, which happened last weekend.

Generally I have very little to add to this, mostly nitpicks.

But one thing stands out and that's how it's tied into HttpRequest.
One thing we didn't exactly get clear on is when the proxy will be used.

It's mostly of concern to indexers, but this isn't enough. coz most/all indexers have their nzb/torrent on some link... you want that proxied too.
So personally I see very little benefit in a 'global'/'indexer only' switch.
I rather go for clarity and go for a global on/off switch. That way you don't have to handle all those special cases where you're downloading nzb/torrents.

Secondly, how to deal with download clients? we're using HttpClient for those too. I didn't see any code that deals with localhost, local subnet filtering.

lloydsparkes:
So looking at your comment, and thinking about it again after looking at all the network requests. I agree.

So I will remove the "Indexer Only" switch and make it a ON / OFF switch

Secondly on the address filtering for local address. I will add a textbox, that takes a comma separated list of domains to avoid the proxy for.

E.g.

'localhost,google.com,*.home'

How does that sound?

Taloth:
Sounds good, can you also check the XMLRPC wrapper that's used for rtorrent?

lloydsparkes:
Sure will do

lloydsparkes:
So I have done all of the requested changes.

I have checked the XMLRPC wrapper that is used for rtorrent. I have also checked all of the other download clients.

It seems they utilise either the XMLRPC client, or the RestSharp RestClient. 

These are not affected by the Proxy code, as they do not seem to utilise any of the Http request/response stack that the rest of the Sonarr code seems to utilise. (TorrentClientBase and UsenetClientBase have a IHttpClient, but it doesn't actually seem to be used)

It should be possible to fairly easily attach the Proxy to these requests if we really wanted to. Its just about passing the correct settings down to them. But it would need to be done for each download client separately.

I suspect though that there is little need for the connection to these download clients to utilise any proxy, given they are all likely to be on the localhost / local network, where you would bypass the proxy anyway. 

Do you want them modified to utilise the proxy?

src/NzbDrone.Core/Configuration/ConfigService.cs
```diff
@@ -344,6 +344,16 @@ public string ProxyPassword
...
         }
 
+        public string ProxySubnetFilter
+        {
+            get { return GetValue("ProxySubnetFilter", string.Empty); }
+        }
+
+        public bool ProxyBypassLocalAddresses
+        {
+            get { return GetValueBoolean("ProxyPassword", true); }
```
Taloth: Take a really close look here.

src/NzbDrone.Core/HealthCheck/Checks/ProxyCheck.cs
```diff
@@ -33,7 +33,6 @@ public override HealthCheck Check()
                 }
 
                 var request = new HttpRequestBuilder("https://sonarr.tv").Build("/");
```
Taloth: @markus101  maybe idea to replace this with `http(s)://services.sonarr.tv/ping`?

src/UI/Settings/General/GeneralViewTemplate.hbs
```diff
@@ -230,6 +237,37 @@
                     <input type="password" name="proxyPassword" class="form-control"/>
                 </div>
             </div>
+            
+            <div class="form-group">
+                <label class="col-sm-3 control-label">Addresses for the proxy to ignore</label>
+
+                <div class="col-sm-1 col-sm-push-4 help-inline">
+                    <i class="icon-sonarr-form-info" title="Use ';' as a seperator"/>
```
Taloth: seperator => separator

src/NzbDrone.Api/Config/HostConfigResource.cs
```diff
@@ -27,11 +27,13 @@ public class HostConfigResource : RestResource
...
         public UpdateMechanism UpdateMechanism { get; set; }
         public string UpdateScriptPath { get; set; }
-        public UseProxy ProxyEnabled { get; set; }
+        public bool ProxyEnabled { get; set; }
         public ProxyType ProxyType { get; set; }
         public string ProxyHostname { get; set; }
         public int ProxyPort { get; set; }
         public string ProxyUsername { get; set; }
         public string ProxyPassword { get; set; }
+        public string ProxySubnetFilter { get; set; }
```
Taloth: WebProxy.BypassList isn't really 'subnet' filtering. it basically filters based on regex/glob (unsure which). So it's probably a good idea to rename it to something like ProxyBypassFilter.

_The current implementation should be sufficient, but if I wanted to make it perfect I would do it a bit differently and support hostnames, address and subnet filtering. eg: "*.sonarr.tv;localhost;127.0.0.1;10.0.0.0/24;192.168.0.0/24"_

src/NzbDrone.Common/Http/HttpRequest.cs
```diff
@@ -54,6 +54,7 @@ public Uri Url
         public Dictionary<string, string> Cookies { get; private set; }
         public bool StoreResponseCookie { get; set; }
         public TimeSpan RateLimit { get; set; }
+        public HttpRequestProxySettings Proxy {get; set;}
```
Taloth: @markus101 Any idea on how to eliminate this property on HttpRequest? It's done this way because the settings are only available in NzbDrone.Core.
My idea atm is to define an interface in NzbDrone.Common `IHttpProxySettingsProvider.GetProxySettings(HttpRequest request)`, but provide an implementation for it in NzbDrone.Core.
The provider would be responsible for checking whether a proxy is needed, and supply the settings if so.

Taloth:
All the download clients have been migrated from RestSharp to HttpClient, so they use that stack. If you rebase your work on latest develop you shouldn't need to change any of those.

Except rtorrent, which uses XmlRpc. but afaik the generated proxy contains a `IXmlRpcProxy.Proxy` property or something.
rtorrent also happens to be the client often used on seedboxes, and thus rarely localhost. If it's possible to provide proxy support for it, it would be nice. But don't spend too much time on it, I've been thinking about removing CookComputing.XmlRpc since it won't work properly on https connections on mono.

as you can see from my comments, I would still like to eliminate the Proxy property on HttpRequest.

Taloth:
On the other hand, ppl likely would talk directly to their seedbox, or use that seedbox as vpn. so probably easier to just leave XmlRpc as it is.

lloydsparkes:
Ok that's fine then. I'll wait to see what @markus101 thinks about your idea to eliminate the Proxy property on HttpRequest before doing any more though

lloydsparkes:
Ok, so I have rebased. Now everything is on the new 'Sonnar' HTTP all the download clients will automatically get the Proxy added by the HttpInterceptor.

I have left the XmlRpc

So the only thing left is any refactoring

src/NzbDrone.Common/Http/HttpRequest.cs
markus101: That sounds like a good idea to me. For my benefit, the reason to do this is so it doesn't need to be set all over the place to be used?

Taloth: yes, but atm it's injected to HttpRequest via a HttpRequestInjector, which was never the intended usage.

lloydsparkes: The crux of why this is an issue, is because IConfigService is not available in the Common module

If we could expose IConfigService to the Common module, then we could just inject it into HttpClient (which could then pass it around to the dispatchers).

This would remove the need for any special interface just for proxy settings, and would remove the interceptor. 

It should just  be a matter of refactoring IConfigServe to the common module.

Taloth:
So I just tried this out. 
Local ssh connection to my ubuntu server with a socks5 tunnel enabled.
Windows Browser to socks5 works fine.
Windows Sonarr (forced CurlDispatcher) to socks5 works fine.
Windows Sonarr (ManagedDispatcher) to socks5 returns BadRequest.
So I changed from localhost to 127.0.0.1, that worked. Is explainable, coz localhost resolves to both ::1 and 127.0.0.1, and I had the proxy only on ipv4. Still, the response statuscode BadRequest is wrong, but in the proxy lib imho.
Then I get an nullrefexception somewhere, Org.Mentalis.Proxy.Client.Dispose. on a IOCompletionCallback. Probably coz it's already disposed. But it's a first-chance exception, coz it gets caught somewhere and possibly handled. Still nasty, again bug I suspect.

Second problem is that SocksWebProxy gets discarded and recreated all the time. meaning that a whole proxy gets created for each request.

lloydsparkes:
Ahh yea.

We could detect this use case and force a 127.0.0.1 for localhost.  But this feels wrong. 

I've looked at trying to pull out the IPv4 address from the list of resolved addresses, but we only want to do this for localhost, not remote addresses. 

So I'm not quite sure how to resolve this...

I'll look into the IOCompletionCallback issue

Taloth:
Would this be an alternative: https://github.com/Yozer/BetterHttpClient/tree/master/BetterHttpClient/Socks

Taloth:
There also some other issues, proxysettings subnet stuff isn't even populated. so I guess it was never tested either.
But I already have a little bit of work done on most of that.

lloydsparkes:
ProxySettings subnet is now filled. I have no idea how I missed that.

What did you do to tunnel to Ubuntu as Socks5?

When I used Putty - it fails (Connection Closed) as Socks5, but works as Socks4

Taloth:
Also putty, socks5 simply worked once I used 127.0.0.1. '4D5678' tunnel.

Taloth:
@lloydsparkes I did some refactoring and cleanup on [proxysupport](https://github.com/Sonarr/Sonarr/tree/proxysupport) branch. But i'm seeing some odd behavior.

When I enable a local socks5 proxy by uncommenting [this stub](https://github.com/Sonarr/Sonarr/commit/b479064abdf29af3e4d8d36b1d1533e060852ce3#diff-26338fe9e947408c04b328c334c26ceaR35) and run the HttpClientFixture tests (ManagedDispatcher), specifically [should_execute_simple_post](https://github.com/Sonarr/Sonarr/blob/2fabe2d198bcba79439f22e6d1a387b76b88b722/src/NzbDrone.Common.Test/Http/HttpClientFixture.cs#L65), it times out. Without proxy it doesn't time out.

Can you look into that?

I also still think we should add some Tests to check the Bypass Filter.

lloydsparkes:
So I am going to spend the weekend hopefully finishing this off. 

I will look at BetterHttpClient too (and first)

I'm also going to go through and retest all the various scenarios, which I have not done since Feb, given the changes. 

Scenario's as I see them:

HTTP
HTTP (auth'd)
Socks 
Socks (auth'd)

I will utilise Putty, and Tor proxy for the Socks Proxies testing and Squid for HTTP

I will also add in some unit tests for the testing of bypass filtering.

I'm going to base all the work on the proxysupport branch you have created. 

I think this PR, will basically become defunct at that point, as we would need to merge proxysupport -> proxysupport. But it probably makes sense to keep most of the conversation on here?

Taloth:
The proxysupport branch is based on your PR, so just use a fast-forward merge, then you can continue pushing changes to your own branch and thus to the PR.

I briefly looked at BetterHttpClient, one nasty part of it is that it quite literally sets up the tcp connection and creates it's own HTTP headers. A bit too low level, sadly. I was hoping it was a simple wrapper.

lloydsparkes:
So I worked out what was causing the issues with the post request tests. 

See my commit here: [SocksWebProxy commit](https://github.com/lloydsparkes/SocksWebProxy/commit/858010b3b45040a57699eeb97d8bcbcb4346ee77)

It seems for some reason the receiving client, was expecting an extra 2 bytes for a POST request, which it never got, so just closed the connection, hence leading to the timeout. 

The problem is, I tested a number of different applications sending post requests, and none of them ever sent these two extra bytes. So I bumped down the minimum number of bytes it expects.

Then it would screw up parsing the headers, as it would parse a JSON post body as a header, so I fixed that up too.

It now seems to be working for me... I'll update nuget and the packages we reference.

Taloth:
not sure where that 6 comes from, the 4 is the two newlines at the end of the headers, but a terminating newline after the response-body is indeed not required.

The more I look at SocksWebProxy the more I suspect it's not good either.
Looks like it doesn't even handle the PUT http verb.

It's better than nothing though. But it might not be the last quirk. And i'm actually seriously worried now coz it will be used to talk to download clients as well. I hope it behaves properly in all cases.

[This](https://github.com/lloydsparkes/SocksWebProxy/blob/858010b3b45040a57699eeb97d8bcbcb4346ee77/Org.Mentalis/Proxy/Http/HttpClient.cs#L219) piece of code makes me wonder how it behaves with https.

Ironically, we first thought curl wouldn't easily support socks4/5, now it's the managed client that's gonna be a problem.

lloydsparkes:
Maybe we should just force to Curl if a SOCKS Proxy is required?

Taloth:
We don't actually use curl on windows.

lloydsparkes:
Ok, the Curl tests do work on windows...

What I will do tomorrow then is to write some tests for PUT / DELETE (should I bother with PATCH?)

There are some SSL tests, "should_execute_https_get" and they seem to pass, so it seems that should be fine....

Taloth:
That's coz in the test environment we deploy some curl stuff so we can run those tests without linux. :)

lloydsparkes:
Ok, had a look at writing tests for PUT/DELETE. In theory its very easy, but HttpClient only supports HEAD, GET, and POST. So we do not need to worry here.

Also I validated that the HTTPS tests work fine with the SocksWebProxy, and do not fallback to utilising Curl

lloydsparkes:
Although I am not sure why the "Ubuntu" - Get HTTPS test fails, especially as they do utilise a proxy on the build server

lloydsparkes:
There we go, added in some tests to test the proxy bypass, which was good actually because the syntax to specify such things is slightly more complicated than I thought. 

I also tweaked how you specify the bypass list, just to make it easier for normal users.

jun6lee:
Any chance you could share the tests?

1. You've addressed giving Sonarr the ability to access indexers via a socks proxy.
2. Tor provides the anonymity based proxy but I'm struggling to configure it right.

In trying to bring the two pieces together, I keep breaking all of Sonarr's web access.

**Here's my presumption for the Sonarr side of things:**
Use Proxy: Y
Proxy Type: Socks5
Hostname: localhost [this is the NAS which tor also sits on]
Port: 9050
Username: xyz
Password: 123 [as per tor config items ]
Address for Proxy to Ignore: -- blank --
Bypass Proxy for local addresses: Y

**As for the Tor Config:**
Socks5Proxy localhost:9050
Socks5ProxyUsername xyz
Socks5ProxyPassword 123
Log notice file /opt/var/log/tor/notices.log
ExitPolicy reject *:* # no exits allowed
DataDirectory /opt/var/lib/tor
RunAsDaemon 1
ORPort 9001

I've set my port-forwarding on router for 9050 and 9001, but I know I'm missing something, since I'm not telling it anywhere what the router address is.

lloydsparkes:
Sorry for the long time to get a reply, life's been busy

I just set this up on my box (I missed it getting released). 

Have you checked your Tor proxy to ensure it working? 

Your Tor Config looks wrong (compared to mine) (I'm use tor headless on Ubuntu - 0.2.7.3-alpha)

I have:

SOCKSPort 9050

Setting Socks5Proxy in the Tor config, tells Tor to connect via the given proxy. It does not config the proxy it provides (see: http://superuser.com/questions/565722/how-to-config-tor-to-use-a-http-socks-proxy)

I don't have any authentication because its localhost only. I don't think Tor supports authenticating its proxy.

So if you remove the Socks5Proxy fields from your tor configuration and restart tor. Then remove the username and password from the sonarr configuration it should work

PR #1119, opened 2016-02-02
Statistics API endpoint

Taloth:
Basically only an endpoint atm. I don't really want to write an UI just yet. But I wanted the stats.

I haven't actually checked it for correctness yet. Probably should have a couple of tests.

Lots of stuff isn't filled in yet. The GlobalStats were supposed to contain like "last week grabs/success rate" kind of things, just "give me bottom line". And when you hover over em or click on it, it would show a modal or popup to get into the detailed IndexerStatistics.

Looking at it, it could probably be just `statistics/*` instead of `system/statistics/*` but either works for me.

Naming is all open for refactoring, this is version 0.1 "get it working".

### DIFF
--- a/src/NzbDrone.Api/System/Statistics/StatisticsModule.cs
+++ b/src/NzbDrone.Api/System/Statistics/StatisticsModule.cs
@@ -1,0 +1,62 @@
+﻿using System;
+using System.Linq;
+using Nancy;
+using Nancy.Routing;
+using NLog;
+using NzbDrone.Api.Extensions;
+using NzbDrone.Common.EnvironmentInfo;
+using NzbDrone.Common.Extensions;
+using NzbDrone.Core.Configuration;
+using NzbDrone.Core.Datastore;
+using NzbDrone.Core.History;
+using NzbDrone.Core.Lifecycle;
+using NzbDrone.Core.Statistics;
+
+namespace NzbDrone.Api.System
+{
+    public class StatisticsModule : NzbDroneApiModule
+    {
+        private readonly IStatisticsService _statisticsService;
+        private readonly Logger _logger;
+
+        public StatisticsModule(IStatisticsService statisticsService, Logger logger)
+            : base("system/statistics")
+        {
+            _statisticsService = statisticsService;
+            _logger = logger;
+
+            Get["/"] = x => GetGlobalStatistics();
+            Get["/indexer"] = x => GetIndexerStatistics();
+        }
+
+        private Response GetGlobalStatistics()
+        {
+            return new
+                {
+                    Generated = DateTime.UtcNow,
+                    Uptime = GetUpTime(),
+                    History = _statisticsService.GetGlobalStatistics()
+                }.AsResponse();
+        }
+
+        private Response GetIndexerStatistics()
+        {
+            var stats = _statisticsService.GetIndexerStatistics();
+
+            return stats.AsResponse();
+        }
+
+        private TimeSpan? GetUpTime()
+        {
+            try
+            {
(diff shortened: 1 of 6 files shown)


src/NzbDrone.Api/System/Statistics/StatisticsModule.cs
```diff
@@ -0,0 +1,62 @@
...
+            var stats = _statisticsService.GetIndexerStatistics();
+
+            return stats.AsResponse();
+        }
+
+        private TimeSpan? GetUpTime()
+        {
+            try
+            {
+                return DateTime.Now - global::System.Diagnostics.Process.GetCurrentProcess().StartTime;
```
kayone: why not just store the time that the process started using a static field rather than something like this that could possibly fail?

also I think this belongs more to `/system` endpoint than it does to statistics.

src/NzbDrone.Core/Statistics/HistoryStatistics.cs
```diff
@@ -0,0 +1,15 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace NzbDrone.Core.Statistics
+{
+    public class HistoryStatistics
+    {
+        public int Grabs { get; set; }
+        public int Replaced { get; set; }
```
kayone: Upgrade?

src/NzbDrone.Core/Statistics/HistoryStatistics.cs
```diff
@@ -0,0 +1,15 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace NzbDrone.Core.Statistics
+{
+    public class HistoryStatistics
```
kayone: `DownloadStatistics`?

src/NzbDrone.Core/Statistics/StatisticsService.cs
```diff
@@ -0,0 +1,139 @@
...
+                stats.AllTime.Failed += group.AllTime.Failed;
+                stats.AllTime.Imported += group.AllTime.Imported;
+            }
+
+            return stats;
+        }
+
+        public Dictionary<string, StatisticsGrouping<HistoryStatistics>> GetIndexerStatistics()
+        {
+            var all = _historyRepository.All().ToArray();
```
kayone: this could get extremely slow. we should be able to run pretty much all of this right on the DB.

thezoggy:
@Taloth have an example stats output?

src/NzbDrone.Core/Statistics/HistoryStatistics.cs
Taloth: The original thought was that this particular statistic comes from the History, I thought that I might find other useful sources to analyse.

src/NzbDrone.Core/Statistics/StatisticsService.cs
Taloth: That's not really possible, the Indexer is currently in the json object, not a db  column. Also we're talking about an aggregation, a running window over the dataset. I don't think I can come up with a sqlite query that handles that properly, might be possible, but probably would be a grotesk unreadable sql query.

src/NzbDrone.Core/Statistics/HistoryStatistics.cs
kayone: I think the source is less relevant than the result.

Taloth: In that case IndexerStatistics would be more applicable. Since we could also have a EpisodeStatistics, which would get data from your series and episodes and say what kind of qualities you have and the amount of data and episodes.

Issue #1122, opened 2016-02-05
Add Safari 9.0 Pinned Tab Icon


nikdoof:
Sonarr currently doesn't have an icon for Safari 9.0's pinned tab feature

[Details on creating an icon](https://developer.apple.com/library/ios/documentation/AppleApplications/Reference/SafariWebContent/pinnedTabs/pinnedTabs.html), Essentially its a black only SVG with transparent background.

NathanTheGr8:
![sonarr2](https://cloud.githubusercontent.com/assets/2585114/12864987/e457d6fc-cc64-11e5-96ec-e57ec964db75.png)

That is obviously a png but I have the svg, how is that?

NathanTheGr8:
![sonarr](https://cloud.githubusercontent.com/assets/2585114/12876810/53101820-cdcd-11e5-86de-06bf25e62fb9.png)

How is that for the black and white logo? svg below, wouldn't render


[sonarr.svg.zip](https://github.com/Sonarr/Sonarr/files/121015/sonarr.svg.zip)

xelra:
Looks very nice. Goes with the whole translucent trend and is definitely a step upwards from the forced white backround. That is if, I hope, everything white in that picture is supposed to be transparent.

I think this would make a very nice new logo in general. Maybe for v3?

NathanTheGr8:
https://gist.github.com/markus101/8b48c9e2a4de9d598338
That is the design I included in my fork. Mark made it obviously. I personally liked it with the circle but his is also much less janky then mine. I am not great with svg software and did some weird things to get the white transparent.

xelra:
I think I kinda like it more with the circle. It also looks a bit like an eye. :)

Issue #1108, opened 2016-01-29
Adding Windows 8/10 Tile info


Gobelet:
Hello,

I was trying to pin a shortcut to my Sonarr install, and I saw that it used the MS Edge icon instead of a Sonarr icon. CouchPotato has a special tile design, however. Here's what's in the header of CouchPotato:

&lt;meta name="msapplication-TileColor" content="#f85c22">
&lt;meta name="msapplication-TileImage" content="/couchpotato/static/images/icons/windows.png">
&lt;meta name="msapplication-square70x70logo" content="/couchpotato/static/images/icons/windows.png">
&lt;meta name="msapplication-square150x150logo" content="/couchpotato/static/images/icons/windows.png">
&lt;meta name="msapplication-square310x310logo" content="/couchpotato/static/images/icons/windows.png">
&lt;link href="/couchpotato/static/images/icons/favicon.ico" rel="icon" type="image/x-icon" />

This would be a nice, easy addition to make Sonarr prettier when pinned on a W10 Mobile / Windows 10 home screen/start menu.

What do you think?

markus101:
Hanselman guide: http://www.hanselman.com/blog/MakeAWindows81PinnedLiveTileForYOURWebsiteInMinutes.aspx

markus101:
This is taken care of in the v3 work.

PR #1136, opened 2016-02-09
added Safari Pinned tab icon

NathanTheGr8:
added a pure black sag of the sonar logo
make a link in the index file


NathanTheGr8:
This is for #1122 , Never done a pull request for a big project before. Also I wasn't able to check that it works, I couldn't figure out how to compile it and run as an OX app. It is probably something simple so please post and I will check that it works.

src/UI/index.html
```diff
@@ -29,6 +29,7 @@
     <link rel="apple-touch-icon" sizes="72x72" href="/Content/Images/touch/72.png"/>
     <link rel="apple-touch-icon" sizes="114x114" href="/Content/Images/touch/114.png"/>
     <link rel="apple-touch-icon" sizes="144x144" href="/Content/Images/touch/144.png"/>
+    <link rel="mask-icon" href="/Logo/sonarr_black_and_white.svg" color="#36C6F4">
```
markus101: `/Logo` isn't accessible by the web server, it needs to be in `/Content/Images`.

markus101:
Not sure that the outline works for the image. Also the SVG needs a transparent background. I'll take a stab at it later today.

src/UI/index.html
NathanTheGr8: ok I will change the file location, thanks

NathanTheGr8:
@markus101 I believe the svg does have a transparent background. The problem about making the svg is that apple just wants only black svg. The sonarr logo can be easy changed to black and white, however the white really need to be transparent. I accomplished this by a kind of crude manor of tracing and png output of the original logo in black and white. I added the think circle line because I thought it looked weird otherwise. It you can make a better svg please do, I am pretty bad at using graphic adobe illustrator.

markus101:
After way too much effort:

https://gist.github.com/markus101/8b48c9e2a4de9d598338

I think it looks less wheel like. Unless anyone else has an opinion I think this is what we should go for.

Name is `logo.svg` in the safari folder you created and I think we're set.

NathanTheGr8:
Looks great! I liked the circle but what ever. You were able to do it much better.

NathanTheGr8:
just added some Tile images for #1108

markus101:
> just added some Tile images for #1108

Each PR should really only address one issue unless they are related and separating them is more effort than its worth.

I'm not seeing the icons load when pinning to from Edge in Windows 10, but the title is there. Looks like the sizes recommended are different than I listed previously. The size should match the meta tag name.

```
<meta name="application-name" content="Sonarr"/>
<meta name="msapplication-TileColor" content="#000000"/>
<meta name="msapplication-square70x70logo" content="tiny.png"/>
<meta name="msapplication-square150x150logo" content="square.png"/>
<meta name="msapplication-wide310x150logo" content="wide.png"/>
<meta name="msapplication-square310x310logo" content="large.png"/>
```

NathanTheGr8:
Sorry about that, I corrected it. I had changed the names to reflect the exact size like elsewhere in the project. I didn't realize it mattered.

markus101:
Its not the file names that are wrong, its that they don't match the size specified, I much prefer the resolution in the file name instead of tiny, square, etc.

NathanTheGr8:
I notice that when I was committing it however when I used this generator http://www.buildmypinnedsite.com/en, It made the file sizes you saw. I assumed it has something to do with display scaling that the wanted a bigger picture then the size in the name tag.

NathanTheGr8:
I just removed the microsoft stuff I will add that in another merge request later.

markus101:
Thanks, the safari icon has been merged.

PR #1146, opened 2016-02-12
Treat XEM aliases as SceneSeasonNumber

### LINKED ISSUE #1140: Only Apply XEM season aliases to Scene season numbers
Fairy Tail TVDB Season 2 currently has the alias for `Fairy Tail S2` applied, which takes effect in in season 4, not TVDB season 2. We'll need to apply XEM season aliases only to scene season numbers that match, not TVDB season numbers. For our own Aliases (Aqua Teen Hunger Force), we need those aliases to apply to the matching TVDB season.

markus101:
This fixes parsing, searching and UI display of aliases that should only apply to scene season numbers, not TVDB season numbers. This will treat anything from TheXEM as a scene season number and anything from Sonarr's services as a season number.

Closes #1140

### DIFF
--- a/src/NzbDrone.Core/Parser/ParsingService.cs
+++ b/src/NzbDrone.Core/Parser/ParsingService.cs
@@ -168,7 +168,7 @@ namespace NzbDrone.Core.Parser
 
             if (parsedEpisodeInfo.IsAbsoluteNumbering)
             {
-                var sceneSeasonNumber = _sceneMappingService.GetSeasonNumber(parsedEpisodeInfo.SeriesTitle);
+                var sceneSeasonNumber = _sceneMappingService.GetSceneSeasonNumber(parsedEpisodeInfo.SeriesTitle);
 
                 foreach (var absoluteEpisodeNumber in parsedEpisodeInfo.AbsoluteEpisodeNumbers)
                 {
@@ -195,8 +195,7 @@ namespace NzbDrone.Core.Parser
 
                             if (episode == null)
                             {
-                                episode = _episodeService.FindEpisode(series.Id, sceneSeasonNumber.Value,
-                                absoluteEpisodeNumber);
+                                episode = _episodeService.FindEpisode(series.Id, sceneSeasonNumber.Value, absoluteEpisodeNumber);
                             }
                         }
 
--- a/src/NzbDrone.Core/Parser/ParsingService.cs
+++ b/src/NzbDrone.Core/Parser/ParsingService.cs
@@ -141,8 +141,6 @@ namespace NzbDrone.Core.Parser
 
         public List<Episode> GetEpisodes(ParsedEpisodeInfo parsedEpisodeInfo, Series series, bool sceneSource, SearchCriteriaBase searchCriteria = null)
         {
-            var result = new List<Episode>();
-
             if (parsedEpisodeInfo.FullSeason)
             {
                 return _episodeService.GetEpisodesBySeason(series.Id, parsedEpisodeInfo.SeasonNumber);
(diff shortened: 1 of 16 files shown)


src/NzbDrone.Core/Parser/ParsingService.cs
```diff
@@ -193,11 +193,11 @@ public List<Episode> GetEpisodes(ParsedEpisodeInfo parsedEpisodeInfo, Series ser
                                 episode = episodes.First();
                             }
 
-                            if (episode == null)
-                            {
-                                episode = _episodeService.FindEpisode(series.Id, sceneSeasonNumber.Value,
-                                absoluteEpisodeNumber);
-                            }
+                            // This will cause mismatches where the the scene season number is used to link to the TVDB season number
```
markus101: Not entirely sure what this case was handling.

markus101:
I think this one is ready to review, nothing too crazy, but wanted a second set of eyes.

src/UI/Series/Details/InfoViewTemplate.hbs
```diff
@@ -55,6 +55,10 @@
             {{#if_eq seasonNumber compare="-1"}}
                 <span class="label label-default">{{title}}</span>
             {{/if_eq}}
+
+            {{#if_eq sceneSeasonNumber compare="-1"}}
```
Taloth: If both seasonNumber and sceneSeasonNumber = -1 then you'll see the title twice. Not sure if that can happen.

Taloth:
Looks good, except for that UI thingy.

Trying to dig in my memory why we had those 4 'Working' testcases, feels like it was relevant. Probably xem naming + services naming together.

src/UI/Series/Details/InfoViewTemplate.hbs
markus101: No, if one is populated the other will be null, at least in the current code (though I don't see that changing).

markus101:
The 4 test cases seemed weird as two conflicted with each other.

src/UI/Series/Details/InfoViewTemplate.hbs
Taloth: No so sure about that, what about Spartacus?
Blood.and.Sand.S01 -> tvdb S01
Gods.of.the.arena.S01 -> tvdb S00
Vengeance.S02 -> tvdb S02
War.of.the.Damned.S03 -> tvdb S03

The special case being Gods.of.the.Arena.

markus101: That doesn't change anything with this though, a single record won't ever have both values. If both XEM and our list have the same mapping then we'll get duplicates, but that would happen right now and we'd remove one of them (a big reason why we don't add anime to our list at all).

Taloth: I thought that with this change you would also support mapping Spartacus properly. So seasonNumber=0 and sceneSeasonNumber=1. Would've been nice and a welcome change.

src/NzbDrone.Core/Parser/ParsingService.cs
```diff
@@ -440,5 +326,142 @@ private Episode GetDailyEpisode(Series series, string airDate, SearchCriteriaBas
...
+        }
+
+        private List<Episode> GetStandardEpisodes(Series series, ParsedEpisodeInfo parsedEpisodeInfo, bool sceneSource, SearchCriteriaBase searchCriteria)
+        {
+            var result = new List<Episode>();
+            var seasonNumber = parsedEpisodeInfo.SeasonNumber;
+
+            if (sceneSource)
+            {
+                var tvdbSeasonNumber = _sceneMappingService.GetTvdbSeasonNumber(parsedEpisodeInfo.SeriesTitle);
```
Taloth: Shouldn't this fetch both scene & tvdb numbers? And verify it the scene/tvdb number matches and throw otherwise?

Coz atm it 'overrides' the parsed season number, which seems wrong imo.
Tbh, I would just pull this all out into a separate method and code it in a way that it applies to both anime and standard.

No big deal though, coz it looks like we have to rethink this whole scene/release mapping stuff.

markus101: Yeah, it either needs to pass in the season number it parsed along with the title or check it when it comes back. For anime we're relying on the scene season number to do the matching, but for standard series (or at least mappings from services), we want the TVDB season number, so not sure how closely they will be, but could still be in one method.

The mapping gets really complicated when you try to remap the season number like we're doing here versus it being applied to the episode when searching, because you need to switch the episode number it searches for, but need to do it without altering it too early, otherwise you end up searching for something else. I played with searching for bit and it was ugly, so I rolled it back.

src/NzbDrone.Core/Parser/ParsingService.cs
```diff
@@ -440,5 +326,142 @@ private Episode GetDailyEpisode(Series series, string airDate, SearchCriteriaBas
...
+
+                if (tvdbSeasonNumber.HasValue && tvdbSeasonNumber.Value >= 0)
+                {
+                    seasonNumber = tvdbSeasonNumber.Value;
+                }
+            }
+
+            if (parsedEpisodeInfo.EpisodeNumbers == null)
+            {
+                return new List<Episode>(); ;
```
Taloth: double ;

src/NzbDrone.Core/Parser/ParsingService.cs
Taloth: Yes, we need to merge thexem data and scene mapping into one service on skyhook, then this whole problem goes away.
tbh, at this point I think it's better to leave it and get this PR merged in... assuming it does what you originally intended.

we just need to be careful that our mappings on services never contain a tvdbSeasonNumber IF there are multiple sceneSeasonNumbers with the same series title.

markus101: I don't see the mappings being a problem, this fixes Spartacus, but it will only fix things if we actually set it, which seems rare, we hardly publish naming for particular seasons, though there are a few.

Yeah merging would certainly help, but would also require us to handle duplicate scene season + episode numbers properly, its valid right now, but we consider them to both be part of the same episode in that case, in the Spartacus case they use conflicting numbers with different names (we may already be smart enough to handle this or something we've already considered, just making note, in hopes of remembering later).

I'll fix this and double `;` below and we can get this merged.

PR #1158, opened 2016-02-28
Use existing special searching logic for anime

smmalis37:
This change just hooks up the anime searching functions to the existing special searching so that we can at least find some anime specials. This isn't going to find everything, and more work will definitely be needed, but it at least finds some specials. Something is better than nothing right?

### DIFF
--- a/src/NzbDrone.Core/IndexerSearch/NzbSearchService.cs
+++ b/src/NzbDrone.Core/IndexerSearch/NzbSearchService.cs
@@ -92,7 +92,7 @@ namespace NzbDrone.Core.IndexerSearch
 
             if (series.SeriesType == SeriesTypes.Anime)
             {
-                return SearchAnimeSeason(series, episodes);
+                return SearchAnimeSeason(series, episodes, seasonNumber);
             }
 
             if (seasonNumber == 0)
@@ -197,9 +197,14 @@ namespace NzbDrone.Core.IndexerSearch
             {
                 searchSpec.AbsoluteEpisodeNumber = episode.AbsoluteEpisodeNumber.Value;
             }
+            // Season 0 indicates specials, not regular episodes
+            else if (episode.SeasonNumber == 0)
+            {
+                return SearchSpecial(series, new List<Episode> { episode });
+            }
             else
             {
-                throw new ArgumentOutOfRangeException("AbsoluteEpisodeNumber", "Can not search for an episode without an absolute episode number");
+                throw new ArgumentOutOfRangeException("AbsoluteEpisodeNumber", "Can not search for a non-special episode without an absolute episode number");
             }
 
             return Dispatch(indexer => indexer.Fetch(searchSpec), searchSpec);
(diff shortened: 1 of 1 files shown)


src/NzbDrone.Core/IndexerSearch/NzbSearchService.cs
```diff
@@ -196,10 +196,15 @@ private List<DownloadDecision> SearchAnime(Series series, Episode episode)
             else if (episode.AbsoluteEpisodeNumber.HasValue)
             {
                 searchSpec.AbsoluteEpisodeNumber = episode.AbsoluteEpisodeNumber.Value;
+            }
+            // Season 0 indicates specials, not regular episodes
```
markus101: Comment not required.

src/NzbDrone.Core/IndexerSearch/NzbSearchService.cs
```diff
@@ -216,13 +221,21 @@ private List<DownloadDecision> SearchSpecial(Series series, List<Episode> episod
...
+        private List<DownloadDecision> SearchAnimeSeason(Series series, List<Episode> episodes, int seasonNumber)
         {
-            var downloadDecisions = new List<DownloadDecision>();
-
-            foreach (var episode in episodes)
-            {
-                downloadDecisions.AddRange(SearchAnime(series, episode));
+            var downloadDecisions = new List<DownloadDecision>();
+
+            // Season 0 indicates specials, not regular episodes
```
markus101: Not required either.

markus101:
Some tests to cover the new behavior would be good, namely that we return an error when appropriate and do a special search instead of an error.

src/NzbDrone.Core/IndexerSearch/NzbSearchService.cs
```diff
@@ -216,13 +221,21 @@ private List<DownloadDecision> SearchSpecial(Series series, List<Episode> episod
...
+            if (seasonNumber == 0)
+            {
+                downloadDecisions.AddRange(SearchSpecial(series, episodes));
```
markus101: What about specials that have an absolute episode number, this should search those episodes normally and sure special search if they are missing.

markus101:
Would also be good to have some examples where this is now working properly, so we can test it out.

src/NzbDrone.Core/IndexerSearch/NzbSearchService.cs
smmalis37: Do any of those exist? All of the anime I have atm have no absolute episode numbers for all of their specials.

markus101: Yeah they do. Offhand I'm not sure which, asked a couple users to comment here though.

markus101: `One Piece` or `Garo: The Animation` are ones to look at.

Issue #347, opened 2014-04-22
Synology Download Station support


SonarrBot:
Since we now (almost) have a public synology package, I would like like to suggest the option of supporting DS Download (Download Station), the nzb and torrent downloader built in to every synology NAS. I believe it would be a snap to support for both NZB and with the upcoming torrent features of nzbdrone. The API of DS Download is documented here http://ukdl.synology.com/ftp/other/Synology_Download_Station_Official_API_V3.pdf.

armdan:
please make this thing happen . Cant wait for Sonarr to support Synology Download manager

hsousa:
Hi!

I'm considering switching from SickRage to Sonarr and currently support for Download Station is my deal breaker (blackhole implies that I have to use/monitor the DS default downloads folder, and that's a recipe for disaster, no?)

I know this is marked as low priority, but anyone got any rough ETA?

markus101:
@hsousa I can tell you its not something that we're actively working on or something that we've slated to pick up.

Watching the default DS folder would be a bit of a mess (even though Sonarr will only process what it can parse), since Blackhole moves files (instead of copying or hardlinking) its use is limited to begin with. If DS supports subfolder watching (which applies a specific label) as well as scripts that can run for a specific label only (or a way for the script to know which label was applied), you could symlink the completed files to a separate folder and have Sonarr watch that folder.

thezoggy:
could do what we did in sb, which was just add it as a 'notifier' so it tells x app when to update. the only issue is then when something gets renamed/removed.. it wont know.. so youd have to fire off updates at that time too...

gnutix:
In the meantime, what's the best alternative ? Is there a tutorial somewhere ?

CorB3n:
Up ! We clearly need Synology Download Station in available downloaders, it's clearly a dealbreaker instead of Sickbeard. Please take attention of it ! Thanks !

baz1860:
Are there Synology disk stations that can run mono/sonarr but can't run
nzbget/transmission? I'd have thought anything that could run mono would be
powerful enough to run a separate downloaded. Download Station, while being
built by Synology isn't actually a great piece of software, you'd be better
off with one of the other options available on SynoCommunity

CorB3n:
[...]

PR #1160, opened 2016-02-28
Synology Download Station client

OllieDay:
I've added a new torrent and usenet client for Synology Download Station.

Basic functionality appears to work so some feedback would be good!

I'm unsure about how to approach tests without duplicating all of the code for each client since they don't contain any logic. Any ideas?

Issue #347.


tutipeti:
@OllieDay any chance you could also add `destination` as a parameter for creating the downloads? Then Sonarr could tell DS where to put the stuff. Would be really useful for me :) Thanks!

tutipeti:
@OllieDay wow, that was super fast. Thanks! :)

Taloth:
It's nice that DS is being worked on, it's been requested for some time now.

However, I do have some big comments:
- I see ALOT of logic in the `DownloadStationProxy` that doesn't belong there. The Proxy is only supposed to read the api, not do businesslogic. The Proxy shouldn't know anything about `DownloadClientItem` and such. So all that gotta be moved to `DownloadStation*` itself, that includes `Test` and `GetItems`. I suggest you look at the other download clients to see how it's done there.
- `DownloadStationUsenetClientFixture` is empty. So... write some tests. This is where the Proxy becomes important, coz when the Proxy is a thin wrapper around the api, you can mock against it and test all the businesslogic in `DownloadStationUsenetClient`.
- `DownloadStationUsenetClient` and `DownloadStationTorrentClient` both call the same `GetItems`, but who filters it? Like it is, and when you have both usenet & torrent clients added you would get dupes and all weird shit happening.
- We've been working on removing RestSharp, examples for how the new Proxies are being implemented can be found in https://github.com/Sonarr/Sonarr/pull/1166.
- Please tell me Download Station has some kind of Category support? If not, look at `Transmission` for an example of how it's dealt with there.

OllieDay:
I'm having issues with remote path mapping since Synology uses "Shared Folders" which don't contain the full path, e.g. I have the Shared Folder "/volume1/Downloads" but the API returns only "Downloads" which causes the remote path mapping to fail.

Any workaround?

Taloth:
Remote path mapping doesn't need the full path, just whatever the api reports. But it only applies when sonarr and download station doesn't run on the same syno, so I don't know how relevant it is. 

Alternatively, Isn't there an api to resolve the shared folder?

OllieDay:
> Remote path mapping doesn't need the full path, just whatever the api reports. But it only applies when sonarr and download station doesn't run on the same syno, so I don't know how relevant it is.

This is my remote path mapping in Sonarr:

```
Host: nas.home
Remote path: Downloads
Local path: X:\
```

The following returns `Downloads` rather than `X:\` as it's `OsPathKind.Unknown`

`_remotePathMappingService.RemapRemoteToLocal("nas.home", new OsPath("Downloads"))`

> Alternatively, Isn't there an api to resolve the shared folder?

There is, but I wanted to avoid any extra complexity if possible.

Taloth:
What what happens if Sonarr runs locally, how would it access 'Downloads'? I assumed it would be `/Downloads`.

OllieDay:
It would be `/volume1/Downloads`. The API accepts download task creation using the full path, but queries return only `Downloads`.

Presumably the solution would be to set up a remote path mapping as follows:

```
Remote path: /volume1/Downloads
Local path: X:\
```

Then when retrieving a download task with destination `Downloads` use their API to resolve the full path as `/volume1/Downloads` before using the `RemotePathMappingService`.

Does that sound right?

Taloth:
Yes, using the File Station api to resolve it to the full path would be best.

foXaCe:
Hello, great job for this torrent client . you're going to merge this commit?

cybrox:
Is there any more progress planned on this?
I would really like to switch to Sonarr from Sickrage, but the lack of feedback and the wrong states in the UI following this really keep me from doing this right now, even though it technically works with my own little blackhole script.

OllieDay:
Been kinda busy lately but I'll try find time to work on it this week.

Prime-Omega:
I would absolutely love to have this commit Keep up the good work, hoping to have it merged with my Sonarr soon!

cybrox:
Awesome! Really appreciate your contribution @OllieDay

elisiano:
👍 
I've been waiting for a while for this to make it to upstream so that I have no more excuses not to switch to sonarr.

vfportero:
I was thinking about implement DS Download integration with Sonarr but I see that there's one almost ready.

Please @OllieDay, solve those conflicts and make this greatness can be added to master!

cristiangauma:
Thank you for your time @OllieDay !! I hope you can get some spare time to finish it. Would be awesome to see this PR accepted on Master!.

markus101:
Obsolete, see #1620

PR #1174, opened 2016-03-08
Revised caching mechanism in XemService and SceneMappingService

Taloth:
Added CachedDictionary that's basically an auto-refreshing cache. Used in both places now.

I haven't tested it thoroughly.

One concern I already noticed is that the SceneMappings are used in the API to add alternativetitles, meaning the api call may wait for an external timeout. Not that bad though, coz in the past we used to do SceneMapping refresh on Appstart event.

### DIFF
--- a/src/NzbDrone.Common/Cache/CachedDictionary.cs
+++ b/src/NzbDrone.Common/Cache/CachedDictionary.cs
@@ -1,0 +1,137 @@
+﻿using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
+
+namespace NzbDrone.Common.Cache
+{
+
+    public class CachedDictionary<TValue> : ICachedDictionary<TValue>
+    {
+        private readonly Func<IDictionary<string, TValue>> _fetchFunc;
+        private readonly TimeSpan? _ttl;
+
+        private DateTime _lastRefreshed = DateTime.MinValue;
+        private ConcurrentDictionary<string, TValue> _items = new ConcurrentDictionary<string, TValue>();
+
+        public CachedDictionary(Func<IDictionary<string, TValue>> fetchFunc = null, TimeSpan? ttl = null)
+        {
+            _fetchFunc = fetchFunc;
+            _ttl = ttl;
+        }
+
+        public bool IsExpired(TimeSpan ttl)
+        {
+            return _lastRefreshed.Add(ttl) < DateTime.UtcNow;
+        }
+
+        public void RefreshIfExpired()
+        {
+            if (_ttl.HasValue && _fetchFunc != null)
+            {
+                RefreshIfExpired(_ttl.Value);
+            }
+        }
+
+        public void RefreshIfExpired(TimeSpan ttl)
+    {
+            if (IsExpired(ttl))
+            {
+                Refresh();
+            }
+        }
+
+        public void Refresh()
+        {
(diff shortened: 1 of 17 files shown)


src/NzbDrone.Common/Cache/CachedDictionary.cs
```diff
@@ -0,0 +1,125 @@
...
+using System.Linq;
+using System.Text;
+
+namespace NzbDrone.Common.Cache
+{
+
+    public class CachedDictionary<TValue> : ICachedDictionary<TValue>
+    {
+        private readonly Action _refreshFunc;
+        private readonly TimeSpan _lifeTime;
```
kayone: `_ttl`

src/NzbDrone.Common/Cache/CachedDictionary.cs
```diff
@@ -0,0 +1,125 @@
...
+
+        private DateTime _lastRefreshed = DateTime.MaxValue;
```
kayone: shouldn't this be `DataTime.MinValue`?

src/NzbDrone.Common/Cache/CachedDictionary.cs
```diff
@@ -0,0 +1,125 @@
...
+
+        public CachedDictionary(Func<Dictionary<string, TValue>> fetchFunc, TimeSpan lifeTime)
+        {
+            _refreshFunc = () => Update(fetchFunc());
+            _lifeTime = lifeTime;
+        }
+
+        protected bool IsExpired(TimeSpan lifeTime)
+        {
+            return _lastRefreshed == DateTime.MaxValue || _lastRefreshed.Add(lifeTime) < DateTime.UtcNow;
```
kayone: if you use `MinValue` then you don't need the first check anymore.

src/NzbDrone.Common/Cache/CachedDictionary.cs
```diff
@@ -0,0 +1,125 @@
...
+        {
+            _refreshFunc();
+            _lastRefreshed = DateTime.UtcNow;
+        }
+
+        public void Update(Dictionary<string, TValue> items)
+        {
+            if (items != null)
+            {
+                _items = items;
```
kayone: `_items = new new Dictionary<string, TValue>(items)`

kayone: also, why not allow update to clear the cache?

kayone:
Also, any reason why not just use, https://github.com/kayone/Purse/blob/master/src/Purse/PrefetchedCache.cs

Taloth:
Coz I don't want it to Prefetch on create, I want it to lazy load.

src/NzbDrone.Common/Cache/CachedDictionary.cs
Taloth: Provide an empty Dictionary and it will. Update(null) is intended to say 'sry couldn't fetch it', but reset the ttl so we don't spam.

kayone:
That seems like a minor change that could be made to `PrefetchedCache` rather than re-implementing the whole thing,

src/NzbDrone.Common/Cache/CachedDictionary.cs
kayone: That seems like a very _hidden feature_ why not just call `ResetTTL` instead of `Update(null)`?

PR #1196, opened 2016-03-22
Fixed: Allow underscore when validating hostnames

iainnicol:
This is particularly useful on FreeNAS.  It uses underscores in the hostnames when creating plugin jails.


Taloth:
Tnx, btw. ;) merged it in

PR #1200, opened 2016-03-23
New: Improved startup performance

Taloth:
@markus101 @kayone 

Tested by starting Sonarr and terminating at a specific point in the code. Measured by Process.StartTime & ExitTime.
Test run 21 times and then taken the median duration (not average).

```
9608 ms     Vanilla
3124 ms     NLog 4.3.0-rc1 (includes ReconfigExistingLoggers fix)
2891 ms     Reconfigure Logging early in the process to set the correct log level.
2002 ms     Don't set ACL if already set.
```

I left out a couple of commits that didn't have as big an impact as I hoped.

I'll run a couple of more profile sessions, but as far as I'm concerned it's already Mission Accomplished :smile:

### DIFF
--- a/src/NzbDrone.Common/Disk/DiskProviderBase.cs
+++ b/src/NzbDrone.Common/Disk/DiskProviderBase.cs
@@ -293,18 +293,30 @@ namespace NzbDrone.Common.Disk
                 var sid = new SecurityIdentifier(accountSid, null);
 
                 var directoryInfo = new DirectoryInfo(filename);
-                var directorySecurity = directoryInfo.GetAccessControl();
+                var directorySecurity = directoryInfo.GetAccessControl(AccessControlSections.Access);
+
+                var rules = directorySecurity.GetAccessRules(true, false, typeof(SecurityIdentifier));
+
+                if (rules.OfType<FileSystemAccessRule>().Any(acl => acl.AccessControlType == controlType && (acl.FileSystemRights & rights) == rights && acl.IdentityReference.Equals(sid)))
+                {
+                    return;
+                }
 
                 var accessRule = new FileSystemAccessRule(sid, rights,
                                                           InheritanceFlags.ContainerInherit | InheritanceFlags.ObjectInherit,
-                                                          PropagationFlags.None, controlType);
+                                                          PropagationFlags.InheritOnly, controlType);
 
-                directorySecurity.AddAccessRule(accessRule);
-                directoryInfo.SetAccessControl(directorySecurity);
+                bool modified;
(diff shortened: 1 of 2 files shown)


src/NzbDrone.Common/Disk/DiskProviderBase.cs
```diff
@@ -293,18 +293,30 @@ public void SetPermissions(string filename, WellKnownSidType accountSid, FileSys
...
+
+                if (modified)
+                {
+                    directoryInfo.SetAccessControl(directorySecurity);
+                }
             }
             catch (Exception e)
             {
-                Logger.Warn(e, string.Format("Couldn't set permission for {0}. account:{1} rights:{2} accessControlType:{3}", filename, accountSid, rights, controlType));
+                Logger.Warn(e, "Couldn't set permission for {0}. account:{1} rights:{2} accessControlType:{3}", filename, accountSid, rights, controlType);
```
markus101: Because `string.Format` is more efficient than nlog?

kayone: even if string format is better, we should be using string interpolation rather than format since it's validated at compile time.

markus101: Making me upgrade to VS 2015 aren't you? Do we need to upgrade .net or not because its purely C# changes that require the new IDE?

kayone: just the IDE, it's syntax sugar + compiler magic.

Taloth: 1. No, it changed from string.Format to the appropriate override. It's just one I forgot when migrating to nlog 4.2. (Note that for Debug/Trace this Nlog override is faster, coz the format only executes when needed.
2. Avoid string interpolation for logging, coz it would execute before nlog is called. Roslyn has a ridiculous design flaw that makes it prefer casting it to string instead of using a FormattedString overload. (So if you have Warn(string) & Warn(FormattedString), then roslyn would cast the string interpolation (which is a FormattedString instance) to string and use Warn(string))

markus101: LOL, reading comprehension `F`, but a good discussion due to it :smile:

kayone: @Taloth I wasn't saying we should use string interpolation for logging, for the same reasons you mentioned, what I was saying is we should us it instead of `string.format`.

Taloth:
Merged into develop.

PR #1204, opened 2016-03-26
WIP: Improved API performance by removing ValueInjecter.

Taloth:
@markus101 @kayone
So, as discussed in chat. Using custom mappings instead of injecter the series api call alone dropped from 400ms to 160ms. (Adding an index to the database dropped it to 130ms, but that's not relevant atm)
The improvement is significant, I expect even bigger improvements for low-powered devices.

So this PR is about entirely retiring ValueInjecter. In hindsight, removing it entirely might've been a little enthusiastic, but it's still the ideal goal as far as I'm concerned.

I'm not entire sure about the design, but I wanted to get the custom mappings typed down. Refactoring them into non-static classes would become fairly easy.

Most of the mappings are done, except for the ProviderModuleBase stuff.

A few design approaches: (atm the prototype is nr1)
1) Use the static Extension methods I've created. The modules themselves occasionally 'fix up' the resource.
2) Add `IMappable<TModel, TMapper> where TMapper : IResourceMapper, new()` to all resources. Then a global MapToResource extension method can locate and instantiate the appropriate mapper.
3) Expose the ResourceMappers via IoC to the modules.

The advantage of option 1 is that the mappers have little scaffolding and are included in the files of the resources and thus easily modified.

Please take a look at it and shoot at it.
I've added '#warning' TODO/FIXME throughout about stuff that needs to be addressed or removed.

The numerous '#warning FIXME: Unused' comments should be evidence that ValueInjection is generally a bad idea.


src/NzbDrone.Api/Config/DownloadClientConfigResource.cs
```diff
@@ -15,4 +16,25 @@ public class DownloadClientConfigResource : RestResource
...
         public bool RemoveFailedDownloads { get; set; }
     }
+
+    public static class DownloadClientConfigResourceMapper
+    {
+        public static DownloadClientConfigResource MapToResource(IConfigService model)
+        {
+            return new DownloadClientConfigResource
+            {
+                Id = 1,
```
kayone: I think this might be a big assumption, I don't think it belongs here, if default it always `1` then maybe move it to `DownloadClientConfigResource` constructor, if not I would keep it in the RestModule.

kayone:
I really like how it turned out, to me it feels much more intuitive and less magical (in a good way) than it was before.

Personally, I would prefer to go with #1, they syntax is clear and concise, and I don't see a benefit in adding a bunch of work both for us, and at runtime to figure out which mapper to use for what, another big benefit is you can always ctrl+click on the `MapToResource()` and it takes you right to the mapping, no need to find the class that does the mapping.

one suggestion, I think it might be nicer to name `MapToResource()` to `ToResource()` so it matches `.ToString()`, `ToJson()` etc, also less wordy.

Taloth:
@markus101 Can you search for '#warning' in this PR and lemme know what we should do about those things?

src/NzbDrone.Api/Indexers/ReleaseResource.cs
```diff
@@ -74,4 +75,97 @@ public DownloadProtocol DownloadProtocol
...
+
+        public static ReleaseInfo ToModel(this ReleaseResource resource)
+        {
+            ReleaseInfo model;
+
+            if (resource.Protocol == DownloadProtocol.Torrent)
+            {
+                model = new TorrentInfo
+                {
+#warning FIXME: Missing on resource?
```
markus101: Sounds like we don't use them, but I'm not opposed to adding them to have a more complete representation of the model in the resource.

src/NzbDrone.Api/ManualImport/ManualImportResource.cs
```diff
@@ -20,13 +21,36 @@ public class ManualImportResource : RestResource
...
+        public static ManualImportResource ToResource(this Core.MediaFiles.EpisodeImport.Manual.ManualImportItem model)
         {
-            get
+            if (model == null) return null;
+
+            return new ManualImportResource
             {
-                return Path.GetHashCode();
-            }
+#warning FIXME: GetHashCode cannot be depended upon for uniqueness.
```
markus101: Can we use something similar to what we used for queue IDs?

src/NzbDrone.Api/Series/SeriesEditorModule.cs
```diff
@@ -20,10 +20,13 @@ public SeriesEditorModule(ISeriesService seriesService)
         private Response SaveAll()
         {
             //Read from request
-            var series = Request.Body.FromJson<List<SeriesResource>>().InjectTo<List<Core.Tv.Series>>();
+            var resources = Request.Body.FromJson<List<SeriesResource>>();
+
+#warning TODO: Shouldn't we somehow handle which properties can be set instead of injecting all of it?
```
markus101: Yes, thats one thing I want in v3 to make the payload A LOT smaller. I'm not sure how we want to handle that between the client sending it and the changes being stored in the DB though.

src/NzbDrone.Api/Series/SeriesModule.cs
```diff
@@ -113,14 +112,19 @@ private List<SeriesResource> AllSeries()
 
         private int AddSeries(SeriesResource seriesResource)
         {
-            var series = _seriesService.AddSeries(seriesResource.InjectTo<Core.Tv.Series>());
+            var model = seriesResource.ToModel();
 
-            return series.Id;
+#warning TODO: Fetch from skyhook in the service before creating? (In AddSeries)
```
markus101: Fetch the details of the series, that'd be nice to reduce what the client _needs_ to send (especially for 3rd party integration).

src/NzbDrone.Api/Series/SeriesModule.cs
```diff
@@ -113,14 +112,19 @@ private List<SeriesResource> AllSeries()
...
+
+            return _seriesService.AddSeries(model).Id;
         }
 
         private void UpdateSeries(SeriesResource seriesResource)
         {
-            GetNewId<Core.Tv.Series>(_seriesService.UpdateSeries, seriesResource);
+#warning TODO: How to do this atomically?
```
markus101: Partial updates here as well?

src/NzbDrone.Api/Series/SeriesModule.cs
```diff
@@ -221,7 +229,8 @@ public void Handle(SeriesEditedEvent message)
 
         public void Handle(SeriesDeletedEvent message)
         {
-            BroadcastResourceChange(ModelAction.Deleted, message.Series.InjectTo<SeriesResource>());
+#warning TODO: Why do we send the resource along with the deletion?
```
markus101: Not sure, at the very least we'd need the ID.

PR #1205, opened 2016-03-27
Allow the user to completly disable logging

DavyLandman:
Hi there,

I like Sonarr, but my raspberry-pi's SD card doesn't. Once Sonarr is running nicely, I don't really need the logs anymore. They cause quite some IO sending the SD to the grave much sooner than needed.

Moving the data dir to a separate HDD is no solution, cause that will just cause spinups of that drive, so, I went through the code and tried to disable logging in the cleanest way I could fine.

I welcome feedback, and I would really love to have this feature :)

Cheers,
Davy


DavyLandman:
never mind, I prefer you don't merge this.

PR #1212, opened 2016-04-01
NUnit3

markus101:
@kayone @Taloth I think this is ready to go, nothing too crazy, just want a second set of eyes.


src/NzbDrone.Common.Test/ServiceProviderTests.cs
```diff
@@ -24,8 +25,10 @@ public void Setup()
         [TearDown]
         public void TearDown()
         {
-            WindowsOnly();
```
markus101: This was causing the test to fail as it was throwing an exception during tear down, not sure if 2.6 didn't care or 3.2 executes the TearDown if the SetUp fails, but this fixes it.

src/NzbDrone.Core.Test/IndexerSearchTests/SearchDefinitionFixture.cs
```diff
@@ -8,16 +9,16 @@ namespace NzbDrone.Core.Test.IndexerSearchTests
...
-        [TestCase("Chicago P.D.", Result = "Chicago+PD")]
-        [TestCase("Kourtney And Khlo\u00E9 Take The Hamptons", Result = "Kourtney+And+Khloe+Take+The+Hamptons")]
-        public string should_replace_some_special_characters(string input)
+        [TestCase("Betty White's Off Their Rockers", "Betty+Whites+Off+Their+Rockers")]
+        [TestCase("Star Wars: The Clone Wars", "Star+Wars+The+Clone+Wars")]
+        [TestCase("Hawaii Five-0", "Hawaii+Five+0")]
+        [TestCase("Franklin & Bash", "Franklin+and+Bash")]
+        [TestCase("Chicago P.D.", "Chicago+PD")]
+        [TestCase("Kourtney And Khlo\u00E9 Take The Hamptons", "Kourtney+And+Khloe+Take+The+Hamptons")]
+        public void should_replace_some_special_characters(string input, string expected)
```
markus101: The old way wasn't supported anymore that I could tell, plus this is more consistent with everything else we do.

src/NzbDrone.Core.Test/InstrumentationTests/DatabaseTargetFixture.cs
```diff
@@ -68,6 +69,7 @@ public void write_long_log()
 
         [Test]
         [Explicit]
+        [ManualTest]
```
markus101: It's dumb, but the new runner sees that we want to run the tests because they we're ignored and includes them, left `Explicit` for other runners (like resharper).

test.sh
```diff
@@ -0,0 +1,41 @@
+PLATFORM=$1
```
markus101: New script, works on windows and is dead simple to operate (or so I think), embracing our BASH overlords.

Taloth:
Looks good to me.

test.sh
```diff
@@ -0,0 +1,41 @@
...
+fi
+
+if [ "$TYPE" = "Unit" ]; then
+  WHERE="$WHERE && cat != IntegrationTest && cat != AutomationTest"
+elif [ "$TYPE" = "Integration" ] ; then
+  WHERE="$WHERE && cat == IntegrationTest"
+elif [ "$TYPE" = "Automation" ] ; then
+  WHERE="$WHERE && cat == AutomationTest"
+else
+  echo "Type must be provided: Unit, Integration or Automation"
```
kayone: just to make it a bit simpler to operate, I would remove these two checks, and add a single check at the top to make sure both platform and type are provided. and if not, show a single message with examples.

test.sh
```diff
@@ -0,0 +1,41 @@
...
+  WHERE="$WHERE && cat != WINDOWS"
+  NUNIT_COMMAND="mono --debug --runtime=v4.0 $NUNIT"
+else
+  echo "Platform must be provided: Windows or Linux"
+  exit 1
+fi
+
+if [ "$TYPE" = "Unit" ]; then
+  WHERE="$WHERE && cat != IntegrationTest && cat != AutomationTest"
+elif [ "$TYPE" = "Integration" ] ; then
```
kayone: `int` instead of `Integration`? less typing :)

test.sh
```diff
@@ -0,0 +1,41 @@
...
+  TEST_DIR="$TEST_DIR/_tests"
+fi
+
+NUNIT="$TEST_DIR/NUnit.ConsoleRunner.3.2.0/tools/nunit3-console.exe"
+NUNIT_COMMAND="$NUNIT"
+NUNIT_PARAMS="--teamcity"
+
+if [ "$PLATFORM" = "Windows" ]; then
+  WHERE="$WHERE && cat != LINUX"
+elif [ "$PLATFORM" = "Linux" ]; then
```
kayone: I'm wondering if this should be mono, since it applies to OSX as well.

test.sh
markus101: Thats how I originally did it, but liked this better as it kept everything together. I think if we add more test groups later it would serve as a good reminder to update the echo statement as well.

test.sh
markus101: I think I prefer the full thing, but maybe handle both?

test.sh
markus101: Yeah... I wasn't sure, I dislike "mono" as a platform when Windows is the alternative. I'd be in favour of having "Mac" or "OSX" added as well in case we need to handle something differently there/add more Mac specific tests.

test.sh
kayone: ok, then maybe update the error message to give examples? this doesn't really help since someone could run 

`test.sh Integration Windows` and it would complain.

test.sh
kayone: sure,

PR #1219, opened 2016-04-04
Cleanup HttpUri.PathCombine

markus101:
@Taloth

### DIFF
--- a/src/NzbDrone.Common.Test/Http/HttpUriFixture.cs
+++ b/src/NzbDrone.Common.Test/Http/HttpUriFixture.cs
@@ -1,0 +1,29 @@
+﻿using FluentAssertions;
+using NUnit.Framework;
+using NzbDrone.Common.Http;
+using NzbDrone.Test.Common;
+
+namespace NzbDrone.Common.Test.Http
+{
+    public class HttpUriFixture : TestBase
+    {
+        private HttpUri GivenHttpUri(string basePath)
+        {
+            return new HttpUri("http", "localhost", 8989, basePath, null, null);
+        }
+
+        [TestCase("", "", "")]
+        [TestCase("base", "", "/base")]
+        [TestCase("/base", "", "/base")]
+        [TestCase("", "relative", "/relative")]
+        [TestCase("", "/relative", "/relative")]
+        [TestCase("base", "relative", "/base/relative")]
+        [TestCase("base", "/relative", "/base/relative")]
+        [TestCase("/base", "relative", "/base/relative")]
+        [TestCase("/base", "/relative", "/base/relative")]
+        public void should_combine_base_path_and_relative_path(string basePath, string relativePath, string expected)
+        {
+            GivenHttpUri(basePath).CombinePath(relativePath).Path.Should().Be(expected);
+        }
+    }
+}
--- a/src/NzbDrone.Common.Test/Http/HttpUriFixture.cs
+++ b/src/NzbDrone.Common.Test/Http/HttpUriFixture.cs
@@ -7,23 +7,78 @@ namespace NzbDrone.Common.Test.Http
 {
     public class HttpUriFixture : TestBase
     {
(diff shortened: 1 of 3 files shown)


src/NzbDrone.Common.Test/Http/HttpUriFixture.cs
```diff
@@ -0,0 +1,84 @@
...
+        [TestCase("base", "", "base")]
+        [TestCase("/base", "", "/base")]
+        [TestCase("/base/", "", "/base/")]
+        [TestCase("", "relative", "relative")]
+        [TestCase("", "/relative", "/relative")]
+        [TestCase("/", "relative", "/relative")]
+        [TestCase("/", "/relative", "/relative")]
+        [TestCase("base", "relative", "relative")]
+        [TestCase("base", "/relative", "/relative")]
+        [TestCase("/base", "relative", "/relative")]
```
markus101: Why would relative overwrite base in this case?

src/NzbDrone.Common/Http/HttpUri.cs
```diff
@@ -144,23 +144,38 @@ private static string CombinePath(string basePath, string relativePath)
...
-                }
+            if (relativePath.StartsWith("/"))
+            {
+                return relativePath;
             }
 
-            return newPath;
+            var baseSlashIndex = basePath.LastIndexOf('/');
+
+            if (baseSlashIndex == 0)
```
markus101: I'm confused by this, unless we have a trailing slash or multiple slashes we're always going to throw out the base path?

src/NzbDrone.Common/Http/HttpUri.cs
```diff
@@ -144,23 +144,38 @@ private static string CombinePath(string basePath, string relativePath)
...
+            {
+                return "/" + relativePath.TrimStart('/');
```
markus101: We shouldn't need to TrimStart since we check above if it starts with a slash.

src/NzbDrone.Common.Test/Http/HttpUriFixture.cs
Taloth: In Uri Combine it does. truncates the 'file' part. Check it out in your browser:
<a href="1024">test 1</a> `<a href="1024">test 1</a>`
<a href="/1024">test 2</a> `<a href="/1024">test 2</a>`

markus101: Huh, when you explain it like that, it makes a lot of sense. For the URL Base issue we should be requiring a trailing slash then?

src/NzbDrone.Common/Http/HttpUri.cs
Taloth: correct, removing it.

src/NzbDrone.Common/Http/HttpUri.cs
Taloth: No, if there's a slash in base then everything after it gets removed. (the last slash).
During some coding substring(0,0) seems no good, but that's not true. I'll clean it up a bit to, but functionally that's still the same:
```
if (baseSlashIndex >= 0)
{
   return basePath.Substring(0, baseSlashIndex) + "/" + relativePath.TrimStart('/');
}
```

src/NzbDrone.Common.Test/Http/HttpUriFixture.cs
Taloth: No, coz for Urlbase in the indexers we use CombinePath (the second set of tests), which is now simplified to always insert a slash were needed.

The top part (uri combine) is only used in the operator + (for example [here](https://github.com/Sonarr/Sonarr/blob/2fabe2d198bcba79439f22e6d1a387b76b88b722/src/NzbDrone.Core/Indexers/RssParser.cs#L270))
And uses all the logic to handle all those different relative uri formats.

PR #1222, opened 2016-04-05
Win10 utorrent

kayone:


src/NzbDrone.Common/Http/HttpRequestBuilder.cs
```diff
@@ -232,6 +234,13 @@ public virtual HttpRequestBuilder Resource(string resourceUrl)
             return this;
         }
 
+        public virtual HttpRequestBuilder KeepAlive()
```
kayone: this feels a bit strange, I think `KeepAlive(true)` `KeepAlive(false)` would be nicer.

src/NzbDrone.Common/Http/HttpRequestBuilder.cs
```diff
@@ -253,6 +262,19 @@ public virtual HttpRequestBuilder SetHeader(string name, string value)
...
 
+        public virtual HttpRequestBuilder AddPrefixQueryParam(string key, object value, bool replace = false)
+        {
+            if (replace)
+            {
+                QueryParams.RemoveAll(v => v.Key == key);
+                SuffixQueryParams.RemoveAll(v => v.Key == key);
+            }
+
+            QueryParams.Insert(0, new KeyValuePair<string, string>(key, value.ToString()));
```
kayone: why do this  for prefix but keep `Suffix`in a separate list?

kayone: does utorrent or anyone else actually care about the order of the querystring?

Taloth: Having less important stuff at the end (apikey, etc) makes it easier to read in log files. That's it.

Having suffix separately saves me from having an integer somewhere keeping track of the insertion point.

kayone: In that case I think we should kill it, if your want your query string in a specific order then insert them in that order.

PR #1241, opened 2016-04-13
Update license expression to SPDX standard

samholmes1337:
When using `npm install` I received an error stating that the `license` field in `package.json` did not conform to the new SPDX standard.

Considering that the previous license was `BSD`, I have updated it to the `BSD-3-Clause`: the most relevant valid expression.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "url": "git://github.com/Sonarr/Sonarr.git"
   },
   "author": "",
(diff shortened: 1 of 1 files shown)


samholmes1337:
I did a silly thing and didn't realise that the npm license didn't reflect the one in the `readme.md`. Opened a new pull request where fewer things went wrong!

PR #1242, opened 2016-04-13
Update package.json license expression

samholmes1337:
The `readme.md` specifies that the license for the project is GPLv3, but the value contained in the `package.json` license field was `BSD`. I have changed it to the `GPL-3.0` SPDX license string both so npm doesn't throw any warnings about license string validity and to better reflect the license of the project.


kayone:
Thanks for the fix.

Issue #309, opened 2013-07-25
Trakt Integration


SonarrBot:
We should integrate with trakt.tv to sync watched status as well as adding downloaded episodes to library (optional of course).

Recommendations and sorting by trakt rating too.

Automatically add series from trakt's lists.

Related Features
- [ ] Delete when series/episode is watched
- [ ] Download only x number of future episodes, auto download next one if one becomes watched.

rickytin:
I will love Trakt integration with Sonarr.

captainnapalm:
My desire is for watched status to toggle the monitored status. If I've already watched an episode, I don't need Sonarr to download the next quality in the profile.

markus101:
@captainnapalm it won't if the cutoff has been met; set your cutoff lower.

captainnapalm:
@markus101 yes that's a good point, but if I haven't watched the episode yet, I want Sonarr to grab the next best quality.

tidusjar:
I would like this to be added! I'd find it very useful for being able to add new series without having to log into Sonarr.

SimonW500:
Now that Trakt.tv has gone through its own update, is this feature request within Sonarr likely to gain traction?

jdfalk:
+1 on prioritization of this issue. Integration would be fantastic and with all the features that the dev's and users have suggested this would most likely attract a much larger following to Sonarr.

greatvc:
Yes please! Having the downloaded episodes automatically mark as collected in trakt gets my vote up

wersplat:
This is the only feature that Sickbeard had that made it difficult to switch..+1 on prioritization a well

cryodream:
+1 for trakt integration. Also, sync trakt lists to sonarr tags.

Producermark:
Please add. +1 on prioritization

donatasben:
+1 for trakt! Was about to suggest this but searched and found it was requested already a long time ago. Any progress? Would love this.

JigSawFr:
:+1:

huvanile:
+1 for trakt!

SamHill77:
Something like SickRage would be awesome.
[http://www.mediafire.com/convkey/0af9/bezck1srf8biujpzg.jpg](url)

dabvid:
+1

kingmatusevich:
+1 for this

bluenote73:
Automatically add from a trakt specified list world be great.

matgardon:
[...]

Issue #276, opened 2015-02-25
Import new Series from IMDB Watch List


SonarrBot:
Example list: http://www.imdb.com/list/export?list_id=ls076157636&author_id=ur58869533p (CSV)

Or RSS: http://rss.imdb.com/list/ls076157636/

We need to provide a way to set the defaults when adding a series via this mechanism, likely on a per feed basis. Trakt is also a candidate for a feature such as this, so we could automatically import a series from a trakt list.

A new series will need a root folder, a profile, monitored state, a type (standard, daily or anime) and whether it should use season folders. When the add series page allows adding tags when adding a series we should provide the same here.

OldhamMade:
:+1:

trymbill:
+1

hsamuelsen:
Radarr already has this feature, so it should be easy to copy?

Popupkiller:
I just discovered this feature in Radarr, and would love to see it in Sonarr as well.

thieverycorp:
Lists works great in Radarr, and would be an amazing addition to Sonarr

Quiksmage:
This would be an amazing addition to Sonarr. Thanks!

jeanpoelie:
I'd love to help out building this, but I got 0 experience with this project, and can't seem to make logic of radarr's version.

AmineI:
Same here, as Radarr can import Trakt lists I'd love to see this on Sonarr :)

Smalfinn:
I'd love to see this too!

D4rkSl4ve:
Works great on Radarr, would be awesome to get it at Sonarr.  That would mean one place to select, and let Sonarr/Radarr read list, search, tag, download, done.

dedren:
I started with Radarr and assumed Sonarr could parse your IMDB list for shows. I was a bit surprised to find out it couldn't so someone pointed me here to chime in and toss my hat into the vote.

mtan93:
I could really do with this feature to allow new releases to sync to sonarr automatically...

benkap:
It hard to believe that this was opened three years ago.... I assumed it's not that complicated since I started with radarr. It's the only thing missing in sonarr to match radarr. Hope it will come soon.

duelistjp:
putting my vote in for what feature i most want from sonarr that it doesn't have yet.  i have discovered a lot of great movies pulling in top 50 stuff on radarr would like the same type of thing for series.

tgrandegger:
[...]

PR #1268, opened 2016-04-30
WIP: Watchlists

matthewdias:
#276, #309

I'm basing this off of Notifications. Started with the UI, because that's what I know. Also, I'm working on a mac, so it seems like some of the file modes have changed. Please let me know if this is an issue. I'll squash commits at significant milestones.

I'm new to .NET, as well as DDD so please give me a heads up if there's anything I should be doing differently.


markus101:
> Please let me know if this is an issue. I'll squash commits at significant milestones.

It looks like a number of files have changed, including sqlite dylibs, none of those changes should be included, if there is a valid reason for the changes they should be in a separate PR.

I take it by the UI you're planning to implement them in a similar fashion to notifications/download clients/indexers? Given trakt gives a lot more than a watch list it may not make sense to scope these to be called watched lists (I don't see a case where we would setup syncing watchlists and other elements from trakt as separate integrations).

src/UI/Settings/Watchlists/Add/WatchlistAddCollectionView.js
```diff
@@ -0,0 +1,8 @@
+var ThingyAddCollectionView = require('../../ThingyAddCollectionView');
+var AddItemView = require('./WatchlistAddItemView');
+
+module.exports = ThingyAddCollectionView.extend({
+	itemView          : AddItemView,
```
markus101: Use spaces instead of tabs (there are a few cases of this).

markus101:
Another thing to keep in mind is the UI is the first thing that will change for v3, so I wouldn't spend a lot of time on it (beyond what is already done) to avoid having to redo it, though the v3 UI is far from ready and a major WIP.

matthewdias:
>It looks like a number of files have changed, including sqlite dylibs, none of those changes should be included, if there is a valid reason for the changes they should be in a separate PR.

Yeah, I didn't change anything here, so don't know why they changed. I'll remove those changes, when I rebase.

>I take it by the UI you're planning to implement them in a similar fashion to notifications/download clients/indexers? Given trakt gives a lot more than a watch list it may not make sense to scope these to be called watched lists (I don't see a case where we would setup syncing watchlists and other elements from trakt as separate integrations).

I was planning on only implementing watchlists, but if there is a desire for deeper trakt integration, (I guess like adding obtained episodes to a `Collected` list?) then I could rename the section to `Lists` or something similarly generic. In #309 there is talk about getting more data such as user ratings, etc. I think this would mean a more drastic change in Sonarr, so it should probably be discussed properly beforehand.

>Another thing to keep in mind is the UI is the first thing that will change for v3, so I wouldn't spend a lot of time on it (beyond what is already done) to avoid having to redo it, though the v3 UI is far from ready and a major WIP.

Didn't know that was happening, but yeah, I'm stealing it straight from notifications, so it's not a big deal.

markus101:
> In #309 there is talk about getting more data such as user ratings, etc. I think this would mean a more drastic change in Sonarr, so it should probably be discussed properly beforehand.

Yeah, it'd definitely be a bigger thing to tackle, but just the watch list aspect of it would be advantageous. I think the whole concept would be part of a bigger Connections discussion, since its another 3rd party integration (like the notifications are). The more I think about it the less I'm worried about what its called now, we can change that in the UI.

iiznrml:
Has any progress been made on this?

matthewdias:
I've put it on hold until v3 is done. If anyone else wants to take a stab at it feel free.

markus101:
Closing pending v3 changes.

Issue #1273, opened 2016-05-02
Vuze support based on Transmission protocol


hmemcpy:
I've been looking into adding Vuze support, based on this [forum request](https://forums.sonarr.tv/t/vuze-torrent-client-support/8700). It seems that Vuze "Support most of Transmission's remote API", according to the developers of the Vuze Web Remote (which are the same people as Transmission, looks like).

According to the post:
> * Vuze uses the same URL base "\transmission\' (confirmed with the in the Plugin logs in Vuze)
> * The version check fails, saying "Transmission version not supported, should be 2.40 or higher."

I've done some digging: sending a `session-get` to Vuze will return the following values:
```
rpc-version=14
...
version=0.5.11
```
0.5.11 being the version number of the Vuze Web Remote plugin. `14` is the RPC version number, corresponding to Transmission v2.40 (https://trac.transmissionbt.com/browser/trunk/extras/rpc-spec.txt#L747)

Now, it seems that [Transmission.cs](https://github.com/Sonarr/Sonarr/blob/develop/src/NzbDrone.Core/Download/Clients/Transmission/Transmission.cs#L214) is checking the value of the `version` property to be at least 2.40, which explains why this fails with Vuze. After changing this to use the `rpc-version` instead, I was able to make it work with Vuze!

However, there is (at least) one more issue remains: Vuze reports downloaded directory incorrectly, whereas in Transmission it expects the path excluding the torrent name, Vuze reports it including the torrent name, so it fails to process the complete download. Also an easy fix.

Question: Given those relatively small changes, would Vuze support be a welcomed PR? It seems there is some reluctance to add it because the client is fairly old and could be abused, but... there are no technical reasons not to.

markus101:
Implemented!

PR #1275, opened 2016-05-02
Changing Transmission to check the protocol (RPC) version instead the client version.

hmemcpy:
This is a pre-requisite for supporting Vuze (#1273), which is also based on the Transmission protocol, but uses a different client version.
I believe this is a more correct approach for validating supported Transmission version, as the RPC version value is absolute.

### DIFF
--- a/src/NzbDrone.Core.Test/Download/DownloadClientTests/TransmissionTests/TransmissionFixture.cs
+++ b/src/NzbDrone.Core.Test/Download/DownloadClientTests/TransmissionTests/TransmissionFixture.cs
@@ -440,17 +440,46 @@ namespace NzbDrone.Core.Test.Download.DownloadClientTests.TransmissionTests
             items.First().OutputPath.Should().Be(@"C:\Downloads\Finished\transmission\" + _title);
         }
 
-        [TestCase("2.84 ()")]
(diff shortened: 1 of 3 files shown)


src/NzbDrone.Core.Test/Download/DownloadClientTests/TransmissionTests/TransmissionFixture.cs
```diff
@@ -440,17 +440,46 @@ public void should_fix_forward_slashes()
             items.First().OutputPath.Should().Be(@"C:\Downloads\Finished\transmission\" + _title);
         }
 
-        [TestCase("2.84 ()")]
-        [TestCase("2.84+ ()")]
-        [TestCase("2.84 (other info)")]
-        [TestCase("2.84 (2.84)")]
-        public void should_version_should_only_check_version_number(string version)
+        [Test]
+        public void should_fix_mixed_slashes()
```
hmemcpy: Oops, a test case from another PR I'm working on got committed...
It's still valid, so I rather keep it in for now.

hmemcpy:
Closing this PR in favor of another one. It has fewer changes, and doesn't modify the existing Transmission checks.

PR #1277, opened 2016-05-03
Vuze support

### LINKED ISSUE #1273: Vuze support based on Transmission protocol

hmemcpy:
Fixes #1273 

This PR adds support for the Vuze client. It is based on the Transmission implementation, as both implement the same protocol.

### DIFF
--- a/src/NzbDrone.Core.Test/Download/DownloadClientTests/TransmissionTests/TransmissionFixture.cs
+++ b/src/NzbDrone.Core.Test/Download/DownloadClientTests/TransmissionTests/TransmissionFixture.cs
@@ -1,201 +1,17 @@
 using System;
-using System.Linq;
 using System.Collections.Generic;
+using System.Linq;
 using FluentAssertions;
 using Moq;
 using NUnit.Framework;
-using NzbDrone.Common.Http;
-using NzbDrone.Core.MediaFiles.TorrentInfo;
 using NzbDrone.Core.Download;
 using NzbDrone.Core.Download.Clients.Transmission;
 
 namespace NzbDrone.Core.Test.Download.DownloadClientTests.TransmissionTests
 {
     [TestFixture]
-    public class TransmissionFixture : DownloadClientFixtureBase<Transmission>
+    public class TransmissionFixture : TransmissionFixtureBase<Transmission>
     {
-        protected TransmissionSettings _settings;
-        protected TransmissionTorrent _queued;
-        protected TransmissionTorrent _downloading;
-        protected TransmissionTorrent _failed;
-        protected TransmissionTorrent _completed;
-        protected TransmissionTorrent _magnet;
-        protected Dictionary<string, object> _transmissionConfigItems;
-
-        [SetUp]
-        public void Setup()
-        {
-            _settings = new TransmissionSettings
-            {
-                Host = "127.0.0.1",
-                Port = 2222,
-                Username = "admin",
-                Password = "pass"
-            };
-
-            Subject.Definition = new DownloadClientDefinition();
-            Subject.Definition.Settings = _settings;
-
-            _queued = new TransmissionTorrent
-                    {
-                        HashString = "HASH",
-                        IsFinished = false,
-                        Status = TransmissionTorrentStatus.Queued,
-                        Name = _title,
-                        TotalSize = 1000,
-                        LeftUntilDone = 1000,
-                        DownloadDir = "somepath"
-                    };
-
-            _downloading = new TransmissionTorrent
-                {
-                    HashString = "HASH",
-                    IsFinished = false,
-                    Status = TransmissionTorrentStatus.Downloading,
-                    Name = _title,
-                    TotalSize = 1000,
-                    LeftUntilDone = 100,
-                    DownloadDir = "somepath"
-                };
-
-            _failed = new TransmissionTorrent
-                    {
-                        HashString = "HASH",
-                        IsFinished = false,
-                        Status = TransmissionTorrentStatus.Stopped,
-                        Name = _title,
-                        TotalSize = 1000,
-                        LeftUntilDone = 100,
-                        ErrorString = "Error",
-                        DownloadDir = "somepath"
-                    };
-
-            _completed = new TransmissionTorrent
-                    {
-                        HashString = "HASH",
-                        IsFinished = true,
-                        Status = TransmissionTorrentStatus.Stopped,
-                        Name = _title,
-                        TotalSize = 1000,
-                        LeftUntilDone = 0,
-                        DownloadDir = "somepath"
-                    };
-
-            _magnet = new TransmissionTorrent
-            {
-                HashString = "HASH",
-                IsFinished = false,
-                Status = TransmissionTorrentStatus.Downloading,
-                Name = _title,
-                TotalSize = 0,
-                LeftUntilDone = 100,
-                DownloadDir = "somepath"
-            };
-
-            Mocker.GetMock<ITorrentFileInfoReader>()
-                  .Setup(s => s.GetHashFromTorrentFile(It.IsAny<byte[]>()))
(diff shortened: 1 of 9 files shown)


hmemcpy:
@markus101 Hey, would you mind reviewing this PR?

src/NzbDrone.Core/Download/Clients/Vuze/Vuze.cs
```diff
@@ -0,0 +1,71 @@
...
+                    IConfigService configService,
+                    IDiskProvider diskProvider,
+                    IRemotePathMappingService remotePathMappingService,
+                    Logger logger)
+            : base(proxy, torrentFileInfoReader, httpClient, configService, diskProvider, remotePathMappingService, logger)
+        {
+            _proxy = proxy;
+        }
+
+        protected override DownloadClientItem CreateDownloadItem(TransmissionTorrent torrent, OsPath outputPath)
```
markus101: Since this method is exactly the same as the base method except for the path `GetOutputPath` would be a better method to override in the Vuze class. Vuze would return the path that its reporting and Transmission would return the path + torrent name. Less code moving and you don't need to alter a path only to recreate it.

src/NzbDrone.Core/Download/Clients/Vuze/Vuze.cs
```diff
@@ -0,0 +1,71 @@
...
+        {
+            // Vuze reports outputPath containing the torrent name already, so we need to subtract it
+            // for the Transmission protocol compatibility
+
+            if (outputPath.FullPath.EndsWith(torrent.Name, StringComparison.Ordinal))
+            {
+                var startOfTorrentName = outputPath.FullPath.LastIndexOf(torrent.Name, StringComparison.Ordinal);
+                outputPath = new OsPath(outputPath.FullPath.Substring(0, startOfTorrentName));
```
markus101: This will be dead, but for what its worth you could do:

`outputPath = new OsPath(outputPath.FullPath.Substring(0, outputPath.FullPath.Length - torrent.Name .Length));`

src/NzbDrone.Core/Download/Clients/Vuze/Vuze.cs
```diff
@@ -0,0 +1,71 @@
...
+using NzbDrone.Core.Configuration;
+using NzbDrone.Core.Download.Clients.Transmission;
+using NzbDrone.Core.MediaFiles.TorrentInfo;
+using NzbDrone.Core.RemotePathMappings;
+
+namespace NzbDrone.Core.Download.Clients.Vuze
+{
+    public class Vuze : Transmission.Transmission
+    {
+        private const int MINIMAL_SUPPORTED_PROTOCOL_VERSION = 14;
```
markus101: `MINIMUM` instead of `MINIMAL`

markus101:
The `Transmission` class should become an abstract class named `TransmissionBase` then a new `Transmission` class that will implement `TransmissionBase`, `Vuze` will implement the base class as well.

src/NzbDrone.Core.Test/Download/DownloadClientTests/TransmissionTests/TransmissionFixture.cs
```diff
@@ -444,13 +261,13 @@ public void should_fix_forward_slashes()
         [TestCase("2.84+ ()")]
         [TestCase("2.84 (other info)")]
         [TestCase("2.84 (2.84)")]
-        public void should_version_should_only_check_version_number(string version)
+        public void should_check_version_number(string version)
```
markus101: `should_only_check_version_number`

src/NzbDrone.Core.Test/Download/DownloadClientTests/TransmissionTests/TransmissionFixtureBase.cs
```diff
@@ -0,0 +1,197 @@
...
+                .Returns<HttpRequest>(r => new HttpResponse(r, new HttpHeader(), new byte[0]));
+
+            _transmissionConfigItems = new Dictionary<string, object>();
+
+            _transmissionConfigItems.Add("download-dir", @"C:/Downloads/Finished/transmission");
+            _transmissionConfigItems.Add("incomplete-dir", null);
+            _transmissionConfigItems.Add("incomplete-dir-enabled", false);
+
+            Mocker.GetMock<ITransmissionProxy>()
+                .Setup(v => v.GetConfig(It.IsAny<TransmissionSettings>()))
```
markus101: Align on `.`

src/NzbDrone.Core.Test/Download/DownloadClientTests/VuzeTests/VuzeFixture.cs
```diff
@@ -0,0 +1,60 @@
...
+
+            Subject.Test().IsValid.Should().BeFalse();
+        }
+
+        [Test]
+        public void should_have_correct_output_directory()
+        {
+            WindowsOnly();
+
+            // Vuze reports DownloadDir including the torrent name
```
markus101: I think this comment is given since its in the Vuze fixture.

src/NzbDrone.Core/Download/Clients/Transmission/Transmission.cs
```diff
@@ -204,24 +212,14 @@ private ValidationFailure TestConnection()
...
-                }
+                return ValidateVersion();
             }
             catch (DownloadClientAuthenticationException ex)
             {
                 _logger.Error(ex, ex.Message);
                 return new NzbDroneValidationFailure("Username", "Authentication failure")
                 {
-                    DetailedDescription = "Please verify your username and password. Also verify if the host running Sonarr isn't blocked from accessing Transmission by WhiteList limitations in the Transmission configuration."
+                    DetailedDescription = string.Format("Please verify your username and password. Also verify if the host running Sonarr isn't blocked from accessing {0} by WhiteList limitations in the {0} configuration.", Name)
```
markus101: I'm tempted to suggest moving this whole method to the concrete class, but perhaps that is overkill and isn't very DRY.

hmemcpy:
@markus101 Thanks for the comments! I fixed it! :)

hmemcpy:
Bump :)

markus101:
Sorry @hmemcpy thanks for implementing this. Its been squashed and merged.

hmemcpy:
Awesome! Thanks a lot :)

PR #1285, opened 2016-05-08
Correctly report local time for HDB

needo37:
Copied from the BTN Parser.


src/NzbDrone.Core/Indexers/HDBits/HDBitsParser.cs
```diff
@@ -63,7 +63,7 @@ public IList<ReleaseInfo> ParseResponse(IndexerResponse indexerResponse)
                     InfoUrl = GetInfoUrl(id),
                     Seeders = result.Seeders,
                     Peers = result.Leechers + result.Seeders,
-                    PublishDate = result.Added
+                    PublishDate = new DateTime(1970, 1, 1, 0, 0, 0, 0, DateTimeKind.Utc).ToUniversalTime().AddSeconds(torrent.Time);
```
markus101: `torrent` isn't defined so this won't compile. The trailing `;` won't work either. The reason this works for BTN is they are returning the time as Epoch, HDB returns a formatted date.

PR #1310, opened 2016-05-21
Replace octal characters from mounts in /proc/mounts

markus101:
Also added some tests and we're now using `DiskProvider` for `ProcMountProvider`.


src/NzbDrone.Mono/ProcMountProvider.cs
```diff
@@ -17,23 +18,26 @@ public class ProcMountProvider : IProcMountProvider
...
         private static string[] _networkDriveTypes = new [] { "cifs", "nfs", "nfs4", "nfsd", "sshfs" };
+        private static readonly Regex OctalRegex = new Regex(@"\\\d{3}", RegexOptions.Compiled);
 
         private static Dictionary<string, bool> _fileSystems;
 
+        private readonly IDiskProvider _diskProvider;
         private readonly Logger _logger;
 
-        public ProcMountProvider(Logger logger)
+        public ProcMountProvider(IDiskProvider diskProvider, Logger logger)
```
Taloth: Can't do that, circular dependency. ProcMountProvider is part of the Mono DiskProvider implementation, basically a composition. That's why it was allowed to directly access System.IO.File and such.

src/NzbDrone.Mono/ProcMountProvider.cs
```diff
@@ -94,7 +98,7 @@ private IMount ParseLine(string line)
             }
 
             var name = split[0];
-            var mount = split[1];
+            var mount = GetMount(split[1]);
```
Taloth: Why not pipe the `split.Select(ExpandEscapes).ToArray()`? Strictly speaking these escapes could be in any field.

src/NzbDrone.Mono/ProcMountProvider.cs
markus101: After we talked I looked at moving `GetMounts` to its own interface, but `GetMounts` is used in a number of places in DiskProvider, so I don't think its worth the effort of unwinding it all. I didn't see this circular dependency and it managed to pass all the tests on TeamCity, maybe we can rename this to something that indicates its a composition in a better way.

PR #1337, opened 2016-06-12
Reenabled calling synoindex after renaming

tutipeti:
AfterRename/OnRename was implemented a year ago already when it was
disabled in 0f2bba061537e2de72f560b8c474f366201a2ab0, I assume it was
by mistake.

Right now it's super annoying, as I have to update the synoindex library manually after every single renaming done in Sonarr.


markus101:
Thanks for catching and fixing this!

PR #1351, opened 2016-06-24
Fix adding label to Sonarr initiated torrents in qBittorrent 3.3.5

### LINKED ISSUE #1347: qBittorrent Labels now called Categories in 3.3.5
qBittorrent has renamed labels as categories in version 3.3.5 and it appears to be causing downloads to not be assigned to the appropriate label and thus to not be detected by Sonarr.

vintage81:
Fix #1347 for the new (3.3.5) version of qBittorent for which the api for setting torrent labels has changed from setLabel to setCategory. Get torrent label still has the same api call as before, so doesn't need updating.

### DIFF
--- a/src/NzbDrone.Core/Download/Clients/qBittorrent/QBittorrentProxy.cs
+++ b/src/NzbDrone.Core/Download/Clients/qBittorrent/QBittorrentProxy.cs
@@ -94,12 +94,26 @@ namespace NzbDrone.Core.Download.Clients.QBittorrent
 
         public void SetTorrentLabel(string hash, string label, QBittorrentSettings settings)
         {
-            var request = BuildRequest(settings).Resource("/command/setLabel")
-                                                .Post()
-                                                .AddFormParameter("hashes", hash)
-                                                .AddFormParameter("label", label);
-
-            ProcessRequest<object>(request, settings);
+            var setCategoryRequest = BuildRequest(settings).Resource("/command/setCategory")
+                                                        .Post()
+                                                        .AddFormParameter("hashes", hash)
+                                                        .AddFormParameter("category", label);
+            try
+            { 
+                ProcessRequest<object>(setCategoryRequest, settings);
+            }
+            catch(DownloadClientException ex)
+            {
+                // if setCategory fails due to method not being found, then try older setLabel command for qbittorent < v.3.3.5
+                if (ex.InnerException is HttpException && (ex.InnerException as HttpException).Response.StatusCode == HttpStatusCode.NotFound)
+                {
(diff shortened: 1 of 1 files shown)


Taloth:
Maybe a version check to remain backwards compatible? Or some other mechanism to accomplish the same thing?

markus101:
With 3.3.5 based on what I know `setLabel` seems to succeed (or Sonarr ignores the error) (if they show up in history then its not causing issues, just silently failing), if thats the case does `setCategory` succeed on 3.3.4 and lower? If so we could try both and not worry about the version check, but only if both are successful. In any event it needs to be backwards compatible.

vintage81:
Thanks for the feedback both. Will look into the effects of using setLabel on 3.3.5 and setCategory on 3.3.4 and lower and then either add in a version check or code to perform both actions.

Taloth:
Try setCategory on 3.3.4, it probably throws an error. Which means you can first do setCategory and fallback to setLabel only when needed.

src/NzbDrone.Core/Download/Clients/qBittorrent/QBittorrentProxy.cs
```diff
@@ -94,12 +94,26 @@ public void RemoveTorrent(string hash, Boolean removeData, QBittorrentSettings s
...
+                                                        .AddFormParameter("hashes", hash)
+                                                        .AddFormParameter("category", label);
+            try
+            { 
+                ProcessRequest<object>(setCategoryRequest, settings);
+            }
+            catch(DownloadClientException ex)
+            {
+                // if setCategory fails due to method not being found, then try older setLabel command for qbittorent < v.3.3.5
+                if (ex.InnerException is HttpException && (ex.InnerException as HttpException).Response.StatusCode == HttpStatusCode.NotFound)
```
markus101: Since `as` won't throw if its not that type, we would use the following:

```
var httpException = ex.InnerException as HttpException;

if (httpException && httpException.Response.StatusCode == HttpStatusCode.NotFound)
```

It makes it a little more clear as its spread out a bit. With C# 6 I think we could even do:

```
if ((ex.InnerException as HttpException)?.Response.StatusCode == HttpStatusCode.NotFound)
```

(https://msdn.microsoft.com/en-CA/library/dn986595.aspx)

markus101: Realized we're doing this in this file already and my JS got into my C#.

markus101:
Thanks for fixing this!

vintage81:
No problem :)

Issue #1050, opened 2016-01-06
After 18-48hrs Sonarr cannot connect to downloaders


994wemj:
I have Sonarr installed on a FreeNAS server jail shared with couch potato: 192.168.0.133
nzbget and transmission are installed on a seperate jail: 192.168.0.139

Every day I have to switch these jails off and on again as each day I get home from work to find that Sonarr can no longer communicate with nzbget or transmission. When I test the connections to the downloaders within the Sonarr UI it says "Test was aborted due to an error: Unable to connect to NzbGet. Error: ConnectFailure (Connection refused)." Here are the log outputs from client: http://pastebin.com/rYhEkBAB

EDIT: additional log from a separate attempt, this time from ssh session: http://pastebin.com/XatNJG76

Couchpotato continues to be able to communicate with both NZBGet and Transmission when Sonarr is not. This tested within cp by intiating downloads, not just clicking the test button.

Rebooting the jails (sometimes multiple times required) typically resolves the problem but sometimes I have to reboot the entire FreeNAS in order to restore the connectivity and it works beautifully again for a day, sometimes two. This is obviously not right. I've actually reinstalled Sonarr onto this most recent jail after it happened before so it must be something in the way I have configured something?

Any help is much appreciated.

FreeNAS-9.3-STABLE-201512121950
Sonarr 2.0.0.3645
Transmission 2.84 (14307)
NZBGet 14.1

PR #1356, opened 2016-06-30
Recreate log database if migration fails

markus101:
Fixes: #1050 

Pretty simplistic approach, happy to discuss alternatives.

### DIFF
--- a/src/NzbDrone.Core/Datastore/DbFactory.cs
+++ b/src/NzbDrone.Core/Datastore/DbFactory.cs
@@ -1,11 +1,14 @@
 ﻿using System;
 using System.Data.SQLite;
+using System.IO;
 using Marr.Data;
 using Marr.Data.Reflection;
+using NLog;
 using NzbDrone.Common.Composition;
+using NzbDrone.Common.Disk;
+using NzbDrone.Common.EnvironmentInfo;
+using NzbDrone.Common.Instrumentation;
 using NzbDrone.Core.Datastore.Migration.Framework;
-using NzbDrone.Core.Instrumentation;
-using NzbDrone.Core.Messaging.Events;
 
 
 namespace NzbDrone.Core.Datastore
@@ -18,8 +21,10 @@ namespace NzbDrone.Core.Datastore
 
     public class DbFactory : IDbFactory
     {
+        private static readonly Logger Logger = NzbDroneLogger.GetLogger(typeof(DbFactory));
         private readonly IMigrationController _migrationController;
         private readonly IConnectionStringFactory _connectionStringFactory;
+        private readonly IDiskProvider _diskProvider;
 
         static DbFactory()
         {
@@ -38,10 +43,13 @@ namespace NzbDrone.Core.Datastore
             container.Register<ILogDatabase>(logDb);
         }
 
-        public DbFactory(IMigrationController migrationController, IConnectionStringFactory connectionStringFactory)
+        public DbFactory(IMigrationController migrationController,
+                         IConnectionStringFactory connectionStringFactory,
+                         IDiskProvider diskProvider)
         {
             _migrationController = migrationController;
             _connectionStringFactory = connectionStringFactory;
+            _diskProvider = diskProvider;
         }
 
         public IDatabase Create(MigrationType migrationType = MigrationType.Main)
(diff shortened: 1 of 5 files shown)


src/NzbDrone.Core/Datastore/DbFactory.cs
```diff
@@ -20,6 +23,10 @@ public class DbFactory : IDbFactory
     {
         private readonly IMigrationController _migrationController;
         private readonly IConnectionStringFactory _connectionStringFactory;
+        private readonly IDiskProvider _diskProvider;
+        private readonly IAppFolderInfo _appFolderInfo;
+        private static readonly Logger Logger = NzbDroneLogger.GetLogger(typeof(DbFactory));
```
Taloth: statics first please.

src/NzbDrone.Core/Datastore/DbFactory.cs
```diff
@@ -72,7 +84,30 @@ public IDatabase Create(MigrationContext migrationContext)
...
+            catch (SQLiteException ex)
+            {
+                var fileName = _connectionStringFactory.GetDatabasePath(connectionString);
+
+                if (migrationContext.MigrationType == MigrationType.Log)
+                {
+                    Logger.Warn(ex, "Logging database is corrupt, attempting to recreate it automatically");
+
+                    _diskProvider.DeleteFile(fileName);
+                    _diskProvider.DeleteFile(fileName + "-shm");
```
Taloth: euhm... the corruption happens on mac,... which doesn't use wal, instead -journal.

src/NzbDrone.Core/Datastore/DbFactory.cs
```diff
@@ -72,7 +84,30 @@ public IDatabase Create(MigrationContext migrationContext)
...
+            {
+                _migrationController.Migrate(connectionString, migrationContext);
+            }
+            catch (SQLiteException ex)
+            {
+                var fileName = _connectionStringFactory.GetDatabasePath(connectionString);
+
+                if (migrationContext.MigrationType == MigrationType.Log)
+                {
+                    Logger.Warn(ex, "Logging database is corrupt, attempting to recreate it automatically");
```
Taloth: Error seems more appropriate. also include link to wiki.

src/NzbDrone.Core/Datastore/DbFactory.cs
```diff
@@ -72,7 +84,30 @@ public IDatabase Create(MigrationContext migrationContext)
...
+
+                    _diskProvider.DeleteFile(fileName);
```
Taloth: How about a try catch around the whole thing? I don't want an support question about something that doesn't delete.
Also might wanna delete -wal and -shm first.

src/NzbDrone.Core/Datastore/DbFactory.cs
```diff
@@ -72,7 +84,30 @@ public IDatabase Create(MigrationContext migrationContext)
...
+                    _diskProvider.DeleteFile(fileName + "-shm");
+                    _diskProvider.DeleteFile(fileName + "-wal");
+
+                    _migrationController.Migrate(connectionString, migrationContext);
+                }
+
+                else
+                {
+                    throw new CorruptDatabaseException("Database file: {0} is corrupt, restore from backup if available", ex, fileName);
```
Taloth: A link to the wiki seems prudent.

src/NzbDrone.Core/Datastore/Migration/Framework/MigrationController.cs
```diff
@@ -37,16 +38,27 @@ public void Migrate(string connectionString, MigrationContext migrationContext)
...
+                {
+                    runner.MigrateUp(migrationContext.DesiredVersion.Value, true);
+                }
+                else
+                {
+                    runner.MigrateUp(true);
+                }
             }
-            else
+            catch (SQLiteException ex)
```
Taloth: nitpick: remove 'ex'. pretty sure it's a compile warning.

Taloth:
I think it might be a good idea to get some 'safe-mode' to ensure the webui shows up and gives an startup error while keeping all logic shut down. I'm stereotyping, but most mac users report 'webui not showing up' and never think to look at the logs.

src/NzbDrone.Core/Datastore/DbFactory.cs
markus101: I went with warning because we're attempting to fix it, if it fails again we'll end up logging another error, which we can catch and throw a bit nicer.

src/NzbDrone.Core/Datastore/DbFactory.cs
markus101: Good idea, will do.

src/NzbDrone.Core/Datastore/DbFactory.cs
markus101: I wasn't sure on the file name, I will delete `-journal` as well.

markus101:
I like the idea of a safe-mode UI, doesn't even have to be complicated, just have some basic information, though without any protection I'd want to keep it limited. Damn you and your good ideas making us do more work.

src/NzbDrone.Core/Datastore/DbFactory.cs
Taloth: It's an Error, coz either way ppl have lost data. It would've been Warn if you recovered the data.

Taloth:
The safe-mode can be simple, just a splash screen with a warning, version, path of the appdir & log dir etc.

Taloth:
Looks good, squash and ship it. File a separate GHI for the safemode UI.

PR #1357, opened 2016-06-30
Put.io Integration

jacobwgillespie:
_Work in progress, I am opening this PR early so that others can comment on the code as it develops._

This begins to implement Put.io as a downloader, extending the work started by @datashaman.

**Description**

Users would sign up for Put.io, create an application, and copy their OAuth token into Sonarr.  For automatic download handling, users would need to mount Put.io as an FTP volume and configure remote path mapping.

**Status**

Downloading works correctly, as does progress tracking, however the actual importing does not as of yet work.

**Expansion**

Some ideas for how this could be more user-friendly:
- Perhaps create a Put.io app for all of Sonarr, and provide a link to start the OAuth flow inside Settings > Downloaders.  Obviously privacy is an issue, so this would only work if Put.io offers an OAuth flow that displays the token for users to copy, rather than redirect back to the initiating URL.  Also would have to check on API rate limits per token, as that might this unfeasible.
- Allow Sonarr to browse / download files directly from Put.io after completion via their API.  This might introduce too much complexity, but would be a large win for user experience (no need to maintain a mapped volume).
- Possibly write an open-source script that automates adding Put.io downloads to Sonarr via API on completion.  This could be nicer than the mapped volume but would be more complex than automatic import built into Sonarr.

Any thoughts / feedback / questions are appreciated.


cosades:
Any news about this feature ?

jacobwgillespie:
I haven't done any work on this and no longer use put.io personally - happy to have somebody else take over this PR.

cosades:
I think I will use put.io soon, but I need plugins for tools I use.
I'm am reading your code. Maybe we could have this PR done one day. ;)

Unfortunality, the tests cannot work because of missing class `PutioTorrentStatus `(see test inside `PutioFixture.cs`)
I know it was a long time ago for you but did you make an error during the clean up in commit 4a8adc5cd9854d54ed2d849b419f272590adf09b ?

jacobwgillespie:
Not totally sure if it'll pass, but I pushed up a version with the tests updated.

cosades:
Thank you ;)
I'll test tonight.

markus101:
Closing as this hasn't seen any progress in some time, if someone wants to pick it up, feel free and send in a PR.

Issue #280, opened 2015-03-24
Releases with Hardcoded Subtitles


SonarrBot:
This has come up a couple times, both VOSTFR and NL-Subs are hardcoded subtitles (at least in my experience).

Personally I'd rather find another release than get hardcoded subs (its distracting when watching), but others are okay with it.

Right now subbed releases are considered other languages (depending on the sub language), which isn't exactly right.

Should we separate subbed language from language or just not worry about hardcoded subs (exclude them another way)?

Example names:

Elementary.S02E24.VOSTFR.HDTV.x264-BRN.[Seedbox].mp4
Helix S01E01 1080p BluRay NL-Subs

raspdealer:
Hi,

Is that thing done ?
I met this problem...

Twanislas:
Hi there !

Having a quick look at the code and such, why not add a second french language setting, thus ending with "French (VOSTFR)" and "French" choices ?

I find it very frustrating to have Sonarr download VOSTFR (English audio, french subtitles) instead of French audio. As line 243 of parser.cs could be easily modified to at least reflect what the french language setting means (-> download movies in french).

Why not change this : `(?<french>(?:\W|_)(?:FR|VOSTFR)(?:\W|_))` (which will match releases with "fr" but not "french", which is already a failure...) to this : `(?<french>(?:\W|_)(?:fr|french|vf|truefrench)(?:\W|_))` for the "French (VF)" setting ?

And then add `(?<french_vost>(?:\W|_)(?:vostfr)(?:\W|_))` for the "French (VOSTFR)" setting ?

Just my 2ct's ;) Hope it helps !

A.

Gromak123:
Hi, 

Is this problem patched ? Cause I juste install Sonarr and change profile setting to French and he downloaded me a lot of Vostfr and not VF :/ this is really a problem when you have some people who don't speak English at all at home ..

Twanislas:
Hi !

From what I understand (and that's how Sonarr works for me), French means anything with "french", "fr" or "vostfr" in the title. So it will match french audio and english audio with french subs.

For my personal case I don't really mind as I speak both but for some show that my GF follows (she doesn't speak english very well) I just added a tag in the indexers with a "must not contain 'vostfr'" restriction. I just apply this label to the shows I want only with french audio and keep using the same profile as normal shows.

Hope this helps :)

Cheers from Belgium !

Gromak123:
[...]

PR #1368, opened 2016-07-06
French Parsing

GiroudMathias:
Update French Parsing

### DIFF
--- a/src/NzbDrone.Core/Parser/Parser.cs
+++ b/src/NzbDrone.Core/Parser/Parser.cs
@@ -240,7 +240,7 @@ namespace NzbDrone.Core.Parser
         private static readonly Regex AnimeReleaseGroupRegex = new Regex(@"^(?:\[(?<subgroup>(?!\s).+?(?<!\s))\](?:_|-|\s|\.)?)",
                                                                 RegexOptions.IgnoreCase | RegexOptions.Compiled);
 
-        private static readonly Regex LanguageRegex = new Regex(@"(?:\W|_)(?<italian>\b(?:ita|italian)\b)|(?<german>german\b|videomann)|(?<flemish>flemish)|(?<greek>greek)|(?<french>(?:\W|_)(?:FR|VOSTFR)(?:\W|_))|(?<russian>\brus\b)|(?<dutch>nl\W?subs?)|(?<hungarian>\b(?:HUNDUB|HUN)\b)",
+        private static readonly Regex LanguageRegex = new Regex(@"(?:\W|_)(?<italian>\b(?:ita|italian)\b)|(?<german>german\b|videomann)|(?<flemish>flemish)|(?<greek>greek)|(?<french>(?:\W|_)(?:FRENCH|VFF|VFQ)(?:\W|_))|(?<russian>\brus\b)|(?<dutch>nl\W?subs?)|(?<hungarian>\b(?:HUNDUB|HUN)\b)",
                                                                 RegexOptions.IgnoreCase | RegexOptions.Compiled);
 
         private static readonly Regex YearInTitleRegex = new Regex(@"^(?<title>.+?)(?:\W|_)?(?<year>\d{4})",


src/NzbDrone.Core/Parser/Parser.cs
```diff
@@ -240,7 +240,7 @@ public static class Parser
         private static readonly Regex AnimeReleaseGroupRegex = new Regex(@"^(?:\[(?<subgroup>(?!\s).+?(?<!\s))\](?:_|-|\s|\.)?)",
                                                                 RegexOptions.IgnoreCase | RegexOptions.Compiled);
 
-        private static readonly Regex LanguageRegex = new Regex(@"(?:\W|_)(?<italian>\b(?:ita|italian)\b)|(?<german>german\b|videomann)|(?<flemish>flemish)|(?<greek>greek)|(?<french>(?:\W|_)(?:FR|VOSTFR)(?:\W|_))|(?<russian>\brus\b)|(?<dutch>nl\W?subs?)|(?<hungarian>\b(?:HUNDUB|HUN)\b)",
+        private static readonly Regex LanguageRegex = new Regex(@"(?:\W|_)(?<italian>\b(?:ita|italian)\b)|(?<german>german\b|videomann)|(?<flemish>flemish)|(?<greek>greek)|(?<french>(?:\W|_)(?:FRENCH|VFF|VFQ)(?:\W|_))|(?<russian>\brus\b)|(?<dutch>nl\W?subs?)|(?<hungarian>\b(?:HUNDUB|HUN)\b)",
```
markus101: French is already handled below.

src/NzbDrone.Core/Parser/Parser.cs
```diff
@@ -240,7 +240,7 @@ public static class Parser
         private static readonly Regex AnimeReleaseGroupRegex = new Regex(@"^(?:\[(?<subgroup>(?!\s).+?(?<!\s))\](?:_|-|\s|\.)?)",
                                                                 RegexOptions.IgnoreCase | RegexOptions.Compiled);
 
-        private static readonly Regex LanguageRegex = new Regex(@"(?:\W|_)(?<italian>\b(?:ita|italian)\b)|(?<german>german\b|videomann)|(?<flemish>flemish)|(?<greek>greek)|(?<french>(?:\W|_)(?:FR|VOSTFR)(?:\W|_))|(?<russian>\brus\b)|(?<dutch>nl\W?subs?)|(?<hungarian>\b(?:HUNDUB|HUN)\b)",
```
markus101: Why are `FR` and `VOSTFR` being removed?

markus101:
Tests for these need to be added to the `LanguageParserFixture`.

src/NzbDrone.Core/Parser/Parser.cs
GiroudMathias: the french torrents are generally written "FRENCH, VFF VFQ".

Leaving the VOSTFR and GB, the files are downloaded in English and therefore not in french

Twanislas:
Hi !

Please see my comment in #280 ;)

Would'nt an even better regexp be : `(?<french>(?:\W|_)(?:fr|french|vf|truefrench|vff|vfq)(?:\W|_))` ?

Cheers !

JigSawFr:
hm, why removing VOSTFR ? I'm using it to download all serie's episodes...
if not available in french, it use VOSTFR actually

Twanislas:
Hi @JigSawFr !

Actually I see this more as a bug than a feature. While it's nice to have it "magically" fallback to VOSTFR, some people may want to download a series only in french.

Moreover, this could end up in a mess with one episode on 2 being VOSTFR and the other in VF. Those of us who are OCD'ed about having a nice and perfeclty organized library would be very upset :D

Yet one could add a tag to a series to exclude the VOSTFR word from search results but then it would mean overriding the behavior we expect when we instruct the software to download things in french.

As I see it, if I put the language setting to "French", that's what I expect Sonarr to download. If I wanted VOSTFR, I'd set it to english, and add a filter. Or try to add another language in Sonarr for that (but that may be overkill ?)

Cheers !

Twanislas:
Sent PR #1389.

Cheers :) !

Taloth:
@GiroudMathias @Twanislas I'm gonna respond to both your PRs right here. As it stands, neither are acceptable to us.

The problem is that you may not consider VOSTFR to be French, but it basically is. Just like NLSub is Dutch. Both may be 'just' subtitles, but the majority of users don't want those.
It's notoriously hard to separate video/audio language and subtitles.
Also, early NLSub releases actually were hardcoded subs, not embedded srts.

Removing those from the language detection basically forces everyone to add Restrictions, which kinda  defeats the goal of Sonarr aiming to be easy to use for 90% of the users.

So, lets think about a solution.
What we need to do is at least be able to make a distinction which language tags are video/audio and which are guaranteed subtitle tags. That's relatively easy, just create separate regexes.
Then we need to come up with a way to let the user decide whether he's okay with foreign sub tags.
Best place to do that atm is the Profile, but an global flag might be an option as well.

We had a PR that created separate Language profiles and multi-lang support, but that's gone (the submitter account is gone).
Sonarr doesn't support foreign language metadata atm, so there's little point in going all the way.

JigSawFr:
@Taloth 
Best way in my mind is possibility to download in French and fallback to subtitles (VOSTFR) if no releases found. And why not to English if nothing found. Maybe an order sytem for languages

Twanislas:
Hi !

While I like you idea @JigSawFr, how would you prevent episodes language mixup in the series ?

Actually, a language order system sounds super nice. Kinda the same system as we already have for qualities, where we can set a cutoff. So we can have early releases in VOSTFR which will be updated when the french version shows up (or not, depending on cutoff point). Me likey !

Tha would mean I don't have to lookup every show in my library to know wether they air in french or english and set tags/filters/restrictions accordingly show by show.

Cheers !

markus101:
Closing, but we're open for more discussion on this if anyone wishes to  start dialog there, please open a GHI.

PR #1370, opened 2016-07-07
WIP - Torrent RSS Feed Generator

equals03:
First implementation of a RSS feed generator client. As the name suggests it creates a feed of the downloads which can be consumed by torrent clients. I find it much more versatile than the blackhole client as this more tightly integrates with the activity/history section.

I haven't completed (or really even started) the tests. Just wanted to get some feedback. If its something the rest of the community wants, i'll get the tests up and running.

I've marked this WIP but its pretty feature complete and has been running well (for me) for the last few days.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -129,3 +129,4 @@ output/*
 
 _start
 _temp_*/**/*
+.vs
--- a/src/NzbDrone.Api/NzbDrone.Api.csproj
+++ b/src/NzbDrone.Api/NzbDrone.Api.csproj
@@ -77,6 +77,8 @@
       <SpecificVersion>False</SpecificVersion>
       <HintPath>..\Libraries\Sqlite\System.Data.SQLite.dll</HintPath>
     </Reference>
+    <Reference Include="System.Xml" />
+    <Reference Include="System.Xml.Linq" />
   </ItemGroup>
   <ItemGroup>
     <Compile Include="..\NzbDrone.Common\Properties\SharedAssemblyInfo.cs">
@@ -221,6 +223,7 @@
     <Compile Include="REST\RestResource.cs" />
     <Compile Include="RootFolders\RootFolderModule.cs" />
     <Compile Include="RootFolders\RootFolderResource.cs" />
+    <Compile Include="Rss\RssFeedModule.cs" />
     <Compile Include="SeasonPass\SeasonPassResource.cs" />
     <Compile Include="Series\AlternateTitleResource.cs" />
     <Compile Include="Series\SeasonResource.cs" />
--- a/src/NzbDrone.Api/Rss/RssFeedModule.cs
+++ b/src/NzbDrone.Api/Rss/RssFeedModule.cs
@@ -1,0 +1,148 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Xml.Linq;
+using Nancy;
+using Nancy.Responses;
+using NzbDrone.Common.Disk;
+using NzbDrone.Common.Extensions;
+using NzbDrone.Core.Download;
+using NzbDrone.Core.Download.Clients.RssGenerator;
+using NzbDrone.Core.Parser.Model;
+
+namespace NzbDrone.Api.Rss {
+
+    public class TorrentGeneratorRssFeedModule : NzbDroneFeedModule {
+
+        private class Rss20Formatter {
+            private const string DateTimeFormat = "ddd, dd MMM yyyy HH:mm:ss";
+            private XNamespace nsSonarr = "https://github.com/Sonarr/Sonarr";
+
+            private static readonly Rss20Formatter __instance = new Rss20Formatter();
+            public static Rss20Formatter Instance { get { return __instance; } }
+
+            public string ContentType { get { return "application/xml"; } }
+            public Encoding Encoding { get { return Encoding.UTF8; } }
+
+            private Rss20Formatter() {
+            }
+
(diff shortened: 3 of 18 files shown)


markus101:
Before looking at this, I'm curious why this would be useful? I don't see how this would fit in the Sonarr ecosystem without Sonarr being able to import it (Drone Factory is not a suitable import method and will be retired in v3).

> I find it much more versatile than the blackhole client as this more tightly integrates with the activity/history section.

How so? With blackhole history is written when the torrent is downloaded, with this I don't see a point where it would be acceptable for Sonarr to write to the history since it can't be assumed that the client will pick it up off the RSS feed. How does it integrate with activity?

Which clients would this be useful for? Wouldn't we be better off support those clients (assuming they have a useful API)?

equals03:
I'll get to the answers to your questions in turn, but let me start by giving some background...

I use Tixati for my torrents. I've been using the blackhole download client for a while now and I noticed two major shortcomings.

1. [primary among them] it does not support magnet links and,
1. while it would show imports in the history, it wouldn't show items in the activity queue (admittedly this could have been because of my setup - I didn't delve into the reasons too much) so I couldn't tell what was going on by just looking at Sonarr.

Over and above those, there are certain quirks in Tixati. For instance, it wouldn't always pick up the torrents in the blackhole folder, and when it did it would display a dialog (irritating). And you couldn't instruct it to assign those downloads to a default category.

The RSS processing in the client, however, is a whole lot better - with better automation. This and the fact the the blackhole doesn't support magnets lead me down the RSS path.

Using the implementation I have, the activity queue is populated (remaining through restarts) and can be maintained (items added/removed). It also listens to two different folders, one for "incomplete/in progress" and one for "completed". Doing some very light matching to determine if the client has started downloading or has completed the download.

So to address your questions directly:

> Before looking at this, I'm curious why this would be useful? I don't see how this would fit in the Sonarr ecosystem without Sonarr being able to import it (Drone Factory is not a suitable import method and will be retired in v3).

It does automated imports. By listening to the different folders, it works autonomously - from start to finish (or I wouldn't have bothered :smile: - I'm lazy like that). It watches the "incomplete" folder (if you have one - optional) to see if the client has started downloading it, and no use of Drone because it watches the "completed" folder and imports them if they exist. The implementation is simple though, if its in the "completed" folder its imported. No checking to see if the files are locked etc.

> How so? With blackhole history is written when the torrent is downloaded, with this I don't see a point where it would be acceptable for Sonarr to write to the history since it can't be assumed that the client will pick it up off the RSS feed. 

It doesn't write to the history directly. It acts as a download client (much like blackhole) and works completely within that framework. Sonarr picks up that the torrent/magnet has been given to the download client and Sonarr logs it into the activity queue as queued - when it detects its complete (moved to the completed folder) it lets Sonarr know and the import happens as per usual.  

> How does it integrate with activity?

As I said above... keeps to the framework. You can remove the item from the queue and it removes itself from the feed so the client won't download it. It also wont be monitored for collection from that point on even if the download client started and then finishes.

> Which clients would this be useful for? Wouldn't we be better off support those clients (assuming they have a useful API)?

Tixati would be top of that list for me. It has no api that I know of. And while you could probably hack its web interface - that seemed like a lot of effort and it was missing certain things like the category assignment. It has been requested, but it doesn't seem to be a priority for them.

So, why did I go to all this trouble? To support magnet links and make automating Tixati easier. I guess I could just have switched, but I like Tixati and I wanted to see if I could make it work :wink: 

I realise this could be pretty niche and YMMMV... but it works for me.

markus101:
Magnet Links:
This is due to what magnet links actually are, without connecting to the swarm they don't be converted to a torrent file and to my knowledge there isn't a common way to present a magnet link in a file to allow people the client to consume it like it would a torrent file. We do have some plans to get torrents from magnets that are available from torcache, which will help albeit not perfect.

Activity for Blackhole:
Sonarr has no reliable way to track it in the download client because there isn't one, file size vs expected final file size isn't reliable, especially when the entire file can be pre-filled to avoid running out of space mid-download.

> Over and above those, there are certain quirks in Tixati. For instance, it wouldn't always pick up the torrents in the blackhole folder, and when it did it would display a dialog (irritating). And you couldn't instruct it to assign those downloads to a default category.

That sounds awful and half baked... I don't think those are good reasons to add support for something thats similar to blackhole, but different at the same time (I can see the potential for confusion for users).

> You can remove the item from the queue and it removes itself from the feed so the client won't download it.

For a proper client it would also remove it from the queue of the client, in this scenario its only useful if the release is removed before the client fetches the RSS feed and in any other case it seems confusing (I clicked remove, but then it came back right away).

While we appreciate the effort we don't think this is a good fit for Sonarr, its too similar to a Blackhole client with some nuances that depend on specific client configuration and seem a bit awkward.

equals03:
> to my knowledge there isn't a common way to present a magnet link in a file to allow people the client to consume it like it would a torrent file.

Apparently Tixati can import a magnet from a .url file. Perhaps more of the clients do that and a simple implementation worth looking at? I was toying with adding it myself, but the lack of queue management (as you mention below) was one of the reasons I didn't.

> Sonarr has no reliable way to track it in the download client because there isn't one ...

Yup. One of the reasons I looked for alternatives.

> For a proper client it would also remove it from the queue of the client...

Doesn't blackhole suffer from the same issue?

> While we appreciate the effort we don't think this is a good fit for Sonarr, its too similar to a Blackhole client with some nuances that depend on specific client configuration and seem a bit awkward.

No problem and understood. I was quite aware that this kind of functionality would probably only be of interest only to myself. Enter the beauty of forks.

Great work on the project btw - it is hands down the best "pvr" like software out there imho :+1:

PR #1381, opened 2016-07-14
Fixed: XEM series with only alternate season titles.

Taloth:
Xem scene naming exceptions are stored in the db as seasonNumber: null sceneSeasonNumber: x.
But for episodes without mapping Sonarr doesn't have a SceneSeasonNumber specified for the episode, so wouldn't use the naming exception.

This commit basically uses episode.SeasonNumber as episode.SceneSeasonNumber if none was specified.

I included several tests to check the specific behavior, in the hopes of not breaking it in the future.

I also adjusted the UI to show the alt title on the episode scene popout.

### DIFF
--- a/src/NzbDrone.Core/DataAugmentation/Scene/SceneMappingService.cs
+++ b/src/NzbDrone.Core/DataAugmentation/Scene/SceneMappingService.cs
@@ -13,8 +13,7 @@ namespace NzbDrone.Core.DataAugmentation.Scene
 {
     public interface ISceneMappingService
     {
-        List<string> GetSceneNamesBySeasonNumbers(int tvdbId, IEnumerable<int> seasonNumbers);
-        List<string> GetSceneNamesBySceneSeasonNumbers(int tvdbId, IEnumerable<int> sceneSeasonNumbers);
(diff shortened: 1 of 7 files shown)


src/NzbDrone.Core/DataAugmentation/Scene/SceneMappingService.cs
```diff
@@ -49,32 +48,21 @@ public class SceneMappingService : ISceneMappingService,
...
         }
 
-        public List<string> GetSceneNamesBySeasonNumbers(int tvdbId, IEnumerable<int> seasonNumbers)
+        public List<string> GetSceneNames(int tvdbId, List<int> seasonNumbers, List<int> sceneSeasonNumbers)
         {
-            var names = FindByTvdbId(tvdbId);
+            var mappings = FindByTvdbId(tvdbId);
 
-            if (names == null)
+            if (mappings == null)
```
markus101: I realize this is mostly existing, but will we ever return null from `FindByTvdbId`?

markus101:
:+1: much better than my hacks :smile:

PR #1387, opened 2016-07-21
MediaInfo VideoBitDepth and AudioChannels

markus101:
#### Database Migration
NO - might need one

#### Description
New file naming tokens `MediaInfo VideoBitDepth` (10 Bit) and `MediaInfo AudioChannels` (5.1).

### DIFF
--- a/src/NzbDrone.Core/MediaFiles/MediaInfo/MediaInfoModel.cs
+++ b/src/NzbDrone.Core/MediaFiles/MediaInfo/MediaInfoModel.cs
@@ -1,4 +1,7 @@
 ﻿using System;
+using System.Globalization;
+using System.Linq;
+using NzbDrone.Common.Extensions;
 using NzbDrone.Core.Datastore;
 
 namespace NzbDrone.Core.MediaFiles.MediaInfo
@@ -7,6 +10,7 @@ namespace NzbDrone.Core.MediaFiles.MediaInfo
     {
         public string VideoCodec { get; set; }
         public int VideoBitrate { get; set; }
+        public int VideoBitDepth { get; set; }
         public int Width { get; set; }
         public int Height { get; set; }
         public string AudioFormat { get; set; }
@@ -14,10 +18,25 @@ namespace NzbDrone.Core.MediaFiles.MediaInfo
         public TimeSpan RunTime { get; set; }
         public int AudioStreamCount { get; set; }
         public int AudioChannels { get; set; }
+        public string AudioChannelPositions { get; set; }
         public string AudioProfile { get; set; }
         public decimal VideoFps { get; set; }
         public string AudioLanguages { get; set; }
         public string Subtitles { get; set; }
         public string ScanType { get; set; }
+        public int SchemaRevision { get; set; }
+
+        public decimal FormattedAudioChannels
+        {
+            get
+            {
+                if (AudioChannelPositions.IsNullOrWhiteSpace())
+                {
+                    return 0;
+                }
+
+                return AudioChannelPositions.Split('/').Sum(s => decimal.Parse(s, CultureInfo.InvariantCulture));
+            }
+        }
     }
 }
--- a/src/NzbDrone.Core/Organizer/FileNameBuilder.cs
+++ b/src/NzbDrone.Core/Organizer/FileNameBuilder.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using System.Globalization;
 using System.IO;
 using System.Linq;
 using System.Text.RegularExpressions;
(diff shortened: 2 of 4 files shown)


src/NzbDrone.Core/Organizer/FileNameBuilder.cs
```diff
@@ -518,12 +518,20 @@ private void AddMediaInfoTokens(Dictionary<string, Func<TokenMatch, string>> tok
                 mediaInfoSubtitleLanguages = string.Format("[{0}]", mediaInfoSubtitleLanguages);
             }
 
-            tokenHandlers["{MediaInfo Video}"] = m => mediaInfoVideo;
-            tokenHandlers["{MediaInfo Audio}"] = m => mediaInfoAudio;
+            var videoBitDepth = episodeFile.MediaInfo.VideoBitDepth.HasValue ? $"{episodeFile.MediaInfo.VideoBitDepth} Bit" : string.Empty;
+            var audioChannels = episodeFile.MediaInfo.AudioChannelPositions.IsNotNullOrWhiteSpace() ? episodeFile.MediaInfo.AudioChannelPositions.Split('/').Sum(s => decimal.Parse(s)).ToString() : string.Empty;
 
-            tokenHandlers["{MediaInfo Simple}"] = m => string.Format("{0} {1}", mediaInfoVideo, mediaInfoAudio);
+            tokenHandlers["{MediaInfo Video}"] = m => videoCodec;
```
markus101: Should we have a migration to convert the current name to `MediaInfo Video` to `MediaInfo VideoBitDepth`? Might not be worth the effort and could drop it in v3.

src/NzbDrone.Core/Organizer/FileNameBuilder.cs
```diff
@@ -518,12 +518,20 @@ private void AddMediaInfoTokens(Dictionary<string, Func<TokenMatch, string>> tok
...
+            tokenHandlers["{MediaInfo VideoCodec}"] = m => videoCodec;
+            tokenHandlers["{MediaInfo VideoBitDepth}"] = m => videoBitDepth;
 
-            tokenHandlers["{MediaInfo Full}"] = m => string.Format("{0} {1}{2} {3}", mediaInfoVideo, mediaInfoAudio, mediaInfoAudioLanguages, mediaInfoSubtitleLanguages);
+            tokenHandlers["{MediaInfo Audio}"] = m => audioCodec;
```
markus101: Same here.

markus101:
One thing I'm unsure about is how to update media info to get the new properties we need to support these tokens, we can either have an option to periodically update it (and a migration to store the last time media info was checked) or check if the properties we're adding exist and update if they don't and have an a similar option to rebuild it (or no option, but that could have adverse effects).

markus101:
Another option is to do nothing and leave it up to the user to either wipe of MediaInfo manually or another way to force Sonarr to rescan the files any existing files wouldn't have the info for the tokens so they would be excluded.

src/NzbDrone.Core/Organizer/FileNameBuilder.cs
Taloth: You mean Video -> VideoCodec? I would stick with Video. Bitdepth etc is rarely needed.

src/NzbDrone.Core/Organizer/FileNameBuilder.cs
```diff
@@ -518,12 +518,20 @@ private void AddMediaInfoTokens(Dictionary<string, Func<TokenMatch, string>> tok
                 mediaInfoSubtitleLanguages = string.Format("[{0}]", mediaInfoSubtitleLanguages);
             }
 
-            tokenHandlers["{MediaInfo Video}"] = m => mediaInfoVideo;
-            tokenHandlers["{MediaInfo Audio}"] = m => mediaInfoAudio;
+            var videoBitDepth = episodeFile.MediaInfo.VideoBitDepth.HasValue ? $"{episodeFile.MediaInfo.VideoBitDepth} Bit" : string.Empty;
+            var audioChannels = episodeFile.MediaInfo.AudioChannelPositions.IsNotNullOrWhiteSpace() ? episodeFile.MediaInfo.AudioChannelPositions.Split('/').Sum(s => decimal.Parse(s)).ToString() : string.Empty;
```
Taloth: Use CultureInvariant to handle the decimal point.
Also, you can add some fallback logic to convert AudioChannels (2,4,6,8 to 2.0, 4.0, 5.1, 7.1).
My suggestion would be to extract it to a separate method.

src/NzbDrone.Core/Organizer/FileNameBuilder.cs
```diff
@@ -518,12 +518,20 @@ private void AddMediaInfoTokens(Dictionary<string, Func<TokenMatch, string>> tok
                 mediaInfoSubtitleLanguages = string.Format("[{0}]", mediaInfoSubtitleLanguages);
             }
 
-            tokenHandlers["{MediaInfo Video}"] = m => mediaInfoVideo;
-            tokenHandlers["{MediaInfo Audio}"] = m => mediaInfoAudio;
+            var videoBitDepth = episodeFile.MediaInfo.VideoBitDepth.HasValue ? $"{episodeFile.MediaInfo.VideoBitDepth} Bit" : string.Empty;
```
Taloth: I would be opposed to hardcoding " Bit".

src/NzbDrone.Core/MediaFiles/MediaInfo/MediaInfoModel.cs
```diff
@@ -7,13 +7,15 @@ public class MediaInfoModel : IEmbeddedDocument
     {
         public string VideoCodec { get; set; }
         public int VideoBitrate { get; set; }
+        public int? VideoBitDepth { get; set; }
```
Taloth: Does it have to be nullable?

src/NzbDrone.Core/Organizer/FileNameBuilder.cs
markus101: >You mean Video -> VideoCodec?

Oops, yes.

src/NzbDrone.Core/Organizer/FileNameBuilder.cs
markus101: Yeah, I was a bit torn on that, no objection on dropping it.

src/NzbDrone.Core/MediaFiles/MediaInfo/MediaInfoModel.cs
markus101: No, it doesn't.

src/NzbDrone.Core/Organizer/FileNameBuilder.cs
markus101: > Also, you can add some fallback logic to convert AudioChannels (2,4,6,8 to 2.0, 4.0, 5.1, 7.1).

I thought about that, but I figured someone would manage to screw that up.

Taloth:
Embed the MediaInfoVersion in the MediaInfoModel? Saves a migration.
Also, I wouldn't add a config option for something that trivial.

markus101:
> Embed the MediaInfoVersion in the MediaInfoModel? Saves a migration.

Contemplated that as well, will do.

> Also, I wouldn't add a config option for something that trivial.

Was a bit torn on that, but erred on the side of doing less without being told to and allowing people to disable it if they don't want it enabled in their setup (but either wanted it initially or didn't notice). Keep the disk idle time people happy with a tiny bit of effort.

Taloth:
> Was a bit torn on that, but erred on the side of doing less without being told to and allowing people to disable it if they don't want it enabled in their setup (but either wanted it initially or didn't notice). Keep the disk idle time people happy with a tiny bit of effort.

I could think of better alternatives (such as lazy-loading) if that's such a big concern.
And it's not about disk idle time, it's a one time deal. It also won't do all series at once. It shouldn't be that excessive either, it reads less than 1% of each file.

My concern isn't the 'tiny bit of effort', it's that you're adding a config option for it. Something we don't normally do unless it's necessary.

PR #1398, opened 2016-07-26
Feature: Add the ability to receive Join notifications.

### LINKED ISSUE #1197: Support sending notifications via Join
Join is a new alternative similar to Pushbullet in its functionality. It would be great if Sonarr supported sending notifications via Join.

http://joaoapps.com/join/

Nedlinin:
#### Description
Allows a user to use the Join API to send a notification of actions taken by Sonarr to all of their devices.

Right now this only supports sending to every device on their account.  However, it could be expanded to allow sending only to specific devices by someone with a bit more time.

Fixes #1197

### DIFF
--- a/src/NzbDrone.Core/Notifications/Join/JoinProxy.cs
+++ b/src/NzbDrone.Core/Notifications/Join/JoinProxy.cs
@@ -1,0 +1,93 @@
+﻿using System;
+using System.Linq;
+using System.Net;
+using FluentValidation.Results;
+using NLog;
+using RestSharp;
+using NzbDrone.Core.Rest;
+using NzbDrone.Common.Serializer;
+
+namespace NzbDrone.Core.Notifications.Join
+{
+    public interface IJoinProxy
+    {
+        void SendNotification(string title, string message, JoinSettings settings);
+        ValidationFailure Test(JoinSettings settings);
+    }
+
+    public class JoinProxy : IJoinProxy
+    {
+        private readonly Logger _logger;
+        private const string URL = "https://joinjoaomgcd.appspot.com/_ah/api/messaging/v1/sendPush?";
+
+        public JoinProxy(Logger logger)
+        {
+            _logger = logger;
+        }
+
+        public void SendNotification(string title, string message, JoinSettings settings)
+        {
+            var request = new RestRequest(Method.GET);
+
+            try
+            {
+                SendNotification(title, message, request, settings);
+            }
+            catch (JoinException ex)
+            {
+                _logger.Error(ex, "Unable to send Join message.");
+                throw new JoinException("Unable to send Join notifications. " + ex.Message);
+            }
+        }
+
+        public ValidationFailure Test(JoinSettings settings)
+        {
+
+            const string title = "Test Notification";
+            const string body = "This is a test message from Sonarr.";
+            try
+            {
+                SendNotification(title, body, settings);
+                return null;
+            }
+            catch (JoinException ex)
+            {
+                _logger.Error(ex, "Unable to send test Join message.");
+                return new ValidationFailure("APIKey", ex.Message);
+            }
+        }
+
+        private void SendNotification(string title, string message, RestRequest request, JoinSettings settings)
+        {
+            try
+            {
+                var client = RestClientFactory.BuildClient(URL);
+
+                request.AddParameter("deviceId", "group.all");
+                request.AddParameter("apikey", settings.APIKey);
+                request.AddParameter("title", "[Sonarr] " + title);
+                request.AddParameter("text", message);
+                request.AddParameter("icon", "https://cdn.rawgit.com/Sonarr/Sonarr/develop/Logo/256.png"); // Use the Sonarr logo.
+
+                var response = client.ExecuteAndValidate(request);
+                var res = Json.Deserialize<JoinResponseModel>(response.Content);
+
+                if (res.success) return;
+
+                if (res.errorMessage != null)
+                {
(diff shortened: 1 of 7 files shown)


Nedlinin:
I should note that I am a JS developer by default so if I've made any mistakes please let me know.  This is heavily based on the Boxcar integration (thanks to whomever did that!)

src/NzbDrone.Core/Notifications/Join/JoinProxy.cs
```diff
@@ -0,0 +1,94 @@
...
+        {
+            try
+            {
+                var client = RestClientFactory.BuildClient(URL);
+
+                request.AddParameter("deviceId", "group.all");
+                request.AddParameter("apikey", settings.APIKey);
+                request.AddParameter("title", "[Sonarr] " + title);
+                request.AddParameter("text", message);
+                request.AddParameter("icon", "https://avatars3.githubusercontent.com/u/1082903"); // Use the Sonarr logo.
```
thezoggy: should use local file.. having every sonarr user hit github probably isnt ideal..

Nedlinin: @thezoggy : The Join API doesn't support sending a local file using Data URI or anything similar.  The only thing they support is a URL to an icon to display.

But, your comment did make me realize I wasn't using the Github CDN.  So I updated the PR to use the rawgit CDN.

src/NzbDrone.Core/Notifications/Join/JoinProxy.cs
```diff
@@ -0,0 +1,93 @@
...
+        {
+
+            const string title = "Test Notification";
+            const string body = "This is a test message from Sonarr.";
+            try
+            {
+                SendNotification(title, body, settings);
+                return null;
+            }
+            catch (JoinException ex)
```
markus101: Since we're throwing JoinExceptions for any exception this could be misleading if the test fails for another reason (Join servers are down for example). We'll still want to catch a JoinException, but we should also have a JoinAuthentication Exception that inherits from JoinException that we can throw when we're sure its an Authentication issue.

src/NzbDrone.Core/Notifications/Join/JoinProxy.cs
```diff
@@ -0,0 +1,93 @@
...
+
+                if (res.errorMessage != null)
+                {
+                    throw new JoinException(res.errorMessage);
+                }
+                if (res.userAuthError)
+                {
+                    throw new JoinException("Authentication failed.");
+                }
+                throw new JoinException("Unknown error.  Join message failed to send.");
```
markus101: Single space after the period. New line before this line.

src/NzbDrone.Core/Notifications/Join/JoinProxy.cs
```diff
@@ -0,0 +1,93 @@
...
+                var response = client.ExecuteAndValidate(request);
+                var res = Json.Deserialize<JoinResponseModel>(response.Content);
+
+                if (res.success) return;
+
+                if (res.errorMessage != null)
+                {
+                    throw new JoinException(res.errorMessage);
+                }
+                if (res.userAuthError)
```
markus101: Nit: New line before new if statement.

src/NzbDrone.Core/Notifications/Join/JoinSettings.cs
```diff
@@ -0,0 +1,29 @@
...
+        {
+            RuleFor(c => c.APIKey).NotEmpty();
+        }
+    }
+
+    public class JoinSettings : IProviderConfig
+    {
+        private static readonly JoinSettingsValidator Validator = new JoinSettingsValidator();
+
+        [FieldDefinition(0, Label = "API Key", HelpText = "The API Key from your Join account settings (click Join API button): https://joinjoaomgcd.appspot.com/", HelpLink = "https://joinjoaomgcd.appspot.com/")]
```
markus101: We don't need to add the URL here as the HelpLink field will be present (and they won't try to copy the URL out of the tooltip awkwardly).

markus101:
Looks pretty straight forward, just a couple little things to address and it should be ready. Thanks!

src/NzbDrone.Core/Notifications/Join/JoinProxy.cs
```diff
@@ -0,0 +1,93 @@
...
+                {
+                    throw new JoinException("Authentication failed.");
```
markus101: This is where we'd want to throw the JoinAuthenticationException.

Nedlinin: @markus101 : I've actually been looking into the API more and it looks like nothing I can do causes userAuthError.  Passing in faked deviceIds or apikey params still simply just returns a success=false.

Would we prefer to simply strip this out as it appears it'll never get hit?

Nedlinin:
@markus101 : Didn't hear back one way or the other so I went ahead and simply throw the JoinAuthException where you noted.  Not sure if this functionality will ever actually be added into the API but at least we will support it when/if it happens.

markus101:
Sorry @Nedlinin I saw it come in and failed to make a note to respond. I think your approach makes the most sense if it ever goes live. Thanks!

Nedlinin:
@markus101 : No worries.  Let me know if there are any other issues you notice with this PR.  Otherwise I'm calling it done in hopes someone else will maintain/expand the functionality down the road :)

markus101:
Thanks, made a minor change and merged it into develop.

PR #1402, opened 2016-07-27
Include afpfs mount points in free space checks

### LINKED ISSUE #1399: "System - Disk Space" inaccurate; OSX autofs mount is "Unknown" type
_Edit:_ `Sonarr` does use longest-path for series' mount computation; updated title to reflect root of the problem.

I'm on Mac OS X with a Synology NAS connected. The "System - Disk Space" only shows "/" despite my media folder being elsewhere. I've seen bugs opened and dismissed under the statement "it depends on what your OS reports". That's true, but not entirely accurate. Here's a sample program that shows my OS does in fact display the mounts (the code is at the bottom of the report):

```
$ mono free.exe
name: / free: 291654406144
name: /dev free: 0
name: /net free: 0
name: /home free: 0
name: /mnt/toaster free: 0
name: /mnt/toaster/videos free: 2215810793472
name: /mnt/toaster/music free: 2215810793472
```

In my particular case, the `autofs` system creates the mount points to `toaster` for `videos` and `music`. While `/mnt/toaster` reports as 0-bytes (because the root of the automounted filesystem isn't writable), the real mountpoints are writable.

What needs to be done in `NzbDrone.Core.DiskSpace.DiskSpaceService`[1] is to find the longest matching `DriveInfo` paths of the series' paths. For example, if I had the series:
```
Foo: /opt/series/Foo/
Bar: /mnt/toaster/videos/tv/Bar/
```
then the algorithm would determine the disk space locations as:
```
Foo -> /
Bar -> /mnt/toaster/videos/
```
because they are the longest matching paths of the series that match a mount.

```
# Psuedo-code
system_disk_space_list = []
for s in series:
    p = s.path
    while p not in mounts:
        p = dirname(p)
    system_disk_space_list.append(p)
system_disk_space_list = distinct(system_disk_space_list)
```

Sample code:
```c#
using System;
using System.IO;

namespace Examples
{
    public class FreeSpace
    {
        public static void Main(string[] args)
        {
            foreach (DriveInfo drive in DriveInfo.GetDrives())
            {
                Console.WriteLine ("name: {0} free: {1}",
                                   drive.Name,
                                   drive.AvailableFreeSpace);
            }
        }
    }
}
```

[1] Note I'm not familiar with `Sonarr`'s structure, so this may be the wrong place.

markus101:
#### Database Migration
NO

#### Description
Includes `afpfs` file systems in free space checks

Fixes #1399

### DIFF
--- a/src/NzbDrone.Common/Disk/DiskProviderBase.cs
+++ b/src/NzbDrone.Common/Disk/DiskProviderBase.cs
@@ -9,13 +9,14 @@ using NzbDrone.Common.EnsureThat;
 using NzbDrone.Common.EnvironmentInfo;
 using NzbDrone.Common.Extensions;
 using NzbDrone.Common.Instrumentation;
-using NzbDrone.Common.Instrumentation.Extensions;
 
 namespace NzbDrone.Common.Disk
 {
     public abstract class DiskProviderBase : IDiskProvider
     {
         private static readonly Logger Logger = NzbDroneLogger.GetLogger(typeof(DiskProviderBase));
+
+        private static readonly HashSet<string> UnknownDriveTypes = new HashSet<string>{ "afpfs" };
 
         public abstract long? GetAvailableSpace(string path);
         public abstract void InheritFolderPermissions(string filename);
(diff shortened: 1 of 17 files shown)


Taloth:
afpfs is linux only, so it probably belongs in NzbDrone.Mono.
Also, I would just add a Dictionary<string, DriveType>, map it to Network. And put in the zfs special case in there too.
Then handle the mapping in a method `GetDriveType(string filesystem, DriveType defaultType) { }` that can be applied before the 'Where(...)'.

Taloth:
Code looks good. Checked it on linux, zfs still shows up properly.

So squash it and ship it.

markus101:
Squashed and merged.

Issue #1410, opened 2016-08-01
RARBG Indexer No Longer Works


a92nj2efnfsa:
The built in rarbg indexer stopped working recently. It now provides an error saying you need to complete a captcha before getting the token.

Unable to connect to indexer: HTTP request failed: [403:Forbidden] [GET] at [https://torrentapi.org/pubapi_v2.php?get_token=get_token]

HTTP Error - Res: [GET] https://torrentapi.org/pubapi_v2.php?get_token=get_token: 403.Forbidden
```
<!DOCTYPE html>
<!--[if lt IE 7]> <html class="no-js ie6 oldie" lang="en-US"> <![endif]-->
<!--[if IE 7]>    <html class="no-js ie7 oldie" lang="en-US"> <![endif]-->
<!--[if IE 8]>    <html class="no-js ie8 oldie" lang="en-US"> <![endif]-->
<!--[if gt IE 8]><!--> <html class="no-js" lang="en-US"> <!--<![endif]-->
<head>
<title>Attention Required! | CloudFlare</title>
<meta charset="UTF-8" />
<meta http-equiv="Content-Type" content="text/html; charset=UTF-8" />
<meta http-equiv="X-UA-Compatible" content="IE=Edge,chrome=1" />
<meta name="robots" content="noindex, nofollow" />
<meta name="viewport" content="width=device-width,initial-scale=1,maximum-scale=1" />
<link rel="stylesheet" id="cf_styles-css" href="/cdn-cgi/styles/cf.errors.css" type="text/css" media="screen,projection" />
<!--[if lt IE 9]><link rel="stylesheet" id='cf_styles-ie-css' href="/cdn-cgi/styles/cf.errors.ie.css" type="text/css" media="screen,projection" /><![endif]-->
<style type="text/css">body{margin:0;padding:0}</style>
<!--[if lte IE 9]><script type="text/javascript" src="/cdn-cgi/scripts/jquery.min.js"></script><![endif]-->
<!--[if gte IE 10]><!--><script type="text/javascript" src="/cdn-cgi/scripts/zepto.min.js"></script><!--<![endif]-->
<script type="text/javascript" src="/cdn-cgi/scripts/cf.common.js"></script>


</head>
<body>
  <div id="cf-wrapper">
    <div class="cf-alert cf-alert-error cf-cookie-error" id="cookie-alert" data-translate="enable_cookies">Please enable cookies.</div>
    <div id="cf-error-details" class="cf-error-details-wrapper">
[...]

PR #1421, opened 2016-08-10
Cloudflare captcha

Taloth:
#### Description
The result of a weeks worth of evenings. (you're welcome ppl)
Basically we extract the cloudflare captcha challenge code from the html page, and present only a reCaptcha widget.
The CAPTCHA token does expire and will need to be manually renewed. (atm rarbg has 7 days, but that will be changed to 30 days later)

#### Todos
- [ ] Add warning about it expiring (optionally we could add a date to the UI?)
- [ ] Retest Twitter (likely broken)

#### Issues

Closes: #1410 
Ref: rarbg/torrentapi#11

#### Preview:

Add Rarbg & run Test:
image

Sonarr notices Captcha Response, click Refresh Token to do the CAPTCHA challenge:
image

UI loads google reCaptcha scripts and widget:
image

Sonarr gets CloudFlare clearance cookie using reCaptcha response as proof:
image

### DIFF
--- a/src/NzbDrone.Core/Http/CloudFlare/CloudFlareCaptchaException.cs
+++ b/src/NzbDrone.Core/Http/CloudFlare/CloudFlareCaptchaException.cs
@@ -1,0 +1,19 @@
+using NzbDrone.Common.Exceptions;
+using NzbDrone.Common.Http;
+
+namespace NzbDrone.Core.Http.CloudFlare
+{
+    public class CloudFlareCaptchaException : NzbDroneException
+    {
+        public HttpResponse Response { get; set; }
+
+        public CloudFlareCaptchaRequest CaptchaRequest { get; set; }
+
+        public CloudFlareCaptchaException(HttpResponse response, CloudFlareCaptchaRequest captchaRequest)
+            : base("Unable to access {0}, blocked by CloudFlare CAPTCHA. Likely due to shared-IP VPN.", response.Request.Url.Host)
+        {
+            Response = response;
+            CaptchaRequest = captchaRequest;
+        }
+    }
+}
--- a/src/UI/Form/CaptchaTemplate.hbs
+++ b/src/UI/Form/CaptchaTemplate.hbs
@@ -1,0 +1,14 @@
+<div class="form-group {{#if advanced}}advanced-setting{{/if}}">
+    <label class="col-sm-3 control-label">{{label}}</label>
+
+    <div class="col-sm-5">
+        <div class="input-group">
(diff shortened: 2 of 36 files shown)


src/NzbDrone.Core/Http/CloudFlare/CloudFlareCaptchaException.cs
```diff
@@ -0,0 +1,19 @@
...
+namespace NzbDrone.Core.Http.CloudFlare
+{
+    public class CloudFlareCaptchaException : NzbDroneException
+    {
+        public HttpResponse Response { get; set; }
+
+        public CloudFlareCaptchaRequest CaptchaRequest { get; set; }
+
+        public CloudFlareCaptchaException(HttpResponse response, CloudFlareCaptchaRequest captchaRequest)
+            : base("Unable to access {0}, blocked by CloudFlare CAPTCHA. Likely due to shared-IP VPN.", response.Request.Url.Host)
```
markus101: `Likely due to shared-IP VPN.` - could also be a VPS, do we need to specify likely part at all?

src/UI/Form/CaptchaTemplate.hbs
```diff
@@ -0,0 +1,14 @@
+<div class="form-group {{#if advanced}}advanced-setting{{/if}}">
+    <label class="col-sm-3 control-label">{{label}}</label>
+
+    <div class="col-sm-5">
+		<div class="input-group">
```
markus101: Spaces, not tabs :smile:

markus101:
Connecting to twitter doesn't work, a new tab is opened but it's URL is: `about:blank`, it looks like the API's response is empty.

src/NzbDrone.Core/Http/CloudFlare/CloudFlareCaptchaException.cs
Taloth: VPS mostly have their own static IP, so that's less of a problem.
We don't need to specify it explicitly, but the error is thrown and logged for all HTTP responses, not just Rarbg. So I thought it would be wise to point ppl to the likely suspect.

ldexterldesign:
Hey,

I'm currently unable to add https://rarbg.to/ as an indexer - am I the only one suffering this issue or will this commit provide a fix when it's released?

Hope to hear back

Kind thanks

PS I see the last Sonarr release was https://github.com/Sonarr/Sonarr/releases/tag/v2.0.0.4230 , so I'm guessing there will be another one soon..?

Taloth:
https://github.com/rarbg/torrentapi/issues/11#issuecomment-239641443

ldexterldesign:
Hey @Taloth,

Cheers for reply

It's worked today, so unsure what the issue was/is 🙄

Kind thanks

PR #1423, opened 2016-08-14
Add VideoBitDepth & AudioChannels to mediaInfo sample.

chrisallen:
#### Database Migration
NO
#### Description
Adds sample mediaInfo values for `VideoBitDepth` `AudioChannels` to be shown in the "Settings > Media Management" page.

### DIFF
--- a/src/NzbDrone.Core/Organizer/FileNameSampleService.cs
+++ b/src/NzbDrone.Core/Organizer/FileNameSampleService.cs
@@ -88,7 +88,10 @@ namespace NzbDrone.Core.Organizer
             var mediaInfo = new MediaInfoModel()
             {
                 VideoCodec = "AVC",
+                VideoBitDepth = 8,
                 AudioFormat = "DTS",
+                AudioChannels = 6,
+                AudioChannelPositions = "3/2/0.1",
                 AudioLanguages = "English",
                 Subtitles = "English/German"
             };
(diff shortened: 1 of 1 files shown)


src/NzbDrone.Core/Organizer/FileNameSampleService.cs
```diff
@@ -88,15 +88,19 @@ public FileNameSampleService(IBuildFileNames buildFileNames)
             var mediaInfo = new MediaInfoModel()
             {
                 VideoCodec = "AVC",
+                VideoBitDepth = "8",
                 AudioFormat = "DTS",
+                AudioChannels = "5.1",
```
markus101: Audio channels is a whole number, which would be `6` in this case and `AudioChannelPositions` would be set to `3/2/0.1`, which would get calculated to 5.1.

src/NzbDrone.Core/Organizer/FileNameSampleService.cs
```diff
@@ -88,15 +88,19 @@ public FileNameSampleService(IBuildFileNames buildFileNames)
             var mediaInfo = new MediaInfoModel()
             {
                 VideoCodec = "AVC",
+                VideoBitDepth = "8",
```
markus101: `VideoBitDepth = 8,`

markus101:
Thanks!

PR #1424, opened 2016-08-16
New: Support for TLS 1.1 and 1.2 connections when .net 4.5 is installed

markus101:
#### Database Migration
NO

#### Description
Support for TLS 1.1 and TLS 1.2 when making HTTP requests. SSL3 and TLS are also still supported, SSL3 should be dropped for security reasons, but left a note to do that in v3.

### DIFF
--- a/src/NzbDrone.Common/Security/SecurityProtocolPolicy.cs
+++ b/src/NzbDrone.Common/Security/SecurityProtocolPolicy.cs
@@ -1,0 +1,40 @@
+﻿using System;
+using System.Net;
+using NLog;
+using NzbDrone.Common.Instrumentation;
+
+namespace NzbDrone.Common.Security
+{
+    public static class SecurityProtocolPolicy
+    {
+        private static readonly Logger Logger = NzbDroneLogger.GetLogger(typeof(SecurityProtocolPolicy));
+
+        private const SecurityProtocolType Tls11 = (SecurityProtocolType)768;
+        private const SecurityProtocolType Tls12 = (SecurityProtocolType)3072;
+
+        public static void Register()
+        {
+            try
+            {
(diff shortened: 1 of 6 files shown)


src/NzbDrone.Host/Owin/OwinHostController.cs
```diff
@@ -34,8 +34,6 @@ public class OwinHostController : IHostController
 
         public void StartServer()
         {
-            X509CertificateValidationPolicy.Register();
```
markus101: We're already doing this in Bootstrap.cs so it seems superfluous here.

src/NzbDrone.Common/Security/SecurityProtocolPolicy.cs
```diff
@@ -0,0 +1,23 @@
...
+namespace NzbDrone.Common.Security
+{
+    public static class SecurityProtocolPolicy
+    {
+        private static readonly Logger Logger = NzbDroneLogger.GetLogger(typeof(SecurityProtocolPolicy));
+
+        public static void Register()
+        {
+            Logger.Debug("Checking if TLS 1.1 and TLS 1.2 are supported");
+            if (Enum.IsDefined(typeof(SecurityProtocolType), 768) && Enum.IsDefined(typeof(SecurityProtocolType), 3072))
```
Taloth: How about adding two `private const`, shall we?

src/NzbDrone.Host/Owin/OwinHostController.cs
Taloth: Agreed

Taloth:
Some constants would be nice. Now, how to test it? :wink:
Could also split TLS 1.1 and 1.2.

Try Catch might be good, afaik the property throws if you add invalid values, and it isn't critical.

The log entry isn't really useful, that only gets logged to console. (debug file isn't initialized yet)

PR #1425, opened 2016-08-17
Changed exit statement to "Press enter to exit..." to match use of ReadLine()

ashleybroughton:
#### Description
Minor fix to match input request in NzbDrone.Console/ConsoleApp.cs
Changed "Press any key to exit..." prompt to "Press enter to exit..."


Taloth:
tbh, I would rather have it on Enter. 'Any key' is a bit too easy, for example, pressing the windows key to open a notepad to copy/paste the error in would be enough.

markus101:
I was thinking the same as @Taloth

ashleybroughton:
I have updated the commit to reflect your preferences.

thezoggy:
Maybe Escape is a better choice?

PR #1428, opened 2016-08-17
Hide options on Settings Page

maxkoryukov:
This is just a **Demo PR**, not a final PR, but an offer to consider this feature.
#### Database Migration

NO
## Description

I want to hide some options from the **Settings page**. But _hide SOME options_ - is not the best approach, so I am trying to improve the UI and the server part, to make them able to:
- hide one/many settings from UI (**done** for "General" tab)
- make a setting read-only (**done** for "General" tab)
- disable modifying one/many settings on server side (**in process**)
- disable access to not-readable settings on server side (**in process**)
### Why?

I want to **preconfigure** the Sonarr and make the part of the config immutable. But additionaly, I want to retain the ability to modify several other settings.
#### Todos
- [ ] Tests - I don't know how to test the UI
- [ ] Documentation - in process


src/UI/Handlebars/Helpers/MetaOption.js
```diff
@@ -0,0 +1,73 @@
+var Handlebars = require('handlebars');
+
+var metaOptionCurrentConfig = {};
+
+metaOptionCurrentConfig = require('../../../../_output/UI/Content/metaOption.json');
```
maxkoryukov: This is an ugly part.. Still don't know, how to make this right..

see maxkoryukov/sonarr#1

markus101:
I don't think it's the UI's responsibility to to make things read-only like this, it would be trivial for someone to bypass the readonly property and make changes or use the API to make changes (albeit undocumented right now). I'm also a little unclear how this would work if this were to be merged in since the UI would be compiled with read-only options that would limit what user's could setup on their own systems, which would definitely not be wanted.

You mention wanting to preconfigure Sonarr and not allow changes, which these changes focus towards, but I'm not sure if it makes sense to have them in the main repo, can you elaborate on what you're trying to achieve?

maxkoryukov:
@markus101 , the desired procedure is following:
1. configure Sonar on a local PC (all settings are accessible) and **save the configuration**
2. restrict access to the settings (make a part of them _readonly_ or _hidden_) using features of this PR. Finally we will get the **meta-configuration**.
3. install Sonarr on client's PC (assuming the access to FS is restricted) with the saved config and meta-config

Done! Now user/client could tune his Sonarr installation, but can't break down the important settings (like an integration with installed torrent-downloaders).

maxkoryukov:
_Regarding the weakness of UI restrictions:_ currently, it is just a demo PR, I will implement server-side restrictions soon.

markus101:
If the server is controlling it you shouldn't need to have config file store on the client, which would get nuked every time Sonarr updated (the entire installation directory is wiped out) and you can't hard code that config file path to live along the rest of the config because it's not always going to be in the same place.

Overall I don't see this as something we want to support and maintain, it seems to fit what you're doing, but I don't see it fitting into most setups.

xvkdev:
These kind of changes help a lot to seedbox providers that want to provide a preconfigured service in which some things just cannot be changed (because they would break functionality, for example port numbers). Not saying this is the right implementation but the functionality itself is a god send.

thezoggy:
or why not just define those breakable changes via an init script, so its always set on startup / can put a little blurp in it to say dont change above this line

maxkoryukov:
@thezoggy, I am sure, when it is possible to change config - somebody _will_ change the config, and will break his own installation.
And he will create a ticket. For us..

xvkdev:
@thezoggy The thing is that those changes can do more than just break temporarily sonarr. For example writing in the wrong directory could fill the wrong disk (best case). Yes, you could argue that this is a permissions problem - but well, some things do need to run with the same system user, which makes sense if they are servicing the same human.

There's also the fact that many humans (me included) love to experiment for the sake of it. So it's best to limit what one can experiment with to what can't harm to one self (or one's data).

Taloth:
@maxkoryukov @xvkdev That's something that should be at the beginning of the PR. "Why?" is a really important question. It's the first thing we wonder about when considering a feature.   

"What?": You want to lock down certain settings to prevent unintentional misconfiguration. 'Hiding' is just a means to an end, not a goal in on itself. It's nice, but making stuff readonly could be good enough for the purpose.   
You'd be surprised to hear how often users propose 'solutions' instead of describing the problem they're having.

No, not going to do all the Five-Ws :)

But while we at the Five-W: Who are you? Any affiliation with seedbox providers or related services?    

But lets get into the tech stuff, some observations:

The template changes and handlebars isn't a very good approach, it's hard to maintain and fairly verbose.   
Consider using a mixin from from code behind. The View/Layout knows it's a Settings page. Make a mixin that uses jquery to get all the input elements on the view. Those things already have the setting name in them, so no need for additional handlebar decorations. The mixin can then find the form-group parent, or whatever is needed to find the appropriate div to encapsulate/hide. It can also set the input to 'readonly'. All it needs is the metadata.
That's just for the normal settings, not for stuff like profiles and tags. (might be useful to make a list of settings that you want to lock down)
Of course it still needs backend logic to provide the metadata and block modifications on the api.

The reason why 'hidden' could be tricky is that if all settings in a fieldset are hidden, then you'll see the fieldset header with no content. It'll make someone frantically press the ugly-button, we can't have that of course. Not a huge obstacle, but will need a bit of work to get it neat, so it might be easier to just set those fields readonly.

I'm on holiday so I can't get into much more details. The sonarr devs will still have to decide whether we would accept a feature like this, so it's important to get straight what you want to achieve and why.

xvkdev:
@Taloth Yes, we run a seedbox service. 

Hiding is also a goal, not a means. It simplifies things enormously for the end user when they can only see what they can modify without risk of harming the system or the data.  In fact most interfaces as they become complex in functionality have this - they just hide all the values the users shouldn't be playing with into an "Advanced" button (which itself could be hidden).

Show users a value they can't change (for example, a port number, or a cache size, or things like that) and you can be sure they're going to ask to change those values just because some tutorial, or forum post, etc, tells them to. 

Now, there are good reasons to show read-only fields, too, such as those that are informative (such as a path) or that needs to be copied somewhere, which is why read-only, and hidden are two options instead of read-only+hidden. 

Again I'm not claiming that the offered implementation is a good one (but I'd be surprised @maxkoryukov wasn't doing a good job), just that there's a real world need for this functionality.

thezoggy:
there is already an advanced button, you prob could just load a extra css that overrides the visibility of that button via css.. `display: none` for it.. and that would **solve** your concern..  ?  or maybe offer a kiosk mode style argument at startup that just does this internally

xvkdev:
@thezoggy I mentioned the advanced button for explanation purposes :-) It's not the same thing here exactly. For example, the download directory is obviously not an advanced setting, yet is one that we need to made static (to avoid files filling the wrong disk, or being written somewhere the FTP user will not have access to, etc).

There are many other similar settings, particularly those that are about interaction with other programs or services (so details must much in both places).

Taloth:
@xvkdev But those aren't just 'settings'... Root Folders, Connect, Download Clients. You're talking about the whole frigging UI. That's a massive undertaking.   
At some point it's gonna be easier to just whip up a simplified UI and talk directly to the API.

xvkdev:
@Taloth Of course, but in general adding two options to *all* fields in the UI is simpler (since you just need to add them to the common class) is cleaner than adding them to just some fields.

If adding the option requires adding code for each field then the problem is different.

markus101:
Hiding settings in this way isn't something viable (especially with the planned UI changes), but I've opened up an issue to discuss multiple users and how that could work: https://github.com/Sonarr/Sonarr/issues/1682

PR #1448, opened 2016-09-02
Support both integer and string ids for HDBits queries

### LINKED ISSUE #1443: Hdbits indexer fails after api change
The hdbits api has changed and the result is the the hdbits indexer is not working.

 i see in the log that sonarr tries to convert the id to an int64. The id is a hash now, right, which means that you will get a  conversion error:

Newtonsoft.Json.JsonSerializationException: Error converting value "2aa43e456d6c5ddb1ad03e7ac70a416ea4a2be84e478045ae390814327c5d2fee581ad21b8f056a49cfbd1fddd5bb8c6fb1353caaa17dd3666b55e1383c27fc5" to type 'System.Int64'. Path '[0].id'. ---> System.FormatException: Input string was not in a correct format. at System.Number.StringToNumber(String str, NumberStyles options, NumberBuffer& number, NumberFormatInfo info, Boolean parseDecimal)
at System.Number.ParseInt64(String value, NumberStyles options, NumberFormatInfo numfmt)
at System.String.System.IConvertible.ToInt64(IFormatProvider provider)
at System.Convert.ChangeType(Object value, Type conversionType, IFormatProvider provider)
at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.EnsureType(JsonReader reader, Object value, CultureInfo culture, JsonContract contract, Type targetType)
--- End of inner exception stack trace ---

 A solution could be to probably just remove the id from the request and set id to string in the response

Dahlgren:
#### Database Migration
NO

#### Description
Adds support for both integer and string IDs in HDBits queries.
A recent change in their API changed all IDs to strings and removed query support for ID entirely.

#### Todos
- [X] Tests
- [X] Documentation

* Fixes #1443


src/NzbDrone.Core.Test/Files/Indexers/HdBits/RecentFeed.json
```diff
@@ -28,7 +28,7 @@
             }
         },
         {
-            "id": 257140,
+            "id": "257140",
```
markus101: We should test this in `HDBitsFixture` to make sure it works (it should, but let's make sure).

If we duplicate and rename this file to `RecentFeedStringIDs.json` and `RecentFeedLongIDs.json` and set the IDs appropriately we can both:

```
[Test]
        public void should_parse_recent_feed_from_HDBits()
        {
            var responseJson = ReadAllText(@"Files/Indexers/HdBits/RecentFeed.json");
```

becomes:

```
[TestCase("@"Files/Indexers/HdBits/RecentFeedLongIDs.json")]
[TestCase("@"Files/Indexers/HdBits/RecentFeedStringIDs.json")]
        public void should_parse_recent_feed_from_HDBits(string fileName)
        {
            var responseJson = ReadAllText(fileName);
```

Dahlgren:
Added separate test case for strings and reverted changes to integer test case

PR #1449, opened 2016-09-04
Allow multiple twitter notification recipients.

navitiello:
#### Database Migration

YES | NO
#### Description

#### Todos
- [ ] Tests
- [ ] Documentation
- 

This change allows for multiple twitter notification recipients. Each twitter user account is separated by comma. Now my girlfriend and I can both be notified via DM when a new episodes are ready to watch.

### DIFF
--- a/src/NzbDrone.Core/Notifications/Twitter/TwitterService.cs
+++ b/src/NzbDrone.Core/Notifications/Twitter/TwitterService.cs
@@ -83,9 +83,19 @@ namespace NzbDrone.Core.Notifications.Twitter
 
                 if (settings.DirectMessage)
                 {
-                    twitter.DirectMessage(message, settings.Mention);
-                }
+                    if(settings.Mention.Contains(","))
+					{
+						var recipients = settings.Mention.Replace(" ",string.Empty).Split(',');
 
+						foreach(var recipient in recipients)
+						{
+							twitter.DirectMessage(message,recipient);
+						}
+					}
+					else
+					{
+						twitter.DirectMessage(message, settings.Mention);	
+					}
                 else
                 {
                     if (settings.Mention.IsNotNullOrWhiteSpace())


src/NzbDrone.Core/Notifications/Twitter/TwitterService.cs
```diff
@@ -83,9 +83,19 @@ public void SendNotification(string message, TwitterSettings settings)
 
                 if (settings.DirectMessage)
                 {
-                    twitter.DirectMessage(message, settings.Mention);
-                }
-
+                    if(settings.Mention.Contains(","))
+					{
+						var recipients = settings.Mention.Replace(" ",string.Empty).Split(',');
```
markus101: `" ",string.Empty` should be `" ", string.Empty`

src/NzbDrone.Core/Notifications/Twitter/TwitterService.cs
```diff
@@ -83,9 +83,19 @@ public void SendNotification(string message, TwitterSettings settings)
...
+
+						foreach(var recipient in recipients)
+						{
+							twitter.DirectMessage(message,recipient);
```
markus101: Same here, needs a space after the comma.

src/NzbDrone.Core/Notifications/Twitter/TwitterService.cs
```diff
@@ -83,9 +83,19 @@ public void SendNotification(string message, TwitterSettings settings)
 
                 if (settings.DirectMessage)
                 {
-                    twitter.DirectMessage(message, settings.Mention);
-                }
-
+                    if(settings.Mention.Contains(","))
+					{
```
markus101: The brace is not aligned properly.

src/NzbDrone.Core/Notifications/Twitter/TwitterService.cs
markus101: Using `StringSplitOptions.RemoveEmptyEntries` would avoid trying to send an invalid mention.

Better yet having spaces in the mention field should be a validation error so it doesn't save it, if we want a safety net we should use `Trim()` after the split, not just blindly remove spaces.

PR #1451, opened 2016-09-08
Updater fixes and logging

Taloth:
When emptying the install dir fails, due to permission issues, the rollback logic kicks in. But instead of just restoring the missing files, it rewrites everything and pukes over the same permission issue.
Usually this doesn't corrupt the install (coz it follows the same sequence), but it logs a Fatal twice.

So I changed it to a MirrorFolder function that compares source and target folders and files and applies only the changes required.

Also added the current version to the logging, so we know what the original version was and which version is being installed.

I didn't test it yet.

### DIFF
--- a/src/NzbDrone.Common/Disk/DiskTransferService.cs
+++ b/src/NzbDrone.Common/Disk/DiskTransferService.cs
@@ -15,6 +15,7 @@ namespace NzbDrone.Common.Disk
     {
         TransferMode TransferFolder(string sourcePath, string targetPath, TransferMode mode, bool verified = true);
         TransferMode TransferFile(string sourcePath, string targetPath, TransferMode mode, bool overwrite = false, bool verified = true);
+        int MirrorFolder(string sourcePath, string targetPath);
     }
 
     public enum DiskTransferVerificationMode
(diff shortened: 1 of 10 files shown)


src/NzbDrone.Common/Disk/DiskTransferService.cs
```diff
@@ -83,6 +84,100 @@ public TransferMode TransferFolder(string sourcePath, string targetPath, Transfe
...
+
+            foreach (var subDir in sourceFolders)
+            {
+                filesCopied += MirrorFolder(subDir.FullName, Path.Combine(targetPath, subDir.Name));
+            }
+
+            var sourceFiles = _diskProvider.GetFileInfos(sourcePath);
+            var targetFiles = _diskProvider.GetFileInfos(targetPath);
+
+            foreach (var targetFile in targetFiles.Where(v => !sourceFiles.Any(d => d.Name == v.Name)))
```
markus101: Should we be testing the relative path of the file in case we moved a file within the directory?

src/NzbDrone.Common/Disk/DiskTransferService.cs
```diff
@@ -83,6 +84,100 @@ public TransferMode TransferFolder(string sourcePath, string targetPath, Transfe
...
+                TransferFile(sourceFile.FullName, targetFile, TransferMode.Copy, true, true);
+                filesCopied++;
+            }
+
+            return filesCopied;
+        }
+
+        private bool CompareFiles(string sourceFile, string targetFile)
+        {
+            if (_diskProvider.FileExists(sourceFile) || _diskProvider.FileExists(targetFile))
```
markus101: File doesn't exist? Won't the source file always exist?

Taloth: Yes, source exists, but I thought it would be smart to validate it anyway. Now both args are validated, and you can call the function without assumptions.

src/NzbDrone.Common/Disk/DiskTransferService.cs
Taloth: A rename is still a change, and to 'undo' it we would still need to write and needs write permissions.
This method isn't written to prevent IO (we even load all file content to compare), it's to prevent trying to write files that haven't changed.

markus101: If the filename still exists in the source, regardless of the path it won't be deleted though.

If the target has a file of `\test.txt` and the source has `\test\test.txt` it will won't be deleted and the target directory will have both files.

Taloth: ah, no, it's only processing the files and folders at the current level, and [recurse](https://github.com/Sonarr/Sonarr/pull/1451/files/4dedc859c2a5d08bad1bfb9b70aaaab2c2a6fed8#diff-7b8398706e513f14fba986e0888b8802R111). So it won't end up comparing those two files.

src/NzbDrone.Common/Disk/DiskTransferService.cs
markus101: I wasn't concerned about the check, but the logic, if either file exists we return false which means we never actually check to see if they are the same before copying the file.

You need two checks here:

```
if (!_diskProvider.FileExists(sourceFile))
{
    return true;
}
```

So it doesn't try to transfer the file.

```
if (!_diskProvider.FileExists(targetFile))
{
    return false;
}
```

So the source file is copied because the target file doesn't exist.

Taloth: The Compare function only compares the files, it doesn't do anything with that conclusion. If either doesn't exist... then they aren't equal. It's entirely separate from what we do later with it (copying the file).
If it eases your mind I can remove the FileExists(sourceFile), it really doesn't matter.

markus101: The source check is fine in a pure-comparison (I'll ignore the caller), but they need to check if they don't exist don't they? Checking if they exist and returning false if they do short circuits the entire method if either exists.

src/NzbDrone.Common/Disk/DiskTransferService.cs
markus101: Ahhh! missed that, then :+1:

PR #1459, opened 2016-09-13
Added the ability to send Join notifications to multiple devices.  Ad…

### LINKED ISSUE #1455: Add deviceID support for Join Notification
Atm we just send to `group.all`. Adding a field for deviceids should be fairly trivial.

Nedlinin:
#### Database Migration
NO

#### Description
Allow users to specify certain devices that receive notifications using the Join API.
Added some better exception handling/setting validation and corresponding error messages.

Fixes #1455

### DIFF
--- a/src/NzbDrone.Core/Notifications/Join/JoinProxy.cs
+++ b/src/NzbDrone.Core/Notifications/Join/JoinProxy.cs
@@ -36,7 +36,7 @@ namespace NzbDrone.Core.Notifications.Join
             catch (JoinException ex)
             {
                 _logger.Error(ex, "Unable to send Join message.");
-                throw new JoinException("Unable to send Join notifications. " + ex.Message);
+                throw ex;
             }
         }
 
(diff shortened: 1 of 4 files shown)


src/NzbDrone.Core/Notifications/Join/JoinProxy.cs
```diff
@@ -50,46 +50,77 @@ public ValidationFailure Test(JoinSettings settings)
...
+            if (res.userAuthError)
+            {
+                throw new JoinAuthException("Authentication failed.");
+            }
 
-                if (res.errorMessage != null)
+            if (res.errorMessage != null)
+            {
+                // I dislike hard coding this string here but nothing else in the Join API lets you determine the error.
+                // No enum or flag contained in the response for this so we have to do a string check..
```
markus101:
```
// Unfortunately hard coding this string here is the only way to determine that there aren't any devices to send to.
// There isn't an enum or flag contained in the response that can be used instead.
```

src/NzbDrone.Core/Notifications/Join/JoinProxy.cs
```diff
@@ -50,46 +50,77 @@ public ValidationFailure Test(JoinSettings settings)
...
                 {
-                    throw new JoinException(res.errorMessage);
+                    throw new JoinInvalidDeviceException(res.errorMessage);
                 }
-
-                if (res.userAuthError)
+                // Oddly enough, rather than give us an "Invalid API key", the Join API seems to assume the key is valid
+                // But fail on doing a device lookup associated with that key.  
+                // In our case we are using "deviceIds" rather than "deviceId" so when the singular form error shows up
+                // we know our API key was the fault.
```
markus101:
```
// Oddly enough, rather than give us an "Invalid API key", the Join API seems to assume the key is valid,
// but fails when doing a device lookup associated with that key.  
// In our case we are using "deviceIds" rather than "deviceId" so when the singular form error shows up
// we know the API key was the fault.
```

src/NzbDrone.Core/Notifications/Join/JoinSettings.cs
```diff
@@ -19,7 +20,10 @@ public class JoinSettings : IProviderConfig
         private static readonly JoinSettingsValidator Validator = new JoinSettingsValidator();
 
         [FieldDefinition(0, Label = "API Key", HelpText = "The API Key from your Join account settings (click Join API button).", HelpLink = "https://joinjoaomgcd.appspot.com/")]
-        public string APIKey { get; set; }
+        public string ApiKey { get; set; }
+
+        [FieldDefinition(1, Label = "Device IDs", HelpText = "Comma separated list of Device IDs you'd like to send notifications.  If unset, all devices will receive notifications.", HelpLink = "https://joinjoaomgcd.appspot.com/")]
```
markus101: `Comma separated list of Device IDs you'd like to send notifications to. If unset, all devices will receive notifications.`

Nedlinin:
@markus101 Requested changes made.  I'm a bit informal about my comments :)

PR #1464, opened 2016-09-15
add downloadvolumefactor and uploadvolumefactor torznab attributes

kaso17:
#### Database Migration
NO

#### Description
Using the attributes it's possible to signal states like freeleech/neutral leech/double upload

#### Todos
- [ ] Documentation (I'll update the wiki once it's merged)

none


Taloth:
I'm not opposed to adding those attributes. But what's the usecase from Sonarr's/CP's/etc perspective? How would it be used?
The minimumseedtime/minimumratio attributes was added so it could properly configure the download client and prevent HnRs (although that was never implemented).

kaso17:
sonarr, etc. could prefere a freeleech/double upload torrent over a normal one

PR #1468, opened 2016-09-22
Filter iCal feed by tags

Taloth:
Filter ical feed by tags. The UI uses the tag input, without the ability to add new tags.
It adds it as integers to the query params.
The feed endpoint supports both integers and tag labels.

Also added housekeeping logic to cleanup unused tags.

### DIFF
--- a/src/NzbDrone.Api/Calendar/CalendarFeedModule.cs
+++ b/src/NzbDrone.Api/Calendar/CalendarFeedModule.cs
@@ -5,17 +5,21 @@ using System.Linq;
 using DDay.iCal;
 using NzbDrone.Core.Tv;
 using Nancy.Responses;
+using NzbDrone.Core.Tags;
+using NzbDrone.Common.Extensions;
 
 namespace NzbDrone.Api.Calendar
 {
     public class CalendarFeedModule : NzbDroneFeedModule
     {
         private readonly IEpisodeService _episodeService;
+        private readonly ITagService _tagService;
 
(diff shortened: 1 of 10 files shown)


src/NzbDrone.Api/Calendar/CalendarFeedModule.cs
```diff
@@ -62,6 +67,12 @@ private Response GetCalendarFeed()
                 premiersOnly = bool.Parse(queryPremiersOnly.Value);
             }
 
+            if (queryTags.HasValue)
+            {
+                var tagInput = (string)queryTags.Value.ToString();
+                tags = tagInput.Split(',').Select(_tagService.GetTag).OfType<Tag>().Select(t => t.Id).ToArray();
```
Taloth: The null-check isn't needed anymore either. GetTag throws if a tag doesn't exist.

markus101: The null check was in reference to line 86, but was part of the thought here.

```
var tags = new List<Tag>();
...
tags.AddRange(tagInput.Split(',').Select(_tagService.GetTag).OfType<Tag>().Select(t => t.Id).ToList());
...
if (tags.None(episode.Series.Tags.Contains))
```

Taloth: And... you just filtered out every event if tags is empty. :smile:

But no worries, I already tweaked the code to be more self-describing.

Issue #1316, opened 2016-05-25
QBittorrent doesn't delete torrent when finished seeding


berserkwarwolf:
It would be really helpful if you could add this option forthe popular open source qbittorrent download client. I've read the API docs and it seems to support this:

https://github.com/qbittorrent/qBittorrent/wiki/WebUI-API-Documentation#delete-torrent

markus101:
This is not a bug, Sonarr has no way of knowing whether the torrent has finished seeding or it was stopped for another reason, Sonarr is playing it safe and not deleting it and won't until we have a way to determine that it finished seeding.

cbodley:
hi @markus101, sorry for commenting on a closed issue, but the inability to clean up completed downloads really limits the qbittorrent client's usefulness for me. is there a better place to hold a discussion about this change?

i would really like to find a way to make progress on this, and would be willing to coordinate with the qbittorrent project to find a proper solution

it would also be nice if, in the meantime, we could add an option to the qbittorrent settings to disable this change, and allow the possibly-not-fully-seeded torrents to be removed anyway. would you consider merging a pull request that adds (and documents) such an option?

markus101:
qbittorrent isn't the only client that has this limitation, rTorrent is has it as well, and Transmission/Vuze probably should but Transmission was released before we considered the consequences so it is grandfathered in (and Vuze only has minor differences from Transmission).

> i would really like to find a way to make progress on this, and would be willing to coordinate with the qbittorrent project to find a proper solution

A proper solution is one of two things:

1. qbittorrent adds a flag so Sonarr can see it's finished seeding
2. Sonarr is aware of the seeding goals (per tracker) that need to be met (time/ratio) and can get the required information from qbittorrent (and all other clients) to determine if it can be stopped and removed.

> in the meantime, we could add an option to the qbittorrent settings to disable this change, and allow the possibly-not-fully-seeded torrents to be removed anyway. would you consider merging a pull request that adds (and documents) such an option?

I'm not in favour of adding another option when there are better solutions even as a quick workaround to add a feature doesn't give any incentive to fix it the right way and hacks have a tendency to become the permanent solution. I'm not opposed to more discussion on this though. @Taloth what are your thoughts on this?

Taloth:
Ref: https://www.reddit.com/r/sonarr/comments/53ihv2/sonarr_not_removing_completed_downloads_qbittorent/

[...]

PR #1484, opened 2016-10-01
qbittorrent: allow removal of torrents that reached the configured max ratio

cbodley:
#### Database Migration
NO

#### Description
Track QBittorrent's configured `max_ratio`, and compare that against each torrent's `share_ratio` to determine whether the torrent has finished seeding. If it has, allow Sonarr to remove and delete the completed torrent.

#### Todos
- [ ] Tests
- [ ] Documentation

* https://github.com/Sonarr/Sonarr/issues/1316

### DIFF
--- a/src/NzbDrone.Core/Download/Clients/qBittorrent/QBittorrent.cs
+++ b/src/NzbDrone.Core/Download/Clients/qBittorrent/QBittorrent.cs
@@ -178,7 +178,7 @@ namespace NzbDrone.Core.Download.Clients.QBittorrent
         {
             var config = _proxy.GetConfig(Settings);
 
-            var destDir = new OsPath((string)config.GetValueOrDefault("save_path"));
+            var destDir = new OsPath(config.SavePath);
 
             return new DownloadClientStatus
             {
--- a/src/NzbDrone.Core/Download/Clients/qBittorrent/QBittorrent.cs
+++ b/src/NzbDrone.Core/Download/Clients/qBittorrent/QBittorrent.cs
@@ -227,6 +227,16 @@ namespace NzbDrone.Core.Download.Clients.QBittorrent
                         DetailedDescription = "Sonarr will not attempt to import completed downloads without a category."
                     };
                 }
+
+                // Complain if qBittorrent is configured to remove torrents on max ratio
+                var config = _proxy.GetConfig(Settings);
+                if (config.MaxRatioEnabled && config.RemoveOnMaxRatio)
+                {
+                    return new NzbDroneValidationFailure(String.Empty, "QBittorrent is configured to remove torrents when they reach their Share Ratio Limit")
+                    {
+                        DetailedDescription = "Sonarr will be unable to perform Completed Download Handling as configured. You can fix this in qBittorrent ('Tools -> Options...' in the menu) by changing 'Options -> BitTorrent -> Share Ratio Limiting' from 'Remove them' to 'Pause them'."
+                    };
+                }
             }
             catch (DownloadClientAuthenticationException ex)
             {
--- a/src/NzbDrone.Core/Download/Clients/qBittorrent/QBittorrent.cs
+++ b/src/NzbDrone.Core/Download/Clients/qBittorrent/QBittorrent.cs
@@ -80,20 +80,14 @@ namespace NzbDrone.Core.Download.Clients.QBittorrent
             }
         }
 
-        public override ProviderMessage Message
-        {
-            get
-            {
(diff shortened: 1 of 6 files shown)


src/NzbDrone.Core/Download/Clients/qBittorrent/QBittorrentTorrent.cs
```diff
@@ -21,5 +21,8 @@ public class QBittorrentTorrent
 
         [JsonProperty(PropertyName = "save_path")]
         public string SavePath { get; set; } // Torrent save path
+
+        [JsonProperty(PropertyName = "share_ratio")]
```
markus101: in 3.3.7 it looks like this is just `ratio`

src/NzbDrone.Core/Download/Clients/qBittorrent/QBittorrent.cs
```diff
@@ -227,6 +219,17 @@ private ValidationFailure TestConnection()
...
                     };
                 }
+
+                // complain if qBittorrent is configured to remove torrents on max ratio
+                var config = _proxy.GetConfig(Settings);
+                if (config.MaxRatioEnabled && config.RemoveOnMaxRatio)
+                {
+                    return new NzbDroneValidationFailure(String.Empty, "QBittorrent is configured to remove torrents when they reach their Share Ratio Limit")
+                    {
+                        IsWarning = true,
```
cbodley: good point. if `EnableCompletedDownloadHandling` is set, that wouldn't work at all. but if it isn't set, does it make sense to just warn in that case? or are there other things that would go wrong?

src/NzbDrone.Core/Download/Clients/qBittorrent/QBittorrentTorrent.cs
cbodley: you're right, i was looking at the wrong section of the api

src/NzbDrone.Core/Download/Clients/qBittorrent/QBittorrent.cs
```diff
@@ -116,11 +110,9 @@ public override IEnumerable<DownloadClientItem> GetItems()
...
 
                 item.OutputPath = _remotePathMappingService.RemapRemoteToLocal(Settings.Host, new OsPath(torrent.SavePath));
-                
-                // At the moment there isn't an easy way to detect if the torrent has
-                // reached the seeding limit, We would need to check the preferences 
-                // and still not be completely sure if that torrent has a limit set for it
-                item.IsReadOnly = true;
+
+                // Avoid removing torrents that haven't reached their max ratio
+                item.IsReadOnly = config.MaxRatioEnabled && config.MaxRatio > torrent.ShareRatio;
```
cbodley: > This should also check the state of the torrent I think, just to make sure it was actually paused as the user could opt to bypass the ratio check on selected torrents.

sorry, are you saying that a user may want to unpause a completed torrent to prevent Sonarr from importing? i.e.:
```
item.IsReadOnly = torrent.State != "pausedUP" || (config.MaxRatioEnabled && config.MaxRatio > torrent.Ratio);
```
or that a user may want to pause an incomplete torrent so that Sonarr could import it before it reached the ratio limit?
```
item.IsReadOnly = torrent.State != "pausedUP" && (config.MaxRatioEnabled && config.MaxRatio > torrent.Ratio);
```
> I take it the API doesn't yet expose per-torrent ratio limits?

correct, that's not exposed by either the torrent list or the torrent generic properties

cbodley:
I've pushed two fixes, but github is hiding some of my followup questions under the outdated diffs

src/NzbDrone.Core/Download/Clients/qBittorrent/QBittorrent.cs
markus101: Sorry, comparing to SABnzbd we throw an error if they enable TV sorting, which causes a number of issues with Sonarr, so a think an error makes sense.

src/NzbDrone.Core/Download/Clients/qBittorrent/QBittorrent.cs
markus101: > sorry, are you saying that a user may want to unpause a completed torrent to prevent Sonarr from importing?

not prevent it from import, but prevent Sonarr from removing it. Another use-case (that seems likely to happen) is if a user increases the ratio required to pause the torrent, since Sonarr only knows about the global ratio it would treat it as finished even if it's still seeding because that torrent hasn't hit it's ratio requirement.

markus101:
Saw that, responded to those inline (after showing them).

cbodley:
@markus101 thanks. i've addressed the comments and squashed the fixes

markus101:
Merged, thanks!

PR #1513, opened 2016-10-19
Added support to override Copy vs Move import logic for Manual Import

Taloth:
#### Database Migration

NO
#### Description
- DownloadedEpisodesScan command now has an importMode=['Move','Copy'] property to override default behavior. It defaults to whatever CDH is saying about the downloadId. If no downloadId is supplied, 'Move' is used as default.
- Manual Import UI now has a dropdown to change the mode. Defaults to 'Move'.
- Activity->Queue->Force Import sets the initial value of the dropdown to whatever CDH is saying about the download (so effectively, seeding torrent defaults to 'Copy')
- https://github.com/Sonarr/Sonarr/issues/606


markus101:
Code looks good and copying seems to be working well. :shipit:

gabo77:
With Copy vs Move, will this give the user the option to select Hardlink also?

markus101:
It will try to hardlink (and fallback to copy if it fails) if `Use Hardlinks instead of Copy` is enabled.

PR #1544, opened 2016-11-12
Import and Rename Extra files

### LINKED ISSUE #459: Move Additional Files

markus101:
#### Database Migration
NO

#### Description
Importing of extra files (new option in the UI) and renaming them when the episode file is renamed.

Closes #459

### DIFF
--- a/src/NzbDrone.Core.Test/ParserTests/LanguageParserFixture.cs
+++ b/src/NzbDrone.Core.Test/ParserTests/LanguageParserFixture.cs
@@ -54,6 +54,8 @@ namespace NzbDrone.Core.Test.ParserTests
 
         [TestCase("2 Broke Girls - S01E01 - Pilot.en.sub", Language.English)]
         [TestCase("2 Broke Girls - S01E01 - Pilot.eng.sub", Language.English)]
+        [TestCase("2 Broke Girls - S01E01 - Pilot.English.sub", Language.English)]
+        [TestCase("2 Broke Girls - S01E01 - Pilot.english.sub", Language.English)]
         [TestCase("2 Broke Girls - S01E01 - Pilot.sub", Language.Unknown)]
         public void should_parse_subtitle_language(string fileName, Language language)
         {
--- a/src/NzbDrone.Core/Extras/ExtraService.cs
+++ b/src/NzbDrone.Core/Extras/ExtraService.cs
@@ -57,10 +57,6 @@ namespace NzbDrone.Core.Extras
                 extraFileManager.CreateAfterEpisodeImport(series, episodeFile);
             }
 
-            // TODO: Remove
-            // Not importing files yet, testing that parsing is working properly first
-            return;
-
             var sourcePath = localEpisode.Path;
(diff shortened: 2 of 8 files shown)


dwatley:
What's the status here? Seems like #459 is on the cusp of being completed and it's much needed.

KalleDK:
Looking forward for this feature 💛 Is about to move my whole library, but I really dont want to move the srt's manually

valantislevas:
any news about ?

markus101:
Issues:

- Do a manual import with a .mkv .en.srt .nl.srt file. Imports the .mkv and a single .srt without suffix. (Both files should be imported with the proper language)
- got to the series folder, rename the .mkv file by adding something like a character in the episode title. hit refresh series. srt file gets deleted/recycled. (File shouldn't be deleted, ideally this would get linked to the file still and renamed? though this is a bit outside the scope of Sonarr, IMO)
- Rename Preview doesn't show extra files (I don't think this is critical, but would be great to have).

markus101:
Rebased and fixed the importing not using the proper language as part of the extension.

Taloth:
Won't that change the extension column in the db?

markus101:
Looks like past me thought of that and we get it from the path again before returning the `TExtraFile`.

Taloth:
Not entirely: https://github.com/Sonarr/Sonarr/pull/1544/files#diff-1778dc79c400fb3c577b08af93f8c40fR61

Don't redefine 'extension', for all ExtraFiles the extension is the same thing, the file extension. Don't concat the language there.

What you need to do is add a function that Builds the new filename in it's entirety, including extension. Not a 'GetExtensionWithSuffix'.
That might also help deal with the code duplication between the MoveFilesAfterRename methods.

markus101:
I'm referring to: https://github.com/Sonarr/Sonarr/blob/develop/src/NzbDrone.Core/Extras/Files/ExtraFileManager.cs#L64 which is getting the extension from the file.

We can get the new file name separately and pass that in instead of building it there and DRY things up a bit though...

markus101:
DRYed out `MoveFilesAfterRename`, but mostly left MetadataService's as-is and can revisit cleaning it up another time as it will be more invasive.

nzbjamz:
Awesome work @markus101! 👏

zjpleau:
How do you enable this feature?

markus101:
Media Management: Import Extra Files and set the extensions to import. It's an advanced setting. It's not released in a master release yet, just develop.

dbancajas:
@markus101  
Do you know when it's scheduled to be merged with the master? Also, how do I use develop branch on-the-fly?

spacecabbie:
hey i am running 0.2.0.654 - Apr 18 2017
I do not see this option ?

Taloth:
@spacecabbie That's a [Radarr](https://github.com/Radarr/Radarr) version number, this is Sonarr.

spacecabbie:
@Taloth Yes i already figured out the diferance :) i have posted in the related treath

PR #1595, opened 2016-12-20
Fixed: Reduce parameters required to add a new series

### LINKED ISSUE #1403: Add Series: Default properties that are NOT NULL in the database
For adding a series we have a limited number of properties that are required, but we have a larger number of fields that are required to successfully save it to the data base (`NOT NULL` DB columns), we should default anything we expect to not be null, but don't care if its been set or remove the `NOT NULL` restriction in the DB (and make sure they are being deserialized properly).

We can default it in the API resource, the model or directly in the DB.

markus101:
#### Database Migration
NO

#### Description
Removes the requirement to send the full JSON body (from `/series/lookup`) when adding a new series and requires only the basic information (so Sonarr knows where to put it and what quality to get it in).

Fixes #1403

### DIFF
--- a/src/NzbDrone.Api/Series/SeasonResource.cs
+++ b/src/NzbDrone.Api/Series/SeasonResource.cs
@@ -41,7 +41,7 @@ namespace NzbDrone.Api.Series
 
         public static List<Season> ToModel(this IEnumerable<SeasonResource> resources)
         {
-            return resources.Select(ToModel).ToList();
+            return resources?.Select(ToModel).ToList();
         }
     }
 }
--- a/src/NzbDrone.Api/Series/SeasonResource.cs
+++ b/src/NzbDrone.Api/Series/SeasonResource.cs
@@ -41,7 +41,7 @@ namespace NzbDrone.Api.Series
 
         public static List<Season> ToModel(this IEnumerable<SeasonResource> resources)
         {
-            return resources?.Select(ToModel).ToList();
+            return resources?.Select(ToModel).ToList() ?? new List<Season>();
         }
     }
 }
--- a/src/NzbDrone.Core.Test/TvTests/AddSeriesFixture.cs
+++ b/src/NzbDrone.Core.Test/TvTests/AddSeriesFixture.cs
@@ -1,0 +1,132 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using FizzWare.NBuilder;
+using FluentAssertions;
+using FluentValidation;
+using FluentValidation.Results;
+using Moq;
+using NUnit.Framework;
+using NzbDrone.Core.Exceptions;
+using NzbDrone.Core.MetadataSource;
+using NzbDrone.Core.Organizer;
+using NzbDrone.Core.Tv;
+using NzbDrone.Core.Test.Framework;
+using NzbDrone.Core.Tv.Events;
+using NzbDrone.Test.Common;
+
+namespace NzbDrone.Core.Test.TvTests
+{
+    [TestFixture]
+    public class AddSeriesFixture : CoreTest<AddSeriesService>
+    {
+        private Series _fakeSeries;
+
+        [SetUp]
+        public void Setup()
+        {
+            _fakeSeries = Builder<Series>
+                .CreateNew()
+                .With(s => s.Path = null)
+                .Build();
+        }
+
+        private void GivenValidSeries(int tvdbId)
+        {
+            Mocker.GetMock<IProvideSeriesInfo>()
+                  .Setup(s => s.GetSeriesInfo(tvdbId))
+                  .Returns(new Tuple<Series, List<Episode>>(_fakeSeries, new List<Episode>()));
+        }
+
+        private void GivenValidPath()
+        {
+            Mocker.GetMock<IBuildFileNames>()
+                  .Setup(s => s.GetSeriesFolder(It.IsAny<Series>(), null))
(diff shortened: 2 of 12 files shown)


src/NzbDrone.Api/Series/SeasonResource.cs
```diff
@@ -41,7 +41,7 @@ public static List<SeasonResource> ToResource(this IEnumerable<Season> models)
 
         public static List<Season> ToModel(this IEnumerable<SeasonResource> resources)
         {
-            return resources.Select(ToModel).ToList();
+            return resources?.Select(ToModel).ToList();
```
Taloth: I'd prefer a separate if like with the other resource converters.

src/NzbDrone.Core.Test/TvTests/AddSeriesFixture.cs
```diff
@@ -0,0 +1,132 @@
...
+                  .Returns(new ValidationResult(new List<ValidationFailure>
+                                                {
+                                                    new ValidationFailure("Path", "Test validation failure")
+                                                }));
+
+            Assert.Throws<ValidationException>(() => Subject.AddSeries(newSeries));
+
+            ExceptionVerification.ExpectedErrors(1);
+        }
+    }
```
Taloth: Maybe add test to check for duplicate slug while we're at it?

src/NzbDrone.Core/Tv/AddSeriesService.cs
```diff
@@ -0,0 +1,103 @@
...
+            newSeries.CleanTitle = newSeries.Title.CleanSeriesTitle();
+            newSeries.SortTitle = SeriesTitleNormalizer.Normalize(newSeries.Title, newSeries.TvdbId);
+            newSeries.Added = DateTime.UtcNow;
+
+            _seriesService.AddSeries(newSeries);
+
+            return newSeries;
+        }
+
+        private Series SeriesGetSeriesInfo(Series newSeries)
```
Taloth: Just `GetSeriesInfo`

src/NzbDrone.Core/Tv/AddSeriesService.cs
```diff
@@ -0,0 +1,103 @@
...
+
+            var series = tuple.Item1;
+
+            series.Path = newSeries.Path;
+            series.RootFolderPath = newSeries.RootFolderPath;
+            series.Monitored = newSeries.Monitored;
+            series.AddOptions = newSeries.AddOptions;
+            series.ProfileId = newSeries.ProfileId;
+            series.SeasonFolder = newSeries.SeasonFolder;
+            series.SeriesType = newSeries.SeriesType;
```
Taloth: Do all these properties have sensible defaults? Imagine someone adding a series without `{ ..., "monitored": true }`.

Also, it might be better to extract setting these properties to a separate method, or even class.
iirc we have some code that defines what the user is allowed to change on a series. It's a powerplay in SeriesModule between seriesService.UpdateService and first having to fetch the current state from the seriesService. Basically, SeriesResource.ToModel only applies properties the user is allowed to change. (https://github.com/Sonarr/Sonarr/blob/816cf608fcef07e375aa750a863d0a7365108981/src/NzbDrone.Api/Series/SeriesResource.cs#L208-L225)
So having the same list of 'mutable' properties on two locations isn't very maintainable.

src/NzbDrone.Api/Series/SeasonResource.cs
markus101: 👍 This was me testing the magic (and the waters).

Taloth: I'm not opposed to the ?. operator in general, but in this case having a separate `if` makes it more clear that it can accept and emit a null.

src/NzbDrone.Core/Tv/AddSeriesService.cs
markus101: We allow the TVDB ID to be changed via the API? That seems not great, also seasons makes sense (for monitoring), which I forgot about (though we'll want to make sure they were actually set, not null or empty).

src/NzbDrone.Api/Series/SeasonResource.cs
markus101: Yeah, returning null for a list is bad form anyways.

src/NzbDrone.Core/Tv/AddSeriesService.cs
kayone: how about `MergeSeriesInfo` or `GetCombinedSeriesInfo`

src/NzbDrone.Core/Tv/AddSeriesService.cs
```diff
@@ -0,0 +1,103 @@
...
+
+            newSeries = SeriesGetSeriesInfo(newSeries);
+
+            if (string.IsNullOrWhiteSpace(newSeries.Path))
+            {
+                var folderName = _fileNameBuilder.GetSeriesFolder(newSeries);
+                newSeries.Path = Path.Combine(newSeries.RootFolderPath, folderName);
+            }
+
+            var validationResult =_addSeriesValidator.Validate(newSeries);
```
kayone: I think this should be done in `SeriesService.AddSeries`

src/NzbDrone.Api/Series/SeasonResource.cs
kayone: +1 for `?`
-1 for returning null.

Issue #328, opened 2013-12-05
Templates for Notifications (Connections)


SonarrBot:
Ability to control what the message header and body looks like for notifications, not sure if it should be global or per notification.

Ultimate control would come from something like HBS templates, storing them in AppData and if they aren't there using defaults. If we went this route I think we would want to cache them on start up/first use and either wait for a restart to rebuild them (since that would be more consistent than a length of time).

simonk83:
Really hoping to see this one come about.  Discussion here:  https://forums.sonarr.tv/t/slight-change-to-pushbullet/1485/14

subzero79:
If this is low priority, can we have at least "on grab" for custom scripts?
If possible provide the indexer as env variable.

tmills1073:
I'd vote for this one as well.  The generic notification just means I have to open the app to see what the message is - it would be ideal to just have that information as the notification.

alexplattfuss:
I'd definitely like to see this one, coming from Sickbeard this is the only major complaint. It is an annoyance on mobile, I flip between disabling and enabling them. Either creating a template or switching the title to $series.' S'.$season.'E'.$episode.' '.$action.' - Sonarr'

deadman36g:
Surprised this has not been added as of yet. Pushbullet is a must.

bobbintb:
I have a custom script for email and SMS templates working. I got some other priorities at the moment but after that I will tweak the script for public consumption and release it.

bobbintb:
I made this workaround for the time being:

https://forums.sonarr.tv/t/custom-email-and-sms-notifications/14401

simonk83:
Nice. Kind of want something for push notifications though (Pushbullet in
my case).
Good job though.

bobbintb:
I don't use Pushbullet anymore but it should only require a small tweak of the curl command. I won't be able to do it but someone else might.

https://docs.pushbullet.com/#api-quick-start

samip5:
What do you think if the hardcoded `sonarr` could be the default which it will use when sending things and have a global setting for it too (if set use it)?

I'm thinking of working on some of those, but wondering what would the Sonarr dev team want it to be like or if they have a preference on how it should be done?

markus101:
A global setting doesn't make as much sense as an option per notification; a field (or two when there is a subject and body) per-notification type. In some ways it's similar to renaming (using tokens).

samip5:
[...]

PR #1598, opened 2016-12-22
Notification API Cleanup

kayone:
#### Database Migration
NO

#### Description
Cleaned up some of the API making them less boiler platy!

Two main parts:

-  `SupportsOnGrab` `SupportsOnDownload` etc, is no longer needed in the implementation, it's dynamically determined at runtime based on whether the implementation has overridden the matching method.  🎩 

- We had 20 different version of `Sonarr - Grabbed` text all over the place, now it all comes from base class, there is still room to make it even more DRY, but requires digging into the proxies etc.

### DIFF
--- a/src/NzbDrone.Core.Test/NotificationTests/NotificationBaseFixture.cs
+++ b/src/NzbDrone.Core.Test/NotificationTests/NotificationBaseFixture.cs
@@ -1,0 +1,120 @@
+﻿using System;
+using FluentAssertions;
+using FluentValidation.Results;
+using NUnit.Framework;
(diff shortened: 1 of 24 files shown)


markus101:
Changes look good though 

`We had 20 different version of Sonarr - Grabbed text all over the place, now it all comes from base class, there is still room to make it even more DRY, but requires digging into the proxies etc.`

Would probably be pulled when https://github.com/Sonarr/Sonarr/issues/328 is implemented (some people aren't thrilled that the notifications are prefixed so having it removable makes sense.).

PR #1603, opened 2016-12-28
Proper detection of OS/Runtime Name/Version 

kayone:
#### Database Migration
NO

#### Description
- Added support of detecting Linux/Unix version and distros
- Added support for detection of OSX version
- Added support for getting actual .NET/Mono versions
- Including the information above in the user-agent sent with HTTP requests

eg. `Sonarr/10.0.0.21189 (Windows; windows 6.2.9200.0) DOTNET 4.6.2`


.gitignore
```diff
@@ -133,3 +133,5 @@ output/*
 
 _start
 _temp_*/**/*
+
+src/.idea/
```
kayone: project Rider

src/Marr.Data/DataHelper.cs
```diff
@@ -109,7 +109,7 @@ public static string GetMemberName<T>(this Expression<Func<T, object>> member)
             var memberExpression = (member.Body as MemberExpression);
             if (memberExpression == null)
             {
-                memberExpression = (member.Body as UnaryExpression).Operand as MemberExpression;
+                memberExpression = ((UnaryExpression) member.Body).Operand as MemberExpression;
```
markus101: Not sure how many upstream changes have been made to Marr, but perhaps we should look at it and see if anything is of use? Changes for changes sake here seems unnecessary.

src/NzbDrone.Api/System/SystemModule.cs
```diff
@@ -13,21 +13,27 @@ public class SystemModule : NzbDroneApiModule
...
         private readonly IMainDatabase _database;
         private readonly ILifecycleService _lifecycleService;
 
         public SystemModule(IAppFolderInfo appFolderInfo,
                             IRuntimeInfo runtimeInfo,
+                            IPlatformInfo platoInfo, 
+                            IOsInfo osInfo,
                             IRouteCacheProvider routeCacheProvider,
-                            IConfigFileProvider configFileProvider,
+                            IConfigFileProvider configFileProvider,
```
markus101: Kill the extra spaces.

src/NzbDrone.Common/Http/HttpClient.cs
```diff
@@ -30,12 +30,14 @@ public class HttpClient : IHttpClient
         private readonly ICached<CookieContainer> _cookieContainerCache;
         private readonly List<IHttpRequestInterceptor> _requestInterceptors;
         private readonly IHttpDispatcher _httpDispatcher;
+        private readonly IUserAgentBuilder _userAgentBuilder;
 
-        public HttpClient(IEnumerable<IHttpRequestInterceptor> requestInterceptors, ICacheManager cacheManager, IRateLimitService rateLimitService, IHttpDispatcher httpDispatcher, Logger logger)
+        public HttpClient(IEnumerable<IHttpRequestInterceptor> requestInterceptors, ICacheManager cacheManager, IRateLimitService rateLimitService, IHttpDispatcher httpDispatcher, IUserAgentBuilder userAgentBuilder, Logger logger)
```
markus101: Getting really long, move them to separate lines?

src/NzbDrone.Host/AccessControl/UrlAclAdapter.cs
```diff
@@ -38,11 +39,14 @@ public List<string> Urls
         public UrlAclAdapter(INetshProvider netshProvider,
                              IConfigFileProvider configFileProvider,
                              IRuntimeInfo runtimeInfo,
+                             IOsInfo osInfo,
+
```
markus101: Extraneous new line.

Taloth:
Back when I added the useragent string, I intentionally left out the OS patch and revision versions from the user agent because I didn't want to leak too much details to servers. I suggest we keep that. (So use `osVersion.ToString(2)` instead)

We should think hard before adding additional info to the user-agent string. Such as distro and .net version, I don't think they're needed. We really should avoid adding all that, just look at the myriad of IE useragents with that horrible tendency.
I'd suggest to only log the new info to the log file and keeping it out of the User Agent string entirely.

A cosmetics note: I believe .net core has a couple of classes to build useragent strings, it might be worth looking into that for consistency.

The feeling I'm getting with this PR: What problem are we trying to solve? Having the distro etc in the logs could be beneficial, but it's not something we've _needed_ so far. Playing Devil's Advocate this time around. 😃

kayone:
The main reason I want the distro/,net/mono version is to know that people are running sonarr under. right now we have NO IDEA which distro/mono version is the most popular, maybe we've only had 2% of users actually running .NET 4 and 98% are running .NET 4.5 and we could've switched to 4.5 months ago, but we don't know. 

Knowing which distros people are using sonarr under gives us the same insight, we could setup test agents on that distro.

I've already suggested to @markus101 that we should make the _extended_ UA string for sonarr requests only and send the old string to 3rd parties but he didn't think it's a big deal and to be honest I'm on the same boat. This is pretty much as anonymous as it gets, EVERY browser out there sends out way more information that this on _every_ request and no one is complaining.

markus101:
I'm not sure there is value in 3rd parties knowing the exact mono version someone is running, that seems unnecessary, knowing that it's running under mono seems like a good compromise there, that way if something went crazy that only affected Windows users or only people running under mono we could still get that information (like the DNS caching issue under mono).

Getting these stats will certainly help us focus on platforms people use (building packages being one thing that comes to mind), but sending them to everything Sonarr connects too does seem a bit much on second thought.

Taloth:
Send the data in the services update requests, coz that's exactly where that information would be useful and you got the infrastructure to handle it. We could hold back sending .net 4.5 updates to windows XP users (which can't get .net 4.5), or force to a legacy branch with extra warning on the UI.
Including them as separate query params, or switch to POST with json should do it. QueryParam's probably best.

This should also allow you get all the stats you need about which distro's to tests against. (That's a valid reason for wanting to get distro details, btw. So you've convinced me of it's usefulness)

I'm gonna stick with my original position when it comes to keeping the User Agent minimal.

kayone:
Ha!, I originally had the data as query string params in the update URL, but then thought that might look a bit strange and it was getting too long, but I'm ok with adding it there.

src/NzbDrone.Api/System/SystemModule.cs
```diff
@@ -13,21 +13,26 @@ public class SystemModule : NzbDroneApiModule
...
+        private readonly IPlatformInfo _platoInfo;
+        private readonly IOsInfo _osInfo;
         private readonly IRouteCacheProvider _routeCacheProvider;
         private readonly IConfigFileProvider _configFileProvider;
         private readonly IMainDatabase _database;
         private readonly ILifecycleService _lifecycleService;
 
         public SystemModule(IAppFolderInfo appFolderInfo,
                             IRuntimeInfo runtimeInfo,
+                            IPlatformInfo platoInfo,
```
markus101: `platformInfo`

src/NzbDrone.Api/System/SystemModule.cs
```diff
@@ -13,21 +13,26 @@ public class SystemModule : NzbDroneApiModule
     {
         private readonly IAppFolderInfo _appFolderInfo;
         private readonly IRuntimeInfo _runtimeInfo;
+        private readonly IPlatformInfo _platoInfo;
```
markus101: `_platformInfo`

PR #1612, opened 2017-01-05
Sentry: Calm the fuck down!

kayone:


src/NzbDrone.Common/Instrumentation/Sentry/SentryTarget.cs
```diff
@@ -52,6 +54,37 @@ public SentryTarget(string dsn)
             _client.Tags.Add("culture", Thread.CurrentThread.CurrentCulture.Name);
             _client.Tags.Add("branch", BuildInfo.Branch);
             _client.Tags.Add("version", BuildInfo.Version.ToString());
+
+            _debounce = new SentryDebounce();
+        }
+
+
+        private List<string> GetFingerPrint(LogEventInfo logEvent)
```
Taloth: Why not just return a string? I think you can put the string.Join in this method.

src/NzbDrone.Common/Instrumentation/Sentry/SentryTarget.cs
```diff
@@ -52,6 +54,37 @@ public SentryTarget(string dsn)
...
+            var fingerPrint = new List<string>
+            {
+                logEvent.Level.Ordinal.ToString(),
+            };
+
+            var lineNumber = "";
+
+            if (logEvent.StackTrace != null)
+            {
+                var stackFrame = logEvent.StackTrace.GetFrame(logEvent.UserStackFrameNumber);
```
Taloth: As mentioned, get rid of the stackframe&line number stuff. Use LogEventInfo.LoggerName + LogEventInfo.Message instead.

src/NzbDrone.Common/Instrumentation/Sentry/SentryDebounce.cs
```diff
@@ -0,0 +1,32 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace NzbDrone.Common.Instrumentation.Sentry
+{
+    public class SentryDebounce
+    {
+        private readonly Dictionary<string, DateTime> _dictionary;
```
Taloth: Why not use Cached? It has a built in expiry iirc.

Also, maybe have a counter on the number of events? If that's relevant to Sentry. (Not a priority, do it later when it's stabilized)

src/NzbDrone.Common/Instrumentation/Sentry/SentryTarget.cs
```diff
@@ -52,6 +54,37 @@ public SentryTarget(string dsn)
...
+                {
+                    lineNumber = $"#{stackFrame.GetFileLineNumber()}";
+                }
+            }
+
+            fingerPrint.Add(logEvent.LoggerName + lineNumber);
+
+            if (logEvent.Exception != null)
+            {
+                fingerPrint.Add(logEvent.Exception.GetType().Name);
```
Taloth: Maybe include the InnerException type as well if present? Helps segregate between different causes.

src/NzbDrone.Common/Instrumentation/Sentry/SentryDebounce.cs
kayone: sentry doesn't it's own internal count.

PR #1620, opened 2017-01-14
Synology diskstation

margaale:
#### Database Migration
NO

#### Description
Here is my attempt to get download station working. No tests writen yet.  Basic functionality is working. All feedback is welcome. I only tested it non locally, still have to query filestation in order to get download directory. 
 It's based on the previous pull request  #1160. So thank you man. 

DownloadStation API http://global.download.synology.com/download/Document/DeveloperGuide/Synology_Download_Station_Web_API.pdf

FileStation API
http://global.download.synology.com/download/Document/DeveloperGuide/Synology_File_Station_API_Guide.pdf

#### Todos
- [x] Tests
- [x] TV category 

#347


src/NzbDrone.Core/Download/Clients/DownloadStation/DownloadStation.cs
```diff
@@ -0,0 +1,302 @@
...
+        {
+            IEnumerable<DownloadStationTask> torrents;
+
+            try
+            {
+                torrents = _proxy.GetItems(Settings);
+            }
+            catch (DownloadClientException ex)
+            {
+                _logger.Error(ex, ex.Message);
```
markus101: Don't use `ex.Message` as the message, if nothing is set it'll default to that anyways. Ideally we'd supply a custom message here for consistency.

src/NzbDrone.Core/Download/Clients/DownloadStation/DownloadStation.cs
```diff
@@ -0,0 +1,302 @@
...
+
+            var items = new List<DownloadClientItem>();
+
+            foreach (var torrent in torrents)
+            {
+                var outputPath = new OsPath(torrent.Additional.Detail["destination"]);
+
+                if (Settings.TvDirectory.IsNotNullOrWhiteSpace())
+                {
+                    if (!Settings.TvDirectory.Equals(outputPath.FullPath))
```
markus101: If statements (even single line ones) either need to be entirely on one line or wrapped in braces, this prevents a future change from being a bug due to if being left outside of the if statement.

`if (!Settings.TvDirectory.Equals(outputPath.FullPath)) continue;`

or

```
if (!Settings.TvDirectory.Equals(outputPath.FullPath)) {
  continue;
}
```

markus101: If `TvDirectory` is used and the torrent has multiple files it will be in a sub directory so this may not work.

src/NzbDrone.Core/Download/Clients/DownloadStation/DownloadStation.cs
```diff
@@ -0,0 +1,302 @@
...
+                        continue;
+                }
+                else if (Settings.TvCategory.IsNotNullOrWhiteSpace())
+                {
+                    var directories = outputPath.FullPath.Split('\\', '/');
+                    if (!directories.Contains(Settings.TvCategory))
```
markus101: Same here with the `if`.

src/NzbDrone.Core/Download/Clients/DownloadStation/DownloadStation.cs
```diff
@@ -0,0 +1,302 @@
...
+                        return torrent.StatusExtra["unzip_progress"];
+                }
+            }
+
+            return null;
+        }
+
+        private DownloadItemStatus GetTorrentStatus(DownloadStationTaskStatus status)
+        {
+            if (StatusTable.ContainsKey(status))
```
markus101: It would be more clear to check the actual statuses here and default to Downloading rather than setting a dictionary and checking if the value exists in it.

src/NzbDrone.Core/Download/Clients/DownloadStation/DownloadStation.cs
```diff
@@ -0,0 +1,302 @@
...
+        {
+            var data = _proxy.GetConfig(Settings);
+            
+            try
+            {
+                var path = data["default_destination"] as string;
+
+                if (Settings.TvCategory.IsNotNullOrWhiteSpace())
+                {
+                    path = string.Format("{0}/.{1}", path, Settings.TvCategory);
```
markus101: `path = $"{path}/.{Settings.TvCategory}";`

The user shouldn't be forced to use a hidden folder though, we reverted that behaviour in Transmission and this should behave the same.

src/NzbDrone.Core/Download/Clients/DownloadStation/DownloadStation.cs
```diff
@@ -0,0 +1,302 @@
...
+            if (items.Any())
+            {
+                try
+                {
+                    return items.Single().Id;
+                }
+                catch (InvalidOperationException)
+                {
+                    var ids = string.Join(",", items.Select(t => t.Id));
+                    _logger.Debug(string.Format("Multiple {0} tasks in Download Station: {1}", magnetLink, ids));
```
markus101: Don't use `string.Format` within the logger, `_logger.Debug("Multiple {0} tasks in Download Station: {1}", magnetLink, ids);` will work.

src/NzbDrone.Core/Download/Clients/DownloadStation/DownloadStation.cs
```diff
@@ -0,0 +1,302 @@
...
+            if (Settings.TvDirectory.IsNotNullOrWhiteSpace())
+            {
+                return Settings.TvDirectory;
+            }
+            else if (Settings.TvCategory.IsNotNullOrWhiteSpace())
+            {
+                var config = _proxy.GetConfig(Settings);
+                var destDir = (string)config.GetValueOrDefault("default_destination");
+
+                return string.Format("{0}/{1}", destDir.TrimEnd('/'), Settings.TvCategory);
```
markus101: Use `$""` instead of `string.Format`.

src/NzbDrone.Core/Download/Clients/DownloadStation/DownloadStation.cs
```diff
@@ -0,0 +1,302 @@
...
+            if (items.Any())
+            {
+                try
+                {
+                    return items.Single().Id;
+                }
+                catch (InvalidOperationException)
+                {
+                    var ids = string.Join(",", items.Select(t => t.Id));
+                    _logger.Debug(string.Format("Multiple {0} tasks in Download Station: {1}", filename, ids));
```
markus101: Same thing here with `string.Format` inside of the logger.

src/NzbDrone.Core/Download/Clients/DownloadStation/DownloadStation.cs
```diff
@@ -0,0 +1,302 @@
...
+            {
+                _logger.Error(ex, ex.Message);
+                return new NzbDroneValidationFailure("Username", "Authentication failure")
+                {
+                    DetailedDescription = string.Format("Please verify your username and password. Also verify if the host running Sonarr isn't blocked from accessing {0} by WhiteList limitations in the {0} configuration.", Name)
+                };
+            }
+            catch (WebException ex)
+            {
+                _logger.Error(ex, ex.Message);
```
markus101: `_logger.Error(ex);`

src/NzbDrone.Core/Download/Clients/DownloadStation/DownloadStation.cs
```diff
@@ -0,0 +1,302 @@
...
+            try
+            {
+                return ValidateVersion();
+            }
+            catch (DownloadClientAuthenticationException ex)
+            {
+                _logger.Error(ex, ex.Message);
+                return new NzbDroneValidationFailure("Username", "Authentication failure")
+                {
+                    DetailedDescription = string.Format("Please verify your username and password. Also verify if the host running Sonarr isn't blocked from accessing {0} by WhiteList limitations in the {0} configuration.", Name)
```
markus101: `$""`

src/NzbDrone.Core/Download/Clients/DownloadStation/DownloadStation.cs
```diff
@@ -0,0 +1,302 @@
...
+                    return new NzbDroneValidationFailure("Host", "Unable to connect")
+                    {
+                        DetailedDescription = "Please verify the hostname and port."
+                    };
+                }
+                return new NzbDroneValidationFailure(string.Empty, "Unknown exception: " + ex.Message);
+            }
+            catch (Exception ex)
+            {
+                _logger.Error(ex, ex.Message);
```
markus101: No `ex.Message`

src/NzbDrone.Core/Download/Clients/DownloadStation/DownloadStationAPIInfo.cs
```diff
@@ -0,0 +1,29 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace NzbDrone.Core.Download.Clients.DownloadStation
+{
+    public class DownloadStationAPIInfo
```
markus101: `DownloadStationApiInfo`

src/NzbDrone.Core/Download/Clients/DownloadStation/DownloadStationProxy.cs
```diff
@@ -0,0 +1,273 @@
...
+        {
+            Resources = new Dictionary<SynologyApi, string>
+            {
+                {SynologyApi.Info, "query.cgi" },
+            };
+        }
+
+        public DownloadStationProxy(ICacheManager cacheManager, IHttpClient httpClient, Logger logger)
+        {            
+            this._httpClient = httpClient;
```
markus101: Same here, don't use `this.`

src/NzbDrone.Core/Download/Clients/DownloadStation/DownloadStationProxy.cs
```diff
@@ -0,0 +1,273 @@
...
+
+        private readonly IHttpClient _httpClient;
+        private readonly Logger _logger;        
+        private bool _authenticated;
+
+        static DownloadStationProxy()
+        {
+            Resources = new Dictionary<SynologyApi, string>
+            {
+                {SynologyApi.Info, "query.cgi" },
```
markus101: `{ SynologyApi.Info, "query.cgi" }`

src/NzbDrone.Core/Download/Clients/DownloadStation/DownloadStationProxy.cs
```diff
@@ -0,0 +1,273 @@
...
+            }
+
+            return new List<DownloadStationTask>();
+        }
+
+        public Dictionary<string, object> GetConfig(DownloadStationSettings settings)
+        {
+            var arguments = new Dictionary<string, object>
+            {
+                {"api", "SYNO.DownloadStation.Info"},
```
markus101: `{ "api", "SYNO.DownloadStation.Info" },` (and the same for the following lines).

src/NzbDrone.Core/Download/Clients/DownloadStation/DownloadStationProxy.cs
```diff
@@ -0,0 +1,273 @@
...
+
+                throw;
+            }
+        }
+        
+        public void RemoveItem(string downloadId, bool deleteData, DownloadStationSettings settings)
+        {
+            var arguments = new Dictionary<string, object>
+            {
+                {"api", "SYNO.DownloadStation.Task"},
```
markus101: `{ "api", "SYNO.DownloadStation.Task" },`

src/NzbDrone.Core/Download/Clients/DownloadStation/DownloadStationProxy.cs
```diff
@@ -0,0 +1,273 @@
...
+                {"force_complete", false}
+            };
+
+            try
+            {
+                var b = ProcessRequest(SynologyApi.DownloadStationTask, arguments, settings);
+            }
+            catch (DownloadClientException)
+            {
+                _logger.Debug(string.Format("Failed to remove item {0} from Download Station", downloadId));
```
markus101: `string.Format` inside the logger.

src/NzbDrone.Core/Download/Clients/DownloadStation/DownloadStationProxy.cs
```diff
@@ -0,0 +1,273 @@
...
+                 { "version", "1"},
+                 { "method", "query"},
+                 { "query", "SYNO.API.Auth, SYNO.DownloadStation.Info, SYNO.DownloadStation.Task"},
+             };
+
+            var downloadStationResponse = ProcessRequest<DownloadStationInfoResponse>(SynologyApi.Info, arguments, settings);
+
+            if (downloadStationResponse.Success == true)
+            {
+                DownloadStationAPIInfo infoResponeDSAuth = downloadStationResponse.Data["SYNO.API.Auth"];
```
markus101: USe `var` instead of `DownloadStationAPIInfo`

src/NzbDrone.Core/Download/Clients/DownloadStation/DownloadStationProxy.cs
```diff
@@ -0,0 +1,273 @@
...
+            }
+
+            var response = _httpClient.Execute(request);
+
+            if (response.StatusCode == HttpStatusCode.OK)
+            {
+                return Json.Deserialize<DownloadStationResponse<T>>(response.Content);
+            }
+
+            throw new Exception("Error " + response.StatusCode);
```
markus101: Throw a `DownloadStationException` and `$"Error {response.StatusCode}"

src/NzbDrone.Core/Download/Clients/DownloadStation/DownloadStationProxy.cs
```diff
@@ -0,0 +1,273 @@
...
+              
+        private HttpRequestBuilder BuildRequest(DownloadStationSettings settings, SynologyApi api, Dictionary<string, object> arguments)
+        {
+            if (!Resources.ContainsKey(api))
+            {
+                GetApiVersion(settings);
+            }
+
+            var requestBuilder = new HttpRequestBuilder(false, settings.Host, settings.Port)
+                .Resource(String.Format("webapi/{0}", Resources[api]));
```
markus101: `$""`

src/NzbDrone.Core/Download/Clients/DownloadStation/DownloadStationSettings.cs
```diff
@@ -0,0 +1,62 @@
...
+    {
+        public DownloadStationSettingsValidator()
+        {
+            RuleFor(c => c.Host).ValidHost();
+            RuleFor(c => c.Port).GreaterThan(0).LessThan(65536);
+
+            RuleFor(c => c.TvCategory).Matches(@"^\.?[-a-z]*$", RegexOptions.IgnoreCase).WithMessage("Allowed characters a-z and -");
+
+            RuleFor(c => c.TvCategory).Empty()
+                .When(c => c.TvDirectory.IsNotNullOrWhiteSpace())
```
markus101: Align with the `.Empty()` above

src/NzbDrone.Core/Download/Clients/DownloadStation/DownloadStationSettings.cs
```diff
@@ -0,0 +1,62 @@
...
+
+        [FieldDefinition(5, Label = "Directory", Type = FieldType.Textbox, HelpText = "Optional location to put downloads in, leave blank to use the default Download Station location")]
+        public string TvDirectory { get; set; }
+
+        [FieldDefinition(6, Label = "Use SSL", Type = FieldType.Checkbox)]
+        public bool UseSsl { get; set; }
+
+        public DownloadStationSettings()
+        {                        
+            this.Host = "192.168.0.10";
```
markus101: And here, drop `this.`

src/NzbDrone.Core/Download/Clients/DownloadStation/DownloadStationTask.cs
```diff
@@ -0,0 +1,87 @@
...
+
+        public DownloadStationTaskAdditional Additional { get; set; }
+
+        private string GetMessage()
+        {
+            if (StatusExtra != null)
+            {
+                if (Status == DownloadStationTaskStatus.Extracting)
+                {
+                    return string.Format("Extracting: %{0}", int.Parse(StatusExtra["unzip_progress"]));
```
markus101: `$""`

src/NzbDrone.Core/Download/Clients/DownloadStation/DownloadStationTorrent.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace NzbDrone.Core.Download.Clients.DownloadStation
+{
+    class DownloadStationTorrent
```
markus101: Empty class, is it used?

src/NzbDrone.Core/Download/Clients/DownloadStation/Responses/DownloadStationError.cs
```diff
@@ -0,0 +1,78 @@
...
+    {
+        private static readonly Dictionary<int, string> CommonMessages;
+        private static readonly Dictionary<int, string> AuthMessages;
+        private static readonly Dictionary<int, string> DownloadStationTaskMessages;
+
+        static DownloadStationError()
+        {
+            CommonMessages = new Dictionary<int, string>
+            {
+                {100, "Unknown error"},
```
markus101: `{ 100, "Unknown error" },` (and for the other dictionaries below).

src/NzbDrone.Core/Download/Clients/DownloadStation/Responses/DownloadStationError.cs
```diff
@@ -0,0 +1,78 @@
...
+                return Code == 105 || Code == 106 || Code == 107;
+            }
+        }
+
+        public string GetMessage(SynologyApi api)
+        {
+            switch (api)
+            {
+                case SynologyApi.Auth:
+                    if (AuthMessages.ContainsKey(Code))
```
markus101: I think this might be cleaner:

```
if (api == SynologyApi.Auth && AuthMessages.ContainsKey(Code)) {
  return AuthMessages[Code];
}
...
```

src/NzbDrone.Core/Download/Clients/DownloadStation/DownloadStation.cs
```diff
@@ -0,0 +1,302 @@
...
+        static DownloadStation()
+        {
+            StatusTable.Add(DownloadStationTask.DownloadStationTaskStatus.Waiting, DownloadItemStatus.Queued);
+            StatusTable.Add(DownloadStationTask.DownloadStationTaskStatus.Error, DownloadItemStatus.Failed);
+            StatusTable.Add(DownloadStationTask.DownloadStationTaskStatus.Paused, DownloadItemStatus.Paused);
+            StatusTable.Add(DownloadStationTask.DownloadStationTaskStatus.Seeding, DownloadItemStatus.Completed);
+        }
+
+        public DownloadStation(IDownloadStationProxy proxy,
+                     ITorrentFileInfoReader torrentFileInfoReader,
```
margaale: Sorry, what does Nit mean??

src/NzbDrone.Core/Download/Clients/DownloadStation/DownloadStation.cs
```diff
@@ -0,0 +1,302 @@
...
+            {
+                return StatusTable[status];
+            }
+
+            return DownloadItemStatus.Downloading;
+        }
+
+        public override DownloadClientStatus GetStatus()
+        {
+            var data = _proxy.GetConfig(Settings);
```
Taloth: This is outside the try catch, how's the catch ever gonna catch a meaningful exception?

src/NzbDrone.Core/Download/Clients/DownloadStation/DownloadStation.cs
Taloth: Nitpick. Like "Yes I know, almost not worth mentioning but I'm gonna mention it anyway."

src/NzbDrone.Core/Download/Clients/DownloadStation/DownloadStationProxy.cs
```diff
@@ -0,0 +1,273 @@
...
+
+            var downloadStationResponse = Json.Deserialize<DownloadStationResponse<DownloadStationAuthResponse>>(response.Content);
+
+            var authResponse = Json.Deserialize<DownloadStationResponse<DownloadStationAuthResponse>>(response.Content);
+
+            _authenticated = authResponse.Success;
+
+            if (!_authenticated)
+            {
+                throw new Exception(downloadStationResponse.Error.GetMessage(SynologyApi.Auth));
```
Taloth: Same here, throw a specific exception such as DownloadClientException or DownloadStationException. There are some options.

src/NzbDrone.Core/Download/Clients/DownloadStation/DownloadStation.cs
margaale: Could you explain me a little more about TvDirectory and Category options? I'm not a user of sonnar yet, i was going to try it because of a friend but when i saw that this client was missing i started coding. What is the purpose of each one of them?

src/NzbDrone.Core/Download/Clients/DownloadStation/DownloadStation.cs
margaale: So, should i return a new DownloadClientStatus()?

Taloth: My comment was more toward markus.
replace it with `$"{path}/{Settings.TvCategory}";` for now and ignore what I said.

src/NzbDrone.Core/Download/Clients/DownloadStation/DownloadStationSettings.cs
margaale: Sorry. That's the is address of my syno :-P

src/NzbDrone.Core/Download/Clients/DownloadStation/DownloadStation.cs
Taloth: Sure.
Category is intended to filter the list of downloads provided by the download client. Many download clients have inherent support for categories (sometimes called labels) and support filtering on the api calls, which drastically improves performance if the client has hundreds of torrents.
Clients like Transmission don't support categories, so we have resorted to a workaround: We take transmission's default download dir and tack on the category name as a subdirectory.
That way we can filter out the api results so we don't try parse downloads that were added directly to transmission by the user or another app.
Sometimes ppl want to 'override' where the download is placed, instead of the default download dir.
Deluge (which supports labels) supports setting a download folder per label. But it's one of the few clients that support that.
So the TvDirectory option is intended to 'override' the default download directory in the torrent client.
Thus TvDirectory is mutually exclusive with Category, coz DownloadDIr=TvDirectory OR DownloadDir=DefaultDownloadDir+Category (OR DownloadDir=DefaultDownloadDir if none are specified).

The download directory can be seen as a 'base' directory.
Torrents with just one file are often placed directly in that directory. But Torrents with multiple files are placed in a new subdirectory (usually the Torrent Name). Sonarr has no control over this, torrent clients simply work that way.
You'll have to handle that properly in the GetOutputPath function.

What markus warned about is that you cannot plainly string compare the TvDirectory (the 'base' directory) with the OutputPath, coz it's a parent-child relation. There are helper functions to deal with that.

margaale: I see, i think i get how that works now. Another question, does sonarr create those directories in the destination or they should be created by the user before hand?

Taloth: Sonarr doesn't create any of those directories.
The subdirectories inside the TvDirectory are created by the torrent client as needed. So might the Category subdirectory, depending on the client.
I'd say just test it out.

margaale: I'm sure that download station needs an existing directory in order to add the task, if its an invalid one it returns "invalid directory error code". But gonna try again, maybe it was another the error the one i saw.

src/NzbDrone.Core/Download/Clients/DownloadStation/DownloadStation.cs
```diff
@@ -0,0 +1,302 @@
...
+            {
+                _logger.Error(ex, ex.Message);
+                return Enumerable.Empty<DownloadClientItem>();
+            }
+
+            var items = new List<DownloadClientItem>();
+
+            foreach (var torrent in torrents)
+            {
+                var outputPath = new OsPath(torrent.Additional.Detail["destination"]);
```
margaale: Not yet, so this is going to change in the future. This directory its not the physical path, i'm going to consume another api (filestation) in order to get that. But if i set a remote path this should be working as it is i think. At least that's what i''ve understood until now of mapping  works. Again, i'm not used to sonarr, so if i'm wrong please correct me. This should be the case when sonarr is not running onto the diskstation, shouldn't be? Sorry if my english is not good enough, i'm from Argentina (spanish here).

src/NzbDrone.Core/Download/Clients/DownloadStation/DownloadStation.cs
```diff
@@ -0,0 +1,302 @@
...
+            }
+
+            return null;
+        }
+
+        protected override string AddFromTorrentFile(RemoteEpisode remoteEpisode, string hash, string filename, byte[] fileContent)
+        {           
+            _proxy.AddTorrentFromData(fileContent, GetDownloadDirectory(), Settings);
+
+            var items = _proxy.GetItems(Settings).Where(t => t.Additional.Detail["uri"] == filename);
```
margaale: To your first question. No i can't the only way to query the api is using its internal id which you have to get asking for all the files and seek it using the filename. Grrrrr, awfull i wish they would listen...

margaale: Your second question, i haven't test this method, so there could be some errors like the one you say.

src/NzbDrone.Core/Download/Clients/DownloadStation/DownloadStation.cs
Taloth: That's exactly the point where Ollie was before other matters took precedence.
The workaround is to use Remote Mapping, but to get it properly automated you need to interface with FileStation. It didn't look horribly complicated, just needs an implementation and some caching.
But if Sonarr runs on another device then yes, you'd still need Remote Mapping.

Don't hesitate to ask about how things work in Sonarr. But you seem to be doing fine.
One note is that you definitely will need to test the different scenarios, such as multi-file torrent, magnet-link, different states, missing subdir. etc Coz we can't see those things from inspecting the code. (This is why the Proxy classes are relatively 'thin', it's to make it easy to mock the api and run unittests on the logic.)

PS: Your english is fine, easy to understand.

margaale:
Done all the changes, or at least thats what i tried. What do you think about handling Authorization problems the way i did? I'd love to get some other user/betatester to try this code too, if you can get someone i'll be very gratefull...

src/NzbDrone.Core/Download/Clients/DownloadStation/DownloadStation.cs
margaale: Gonna start working on this today.  Could you give some high level steps to follow? Is there any class that I should inherit from? I was thinking adding a method to the proxy to consume the new api. Is there any facility already implemented for caching? How does remote mapping come into play here?

Taloth: 1) You already have DownloadStationFixture , which correctly inherit from DownloadClientFixtureBase<TClient>, look at existing subclasses for inspiration, TransmissionFixtureBase is probably pretty close.

2) For filestation you might want to create a second IFileStationProxy (api) and ISharedFolderResolver (service+caching), considering it's a different service on the Syno. IDownloadStationProxy instead of IFileStationProxy might be okay too, I think, but separate class is cleaner. @markus101  may have an opinion.

3) Place it in the *.Clients.DownloadStation namespace.

4) Check out ICacheManager (factory) and ICached, there are quite a few places where caching is used. In your case I'd use ICached<string> so you can store and track individual mappings.

5) Good question, coz this is tricky. We can add downloads to DownloadStation, for that we'd use the Shared Folder path, not the filesystem path.
The GetItems function should first filter by Category/TvDirectory.
Then it runs GetOutputPath, which should resolve the Shared Folder.
Finally RemotePathMapping is applied.
Probably an idea to write a list/table of all scenarios.

margaale: Yeah, I like better separating the proxies, but that is gonna cost me a little work because i'll have to refactor the code in order to create a base class to share some common functionality between both of them. After doing that i'll come back here to ask you some more questions.

Regards

margaale: Done. Caching is working with no expiration date because I don't know how to estimate a good number.

 Now to the remote mapping part.

Possible scenarios:

Sonnar running on a Diskstation
	I guess SharedFolderResolver is good enough here because it will always resolve to a phisical path that sonarr has access to.
	
Sonnar not running on a Diskstation
	Here we need to configure RemoteMapping, the question is with which one? The shared folder or the physical one? I'm voting for shared because it's easier for the user to remember.

What do you think of this, am i missing something?

margaale:
Is there a way to force sonarr into using a torrent file instead of a Magnet link to download an episode? I wanna test that path in the code...

margaale:
Sorry, did you guys have time to see the question thah i left here?

Taloth:
I've been having a couple of busy days at work, so haven't been able to sit down properly.

Test the torrent file in live environment or unittest? If live, then just use the debugger, or use the [push api](https://github.com/Sonarr/Sonarr/wiki/Release-Push). If unittests, don't supply a torrent url, only magnet.

As for the auth code. Personally I would've have used re-entrancy and instead just structured the code differently. But that's somewhat of a personal choice.

I'll look through the latest changes later, probably tomorrow.

margaale:
Getting back to you, i'm leaving some thoughts here to discuss with you.

- **_Caching:_**
It's working with no expiration date because I don't know how to estimate a good number.

- **_Remote mapping:_**
*_Sonnar running on a Diskstation:_ I guess SharedFolderResolver is good enough here because it will always resolve to a phisical path that sonarr has access to.
*_Sonnar not running on a Diskstation:_ I need to configure RemoteMapping, the question is with which one? The shared folder or the physical one? I'm voting for shared because it's easier for the user to remember.

What do you think of this, am i missing something?

foXaCe:
@margaale hello, do you project to merge ?

margaale:
@foXaCe, i'm waiting for @Taloth to discuss some issues with him/her. Cannot give you an ETA right now.

margaale:
Eventually yes. But it's not me who is gonna take that decision...

foXaCe:
@margaale it possible to push this on Radarr please  ?

https://github.com/Radarr/Radarr

margaale:
@foXaCe I guess I could, talk to them to see if they want me to. The might want to wait until this PR is merged so they could merge into their repo.

foXaCe:
@Taloth @markus101  please merge the pull for my synology
thank

markus101:
@foXaCe that will happen after it's reviewed and tested, not before.

margaale:
@foXaCe I mean you need talk to the developers of radarr, not here...

src/NzbDrone.Core/Download/Clients/DownloadStation/DownloadStation.cs
```diff
@@ -0,0 +1,323 @@
...
+            _proxy.RemoveTorrent(downloadId, deleteData, Settings);            
+        }
+
+        protected override string AddFromMagnetLink(RemoteEpisode remoteEpisode, string hash, string magnetLink)
+        {
+            _proxy.AddTorrentFromUrl(magnetLink, GetDownloadDirectory(), Settings);
+
+            var items = _proxy.GetTorrents(Settings).Where(t => t.Additional.Detail["uri"] == magnetLink);
+
+            if (items.Any())
```
Taloth:
```
var item = items.SingleOrDefault();

if (item != null)
{
    return item.Id;
}

_logger.Debug("No such task {0} in Download Station", magnetLink);

throw new DownloadClientException("Failed to add task to Download Station");
```

src/NzbDrone.Core/Download/Clients/DownloadStation/DownloadStation.cs
```diff
@@ -0,0 +1,323 @@
...
+            }
+
+            return null;
+        }
+
+        protected override string AddFromTorrentFile(RemoteEpisode remoteEpisode, string hash, string filename, byte[] fileContent)
+        {           
+            _proxy.AddTorrentFromData(fileContent, GetDownloadDirectory(), Settings);
+
+            var items = _proxy.GetTorrents(Settings).Where(t => t.Additional.Detail["uri"] == filename);
```
Taloth: Did you test this method? Coz my previous comment was that the filename isn't sent to DownloadStation, so it's not something you can compare against.
Also, it might be prudent to stash an actual response in the Tests somewhere, even if it's not tested against.

src/NzbDrone.Core/Download/Clients/DownloadStation/Proxies/FileStationProxy.cs
```diff
@@ -0,0 +1,51 @@
...
+        }
+
+        public string GetPhysicalPath(string sharedFolder, DownloadStationSettings settings)
+        {
+            var arguments = new Dictionary<string, object>
+            {
+                { "api", "SYNO.FileStation.List" },
+                { "version", "2" },
+                { "method", "getinfo" },
+                { "path", $"[\"{ sharedFolder }\"]" },
```
Taloth: This looks like json to me. Why not `new [] { sharedFolder }.ToJson()`? That way everything in sharedFolder gets properly escaped.

src/NzbDrone.Core/Download/Clients/DownloadStation/Proxies/FileStationProxy.cs
```diff
@@ -0,0 +1,51 @@
...
+                { "additional", $"[\"real_path\"]" }
+            };
+
+            try
+            {
+                var response = ProcessRequest<FileStationListResponse>(SynologyApi.FileStationList, arguments, settings);
+
+                if (response.Success == true)
+                    return response.Data.Files.First().Additional["real_path"].ToString();
```
Taloth: Add `{ ... }`

(I know, when I first joined the Sonarr team I had to get used to the coding style as well :wink:)

src/NzbDrone.Core/Download/Clients/DownloadStation/SharedFolderResolverService.cs
```diff
@@ -0,0 +1,45 @@
...
+using NzbDrone.Core.Download.Clients.DownloadStation.Proxies;
+using NzbDrone.Core.ThingiProvider;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace NzbDrone.Core.Download.Clients.DownloadStation
+{
+    public class SharedFolderResolverService : ISharedFolderResolverService
```
Taloth: Nitpick, but `SharedFolderResolver` is enough, no need to tack `Service` on everything, coz everything is a 'service' anyway.

src/NzbDrone.Core/Download/Clients/DownloadStation/SharedFolderResolverService.cs
```diff
@@ -0,0 +1,45 @@
...
+            var physicalPath = _proxy.GetPhysicalPath(sharedFolder, Settings);
+            var sharedFolderCache = new SharedFolderCache(physicalPath);
+            _cache.Set(sharedFolder, sharedFolderCache);
+
+            return physicalPath;
+        }
+
+        public OsPath ResolvePhysicalPath(string sharedFolder)
+        {
+            return new OsPath(_cache.Find(sharedFolder)?.PhysicalPath ?? GetPhysicalPath(sharedFolder));
```
Taloth: ICached has a nice [method](https://github.com/Sonarr/Sonarr/blob/816cf608fcef07e375aa750a863d0a7365108981/src/NzbDrone.Common/Cache/ICached.cs#L17), including the lifetime property.
So `_cache.Get(sharedFolder, () => GetPhysicalPath(sharedFolder), TimeSpan.FromHour(1));` then you remove the _cache.Set from GetPhysicalPath, and everything sorts itself out and expires every once in a while. It's also more atomic.

You asked about the Cache lifetime. Just pick something sensible. It's fairly arbitrary, but 1h seemed reasonable, but feel free to increase it to 3h or so.

src/NzbDrone.Core/Download/Clients/DownloadStation/DownloadStation.cs
```diff
@@ -0,0 +1,323 @@
...
+                               IDiskProvider diskProvider,
+                               IRemotePathMappingService remotePathMappingService,
+                               Logger logger,
+                               ICacheManager cacheManager,
+                               ISharedFolderResolverService sharedFolderResolver)
+            : base(torrentFileInfoReader, httpClient, configService, diskProvider, remotePathMappingService, logger)
+        {
+            _proxy = proxy;
+            _sharedFolderResolver = sharedFolderResolver;
+            _sharedFolderResolver.Settings = Settings;
```
Taloth: Can't do that, Settings is assigned after the constructor. Easiest is probably to pass it along on the GetPhysicalPath method call. 
Also, SharedFolderResolver is a singleton, while DownloadStation might not be if ppl create multiple, so it has to be stateless. Meaning the Settings.Host+Port needs to be used as an additional key (prefix maybe) in the _cache.
It's really unlikely ppl add multiple diskstations, but if they do then it goes bad.

src/NzbDrone.Core/Download/Clients/DownloadStation/DownloadStation.cs
```diff
@@ -0,0 +1,323 @@
...
+
+                var physicalPath = _sharedFolderResolver.ResolvePhysicalPath(outputPath.FullPath);
+
+                outputPath = _remotePathMappingService.RemapRemoteToLocal(Settings.Host, outputPath);
+                
+                var item = new DownloadClientItem()
+                {
+                    Category = Settings.TvCategory,
+                    DownloadClient = Definition.Name,
+                    DownloadId = torrent.Id,
```
Taloth: Prefix this with something. The DownloadId is used in History to find related history items and should be as globally unique as possible.
Imagine someone buying a new DiskStation, plugging it in and suddenly get weird behavior coz the Id starts at 1 again and newly added torrents start getting tracked completely wrong.
I couldn't find a reference of how 'Id' looks, whether it's just an ID or some Guid, if it's a Guid then disregard this comment, coz then it's already unique.

Taloth:
@margaale my apologies for not reviewing any sooner. I honestly forgot. 😞

Your final question was about Remote Path Mapping.
I'm wondering if you could automatically deal with it, by detecting if it's on a local network. But I seriously doubt if it's worth the effort. Most users will run Sonarr on the DiskStation or in docker on the DiskStation and be happy with the SharedFolderResolver and need nothing else.

So tbh, I'd leave the physical mapping.

Some final remarks:
- I suggest you add a test Fixture for the SharedFolderResolver, and add tests to verify what happens when FileStation can't resolve a folder. Such as, does it log a Warning? or does it throw? If it throws, then that must be expected behavior coz DownloadStation might depend on it throwing. Otherwise DownloadStation might return a downloaditem with a shared folder instead of physical path.
- Likewise, DownloadStation needs a test to check whether it behaves appropriately if SharedFolderResolver throws. For example, by returning the DownloadItem, but with a status that prevents it from being imported... such as setting OutputPath = "". Or return nothing at all and log instead.
- Please manually test the AddFromTorrentFile with torrents containing a single file, and torrents containing multiple files. I just fear that the filename check will wreak havoc and prevent Sonarr from tracking downloads. Same goes with the DownloadId comment, if that doesn't get sorted we'll get all kinds of weird behavior.

src/NzbDrone.Core/Download/Clients/DownloadStation/DownloadStation.cs
margaale: I tested it, and i think it's not going to work either. The api doesn't receive a filename, that's why it's not being passed to the proxy...
I need further investigation on this issue

src/NzbDrone.Core/Download/Clients/DownloadStation/DownloadStation.cs
margaale: The id is somesthing like "task_xx" where xx is an incremental integer number. Let me see if i can get some other info from the diskstation that i can use, a serial number or something like that...

margaale:
@Taloth on the mapping comment, i could check if its diskstation is running on 127.0.0.1, that way i could skip the mapping, otherwise it would be done. Pretty straightfoward solution, i assume you has already came up with it, but i have to ask if what do you think of it anyway.

Taloth:
>On the mapping issue i could check if its diskstation is running on 127.0.0.1, that way i could skip the mapping, otherwise it would be done. Pretty straightfoward solution, i assume you has already came up with it, but i have to ask if what do you think of it anyway.

I'm assuming that with 'mapping' you mean RemotePathMapping, not the shared folder stuff.
The user specifically has to enter 'Host = 127.0.0.1' if they want to remap something locally (which does have some rare usecases). For a remote host you simply don't know how to map, you could turn it into an UNC path, but that only works when Sonarr runs on Windows. It's simply not worth the confusion. So my recommendation stands, just leave it as is.

The DownloadId & Cache Id are fine, it'll do the job. Better would be something unique to the syno but that's harder to get and using host:port is suffice for now. I really wish downloaders would use guids.
If someone replaces the syno and copies over the database it'll still go wrong, but it'll have to do.

>Have to do the tests yet.

That's the last thing left, afaik. It's looking good.

Taloth:
And do a complete squash and rebase on top of latest develop, instead of merges.

margaale:
That's what I been trying to do, hehehe. It's my first time with git and I've been trying to figure it out. I can rebase my develop branch against yours (upstream), but I cannot do the same with my feature branch. Could you give me a hand with that?

Taloth:
There are various ways to do it. Most UIs offer a mechanism one way or another. git [cmdline](https://git-scm.com/docs/git-rebase) (--interactive) works too.

margaale:
I was able to rebase the branch, but when i try to squash the commits i get merge errors. Any help that you could give me would be highly appreciated.

markus101:
You'll need to do an interactive rebase (`git rebase -i upstream/develop` will work if `Sonarr/Sonarr` is treated as upstream in your repository) and delete our commits (any commit that isn't new in this branch and already exists in develop).

margaale:
Could you tell me if 3870c514e2ea309eed807753dd7941174d58a14e is ok? I think that i did it right this time...

markus101:
There are still other commits there (55 total).

margaale:
There are a few of them(my commits) that are repeated, and there are others that are not mine in that number. All the important ones are in that squash

markus101:
Squashing eliminates the extra commits and you'll need to drop the ones not done by you because they shouldn't be there (they already exist in develop). How did you attempt to squash them?

margaale:
git rebase -i <hash>

i put an s in the beging of each line that i wanted to stay, i guess i didn't remove de other ones...

Taloth:
The final 'merge' is the problem.
`git checkout synology-diskstation`
`git reset --hard 3870c51`
That'll reset the branch to that specific commit, dropping the merge commit.

Then you need to _force_ push to github, basically saying: "ignore history, my local branch & history is the truth from now on."
`git push --force origin synology-diskstation` (or 'upstream' instead of 'origin', depends on your remote)

The _force_ part is what's been going wrong, unless you choose that option specifically it'll probably suggest to do _merge_ for you, which we obviously don't want.

margaale:
You were absolutely right about that _force_, that's was what i was missing. Sorry, as I've told you before, it's my first time with git, it can be really confusing...

margaale:
I've just got working addtorrentfile, it's not the most reliable method in the world but it does the job.

Taloth:
Commits since previous review:
https://github.com/Sonarr/Sonarr/commit/fd55c82 (reviewed)
https://github.com/Sonarr/Sonarr/commit/6ed90cf (reviewed)
https://github.com/Sonarr/Sonarr/commit/b4e114e (reviewed)
unnamed commit (reviewed via diff with a local copy I had)

@margaale Don't squash commits that need to be reviewed, otherwise we don't know what changed.

Some things:
- I don't think we need to have a separate object for SerialNumberCache, you can put it directly in ICached<string> in SerialNumberProvider. The reason I didn't comment on it for SharedFolderCache was because I figured the SharedFolderCache could be named SharedFolderMapping and contain both Name and PhysicalPath and potentially other information about the SharedFolder and potentially be the model the Proxy returns. Just a thought.
- For bonus points: Use HashConverter.GetHash on the serialnumber, not watertight, but at least the serialnumber won't be in plaintext.
- namespace usings. There's a visual studio setting that makes the System namespaces go on top. we have that option enabled.
- Are you absolutely sure that Transfer["size_downloaded"] and speed_downloaded will be valid and not null at all times? Magnet downloads may not have a size yet. Just mentioned it coz we've been burned on this with other downloaders that started returning weird values such as null and negative depending on the state of the torrent.

margaale:
Done all the sugestions.
Regarding size_downloaded and speed_download, i'm not sure but i've been using it and i didn't see any wrong number there. I've added a check just to be sure.

margaale:
@Taloth Finally had the time to dig into the tests. Make a list if you want me to add more please.

margaale:
@Taloth Just a thought, the downloadstation api allows to specify a different username than the one used to log in. Do you think it could be useful?

Taloth:
Shouldn't the username in settings be enough? I'd imagine it'll be the same user in 99.9% of the cases, so lets not needlessly complicate it.

One other question: is a shared folder always `/name` or can users create a shared folder `/group/name`? I mean from filestation perspective, not from downloadstation.

Atm I'm working on cleaning up some stuff, tweaked some stuff, nothing too drastic. But that's something I ran into.

margaale:
In my syno i have shared folder named "video" and some subfolders like, series, anime, movies, misc, etc... The shared folder is video, you cannot only share /video/movie for example. Think of it more link an alias than anything else.
That answers your question?

As of the user. It would be useful for some paranoic user regarding permissions. It could be think of as some sort of label too. But it's not necessary at all...

Taloth:
> That answers your question?

yes, it means my theory about what SharedFolderResolver should do is correct and i'm on the right track to use the same mechanism as RemotePathMapping in order to properly map subfolders.

margaale:
You going to cache the main folder and infer the rest from there?

Taloth:
Even if filestation can resolve subfolders (which based on your reaction it probably does) it would mean we need a FileStation api call for every torrent, albeit cached. By only doing the sharedfolder we'll only need one.

margaale:
Yes, you are right with that, i thought something like that for the serial number, like getting it a the beginning of GetItems and reuse it all over the method..

margaale:
@Taloth like this 1a62338bc90bcc69562d45650d502786d6f19967

Taloth:
With the SerialNumberProvider already having the cache you don't strictly need to keep it local as well since it's still just one api call, but having it local is a bit faster.

Another question: What does the TaskStatus = Finished mean vs Seeding?
One thing I noticed was that the _completed item in Fixture wasn't actually considered Completed, but still Downloading.

margaale:
Wait. Now I understand what you were trying to say before. The shared folder that I'd being passed to shared folder resolver is always the same, it doesn't change from torrent to torrent because the only ones that get there are the ones matching the directory or category in the settings. 
All my torrents returns video/series it doesn't matter if they are single or multiple files in it.

margaale:
Back to the status issue, it's a bug, they should be both completed. If I not mistaken, finished is the final status. It means that is done with the seeding, you can configure to stop seeding after some time elapsed, shared ratio or both.

Taloth:
Ok, then allow me to toss in a couple of requirements and a boatload of commits:
- If "it's a bug", then "it's a missing test", so fix that. Feel free to look at the Fixtures of the other torrent clients.
- OutputPath must point to the exact file to be imported, or the directory containing multiple files for ONE torrent. It cannot, and must not, point to 'series' in general. Sonarr will attempt to import the entire directory or file indiscriminately. So... again, unittests. If the Proxy does not provide the directory/file name, you'll have to infer it. Hopefully we can do that without having to query Syno for the files inside the torrent.
I can't emphasis enough that we need unittests that mock the Proxy instances to return the various scenarios as Synology would.
- What status does it get if it's 'scheduled' for seeding? For example because the the max number of concurrent seeding torrents is already reached? What I mean, is that I need to know if Finished is an intermediate state, or guaranteed to be 'Finished seeding, won't ever be needed again'. This is because the IsReadOnly flag tells Sonarr whether the files can be moved or should be treated readonly and copied/hardlinked instead. It will also allow Sonarr to Remove the torrent automatically after seeding finishes.

Finally, I did a bit of cleanup and a few enhancements, but tried to commit them separately. Started with it yesterday, but didn't split up the commits till today.
You can find it here: https://github.com/Sonarr/Sonarr/tree/synology-diskstation
You can 'pull' those changes by using a fast-forward merge. (basically, rebase on top of that branch). Try avoid using merge-commits and use rebase/fast-forward instead.
Btw. two tests are failing, and they're supposed to till the seeding/isreadonly stuff is dealt with.

If you disagree with any of the changes, lemme know.

margaale:
To your previous comments:

-Finished status is handled correctly.
-What i meant in my previous comment was that diskstation returns the same path that was set when the task was created (or the default in case no path was provided). I'm already infering the full path, but i have to see what do i get when the torrent has a single file in it. Starting to work on this. I can query which files are in a torrent (1e38e988670b0e48dda44b0e8865f121dcff18db) but have a few doubts about what conditions should i check.
-To your question, the status is waiting. Finished is the final status after seeding. Finishing is after downloaded and before seeding.

Taloth:
1) I kinda dislike having the logic in the DownloadStationTorrent, it's a model that the API spits out. It would be better to keep that in DownloadStation itself. An added benefit is that you can then properly set IsReadOnly and make decisions based on other parameters (such as size).

2) It's unfortunate we need to pull the 'file' info from DownloadStation, but so be it. I'll wait for the single-file scenario/test.

3) Sounds we can safely set IsReadOnly to false if the status is Finished.
If 'Waiting' is 'scheduled for seeding', then the 'Waiting' api status should be considered 'Completed', not 'Queued'. 'Queued' implies it hasn't been downloaded yet.

margaale:
Mmmm, i'm not absolutely sure, but there is a chance that waiting it's beign used for both of the statuses, like waiting for download and waiting to seed...

margaale:
Another question, don't you use some kind of playlist in order to run the tests? I've added one to test only DownloadStationFixture.

Taloth:
If Waiting is used for both cases, then checking the totalsize vs downloadedsize might be useful.

As for testing, we don't use playlists. We got separate unittest and integrationtest categories.
In this case I just filter on what i need at a particular time:
image

cybrox:
Been watching this for a while now, even though I don't use Sonarr on my DS anymore but I always felt like Download Station support was a rather important piece missing.
Huge thanks to you, @margaale for finally filling this gap! ❤️

margaale:
@cybrox Thank you for your words, really appreciate them...

margaale:
@Taloth I found a little problem, once the status of torrent changes to finished i cannot get the files in it because the api returns null in that field. Have any ideas how to overcome this? I don't want to consume filestation to ask if exists the folder or file in the item, but i think i will have no choice...,

Taloth:
Before I answer that question, some things I need to know:
What is torrent.Title? Is it based on the 'filename' we supply? What is it for a single file torrent? What about for a multi file download?
I'm thinking about two possible solutions that don't involve fetching the file objects (which isn't reliable, consider Sonarr restarting and losing the cache):
1) Use side-channel info such as torrent.Title to determine the folder.
2) Force Download Station to always store downloads inside subfolders.

How either can be implemented depends heavily on what torrent.Title does. So:
- If I send a new download, .torrent file with a `Movie.mkv` as content, with filename `My.Movie.torrent`, what does the torrent.Title become? (don't guess, we need to know for sure)
- If I send a new download, .torrent file with `Movie.mkv` + `Movie.srt` as content, with filename `My.Movie.torrent`, what does the torrent.Title become? And in what subfolder (if any) will DownloadStation put the two files?

margaale:
Had the time to run some tests. 

a) While queued title is My.Movie.torrent, after starting is Movie.mkv
b) While queued title is My.Movie.torrent, after starting is the directory where both files will be left after finishing.

Taloth:
in case b, is that a new subdirectory created by DownloadStation? or the root directory? (are the files mixed with other downloads?)

margaale:
It's created for each torrent

Taloth:
In that case, OutputPath + Title should be enough. In case of a single file it would be `/movies/share/Movie.mkv`. In case of MultiFile it would be `/movies/share/My.Movie`.
Doesn't that mean you can do without the 'file' info on the DownloadStation api? OutputPath and Title should be enough.

margaale:
Absolutely, it was that way before you made me doubt about it and It's been like that again since Wednesday I think. Just have to clean up proxy now....

Taloth:
I may have been unclear before. I tried to derive step-by-step how DownloadStation behaved so we can get a solution that works 100% of the time.

My concern posted [last week](https://github.com/Sonarr/Sonarr/pull/1620#issuecomment-277837785) was that, with the implementation at the time, OutputPath was assigned the value that the Syno api supplied. Title wasn't included there in any way. It was something that needed to be solved, I didn't say how.

Either way, it's a relieve to know we can make it work properly in all cases.

margaale:
Not trying to argue here. But yes it was, if you inspect your branch you can see that it's what getoutputpath returns... But I understand your concern do don't worry about it. 
Changing subject, Do you think of any other test that should be added?

Taloth:
> Not trying to argue here. But yes it was,

You're totally correct, I overlooked that, probably because it's appended after the Remap. So my initial concern was unwarranted. My apologies for the extra work that caused. I know it's fine, had to be said though.

The tests are pretty complete now, with the most important behavioral stuff covered. I'd keep the _singleFile & _multipleFIles related tests, Just without the Files. Obviously we won't need those file details anymore.

https://github.com/Sonarr/Sonarr/pull/1620/files#diff-045540c1f6509c59d0838edfc37ada0aR557 needs to be updated to physicalPath + Title, then we're good to go.

(Also do a search for `ouputhPath`->`outputPath` ;) )

If you can fix everything up + squash and rebase. Then i'll get it merged to develop today.

margaale:
Guess i got it right with 554e14370a826e518c01e34b567094ee95bb6acf

Taloth:
Looks like you rebased on another branch instead of develop, but no matter, I fixed it in https://github.com/Sonarr/Sonarr/commit/82a99b7f806fbfb756ae5fe1296fb45c2047a9cb.

The build agent is building and running tests in develop now, so the next dev release should have it.

lpicanco:
Anyone notice a high cpu usage after setting the synology disk station as download client?

margaale:
How many tasks were added? Try lowering your concurrent torrents in download station settings

lpicanco:
There are no torrents in download queue. When I enable the DS client, the mono process go 60% of the cpu. I think there is something with the connection between Sonnar and DS.

margaale:
What about finished tasks? If your download station list is empty then it is not related to the client at all

cristiangauma:
Hi,

could you finally try to add magents to Download Station?. They are not working as far as I could test. They are in "waiting" status on Download Station.

Also, I think the the title on "File Name" is not being parsed correctly. They have HTML character codes instead of the correct charactes. Here you have an [screenshot](https://i.imgur.com/544gM7d.png) so you will be able to see what I'm referring to.

lpicanco:
I had the same problem @cristiangauma. In my case, i created a new user in synology only to use with Sonarr. I had to login  with the new user and choose the default download folder of download station using the app.

cristiangauma:
@lpicanco Yep, it worked this way. Thanks!

Issue #749, opened 2015-08-18
Localization of the UI


markus101:
This would be a pretty massive undertaking, not only would we need to add support for it we'd need to get proper translations. I'm not entirely sure how reliable user contributed translations would be and we're not in a position to go for paid translations.

Mirx:
Plex does it through getlocalization.com. 
As a former member of a custom subtitling team back in the day, I can surely say that correct contextual translation is hard and takes a special touch.

F.e. in Plex, account was translated to "rekening", which is actually a correct translation, but in relation to a checking account, not to a user account. 

So I share your concerns that bad contextual translations can lead to misinterpretation and confusion. That's why I always use the native GUI language (if I understand that).

I wouldn't worry to much about paid versus free, as the paid firms are mostly production houses with low quality control, which in my believe leads to more contextual errors, then a translation made by someone who actually cares about the end result.

If you do add localization, I'll pitch in for the Dutch translation.

kayone:
I agree with @Mirx concerns, one thing that is worst than an english only UI is a badly translated or partially translated UI.

Maybe once we are more popular and could source reliable translation from users _that actually know and understand_ the product it would be a more viable option. Keep in mind sonarr has a decent amount of domain specific terms that might not be easily translated.

Thirrian:
Just my 2 cents: I feel translations are the spawn of the devil. Please don't.
There's plenty of far more interesting (and beneficial for all users) stuff in your todo list in my opinion. If only I could code...

Also think about the support nightmare. I mean, *really* think about it. Yeah. That's right. No sleep for you tonight. :laughing:

(Note: I'm not saying this as an English-only speaker. I'm learning my 4th language at the moment, not counting my native tongue. If you really must: I guess I'll volunteer as I'd hate to see such a fine application lose value by crappy translations.)

Blackspirits:
Add Portuguese language, I translate

markus101:
This will be done in v4.

PR #1623, opened 2017-01-16
Add French language and French broadcast calendar

kestouf:
Hello
- Add the French language to sonarr. I am willing to do the translation (I am French)
- Add the possibility of having the French broadcast calendar
Thx
Best regards


markus101:
Not sure why this is a PR, it sounds like it should be an issue.

There is already an issue for Translations, though it's not a priority and has it's downsides. https://github.com/Sonarr/Sonarr/issues/749
A calendar from another source isn't possible since all the information comes from TheTVDB and it only has the original broadcast info.

PR #1624, opened 2017-01-16
Option to convert ical feed items to all-day events

sanderploegsma:
#### Database Migration
NO

#### Description
As a European watching mostly American shows, the iCal feed events all take place around midnight, which (in my case) is collapsed by default. Since I don't really care for the exact air time of each episode (it takes a while before they are available anyway), I would prefer the ability of having all-day events for each episode. 

The following toggle enables this feature.
image

* None. This was just a random idea that came up. Since it's not that intrusive, I thought I'd just propose it anyway.

### DIFF
--- a/src/NzbDrone.Api/Calendar/CalendarFeedModule.cs
+++ b/src/NzbDrone.Api/Calendar/CalendarFeedModule.cs
@@ -37,6 +37,7 @@ namespace NzbDrone.Api.Calendar
             var end = DateTime.Today.AddDays(futureDays);
             var unmonitored = false;
             var premiersOnly = false;
+            var asAllDay = false;
             var tags = new List<int>();
 
             // TODO: Remove start/end parameters in v3, they don't work well for iCal
@@ -46,6 +47,7 @@ namespace NzbDrone.Api.Calendar
             var queryFutureDays = Request.Query.FutureDays;
             var queryUnmonitored = Request.Query.Unmonitored;
             var queryPremiersOnly = Request.Query.PremiersOnly;
+            var queryAsAllDay = Request.Query.AsAllDay;
             var queryTags = Request.Query.Tags;
 
             if (queryStart.HasValue) start = DateTime.Parse(queryStart.Value);
@@ -71,6 +73,11 @@ namespace NzbDrone.Api.Calendar
             if (queryPremiersOnly.HasValue)
             {
                 premiersOnly = bool.Parse(queryPremiersOnly.Value);
+            }
+
+            if (queryAsAllDay.HasValue)
+            {
+                asAllDay = bool.Parse(queryAsAllDay.Value);
             }
 
             if (queryTags.HasValue)
@@ -102,10 +109,18 @@ namespace NzbDrone.Api.Calendar
                 var occurrence = calendar.Create<Event>();
                 occurrence.Uid = "NzbDrone_episode_" + episode.Id;
                 occurrence.Status = episode.HasFile ? EventStatus.Confirmed : EventStatus.Tentative;
-                occurrence.Start = new CalDateTime(episode.AirDateUtc.Value) { HasTime = true };
-                occurrence.End = new CalDateTime(episode.AirDateUtc.Value.AddMinutes(episode.Series.Runtime)) { HasTime = true };
                 occurrence.Description = episode.Overview;
                 occurrence.Categories = new List<string>() { episode.Series.Network };
+
+                if (asAllDay)
+                {
+                    occurrence.Start = new CalDateTime(episode.AirDateUtc.Value) { HasTime = false };
+                }
+                else
+                {
+                    occurrence.Start = new CalDateTime(episode.AirDateUtc.Value) { HasTime = true };
+                    occurrence.End = new CalDateTime(episode.AirDateUtc.Value.AddMinutes(episode.Series.Runtime)) { HasTime = true };
+                }
 
                 switch (episode.Series.SeriesType)
                 {
--- a/src/UI/Calendar/CalendarFeedView.js
+++ b/src/UI/Calendar/CalendarFeedView.js
@@ -9,6 +9,7 @@ module.exports = Marionette.Layout.extend({
     ui : {
         includeUnmonitored : '.x-includeUnmonitored',
         premiersOnly       : '.x-premiersOnly',
+        asAllDay           : '.x-asAllDay',
         tags               : '.x-tags',
         icalUrl            : '.x-ical-url',
         icalCopy           : '.x-ical-copy',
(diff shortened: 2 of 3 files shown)


Taloth:
One concern I have over this is that it feels like a 1% feature that just adds another option in terms of complexity. Another is that it's basically intended as workaround for an issue with your device / calendar app, rather than Sonarr, that doesn't sit right with me.
But I'll let @markus101 chime in first.

markus101:
I have the same concern.

Why are the midnight times automatically collapsed?
Not within working hours?
Are you able to define that on a calendar-by-calendar basis?

@Taloth how would you feel about the option existing in the API, but not the UI?

sanderploegsma:
Yeah it's due to working hours and is indeed an option of most calendar apps, so no issue with the Sonarr functionality. I just figured that I might not be the only one though. 

Anyway, I can see how an extra UI component might be too much seeing as this might not matter to most of the users. API only functionality would be a viable solution in my case.

Taloth:
We had a vote with the other devs. We'll accept the change, because there might be a larger percentage of users that don't care about specific air times, and just want the date, and also because calendar apps don't usually render overlapping events (primetime shows) nicely.
That it happens to fix your specific issue too is just a totally 'unintended side-effect'. :wink:

However it'll need to be cleaned up. we'll leave some comments.

src/NzbDrone.Api/Calendar/CalendarFeedModule.cs
```diff
@@ -37,6 +38,7 @@ private Response GetCalendarFeed()
             var end = DateTime.Today.AddDays(futureDays);
             var unmonitored = false;
             var premiersOnly = false;
+			var allDay = false;
```
Taloth: Replace tabs with the appropriate number of spaces.
Advice: Install EditorConfig extension in your editor. It'll automatically adjust the editor settings based on `.editorconfig` files in the sources.

src/NzbDrone.Api/Calendar/CalendarFeedModule.cs
```diff
@@ -11,6 +11,7 @@
 using Nancy.Responses;
 using NzbDrone.Core.Tags;
 using NzbDrone.Common.Extensions;
+using System.Diagnostics;
```
Taloth: Nitpick: We sort System.* first, it's vanilla Visual Studio setting btw.

src/NzbDrone.Api/Calendar/CalendarFeedModule.cs
```diff
@@ -46,6 +48,7 @@ private Response GetCalendarFeed()
             var queryFutureDays = Request.Query.FutureDays;
             var queryUnmonitored = Request.Query.Unmonitored;
             var queryPremiersOnly = Request.Query.PremiersOnly;
+			var queryAllDay = Request.Query.AllDay;
```
Taloth: I'd prefer 'AsAllDay' since it's not a filter option.

src/NzbDrone.Api/Calendar/CalendarFeedModule.cs
```diff
@@ -101,12 +109,23 @@ private Response GetCalendarFeed()
...
+				occurrence.Status = episode.HasFile ? EventStatus.Confirmed : EventStatus.Tentative;
+				occurrence.Description = episode.Overview;
                 occurrence.Categories = new List<string>() { episode.Series.Network };
 
+				if (allDay)
+				{
+					DateTime startDate = episode.AirDateUtc.Value;
+					DateTime endDate = episode.AirDateUtc.Value.AddDays(1);
+					occurrence.Start = new CalDateTime(startDate.Year, startDate.Month, startDate.Day) { HasTime = false };
+					occurrence.End = new CalDateTime(endDate.Year, endDate.Month, endDate.Day) { HasTime = false };
```
Taloth: Based on webcal spec you can just leave `.End` unset if you have `.Start` without time, can you try that out? Saves you the arithmetic, and other calendar apps may freak out with the end date turning up as the next day... or the next year if it happens to air on December 31st.

src/NzbDrone.Api/Calendar/CalendarFeedModule.cs
```diff
@@ -120,6 +139,7 @@ private Response GetCalendarFeed()
 
             var serializer = (IStringSerializer) new SerializerFactory().Build(calendar.GetType(), new SerializationContext());
             var icalendar = serializer.SerializeToString(calendar);
+			Debug.WriteLine(icalendar);
```
Taloth: Stray debug code. -1 dory points for you. :smile:

src/UI/Calendar/CalendarFeedView.js
```diff
@@ -9,6 +9,7 @@ module.exports = Marionette.Layout.extend({
     ui : {
         includeUnmonitored : '.x-includeUnmonitored',
         premiersOnly       : '.x-premiersOnly',
+        allDay             : '.x-allDay',
```
Taloth: Change all of these to 'asAllDay' too, for consistency.

src/UI/Calendar/CalendarFeedViewTemplate.hbs
```diff
@@ -42,6 +42,24 @@
                 </div>
             </div>
             <div class="form-group">
+                <label class="col-sm-3 control-label">Convert To All-day Events</label>
```
Taloth: `Show as All-Day events` should fit the horizontal space better.

sanderploegsma:
Great! Yeah, leave some comments and I'll get to it asap. I noticed already I forgot to remove some debug stuff before committing. (also, Visual Studio did some weird stuff to the indenting...)

src/UI/Calendar/CalendarFeedView.js
```diff
@@ -18,6 +19,7 @@ module.exports = Marionette.Layout.extend({
     events : {
         'click .x-includeUnmonitored' : '_updateUrl',
         'click .x-premiersOnly'       : '_updateUrl',
+        'click .x-allDay'             : '_updateUrl',
         'itemAdded .x-tags'           : '_updateUrl',
         'itemRemoved .x-tags'         : '_updateUrl'
     },
```
Taloth: While you're at it can you replace 'NzbDrone' with 'Sonarr' a few lines down, the backend already supports it and evidently we forgot to replace it in the UI.

sanderploegsma:
Ok thanks for the feedback, I processed everything! You were right by the way, the end time is not needed for all-day events.

Taloth:
Look good. Merged in.

Tnx for the contribution.

sanderploegsma:
No problem, thanks for merging 👍

PR #1635, opened 2017-01-21
Remove torrent from client when they're moved

beschoenen:
#### Database Migration
NO

#### Description
This PR will make it possible to remove seeding torrents from clients when the files are moved/removed from the downloads folder by Sonarr.

#### Todos
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/src/NzbDrone.Core/Download/DownloadEventHub.cs
+++ b/src/NzbDrone.Core/Download/DownloadEventHub.cs
@@ -37,11 +37,12 @@ namespace NzbDrone.Core.Download
         {
             if (!_configService.RemoveCompletedDownloads ||
                 message.TrackedDownload.DownloadItem.Removed ||
-                message.TrackedDownload.DownloadItem.IsReadOnly ||
                 message.TrackedDownload.DownloadItem.Status == DownloadItemStatus.Downloading)
             {
                 return;
             }
+
+            if (message.TrackedDownload.DownloadItem.IsReadOnly && _configService.CopyUsingHardlinks) return;
 
             RemoveFromDownloadClient(message.TrackedDownload);
         }


Taloth:
But... Sonarr doesn't move/remove files for seeding torrents, it only copies or hardlinks them.

beschoenen:
OK, forgive me, I started using Sonarr today and when I noticed that torrents weren't getting removed I started debugging and found this.

Does the "remove from client" settings also ask clients to remove the files? Cause I tested this and the original files were removed.

What do I have to do to always remove the torrent from the client, even when it's seeding?

Taloth:
> Does the "remove from client" settings also ask clients to remove the files? Cause I tested this and the original files were removed.

Yes, at least for the clients that [support it](https://github.com/Sonarr/Sonarr/blob/0bdc1370937a35377925bb2a07136ab9b8533acf/src/NzbDrone.Core/Download/DownloadEventHub.cs#L68).

> What do I have to do to always remove the torrent from the client, even when it's seeding?

Configure your client to stop seeding is the only way. Usually this is done by specifying a desired ratio.
If Sonarr sees that the download finished seeding, it'll ask the download client to remove the torrent and files. This feature depends on the download client, not all have a 'Finished seeding' state. Notably, rtorrent doesn't support it.

beschoenen:
What would think about a new setting? 
Can only be toggled when "Remove" is enabled.
"Always"
Description: "Always from the torrent from the client, even when it's still seeding"

Taloth:
Can I ask why you won't consider setting your download client seeding ratio appropriately?

Having Sonarr force the removal is not the right approach. For one, it's way too easy to get HnRs on private trackers that way. Many users won't realize the consequences of turning it on. It also raises the question on the sequence of events for imports, since it would have to stop the torrent first to be able to move the files.
So, no, I don't see that option added.

PR #1644, opened 2017-01-27
Change default Newznab/Torznab indexer categories to TV

kaso17:
#### Database Migration
NO

#### Description
Over at the Jackett project we occasionally get issues due to an incorrect category configuration. The root cause is the fact that some trackers don't have TV SD/HD categories (just a TV cat). In this case the releases won't be found with the current default cat settings.
Due to that I would like to propose this change of the default categories.

### DIFF
--- a/src/NzbDrone.Core/Indexers/Newznab/NewznabSettings.cs
+++ b/src/NzbDrone.Core/Indexers/Newznab/NewznabSettings.cs
@@ -60,7 +60,7 @@ namespace NzbDrone.Core.Indexers.Newznab
 
         public NewznabSettings()
         {
-            Categories = new[] { 5030, 5040 };
+            Categories = new[] { 5000 };
             AnimeCategories = Enumerable.Empty<int>();
         }
 


markus101:
Changing it to `5000` has the negative effect of pulling in foreign and other categories that are not needed by most users at least in the usenet world. If this is affecting Jackett only we should address it for Torznab only and not for Newznab as well or deal with it in a better way.

How does Jackett translate the categories back to the indexers? Can it default to all TV categories if the selected categories aren't found? I'd expect indexers that provide a Torznab API to use the categories properly so even setting Torznab to 5000 may have a negative impact.

kaso17:
Jackett maps the Newznab cats back to the tracker specific cats.
Often the Newznab cats don't fit very well to the tracker specific caps, so they are mapped to the closest parent category.

Would you accept a bigger patch making newznab/torznab tracker addition a two step process?
In the second step Sonarr would query the indexer caps and show the available categories so the user can select them.

markus101:
Not in the current UI state, perhaps something we could look at after v3 lands (whenever that is). For now I think we could enforce this for Torznab only and set the default for current Torznab templates to use the the current default qualities. That way existing templates work as-is and anything custom uses the default of `5000`.

Taloth:
The current UI would be able to do it too. I implemented the necessary backend architecture in the rarbg captcha commit. The biggest hurdle would be to visualize the category hierarchy (yes, it's a hierarchy).
I thought we had a Github issue about that, but I can't find it.

Please be aware that newznab (and by extension torznab), is designed to be able to query multiple categories at once. 5000 already contains multiple sub-categories.
It entirely depends on the tracker website on whether Jackett can use a single query to search for multiple categories at once and thus is the responsibility of Jackett to handle if the tracker can't (for example by using multiple trackers queries to service a single torznab api call).

There doesn't have to be a 1 to 1 mapping between torznab and tracker categories. For example [this](https://github.com/Sonarr/Sonarr/blob/816cf608fcef07e375aa750a863d0a7365108981/src/NzbDrone.Core.Test/Files/Indexers/Torznab/torznab_hdaccess_net.xml#L31-L34), that single release exists in two tracker categories (9 and 36) and is assigned to the 5040 torznab category (which is a child of 5000).
hdaccess had torznab support and simply translated a query for 5040 (or 5000) by a database query WHERE category IN (9, 36). hd4free uses the same logic.

Taloth:
If 5030,5040 doesn't exist, then we could just test that by adding another validation [here](https://github.com/Sonarr/Sonarr/blob/f16dd069b5f40fadf64e5d1b334e8c7dbda142de/src/NzbDrone.Core/Indexers/Torznab/Torznab.cs#L83). Checking whether the provided categories actually exist and report an error if it doesn't (it's an advanced field though and not always visible, so be careful).
Jackett can also do that by returning an error on a query that has an unsupported category.

So there are several solutions available, each with an increased level of complexity. But simply changing the default category isn't acceptable.

- Have Jackett simply returns all tv items if queried by 5030,5040, for that specific tracker of course.
- Have Jackett throw an error if an unknown category is queried.
- Have Sonarr check caps to reject the settings if an unknown category is selected.
- Add UI element that uses the [callback api (see rarbg captcha)](https://github.com/Sonarr/Sonarr/blob/f16dd069b5f40fadf64e5d1b334e8c7dbda142de/src/NzbDrone.Core/Indexers/Rarbg/Rarbg.cs#L39) to fetch the caps, and show a proper category selection UI.

Take your pick.

Issue #1671, opened 2017-02-12
Download Station


foXaCe:
hello, send torrent do not work.

the log

Request Failed. POST /api/release: Failed to add torrent task to Download Station

NzbDrone.Core.Download.Clients.DownloadClientException: Failed to add torrent task to Download Station
  at NzbDrone.Core.Download.Clients.DownloadStation.DownloadStation.AddFromTorrentFile (NzbDrone.Core.Parser.Model.RemoteEpisode remoteEpisode, System.String hash, System.String filename, System.Byte[] fileContent) [0x000a0] in M:\BuildAgent\work\b69c1fe19bfc2c38\src\NzbDrone.Core\Download\Clients\DownloadStation\DownloadStation.cs:179 
  at NzbDrone.Core.Download.TorrentClientBase`1[TSettings].DownloadFromWebUrl (NzbDrone.Core.Parser.Model.RemoteEpisode remoteEpisode, System.String torrentUrl) [0x001aa] in M:\BuildAgent\work\b69c1fe19bfc2c38\src\NzbDrone.Core\Download\TorrentClientBase.cs:181 
  at NzbDrone.Core.Download.TorrentClientBase`1[TSettings].Download (NzbDrone.Core.Parser.Model.RemoteEpisode remoteEpisode) [0x00148] in M:\BuildAgent\work\b69c1fe19bfc2c38\src\NzbDrone.Core\Download\TorrentClientBase.cs:117 
  at NzbDrone.Core.Download.DownloadService.DownloadReport (NzbDrone.Core.Parser.Model.RemoteEpisode remoteEpisode) [0x00171] in M:\BuildAgent\work\b69c1fe19bfc2c38\src\NzbDrone.Core\Download\DownloadService.cs:66 
  at NzbDrone.Api.Indexers.ReleaseModule.DownloadRelease (NzbDrone.Api.Indexers.ReleaseResource release) [0x0002c] in M:\BuildAgent\work\b69c1fe19bfc2c38\src\NzbDrone.Api\Indexers\ReleaseModule.cs:67 
  at NzbDrone.Api.Indexers.ReleaseModule.<.ctor>b__7_0 (System.Object x) [0x00000] in M:\BuildAgent\work\b69c1fe19bfc2c38\src\NzbDrone.Api\Indexers\ReleaseModule.cs:46 
[...]

PR #1672, opened 2017-02-12
Throw exception with error message return by diskstation

margaale:
#### Database Migration
NO

#### Description
Throws a DownloadClientException with the message return by the Diskstation when the request was unsuccessful

#1671

### DIFF
--- a/src/NzbDrone.Core/Download/Clients/DownloadStation/DownloadStation.cs
+++ b/src/NzbDrone.Core/Download/Clients/DownloadStation/DownloadStation.cs
@@ -146,7 +146,7 @@ namespace NzbDrone.Core.Download.Clients.DownloadStation
 
             _proxy.AddTorrentFromUrl(magnetLink, GetDownloadDirectory(), Settings);
 
-            var item = _proxy.GetTorrents(Settings).Where(t => t.Additional.Detail["uri"] == magnetLink).SingleOrDefault();
+            var item = _proxy.GetTorrents(Settings).SingleOrDefault(t => t.Additional.Detail["uri"] == magnetLink);
 
             if (item != null)
             {
--- a/src/NzbDrone.Core/Download/Clients/DownloadStation/DownloadStation.cs
+++ b/src/NzbDrone.Core/Download/Clients/DownloadStation/DownloadStation.cs
@@ -117,13 +117,16 @@ namespace NzbDrone.Core.Download.Clients.DownloadStation
 
         public override void RemoveItem(string downloadId, bool deleteData)
         {
-            if (_proxy.RemoveTorrent(ParseDownloadId(downloadId), deleteData, Settings))
+            try
             {
+                _proxy.RemoveTorrent(ParseDownloadId(downloadId), deleteData, Settings);
                 _logger.Debug("{0} removed correctly", downloadId);
                 return;
             }
-
-            _logger.Error("Failed to remove {0}", downloadId);
+            catch (DownloadClientException e)
+            {
+                _logger.Error(e);
+            }
         }
 
         protected OsPath GetOutputPath(OsPath outputPath, DownloadStationTorrent torrent, string serialNumber)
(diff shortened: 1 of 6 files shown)


Taloth:
Commit itself is OK, but it bypasses all the logic in:
https://github.com/Sonarr/Sonarr/blob/develop/src/NzbDrone.Core/Download/Clients/DownloadStation/Proxies/FileStationProxy.cs#L57 
https://github.com/Sonarr/Sonarr/blob/develop/src/NzbDrone.Core/Download/Clients/DownloadStation/Proxies/DSMInfoProxy.cs#L36

Wouldn't it be better to replace those and add similar constructs in AddTorrentFromUrl and AddTorrentFromData.
Using the pattern `"Failed to add torrent {0}: {1}", filename, reason`.
So, throwing an exception that describes what Sonarr wanted to do and what error it got.

margaale:
Sorry, i don't fully understand what you are asking me to do.

Replace those lines with what?
And you mean AddTorrentFromUrl and AddTorrentFromData in the proxy too?

margaale:
Or you mean catching the newly added exception? Or don't throw it at all?

Taloth:
On develop we throw an exception in FileStation and DSMInfo if something goes wrong `"Failed to get Download Station serial number"`. That should be `"Failed to get Download Station serial number: {0}", reason"` where reason is the error message syno gives.   
The same should be done in the other methods.

AddTorrentFromUrl and AddTorrentFromData don't throw an exception atm, but they probably should.

margaale:
But now ProcessRequest throws the exception first. Should i try/catch that exception in the proxies an throw a new one like the one you describe?

Taloth:
Don't implement the exception in ProcessRequest, instead do it in AddTorrentFromUrl and AddTorrentFromData.

Try-Catch works too, it just is a bit redundant in this case. I just want to make sure the user gets an error that both includes what Sonarr wants to do, and what went wrong.

margaale:
What do you think of it now, different approach...

Taloth:
It looks like that will work. Little different than we usually do.

@markus101 Your opinion?

I've gotta go for a few hours. I'll see about merging it in afterwards.

src/NzbDrone.Core/Download/Clients/DownloadStation/DownloadStation.cs
```diff
@@ -141,41 +144,39 @@ protected override string AddFromMagnetLink(RemoteEpisode remoteEpisode, string
...
         {
             var hashedSerialNumber = _serialNumberProvider.GetSerialNumber(Settings);
 
-            if (_proxy.AddTorrentFromData(fileContent, filename, GetDownloadDirectory(), Settings))
-            {
-                var items = _proxy.GetTorrents(Settings).Where(t => t.Additional.Detail["uri"] == Path.GetFileNameWithoutExtension(filename));
+            _proxy.AddTorrentFromData(fileContent, filename, GetDownloadDirectory(), Settings);
 
-                var item = items.SingleOrDefault();
+            var items = _proxy.GetTorrents(Settings).Where(t => t.Additional.Detail["uri"] == Path.GetFileNameWithoutExtension(filename));
```
markus101: You can use `SingleOrDefault` instead of `Where` here.

src/NzbDrone.Core/Download/Clients/DownloadStation/Proxies/DiskStationProxyBase.cs
```diff
@@ -62,14 +64,22 @@ public DiskStationProxyBase(IHttpClient httpClient, Logger logger)
...
                 {
-                    return responseContent;
+                    if (responseContent.Error.SessionError)
+                    {
+                        _authenticated = false;
+                        return ProcessRequest<T>(api, arguments, settings, operation, method, retries++);
+                    }
+
+                    var msg = $"Failed to {operation}. Reason: {responseContent.Error.GetMessage(api)}";
+                    _logger.Debug(msg);
```
markus101: Message looks error-like but we only log a debug message, is that correct?

src/NzbDrone.Core/Download/Clients/DownloadStation/Proxies/DownloadStationProxy.cs
```diff
@@ -76,14 +72,17 @@ public IEnumerable<DownloadStationTorrent> GetTorrents(DownloadStationSettings s
...
             {
+                var response = ProcessRequest<DownloadStationTaskInfoResponse>(DiskStationApi.DownloadStationTask, arguments, settings, "get torrents");
+
                 return response.Data.Tasks.Where(t => t.Type == DownloadStationTaskType.BT);
             }
-
-            return new List<DownloadStationTorrent>();
+            catch (DownloadClientException e)
+            {
+                _logger.Debug(e);
```
markus101: Debug for an exception?

markus101:
The approach seems fine, a bit different, but nothing earth shattering.

margaale:
Done

Issue #1676, opened 2017-02-12
Download Station and Completed Download Handling


obouallou:
Hello,

First of all thanks for adding support for DS in Sonarr !

I've started to try it since it was merged in develop branch, and got some questions related to completed download handling.
Here's the description of the feature from the wiki : 

If the "Completed Download Handling - Remove" option is enabled in Sonarr's settings, Sonarr will delete the original file and torrent from your client, but only if the client reports that seeding is complete and torrent is stopped.

So if I understand correctly, once seeding is done, Sonarr is supposed to remove the torrent from Synology client (which it does) and delete the original files from the download directory (which it doesn't)

Is it working as inteded or do you think it should delete files ?

margaale:
Hi, a few questions until some of the devs see this. How is your setup? Do you have sonarr runing in the diskstation? Any remote mapping? Permissions of your download directory? Laslty, any do you see any error in logs or any other info that could be useful?

obouallou:
Sonarr is running on a different machine but the video folder is mounted with R/W permissions.
torrents are downloaded to /video/downloaded/sonarr and then processed to /video/series.

There are no errors in logs  (removed data & filename : 

```
17-2-12 20:34:38.4|Trace|CommandQueueManager|Publishing CheckForFinishedDownload
17-2-12 20:34:38.4|Trace|CommandQueueManager|Checking if command is queued or started: CheckForFinishedDownload
17-2-12 20:34:38.4|Trace|CommandQueueManager|Inserting new command: CheckForFinishedDownload
17-2-12 20:34:38.4|Trace|CommandExecutor|CheckForFinishedDownloadCommand -> DownloadMonitoringService
17-2-12 20:34:38.4|Trace|CommandQueueManager|Marking command as started: CheckForFinishedDownload
17-2-12 20:34:38.7|Trace|HttpClient|Req: [GET] http://192.168.138.1:5000/webapi/DownloadStation/task.cgi?api=(removed)&version=1&method=list&additional=detail%2Ctransfer
17-2-12 20:34:38.7|Trace|ConfigService|Using default config value for 'proxyenabled' defaultValue:'False'
17-2-12 20:34:38.9|Trace|HttpClient|Res: [GET] http://192.168.138.1:5000/webapi/DownloadStation/task.cgi?api=(removed)&version=1&method=list&additional=detail%2Ctransfer: 200.OK (263 ms)
[...]

PR #1678, opened 2017-02-13
Delete data manually after finished seeding

margaale:
#### Database Migration
NO

#### Description
DS doesn't support deleting files when removing the torrent after it has finished seeding.

#1676

### DIFF
--- a/src/NzbDrone.Core/Download/Clients/DownloadStation/DownloadStation.cs
+++ b/src/NzbDrone.Core/Download/Clients/DownloadStation/DownloadStation.cs
@@ -119,7 +119,11 @@ namespace NzbDrone.Core.Download.Clients.DownloadStation
         {
             try
             {
-                _proxy.RemoveTorrent(ParseDownloadId(downloadId), deleteData, Settings);
+                if (deleteData)
+                {
+                    DeleteItemData(downloadId);
+                }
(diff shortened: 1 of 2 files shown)


margaale:
@Taloth Is there a way to determine if DeleteItemData has actually deleted the data?

Taloth:
No, but that's trivial to add to that method.
But why would you need it? Don't you have to delete the torrent itself from the Syno itself regardless of that?

margaale:
Yes i have and that it's what already did. Just asking cause if i delete the torrent, RemoveItem will never be called again and if files are still there they will never get deleted. Or at least that is what i understand from inspecting the code....

Taloth:
You're correct, it would never get deleted if it went wrong at one point. But how often does that happen? If it failed to remove, it likely would've failed to import in the first place.
You could try handle things differently based on a `DeleteItemData` return value, such as not remove the torrent, but you'll end up having it retry every minute. That might be equally undesirable.
So my question is: Do you really need it? Or will you accept the risk that in some edge-case, you'll have leftovers in the download dir.

margaale:
As a developer it bothers me that is not beiign atomic, as a user i guess i could live with that. So, let's keep it simple now and leave it as it is. If we get enough complaints from the users we can change it later.

PR #1679, opened 2017-02-13
Usenet DownloadStation support 

margaale:
#### Database Migration
NO

#### Description
Now that download station is working i'm planning to add usenet support to it. I'm requesting very early in order to disscuss some issues first.

#### Todos
- [x] Tests
- [x] Coding

### DIFF
--- a/src/NzbDrone.Core/Download/Clients/DownloadStation/Proxies/DownloadStationProxy.cs
+++ b/src/NzbDrone.Core/Download/Clients/DownloadStation/Proxies/DownloadStationProxy.cs
@@ -12,7 +12,7 @@ namespace NzbDrone.Core.Download.Clients.DownloadStation.Proxies
     {
         IEnumerable<DownloadStationTask> GetTasks(DownloadStationTaskType type, DownloadStationSettings settings);
         Dictionary<string, object> GetConfig(DownloadStationSettings settings);
-        void RemoveTorrent(string downloadId, DownloadStationSettings settings);
+        void RemoveTask(string downloadId, DownloadStationSettings settings);
         void AddTaskFromUrl(string url, string downloadDirectory, DownloadStationSettings settings);
         void AddTaskFromData(byte[] data, string filename, string downloadDirectory, DownloadStationSettings settings);
         IEnumerable<int> GetApiVersion(DownloadStationSettings settings);
@@ -99,7 +99,7 @@ namespace NzbDrone.Core.Download.Clients.DownloadStation.Proxies
             return response.Data;
         }
 
-        public void RemoveTorrent(string downloadId, DownloadStationSettings settings)
+        public void RemoveTask(string downloadId, DownloadStationSettings settings)
         {
             var arguments = new Dictionary<string, object>
             {
--- a/src/NzbDrone.Core/Download/Clients/DownloadStation/Proxies/DownloadStationProxy.cs
+++ b/src/NzbDrone.Core/Download/Clients/DownloadStation/Proxies/DownloadStationProxy.cs
@@ -10,11 +10,11 @@ namespace NzbDrone.Core.Download.Clients.DownloadStation.Proxies
 {
     public interface IDownloadStationProxy
     {
-        IEnumerable<DownloadStationTorrent> GetTorrents(DownloadStationSettings settings);
+        IEnumerable<DownloadStationTask> GetTasks(DownloadStationTaskType type, DownloadStationSettings settings);
         Dictionary<string, object> GetConfig(DownloadStationSettings settings);
         void RemoveTorrent(string downloadId, DownloadStationSettings settings);
-        void AddTorrentFromUrl(string url, string downloadDirectory, DownloadStationSettings settings);
-        void AddTorrentFromData(byte[] torrentData, string filename, string downloadDirectory, DownloadStationSettings settings);
+        void AddTaskFromUrl(string url, string downloadDirectory, DownloadStationSettings settings);
+        void AddTaskFromData(byte[] data, string filename, string downloadDirectory, DownloadStationSettings settings);
         IEnumerable<int> GetApiVersion(DownloadStationSettings settings);
     }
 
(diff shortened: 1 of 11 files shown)


margaale:
I've been trying to figure how to add this out. I'll have a lot of duplicate code between torrent and usenet classes. I was thinking to add a helper class  or some extensions to mitigate that, for example createid, parseid, getstatus, etc. What approach would you use, my first attempt would be to create a base class but that cannot be done here given the actual hierarchy...

Taloth:
The renames for DownloadStationTask are good, and the sharing of the Proxy between the two as well.
I'm not fond of the Extensions, they're not horrible, but I try to avoid extension methods like that.
I talked to markus, we're ok with some code duplication between DownloadStation & UsenetDownloadStation.
If there's complex logic that is shared between the two, create a separate service+interface, like we do with the serialnumberprovider. If you insist for the DownloadId, then I guess we'll need a IDownloadIdGenerator or something. But it isn't too bad to have the code duplicated. They are separate clients. And the 'status' behavior could well be different between the two, now or in the future.

Btw. I'm pushing a rename to develop. with db migration to rename 'DownloadStation' to 'TorrentDownloadStation'.

margaale:
Thanks for that commit, i wanna rename it like that. Reverted all extensions methods. If you both are ok with duplicated code i'ml leaving it as it is right now... So, tests, any particular test should be taken into account? 

A note on this, i've got no idea about nzb, i've never used it actually. Tasks got queued and finished but not the episode....

Taloth:
tests: nothing special, IsReadOnly should be false. since there's no seeding. Nzbs probably always end up in a subfolder, but I doubt that changes the implementation on our end. I'd just copy the Fixture and clean it up. The logic is largely the same, afaik.

I don't know what you mean with "but not the episode"

margaale:
I mean that i ended up with a bunch of folders named like the nzb file but no files inside them.

Taloth:
it is possible sonarr imports them quickly? by moving em :)

margaale:
jajajaja, i don't think so. They were quite old episodes, like the walking dead season 3. And by the way i don't know how to configure my syno either for nzb :-P

margaale:
> tests: nothing special, IsReadOnly should be false. 

You mean when (status == Completed) don't you?

Taloth:
Yes.

margaale:
Don't forget about me please ;-)

src/NzbDrone.Core/Download/Clients/DownloadStation/Proxies/DownloadStationProxy.cs
```diff
@@ -39,12 +39,12 @@ public void AddTorrentFromData(byte[] torrentData, string filename, string downl
                 arguments.Add("destination", downloadDirectory);
             }
 
-            arguments.Add("file", new Dictionary<string, object>() { { "name", filename }, { "data", torrentData } });
+            arguments.Add("file", new Dictionary<string, object>() { { "name", filename }, { "data", data } });
            
             var response = ProcessRequest(DiskStationApi.DownloadStationTask, arguments, settings, $"add torrent from data {filename}", HttpMethod.POST);
```
Taloth: "add torrent" is no longer accurate. Same for other locations. replace it with 'task' and it's ok.

margaale: Sorry, forgot those ones...
Did you have time to see #1686?

Taloth:
@foXaCe You use usenet right? Do you have time to test it? I'll push the code to a branch so it gets a published release.

foXaCe:
ok tomorrow I will test

Taloth:
branch=usenet-download-station (you can switch to it the same way as develop)
version=2.0.0.4603

The build server is still building it. If you have auto-update, you can already switch to the branch and you'll get it automatically.

foXaCe:
@margaale @Taloth  hello, thanks for all job, usenet work very good. 
sonarr .nzb control work fine.

Taloth:
@margaale One thing, RemainingTime is different for torrent vs usenet.
With torrents we simply calculate ETA based on the current speed of the given torrent. So all torrents in parallel.
With usenet we use remaining size and the global usenet download speed and calculate it cumulative.
Unless the downloader already provides an eta of it's own.
So that may require a little bit of tweaking.

foXaCe:
question, option Completed Download Handling , remove after import or waits the parameters of DS station ?
(x days or x % ratio)

Taloth:
Sonarr doesn't wait for a particular ratio, it waits for DL station to report seeding has finished.

Taloth:
yes on the sum of download speeds (usenet only of course). Preferably a global rate, if it's available. Otherwise you may need an additional pass over the unfiltered tasks list to get all the speeds.
`var globalSpeed = nzbTasks.Select(GetRemainingSize).Sum() ?? 0.0;` or something like it.

Btw we do it with Nzbget:
https://github.com/Sonarr/Sonarr/blob/ddd119a4ebf8c80eefa1dfd8cacccf7e3ac3c265/src/NzbDrone.Core/Download/Clients/Nzbget/Nzbget.cs#L68
https://github.com/Sonarr/Sonarr/blob/ddd119a4ebf8c80eefa1dfd8cacccf7e3ac3c265/src/NzbDrone.Core/Download/Clients/Nzbget/Nzbget.cs#L105-L106

margaale:
Did i get it right?

Taloth:
Some considerations:
- In the initial sum: only count speed_download for actually downloading items. I don't know if speed_download is a historic average.
- Paused nzbs shouldn't be take into account for size/speed/eta
- You should also move the GetRemainingSize + sum above the category filter, coz downloading and queued nzbs from other categories should be taken into account.

That's assuming downloadstation actually returns them in download order.

margaale:
Since i have no previous knwoledge about nzb i've done all the changes that you've requested before.

foXaCe:
@Taloth hello, it possbile to push all in develop branch ?

valeriomazzeo:
really looking forward to the next release :D

Taloth:
Closed via https://github.com/Sonarr/Sonarr/commit/c56c83e169a2fa64579d9986b107a9fd89c429ee

PR #1686, opened 2017-02-16
Check output path on diskstation settings

margaale:
#### Database Migration
NO

#### Description
Some radarr users have been experiencing some issues regarding permissions and access to the output path, it seems to me that they don't know how to properly setup download station client, so it looks like it's  a good idea to check if the output path exists at the moment of adding this client.

### DIFF
--- a/src/NzbDrone.Core/Download/Clients/DownloadStation/TorrentDownloadStation.cs
+++ b/src/NzbDrone.Core/Download/Clients/DownloadStation/TorrentDownloadStation.cs
@@ -23,6 +23,7 @@ namespace NzbDrone.Core.Download.Clients.DownloadStation
         protected readonly IDownloadStationProxy _proxy;
         protected readonly ISharedFolderResolver _sharedFolderResolver;
         protected readonly ISerialNumberProvider _serialNumberProvider;
+        protected readonly IFileStationProxy _fileStationProxy;
 
         public TorrentDownloadStation(IDownloadStationProxy proxy,
                                ITorrentFileInfoReader torrentFileInfoReader,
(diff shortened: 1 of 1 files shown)


Taloth:
No need to fix it, but question: What does download station do if a subfolder does not exist, but the shared folder does? Like what happens when you use TvCategory.

margaale:
It returns isDir = false. That's why i wrote that extra check in there. And if the shared folder does not exists it returns additional_info in null.

Taloth:
What I mean, what happens if you actually create the download task?

Taloth:
Why I ask, is because I think we might only need to test this if TvDirectory is set, but not if TvCategory is. If the tvcategory directory is created automatically by downloadstation.

margaale:
Sorry, i thought you were talking about this PR. No, the full path must exists otherwise it returns an error when creating the task.

Taloth:
k, and yes I was talking about the PR-related stuff. I was just cleaning it up, copying it to usenetDLstation etc.

Taloth:
Here https://github.com/Sonarr/Sonarr/commit/3656807522fe2af0a5ee37bbf9c430a184545edf

What do you think?
Obviously I didn't test it.

Significant changes:
- `if (failures.Any()) return;` only needed to abort after connection failures, the other failures can be tested and reported on concurrently.
- Only test directory if either TvDirectory or TvCategory is specified.
- Report the failure for the relevant field with a short message, that way it fits in the UI better.
- Report exceptions globally without explanation, if an exception occurs, it's a failure on our part.

margaale:
Seems ok to me. Gonna test it at home later.

Taloth:
ok lemme know and I'll push to dev tomorrow.

Taloth:
Closed via https://github.com/Sonarr/Sonarr/commit/9304547c957037e505f8eb8b0b92c0e6b4d0b8ec

Issue #1395, opened 2016-07-25
Deleted episode not being unmonitored


ThatWhiteMonster:
Hi,

I've been using Sonarr on Windows to some extent for a few month now. It's a great app! I love it!
There are some features that I wish it had but that's another conversation (*cough*cough*subtitle handling*cough*cough*) .
;)

Anyway . . .

Recently, I've reconfigured it to actually handle the media management properly (monitoring existing episodes and such), instead of just acting as an "indexer watcher". It worked nice for a few days but now I've noticed something is wrong.

I have "Ignore Deleted Episodes" option enabled and it was working fine. 
I tend to only keep the recent episode and delete the whole series folder with that one episode after I watched it.
But now I see that after I delete an episode manually via Explorer and go and hit "Update series info and scan disk" button in the series page, the episode remains monitored and the path to deleted file is still being displayed.
Then if I try to delete via Sonarr and hit rescan, the episode remains monitored and the icon describing the quality profile comes back from "Missing" to let's say "HDTV-1080p". In both attempts, the episode in question is completely deleted - it's not in the library, it's not in the Sonarr recycle bin, it's not in the torrent client queue or downloads folder.

There's nothing in the logs if I rescan after deleting the episode. Obviously, when I try to delete it again via Sonarr, it says that the file doesn't exists. Yet it still comes back (as if it exists) with the quality indicator and previous path after another rescan.

Does this mean I can't delete the series folder along with the episode? Please help, as it would be nice if I didn't have to keep empty series folders in my library folder. That's why I have "Create empty series folders" unchecked after all.

Thank you in advance,
mnstr

markus101:
>Does this mean I can't delete the series folder along with the episode?

Yes, when the series folder is missing Sonarr doesn't assume that the it was deleted because that can cause issues in some setups. Ideally it would detect that they are missing regardless of the folder being present and either handle that as deleted or temporarily missing to deal with the issues that are currently caused in some setups. This is the relevant issue for that: https://github.com/Sonarr/Sonarr/issues/976

If you use Sonarr to delete the file, delete only the file from explorer or have Sonarr create empty series folders you won't run into the current limitation.

ThatWhiteMonster:
Thank you for the explanation, @markus101 .

> when the series folder is missing Sonarr doesn't assume that the it was deleted because that can cause issues in some setups

[...]

Issue #1257, opened 2016-04-22
Manually deleted series/episodes are remembered.


deadlyknapsack:
I wast testing a new download client with Sonarr and I manually deleted an episode to redownload after for testing purposes. When I deleted the episode manually from my diskstation, and restarted Sonarr, hit all the refresh tasks, it still said the episode was there. I am 100% confident it wasn't. Then, I deleted an entire series from my diskstation (at the series top directory), I went back to Sonarr and of course the series was still mentioned there, this is normal behaviour. However, Sonarr claimed all the episodes were supposedly still there, at the original location on my diskstation. Seems like a bug to me.

Details of setup:

- Synology DS716+ with DSM 5.2-5644 Update 7
- Sonarr version 2.0.0.3953

Please let me know what to do. I can provide logs if needed, no problem.

Taloth:
The series folder must exist for Sonarr to remove the episodes from the db. That's by design to handle unmounted root folders.

deadlyknapsack:
Pfff. First you fail to read my comment properly. Second you immediately close it. What's up with that?!

Again:

"I manually deleted an episode to redownload after for testing purposes. When I deleted the episode manually from my diskstation, and restarted Sonarr, hit all the refresh tasks, it still said the episode was there."

You completely ignore this part, where is this even remotely related to deleting a root series directory? This step (misunderstood, see below) is after this.

Second of all, I never stated that I deleted the root series folder. Maybe I didn't explain it very well, my bad, but I deleted a single series (sub)folder, like "The Simpsons" or "The Sopranos". 

Even after deleting the entire directory of the "The Simpsons" the episodes were still visible in Sonarr as if they were located within a folder called "The Simpsons" when I know for a fact, it wasn't.

Please re-open.

Taloth:
I actually never mentioned deleted root folders (D:/tv), I mentioned deleted series folders (D:/tv/Arrow/). And it was an answer to the second part of your post. (considering that was your current state, it seemed the most relevant)

My post may have been a bit short, sry for that, but I'm on a phone and we usually direct support questions to our forum.

Either way, if you can reproduce the original problem with the single episode. Post some debug logs on the forum.

Issue #1228, opened 2016-04-08
"Ignore Deleted Episodes" feature and file detection breaks when series folder is deleted


Ptgtp:
The "Ignore Deleted Episodes" feature seems to work great when I delete individual episode files or season folders. However, when I delete the entire series folder, not only do files not get unmonitored, Sonarr seems to keep thinking that the files are still there when they are clearly not (even after series refresh).

To "fix" Sonarr in these cases, I had to re-create an empty folder for the series, run the refresh and then delete that empty folder. Then the refresh realises that the files are gone and un-monitors everything as expected.

Personally, when I finish a season, I delete the whole series folder because I don't want to have hundreds of empty series folders in my media. But whether this is good practice or not, the fact that Sonarr keeps thinking the files are there when they aren't definitely seems broken to me.

Taloth:
It's by design to handle unmounted drives and shares.

Ptgtp:
Is there an easy way to recover from this, when series folders were deleted?

Right now, my Sonarr is in a pretty broken state as a result. I have probably over 50 shows that I deleted, but Sonarr still thinks it has it's files and will continue to think it has those files forever.

Of course I can manually re-create the series folders, making sure their names are exactly as Sonarr expects, but I was wondering if there was a less manual/quicker way?

Taloth:
Enable create empty series folders in the settings, then run Update Library twice. (first time creates the folders, second time to clear episodes from the db)

Issue #976, opened 2015-12-04
Missing Series Folder when intentionally deleted


markus101:
This comes up enough that it warrants some additional consideration I think, right now if the series folder is missing we don't rescan the series and subsequently treat the files as missing, this saves users that have disconnected network or USB drives from Sonarr forgetting about their episodes and possibly them being automatically ignored.

This behaviour isn't obvious to everyone, especially new users and they will often rescan the series manually a few times and not realize why its failing, there are a few things we could do here:
1. Make it more obvious, UI messaging, logging, etc
2. Add an option to skip this behaviour (yuck!)
3. Ignore the series folder missing when doing a manual refresh (from the UI or API)
4. Check for sibling folders of the series folder and use them to determine if we should rescan the series folder

`1` Seems like a good idea, but would also be complicated
`3` I like this idea, but it requires manual interaction
`4` Any major issues with this one? I think I like it the best, it would fail if things are symlinked and the series folder is actually missing, but there are siblings still there, but that seems like an acceptable trade-off

kayone:
we can do what plex does, mark them as unavailable. that way it's obvious what's happening and we aren't removing stuff from the DB either.

kayone:
1. this one is my favorite, tell the user what is happening, and let them decide what to do.
2. I don't like this because people that would get confused by this behaviour most likely won't know about this setting anyways.
3. it's still as destructive as automatic scan
4.I don't mind this, just feels too magical. how many of siblings are we gonna check? what if they delete two folders? what if they intentionally delete half of their series?

ZaLiTHkA:
Keeping within the mentality of not breaking the existing concepts across a few different pages, what about assigning this "status" a new series status colour?

Currently we have "continuing", "ended", "missing - monitored", "missing - not monitored"; so what about another called "unavailable" with a dark grey progress bar? This way the bar and the numbers could still reflect the last known count, while the physical file availability is clear to see. Naturally, this same colour would work on the individual season tags as well.

This begs one question in my mind though: will people expect the individual file rows in each season panel to reflect their individual availability as well, or will the availability check behind this not go that far? For the record, I don't think it should.

What about checking availability by seeing if there is any physical file that matches the chosen episode naming pattern. On the first match, the folder is considered "available" and the normal processes continue. If nothing matches after a short timeout, then the show is marked as "unavailable".

markus101:
Option 4 was implemented, closing this out.

PR #1690, opened 2017-02-18
Update series info with missing series folder

StormPooper:
#### Database Migration
NO

#### Description
I've just started using Sonarr and I'm really liking it. However, it was confusing when I had deleted a series folder (since there are no more episodes and due to various other factors leaving the empty folder in place is an issue) and the interface still insisted that there are episodes available until another is redownloaded. I understand from quickly Googling it that this is a common issue. 

Since for some users this will be desired behaviour (external drive, network issues, etc) I figured an advanced option that will remove episodes if the folder is missing would make the most sense - it is off by default to prevent clashing. I've also updated it so that if using the create folders option you don't have to run it twice.

Here's what the setting looks like along with the hover description (if anyone has a better way of wording this stuff, let me know):
settings

I've left it as an advanced setting like creating the empty folder since this is a change from the default behaviour, but if people would prefer it as a normal setting I can change that, or if anyone can think of a better way of handling this.

There are tests around this behaviour and I've added a couple around the creating folder behaviour too.

#### Todos
- [ ] Documentation?

- #1395
- #1257 (not directly asking about this, but answer refers to it)
- #1228
- #976 (though I obviously think being able to change the behaviour is better than just notifying, though maybe combined with some other UI changes it could work)
- #765 
- Possibly more using different phrases for the same issue

### DIFF
--- a/src/NzbDrone.Core/Configuration/ConfigService.cs
+++ b/src/NzbDrone.Core/Configuration/ConfigService.cs
@@ -148,13 +148,6 @@ namespace NzbDrone.Core.Configuration
             set { SetValue("RemoveFailedDownloads", value); }
         }
 
-        public bool UpdateSeriesIfFolderMissing
-        {
-            get { return GetValueBoolean("UpdateSeriesIfFolderMissing", false); }
-
-            set { SetValue("UpdateSeriesIfFolderMissing", value); }
-        }
-
         public bool CreateEmptySeriesFolders
         {
             get { return GetValueBoolean("CreateEmptySeriesFolders", false); }
--- a/src/NzbDrone.Core/Configuration/ConfigService.cs
+++ b/src/NzbDrone.Core/Configuration/ConfigService.cs
@@ -148,6 +148,13 @@ namespace NzbDrone.Core.Configuration
             set { SetValue("RemoveFailedDownloads", value); }
         }
 
+        public bool ClearSeriesIfFolderMissing
+        {
+            get { return GetValueBoolean("ClearSeriesIfFolderMissing", false); }
+
+            set { SetValue("ClearSeriesIfFolderMissing", value); }
+        }
+
         public bool CreateEmptySeriesFolders
         {
             get { return GetValueBoolean("CreateEmptySeriesFolders", false); }
--- a/src/NzbDrone.Core/Configuration/ConfigService.cs
+++ b/src/NzbDrone.Core/Configuration/ConfigService.cs
@@ -148,11 +148,11 @@ namespace NzbDrone.Core.Configuration
             set { SetValue("RemoveFailedDownloads", value); }
         }
 
-        public bool ClearSeriesIfFolderMissing
+        public bool UpdateSeriesIfFolderMissing
         {
-            get { return GetValueBoolean("ClearSeriesIfFolderMissing", false); }
+            get { return GetValueBoolean("UpdateSeriesIfFolderMissing", false); }
 
-            set { SetValue("ClearSeriesIfFolderMissing", value); }
+            set { SetValue("UpdateSeriesIfFolderMissing", value); }
         }
 
         public bool CreateEmptySeriesFolders
--- a/src/NzbDrone.Core/MediaFiles/DiskScanService.cs
+++ b/src/NzbDrone.Core/MediaFiles/DiskScanService.cs
@@ -67,27 +67,23 @@ namespace NzbDrone.Core.MediaFiles
         {
             var rootFolder = _diskProvider.GetParentFolder(series.Path);
 
-            var rootFolderExists = _diskProvider.FolderExists(rootFolder);
-            if (!rootFolderExists)
+            if (!_diskProvider.FolderExists(rootFolder))
             {
                 _logger.Warn("Series' root folder ({0}) doesn't exist.", rootFolder);
-                _eventAggregator.PublishEvent(new SeriesScanSkippedEvent(series,
-                    SeriesScanSkippedReason.RootFolderDoesNotExist));
+                _eventAggregator.PublishEvent(new SeriesScanSkippedEvent(series, SeriesScanSkippedReason.RootFolderDoesNotExist));
                 return;
             }
 
-            if (_diskProvider.GetDirectories(rootFolder).Empty() && !_configService.UpdateSeriesIfFolderMissing && !_configService.CreateEmptySeriesFolders)
+            if (_diskProvider.GetDirectories(rootFolder).Empty())
             {
-                _logger.Warn("Series' root folder ({0}) is empty and no options set to create folders or clean DB.", rootFolder);
-                _eventAggregator.PublishEvent(new SeriesScanSkippedEvent(series,
-                    SeriesScanSkippedReason.RootFolderIsEmpty));
+                _logger.Warn("Series' root folder ({0}) is empty.", rootFolder);
+                _eventAggregator.PublishEvent(new SeriesScanSkippedEvent(series, SeriesScanSkippedReason.RootFolderIsEmpty));
                 return;
             }
 
             _logger.ProgressInfo("Scanning disk for {0}", series.Title);
 
-            var seriesFolderAlreadyExists = _diskProvider.FolderExists(series.Path);
-            if (!seriesFolderAlreadyExists)
+            if (!_diskProvider.FolderExists(series.Path))
             {
                 if (_configService.CreateEmptySeriesFolders)
                 {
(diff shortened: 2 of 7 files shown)


src/NzbDrone.Core/MediaFiles/DiskScanService.cs
```diff
@@ -67,26 +67,29 @@ public void Scan(Series series)
...
             {
                 _logger.Warn("Series' root folder ({0}) doesn't exist.", rootFolder);
-                _eventAggregator.PublishEvent(new SeriesScanSkippedEvent(series, SeriesScanSkippedReason.RootFolderDoesNotExist));
+                _eventAggregator.PublishEvent(new SeriesScanSkippedEvent(series,
+                    SeriesScanSkippedReason.RootFolderDoesNotExist));
                 return;
             }
 
-            if (_diskProvider.GetDirectories(rootFolder).Empty())
+            if (_diskProvider.GetDirectories(rootFolder).Empty() && !_configService.UpdateSeriesIfFolderMissing && !_configService.CreateEmptySeriesFolders)
```
markus101: We shouldn't try to create a folder or rescan the folder if the root folder is empty. It's a safety net against `/mnt/tv` being present, but not properly mounted.

src/NzbDrone.Core/MediaFiles/DiskScanService.cs
```diff
@@ -67,26 +67,29 @@ public void Scan(Series series)
         {
             var rootFolder = _diskProvider.GetParentFolder(series.Path);
 
-            if (!_diskProvider.FolderExists(rootFolder))
+            var rootFolderExists = _diskProvider.FolderExists(rootFolder);
+            if (!rootFolderExists)
             {
                 _logger.Warn("Series' root folder ({0}) doesn't exist.", rootFolder);
-                _eventAggregator.PublishEvent(new SeriesScanSkippedEvent(series, SeriesScanSkippedReason.RootFolderDoesNotExist));
+                _eventAggregator.PublishEvent(new SeriesScanSkippedEvent(series,
```
markus101: Curious why this line was changed, but the long `if` below is longer than this ever was.

src/NzbDrone.Core/MediaFiles/DiskScanService.cs
```diff
@@ -67,26 +67,29 @@ public void Scan(Series series)
         {
             var rootFolder = _diskProvider.GetParentFolder(series.Path);
 
-            if (!_diskProvider.FolderExists(rootFolder))
+            var rootFolderExists = _diskProvider.FolderExists(rootFolder);
```
markus101: Why extract this to a variable?

src/NzbDrone.Core/MediaFiles/DiskScanService.cs
markus101: Come to think of it, if we only rescan _if_ there is at least one folder present in the root folder I think that would eliminate there needing to be an option at all to rescan missing folders. We'd still have two safety nets (root folder exists and isn't empty) and not have to rely on a setting that has a chance of getting enabled and causing issues. @Taloth @kayone thoughts?

src/NzbDrone.Core/MediaFiles/DiskScanService.cs
```diff
@@ -67,26 +67,29 @@ public void Scan(Series series)
...
+                _eventAggregator.PublishEvent(new SeriesScanSkippedEvent(series,
+                    SeriesScanSkippedReason.RootFolderIsEmpty));
+                return;
             }
 
             _logger.ProgressInfo("Scanning disk for {0}", series.Title);
-            
-            if (!_diskProvider.FolderExists(series.Path))
+
+            var seriesFolderAlreadyExists = _diskProvider.FolderExists(series.Path);
```
markus101: I see we're using this below, but if we return early after creating the series folder and publish the `_eventAggregator.PublishEvent(new SeriesScannedEvent(series));` event we save ourselves a bunch of repeated checks.

src/NzbDrone.Core/Configuration/ConfigService.cs
```diff
@@ -148,6 +148,13 @@ public bool RemoveFailedDownloads
             set { SetValue("RemoveFailedDownloads", value); }
         }
 
+        public bool UpdateSeriesIfFolderMissing
```
markus101: I think `Update` is the wrong term here, we're going to upgrade regardless (fetch info from skyhook), but we're going to "scan" it. `ScanMissingSeriesFolders` would be better I think.

src/UI/Settings/MediaManagement/Sorting/SortingViewTemplate.hbs
```diff
@@ -2,6 +2,29 @@
...
+                    <p>
+                        <span>Yes</span>
+                        <span>No</span>
+                    </p>
+
+                    <div class="btn btn-primary slide-button"/>
+                </label>
+
+                <span class="help-inline-checkbox">
+                    <i class="icon-sonarr-form-info" title="Update series info for existing episodes even if the series folder is missing during disk scan"/>
```
markus101: Again, update is the wrong term here. It may be moot if we decide the foot folder existing and not being empty is enough safety though.

src/NzbDrone.Core/MediaFiles/DiskScanService.cs
StormPooper: Good spot, this was from a previous commit and I forgot to inline it again.

src/NzbDrone.Core/MediaFiles/DiskScanService.cs
StormPooper: This is just a formatting thing that my IDE did automatically, the rest of the line is on the line below, I'll undo that.

src/NzbDrone.Core/Configuration/ConfigService.cs
StormPooper: That's a much better name for it, I'll change it if we go with the setting.

src/NzbDrone.Core/MediaFiles/DiskScanService.cs
StormPooper: Since I'm on Windows and use an internal HDD it will always be mounted, so I thought it would be useful to be able to override it, but I see the point, I'll change that back.

I like the idea of removing the setting and just doing it if the folder has content since you know it's mounted. I avoided that at first because there's an edge case that if you do in fact completely empty the folder out it won't reflect that, but I think since it'll update as soon as something is in there it's a decent compromise.

src/NzbDrone.Core/MediaFiles/DiskScanService.cs
StormPooper: I did it this way so that the `_mediaFileTableCleanupService.Clean` call can still be done on the created folder so it's up to date, but I could do the check at that point and then exit early.

Taloth:
Yeah, I definitely don't want to see an config option added for this.

The unmount detection just needs some tweaking. I'm not worried about unmounted series folder, only root folders. If the root folder is missing, or empty, then it should consider the series unmounted.

Also, we need a healthcheck to report to the use that episode file information is being kept in the database due to a likely unmounted drive.

StormPooper:
Updated based on the suggestions from @markus101 and @Taloth. If there is at least one series folder around, it will clean the DB if the series folder does not exist, but if the root folder does not exist or there are no series folders within the root folder, it will skip the scan. The config option has been removed.

StormPooper:
@markus101, @Taloth, any thoughts on the latest version of this?

Taloth:
PR looks fine as is.

@StormPooper Regarding the HealthCheck. Even with only the root folder check, the user isn't aware that Sonarr is keeping the information in the database due to an unmounted root folder.
The thing is, we can't just throw a healthcheck if the root folder is empty. We can only do that if there's a series in that root folder with episodefiles in the database.
So maybe an UnmountedSeriesHealthCheck that checks whether a series has episodefiles in an unmounted root folder.

On the long run we might want to add an 'archived' flag to the episode/series. Add add the appropriate UI flair to show that. But that's more complicated and requires a discussion about design.

Lemme know if you want to tackle that UnmountedSeriesHealthCheck, if not then it's fine too, then we'll merge it in as is.

StormPooper:
@Taloth that makes sense, happy to add the HealthCheck but it might be a few days before I get a chance to do it. How about merging this part in and I'll raise a separate pull request for it when I've done the HealthCheck?

Taloth:
@StormPooper  Agreed.

Issue #1700, opened 2017-02-21
KeyNotFoundException in DownloadStation


Taloth:
@margaale Looks to me like a error Code in DiskStationError that isn't listed.
No need to fix it immediately, coz it's likely a temporary misconfiguration by the user that caused the error.

KeyNotFoundException: The given key was not present in the dictionary.
  File "\NzbDrone.Core\Download\Clients\DownloadStation\SerialNumberProvider.cs", line 34, in GetSerialNumber
    System.String GetSerialNumber(NzbDrone.Core.Download.Clients.DownloadStation.DownloadStationSettings)
  File "\NzbDrone.Common\Cache\Cached.cs", line 81, in Get
    System.String Get(System.String, System.Func`1[System.String], System.Nullable`1[System.TimeSpan])
  File "\NzbDrone.Core\Download\Clients\DownloadStation\SerialNumberProvider.cs", line 34, in <GetSerialNumber>b__0
    System.String <GetSerialNumber>b__0()
  File "\NzbDrone.Core\Download\Clients\DownloadStation\SerialNumberProvider.cs", line 45, in GetHashedSerialNumber
    System.String GetHashedSerialNumber(NzbDrone.Core.Download.Clients.DownloadStation.DownloadStationSettings)
  File "\NzbDrone.Core\Download\Clients\DownloadStation\Proxies\DSMInfoProxy.cs", line 29, in GetSerialNumber
    System.String GetSerialNumber(NzbDrone.Core.Download.Clients.DownloadStation.DownloadStationSettings)
  File "\NzbDrone.Core\Download\Clients\DownloadStation\Proxies\DiskStationProxyBase.cs", line 81, in ProcessRequest
    NzbDrone.Core.Download.Clients.DownloadStation.Responses.DiskStationResponse`1[NzbDrone.Core.Download.Clients.DownloadStation.Responses.DSMInfoResponse] ProcessRequest[DSMInfoResponse](NzbDrone.Core.Download.Clients.DownloadStation.DiskStationApi, Syste...
  File "\NzbDrone.Core\Download\Clients\DownloadStation\Responses\DiskStationError.cs", line 95, in GetMessage
    System.String GetMessage(NzbDrone.Core.Download.Clients.DownloadStation.DiskStationApi)
  File "/spksrc/native/mono/work-native/mono-4.6.2/mcs/class/referencesource/mscorlib/system/collections/generic/dictionary.cs", line 183, in get_Item
    System.String get_Item(Int32)

[...]

PR #1703, opened 2017-02-21
Diskstation: Fix for key not found, returning a generic error instead

margaale:
#### Database Migration
NO

#### Description
Returns a generic error message if error-code is not in the dictionary.

#1700

### DIFF
--- a/src/NzbDrone.Core.Test/Download/DownloadClientTests/DownloadStationTests/TorrentDownloadStationFixture.cs
+++ b/src/NzbDrone.Core.Test/Download/DownloadClientTests/DownloadStationTests/TorrentDownloadStationFixture.cs
@@ -60,7 +60,7 @@ namespace NzbDrone.Core.Test.Download.DownloadClientTests.DownloadStationTests
                 Id = "id1",
                 Size = 1000,
                 Status = DownloadStationTaskStatus.Waiting,
-                Type = DownloadStationTaskType.BT,
+                Type = DownloadStationTaskType.BT.ToString(),
                 Username = "admin",
                 Title = "title",
                 Additional = new DownloadStationTaskAdditional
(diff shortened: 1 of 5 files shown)


Taloth:
Another user has authentication issues. Must something be set on the download station for auth to work on the api? https://www.reddit.com/r/sonarr/comments/5vdxdn/sonarr_v2004613_and_btn_update/de2aqe0/
If so, then clarifying that in the Test ValidationError would be a good idea.

Taloth:
And another, which is a significant bug: https://forums.sonarr.tv/t/synology-download-station-ssl-sonarr-v2-0-0-4613/13779

margaale:
1) Not aware that any special configuration should be done.
2) That seems pretty bad, "type string Possible types: BT, NZB, http, ftp, eMule 1 and later" taken from https://global.download.synology.com/download/Document/DeveloperGuide/Synology_Download_Station_Web_API.pdf page 28. Guess they didn't include it in the list. I' gonna submit a patch right now.

Taloth:
Maybe the api object should be a string, that way we can handle new ones.

margaale:
Yeah, thats's a good idea...

margaale:
Done

Taloth:
Missing one thing: A unittest.  But i'll take that on and merge to dev.

margaale:
Ha, sorry, i'm at work now. No much time...

Taloth:
It was a 'little' bit more involved than I wanted, but I think I made the right choice.
I pulled out the filter from the proxy, it didn't really belong there coz proxy needs to be as thin as possible. That allowed me to add the appropriate tests to Torrent/UsenetDownloadStation.

margaale:
Yep, that was my next step. But i had no time now to do it. Thanks

PR #1715, opened 2017-02-25
Sorting by torrent peers priority increased

Xedecimal:
#### Database Migration
NO

#### Description
This would first sort by peers in a torrent result, then by the other sorts. Hopefully making a given quality have a ordered list by quality, then health.

### DIFF
--- a/src/NzbDrone.Core/DecisionEngine/DownloadDecisionComparer.cs
+++ b/src/NzbDrone.Core/DecisionEngine/DownloadDecisionComparer.cs
@@ -24,7 +24,7 @@ namespace NzbDrone.Core.DecisionEngine
             var comparers = new List<CompareDelegate>
             {
                 ComparePeersIfTorrent,
-				CompareQuality,
+                CompareQuality,
                 CompareProtocol,
                 CompareEpisodeCount,
                 CompareEpisodeNumber,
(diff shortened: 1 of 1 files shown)


markus101:
This would rank something with 400 peers and SDTV above something above something with 300 peers an Bluray 1080p. It would also ignore the preferred protocol, the correct preference of episodes and the episode order (this is so it attempts to grab episode 1 before episode 2, etc).

Given releases with different qualities, the same protocol and and exact same episode(s) the order should be ranked by quality then peers already.

Xedecimal:
Where would the episode order come into play? I've only ever seen the capability to search for a single file for an entire season, or a single episode at a time.

Maybe just move quality just above peers ? I've found that it still finds the quality I was asking for just fine with this in testing. Maybe it's just something in my configuration that causes this to work perfectly for me.

markus101:
> Where would the episode order come into play?

A season search where there aren't any season packs in the results it would apply.

>Maybe just move quality just above peers

Protocol still needs to come before peers and the episode ordering won't affect it if it's comparing peers for the same episode. If you only used torrents, didn't care about episode order or the number of episodes in the release (preferring less means there won't be an issue later trying to upgrade when another quality has them split) then the changes work, but it shouldn't change anything (and if it does something else may be wrong).

PR #1717, opened 2017-02-27
Consistent formatting for MediaInfo in various locations

### LINKED ISSUE #1534: Kodi NFO Stream Details - MP3 Error
Sonarr is currently saving `<codec>MPEG Audio</codec>` for MP3 audio tracks.
Kodi needs to see `<codec>MP3</codec>` to recognise the MP3 track.

This fix has already been done for [AC3](https://github.com/Sonarr/Sonarr/commit/c1f59a55c66515e0510784a861c51ff32912b0da)

markus101:
Fixed: Stream details for MP3 and EAC3 in Kodi metadata
Closes #1534

#### Database Migration
NO

#### Description
This will allow us to consistently format MediaInfo in various locations as well as expose in on the API with the same consistency. There is a bit of duplication everywhere this is used right now, we could either add additional properties to the `MediaInfoModel` and always format them when returning them or only do that in the API module and leave the rest as-is. Even if we did it when returning the `EpisodeFile`s we'd still need to handle the case when importing a new episode file. Thoughts?

* #1534

### DIFF
--- a/src/NzbDrone.Core.Test/MediaFiles/MediaInfo/MediaInfoFormatterTests/FormatAudioCodecFixture.cs
+++ b/src/NzbDrone.Core.Test/MediaFiles/MediaInfo/MediaInfoFormatterTests/FormatAudioCodecFixture.cs
@@ -1,0 +1,77 @@
+﻿using FluentAssertions;
+using NUnit.Framework;
+using NzbDrone.Core.MediaFiles.MediaInfo;
+
+namespace NzbDrone.Core.Test.MediaFiles.MediaInfo.MediaInfoFormatterTests
+{
+    [TestFixture]
+    public class FormatAudioCodecFixture
+    {
+        [Test]
+        public void should_return_AC3_for_AC_3()
+        {
+            var mediaInfoModel = new MediaInfoModel
+            {
+                AudioFormat = "AC-3"
+            };
+
+            MediaInfoFormatter.FormatAudioCodec(mediaInfoModel).Should().Be("AC3");
+        }
+
+        [Test]
+        public void should_return_EAC3_for_E_AC_3()
+        {
+            var mediaInfoModel = new MediaInfoModel
+            {
+                AudioFormat = "E-AC-3"
+            };
+
+            MediaInfoFormatter.FormatAudioCodec(mediaInfoModel).Should().Be("EAC3");
+        }
+
+        [Test]
+        public void should_return_MP3_for_MPEG_Audio_with_Layer_3_for_the_profile()
+        {
+            var mediaInfoModel = new MediaInfoModel
+            {
+                AudioFormat = "MPEG Audio",
+                AudioProfile = "Layer 3"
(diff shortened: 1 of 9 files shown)


src/NzbDrone.Core.Test/MediaFiles/MediaInfo/MediaInfoFormatterTests/FormatAudioCodecFixture.cs
```diff
@@ -0,0 +1,77 @@
...
+using NUnit.Framework;
+using NzbDrone.Core.MediaFiles.MediaInfo;
+
+namespace NzbDrone.Core.Test.MediaFiles.MediaInfo.MediaInfoFormatterTests
+{
+    [TestFixture]
+    public class FormatAudioCodecFixture
+    {
+        [Test]
+        public void should_return_AC3_for_AC_3()
```
Taloth: I'd simply these tests like the parser tests, with `[TestCase(format, profile, result)]should_simplify_format`  and `[TestCase(format)]should_keep_format`.
That way it's easier to add new cases.

Same for video, of course.

src/NzbDrone.Core/MediaFiles/MediaInfo/MediaInfoFormatter.cs
```diff
@@ -0,0 +1,87 @@
...
+            }
+
+            if (audioFormat == "E-AC-3")
+            {
+                return "EAC3";
+            }
+
+            if (audioFormat == "MPEG Audio")
+            {
+                return mediaInfo.AudioProfile == "Layer 3" ? "MP3" : audioFormat;
```
Taloth: Merge ternary expression with if statement and use fall-through for the non Layer 3 case.

Taloth:
How about we log 'unknown' codecs? At Debug level, but pipe it to sentry as well so we can add em?

markus101:
> How about we log 'unknown' codecs? At Debug level, but pipe it to sentry as well so we can add em?

That might cause a bit of noise that we don't need (or the code to handle it), `DTS` is caught by the default case, but that's because it's formatted well enough to begin with. I think having the ability to show it in the UI might expose some more that are wrong so it should improve naturally as people start using it. We can turn it on and see how noisy it gets though.

Taloth:
I'd argue we should code all of em, and the sentry logger has a rate limiter.

src/NzbDrone.Core/MediaFiles/MediaInfo/MediaInfoFormatter.cs
markus101: I didn't let it fall through since we want to handle it anyways, though I could handle it as a separate `if` instead of the ternary.

markus101:
I did the fix ups 2 weeks ago and forgot to push them... rebased and pushed now.

markus101:
This could use another quick review.

PR #1726, opened 2017-03-02
Testing Custom Script executes the script and verifies the exit code

markus101:
#### Database Migration
NO

#### Description
Verifies that the script executes, without error and that the exit code is zero.


Taloth:
Won't this be breaking for scripts that don't handle it gracefully?

markus101:
Yeah, that's my concern. A couple ways to mitigate that:

- Warning message (which inevitably won't be read)
- Implement it in v3

markus101:
Merged into the v3 work.

Issue #1734, opened 2017-03-04
Download Station: Failed to get serial number


Faerus:
Hi there,

I'm not able to use Download Station as torrent download client as following error always occurs.
```
17-3-5 00:17:20.6|Error|DSMInfoProxy|Failed to get serial number. Reason: The requested version does not support the functionality
17-3-5 00:17:20.6|Warn|SerialNumberProvider|Could not get the serial number from Download Station 127.0.0.1:5000

[v2.0.0.4613] NzbDrone.Core.Download.Clients.DownloadClientException: Failed to get serial number. Reason: The requested version does not support the functionality
  at NzbDrone.Core.Download.Clients.DownloadStation.Proxies.DiskStationProxyBase.ProcessRequest[DSMInfoResponse] (DiskStationApi api, System.Collections.Generic.Dictionary`2 arguments, NzbDrone.Core.Download.Clients.DownloadStation.DownloadStationSettings settings, System.String operation, HttpMethod method, Int32 retries) [0x000c2] in M:\BuildAgent\work\b69c1fe19bfc2c38\src\NzbDrone.Core\Download\Clients\DownloadStation\Proxies\DiskStationProxyBase.cs:84 
  at NzbDrone.Core.Download.Clients.DownloadStation.Proxies.DSMInfoProxy.GetSerialNumber (NzbDrone.Core.Download.Clients.DownloadStation.DownloadStationSettings settings) [0x00036] in M:\BuildAgent\work\b69c1fe19bfc2c38\src\NzbDrone.Core\Download\Clients\DownloadStation\Proxies\DSMInfoProxy.cs:29 
  at NzbDrone.Core.Download.Clients.DownloadStation.SerialNumberProvider.GetHashedSerialNumber (NzbDrone.Core.Download.Clients.DownloadStation.DownloadStationSettings settings) [0x00000] in M:\BuildAgent\work\b69c1fe19bfc2c38\src\NzbDrone.Core\Download\Clients\DownloadStation\SerialNumberProvider.cs:45 
  at NzbDrone.Core.Download.Clients.DownloadStation.SerialNumberProvider+<>c__DisplayClass4_0.<GetSerialNumber>b__0 () [0x00000] in M:\BuildAgent\work\b69c1fe19bfc2c38\src\NzbDrone.Core\Download\Clients\DownloadStation\SerialNumberProvider.cs:34 
[...]

PR #1746, opened 2017-03-09
Diskstation - Use available api version number in Syno.DSM.Info

margaale:
#### Database Migration
NO

#### Description
It seems that older version of DSM use api version 1 while newers use version strictly version 2 of it. So, this PR use whatever version the DSM reports. Refactor some code in the way.

* #1734


margaale:
I was thinking about making Resources non static. As it is it could bring troubles if an user adds two differents DiskStation, one with an older and the other with a newer DSM version.

Taloth:
I agree, shouldn't use a static for that.

But a question, what happens if you don't provide `&version=1/2` as query param? error? or does it magically return something?

Alternative would be to just cache the version in a ICached in DSMInfoProxy.
Basically:
- fetch min/maxversion & store in cache.
- Try version=x.
- try catch: if failure due to version, re-fetch min/maxversion & store in cache.
- Try version=x.
- try catch: throw.

margaale:
I'll try that when i get home (not send version). Making it non static should solve the problem, but i was thinking about adding a cache that stores Resources too in case of an update of the DSM.

margaale:
> But a question, what happens if you don't provide &version=1/2 as query param? error? or does it magically return something?

Yep, it returns an error

margaale:
@Taloth  What do you think of it now?

Taloth:
It all seems quite a big change for something that trivial. I don't think you need a separate Task/Info proxy. If you had extracted the version check into an appropriate service. You could've done a `if (IsVersionSupported(2)) { do v=2 } else if (IsVersionSupported(1)) { do v=1 } else { throw; }`, instead of using an abstract Api property and related methods.

But I need to look it over again to be able to point at more specifics.

margaale:
The big change is not just for this. It is something I had in mind for a while...

src/NzbDrone.Core/Download/Clients/DownloadStation/Proxies/DiskStationProxyBase.cs
```diff
@@ -174,46 +202,46 @@ private HttpRequest BuildRequest(DownloadStationSettings settings, DiskStationAp
...
-                    throw new DownloadClientException("Api not implemented");
+                return Enumerable.Range(_resources[Api.Value].MinVersion, _resources[Api.Value].MaxVersion - _resources[Api.Value].MinVersion + 1);
             }
+
+            throw new DownloadClientException("Api not implemented");
+        }
+
+        protected int GetMinVersion(DownloadStationSettings settings)
+        {
+            return _cache.Get(Api.Key, () => GetApiVersion(settings).Min(), TimeSpan.FromHours(1));
```
Taloth: The cache key isn't unique per host:port, so you still have the 'two synology' issue.

src/NzbDrone.Core/Download/Clients/DownloadStation/Proxies/DiskStationProxyBase.cs
```diff
@@ -1,58 +1,86 @@
...
         private readonly IHttpClient _httpClient;
         protected readonly Logger _logger;
+        private ICached<int> _cache;
         private bool _authenticated;
-
-        static DiskStationProxyBase()
+        
+        public DiskStationProxyBase(IHttpClient httpClient,
+                                    Logger logger,
+                                    ICacheManager cacheManager)
```
Taloth: We generally do Logger as last parameter and new stuff from derived classes get prepended.

src/NzbDrone.Core/Download/Clients/DownloadStation/Proxies/DSMInfoProxy.cs
```diff
@@ -13,20 +14,28 @@ public interface IDSMInfoProxy
...
+                return new KeyValuePair<string, DiskStationApi>("SYNO.DSM.Info", DiskStationApi.DSMInfo);
+            }
+        }
+
         public string GetSerialNumber(DownloadStationSettings settings)
         {
             var arguments = new Dictionary<string, object>() {
-                { "api", "SYNO.DSM.Info" },
-                { "version", "2" },
+                { "version", GetMinVersion(settings) },
```
Taloth: The risk here is that you know how to handle version 1 & 2 responses. What if you suddenly talk to a future DSM with version 3 that returns different data?

src/NzbDrone.Core/Download/Clients/DownloadStation/Proxies/DiskStationProxyBase.cs
```diff
@@ -1,58 +1,86 @@
...
         {
-            Resources = new Dictionary<DiskStationApi, string>
+            _httpClient = httpClient;
+            _logger = logger;
+            _cache = cacheManager.GetCache<int>(GetType());
```
Taloth: This cache is per derived type, is that intended?

Taloth:
Wouldn't it have been easier and cleaner to have a separate Proxy to handle (and cache) all SYNO.DSM.Info stuff? (Afaik it doesn't need auth)
With an interface so you can implement `bool SupportsVersion(string api, int version, DownloadStationSettings settings)`, which could then be used by all the proxies to select the right version and request/response handling.

src/NzbDrone.Core/Download/Clients/DownloadStation/Proxies/DiskStationProxyBase.cs
margaale: Yes it is. That's what i wanted to do. Before the proxybase knew about all apis, that is why i added the abstract property now.

src/NzbDrone.Core/Download/Clients/DownloadStation/Proxies/DSMInfoProxy.cs
margaale: If they return different data it's going to break, but it's not the purpose of this to handle that scenario

margaale:
In my tests i have to login before consuming SYNO.DSM.Info otherwise it returns an error.
Sure, i'm with you if you want me to add support for diferent versions of the api. I can remove the cache from baseproxy if you really want that.

Taloth:
I meant SYNO.API.Info not SYNO.DSM.Info. API.Info is for the versions, DSM.Info for the serial. But nevermind.

margaale:
Ohhhh, OK. My goal was to keep changes at a minimum, that's why I kept a DownloadStation proxy that is a composite(is ok using  this word in this context?) of the other two...

src/NzbDrone.Core/Download/Clients/DownloadStation/Proxies/DownloadStationProxy.cs
```diff
@@ -1,121 +1,56 @@
...
 
-    public class DownloadStationProxy : DiskStationProxyBase, IDownloadStationProxy
+    public class DownloadStationProxy : IDownloadStationProxy
     {
-        public DownloadStationProxy(IHttpClient httpClient, Logger logger)
-            : base(httpClient, logger)
+        private IDownloadStationTaskProxy _taskProxy;
+        private IDownloadStationInfoProxy _infoProxy;
+
+        public DownloadStationProxy(IDownloadStationTaskProxy taskProxy, IDownloadStationInfoProxy infoProxy)
```
markus101: A proxy shouldn't call other proxies, a proxy is meant to be a simple interface for mocking so it's easy to mock API calls without having to build the URL each time. Calling another proxy likely means there is too much business logic there. In this case it looks like an unnecessary layer though.

margaale:
@markus101 did you even bother to read any of the comments that I wrote? What business logic your business is managing episode's and series not to connect and authorize with the client? There's is one proxy per api except authentication and info which are needed for every proxy. If you don't like this PR yo can close it or tell me and I'll do it myself.

markus101:
> did you even bother to read any of the comments that I wrote? 

Code comments or comments on the PR?

> What business logic your business is managing episode's and series not to connect and authorize with the client?

I don't follow, if you're responding to one of my comments it'd be better to respond to the comment itself to maintain context. For `DownloadStationProxy` I don't see why it is there, all it's methods just call a method in another proxy, can't the caller of those methods call those methods directly instead of the extra method in `DownloadStationProxy`?

> If you don't like this PR yo can close it or tell me and I'll do it myself.

It's a PR review, it's not personal and I'm not attacking you, I'm asking questions about the code. I like the direction the PR is going, I just don't like how it's getting there at the moment, but that's why we have peer code reviews, so we can improve it.

margaale:
One of firsts Taloth's comments was that this change was too big, i wrote
in the code review comments that i was trying to change as little as i
could and, at the same time, fix the related error. Eventually  it will get
to something closer towards what you are heading, now i'm trying to avoid
touching the client logic and that is why a kept DownloadStationProxy as
awful as it is now.

Sorry if i was too aggressive...

Taloth:
@margaale I asked markus to look over the PR since I had my concerns about a few points and needed a second opinion. I've had my fair share of frustrations over comments the other devs made about my code, but it got better because of it. It's constructive criticism.

I did say that I felt the change was fairly big for what we were trying to solve, I thought that only tweaking the DSM.Info proxy would've been sufficient to patch it up.
However, if we're going to split up DownloadStationProxy, we should do it right.

To explain my thoughts:
It is fairly complex (for a proxy)... the code that pulls together _resources, the api info etc. I should've commented on it back when we first started out in december, but I didn't.
When we try to make a call (via ProcessRequest) we do a GetApiVersion (really an UpdateApiInfoAndReturnVersion), which pulls info from the Api property, add 'API.Auth' to the list and then queries the API. But ProcessRequest also accesses the ApiInfo directly via the _resources field, it's convoluted.

First part is to deal with GetApiVersion name and dual role, what we need is a single method that returns information about the Api we're interested in, a `GetApiInfo` method.
- Replace _cache with a `ICached<DiskStationApiInfo>`, key by host+port+apiname.
- Remove _resources, use _cache instead. (That fixes the multi synology edge-case)
- Move update logic from `GetApiVersion` to `private UpdateApiInfo(settings)`, which has the sole job of querying the api and updating _cache, no return value.
- Create `protected DiskStationApiInfo GetApiInfo(DiskStationApi/string api, ...)` replacing your existing calls to GetApiVersion & _resources, call UpdateApiInfo as needed otherwise return the _cache value.
- Remove GetMinVersion, DSMInfoProxy can call GetApiInfo.
- We also don't really need Enumerable.Range, Minmax is enough.

The result is that all logic involving getting the api info resides in `GetApiInfo`. Internally it takes care of fetching and caching it.

Having a property Api with KeyValuePair as return is really a symptom of a problem. Ideally I'd have the constructor 'register' the DiskStationApi & string, the base-class can then register API.Auth in the same way, and even API.Info (by including a predefined DiskStationApiInfo).
And then we can refer to the api via the enum value and UpdateApiInfo can do it's job.
But i'm ok with leaving the Api property for now.

If we want to go further we can, especially around ProcessRequest, to better leverage the HttpRequestBuilder. But I fear i'm causing you enough pain as is.

margaale:
@Taloth Have a look now please... Did i get you right?

Taloth:
Yup, that's what I meant.

Part 2:
Two options:
1) Have DiskStationApi & string part of the constructor base(...) call, and store them separately in two readonly fields.
or
2) Have two separate abstract properties instead of the KeyValuePair. (this is my least favorite option)
or
3) Add `RegisterApi(DiskStationApi api, string name)` this would of course store it in an `Dictionary<DiskStationApi, string> _registeredApis` field.

Either option would be ok.

Btw. Adding an if(.. == DiskStationApi.SynoApiInfo) in GetApiInfo to handle the Api.Info edgecase and I'd be happy.

If I wanted to go on refactoring, I'd change BuildRequest to accept a DiskStationApiInfo instance, have it return a httpRequestBuilder and call AddQueryParam & AddFormUpload instead of the 'arguments' dict.
You'd then call ProcessRequest with the httpRequestBuilder as argument.
That would allow UpdateApiInfo to bypass the GetApiInfo logic _and_ potentially bypass the authentication as well.
But that's not needed to fix the original issue, so i'd leave that as is.

margaale:
Something like latest commit? Tested against my diskstation, it's working ok.

src/NzbDrone.Core/Download/Clients/DownloadStation/Proxies/DiskStationProxyBase.cs
```diff
@@ -105,115 +179,109 @@ private void AuthenticateClient(DownloadStationSettings settings)
...
+                 { "format", "sid" },
                  { "session", "DownloadStation" },
              };
 
-            var authLoginRequest = BuildRequest(settings, DiskStationApi.Auth, arguments, HttpMethod.GET);
-            authLoginRequest.StoreResponseCookie = true;
+            var authInfo = GetApiInfo(DiskStationApi.Auth, settings);
 
-            var response = _httpClient.Execute(authLoginRequest);
+            var requestBuilder = BuildRequest(settings, authInfo, HttpMethod.GET);
```
Taloth: Now you have the requestBuilder, isn't it easier to just call AddQueryParam, instead of the dictionary? This applies in multiple cases.
``` 
var requestBuilder = BuildRequest(settings, authInfo, HttpMethod.GET);
requestBuilder.AddQueryParam("version", "1");
requestBuilder.AddQueryParam("method", "login");
requestBuilder.AddQueryParam("account", settings.Username);
requestBuilder.AddQueryParam("passwd", settings.Password);
requestBuilder.AddQueryParam("format", "sid");
requestBuilder.AddQueryParam("session", "DownloadStation");
```

BuildRequest can add the 'api' queryparam, and potentially deal with version & method too via an method parameter, simply because all api calls have those query params. And gives the BuildRequest a chance to verify your specified version vs the MinMax.

src/NzbDrone.Core/Download/Clients/DownloadStation/Proxies/DiskStationProxyBase.cs
```diff
@@ -105,115 +179,109 @@ private void AuthenticateClient(DownloadStationSettings settings)
...
-                }
-            }
-            else
-            {
-                foreach (var arg in arguments)
-                {
-                    requestBuilder.AddQueryParam(arg.Key, arg.Value);
-                }
-            }
+            return requestBuilder;
```
Taloth: As mentioned, you already know the api here via apiInfo, so afaik you can add the &api= queryparam.

src/NzbDrone.Core/Download/Clients/DownloadStation/Proxies/DiskStationProxyBase.cs
```diff
@@ -105,115 +179,109 @@ private void AuthenticateClient(DownloadStationSettings settings)
...
-                    return Enumerable.Range(infoResponseFSList.MinVersion, infoResponseFSList.MaxVersion - infoResponseFSList.MinVersion + 1);
-                case DiskStationApi.DSMInfo:
-                    return Enumerable.Range(infoResponseDSMInfo.MinVersion, infoResponseDSMInfo.MaxVersion - infoResponseDSMInfo.MinVersion + 1);
-                default:
-                    throw new DownloadClientException("Api not implemented");
+                if (apis.ContainsKey(data.Key))
+                {
+                    var key = GenerateInfoCacheKey(settings, apis[data.Key]);
+
+                    var info = _cacheInfo.Find(key);
```
Taloth: Don't check Find, just call Set. You'll want to refresh the data in the cache no matter what.

src/NzbDrone.Core/Download/Clients/DownloadStation/DiskStationApiInfoCache.cs
```diff
@@ -0,0 +1,15 @@
...
+using System.Linq;
+using System.Text;
+
+namespace NzbDrone.Core.Download.Clients.DownloadStation
+{
+    public class DiskStationApiInfoCache
+    {
+        public DiskStationApi Type { get; set; }
+        public string Name { get; set; }
+        public DiskStationApiInfo Info;
```
Taloth: Typo? Should be a property.

src/NzbDrone.Core/Download/Clients/DownloadStation/Proxies/DiskStationProxyBase.cs
```diff
@@ -40,7 +51,7 @@ static DiskStationProxyBase()
         {
             _httpClient = httpClient;
             _logger = logger;
-            _cacheInfo = cacheManager.GetCache<DiskStationApiInfo>(GetType());
+            _infoCache = cacheManager.GetCache<DiskStationApiInfoCache>(typeof(DiskStationApiInfoCache)); // with typeof(DiskStationProxyBase) throws an exception, i guess it cannot be an abstract class
```
Taloth: No, you need unique keys coz you have a conflict with _sessionCache.

```
cacheManager.GetCache<DiskStationApiInfoCache>(typeof(DiskStationProxyBase), "apiInfo");
cacheManager.GetCache<string>(typeof(DiskStationProxyBase), "sessions");
```

src/NzbDrone.Core/Download/Clients/DownloadStation/Proxies/DiskStationProxyBase.cs
```diff
@@ -151,14 +93,12 @@ private string GenerateSessionCacheKey(DownloadStationSettings settings)
 
                     if (responseContent.Error.SessionError)
                     {
-                        _cacheInfo.Remove(GenerateSessionCacheKey(settings));
+                        _infoCache.Remove(GenerateSessionCacheKey(settings));
```
Taloth: I overlooked this earlier. supposed to be _sessionCache, isn't it?

src/NzbDrone.Core/Download/Clients/DownloadStation/Proxies/DiskStationProxyBase.cs
```diff
@@ -172,40 +112,62 @@ private string GenerateSessionCacheKey(DownloadStationSettings settings)
...
-                 { "session", "DownloadStation" },
-             };
-
             var authInfo = GetApiInfo(DiskStationApi.Auth, settings);
 
-            var requestBuilder = BuildRequest(settings, authInfo, HttpMethod.GET);
+            var requestBuilder = BuildRequest(settings, authInfo, "login", 2);
+            requestBuilder.AddQueryParam("account", settings.Username);
+            requestBuilder.AddQueryParam("passwd", settings.Password);
+            requestBuilder.AddQueryParam("format", "cookie");
```
Taloth: cookie? wasn't this 'sid'?

PS: The use of AddQueryParam is exactly what I meant.

src/NzbDrone.Core/Download/Clients/DownloadStation/Proxies/DiskStationProxyBase.cs
```diff
@@ -172,40 +112,62 @@ private string GenerateSessionCacheKey(DownloadStationSettings settings)
...
-                 { "account", settings.Username },
-                 { "passwd", settings.Password },
-                 { "format", "sid" },
-                 { "session", "DownloadStation" },
-             };
-
             var authInfo = GetApiInfo(DiskStationApi.Auth, settings);
 
-            var requestBuilder = BuildRequest(settings, authInfo, HttpMethod.GET);
+            var requestBuilder = BuildRequest(settings, authInfo, "login", 2);
```
Taloth: On the left side of the compare you have version=1. Is that correct, or was version=2 already supported by DSM 5.x? (If v2 was supported then that's obviously the better option)

src/NzbDrone.Core/Download/Clients/DownloadStation/Proxies/DiskStationProxyBase.cs
```diff
@@ -172,40 +112,62 @@ private string GenerateSessionCacheKey(DownloadStationSettings settings)
...
+            requestBuilder.AddQueryParam("passwd", settings.Password);
+            requestBuilder.AddQueryParam("format", "cookie");
+            requestBuilder.AddQueryParam("session", "DownloadStation");
 
-            foreach (var arg in arguments)
-            {
-                requestBuilder.AddQueryParam(arg.Key, arg.Value);
-            }
+            var request = requestBuilder.Build();
+            request.StoreResponseCookie = true;
```
Taloth: It's my understanding that with sid we don't need the cookie.

src/NzbDrone.Core/Download/Clients/DownloadStation/Proxies/DiskStationProxyBase.cs
```diff
@@ -172,40 +112,62 @@ private string GenerateSessionCacheKey(DownloadStationSettings settings)
...
+        protected HttpRequestBuilder BuildRequest(DownloadStationSettings settings, string methodName, int methodVersion, HttpMethod httpVerb = HttpMethod.GET)
         {
-            var requestBuilder = new HttpRequestBuilder(settings.UseSsl, settings.Host, settings.Port).Resource($"webapi/{apiInfo.Path}");
+            var info = GetApiInfo(_apiType, settings);
+
+            return BuildRequest(settings, info, methodName, methodVersion, httpVerb);
+        }
 
-            requestBuilder.Method = method;
+        private HttpRequestBuilder BuildRequest(DownloadStationSettings settings, DiskStationApiInfoCache apiInfo, string methodName, int methodVersion, HttpMethod httpVerb = HttpMethod.GET)
```
Taloth: Nitpick: methodVersion -> apiVersion :wink:

src/NzbDrone.Core/Download/Clients/DownloadStation/Proxies/DiskStationProxyBase.cs
```diff
@@ -172,40 +112,62 @@ private string GenerateSessionCacheKey(DownloadStationSettings settings)
...
+        {
+            var requestBuilder = new HttpRequestBuilder(settings.UseSsl, settings.Host, settings.Port).Resource($"webapi/{apiInfo.Info.Path}");
+            requestBuilder.Method = httpVerb;
             requestBuilder.LogResponseContent = true;
             requestBuilder.SuppressHttpError = true;
             requestBuilder.AllowAutoRedirect = false;
 
+            if (httpVerb == HttpMethod.POST)
+            {
+                requestBuilder.Headers.ContentType = "multipart/form-data";
```
Taloth: AddForm* overrides ContentType, so no need to set it.

src/NzbDrone.Core/Download/Clients/DownloadStation/DiskStationApiInfoCache.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace NzbDrone.Core.Download.Clients.DownloadStation
+{
+    public class DiskStationApiInfoCache
```
Taloth: I'm wondering if we should just add Name, Type, and NeedsAuth directly to DiskStationApiInfo. That way it's entirely descriptive of the api.

src/NzbDrone.Core/Download/Clients/DownloadStation/Proxies/DownloadStationTaskProxy.cs
```diff
@@ -24,52 +24,39 @@ public DownloadStationTaskProxy(IHttpClient httpClient, ICacheManager cacheManag
...
-            {
-                { "version", "2" },
-                { "method", "create" }
-            };
+            var requestBuilder = BuildRequest(settings, "create", 2, HttpMethod.POST);
 
             if (downloadDirectory.IsNotNullOrWhiteSpace())
             {
-                arguments.Add("destination", downloadDirectory);
+                requestBuilder.AddQueryParam("destination", downloadDirectory);
```
Taloth: AddFormParameter?

src/NzbDrone.Core/Download/Clients/DownloadStation/Proxies/DownloadStationTaskProxy.cs
```diff
@@ -24,52 +24,39 @@ public DownloadStationTaskProxy(IHttpClient httpClient, ICacheManager cacheManag
...
         public void AddTaskFromUrl(string url, string downloadDirectory, DownloadStationSettings settings)
         {
-            var arguments = new Dictionary<string, object>
-            {
-                { "api", "SYNO.DownloadStation.Task" },
-                { "version", "3" },
-                { "method", "create" },
-                { "uri", url }
-            };
+            var requestBuilder = BuildRequest(settings, "create", 3);
```
Taloth: Wait a sec, create by file = v2 and create by url = v3?

src/NzbDrone.Core/Download/Clients/DownloadStation/Proxies/FileStationProxy.cs
```diff
@@ -34,15 +34,11 @@ public SharedFolderMapping GetSharedFolderMapping(string sharedFolder, DownloadS
...
-                { "version", "2" },
-                { "method", "getinfo" },
-                { "path", new [] { path }.ToJson() },
-                { "additional", $"[\"real_path\"]" }
-            };
-
-            var response = ProcessRequest<FileStationListResponse>(arguments, settings, $"get info of {path}");
+            var requestBuilder = BuildRequest(settings, "getinfo", 2);
+            requestBuilder.AddQueryParam("path", new[] { path }.ToJson());
+            requestBuilder.AddQueryParam("additional", $"[\"real_path\"]");
```
Taloth: Nitpick "$" not needed.

Taloth:
I like the changes, to me it seems more understandable.

Most of the comments in my recent review show 'outdated' but they still apply. You pushed new changes while I was reviewing :smile:

src/NzbDrone.Core/Download/Clients/DownloadStation/Proxies/DiskStationProxyBase.cs
margaale: Thanks for the tip

src/NzbDrone.Core/Download/Clients/DownloadStation/Proxies/DiskStationProxyBase.cs
margaale: Yep

src/NzbDrone.Core/Download/Clients/DownloadStation/DiskStationApiInfoCache.cs
margaale: Already fixed

src/NzbDrone.Core/Download/Clients/DownloadStation/DiskStationApiInfoCache.cs
margaale: I was wondering the same, lets discuss it later...

src/NzbDrone.Core/Download/Clients/DownloadStation/Proxies/DiskStationProxyBase.cs
margaale: The problem was with DSM.Info, Auth was working fine if i remember correctly

src/NzbDrone.Core/Download/Clients/DownloadStation/Proxies/DiskStationProxyBase.cs
margaale: i couldn't make it work with sid.

src/NzbDrone.Core/Download/Clients/DownloadStation/Proxies/DiskStationProxyBase.cs
margaale: That would be the case if i would have been able to get it right...

src/NzbDrone.Core/Download/Clients/DownloadStation/Proxies/DownloadStationTaskProxy.cs
margaale: Guess could be 3 both of them, but version 2 there was no way to specify uri.

margaale:
Yes, i'm very pleased as it is now. It seems to be working as before the changes. I was hesitating about adding those properties to ApiInfo or leave the new class, and i cannot decide yet...

src/NzbDrone.Core/Download/Clients/DownloadStation/Proxies/DiskStationProxyBase.cs
Taloth: 'Was working'? I'm asking because on master&develop you send v=1, see: https://github.com/Sonarr/Sonarr/blob/cf306f4aba051c0a8011dc03b047993297cef8e6/src/NzbDrone.Core/Download/Clients/DownloadStation/Proxies/DiskStationProxyBase.cs#L98

Taloth:
In Sabnzbd I do the same, I pull in a SabnzbdCategory object list. And then convert a Dir relativepath to populate a FullPath field.
I'm fine with either solution, but a single ApiInfo object seemed to be more useful.

If there's no way to get sid working, then we don't really need _sessionCache either anymore. But it might be worth trying another time, since it's better than fiddling with cookies.

src/NzbDrone.Core/Download/Clients/DownloadStation/Proxies/DiskStationProxyBase.cs
margaale: Sorry, you are right, I need version 2 in order to be able to use sid. According to documentation v2 is available since DSM 4.1. I wonder how many people have not been able to update theirs DiskStation...

Taloth: I don't think Sonarr was available on Syno for those earlier DSM versions, and if any tried, those devices would've likely been under-powered. 
Sounds like v2 is the right one to use.

margaale: But someone may want to use it as client only (sonarr running on other server). That said, i agree with you that we should use v2.

margaale:
As it is right now , _sessionCache is being used as a bool to check whether or not the user is authenticated. I'm still trying to get sid right, but there must to be something i'm not seeing. Same request, tried with postman and it worked while sonarr didn't.

Taloth:
Install Fiddler, you'll be able to see the http traffic accurately for postman and Sonarr.

margaale:
That's what i did. I see that they always return the cookie no matter what format i send 

Request: GET /webapi/auth.cgi?api=SYNO.API.Auth&version=2&method=login&account=admin&passwd=qwftyy654&session=DownloadStation&format=sid HTTP/1.1
Response: 
HTTP/1.1 200 OK
Server: nginx
Date: Sat, 18 Mar 2017 18:43:48 GMT
Content-Type: text/plain; charset="UTF-8"
Connection: keep-alive
Keep-Alive: timeout=20
Vary: Accept-Encoding
X-Content-Type-Options: nosniff
X-XSS-Protection: 1; mode=block
P3P: CP="IDC DSP COR ADM DEVi TAIi PSA PSD IVAi IVDi CONi HIS OUR IND CNT"
Set-Cookie: id=ne2Tpoh6yha9YAAH1N00944;expires=Sat, 25-Mar-2017 18:43:47 GMT;path=/
Content-Length: 58

{"data":{"sid":"ne2Tpoh6yha9YAAH1N00944"},"success":true}

Taloth:
During real requests you need `&_sid=` not `&sid=`, based on the webapi doc. That's probably what's causing the problem.

Sidenote: looks like during auth session= should be a random one. (`"session-" + Guid.NewGuid().ToString()` maybe)

margaale:
What doc says that? I mean the _sid parameter. 
The one i've got says: 
sid Authorized session ID. When the user log in with format=sid, cookie will not be set and each API request should provide a request parameter sid=<sid> along with other parameters.

margaale:
With _sid works fine. I hate them...

Taloth:
https://global.download.synology.com/download/Document/DeveloperGuide/Synology_Download_Station_Web_API.pdf
Page 6 "Making Requests"

margaale:
I was reading in the same document the chart in page 14. Thanks!!!

margaale:
Just did a rebase. Do you want me to squash this?

Taloth:
Squash it, but change the commit message to
`Fixed: DownloadStation api client for DSM 5.x.`

Taloth:
Not sure why github didn't do a fast-forward merge, but meh... it worked.

margaale:
No idea, i did another rebase before squashing...

PR #1754, opened 2017-03-11
Don't grab upgrades if existing file was deleted and will be unmonitored

Taloth:
We want to block upgrades, if the existing file was already deleted. But Sonarr isn't aware of that till the next disk scan.
This PR adds a Specification that will check if the file exists.

For performance reasons this Specification can only run if no other rejections exist. So I refactored the specifications to use a tiered approach. If any rejection exist in the current tier, the the next one won't be evaluated, thus avoiding the disk IO.

I haven't tested it, nor are there any unit tests. First I want to know how happy we are with this approach.

### DIFF
--- a/src/NzbDrone.Core.Test/DecisionEngineTests/DownloadDecisionMakerFixture.cs
+++ b/src/NzbDrone.Core.Test/DecisionEngineTests/DownloadDecisionMakerFixture.cs
@@ -28,6 +28,8 @@ namespace NzbDrone.Core.Test.DecisionEngineTests
         private Mock<IDecisionEngineSpecification> _fail2;
         private Mock<IDecisionEngineSpecification> _fail3;
 
+        private Mock<IDecisionEngineSpecification> _failDelayed1;
+
         [SetUp]
         public void Setup()
         {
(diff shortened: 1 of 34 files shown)


markus101:
I like this approach, I think it fits well with Manual Search and wanting to show all the reasons something was rejected since we'd never hit it during a search anyways. I say let's do it and save people from wanting to rescan their disk more often because the current rescan interval doesn't always work out.

Taloth:
Hm, does that mean I should run ALL decision specs if it was a manual search?

markus101:
Oh sorry, I was thinking no (so the behaviour you have here makes sense). Maybe in a future version we could have decisions that were tested and rejected, but were ignored because it was a search, but we still can show it as a warning (which would give more visibility why something wasn't grabbed via RSS), but I think that's a separate problem.

Taloth:
@markus101 Added tests, surprisingly they're all green.

I'm not sure whether I should cache the FileExists result for 5 second, in case it has to evaluate multiple releases (or more likely, the same release from multiple indexers). The reason why I don't is because we already hit the disk, so the FS driver should already have cached it after the first call.

Issue #621, opened 2015-06-15
[Feature] Minimum seeders


d2dyno:
Hello,

Unless I'm blind, I cannot find a way to only allow torrents with x amount of seeders to be downloaded. If this is not already a feature, would it be possible to add it? I know just today, torrents with 1/0 seeders were being added instead of ones with 24/5 seeders. I was forced to manually correct this.

Anyways, let me know :)

markus101:
We don't plan to expose a configurable minimum, but do plan to make the selection within the same quality smarter, assuming these are the same quality.

vinchi007:
Please expose "min seed" option so that users can change it at will. There are so many older backlog series that have few seeds - current automatic select does not pick the highest seed count.

meffect:
i get lots of 0 seeder torrents

Kyrluckechuck:
Completely agreeing with this -- it's needed. There are far too many shows that keep getting put into my queue with 0-2 seeders where if it's more than 0, they usually flake out before I even have 1% of it done.. 

I keep needing to remove them manually and go find them since I can't know from the WebUI if it's a decently seeded file

meffect:
I think issue still cause what happens is my torrent client is behind blocklists. my pvr software may or may not be behind the same blocklists depending. so while the pvr software thinks it's being smart, it doesn't know that their may be 4 out of 5 seeds getting blocked from my blocklists. likewise, the person on the other end may be blocking me. i'm sure their's various other conditions that cause the tracker reported seeds to not be accurate. just adding a pov maybe not considered yet.

it would be nice to set a minimum so I can add my own educated buffer of seeds I think I wont be able to connect with. even nicer would be sonarr evaluating torrent state through acquisition process and making decisions.

donisewell:
Would love a configurable minimum. On private trackers I don't really need to worry about it having 0 seeders. All this is currently doing for me is making me "late" to torrents and cutting the amount of time I could have been seeding to others.

cmgraz:
Is there a reason that this isn't exposed?  Without it, I'm forced to go through and do quite a few things manually, especially for older seasons.

What if I'm looking at a show that has 1 seeder at 720p, but 30 seeders at 480p?  I'd prefer to take the lower quality than have something sit in Deluge and not actually download.

What's the negative recourse for exposing some sort of parameter that allows a user to select minimum # of seeders?

meffect:
any response from dev's?

wdaniel526:
[...]

PR #1763, opened 2017-03-15
Ability to set minimum seeders on a per indexer basis

markus101:
#### Database Migration
NO

#### Description
Adds a setting to torrent indexers to specify the minimum number of seeders on a per indexer basis. Currently this only adds the setting to RARBG, but could be added to others. Instead of a setting we could do this as a prop on the indexer itself (similar to enable search/name).

#### Todos
- The rest of the torrent indexers

* https://github.com/Sonarr/Sonarr/issues/621


src/NzbDrone.Core/DecisionEngine/Specifications/Search/TorrentSeedingSpecification.cs
```diff
@@ -25,10 +30,14 @@ public Decision IsSatisfiedBy(RemoteEpisode remoteEpisode, SearchCriteriaBase se
                 return Decision.Accept();
             }
 
-            if (torrentInfo.Seeders != null && torrentInfo.Seeders < 1)
+            var indexer = _indexerFactory.Get(torrentInfo.IndexerId);
+            var minimumSeedersProperty = indexer.Settings.GetType().GetSimpleProperties().SingleOrDefault(s => s.Name == "MinimumSeeders");
+            var minimumSeeders = minimumSeedersProperty == null ? 1 : (int)minimumSeedersProperty.GetValue(indexer.Settings, null);
```
markus101: I'm not very happy with this, if `MinimumSeeders` was a property on the indexer this could be avoided, but would be extraneous for Usenet indexers (though not really causing harm).

Taloth:
At least you should use an interface and cast, instead of reflection.

Adding a decision logic config option to specific indexers? That's something we've never done before.... 😲

So why not just add a global setting?
I know you want to avoid the inevitable "but I want to set it only for this indexer" support question, but we were already extremely reluctant to add min seeds, I see no need to add 5 separate config options that can be tweaked individually.

markus101:
Yeah, decision logic based on the specific indexer has never happened before.

I was torn on global vs per indexer and went per indexer to avoid the support questions, but also because you could want different rules for different indexers, like one user pointed out on a private tracker 0 seeds is fine to jump on something early, but for a public tracker you may want 10 or more (to deal with block lists, regardless of their helpfulness). I feel less reluctant to do it at the indexer level than the global level because it feels like a more complete option over a global option which is there just to stop the questions about why it isn't there.

LSL1337:
so what did you guys decide?

in my opinion a global filter is ok.

if somebody has private trackers, they want a 0 setting (maybe only for rss sync, but if i do a manual search, i can see the seeders anyway).
if somebody is not on private trackers, they want the current setting.
i'm not sure how big is the overlap between the two camps.

i'm sure there is some overlap (meaning guys with private access using public trackers), but as a first version, the global setting would be an improvement (if the per indexer setting is too much of a headache)

what's the current limit for torrent btw? i didn't even know there is a minimum? Is it just 1?
thanks

markus101:
I think if we think we may want a per indexer setting it makes sense to do it. Personally I'm still in favour of a per indexer setting, it's a bit more verbose, but it' more flexible and does eliminate the inevitable questions about it and the compromises that may need to be made.

> what's the current limit for torrent btw? i didn't even know there is a minimum? Is it just 1?

Yeah, it's a hardcoded minimum of 1.

src/NzbDrone.Core/Indexers/BitMeTv/BitMeTvSettings.cs
```diff
@@ -23,13 +23,14 @@ public BitMeTvSettingsValidator()
...
 
-    public class BitMeTvSettings : IProviderConfig
+    public class BitMeTvSettings : IProviderConfig, ITorrentIndexerSettings
     {
         private static readonly BitMeTvSettingsValidator Validator = new BitMeTvSettingsValidator();
 
         public BitMeTvSettings()
         {
             BaseUrl = "https://www.bitmetv.org";
+            MinimumSeeders = 1;
```
Taloth: Should we put this in a constant somewhere? coz it's repeated a dozen times.

src/NzbDrone.Core/Indexers/Torznab/TorznabSettings.cs
```diff
@@ -46,10 +47,18 @@ public TorznabSettingsValidator()
...
+    public class TorznabSettings : NewznabSettings, ITorrentIndexerSettings
     {
         private static readonly TorznabSettingsValidator Validator = new TorznabSettingsValidator();
 
+        public TorznabSettings()
+        {
+            MinimumSeeders = 1;
+        }
+
+        [FieldDefinition(5, Type = FieldType.Textbox, Label = "Minimum Seeders", HelpText = "Minimum number of seeders required.", Advanced = true)]
```
Taloth: I suggest we add a comment to NewznabSettings that Field 5 is in use, to avoid conflicts in the future.

francesco2013:
I totally agree this option is a must have. So many torrents start and then they die after a bit because the number of seeders is so little. It would be really helpful to have this ability and to set the minimum seeders number IMHO it should be at least 10.

PR #1773, opened 2017-03-17
Use MaterialisingResponse for static resource responses

markus101:
#### Database Migration
NO

#### Description
Wraps responses for static resources in [MaterialisingResponse](https://github.com/NancyFx/Nancy/blob/master/src/Nancy/Responses/MaterialisingResponse.cs) so the response is properly calculated before returning to the client and the logs should the correct response (and `OnError` handling works in the pipeline).

PRed such a simple change because there is a warning on `MaterialisingResponse` and while our responses aren't for giant files I wanted to play it safe.

### DIFF
--- a/src/NzbDrone.Api/Frontend/Mappers/StaticResourceMapperBase.cs
+++ b/src/NzbDrone.Api/Frontend/Mappers/StaticResourceMapperBase.cs
@@ -1,4 +1,4 @@
-using System;
+﻿using System;
 using System.IO;
 using NLog;
 using Nancy;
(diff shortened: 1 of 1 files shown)


Taloth:
We already load the entire logfile into memory when returning it, but that's incidental.
However, streaming static resources directly probably has better performance, especially on RPi. Not by much.

What reason do we have for needing this?

markus101:
Right now if `index.html` is requested we process the request, log a 200 response in the logs and start streaming it back, but if that fails for some reason (for example a css file is missing) the request fails and the client receives a 500. When asking users for logs the server thinks everything was fine but the browser either shows a generic 500 error page or nothing (depends on the browser). The client doesn't actually see the underlying reason why the request failed (though that could be remedied by handling errors within the stream response, but our logs would still show the original 200 response.

Both:

https://forums.sonarr.tv/t/white-or-blank-screen-after-update/14043 and https://forums.sonarr.tv/t/form-authentication-leads-to-blank-page/14004 are recent examples of the issue.

Taloth:
Are you sure this change actually addresses that issue?

Im asking coz once we're returning a stream for index.html then that's it, the file exists and is readable.
css files listed in index.html are a separate request and wouldn't cause the index.html request to fail.
(I know it was an example, but it simply doesn't apply)

You're free to push this to develop, I don't see any harm in it, but I doubt it has the desired effect.

Issue #1786, opened 2017-03-22
DownloadStation authentication just started failing


demonbane:
DSM Version: DSM 6.1-15047 Update 1
Sonarr Version: 2.0.0.4668

This may have started as early as .4654 btw, I just haven't checked Sonarr in a few days and the auto-updates have been happening behind the scenes.

This is an issue that I brought up in #1784 before realizing that it was my mistake and the two actually aren't related. The error, however, is the same.

```
[v2.0.0.4668] NzbDrone.Core.Download.Clients.DownloadClientException: Failed to login. Reason: Permission denied
  at NzbDrone.Core.Download.Clients.DownloadStation.Proxies.DiskStationProxyBase.ProcessRequest[T] (NzbDrone.Common.Http.HttpRequestBuilder requestBuilder, System.String operation, DiskStationApi api, NzbDrone.Core.Download.Clients.DownloadStation.DownloadStationSettings settings) <0x41ea39f0 + 0x00287> in <filename unknown>:0 
  at NzbDrone.Core.Download.Clients.DownloadStation.Proxies.DiskStationProxyBase.AuthenticateClient (NzbDrone.Core.Download.Clients.DownloadStation.DownloadStationSettings settings) <0x41f782f0 + 0x00143> in <filename unknown>:0 
  at NzbDrone.Core.Download.Clients.DownloadStation.Proxies.DiskStationProxyBase+<>c__DisplayClass14_0.<BuildRequest>b__1 () <0x41f782c0 + 0x00017> in <filename unknown>:0 
  at NzbDrone.Common.Cache.Cached`1[T].Get (System.String key, System.Func`1 function, Nullable`1 lifeTime) <0x41d6c5a0 + 0x0021e> in <filename unknown>:0 
  at NzbDrone.Core.Download.Clients.DownloadStation.Proxies.DiskStationProxyBase.BuildRequest (NzbDrone.Core.Download.Clients.DownloadStation.DownloadStationSettings settings, NzbDrone.Core.Download.Clients.DownloadStation.DiskStationApiInfo apiInfo, System.String methodName, Int32 apiVersion, HttpMethod httpVerb) <0x41ea1720 + 0x0053e> in <filename unknown>:0 
[...]

Issue #1784, opened 2017-03-21
Download Station client - Permission Denied with 2-factor authentication


nswali:
**Synology: DSM 6.1-15047 Update 1
Sonaar: 2.0.0.4645**

I'm having trouble adding the Download Station download client; I'm getting the following permission denied exception:

NzbDrone.Core.Download.Clients.DownloadClientAuthenticationException: Permission denied
  at NzbDrone.Core.Download.Clients.DownloadStation.Proxies.DiskStationProxyBase.AuthenticateClient (NzbDrone.Core.Download.Clients.DownloadStation.DownloadStationSettings settings) [0x000d3] in M:\BuildAgent\work\b69c1fe19bfc2c38\src\NzbDrone.Core\Download\Clients\DownloadStation\Proxies\DiskStationProxyBase.cs:127 
  at NzbDrone.Core.Download.Clients.DownloadStation.Proxies.DiskStationProxyBase.ProcessRequest[T] (NzbDrone.Core.Download.Clients.DownloadStation.DiskStationApi api, System.Collections.Generic.Dictionary`2[TKey,TValue] arguments, NzbDrone.Core.Download.Clients.DownloadStation.DownloadStationSettings settings, System.String operation, NzbDrone.Common.Http.HttpMethod method, System.Int32 retries) [0x0003c] in M:\BuildAgent\work\b69c1fe19bfc2c38\src\NzbDrone.Core\Download\Clients\DownloadStation\Proxies\DiskStationProxyBase.cs:57 
  at NzbDrone.Core.Download.Clients.DownloadStation.Proxies.DownloadStationProxy.GetTasks (NzbDrone.Core.Download.Clients.DownloadStation.DownloadStationSettings settings) [0x00046] in M:\BuildAgent\work\b69c1fe19bfc2c38\src\NzbDrone.Core\Download\Clients\DownloadStation\Proxies\DownloadStationProxy.cs:77 : Permission denied

I have 2-factor authentication switched on, could this be a cause?

I have Transmission and SABnzbd configured and these work, but would like to replace both with Download Station.

margaale:
Hi, there is no current support for that type of authentication yet.

nswali:
OK. Can this be added as an enhancement?

margaale:
How often does the code change? If it changes too frequently is not going to wor, cause sonarr needs always to be connected with the client and you will end up with a lot of authentication errors in the logs...

demonbane:
[...]

PR #1790, opened 2017-03-23
DownloadStation - Revert Session name

margaale:
#### Database Migration
NO

#### Description
As it seems guid is causing some troubles i think we should get back to "DownloadStation" for the session name. Just tried here with filestation api and it's working fine.

* #1786, #1784


Taloth:
But do we want to just push this without knowing why ppl get different results?

margaale:
No, but can we know what is causing it? I would make a branch for him to
test

demonbane:
I'd be happy to try a different branch if you want me to test it first.

Taloth:
Good idea, I've pushed it to the `revert-downloadstation` branch, it's being built now, should be ready in 15-30min or so.

Taloth:
Released, branch `revert-downloadstation` (you can switch to it just like master&develop), version 4670.

demonbane:
Looks good. Connection test worked fine, and I was able to successfully add a download from my wanted list.

Taloth:
Merged it. FYI, the system will push you back to develop automatically.

Issue #1782, opened 2017-03-20
Deleting Episode Files that do not exist fails


stuntguy3000:
Show List: (Directory was manually deleted)
![screenshot_2](https://cloud.githubusercontent.com/assets/1522389/24097748/24eb2fd6-0db7-11e7-8926-be4f268d8c65.png)

HTTP Responses when deleting files:
![screenshot_1](https://cloud.githubusercontent.com/assets/1522389/24097758/2d02cbca-0db7-11e7-959d-659d859a37e9.png)

Whilst this was fixed by performing a disk rescan, this error should be resolved.

markus101:
What do the logs show for this scenario?
What response are you expecting here?

If the series folder is on an unmounted you'd get the same behaviour and it wouldn't make sense to remove Sonarr's record of it. It shouldn't return a 500 response, but whether it's successful or not will depend on whether Sonarr thinks the series folder is likely on an unmounted volume (root folder is missing/empty) or just that series folder is missing and the records should be cleaned up.

stuntguy3000:
"What response are you expecting here?"
What kind of a question is this? The UI says the files are deleted except that's not the case.

"...and it wouldn't make sense to remove Sonarr's record of it."
If the user is attempting to delete the files and the file operation fails because the folder can't be found, then they should **absolutely** be removed from Sonarr.

"If the series folder is on an unmounted you'd get the same behaviour"
If the series folder is unmounted, it should say so and not let operations be performed.

"What do the logs show for this scenario?"
I don't have any, but in this scenario logs are irrelevant.

markus101:
> "What response are you expecting here?"
What kind of a question is this? The UI says the files are deleted except that's not the case.

That's an important distinction vs Sonarr returned and nothing happened. The only information you provided was that there was a 500 response and it shouldn't.

> "...and it wouldn't make sense to remove Sonarr's record of it."
If the user is attempting to delete the files and the file operation fails because the folder can't be found, then they should absolutely be removed from Sonarr.

Again, not if the series is on an unmounted share/volume.

>"What do the logs show for this scenario?"
I don't have any.

Logs would be helpful to see exactly where it failed, though in this case it should be easy to reproduce.

stuntguy3000:
> Again, not if the series is on an unmounted share/volume.

How realistic, common scenario though? _Hint:_ It's not.

> That's an important distinction vs Sonarr returned and nothing happened. The only information you provided was that there was a 500 response and it shouldn't.

[...]

PR #1800, opened 2017-03-26
Deleting Episode Files that no longer exist on disk

markus101:
#### Database Migration
NO

#### Description
This improves the deleting of episode files and a small refactor to deleting episode files after a series is deleted.

There are a few rules (consistent with rescanning a series folder):

- If the root folder for a series is missing or empty the file will not be deleted from the DB
- If only the series folder or the episode file is missing the file will be deleted from the DB
- If the file exists on disk we'll delete it from disk and then the DB

* #1782

### DIFF
--- a/src/NzbDrone.Core.Test/MediaFiles/MediaFileDeletionService/DeleteEpisodeFileFixture.cs
+++ b/src/NzbDrone.Core.Test/MediaFiles/MediaFileDeletionService/DeleteEpisodeFileFixture.cs
@@ -1,0 +1,140 @@
+﻿using System.IO;
+using FizzWare.NBuilder;
+using Moq;
+using NUnit.Framework;
+using NzbDrone.Common.Disk;
+using NzbDrone.Core.Exceptions;
+using NzbDrone.Core.MediaFiles;
+using NzbDrone.Core.Test.Framework;
+using NzbDrone.Core.Tv;
+using NzbDrone.Test.Common;
+
+namespace NzbDrone.Core.Test.MediaFiles.MediaFileDeletionService
+{
+    [TestFixture]
+    public class DeleteEpisodeFileFixture : CoreTest<Core.MediaFiles.MediaFileDeletionService>
+    {
+        private static readonly string RootFolder = @"C:\Test\TV";
+        private Series _series;
+        private EpisodeFile _episodeFile;
+
+        [SetUp]
+        public void Setup()
+        {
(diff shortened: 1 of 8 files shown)


src/NzbDrone.Core/MediaFiles/MediaFileDeletionService.cs
```diff
@@ -0,0 +1,76 @@
...
+        }
+
+        public void DeleteEpisodeFile(Series series, EpisodeFile episodeFile)
+        {
+            var fullPath = Path.Combine(series.Path, episodeFile.RelativePath);
+            var rootFolder = _diskProvider.GetParentFolder(series.Path);
+
+            if (!_diskProvider.FolderExists(rootFolder))
+            {
+                throw new NzbDroneClientException(HttpStatusCode.NotFound, "Series' root folder ({0}) doesn't exist.", rootFolder);
```
markus101: If you have a better suggestion for the status code, I'll take it.

Taloth: If the episodefile doesn't exist in the db at all then 404. Otherwise I'm leaning to Conflict, we already use it in the ReleaseModule if adding the release failed. But we need something that shows up on UI as a proper error notification.

src/NzbDrone.Core.Test/MediaFiles/MediaFileDeletionService/DeleteEpisodeFileFixture.cs
```diff
@@ -110,6 +113,29 @@ public void should_delete_from_disk_and_db_if_episode_file_exists()
...
+
+            Mocker.GetMock<IRecycleBinProvider>()
+                  .Setup(s => s.DeleteFile(_episodeFile.Path, "Series Title"))
+                  .Throws(new IOException());
+
+            Assert.Throws<NzbDroneClientException>(() => Subject.DeleteEpisodeFile(_series, _episodeFile));
+
+            ExceptionVerification.ExpectedErrors(1);
+            Mocker.GetMock<IRecycleBinProvider>().Verify(v => v.DeleteFile(_episodeFile.Path, "Series Title"), Times.Once());
+            Mocker.GetMock<IRecycleBinProvider>().Verify(v => v.DeleteFile(_episodeFile.Path, It.IsAny<string>()), Times.Once());
```
Taloth: Missing assertion verifying that the db entry isn't removed?

PR #1880, opened 2017-04-29
use cleancss for minification

drewfreyling:
#### Database Migration
NO

#### Description
Use a better css compressor. We save 3KB overall. CleanCSS performs well -> https://goalsmashers.github.io/css-minification-benchmark/

### DIFF
--- a/gulp/less.js
+++ b/gulp/less.js
@@ -5,7 +5,7 @@ var postcss = require('gulp-postcss');
 var sourcemaps = require('gulp-sourcemaps');
 var autoprefixer = require('autoprefixer-core');
 var livereload = require('gulp-livereload');
-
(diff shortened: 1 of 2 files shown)


gulp/less.js
```diff
@@ -5,7 +5,7 @@ var postcss = require('gulp-postcss');
 var sourcemaps = require('gulp-sourcemaps');
 var autoprefixer = require('autoprefixer-core');
 var livereload = require('gulp-livereload');
-
+var  cleanCSS = require('gulp-clean-css');
```
MindTooth: Extra space? :-)

drewfreyling: Fixed. Also changed casing to match `postcss`

PR #1884, opened 2017-04-29
remove redundant IE meta tag as we use http header instead

drewfreyling:
#### Database Migration
NO

#### Description
Since we already use the recommended way of sending down the `X-UA-Compatible` http header we don't need the meta tag as well.

#### Todos
- [ ] Documentation about browser support?

### DIFF
--- a/src/UI/index.html
+++ b/src/UI/index.html
@@ -2,7 +2,6 @@
 <html>
 <head>
     <title>Sonarr</title>
-    <meta http-equiv="X-UA-Compatible" content="IE=edge"/>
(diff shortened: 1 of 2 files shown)


thezoggy:
read: https://github.com/h5bp/html5-boilerplate/blob/master/dist/doc/html.md#x-ua-compatible

drewfreyling:
@thezoggy yup that is what i've done - "If possible, we recommend that you remove the meta tag and send only the HTTP response header". Am I missing something here?

Issue #1867, opened 2017-04-24
Health Checks on mount points


kyse:
Background: I noticed when messing with Ubuntu as a vm inside ESXi, using SATA controller instead of SCSI controller when using sata drives I exposed to the VM as a raw device mapping was a mistake.  The smart info wasn't available for the drives using the SATA controller, but worked fine with SCSI controller, and also I experienced errors in Sonarr once in a while where an import failed.  Obviously checking the logs it became apparent the access to path denied error wasn't so much related to permissions, but more so because of some error on the mount causing it to remount as read only.

So, I downloaded the source, and have been playing around with the health checker code, and grabbing series paths to determine mount points of interest to be monitored.  I figure mixing this with the procmount info, could we make the error message for a failed import a little more obvious for this scenario?

Or even a status alert that a mount point used in one of the existing series paths is mounted as read only.

Thoughts?  Is this something I should continue working towards a PR for you all?  Pretty sure I already fixed up my issue with my mount remounting read-only when I switched back to the SCSI controller, but figure this could be helpful during an issue in the future one day when drives start failing.

kyse:
I was able to get this working.

Exposed the mount options in the underlying mono IMount interface implementation.  I ran into a problem at first because certain mounts (mainly the actual mounts one would probably care about) weren't having their mountoptions value passed in to the final list returned from the service.  This was due to the list being built by first grabbing disks from the System.IO library, then concating the list with any additional mounts found from the mono implementation.  Reversing this fixed that problem, and I was able to successfully parse mount options for rw, or ro to determine how the mount was mounted.  In the health check, comparing a distinct list of series mounts (cause why do we care about mounts series aren't targeting) to find any ro mounts resulted in the desired warning output.

I noticed a list being build in one of the services ignores mounts with a path of /var/lib saying folks wouldn't care about anything mounted here, but for my case, I have one drive mounted under /var/lib/plexmediaserver/... because that's the home dir of the plex user.  So just a thought, that if block might want to be rethought.

markus101:
Improving messaging to users is definitely useful. This seems like a worthy improvement and we'd be interested in a PR for it.

> This was due to the list being built by first grabbing disks from the System.IO library, then concating the list with any additional mounts found from the mono implementation.

This is intentional to prefer the information in mono over the information in proc/mounts in the event we're parsing something wrong (or something similar), but perhaps a better middle-ground would be to take the information from mono and add in anything else provided by parsing proc/mount.

[...]

PR #1903, opened 2017-05-05
Health Check - Mount Read-Only

kyse:
#### Database Migration
NO

#### Description
Adds a health check to determine if any series targets a mount that is currently read-only.

#### Todos
- [ ] Tests
- [ ] Documentation

* #1867

### DIFF
--- a/src/NzbDrone.Common/Disk/DriveInfoMount.cs
+++ b/src/NzbDrone.Common/Disk/DriveInfoMount.cs
@@ -1,4 +1,5 @@
-﻿using System.IO;
+﻿using System.Collections.Generic;
+using System.IO;
 using NzbDrone.Common.Extensions;
 
 namespace NzbDrone.Common.Disk
@@ -8,10 +9,11 @@ namespace NzbDrone.Common.Disk
         private readonly DriveInfo _driveInfo;
         private readonly DriveType _driveType;
 
-        public DriveInfoMount(DriveInfo driveInfo, DriveType driveType = DriveType.Unknown)
+        public DriveInfoMount(DriveInfo driveInfo, DriveType driveType = DriveType.Unknown, Dictionary<string,string> options = null)
         {
             _driveInfo = driveInfo;
             _driveType = driveType;
+            MountOptions = options;
         }
 
         public long AvailableFreeSpace => _driveInfo.AvailableFreeSpace;
@@ -32,6 +34,8 @@ namespace NzbDrone.Common.Disk
         }
 
         public bool IsReady => _driveInfo.IsReady;
+
+        public Dictionary<string, string> MountOptions { get; private set; }
 
         public string Name => _driveInfo.Name;
 
--- a/src/NzbDrone.Common/Disk/DriveInfoMount.cs
+++ b/src/NzbDrone.Common/Disk/DriveInfoMount.cs
@@ -9,11 +9,11 @@ namespace NzbDrone.Common.Disk
         private readonly DriveInfo _driveInfo;
         private readonly DriveType _driveType;
 
-        public DriveInfoMount(DriveInfo driveInfo, DriveType driveType = DriveType.Unknown, Dictionary<string,string> options = null)
+        public DriveInfoMount(DriveInfo driveInfo, DriveType driveType = DriveType.Unknown, MountOptions mountOptions = null)
         {
             _driveInfo = driveInfo;
             _driveType = driveType;
-            MountOptions = options;
+            MountOptions = mountOptions;
         }
 
         public long AvailableFreeSpace => _driveInfo.AvailableFreeSpace;
@@ -35,7 +35,7 @@ namespace NzbDrone.Common.Disk
 
         public bool IsReady => _driveInfo.IsReady;
 
-        public Dictionary<string, string> MountOptions { get; private set; }
+        public MountOptions MountOptions { get; private set; }
 
         public string Name => _driveInfo.Name;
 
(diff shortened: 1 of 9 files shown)


src/NzbDrone.Common/Disk/DriveInfoMount.cs
```diff
@@ -8,10 +9,11 @@ public class DriveInfoMount : IMount
         private readonly DriveInfo _driveInfo;
         private readonly DriveType _driveType;
 
-        public DriveInfoMount(DriveInfo driveInfo, DriveType driveType = DriveType.Unknown)
+        public DriveInfoMount(DriveInfo driveInfo, DriveType driveType = DriveType.Unknown, Dictionary<string,string> options = null)
```
markus101: A dictionary or any enumerable type shouldn't be null, so you don't need to do a null check on it, so we should have a separate constructor or make it required.

markus101: Also: `Dictionary<string, string>` (with a space).

src/NzbDrone.Common/Disk/IMount.cs
```diff
@@ -8,6 +9,7 @@ public interface IMount
         string DriveFormat { get; }
         DriveType DriveType { get; }
         bool IsReady { get; }
+        Dictionary<string,string> MountOptions { get; }
```
markus101: `Dictionary<string, string>` (with a space) here as well.

src/NzbDrone.Mono/Disk/DiskProvider.cs
```diff
@@ -86,11 +86,13 @@ public override void SetPermissions(string path, string mask, string user, strin
...
         {
-            return GetDriveInfoMounts().Select(d => new DriveInfoMount(d, FindDriveType.Find(d.DriveFormat)))
-                                       .Where(d => d.DriveType == DriveType.Fixed || d.DriveType == DriveType.Network || d.DriveType == DriveType.Removable)
-                                       .Concat(_procMountProvider.GetMounts())
-                                       .DistinctBy(v => v.RootDirectory)
-                                       .ToList();
+
+            var procMounts = _procMountProvider.GetMounts();
+            return GetDriveInfoMounts().Join(procMounts, d => d.RootDirectory.FullName, m => m.RootDirectory, (d, m) => new DriveInfoMount(d, FindDriveType.Find(d.DriveFormat), m.MountOptions))
+                .Where(d => d.DriveType == DriveType.Fixed || d.DriveType == DriveType.Network || d.DriveType == DriveType.Removable)
```
markus101: Can we align this on the `.` as well?

src/NzbDrone.Mono/Disk/DiskProvider.cs
```diff
@@ -86,11 +86,13 @@ public override void SetPermissions(string path, string mask, string user, strin
 
         public override List<IMount> GetMounts()
         {
-            return GetDriveInfoMounts().Select(d => new DriveInfoMount(d, FindDriveType.Find(d.DriveFormat)))
-                                       .Where(d => d.DriveType == DriveType.Fixed || d.DriveType == DriveType.Network || d.DriveType == DriveType.Removable)
-                                       .Concat(_procMountProvider.GetMounts())
-                                       .DistinctBy(v => v.RootDirectory)
-                                       .ToList();
+
+            var procMounts = _procMountProvider.GetMounts();
```
markus101: New line after instead of before.

Taloth:
First of, add a MountOptions class (or something like it). Give that an IsReadOnly property. Have that check the internal dictionary for 'ro'.
That way you don't give the HealthCheck knownledge about those internals AND you can use the flag elsewhere.
Bonus is that it gets rid of the nullable dictionary.

Also, GetMounts is now broken. Imagine ProcMounts gives an empty list as result. What will happen with GetMounts? Reason is coz Join is an InnerJoin.

kyse:
I was using the null to know if the dictionary came from /proc/mounts or not.  Simply checking for a ro key in the dictionary whether contained in a class or directly would result in drives returned from System.IO library appearing as rw by default in a windows env.  If you're fine with using such an assumption, I'll run with it.  I just wanted a sure way to only pull up warnings/errors if the details retrieved from the dictionary actually came from /proc/mounts, thus the null optional parameter.

Taloth:
You can still have MountOptions nullable, I just wanted to be sure the internals are hidden and that the HealthCheck knows nothing about fstab mount options.

src/NzbDrone.Core/HealthCheck/Checks/MountCheck.cs
```diff
@@ -0,0 +1,35 @@
...
+        {
+            _diskProvider = diskProvider;
+            _seriesService = seriesService;
+        }
+
+        public override HealthCheck Check()
+        {
+            var mounts = _seriesService.GetAllSeries()
+                                       .Select(series => _diskProvider.GetMount(series.Path))
+                                       .DistinctBy(m => m.RootDirectory)
```
Taloth: Move DistinctBy one line up: `.DistinctBy(m => m.Path)`. That way you don't call GetMount per series, but only per Path.

markus101: You'd need to select all root folders for the series first since every series path should be unique.

```
.Select(s => s.Path.GetParentFolder())
.Where(m => m != null)
.Distinct()
```

The `Where` being a safety check to ensure it's not null (in case someone added a series path at the top of a drive/root, _shudders_).

Taloth: Already discussed it on IRC, we can't use Distinct coz we need mono to do the SymLinkResolver per series path to get the right mount. So we opted to just add a comment to explain why. It's not like this code gets call often.

PR #1907, opened 2017-05-05
Added ability to filter scene mappings by regex via services.

Taloth:
#### Database Migration
YES: 112

#### Description

- First is adding a FilterRegex field to SceneMapping. If it's not-null then the regex is used against the releaseTitle. In order to make that happen I needed to make sure the releaseTitle was passed through from the various places.
Along the way did some cleanup (SeriesService no longer has a ref to SceneMapping, ParsingService does that.)

- Second is dealing with potential duplicates. Atm I throw if there are conflicting scenemappings. Although one with regex will never conflict with one without regex.
The idea is that if you add a mapping to services that for example maps a scene name without year to a tvdb name with year, that you'll ALSO add the original non-year tvdb series 'mapping' unless the original has no scene releases. That way Sonarr is aware that there's a dupe. This means that a regex is basically mandatory in those cases.
The TODO is to turn this into a DecisionSpec as well, so we don't throw inside the parsing service.

There's a little bit of risk involved since I did change some stuff and might not have updated the title vs seriesTitle vs releaseTitle stuff correctly. So it will need some testing.

This requires a change in services too, but I got that ready, just a small commit to add the field.

_This change incorporates an idea I had for TheXEMv2. What TheXEMv2 is about is mapping multiple scene variants to tvdb by using regex and multi-mapping tables instead of just one 'SceneSeasonNumber' fields. 
The idea was that TheXEM/skyhook would know which potential conflicts exist. skyhook could then product a list of 'global mappings', which map to the correct series and prevents series level conflicts (this sometimes requires numbering so isn't just our scenemapping api, but a filtered thexem). And detailed 'local mappings', which map between multiple scene variants to tvdb numbering for individual series, which would only be loaded if that particular series is actually added._

### DIFF
--- a/src/NzbDrone.Core.Test/DataAugmentation/Scene/SceneMappingServiceFixture.cs
+++ b/src/NzbDrone.Core.Test/DataAugmentation/Scene/SceneMappingServiceFixture.cs
@@ -20,11 +20,14 @@ namespace NzbDrone.Core.Test.DataAugmentation.Scene
 
         private Mock<ISceneMappingProvider> _provider1;
         private Mock<ISceneMappingProvider> _provider2;
-            
+
         [SetUp]
         public void Setup()
         {
-            _fakeMappings = Builder<SceneMapping>.CreateListOfSize(5).BuildListOfNew();
+            _fakeMappings = Builder<SceneMapping>.CreateListOfSize(5)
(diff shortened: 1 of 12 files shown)


Taloth:
image

PR #1910, opened 2017-05-06
Skyhook notifications and url substitutes

Taloth:
#### Database Migration
YES 113 (for now, I'd like to merge that in separately)

#### Description
We need a quicker and easier way to disable indexer that shut down and or substitute their Urls.
I wanted to roll this into a feed of 'notifications' from services.
3 types were defines:
- Simple user notifications. I have no plans to implement the clientside atm, but the sole goal is to show these notifications to the user.
- UrlBlacklist. A regex match for urls to disable the relevant indexer.
- UrlReplace. A regex match+replace to replace the url with a new one.

So far I've added a simple substitution logic in the `IndexerFactory`, I've added an `IIndexerSettings` interface so that all indexers have the same `BaseUrl` property. The substitution isn't persistent (yet), it's purely the logic that runs if the user doesn't touch it.

What I'd like to do is to add a HealthCheck to handle the UrlBlacklist entries, notifying the user that the indexer is disabled.
And to have a Housekeeping task that applies the substitutions.

#### Todos
- [ ] More Tests
- [ ] Improve logic in IndexerFactory
- [ ] HealthCheck for UrlBlacklist
- [ ] Housekeeper for UrlSubstitute
- [ ] Validation/Test logic to notify user of invalid url while editing settings
- [ ] Security, since we're modifying urls we need to be sure it's safe. Ideally we'd start signing the api response some way.

### DIFF
--- a/src/NzbDrone.Core/Indexers/IndexerFactory.cs
+++ b/src/NzbDrone.Core/Indexers/IndexerFactory.cs
@@ -2,7 +2,9 @@
 using System.Linq;
 using NLog;
 using NzbDrone.Common.Composition;
+using NzbDrone.Common.Extensions;
 using NzbDrone.Core.Messaging.Events;
+using NzbDrone.Core.SkyhookNotifications;
 using NzbDrone.Core.ThingiProvider;
 
 namespace NzbDrone.Core.Indexers
@@ -15,17 +17,20 @@ namespace NzbDrone.Core.Indexers
 
     public class IndexerFactory : ProviderFactory<IIndexer, IndexerDefinition>, IIndexerFactory
     {
+        private readonly ISubstituteIndexerUrl _substituteIndexerUrl;
         private readonly IIndexerStatusService _indexerStatusService;
         private readonly Logger _logger;
 
-        public IndexerFactory(IIndexerStatusService indexerStatusService,
+        public IndexerFactory(ISubstituteIndexerUrl replaceIndexerUrl,
+                              IIndexerStatusService indexerStatusService,
                               IIndexerRepository providerRepository,
                               IEnumerable<IIndexer> providers,
-                              IContainer container, 
+                              IContainer container,
                               IEventAggregator eventAggregator,
                               Logger logger)
             : base(providerRepository, providers, container, eventAggregator, logger)
         {
+            _substituteIndexerUrl = replaceIndexerUrl;
             _indexerStatusService = indexerStatusService;
             _logger = logger;
         }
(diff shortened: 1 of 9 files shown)


src/NzbDrone.Core/SkyhookNotifications/SubstituteUrlService.cs
```diff
@@ -0,0 +1,51 @@
...
+                    if (action.RegexMatch == null) continue;
+
+                    if (Regex.IsMatch(baseUrl, action.RegexMatch))
+                    {
+                        return null;
+                    }
+                }
+                else if (action.Type == SkyhookNotificationType.UrlReplace)
+                {
+                    if (action.RegexMatch == null) continue;
```
markus101: `IsNullOrWhitespace()`?

src/NzbDrone.Core/Indexers/IndexerFactory.cs
```diff
@@ -62,12 +67,31 @@ public List<IIndexer> SearchEnabled(bool filterBlockedIndexers = true)
...
             if (filterBlockedIndexers)
             {
-                return FilterBlockedIndexers(enabledIndexers).ToList();
+                return FilterBlockedIndexers(SubstituteIndexerUrls(enabledIndexers)).ToList();
             }
 
             return enabledIndexers.ToList();
         }
 
+        private IEnumerable<IIndexer> SubstituteIndexerUrls(IEnumerable<IIndexer> indexers)
```
markus101: Would we want to persist these changes to the DB instead of applying them when they are used?

Definitely pros and cons to both, but changing it might make for a more obvious fix to the user, unless the notification message will be displayed when editing (at some point?).

src/NzbDrone.Core/SkyhookNotifications/SkyhookNotificationProxy.cs
```diff
@@ -0,0 +1,61 @@
...
+using NzbDrone.Common.Http;
+
+namespace NzbDrone.Core.SkyhookNotifications
+{
+    public interface ISkyhookNotificationProxy
+    {
+        List<SkyhookNotification> GetNotifications();
+    }
+
+    public class SkyhookNotificationProxy : ISkyhookNotificationProxy
```
markus101: Do we want to use Skyhook as a name here? That seems a bit confusing since we generally refer to our TVDB as Skyhook. Totally up for a name for it, but I don't think Skyhook is it.

src/NzbDrone.Core/Indexers/IndexerFactory.cs
Taloth: I didn't want Factory to change settings, only overrule them but UrlBlacklist can only be done here. The substitute would actually be persisted by a periodic task, Housekeeping was my thought.

src/NzbDrone.Core/SkyhookNotifications/SkyhookNotificationProxy.cs
Taloth: I didn't want it as ServicesNotifications as it's more easily confused and simply picked Skyhook as working title I'm open for suggestions. Notifications is already taken, Blog isn't it, Feed nope, Updates, already taken....

PR #1916, opened 2017-05-10
Downloadclient status

Taloth:
#### Database Migration
Yes 114 (already pushed to dev)
Yes 115

#### Description
Basically the same as for Indexer Backoff. But with a few twists.
First, we need to ONLY do the backoff for CDH, and not for download attempts. Download attempts should bypass the backoff logic.
Second, we should have a maximum escalation of about 1-3h max instead of 24h. We just want to avoid pointless errors to the user logs and sentry.
Third, we don't want Backoff to kick in immediately, I'd suggest to set that it must have been failing for at least 5 minutes. It's stored in the DB, so we need not be afraid of restarts.

Optional is to add the logic that allows specific HealthChecks to be rerun based on certain events.

#### Checklist
- [x] Convert IndexerStatusService into Thingy Provider arch
- [x] Add DownloadClientStatusService
- [x] Integrate backoff logic into CDH checks
- [ ] Tests for Download Client specific stuff
- [x] Add improved health check auto-rerun logic
- [x] Add health check explanation to Wiki
- [ ] Documentation


src/NzbDrone.Core.Test/Download/DownloadClientTests/DownloadClientStatusServiceFixture.cs
```diff
@@ -0,0 +1,93 @@
...
+using System.Linq;
+using FluentAssertions;
+using Moq;
+using NUnit.Framework;
+using NzbDrone.Core.Download;
+using NzbDrone.Core.Indexers;
+using NzbDrone.Core.Test.Framework;
+using NzbDrone.Core.ThingiProvider.Status;
+
+namespace NzbDrone.Core.Test.IndexerTests
```
Taloth: Wrong namespace.

PR #1932, opened 2017-05-15
Scene numbering override

alzeebum:
#### Database Migration
YES

#### Description
There are instances where XEM's episode numbering/mapping is incorrect.  This changeset gives users a way to ignore that mapping when they believe it is appropriate, on a per-season basis.  As a user myself, this is a better option than sitting with my fingers crossed hoping that XEM will fix their mapping or that the uploaders will rename their files to conform with the syntax.

This code adds a new boolean database field to the series table (IgnoreSceneNumbering), a toggle for that setting to the series edit panel of the UI, and two new properties to series -- IgnoreSceneNumbering and ActuallyUseSceneNumbering.  Where appropriate, calls to UseSceneNumbering were replaced with calls to ActuallyUseSceneNumbering.

#### Todos
- [ ] Unit/functional tests.  Manual testing was performed and seemed ok.  I'm not a node.js guy nor a C# guy so getting this much done was a bit of a trial;  I made no attempt to grok any existing test framework in the project.

- [ ] Documentation

*

### DIFF
--- a/src/NzbDrone.Api/Series/SeriesResource.cs
+++ b/src/NzbDrone.Api/Series/SeriesResource.cs
@@ -58,6 +58,7 @@ namespace NzbDrone.Api.Series
         public bool Monitored { get; set; }
 
         public bool UseSceneNumbering { get; set; }
+        public bool IgnoreSceneNumbering { get; set; }
         public int Runtime { get; set; }
         public int TvdbId { get; set; }
         public int TvRageId { get; set; }
@@ -131,6 +132,7 @@ namespace NzbDrone.Api.Series
                 Monitored = model.Monitored,
 
                 UseSceneNumbering = model.UseSceneNumbering,
+                IgnoreSceneNumbering = model.IgnoreSceneNumbering,
                 Runtime = model.Runtime,
                 TvdbId = model.TvdbId,
                 TvRageId = model.TvRageId,
@@ -185,6 +187,7 @@ namespace NzbDrone.Api.Series
                 Monitored = resource.Monitored,
 
                 UseSceneNumbering = resource.UseSceneNumbering,
+                IgnoreSceneNumbering = resource.IgnoreSceneNumbering,
                 Runtime = resource.Runtime,
                 TvdbId = resource.TvdbId,
                 TvRageId = resource.TvRageId,
--- a/src/NzbDrone.Core/DataAugmentation/Xem/XemService.cs
+++ b/src/NzbDrone.Core/DataAugmentation/Xem/XemService.cs
@@ -223,7 +223,7 @@ namespace NzbDrone.Core.DataAugmentation.Xem
                 return;
             }
 
-            if (!_cache.Find(message.Series.TvdbId.ToString()) && !message.Series.UseSceneNumbering)
+            if (!_cache.Find(message.Series.TvdbId.ToString()) && !message.Series.ActuallyUseSceneNumbering)
             {
                 _logger.Debug("Scene numbering is not available for {0} [{1}]", message.Series.Title, message.Series.TvdbId);
                 return;
--- a/src/NzbDrone.Core/Datastore/Migration/115_add_scene_number_override_flag.cs
+++ b/src/NzbDrone.Core/Datastore/Migration/115_add_scene_number_override_flag.cs
@@ -1,0 +1,15 @@
+﻿using FluentMigrator;
+using NzbDrone.Core.Datastore.Migration.Framework;
+
+namespace NzbDrone.Core.Datastore.Migration
+{
+    [Migration(115)]
+    public class add_scene_number_override_flag : NzbDroneMigrationBase
+    {
+        protected override void MainDbUpgrade()
+        {
+            Alter.Table("Series").AddColumn("IgnoreSceneNumbering").AsBoolean().Nullable();
+            Execute.Sql("UPDATE Series SET IgnoreSceneNumbering = 0");
+        }
+    }
+}
--- a/src/NzbDrone.Core/DecisionEngine/Specifications/AcceptableSizeSpecification.cs
+++ b/src/NzbDrone.Core/DecisionEngine/Specifications/AcceptableSizeSpecification.cs
@@ -77,7 +77,7 @@ namespace NzbDrone.Core.DecisionEngine.Specifications
                     List<Episode> seasonEpisodes;
 
                     var seasonSearchCriteria = searchCriteria as SeasonSearchCriteria;
-                    if (seasonSearchCriteria != null && !seasonSearchCriteria.Series.UseSceneNumbering && seasonSearchCriteria.Episodes.Any(v => v.Id == episode.Id))
+                    if (seasonSearchCriteria != null && !seasonSearchCriteria.Series.ActuallyUseSceneNumbering && seasonSearchCriteria.Episodes.Any(v => v.Id == episode.Id))
                     {
                         seasonEpisodes = (searchCriteria as SeasonSearchCriteria).Episodes;
                     }
--- a/src/NzbDrone.Core/IndexerSearch/NzbSearchService.cs
+++ b/src/NzbDrone.Core/IndexerSearch/NzbSearchService.cs
@@ -66,12 +66,15 @@ namespace NzbDrone.Core.IndexerSearch
 
                 return SearchDaily(series, episode, userInvokedSearch);
             }
-            if (series.SeriesType == SeriesTypes.Anime)
+
+            // The anime type just does absolute searching, so don't use it if we're not using scene numbering.
+            if (series.SeriesType == SeriesTypes.Anime && series.ActuallyUseSceneNumbering)
             {
                 return SearchAnime(series, episode, userInvokedSearch);
             }
 
-            if (episode.SeasonNumber == 0)
+            // S00Eyy is a common pattern so don't use special if we're ignoring scene numbering.
+            if (episode.SeasonNumber == 0 && series.ActuallyUseSceneNumbering)
             {
                 // search for special episodes in season 0 
                 return SearchSpecial(series, new List<Episode> { episode }, userInvokedSearch);
@@ -103,7 +106,7 @@ namespace NzbDrone.Core.IndexerSearch
 
             var downloadDecisions = new List<DownloadDecision>();
 
-            if (series.UseSceneNumbering)
+            if (series.ActuallyUseSceneNumbering)
             {
                 var sceneSeasonGroups = episodes.GroupBy(v =>
                 {
@@ -162,7 +165,7 @@ namespace NzbDrone.Core.IndexerSearch
         {
             var searchSpec = Get<SingleEpisodeSearchCriteria>(series, new List<Episode> { episode }, userInvokedSearch);
 
-            if (series.UseSceneNumbering && episode.SceneSeasonNumber.HasValue && episode.SceneEpisodeNumber.HasValue)
+            if (series.ActuallyUseSceneNumbering && episode.SceneSeasonNumber.HasValue && episode.SceneEpisodeNumber.HasValue)
             {
                 searchSpec.EpisodeNumber = episode.SceneEpisodeNumber.Value;
                 searchSpec.SeasonNumber = episode.SceneSeasonNumber.Value;
--- a/src/NzbDrone.Core/NzbDrone.Core.csproj
+++ b/src/NzbDrone.Core/NzbDrone.Core.csproj
@@ -293,6 +293,7 @@
     </Compile>
     <Compile Include="Datastore\Migration\105_rename_torrent_downloadstation.cs" />
     <Compile Include="Datastore\Migration\111_create_language_profiles.cs" />
+    <Compile Include="Datastore\Migration\115_add_scene_number_override_flag.cs" />
     <Compile Include="Datastore\Migration\Framework\MigrationContext.cs" />
     <Compile Include="Datastore\Migration\Framework\MigrationController.cs" />
     <Compile Include="Datastore\Migration\Framework\MigrationDbFactory.cs" />
--- a/src/NzbDrone.Core/Parser/ParsingService.cs
+++ b/src/NzbDrone.Core/Parser/ParsingService.cs
@@ -430,7 +430,7 @@ namespace NzbDrone.Core.Parser
 
             foreach (var episodeNumber in parsedEpisodeInfo.EpisodeNumbers)
             {
-                if (series.UseSceneNumbering && sceneSource)
+                if (series.ActuallyUseSceneNumbering && sceneSource)
                 {
                     List<Episode> episodes = new List<Episode>();
 
--- a/src/NzbDrone.Core/Tv/Series.cs
+++ b/src/NzbDrone.Core/Tv/Series.cs
@@ -36,7 +36,14 @@ namespace NzbDrone.Core.Tv
         public List<MediaCover.MediaCover> Images { get; set; }
         public SeriesTypes SeriesType { get; set; }
         public string Network { get; set; }
-        public bool UseSceneNumbering { get; set; }
+
+        protected bool _UseSceneNumbering;
+        protected bool _IgnoreSceneNumbering = false;
+        public bool UseSceneNumbering {
+            get { return _UseSceneNumbering && !_IgnoreSceneNumbering; }
+            set { _UseSceneNumbering = value; }
+        }
+
         public string TitleSlug { get; set; }
         public string Path { get; set; }
         public int Year { get; set; }
--- a/src/NzbDrone.Core/Tv/Series.cs
+++ b/src/NzbDrone.Core/Tv/Series.cs
@@ -37,11 +37,11 @@ namespace NzbDrone.Core.Tv
         public SeriesTypes SeriesType { get; set; }
         public string Network { get; set; }
 
-        protected bool _UseSceneNumbering;
-        protected bool _IgnoreSceneNumbering = false;
-        public bool UseSceneNumbering {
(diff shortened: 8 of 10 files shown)


markus101:
It looks like two issues are being addressed here, not using the scene mappings or searching for releases by season + episode number even when the series type is anime?

The first part of the problem seems to be primarily what you're aiming to fix, but then you also added in a circumvention to search by season + episode instead of absolute episode number when the mappings are turned off, which doesn't make sense, becasue the absolute episode number is still valid, it's not a scene number, it comes from TheTVDB directly and should be used when the series type is anime (that is a primary reason for setting the series type to anime). I still maintain that searching for season + episode number for anime for every search is a waste of resources, no matter how that is perceived, but searching for season + episode when an absolute search fails (similar to falling back to a title search when searching by ID fails), is not a waste (on newznab indexers at least) and would be a welcomed improvement.

As I mentioned on Reddit, XEM falls short in a few areas, one that you've noticed, I wasn't aware of this one, but probably because most indexers keep the absolute episode number, but nzbgeek appears to have re-uploaded the release with the season and episode number, (which IMO is a better format, but I digress), but that seems to be why you discovered it. Another is there is no difference between scene and p2p releases so for something like American Dad the scene's season number is off by one, but p2p groups release episodes following TheTVDB numbering, being able to disable scene numbering for American Dad appears to fix the problem, but introduces another problem in that Sonarr could start grabbing the scene releases for another season because it doesn't know that there it's actually for another episode in another season.

An idea we've talked about on IRC (@Taloth and I as well as some XEM admins) is the concept of XEMv2 which would allow for multiple mappings at different qualities and different release groups (matching by regex or something else), this would give a lot of flexibility and we building it from the ground up (and with years of experience now) we could improve how moderation is handled. The idea is all well and good, but it needs a lot of time/man power to implement it and test it and with work, life, Sonarr there isn't much time to implement it.

I don't know what our plan moving forward is for XEMv2, but it needs to go somewhere, otherwise changing Sonarr looks like the easy way out, but it's not the right place to do it.

I don't think the changes here, properly address the underlying issues, but I don't think it has to stop with XEMv2 will solve it, one day, I think we can achieve falling back to season + episode searching if an absolute episode search fails quite easily (the system is there to do it, it just needs to be implemented in the `NewznabRequestGenerator`). For the issue with the season/episode numbers being out of wack, if it's not something XEM can fix for whatever reason (on the surface it looks like it's misconfigured, but there may be something I'm missing) I think we could improve our handling of that and throw away the scene numbering if the absolute episode numbers match between the scene, anidb and TVDB and the season + episode numbers match between the scene and anidb (or something similar to that), with the added benefit that we don't show a mapping in the UI that's useless from the absolute number perspective.

If XEMv2 is something you feel you could contribute to, let us know, it's still in the thought stage (not even planning), but really needs someone capable and willing to take it on.

alzeebum:
From the perspective of an end user, there is only one issue being addressed here:  Sonarr failing to find episodes at an indexer when they exist.  All I care about is that when it searches an indexer for episodes, if they exist, they are found.  I am certain this is all that the majority of users care about as well.  If they are not found, nobody actually cares if it's because XEM has some mapping wrong, if the indexer doesn't have all the data entered correctly, or whatever else the reason may be.  All we care about is that we aren't getting the results we expect.

Since sonarr is the program we're using to search our indexer(s), and our indexer(s) are returning matching results, sonarr is ultimately to blame.  This applies even when it's some weirdness happening at XEM -- since sonarr gives us no choice but to use XEM, and provides no ability to configure *how* XEM is used.

FWIW, the anime toggle does not seem to work properly anyway.  I expect it to treat the series as normal when set to standard, and thus for example to prefer thetvdb's numbering to anidb's when the two are in conflict.  As you explained in my previous issue though, when I turn it off, sonarr is still using the absolute episode numbering if it is present -- still treating the show as 'anime' in otherwords.  This is not desirable behavior.

I know it is doing this because when I first ran into this problem and tracked it down in the source, I found it by debugging the http calls to the indexer.  You mentioned that it searches on show id but what I saw it do was something else.  It was searching only with straight query strings; either "Series+Exxx" when in "anime" mode, or "Series+SxxEyy" when in standard mode, for that particular show.  When coupled with the mapping, this lead to the situation I outlined in the initial issue I created:  Searching for e.g. absolute episode 11 (season 2 episode 1) resulted in searches for either "Series+11" ("anime" mode) or "Series+S01E11" ("standard" mode), both of which returned no results.

With the changes I made in place, and the flag on, it issued the search for "Series+S02E01" which is exactly what I want to happen -- regardless of the "type" of series, which is what I (as a user) would want and expect such a setting to do.

I understand the time constraints with getting XEM moderation accomplished but to be blunt, I don't really care.  As a user, I am interested in simply being able to turn knobs and flip levers until I get the results I want.  As a developer, I am interested in giving those knobs and levers to other end users.  This particular lever only took an afternoon to code and, as far as I can tell, has absolutely no downside beyond making the UI marginally more complex.  Do you disagree?

If there were more control in sonarr so that I could choose which datasource (anidb, thetvb, etc) to use for a series when getting data from XEM that would be one thing -- but since there isn't, I find XEM integration itself to be rather useless.

Waiting to hear back from XEM holds no interest for me.  I don't care if they fix the numbering or not since I have a workaround.  Working on XEM2 may be of interest in the future, but not if users aren't going to be given a much greater say in how it is used, and not as a "substitute" for some other fix within sonarr itself that can address the issue *now* rather than months from now.

Right now, what are you are suggesting as a course of action?

My gut says we just have very different philosophies as developers and that we probably aren't going to see eye to eye here.  I'm basing this somewhat on the comment in the reddit discussion where you said you don't want to provide a customizable search field in the series and would rather continue to maintain a curated list.  I quite strongly disagree.

markus101:
> FWIW, the anime toggle does not seem to work properly anyway. I expect it to treat the series as normal when set to standard, and thus for example to prefer thetvdb's numbering to anidb's when the two are in conflict. As you explained in my previous issue though, when I turn it off, sonarr is still using the absolute episode numbering if it is present -- still treating the show as 'anime' in otherwords. This is not desirable behavior.

Scene mappings apply to shows regardless of the type, the reason it doesn't work properly for Bleach because of how the mapping is setup on XEM, it's creating a scene mapping that follows anidb's numbering, which isn't of any use to Sonarr and in the case of wanting to search by TVDB's numbering is broken.

>With the changes I made in place, and the flag on, it issued the search for "Series+S02E01" which is exactly what I want to happen -- regardless of the "type" of series, which is what I (as a user) would want and expect such a setting to do.

You're getting the result you wanted, but your setting is hiding the underlying change that gets you the result by changing something it shouldn't, you'd get the same result if the series type was set to standard with the same `IgnoreSceneNumbering` setting and not have to hack the underlying code.

> This particular lever only took an afternoon to code and, as far as I can tell, has absolutely no downside beyond making the UI marginally more complex. Do you disagree?

I do, as mentioned above it breaks searching by absolute episode number if scene mapping is disabled in order to find the results you want, it shouldn't do that.

> If there were more control in sonarr so that I could choose which datasource (anidb, thetvb, etc) to use for a series when getting data from XEM that would be one thing -- but since there isn't, I find XEM integration itself to be rather useless.

Either you don't have any series that XEM helps correct or those series are working without further intervention, there is no doubt there is something wrong with how Bleach is handled, whether that's on the XEM side or Sonarr's expectation of XEM is a bit blurred, but calling it useless is inaccurate.

> Right now, what are you are suggesting as a course of action?

I'm not completely opposed to an option to disable scene mappings, but I don't think it's required here and lacking a concrete reason to add it I don't think it should be added, it opens up another avenue for issues and is guaranteed to be used wrong by some users. 6+ years working on Sonarr has taught me that if there is an option it will be used wrong, so if we have a better solution let's go for that.

1. Instead of an option to disable scene mapping we should intelligently disable it if it doesn't make sense, which would be to check the mapping and not apply it if it isn't doing anything
2. For Newznab indexers  add a fallback to season + episode searching if a search by absolute episode number returns no results

> My gut says we just have very different philosophies as developers and that we probably aren't going to see eye to eye here. I'm basing this somewhat on the comment in the reddit discussion where you said you don't want to provide a customizable search field in the series and would rather continue to maintain a curated list. I quite strongly disagree.

I'm not basing that solely on development philosophies, but on the fact that after it's added it needs to supported, which largely falls on @Taloth and myself, occasionally other developers add features and help out with any issues, but that's uncommon in our experience. Maintaining a list allows us to fix it once and for everyone, yes the downside is that a user can't simply add a title and have it work immediately, but with the issues I pointed out having to set an alias to search shouldn't be required.

alzeebum:
> You're getting the result you wanted, but your setting is hiding the underlying change that gets you the result by changing something it shouldn't, you'd get the same result if the series type was set to standard with the same IgnoreSceneNumbering setting and not have to hack the underlying code

What "underlying change" is it hiding?  It does what it's supposed to do with with both the "anime" and "standard" settings, which is exactly what I wanted it to do.

>it breaks searching by absolute episode number

it doesn't "break" it, it disables it.  That's what it's intended to do.  It may be named poorly -- the goal is to search by season/episode as it is presented to the user, which I suppose is simply thetvdb no matter what?  I had never heard of XEM or "scene mapping" or "scene order" before all this, so it's quite possible I'm using those terms incorrectly.

>there is no doubt there is something wrong with how Bleach is handled

No longer an issue.  Some person named "xelra" @ XEM has deleted the entire series.  Another reason for me to hate on XEM.

>lacking a concrete reason to add it I don't think it should be added

The "concrete reason" for me is the same as for any feature or knob.  User choice, freedom, opinion, mood, whatever you choose to call it.

>if there is an option it will be used wrong

Perhaps you shouldn't think of it as being used "wrong" but rather "differently than intended."  I've learned the same lesson in my 20 years of developing software in general.  What I took away from it is "users will do unexpected things.  Embrace and encourage it."

> Instead of an option to disable scene mapping

There is no reason for this to be an "instead of" situation.  The more explicit options the better.  I don't really care if they're only available in the db and not even exposed in the UI, but they should still be there.

> For Newznab indexers add a fallback to season + episode searching if a search by absolute episode number returns no results

This is not a solution.  Sometimes the absolute numbering returns a single hit while the S+E numbering returns more.  Sometimes it's the other way around.  If there isn't going to be a knob to explicitly switch between the two, then both of them should be used every single time.  IIRC you already rejected this suggestion in my previous issue.

> I'm not basing that solely on development philosophies, but on the fact that after it's added it needs to supported

If you don't respond to a support request, the community will.  That is what communities are for.  Not adding a useful feature simply because it may generate more support requests doesn't strike me as a particularly good reason.

markus101:
> What "underlying change" is it hiding?

The change to that disables disables searching by the absolute episode number because scene mapping is disabled. If that's the intention of the change it's incorrect, falling back to searching the season + episode number in the case where the absolute episode number search fails would be a better solution, that way even if they disabled the mapping it'd still work.

> No longer an issue. Some person named "xelra" @ XEM has deleted the entire series. Another reason for me to hate on XEM.

Yes, deleted because it wasn't providing anything useful, which is what I've been saying. Why is that a reason to hate on XEM? It was deleted to fix part of your issue, the other part being how Sonarr searches for anime episodes, but this was a problem. I'm not even sure what reason there is to hate XEM at all, it's not perfect, but it's still helpful.

>> For Newznab indexers add a fallback to season + episode searching if a search by absolute episode number returns no results

>This is not a solution. Sometimes the absolute numbering returns a single hit while the S+E numbering returns more. Sometimes it's the other way around. If there isn't going to be a knob to explicitly switch between the two, then both of them should be used every single time. IIRC you already rejected this suggestion in my previous issue.

The majority of anime is not released with season + episode numbers, only absolute episode numbers, always searching doesn't sounds like the best use of API calls, if it doesn't yield any more results, maybe it will, maybe it won't and like you said it's going to differ. In this case the knob already exists, it's the series type option, that changes the primary format used for searching, the reason it was broken for Bleach before was due to the fault mapping on XEM.

alzeebum:
> falling back to searching the season + episode number in the case where the absolute episode number search fails would be a better solution

That isn't a solution at all, let alone a better solution.  Having sonarr "automatically fall back" without any possibility of the user overriding this behavior is the *problem*.  Sometimes you want to use the "fallback" no matter what sonarr or XEM think.

I'm not sure why this isn't as obvious to you as it is to me -- sometimes XEM is wrong.  Sometimes sonarr's assumptions are wrong.  The solution is to provide a way to override those results and assumptions -- not bake even more assumptions in.

> Yes, deleted because it wasn't providing anything useful, which is what I've been saying.

Sure.  I've been saying to delete them all!  ;)

> Why is that a reason to hate on XEM

You don't find it even slightly absurd that the solution to a bad mapping was to delete the entire show instead of fixing the mapping, whatever fixing means in this case?

> The majority of anime is not released with season + episode numbers, only absolute episode numbers, 

That was true 5 or 10 years ago, but it no longer is.  I am subscribed to over a dozen currently airing anime series and only 4 of them are being released with *only* absolute numbering right now, all the others are released in both naming formats as they air.  It just depends on the group doing the releasing, and most these days do both and have for a few years now. 

> In this case the knob already exists, it's the series type option

That knob does not perform the same function.  If it did, then changing the series type would have solved my problem and I wouldn't have had to dive into visual studio to figure out what was going on.  Pointing the finger at XEM isn't productive unless there is an easy way to circumvent it when it misbehaves, which there currently isn't.

> always searching doesn't sounds like the best use of API calls

I completely disagree.  They are not a precious resources, and as long as they are returning different results (which they will), not searching for both does a disservice to the user.  Nearly every series I'm subscribed to has releases in both formats by different groups, each with different quality levels and other properties.

The more I think about it, the more I believe that this is really what I want it to do.  It should be configurable of course so I'm thinking about how to approach that.  I'm torn between making it another on/off toggle for the series, or adding another 'type' called 'both' or something to the series type dropdown.

I'll close this pull request now since it's pretty obvious you're not going to take it.   Thanks for taking the time to discuss it anyway.  I think I'm probably just going to keep the fork going and take things my own way.  As a user, I want as much control over the program as possible, and we have a fundamental disagreement in that regard.  I don't really care how much the support requests jump -- I'm under no obligation to handle them personally, and neither are you.

markus101:
> The solution is to provide a way to override those results and assumptions -- not bake even more assumptions in.

Your solution assumes that an absolute number is only for scene numbering and bypasses that when scene numbering is disabled, disabling the scene numbering is one thing, but also having that setting change how the episode is searched is confusing. 

> That knob does not perform the same function. If it did, then changing the series type would have solved my problem and I wouldn't have had to dive into visual studio to figure out what was going on. Pointing the finger at XEM isn't productive unless there is an easy way to circumvent it when it misbehaves, which there currently isn't.

It does, which is why you were seeing it search for `S01Exxx` instead of `xxx`, there were two issues there, one the mapping didn't make sense and two you wanted to search a different format,\. With the option to disable the mapping you fix problem one and by changing the series type you fixed problem two.

> You don't find it even slightly absurd that the solution to a bad mapping was to delete the entire show instead of fixing the mapping, whatever fixing means in this case?

Not when the mapping served no purpose, it was a waste of bytes, not really different from ignoring it since it was wrong.

> Nearly every series I'm subscribed to has releases in both formats by different groups, each with different quality levels and other properties.

I'm curious about which series you're currently following that release in both formats, it's not something I've seen done consistently enough to warrant searching it, but my anime searching is quite limited.

> The more I think about it, the more I believe that this is really what I want it to do. It should be configurable of course so I'm thinking about how to approach that. I'm torn between making it another on/off toggle for the series, or adding another 'type' called 'both' or something to the series type dropdown.

If that's what you want to do I'm not even sure it needs to be part of the series, unless you explicitly want to turn it off at a series level (but that seems like an extra step a user needs to do to get more results for every series they want to search both formats for), you could have a per-indexer or global setting handle it, but i don't have a horse in that race, just my 2c.

Best of luck, thanks for the discussion, even though we disagree on several points it's good to get the insight.

alzeebum:
> Your solution assumes that an absolute number is only for scene numbering and bypasses that when scene numbering is disabled

Well I guess you can put it that way, but I wouldn't call it an assumption so much as pure unadulterated ignorance.  I have no idea what "scene numbering" is even supposed to mean, so I wasn't assuming it was used one way or another.  What I found when I was stepping through the code was that in the problematic series and episodes, UseSceneNumbering was always true, and forcing it to false completely solved my issue -- so I set about making it switchable.

> It does, which is why you were seeing it search for S01Exxx instead of xxx, there were two issues there, one the mapping didn't make sense and two you wanted to search a different format,. With the option to disable the mapping you fix problem one and by changing the series type you fixed problem two

By "disable the mapping" you mean the series deletion (or repair) from XEM?  That is hardly a realistic option for your average user.  I'm sure most users, like me three days ago, have never *heard* of XEM.  Sonarr itself isn't exactly forthcoming that it's being used or that it even exists.

> Not when the mapping served no purpose, it was a waste of bytes, not really different from ignoring it since it was wrong

There are plenty of series in XEM that serve no purpose, other than to provide a list of alternate series names (which Bleach also had, or at least, should have had "ブリーチ" at a minimum), yet they aren't deleted.

> I'm curious about which series you're currently following that release in both formats

These four off the top of my head.  I'll get you an exhaustive list from my library if you like:
March Comes In Like A Lion
Clockwork Planet
Akashic Records Of A Bastard Magic Instructor
Eromanga Sensei

It is common these days.  Some groups like PHF only release in SxxEyy format.  Some like horriblesubs & bakedfish only seem to release in absolute format.  ANiURL doesn't seem to have any rhyme or reason to which format they release in.  Restricting searches to just one format or the other means missing out on all the rest.

> If that's what you want to do I'm not even sure it needs to be part of the series
You're right, that doesn't really make any sense.  A top level toggle, or per-indexer would be better.

PR #1938, opened 2017-05-21
Added [Sonarr] tag for Twitter Notifications

jcostom:
#### Database Migration
NO

#### Description
Just a simple addition to Twitter notifications to identify the source of the tweet as coming from Sonarr.

* None - just a simple enhancement, dare I say, nearly trivial.

### DIFF
--- a/src/NzbDrone.Core/Notifications/Twitter/Twitter.cs
+++ b/src/NzbDrone.Core/Notifications/Twitter/Twitter.cs
@@ -21,12 +21,12 @@ namespace NzbDrone.Core.Notifications.Twitter
 
         public override void OnGrab(GrabMessage message)
         {
-            _twitterService.SendNotification($"Grabbed: {message.Message}", Settings);
+            _twitterService.SendNotification($"[Sonarr] Grabbed: {message.Message}", Settings);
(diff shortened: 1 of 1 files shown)


markus101:
We'd rather let it be customizable and not force this change on everyone using, that'd be part of a longer standing issue: https://github.com/Sonarr/Sonarr/issues/328 but not something we're looking to implement until after v3 is out.

jcostom:
It's a shame you're not open to this, even if it's only until after you get around to coding what you'd really like to have.  It's such a simple change that reverting the tags in the _2 lines of code_ would be trivial.  At least this is progress toward a goal.

The Radarr team was happy to accept the pull.

markus101:
It's not that we wouldn't want to to revert it, it's that we don't want to change the pattern that is already in place (and affect existing users, which we've gone through before), because there will be  people that don't want the change. In addition sending through a Sonarr specific account could help remove confusion on where the tweet came from, albeit in a different way.

jcostom:
Can't say I agree, but it's not my project, so cheers.

Love the tool, works fantastically.  Thanks for your work on it!

PR #1957, opened 2017-06-02
qbittorrent: add Origin header for webui CSRF

### LINKED ISSUE #1956: qBitTorrent 3.3.14 broken due to changed api
Looks like qbit added Origin: or Referer: header requirements on the api for 3.3.13: https://github.com/qbittorrent/qBittorrent/commit/087856d3d8d6d77881c3d8378a95d80d693af2dd

Needs to be verified first, and then we should probably just send out the Origin header as appropriate.

cbodley:
#### Database Migration
NO

#### Description
Add Origin header for CSRF enforced by qBitTorrent 3.3.13

Without this header, requests to the qBitTorrent WebUI are rejected with 401. With the header (i.e. `Origin: http://localhost:8080`), the requests succeed as normal.

* Fixes #1956


Taloth:
I'd prefer to wait for the discussion on the qbittorrent issue that was created for it. Coz it seems like they are willing to reevaluate the decision.

Also, while Referer is the original URL, the Origin is just scheme+host+port. BaseUrl may include a sub path, so you should use Uri or HttpUrl to get the correct value.

cbodley:
cool, looks like this won't be needed

PR #1606, opened 2016-12-30
crushed with pingo

drewfreyling:
#### Database Migration
NO

#### Description
PNG files made smaller with https://css-ig.net/pingo

sonarr-pngcrush

PR #1959, opened 2017-06-04
lossless compression of images saved 14KB

Fish2:
Save Bandwidth and speed up website load time but lossless completion of all images.

doing this saved 14KB docent sound like much but why not


Taloth:
Please... this is a waste of time. You saved 291 bytes effectively, not 14 KB. And that's like 0.8 % of the total traffic involved in fetching them, and 0.03 % vs the entire page load.
Prove to me that load time is faster or that any meaningful bandwidth is saved.

In development we call this premature optimization.

markus101:
These inages we're previously run through pingo, I'm not sure what's was used here, but I know additional passes in pingo will save a bit more each pass (to a point), but Ithe savings is minimal.

https://github.com/Sonarr/Sonarr/pull/1606

Fish2:
It has saved exactly 13.883KB but i don't know how many of the image files are used on the hosting side. but any optimization for free is surely a positive thing.

I used also used 
AdvPNG
PNGOUT

Fish2:
just seen you use so you could use gulp-imagemin

Issue #1541, opened 2016-11-08
OnRename - Source and Destination


nullb32:
Is it possible to add new environment variables to CustomScript.cs OnRename() function to expose the source path&file and destination path&file?

markus101:
It's an event emitted for a single series, it could be possible to track which files have changed with their old paths/filenames communicating them all would result in a really ugly environment variable (or list of environment variables).

What are you looking to achieve with this?

nullb32:
Currently I'm pointing Sonarr at an overlay fs, which breaks hardlinks.
Even though the rw portion of the overlay fs is on the same drive as the downloads folder Sonarr can't seem to hardlink but neither can regular ln.

Technically an overwrite option for hardlinks would work, such that:
Sonarr path = /home/something/overlayfs-all/a.mp4
Overwrite just for hardlinks = /home/something/overlayfs-local/a.mp4

But by adding something OnRename() I can just execute a shell script to delete the duplicated file and create a hardlink in it's place myself, I could disable renaming and do that now with OnDownload() option but I don't really want to disable renaming :/

mattmac24:
I think this would be useful in many different situations. I use a script on download to copy the file to several rclone destinations. On rename I have no easy way of renaming my files without deleting the whole series and reuploading from my local source.

sepbot:
Need the information to delete left over extracted file from .rar archives in torrents.

If OnRename can at the very least say what episode was renamed the additional coupling can happen by the consumer since source information can be cached from OnDownload event.

For the time being I will have to delegate renaming to Filebot and pass the output to Sonarr using black hole, but it will be a nice to have piece of information.

markus101:
@sepbot `OnRename` is triggered when renaming existing files, `OnDownload` is triggered when importing new files and has the information you're looking for.

markus101:
I've been thinking about this recently and while we could send all the renamed files via a custom script it's not very practical, I have an idea that would work in conjunction with a custom script though:

1. Store rename history for each episode that is renamed
2. Custom script is called
3. Custom script fetches history from Sonarr (filtered by series ID) - https://github.com/Sonarr/Sonarr/wiki/History
4. Script processes history items, filtering out non-renames and anything not relevant.

Sonarr would gain an audit log of renames and we would need to force things a list of items into environment variables.

I've opened a PR with the audit log portion so the other devs can take a look: https://github.com/Sonarr/Sonarr/pull/1963

gcrevell:
[...]

PR #1963, opened 2017-06-06
Store episode renames in History

markus101:
#### Database Migration
NO

#### Description
This would help for custom scripts where they want to know exactly which files were renamed as well as serve as an audit log when files are renamed, if you rename files often this will add some noise, but otherwise it'll be invisible.

* #1541


src/NzbDrone.Core/History/HistoryService.cs
```diff
@@ -257,6 +258,34 @@ public void Handle(EpisodeFileDeletedEvent message)
...
+                    EventType = HistoryEventType.EpisodeFileRenamed,
+                    Date = DateTime.UtcNow,
+                    Quality = message.EpisodeFile.Quality,
+                    SourceTitle = message.OriginalPath,
+                    SeriesId = message.EpisodeFile.SeriesId,
+                    EpisodeId = episode.Id,
+                };
+
+                history.Data.Add("SourcePath", sourcePath);
+                history.Data.Add("SourceRelativePath", sourceRelativePath);
```
Taloth: Just SourceRelativePath & RelativePath is enough, no need to add more data to the database.

markus101: Should we also store the series path at the time of the move? That way the absolute path is known and not changing if the series folder changes. (I think that was the reason I included relative and absolute).

Taloth: It's not a showstopper for me to keep both, it just felt extraneous. SeriesPath might be an idea.
Either keep what you have now, or use SeriesPath + 2x Relative.

PR #1968, opened 2017-06-10
Redirect calls missing URL Base

markus101:
#### Database Migration
NO

#### Description
Currently if a user has set a URL Base of `/sonarr` and then hits `localhost:8989` they're redirected to the login page at `localhost:8989/sonarr/login?returnUrl=/` and when they login they get redirected back to the login page because the cookie is stored on `/sonarr` not `/`. This hook needs to execute before the forms auth login hook.

### DIFF
--- a/src/NzbDrone.Api/Extensions/Pipelines/UrlBasePipeline.cs
+++ b/src/NzbDrone.Api/Extensions/Pipelines/UrlBasePipeline.cs
@@ -20,15 +20,24 @@ namespace NzbDrone.Api.Extensions.Pipelines
 
         public void Register(IPipelines pipelines)
         {
-            pipelines.BeforeRequest.AddItemToStartOfPipeline((Func<NancyContext, Response>) Handle);
+            if (_urlBase.IsNotNullOrWhiteSpace())
+            {
+                pipelines.BeforeRequest.AddItemToStartOfPipeline((Func<NancyContext, Response>) Handle);
+            }
         }
 
         private Response Handle(NancyContext context)
         {
+            var basePath = context.Request.Url.BasePath;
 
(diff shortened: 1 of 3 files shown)


src/NzbDrone.Api/Extensions/Pipelines/UrlBasePipeline.cs
```diff
@@ -0,0 +1,37 @@
...
+        public UrlBasePipeline(IConfigFileProvider configFileProvider)
+        {
+            _urlBase = configFileProvider.UrlBase;
+        }
+
+        public int Order => 99;
+
+        public void Register(IPipelines pipelines)
+        {
+            pipelines.BeforeRequest.AddItemToStartOfPipeline((Func<NancyContext, Response>) Handle);
```
Taloth: If _urlBase is null, then we don't need to attach the pipeline.

src/NzbDrone.Api/Extensions/Pipelines/UrlBasePipeline.cs
```diff
@@ -0,0 +1,37 @@
...
+        }
+
+        private Response Handle(NancyContext context)
+        {
+
+            if (_urlBase.IsNotNullOrWhiteSpace() && _urlBase != context.Request.Url.BasePath)
```
Taloth: What if someone visits :8989/somethingelse/api/ then the redirect will be wrong :8989/base/somethingelse/api.
How about redirect for the root, and 404 for anything else?

markus101: I wouldn't expect that to ever happen, but worth adding a check. 👍

Issue #1967, opened 2017-06-10
Fix for missing libmediainfo.so.0 causing bad sample detection


smcpeck:
Looking at `/src/NzbDrone.Core/MediaFiles/EpisodeImport/DetectSample.cs`, I am seeing what seems to be a defect.

I've suffered the unfortunate issue where `libmediainfo.so.0` isn't being properly consumed.  Obviously, this is a configuration issue, however, I got curious as to why the lack of the library causes good media to be erroneously detected as a sample.

That led me to the chunk of code below in the file mentioned above:

```
            try
            {
                var runTime = _videoFileInfoReader.GetRunTime(path);
                var minimumRuntime = GetMinimumAllowedRuntime(series);

                if (runTime.TotalMinutes.Equals(0))
                {
                    _logger.Error("[{0}] has a runtime of 0, is it a valid video file?", path);
                    return true;
                }

                if (runTime.TotalSeconds < minimumRuntime)
                {
                    _logger.Debug("[{0}] appears to be a sample. Runtime: {1} seconds. Expected at least: {2} seconds", path, runTime, minimumRuntime);
                    return true;
                }
            }

            catch (DllNotFoundException)
            {
                _logger.Debug("Falling back to file size detection");

                return CheckSize(size, quality);
            }
```

The first line here in the **try** block tries to get the media's runtime.  This fails and ends up returning 0.  If you look in `VideoFileInfoReader.cs` you will see:
`            catch (DllNotFoundException ex)
            {
                _logger.Error(ex, "mediainfo is required but was not found");
            }`

So the **catch** in the `DetectSample.cs` snippet above never actually happens in this case.  My thought is that simply adding `throw;` to the `VideoFileInfoReader.cs` catch of the DllNotFoundException would fix this exception.

Just wanted to get my analysis added here for someone to potentially put into action.  I'm lazy right now, but maybe someday I'll actually clone the repo and add this one line of code.  😆

markus101:
Thanks for digging in, looks like we haven't fallen back to the fie size in a long time, so making some changes to rip that out and report a better message when Sonarr can't determine if the file is a sample or not.

PR #1969, opened 2017-06-11
Better handling when MediaInfo is not available

markus101:
#### Database Migration
NO

#### Description
Turns out we haven't fallen back to file size to determine if a file is a sample for a long time (years maybe?), so I've fixed the glitch and also improved what Sonarr logs/reports when getting the mediainfo fails completely.

* #1967

### DIFF
--- a/src/NzbDrone.Core.Test/MediaFiles/DownloadedEpisodesImportServiceFixture.cs
+++ b/src/NzbDrone.Core.Test/MediaFiles/DownloadedEpisodesImportServiceFixture.cs
@@ -164,11 +164,9 @@ namespace NzbDrone.Core.Test.MediaFiles
 
             Mocker.GetMock<IDetectSample>()
                   .Setup(s => s.IsSample(It.IsAny<Series>(),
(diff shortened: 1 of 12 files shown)


Taloth:
Looks good, but one question: Do we really want to get rid of the size check? I know it wasn't used anymore, but it's a reasonable fallback. Not likely a 1GB file is gonna be a sample.

markus101:
I think as it was it had the possibility to screw something up, such as we added 2160p qualities and a large sample file (I'm not even sure what the rough size for a sample is for 2160p content) could have exceeded the default and been imported instead of the actual video file and the user would have to dig in, see what went wrong, find the download and import it. By not importing if mediainfo is unavailable we make it mandatory to have, it's black or white and when we have mediainfo checking resolutions we know it's there, not it might be there.

We also haven't had any kicking or screaming because it was gone, so it's extra cruft we don't need to maintain/worry about, though if it comes to it I'm not opposed to revisiting it.

Taloth:
Agreed

PR #1975, opened 2017-06-13
Upgrade Bootstrap and related dependencies

mitchellcash:


thezoggy:
this may be moot to include due to v3 (new ui).. as this change while small could introduce bugs/issues unplanned for...

mitchellcash:
@thezoggy I didn't realise! But it that's the case, I will close this :)

PR #1985, opened 2017-06-18
Quality groups

markus101:
#### Database Migration
YES (116)

#### Description
Adds support for grouping qualities in a profile and groups Webrip with WEBDL.

*


markus101:
Rebased and fixed the migration.

Taloth:
Not reviewing the code just get. But first the UI:

1) It's it confusing to the average user that Settings->Quality contains the detailed list, but Profiles doesn't show them unless Advanced is on. I don't have a good answer for this atm.
One option could be to add the 'subitems' as badges in the profile, like this:
image
Obviously gonna be a problem if ppl group a lot of items, but an option nonetheless.

2) I dislike that the quality list in the profile is so big that it requires a scroll bar by default. The quality profile is already confusing for the average user.
We could play with the margins, and maybe make the checkboxes a bit smaller, that'll set them apart from the other checkbox fields we have.
Additionally, I wonder if we move the Name to the left (unless in small viewport mode), put the quality list on the right so it can use the full height. Won't give it enough height to be single-page in advanced mode, but it will for simplified.
In another GHI I think I talked about idea to make the cutoff more intuitive by replacing the cutoff dropdown with an sliding arrow snapping to between the qualities. And rename the cutoff to 'upgrade until'.

3) Consider using the [indent](http://fontawesome.io/icon/indent/) or [object-group](http://fontawesome.io/icon/object-group/) icons instead of the plus for grouping. I think it's more intuitive. There are some other icons that might be useful too, but these were the most obvious imho.

4) In advanced mode it's maybe a bit too easy to modify groups, if you want to reorder something, say Bluray-2160p, it's quite easy to accidentally drop it inside the group and you actually have to do your best to move it between WEB 2160p and HDTV 2160p.
If you drag it down, but hover over the checkboxes, it's easy. but once you move over the subitems....
anyway, there are a few possible solutions for this. One could be to increase the right margin of the subitems, so you're not actually dragging it over the subitems. An alternative could be to add a 'Edit Groups' button, that expands the groups (from the simplified view suggested in 1), to the current view.
I'm not sure about the right route, 'Edit Groups' has some appeal coz it actually makes the whole thing easier.

5) Cosmetic: In the original UI we grayed out the deselected qualities to make it more apparent they're disabled.

markus101:
1. That makes sense.
2. That's going to depend entirely on screen size, at 1440p I have to scroll to see them all, but they are all there, at 1080p it's close, but two qualities are missing.
 - Moving the qualities to the right would work, but that would result in multiple drag icons close by unless we move the group drag icon to the left of the qualities, also what would we do about the non-group ones? they would be really wide with not much information.
 - Yes there is another GHI (probably more than one) that talks about/some other changes to quality profiles
3. Not opposed to another icon, we also need one for destroying the group (currently an `x`)
4. Yeah, the dragging out of a group is a bit complicated and seems to be sticking to the group even though the drop indicator is below the group, I can take a look at that. Edit group seems like an unnecessary action to do because something might get dropped in the wrong place. What does the `Edit Groups` button do to make it make easier?
5. That's unrelated to this work, but noted.

Taloth:
2. I meant giving it two columns, one of the normal fields like 'Name', and the second column to contain the quality list. It wouldn't change how the list itself looks, only where it's positioned in the dialog so it can use the full height of the dialog.

3. There's an outdent icon too (or ungroup icon).

4. 'Edit Groups' would effectively be what Advanced Settings does right now. Except that Edit Groups is focused on the groups, and not re-ordering or selecting. So the checkboxes should be hidden, for example. Edit groups could even be in a separate dialog, but I don't think that's needed, although that would allow you to add more information about what a group actually is.

markus101:
2. Not sure I follow what you mean.
4. I do like the idea of not having it tied to the advanced feature since you can't toggle it without leaving the modal. Would the view be different in Edit Groups mode or would that allow drag and drop into groups? Or something else? How would I add/remove items from a group?

Taloth:
2. You mean that if we reduce the margins and such it'll put the drag handles closer together?
There are two things here: we can tweak the style a bit, reduce margins, smaller checkbox, things like that.
And we can change the layout of the dialog, which doesn't affect how the drag handles appear at all.
Like:
image
You need the additional whitespace between label and drag handle for the subgroup badges.

4. I think the view should be different, mockups:
image
I'm not sure about how to drag from outside group to inside. in the mockup you'll have to 'indent' then drag it from one group to the other, but i'm not sure if that's intuitive enough.
It's also possible to simply keep the reorder-icons for top level items, allowing them to be dragged into groups, but when I was doing the mockup that felt confusing.
An alternative could be to show 'group' icons between the items like 'group these two together'.
A final alternative would be to add temporary selection checkboxes _outside_ the items to the left, with a 'group selected items' button. In that case you could ensure that items can only be dragged _within_ their own group, and you can remove the group.

markus101:
2. I was envisioning only moving the qualities within a group to the side so you'd have wider columns and start the qualities beside the name, which is how I saw the two drag icons getting beside each other.

3. (placeholder)


4. This all sounds a lot more complicated that simple drag and drop with the group/ungroup buttons to create a group in the first place. I'm not seeing how this helps.

The `indent` button would do something, but it seems magically what it;'d do, which group would it be added to, would it become it's own group, etc.

> An alternative could be to show 'group' icons between the items like 'group these two together'.

So if they're next to each to each other there would be an icon between the two (a bit like the xem linking between entities, if you've used that?) I'm not sure how discoverable that would be, I'd probably have to get lucky and hover between the two to see it.

> A final alternative would be to add temporary selection checkboxes outside the items to the left, with a 'group selected items' button.

I thought about something like that, but with a check being an indicator that the qualitiy is wanted that seems awkward.

A couple other ideas:

We could get rid of the check boxes and simply change the colour of the field (and maybe show an icon to indicate it's wanted). We could show a checkbox/something else selectable for grouping. If you selected 2 or more qualities we'd show/enable a `group` button and create a group out of those qualities. If the qualities are within a group they can be be reordered within, and there would be a button to destroy the group (and remove all items) as well as a button to remove a single item from the group, or maybe just the latter and you destroy the group by removing all the items (and we could validate that a group has at least 2 items in it). We'd also need a way to add items to an existing group, I'm not sure what that looks like though.

Taloth:
> 2. I was envisioning only moving the qualities within a group to the side

Ah, wouldn't that get problematic if a group gets bigger than 2? And I'm okay with scrolling when editing the groups, coz I think that's unavoidable. I just want the basic view to be scrolless.

> 4. This all sounds a lot more complicated that simple drag and drop

First thing first: Get rid of the 'advanced settings' and replace it with a 'Edit Quality Groups' checkbox/toggle button. And make it mimic the current behavior of advanced settings.

> The indent button would do something, but it seems magically what it;'d do, which group would it be added to, would it become it's own group, etc.

Yup, atm it would be it's own group, but it isn't exactly clear.

> So if they're next to each to each other there would be an icon between the two

Yes, and those icons would be visible permanently in 'Edit Quality Groups' mode, so no guessing for the user. While hovering over the already visible icon, you could show a 'bracket' around the two items about to be grouped, if that improves things, but I think it will be clear enough for the user if we position the icons correctly.
I think we should make a mockup of this. Either this or the checkbox scenario might be the most intuitive.

> I thought about something like that, but with a check being an indicator that the qualitiy is wanted that seems awkward.

That's why I wanted those selector checkboxes to be in front of the rows instead of inside.

> We could get rid of the check boxes and simply change the colour of the field (and maybe show an icon to indicate it's wanted). We could show a checkbox/something else selectable for grouping. If you selected 2 or more qualities we'd show/enable a group button and create a group out of those qualities.

That's basically what I suggested.

> If the qualities are within a group they can be be reordered within, and there would be a button to destroy the group (and remove all items)

Similar to what we have now, except with the reordering being restricted to the group.

> as well as a button to remove a single item from the group, or maybe just the latter and you destroy the group by removing all the items (and we could validate that a group has at least 2 items in it).

The edge-case for sure, but not a big problem. I'd just make ppl destroy the group if they want to remove an item. It's not the primary usecase anyway. Normally you'll reorder the qualities, then select two or three and group em.

> We'd also need a way to add items to an existing group, I'm not sure what that looks like though.

You select the group and the item then group em together.

markus101:
> I just want the basic view to be scrolless.

That's going to be dependent on window size and how many qualities there are, I don't think we should optimize for that.

> I'd just make ppl destroy the group if they want to remove an item.

That sounds clunky. I'd rather have a remove button per icon and the group either dissolves when only one item is left or they click to remove the last item.

I'm not seeing how creating a group will work well unless we use a checkbox system for it, otherwise you're forced to reorder qualities, then to switch to edit groups mode, make a group and back and forth, which sounds like an awful user experience.

So what's the issue we're actually trying to solve here?
Is the drag and drop in and out of groups a problem? There is one issue dragging a quality from a group back to the main list at the same index as the group, but I have some ideas to overcome that.

Taloth:
> That's going to be dependent on window size and how many qualities there are, I don't think we should optimize for that.

Of course on a small enough screen it won't matter, but all I'm asking for is a bit of consideration for the average user. Having to scroll in this view it pretty bad for the understanding of the user.

> That sounds clunky. I'd rather have a remove button per icon and the group either dissolves when only one item is left or they click to remove the last item.

Maybe it's a bit clunky, but it would avoid having two different buttons in the same item. And I ask you to come up with an example where the user would want to remove a single item without decomposing the entire group. Again, I want to make this easy for the most common scenarios.

> So what's the issue we're actually trying to solve here?
Is the drag and drop in and out of groups a problem? There is one issue dragging a quality from a group back to the main list at the same index as the group, but I have some ideas to overcome that.

I'm going to assume we're agree on the 'Edit Groups' switch instead of 'Advanced Settings', coz that was my primary concern. And solves the 'drag and drop' issue in most cases because you can do the reordering in the default view.

The issue I wanted to address is to reduce the visual complexity of this feature. Atm we have 2 buttons per item in close proximity. You can both enable/disable, reorder qualities, and group em, and grouping them is not even that obvious. And yes, dragging the item out again isn't obvious either.
Separating it in to a Normal and Edit Group modes already helps, because in Normal mode you can enable/disable and reorder. And in Edit Group mode you can reorder, group/ungroup.

What about moving the group/ungroup icons to the left, putting them in place of the checkboxes?

markus101:
> Maybe it's a bit clunky, but it would avoid having two different buttons in the same item. And I ask you to come up with an example where the user would want to remove a single item without decomposing the entire group. Again, I want to make this easy for the most common scenarios.

Group all 720p qualities, then realize you want to ungroup BR 720p so you can upgrade to it later, maybe not super common. I think it'd be better to have a way to remove from a group without being forced to destroy it and possibly recreate it, but we can work on that.

> I'm going to assume we're agree on the 'Edit Groups' switch instead of 'Advanced Settings', coz that was my primary concern. And solves the 'drag and drop' issue in most cases because you can do the reordering in the default view.

Yeah, I think that's a good change. Now can we agree where the button should live? I feel like it should be near the qualities in the list, but not sure how that should look.

I think dragging in/out is fairly obvious if you're dragging items because there is a placeholder where the item should end up (granted the placeholder sometimes isn't where the item ends up, which needs to be addressed). Selecting items then creating a group and having a single button to remove an item from a group/destroying a group could be more intuitive though.

> What about moving the group/ungroup icons to the left, putting them in place of the checkboxes?

I think dropping the check boxes and having another indication that the quality is wanted would be good.

We can move the button as well, but I wonder if we could have a check box show up next to items in edit mode and you select them and then see a `Create Group` button, but I don't know where that button could live. I think having the buttons on the same side is fine though, it keeps the actions in one place.

Two things we can do right now:

- Edit groups button (need a place to put it)
- Remove the check boxes and use another visual indicator to indicate the quality is wanted.

From there I think we can do one of two things:

- Leave the drag and drop as-is (and fix the current issues)
- Allow items to be selected and have a button to make a group from those selected items

Questions:
- In edit mode we'd hide re-order buttons for top-level groups/qualities?
- What do we want allowed qualities to look like? Have them as-is and gray out disabled ones?

Taloth:
So based on our hangouts conversation:

- Add 'Edit Quality Groups' toggle, should definitely be near the list.
- In 'Enable/Reorder' mode, we show the checkboxes and reorder icons.
- Gray out the disabled qualities like in v2.
- We leave the drag-drop logic as is, the Edit Groups toggle already solves most of that.
- In 'Edit Quality Groups' mode, we hide the checkboxes and keep the reorder icons.
- In 'Edit Quality Groups' mode, show the + and x icons, in place of the checkboxes.
- Some minor cosmetics, like group border & background, margins.

I think at that point we're 'good enough'. If we don't like it we can always tweak it in the future, replace icons if we find something better, or tweak the drag drop logic a bit.

I suppose we could also show it to our regulars at one point, see how they react.

Taloth:
The UI looks pretty good now, but can we do a full rebase+squash of the code so the diff doesn't contain the full v3 UI change history?

markus101:
Squashed and rebased for your reviewing pleasure.

Taloth:
Didn't finish the review yet, I've been looking at it on/off for the last few days.
There are a couple of things I didn't like, but I first want to get a feel for the bigger changes before nitpicking.

Taloth:
btw. performance issue: Edit Group -> one group of WEBDL-2160p + WEBRip-2160p, drag webdl-2160p, drag+hover over WEBRIP-2160p (one item lower in the group), then move it left and right over the WEBRIP-2160p item without dropping.

src/NzbDrone.Core.Test/Datastore/Migration/117_add_webrip_qualites_in_profileFixture.cs
```diff
@@ -0,0 +1,94 @@
...
+                    Name = "SDTV",
+                    Cutoff = 1,
+                    Items = $"[{GenerateQualityJson(1, true)}, {GenerateQualityJson((int)Quality.WEBRip480p, false)}, {GenerateQualityJson((int)Quality.WEBRip720p, false)}, {GenerateQualityJson((int)Quality.WEBRip1080p, false)}, {GenerateQualityJson((int)Quality.WEBRip2160p, false)} ]"
+                });
+            });
+
+            var profiles = db.Query<Profile116>("SELECT Items FROM Profiles LIMIT 1");
+            var items = profiles.First().Items;
+
+            items[1].Items.First().Quality.Should().Be((int)Quality.WEBRip480p);
```
markus101: Yeah, it is, the UI shows bottom up, not top down, so we want the lower quality first.

src/NzbDrone.Api/Profiles/ProfileSchemaModule.cs
```diff
@@ -27,7 +27,7 @@ private List<ProfileResource> GetAll()
                 .ToList();
 
             var profile = new Profile();
-            profile.Cutoff = Quality.Unknown;
+            profile.Cutoff = Quality.Unknown.Id;
```
markus101: image

markus101:
Fixed the perf issue.

src/NzbDrone.Api/Profiles/ProfileSchemaModule.cs
Taloth: Ah, because cutoff changed from Quality to int. Does that mean int has dual meaning here: 'quality' or a group id? Wouldn't it be better then to switch to an index in the array?

markus101: Yes, it's quality or group ID. We could switch it to the index of the array, but we'd need to keep it updated whenever we re-order qualities so the UI would still need to track it by ID, but we'd store the index.

src/NzbDrone.Core/Qualities/Quality.cs
```diff
@@ -67,12 +67,12 @@ public override bool Equals(object obj)
         public static Quality HDTV1080p => new Quality(9,  "HDTV-1080p");
         public static Quality RAWHD => new Quality(10, "Raw-HD");
         //public static Quality HDTV480p    { get { return new Quality(11, "HDTV-480p"); } }
```
markus101: SDTV as a source is hardly seen anymore, adding HDTV 480p seems mostly ceremonious, but could also be confusing, since pretty much everything "SDTV" would become it.

Bluray 480p makes sense, which would include `BDRip` (direct from BR Source to 480p) and `BRRip` (re-encode from a 1080p release). What do we want to name this group? `Disk 480p` maybe?

Also what do we want to do for Remux? There is a separate issue for it, but I'd like to land it as part of this PR if we're going to join group it with BR 1080p/2160p (and people can break the group), otherwise we can add it as a separate quality or would this fall under preferred words? I realize the source is the same, but it feels different enough in quality it may warrant it's own quality.

Taloth: Not sure about the group name 'disk 480p', i'd be ok with keeping it DVD.

Remux is more like 'Remux-1080p' and 'Remux-2160p'. If we add them separately then they have their own size limits.
But I don't know which group it should be shoved under. Most people won't even want remuxes. Where previously sizelimits would make sure it didn't, now with separate qualities and thus separate limits (but the same group) it would cause undesired behavior.
It doesn't fall under preferred words, I think that's what we got out of that big debate in the other issue.

Crap, I don't know, I'm feeling like we're adding too many qualities without a means to make it easier for normal users.

src/NzbDrone.Core/Qualities/QualityModelComparer.cs
```diff
@@ -16,17 +17,45 @@ public QualityModelComparer(Profile profile)
...
 
         public int Compare(QualityModel left, QualityModel right)
         {
-            int result = Compare(left.Quality, right.Quality);
+            return Compare(left, right, false);
+        }
+
+        public int Compare(QualityModel left, QualityModel right, bool respectGroupOrder)
+        {
+            int result = Compare(left.Quality, right.Quality, respectGroupOrder);
```
markus101: > Is it valid to compare revision if quality is not exactly the same?

No, it doesn't make sense.

When comparing for ordering we're explicitly getting the "index" which is the group index + quality in group index, so we're ordering consistently and only comparing the revision/real if the quality is an exact match.

src/NzbDrone.Core/Qualities/Quality.cs
markus101: I don't think we're going too overboard with qualities, but it is more complicated than not adding them. From the initial view we're adding two qualities (remuxes), underneath we're adding quite a bit more, but if you don't care/don't look it stays pretty simple and beyond having a way to completely disable editing the groups I can't think of a way that we could keep it super simple. I don't think we're doing a disservice to normal/new users, but we are definitely making things better for advanced users.

TODO:
- Add `Remux 1080p` and `Remux 2160p`, as their own qualities
- Add `Bluray 480p` (`BDRip` and `BRRip`), Grouped with `DVD`

I'm still not sure about adding `HDTV 480p`, grouped with SDTV because this is a pretty big change since most "SD" content comes from an HDTV feed and would fall under this category, I think this might be the most confusing to normal/new users and I'm not sure they really care they just want their episodes in the smallest size they can.

src/NzbDrone.Core/Qualities/QualityModelComparer.cs
Taloth: For Ordering we need to consider the qualities in one group to be equal. That's the whole point of grouping, so that webdl & webrip are considered equal in priority unless the user breaks the group.

markus101: > For Ordering we need to consider the qualities in one group to be equal.

For upgrade purposes we need to, but respecting the group ordering to make a decision on which to download makes sense. For example: preferring WEB-DL over Webrip would grab the WEB-DL if both are seen, but wouldn't upgrade Webrip to WEB-DL even if the cutoff is higher.

markus101:
@Taloth anything else for this one?

markus101:
Squashed and merged.

PR #2010, opened 2017-06-28
Add 'Series Title, The' filename option

Neaox:
* Adds 'Series Title, The' filename option
* Updates the FileNameBuilder.SeriesTitleRegex to include new format
* Adds tests for this new format

#### Database Migration
NO

#### Description

#### Todos
- [x] Tests
- [ ] Documentation

* 371

### DIFF
--- a/src/NzbDrone.Core/Organizer/FileNameBuilder.cs
+++ b/src/NzbDrone.Core/Organizer/FileNameBuilder.cs
@@ -293,7 +293,7 @@ namespace NzbDrone.Core.Organizer
         {
             tokenHandlers["{Series Title}"] = m => series.Title;
             tokenHandlers["{Series CleanTitle}"] = m => CleanTitle(series.Title);
-            tokenHandlers["{Series Title The}"] = m => TitleThe(series.Title);
+            tokenHandlers["{Series TitleThe}"] = m => TitleThe(series.Title);
         }
 
         private string AddSeasonEpisodeNumberingTokens(string pattern, Dictionary<string, Func<TokenMatch, string>> tokenHandlers, List<Episode> episodes, NamingConfig namingConfig)
--- a/src/NzbDrone.Core/Organizer/FileNameBuilder.cs
+++ b/src/NzbDrone.Core/Organizer/FileNameBuilder.cs
@@ -67,7 +67,7 @@ namespace NzbDrone.Core.Organizer
 
         private static readonly char[] EpisodeTitleTrimCharacters = new[] { ' ', '.', '?' };
 
-        private static readonly string[] TitlePrefixes = { "The ", "An ", "A " };
+        private static readonly Regex TitlePrefixRegex = new Regex(@"^(The|An|A) (.*?)( \(.*?\))?$", RegexOptions.Compiled | RegexOptions.IgnoreCase);
 
         public FileNameBuilder(INamingConfigService namingConfigService,
                                IQualityDefinitionService qualityDefinitionService,
@@ -257,16 +257,7 @@ namespace NzbDrone.Core.Organizer
 
         public static string TitleThe(string title)
         {
-            foreach (string prefix in TitlePrefixes)
-            {
-                if (title.StartsWith(prefix, StringComparison.OrdinalIgnoreCase))
-                {
-                    title = title.Substring(prefix.Length) + ", " + prefix.Trim();
-                    break;
-                }
-            }
-
-            return title.Trim();
+            return TitlePrefixRegex.Replace(title, "$2, $1$3");
         }
 
         public static string CleanFileName(string name, bool replace = true)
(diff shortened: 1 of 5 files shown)


src/NzbDrone.Core/Organizer/FileNameBuilder.cs
```diff
@@ -253,6 +253,22 @@ public static string CleanTitle(string title)
             return title;
         }
 
+        public static string TitleThe(string title)
+        {
+            string[] prefixes = { "The ", "An ", "A " };
```
Taloth: Lets make that static somewhere.

src/NzbDrone.Core/Organizer/FileNameBuilder.cs
```diff
@@ -253,6 +253,22 @@ public static string CleanTitle(string title)
...
+            foreach (string prefix in prefixes)
+            {
+                int prefix_length = prefix.Length;
+                if (prefix.ToLower() == title.Substring(0, prefix_length).ToLower())
```
Taloth: Failing testcase: (always make negative testcases too)

```
        [TestCase("A")]
        [TestCase("Anne")]
        [TestCase("Theodore")]
        [TestCase("3%")]
        public void should_not_change_title(string title)
        {
            _series.Title = title;
            _namingConfig.StandardEpisodeFormat = "{Series TitleThe}";

            Subject.BuildFileName(new List<Episode> { _episode }, _series, _episodeFile)
                   .Should().Be(title);
        }
```

Taloth: And we got something called `string.StartsWith(..., StringComparison.OrdinalIgnoreCase)`

src/NzbDrone.Core/Organizer/FileNameBuilder.cs
```diff
@@ -277,6 +293,7 @@ private void AddSeriesTokens(Dictionary<string, Func<TokenMatch, string>> tokenH
         {
             tokenHandlers["{Series Title}"] = m => series.Title;
             tokenHandlers["{Series CleanTitle}"] = m => CleanTitle(series.Title);
+            tokenHandlers["{Series Title The}"] = m => TitleThe(series.Title);
```
Taloth: I prefer `TitleThe` over `Title The` because it otherwise creates ambiguity about the chosen separator.

src/NzbDrone.Core/Organizer/FileNameBuilder.cs
Neaox: Of course `string.StartsWith` is a better option, I would have used it had I wrote the code from scratch. I copied and pasted this code from Radarr. No excuse for not reviewing the code correctly myself, I should have put more than the minimum amount of effort in, my apologies.

Taloth:
I was about to squash and merge it in, but I was checking the original issue and...

How does it behave with "The Flash (2010)" would be "Flash, The (2010)"?

I think atm it'll spit out `Flash (2010), The`

Some testcases I pulled from Skyhook/Tvdb for your convenience, I think i covered all the patterns I saw in the database:
```
The Flash (2010)
A League Of Their Own (AU)
The Fixer (ZH) (2015)
The Sixth Sense 2 (Thai)
The Amazing Race (Latin America)
The Rat Pack (A&E)
```

I'd do a regex that avoid () somewhere in the middle, there aren't any cases with an article at the start, but it's better to be safe than sorry, example: `I (Almost) Got Away With It` -> `The Challenge: I (Almost) Got Away With It` -> `Challenge: I (Almost) Got Away With It, The`
Heck, you can probably roll the entire logic in a single Compiled Regex, and then use TitleTheRegex.Replace(..) as a single statement in your TitleThe method.

If you can't get the regex right lemme know, I'm fairly adept at those things.

Neaox:
I made the requested change to support the preservation of title suffixes. I generally consider myself pretty versed in regex, but if you have any suggestions beyond my pattern, by all means.

Taloth:
It looks pretty good to me, doesn't have any stray whitespace.
However i'd change the () section to `((?: *\([^)]+\))*)$` It'll prevent cases like `The Climax: I (Almost) Got Away With It (2016)` from being mishandled.
The ` *`is to gobble up optional whitespace, in case people put down (UK)(2016) without a space. I didn't see that in the database, but you never know, tvdb can be a mess at times.

Neaox:
No worries. I like that your standards are so high. Thanks for being patient with me.

M0UL:
that pulls request does work with clean titles

Taloth:
No, only with the standard title.

Neaox:
Thanks for merging this in @Taloth when is the next release likely to be... released? Thanks!

M0UL:
this is already available on the develop branch

Taloth:
@M0UL  @thezoggy Can you give me a reason why 'Series CleanTitleThe' should exist as well? My reason was that CleanTitle is often used with the scene dot-notation, and thus moving articles to the end wouldn't be very useful in that case.

M0UL:
mostly anime for example The Ancient Magus' Bride

PR #2016, opened 2017-07-04
WIP: Improve functionality and testability of Normalize and CleanTitle.

Taloth:
At & Percentage sign handling, and general cleanup.
I added quite a bit of testcases, but I change the behavior as well.

What I want to do is create several NormalizeTitle variants.
- NormalTitle(string title)
- CleanTitle(string title)
- ScenifyTitle(string title)

The idea is that the ScenifyTitle is tailored for the scene dot-notation, but otherwise much the same to CleanTitle. This gives different results for certain series like Agents of Shield.

#### Todos
- [ ] TBD

* TBD


src/NzbDrone.Core/Parser/NormalizeParsedTitle.cs
```diff
@@ -0,0 +1,69 @@
...
+        private static readonly Regex DuplicateSpacesRegex = new Regex(@"\s{2,}", RegexOptions.Compiled);
+        private static readonly Regex SpecialEpisodeWordRegex = new Regex(@"\b(part|special|edition|christmas)\b\s?", RegexOptions.IgnoreCase | RegexOptions.Compiled);
+
+        public static string CleanSeriesTitle(this string title)
+        {
+            long number = 0;
+
+            //If Title only contains numbers return it as is.
+            if (long.TryParse(title, out number))
+                return title;
```
markus101: I know this is copy/paste, but braces please!

Taloth:
I think foremost what we need is to make a decision on what we want renamed and how. And some examples, so I plan to run the logic over skyhook's database and list the differences between before and after. Then have a RFC with some of our regulars.

PR #2039, opened 2017-07-11
Store Plex Auth Token in DB

markus101:
#### Database Migration
NO

#### Description
No longer stores the plex username/password in the DB and instead will store an auth token. This will force users updating the Plex Media Server connection to authenticate so the token is stored instead and in v3 we can force remove those settings and rely only on the stored token. Additionally we can investigate using the PIN system to authenticate instead of/addition to username and password.


Taloth:
I think the 'Persisted' logic is unnecessary. Wouldn't it be better to do it like Captcha? with a 'OAuth' Field type. That way you can present the appropriate oauth window or possibly do the Pin check.

Issue #289, opened 2014-05-13
Use metadata files when importing existing series to improve matches


SonarrBot:
Match probability, based on metadata files and title matching (simple title).

Use series metadata to match shows
- Scan for folders
- Check each one for metadata
- Grab series ID from file
- Use series ID and grab info from trakt
- Confirmed shows can be added

cryodream:
This is the most annoying thing about Sonarr. I have almost 900+ shows and at the same time I do not spaces in the folder names. At the moment I am testing Sonarr on unraid (thinking of switching my server to run unraid). To makes things super easy, I used Directory Opus with to filter on tvshow.nfo and copied all of them automaticly recreating the folders. This way I had a copy of all tv shows folders with tvshow.nfo meta file in them. All I had to do was point the new sonarr to the folder and import from disk.
Or so I thought. All I got was a long list of searches which find nothing, because of the dots instead of spaces in the folder names! It never even entered my mind, that Sonarr does not use tvshow.nfo files! Not to mention, that most of them were created by the sonarr itself. I'm sorry, but WTF?

Checking if tvshow.nfo is present and trying to deserialize the show ID is like 3 lines of code...

markus101:
> Checking if tvshow.nfo is present and trying to deserialize the show ID is like 3 lines of code...

Its definitely not, hyperbole aside, if we support tvshow.nfo we want to support every other metadata type that we support. We also accept PRs.

cryodream:
@markus101 sorry for that "3 lines" overreach.
I was just really surprised you do not read meta files, while you do write them...
I see you do not use serialization for writing xml. I'm sure there's a reason, just interesting, why not?

markus101:
We use XmlWriter: https://github.com/Sonarr/Sonarr/blob/develop/src/NzbDrone.Core/Metadata/Consumers/Xbmc/XbmcMetadata.cs#L177

Kodi episode metadata is a bit tricky because Kodi expects invalid XML(multiple root elements) when multi episodes exist in a single file, but series metadata is pretty standard.

cryodream:
@markus101 thanks for explanation.

Then maybe I can suggest a temporary simpler solution: to replace the non alphanumeric chars with spaces in the folder name when searching.
[...]

PR #2096, opened 2017-08-03
Add support of nfo when "Import existing series on Disk"

j0nathan33:
#### Database Migration
NO

#### Description
Add some code to read nfo when do "Importing existing series on Disk"

#### Todos
- [x] Tests
- [x] Documentation

Fix issues #289


src/NzbDrone.Core/RootFolders/RootFolderService.cs
```diff
@@ -153,7 +154,34 @@ private List<UnmappedFolder> GetUnmappedFolders(string path)
             foreach (string unmappedFolder in unmappedFolders)
             {
                 var di = new DirectoryInfo(unmappedFolder.Normalize());
-                results.Add(new UnmappedFolder { Name = di.Name, Path = di.FullName });
+                var fileNfoName = String.Format("{0}\\tvshow.nfo", di.FullName);
```
markus101: Use `Path.Combine` to join paths, which works on different operating systems, currently this would break on anything other than Windows.

markus101: Use `Path.Combine` to join paths, which works on different operating systems, currently this would break on anything other than Windows.

src/NzbDrone.Core/RootFolders/RootFolderService.cs
```diff
@@ -153,7 +154,34 @@ private List<UnmappedFolder> GetUnmappedFolders(string path)
...
+                if (File.Exists(fileNfoName))
```
markus101: Use `DiskProvider.FileExists` so it can be mocked for unit tests.

src/NzbDrone.Core/RootFolders/RootFolderService.cs
```diff
@@ -153,7 +154,34 @@ private List<UnmappedFolder> GetUnmappedFolders(string path)
...
+                        doc.Load(File.Open(fileNfoName, FileMode.Open));
+                        XmlNode node = doc.DocumentElement.SelectSingleNode("/tvshow/id");
+                        if (node == null)
+                        {
+                            results.Add(new UnmappedFolder { Name = di.Name, Path = di.FullName });
+                        }
+                        else
+                        {
+                            String tvid = node.InnerText;
+                            results.Add(new UnmappedFolder { Name = String.Format("tvdb:{0}", tvid), Path = di.FullName });
```
markus101: The name should still be the name, any additional information should be returned via a separate property, which also means the API and UI will need to be updated to pull in that information.

src/NzbDrone.Core/RootFolders/RootFolderService.cs
```diff
@@ -153,7 +154,34 @@ private List<UnmappedFolder> GetUnmappedFolders(string path)
             foreach (string unmappedFolder in unmappedFolders)
             {
                 var di = new DirectoryInfo(unmappedFolder.Normalize());
-                results.Add(new UnmappedFolder { Name = di.Name, Path = di.FullName });
+                var fileNfoName = String.Format("{0}\\tvshow.nfo", di.FullName);
+                if (File.Exists(fileNfoName))
+                {
+                    try
```
markus101: The logic of pulling the ID from an NFO file should be pulled out to a separate method, possibly another class/interface so additional metadata types could be also be handled. It'd be worth looking at extending the existing metadata handling to be able to check for an existing file and check that file for an ID if the file doesn't exist or doesn't support storing an an ID Sonarr can use for a lookup it would return null.

j0nathan33:
I don't know why junit fail when I only edit javascript file

src/UI/AddSeries/Existing/AddExistingSeriesCollectionView.js
```diff
@@ -30,9 +30,12 @@ module.exports = Marionette.CompositeView.extend({
 
         if (model) {
             var currentIndex = index;
-            var folderName = model.get('folder').name;
+            var termSearch = model.get('folder').name;
+            if (typeof model.get('folder').id !== "undefined") {
```
markus101: `model.get('folder').id != null` (instead of the `!==` you had before as `!=` will catch undefined and null.

src/NzbDrone.Core/RootFolders/UnmappedFolder.cs
```diff
@@ -4,5 +4,6 @@ public class UnmappedFolder
     {
         public string Name { get; set; }
         public string Path { get; set; }
+        public string Id { get; set; }
```
markus101: `Id` is reserved for the `Id` of the record, `TvdbId` would be more accurate since that assumption is being made in the client. It should also be an integer instead of a string. Probably a nullable int so we don't return 0 if it's unknown.

src/NzbDrone.Core/RootFolders/RootFolderService.cs
```diff
@@ -129,6 +130,30 @@ public void Remove(int id)
...
+            if (_diskProvider.FileExists(fileNfoName))
+            {
+                try
+                {
+                    XmlDocument doc = new XmlDocument();
+                    doc.Load(File.Open(fileNfoName, FileMode.Open));
+                    XmlNode node = doc.DocumentElement.SelectSingleNode("/tvshow/id");
+                    if (node != null)
+                    {
+                        returnValue = node.InnerText;
```
markus101: Why not return directly from here? This should also parse it to an integer to ensure it's valid.

markus101:
After talking with @Taloth there are a couple things we want to improve here:

1. The logic for pulling the TVDB ID needs to be moved outside of the `RootFolderService`
2. The host should tell the client what to search for, that is, if it has a TVDB ID, it should set the search term to `tvdbid:12345` and if it doesn't it should return the folder name, the client will then use the search term from the server and not decide which to use on it's own

PR #2104, opened 2017-08-06
Fix inconsistent naming of qBittorrent

jamesmacwhite:
#### Database Migration
NO

#### Description
I noticed that the naming of "qBittorrent" in exception/log messages is a little inconsistent. This updates all the references to the correct name of the application.

### DIFF
--- a/src/NzbDrone.Core/Download/Clients/QBittorrent/QBittorrent.cs
+++ b/src/NzbDrone.Core/Download/Clients/QBittorrent/QBittorrent.cs
@@ -120,7 +120,7 @@ namespace NzbDrone.Core.Download.Clients.QBittorrent
                 {
                     case "error": // some error occurred, applies to paused torrents
                         item.Status = DownloadItemStatus.Failed;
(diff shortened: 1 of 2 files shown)


src/NzbDrone.Core/Download/Clients/QBittorrent/QBittorrent.cs
```diff
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
```
markus101: GitExtensions is showing me the `u` changed, but I see no difference in Visual Studio. What's up with these two changes?

Taloth: probably the BOM.

jamesmacwhite: Hi. I noticed this. Likely do with my editor/git settings? I've just gone back and rebased. It appears to be fixed now.

Issue #2115, opened 2017-08-10
Add Hebrew Language


rotem925:
Please add support for Hebrew language to the profiles.
Usually it is something like this:
Smurfs.​The.​Lost.​Village.​2017.​1080p.​BluRay.​REMUX.​HebDub.​x264-​iSrael

The HebDub is what matter

markus101:
Is the same `HebDub` used for TV show? As you refer to a movie title here.

rotem925:
Ohh, sorry, you are right. HebDub is for movies. 
I was sure that had to like a movie.
Just checked Israeli torrent files, They put nothing there:
Metim.Lerega.S02E30.720p.HDTV.x264-Yonidan

rotem925:
Ok,
Checked again,
If it's not an Hebrew movie that is dubbed, then HebDub is correct:

Avatar.The.Last.Airbender.S01-03.DVDRip.HebDub.

So I guess the Hebrew language just have to support HebDub

PR #2116, opened 2017-08-11
Added Hebrew lanugage

rotem925:
#### Database Migration
NO

#### Description
Added support for Hebrew language.

#### Todos
- [x] Tests

* #2115


rotem925:
I've made the same change for Radarr.

markus101:
What was the purpose of this reversion? https://github.com/Radarr/Radarr/pull/1922

rotem925:
I've added it to Radarr... what do you mean re-version? If you referring the remove of the iso.
It is just to disable the search by Hebrew language, since most of the movies/tv shows are still in english and does not come up in the search result of the torrents when they appear in Hebrew.

markus101:
Sorry, I meant reverting. Does Radarr use the ISO Language when searching? We don't in Sonarr, but I want to make sure there wasn't a reason for us to also change that.

rotem925:
@markus101 Yes, I the search results come in Hebrew in Radarr, that what it looked like. After I've removed the the iso and checked, then I could find torrents. If not the same for Sonarr, then this could stay... Thank you for taking care of this.

markus101:
Merged in https://github.com/Sonarr/Sonarr/commit/7d21585f50cc57d4e6c6048584150a2114bb9530

PR #2119, opened 2017-08-12
Remove pending items if the release is no longer on the indexer.

Taloth:
If the indexer returns 404 we throw a ReleaseUnavailableException which causes the release to be put on the Rejected list instead. That should prevent the release from being stuck on the pending queue forever.

### DIFF
--- a/src/NzbDrone.Core/Download/ProcessDownloadDecisions.cs
+++ b/src/NzbDrone.Core/Download/ProcessDownloadDecisions.cs
@@ -6,6 +6,7 @@ using NLog;
 using NzbDrone.Core.DecisionEngine;
 using NzbDrone.Core.Download.Clients;
 using NzbDrone.Core.Download.Pending;
+using NzbDrone.Core.Exceptions;
 using NzbDrone.Core.Indexers;
 
 namespace NzbDrone.Core.Download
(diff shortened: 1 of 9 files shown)


src/NzbDrone.Core/Download/ProcessDownloadDecisions.cs
```diff
@@ -74,6 +76,11 @@ public ProcessedDecisions ProcessDecisions(List<DownloadDecision> decisions)
                     _downloadService.DownloadReport(remoteEpisode);
                     grabbed.Add(report);
                 }
+                catch (ReleaseUnavailableException)
+                {
+                    _logger.Warn("Failed to download release from indexer, no longer available. " + remoteEpisode);
```
markus101: Is there a reason not to use `_logger.Warn("Failed to download release from indexer, no longer available. {0}", remoteEpisode);`?

Taloth: No reason, just copy paste from the cache (Exception ex) that comes after.

markus101:
Somewhat related issue (so we don't forget): https://forums.sonarr.tv/t/rss-feed-grabs-over-3000-never-finishes-causing-excessive-cpu-use/15809

Almost 2000 torrent releases without a torrent indexer/download client so they just sit there and are continually processed, which is slow due to the issue you pointed out. We should clean up pending releases when an indexer is removed (along with a housekeeping task).

markus101:
Was this completed when the pending release fixes landed?

Taloth:
closed by https://github.com/Sonarr/Sonarr/commit/20af2c8c0f18c1f1d3f8d915e04e3401ea3684f0

Yup was rebased and merged so didn't autoclose.

PR #2136, opened 2017-08-16
[WIP] Improve ARIA accessibility on icons and other UI elements

jamesmacwhite:
#### Database Migration
NO

#### Description

I noticed a fair bulk of the icon elements are other UI components could benefit from a bit of an accessibility improvement using `aria-hidden="true"` in places and also adding missing `title` attributes along with other areas, so I was wondering if what I'm working on here would be something accepted into the project? I'm aware there is a rework of the Sonarr UI, so I haven't gone too far, but done enough to showcase my purpose of this PR in a general sense.

In addition, while investigating icon based elements, I have also fixed up a few areas of the code for readability.

Please let me know any feedback on this and if its worth me continuing it.

### DIFF
--- a/src/UI/Activity/ActivityLayoutTemplate.hbs
+++ b/src/UI/Activity/ActivityLayoutTemplate.hbs
@@ -1,7 +1,19 @@
 <ul class="nav nav-tabs">
-    <li><a href="#queue" class="x-queue-tab no-router">Queue</a></li>
-    <li><a href="#history" class="x-history-tab no-router">History</a></li>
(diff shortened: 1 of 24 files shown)


jamesmacwhite:
I'll close, this as I know the answer ;)

markus101:
We'd definitely support adding these changes to the v3 UI once it's in a state where we're ready for testing and fixes before release though. Not sure when that will be though.

jamesmacwhite:
@markus101 Good to know. I'll certainly be willing to assess the V3 UI to incorporate them :).

Issue #2098, opened 2017-08-04
delay import with episodes listed as TBD // netflix


thezoggy:
episodes for netflix shows are in tvdb as TBD until the show airs.. and then what ends up happening is that sonarr doesnt refresh the series to get said episode names until AFTER the show gets ripped.. thus you end up with TBD names and have to go back and tell sonarr to rename. 

either can we have sonarr know this scenario and update the show more aggressively (6 hours?) or offer a way to delay importing files that have TBD listed as the episode name inside sonarr. that way when sonarr finally does update the series it will get imported.. however long that may be. 

as we would want the files to be grabbed.. they are valid and we dont want things to get dmca'ed / missing out on seeding / etc just because sonarr is lacking the metadata.

thezoggy:
thanks!

PR #2140, opened 2017-08-19
New: Delay import of episodes without titles temporarily

markus101:
#### Database Migration
NO

#### Description
Rejects importing episodes when the title is unknown and it recently aired (for up 24 hours) and will attempt to refresh the series every hour until a title is available. Open to tweaking the maximum delay and refetch intervals as we see fit.

#### Todos
- [x] Tests
- [ ] Documentation

* #2098

### DIFF
--- a/src/NzbDrone.Core.Test/MediaFiles/EpisodeImport/Specifications/EpisodeTitleSpecificationFixture.cs
+++ b/src/NzbDrone.Core.Test/MediaFiles/EpisodeImport/Specifications/EpisodeTitleSpecificationFixture.cs
@@ -57,28 +57,12 @@ namespace NzbDrone.Core.Test.MediaFiles.EpisodeImport.Specifications
         }
 
         [Test]
-        public void should_reject_when_title_is_a_placeholder()
+        public void should_accept_when_did_not_air_recently_but_title_is_TBA()
         {
-            _localEpisode.Episodes.First().Title = $"Episode {_localEpisode.Episodes.First().EpisodeNumber}";
+            _localEpisode.Episodes.First().AirDateUtc = DateTime.UtcNow.AddDays(-7);
+            _localEpisode.Episodes.First().Title = "TBA";
 
-            Subject.IsSatisfiedBy(_localEpisode, null).Accepted.Should().BeFalse();
-        }
-
-        [Test]
-        public void should_reject_when_title_is_a_two_digit_placeholder()
-        {
-            _localEpisode.Episodes.First().Title = $"Episode {_localEpisode.Episodes.First().EpisodeNumber:00}";
-
-            Subject.IsSatisfiedBy(_localEpisode, null).Accepted.Should().BeFalse();
-        }
-
-        [Test]
-        public void should_accept_when_did_not_air_recently_but_title_is_null()
-        {
-            _localEpisode.Episodes.First().AirDateUtc = DateTime.UtcNow.AddDays(7);
-            _localEpisode.Episodes.First().Title = null;
-
-            Subject.IsSatisfiedBy(_localEpisode, null).Accepted.Should().BeFalse();
+            Subject.IsSatisfiedBy(_localEpisode, null).Accepted.Should().BeTrue();
         }
     }
 }
--- a/src/NzbDrone.Core.Test/MediaFiles/EpisodeImport/Specifications/EpisodeTitleSpecificationFixture.cs
+++ b/src/NzbDrone.Core.Test/MediaFiles/EpisodeImport/Specifications/EpisodeTitleSpecificationFixture.cs
@@ -1,0 +1,84 @@
+﻿using System;
+using System.Linq;
+using FizzWare.NBuilder;
+using FluentAssertions;
+using NUnit.Framework;
+using NzbDrone.Core.MediaFiles.EpisodeImport.Specifications;
+using NzbDrone.Core.Parser.Model;
+using NzbDrone.Core.Test.Framework;
+using NzbDrone.Core.Tv;
+using NzbDrone.Test.Common;
+
+namespace NzbDrone.Core.Test.MediaFiles.EpisodeImport.Specifications
+{
+    [TestFixture]
+    public class EpisodeTitleSpecificationFixture : CoreTest<EpisodeTitleSpecification>
+    {
+        private Series _series;
+        private LocalEpisode _localEpisode;
+
+        [SetUp]
+        public void Setup()
+        {
+            _series = Builder<Series>.CreateNew()
+                                     .With(s => s.SeriesType = SeriesTypes.Standard)
+                                     .With(s => s.Path = @"C:\Test\TV\30 Rock".AsOsAgnostic())
+                                     .Build();
+
+            var episodes = Builder<Episode>.CreateListOfSize(1)
+                                           .All()
+                                           .With(e => e.SeasonNumber = 1)
+                                           .With(e => e.AirDateUtc = DateTime.UtcNow)
+                                           .Build()
+                                           .ToList();
+
+            _localEpisode = new LocalEpisode
+                                {
+                                    Path = @"C:\Test\Unsorted\30 Rock\30.rock.s01e01.avi".AsOsAgnostic(),
+                                    Episodes = episodes,
+                                    Series = _series
+                                };
+        }
+
+        [Test]
+        public void should_reject_when_title_is_null()
+        {
+            _localEpisode.Episodes.First().Title = null;
+
+            Subject.IsSatisfiedBy(_localEpisode, null).Accepted.Should().BeFalse();
+        }
+
+        [Test]
+        public void should_reject_when_title_is_TBA()
+        {
+            _localEpisode.Episodes.First().Title = "TBA";
+
+            Subject.IsSatisfiedBy(_localEpisode, null).Accepted.Should().BeFalse();
+        }
+
+        [Test]
+        public void should_reject_when_title_is_a_placeholder()
+        {
+            _localEpisode.Episodes.First().Title = $"Episode {_localEpisode.Episodes.First().EpisodeNumber}";
+
+            Subject.IsSatisfiedBy(_localEpisode, null).Accepted.Should().BeFalse();
+        }
+
+        [Test]
+        public void should_reject_when_title_is_a_two_digit_placeholder()
+        {
+            _localEpisode.Episodes.First().Title = $"Episode {_localEpisode.Episodes.First().EpisodeNumber:00}";
+
+            Subject.IsSatisfiedBy(_localEpisode, null).Accepted.Should().BeFalse();
+        }
+
+        [Test]
+        public void should_accept_when_did_not_air_recently_but_title_is_null()
+        {
+            _localEpisode.Episodes.First().AirDateUtc = DateTime.UtcNow.AddDays(7);
+            _localEpisode.Episodes.First().Title = null;
+
+            Subject.IsSatisfiedBy(_localEpisode, null).Accepted.Should().BeFalse();
+        }
+    }
+}
--- a/src/NzbDrone.Core.Test/MediaFiles/EpisodeImport/Specifications/EpisodeTitleSpecificationFixture.cs
+++ b/src/NzbDrone.Core.Test/MediaFiles/EpisodeImport/Specifications/EpisodeTitleSpecificationFixture.cs
@@ -1,9 +1,10 @@
-﻿using System;
+using System;
 using System.Linq;
 using FizzWare.NBuilder;
 using FluentAssertions;
 using NUnit.Framework;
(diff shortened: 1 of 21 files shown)


src/NzbDrone.Core.Test/MediaFiles/EpisodeImport/Specifications/EpisodeTitleSpecificationFixture.cs
```diff
@@ -0,0 +1,84 @@
...
+        {
+            _localEpisode.Episodes.First().Title = "TBA";
+
+            Subject.IsSatisfiedBy(_localEpisode, null).Accepted.Should().BeFalse();
+        }
+
+        [Test]
+        public void should_reject_when_title_is_a_placeholder()
+        {
+            _localEpisode.Episodes.First().Title = $"Episode {_localEpisode.Episodes.First().EpisodeNumber}";
```
Taloth: No, should accept, coz these are valid titles and there are countless of series that do this.

src/NzbDrone.Core.Test/MediaFiles/EpisodeImport/Specifications/EpisodeTitleSpecificationFixture.cs
```diff
@@ -0,0 +1,84 @@
...
+            Subject.IsSatisfiedBy(_localEpisode, null).Accepted.Should().BeFalse();
+        }
+
+        [Test]
+        public void should_accept_when_did_not_air_recently_but_title_is_null()
+        {
+            _localEpisode.Episodes.First().AirDateUtc = DateTime.UtcNow.AddDays(7);
+            _localEpisode.Episodes.First().Title = null;
+
+            Subject.IsSatisfiedBy(_localEpisode, null).Accepted.Should().BeFalse();
```
Taloth: "should_accept" and BeFalse()? Should be BeTrue()

src/NzbDrone.Core/MediaFiles/EpisodeImport/Specifications/EpisodeTitleSpecification.cs
```diff
@@ -0,0 +1,73 @@
...
+
+                if (title.Equals("TBA"))
+                {
+                    _logger.Debug("Episode has a TBA title and recently aired");
+                    _commandQueueManager.Push(new RefreshSeriesCommand(seriesId));
+
+                    return Decision.Reject("Episode has a TBA title and recently aired");
+                }
+
+                if (title.Equals($"Episode {episodeNumber}", StringComparison.InvariantCultureIgnoreCase))
```
Taloth: So remove these two cases.

src/NzbDrone.Core/MediaFiles/EpisodeImport/Specifications/EpisodeTitleSpecification.cs
```diff
@@ -0,0 +1,73 @@
...
+                if (airDateUtc.HasValue && airDateUtc.Value.Before(DateTime.UtcNow.AddDays(-1)))
+                {
+                    _logger.Debug("Episode aired more than 1 day ago");
+                    continue;
+                }
+
+                if (title.IsNullOrWhiteSpace())
+                {
+                    _logger.Debug("Episode does not have a title and recently aired");
+                    _commandQueueManager.Push(new RefreshSeriesCommand(seriesId));
```
Taloth: I don't think we should refresh, just delay the import.
If you're going to refresh, then you should do it synchronous and check the results again.
If you're adamant about refreshing, then use a single threshold: 
- Define a constant: 1 h
- If we're < 1 h since airdate. Do not refresh. Delay/Reject import.
- If we're >= 1 h since airdate, and lastsync < 1 h since airdate. Refresh, and test again.
- If we're >= 1 h since airdate, and lastsync >= 1 h since airdate. Do not refresh. (so we only forcible refresh once)

That way we only refresh once. I still don't like it, coz it causes all those instances to refresh the same series at the same time. The effect on skyhook could be noticeable.

src/NzbDrone.Core/MediaFiles/EpisodeImport/Specifications/EpisodeTitleSpecification.cs
```diff
@@ -0,0 +1,73 @@
...
+
+                    return Decision.Reject("Episode does not have a title and recently aired");
```
Taloth: We should only reject if this is CDH (so it gets retried), if triggered via api we should import even if TBA coz we won't get another retry.

Also update the message so it's clear it's being delayed. Perhaps Decision.Delay(..) so we can show that on the UI?

src/NzbDrone.Core/Tv/ShouldRefreshSeries.cs
```diff
@@ -28,9 +28,9 @@ public bool ShouldRefresh(Series series)
                 return true;
             }
 
-            if (series.LastInfoSync >= DateTime.UtcNow.AddHours(-6))
+            if (series.LastInfoSync >= DateTime.UtcNow.AddHours(-1))
```
Taloth: No reason to do this. You're not using this interface in the single refresh since you added the if explicitly.
There's also no reason to lower this threshold for all series.

src/NzbDrone.Core/Tv/SeriesEditedService.cs
```diff
@@ -1,24 +1,23 @@
...
-            _commandQueueManager = commandQueueManager;
+            _diskScanService = diskScanService;
         }
 
         public void Handle(SeriesEditedEvent message)
         {
             if (message.Series.SeriesType != message.OldSeries.SeriesType)
             {
-                _commandQueueManager.Push(new RefreshSeriesCommand(message.Series.Id));
+                _diskScanService.Scan(message.Series);
```
Taloth: Euh, no... if series type changes then a RefreshSeries is mandatory since the refresh logic has series type specific behavior.

src/NzbDrone.Core/Tv/RefreshSeriesService.cs
```diff
@@ -146,20 +146,26 @@ private List<Season> UpdateSeasons(Series series, Series seriesInfo)
...
+            var manualTrigger = message.Trigger == CommandTrigger.Manual;
+            _eventAggregator.PublishEvent(new SeriesRefreshStartingEvent(manualTrigger));
 
             if (message.SeriesId.HasValue)
             {
                 var series = _seriesService.GetSeries(message.SeriesId.Value);
-                RefreshSeriesInfo(series);
+
+                // Only refresh automatically if the triggering manually or refreshed more than 1 hour ago
+                if (manualTrigger || series.LastInfoSync < DateTime.UtcNow.AddHours(-1))
```
Taloth: Given the earlier comments I don't think th is code is needed. But if you want to keep it, then use _checkIfSeriesShouldBeRefreshed instead.

src/NzbDrone.Core.Test/MediaFiles/EpisodeImport/Specifications/EpisodeTitleSpecificationFixture.cs
markus101: 256k / 1.8M... more popular than I thought.

src/NzbDrone.Core/MediaFiles/EpisodeImport/Specifications/EpisodeTitleSpecification.cs
markus101: > If you're going to refresh, then you should do it synchronous and check the results again.

With CDH it'll check again in a minute, for API calls no, but you suggested special casing those anyways.

I think a single refresh is about as useless as no refreshes since there are caches and it'd come down to when that refresh happened.

src/NzbDrone.Core/MediaFiles/EpisodeImport/Specifications/EpisodeTitleSpecification.cs
markus101: Do we know the source of the import right now?

src/NzbDrone.Core/Tv/SeriesEditedService.cs
markus101: https://github.com/Sonarr/Sonarr/commit/25abeb8c9c267218f66e449fff9eab377615af42 - I specifically called out rescanning in the commit message, but that was before the special refresh logic, but when we were adding absolute episode numbers ourselves, but only for anime: https://github.com/Sonarr/Sonarr/blob/25abeb8c9c267218f66e449fff9eab377615af42/src/NzbDrone.Core/Tv/RefreshEpisodeService.cs#L156

The only really benefit I see with a refresh/rescan after editing now is for picking up air-by-date episodes when switch to daily.

Thirrian:
Would there be a way to override this? I don't use the {Episode Title} token for renaming files, and I'm not sure there is anything else that would really _need_ the title so as to block importing the episode?

markus101:
@Thirrian it should be possible for us to detect whether episode title is wanted in the filename and only delay the import if it's there.

src/NzbDrone.Core/Tv/ShouldRefreshSeries.cs
markus101: My thinking was, if I updated a series on TVDB and refreshed not knowing about the various caches nothing would update, then if a couple hours later Sonarr refreshed it automatically it'd skip the update it and not update automatically for about 15 hours.

Reverted it though.

markus101:
Apparently I addressed these comments months ago and forgot to push them, everything should be addressed now.

src/NzbDrone.Core/Organizer/FileNameBuilder.cs
```diff
@@ -280,6 +281,37 @@ public static string CleanFolderName(string name)
             return name.Trim(' ', '.');
         }
 
+        public bool RequiresEpisodeTitle(Series series, List<Episode> episodes)
```
Taloth: Encapsulate this in a cache, it's going to be called a lot and you don't want the same regex to be evaluated over and over again.
When config is changed, pull all formats, check which ones have Episode Title and store the resulting 3 booleans in the cache.

src/NzbDrone.Core/Organizer/FileNameBuilder.cs
```diff
@@ -280,6 +281,37 @@ public static string CleanFolderName(string name)
...
+        {
+            var namingConfig = _namingConfigService.GetConfig();
+            var pattern = namingConfig.StandardEpisodeFormat;
+
+            if (series.SeriesType == SeriesTypes.Daily)
+            {
+                pattern = namingConfig.DailyEpisodeFormat;
+            }
+
+            if (series.SeriesType == SeriesTypes.Anime && episodes.All(e => e.AbsoluteEpisodeNumber.HasValue))
```
Taloth: Why check if the episode has absolute numbers filled in? The renamer itself doesn't. Afaik it won't even import if there's no abs nr.

markus101: If the absolute episode numbers are not set the standard format is used instead.

https://github.com/Sonarr/Sonarr/blob/968d8159a6c8bc3f9d4168a8940a745c240d1b25/src/NzbDrone.Core/Organizer/FileNameBuilder.cs#L121-L124

Whether that's what we should do or not is questionable, with another PR discussing it here: https://github.com/Sonarr/Sonarr/pull/2283

src/NzbDrone.Core/Organizer/FileNameBuilder.cs
markus101: Added the cache, based on the format string, it could lead to more than 3 values stored, but only if the user is changing formats and that seems like an unlikely use case to worry about.

Taloth: That's fine. Not like the user is going to have hundreds of patterns and never reboot.

src/NzbDrone.Core/Organizer/FileNameBuilder.cs
Taloth: Understood, we'll revisit that later then.

src/NzbDrone.Core/MediaFiles/EpisodeImport/Specifications/EpisodeTitleSpecification.cs
```diff
@@ -0,0 +1,60 @@
...
+using NzbDrone.Core.Parser.Model;
+
+namespace NzbDrone.Core.MediaFiles.EpisodeImport.Specifications
+{
+    public class EpisodeTitleSpecification : IImportDecisionEngineSpecification
+    {
+        private readonly IBuildFileNames _buildFileNames;
+        private readonly Logger _logger;
+
+        public RejectionType Type => RejectionType.Temporary;
```
Taloth: You removed Type from the interface, but kept all implementations. Was that intentional?

Issue #263, opened 2015-01-20
Verify nzb before sending to download client


SonarrBot:
NZBGet occasionally reports an error parsing, after accepting the NZB it seems, which causes it to be tossed out (and likely blacklisted).

markus101:
(From #2221) Apparently some indexers set this in the body instead of using a header:

```xml
<error code="501" description="Download limit reached, please upgrade your account or wait 24 hours. See https://drunkenslug.com for details"/>
```

thezoggy:
any update on that validate-nzbs branch?

markus101:
There is a PR awaiting review: https://github.com/Sonarr/Sonarr/pull/2164

PR #2164, opened 2017-09-05
New: Validate NZBs before sending to download client

markus101:
#### Database Migration
NO

#### Description
Validates NZBs before sending them to the download client.

* #263


Safihre:
@markus101 Great to find this PR, just some feedback:
`should_throw_when_no_group_in_groups`, is not a nessecary reason to dismiss an NZB, no newsserver nowadays uses groups anymore, only article-ID's.
`should_throw_when_no_segment_in_segments` also not a reason to reject an NZB, SAB and NZBGet just ignore the file.

markus101:
Thanks @Safihre I dropped the following 4 checks:

- `should_throw_when_no_groups_in_file`
- `should_throw_when_no_group_in_groups`
- `should_throw_when_no_segments_in_file`
- `should_throw_when_no_segment_in_segments`

I assumed `groups` and `segments` don't need to be children of `file` if SAB and NZBGet don't care about their being groups or segments within.

Safihre:
Looks good! 
One can always think of more ways in which the NZB could be invalid, but these basic checks will effectively catch 99% of it :+1:

Safihre:
I couldn't make it up from the code, but what happens if the NZB is invalid? Is the whole indexer ignored in general or just the 1 specific NZB or release?

markus101:
That particular NZB is ignored and if available, another will be tried (could be the same release from another indexer or a different one altogether). I suppose we could also look at using invalid NZBs from an indexer as a sign that the indexer should be ignored for a period of time.

Safihre:
Yes, I would definitely think so. Maybe a minimal delay, but at least that it's not hit again in the same fetch/RSS session.

Safihre:
@markus101 lets get it in? In its current form it's already a great addition :)

markus101:
Thanks for the reminder @Safihre, merged.

PR #2172, opened 2017-09-13
new-feature: Add Plex Metadata Option

RandomNinjaAtk:
This adds Plex as a Metadata option for naming image files to the proper plex naming convention

#### Database Migration
NO

#### Description
This pull request adds Plex as a Metadata option for proper image naming conventions. Plex is a rather popular option that is similar to kodi/emby but the file naming in those metadata options do not fully match up with Plex, so this resolves that.

#### Todos
- [x] Tests
- [ ] Documentation

* None, new feature

### DIFF
--- a/src/NzbDrone.Core/Extras/Metadata/Plex/PlexMetadata.cs
+++ b/src/NzbDrone.Core/Extras/Metadata/Plex/PlexMetadata.cs
@@ -1,0 +1,214 @@
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text;
+using System.Text.RegularExpressions;
+using System.Xml;
+using System.Xml.Linq;
+using NLog;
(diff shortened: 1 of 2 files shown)


RandomNinjaAtk:
I'm really new to this and would greatly appreciate if someone with more expertise reviewed this helped getting it added. I've tested it and it was functioning for me as intended.

markus101:
Which provider are you using in Plex to read these files in after they are created?

RandomNinjaAtk:
You don't need a specific provider, Plex pulls the images in via its normal local scrapper.

For TV: https://support.plex.tv/hc/en-us/articles/200220717-Local-Media-Assets-TV-Shows

matmuilwijk:
From what I know, Plex doesn't support it's metadata to be exported. The builders of webtools have tried to do this, and all they could do was subtitles.

PR #2188, opened 2017-09-23
Add Jackett preset for Torznab Indexer

Wyall:
copied from radarr, added "categories", not sure about "apiPath" as it's not in any of the radarr presets.

Goal is just convenience for adding Jackett Indexers

### DIFF
--- a/src/NzbDrone.Core/Indexers/Torznab/Torznab.cs
+++ b/src/NzbDrone.Core/Indexers/Torznab/Torznab.cs
@@ -42,7 +42,7 @@ namespace NzbDrone.Core.Indexers.Torznab
                 yield return GetDefinition("HD4Free.xyz", GetSettings("http://hd4free.xyz"));
                 yield return GetDefinition("AnimeTosho Torrents", GetSettings("https://animetosho.org", apiPath: @"/feed/nabapi", categories: new int[0], animeCategories: new[] { 5070 }));
                 yield return GetDefinition("Nyaa Pantsu", GetSettings("https://nyaa.pantsu.cat", apiPath: @"/feed/torznab", categories: new int[0], animeCategories: new[] { 5070 }));
-                yield return GetDefinition("Jackett", GetSettings("http://localhost:9117/torznab/YOURINDEXER", categories: new int[0], animeCategories: new[] { 5070 }));
+                yield return GetDefinition("Jackett", GetSettings("http://localhost:9117/torznab/YOURINDEXER", categories: new int[] { 5000, 5020, 5030, 5040, 5045, 5050, 5080 }, animeCategories: new[0]));
             }
         }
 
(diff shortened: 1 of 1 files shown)


src/NzbDrone.Core/Indexers/Torznab/Torznab.cs
```diff
@@ -42,6 +42,7 @@ public override IEnumerable<ProviderDefinition> DefaultDefinitions
                 yield return GetDefinition("HD4Free.xyz", GetSettings("http://hd4free.xyz"));
                 yield return GetDefinition("AnimeTosho Torrents", GetSettings("https://animetosho.org", apiPath: @"/feed/nabapi", categories: new int[0], animeCategories: new[] { 5070 }));
                 yield return GetDefinition("Nyaa Pantsu", GetSettings("https://nyaa.pantsu.cat", apiPath: @"/feed/torznab", categories: new int[0], animeCategories: new[] { 5070 }));
+                yield return GetDefinition("Jackett", GetSettings("http://localhost:9117/torznab/YOURINDEXER", categories: new int[0], animeCategories: new[] { 5070 }));
```
markus101: Probably want categories, not animeCategories for most Torznab indexers.

Taloth:
Wouldn't it be easier to simply give jackett the sonarr api key and have it add, remove and configure the indexers as required?

galli-leo:
@Taloth This is already "tracked" as an issue on the Jackett Github: https://github.com/Jackett/Jackett/issues/1413 But it probably won't come to fruition, as our Indexer add API is not very generic (see my comment on the above issue)

src/NzbDrone.Core/Indexers/Torznab/Torznab.cs
```diff
@@ -42,6 +42,7 @@ public override IEnumerable<ProviderDefinition> DefaultDefinitions
                 yield return GetDefinition("HD4Free.xyz", GetSettings("http://hd4free.xyz"));
                 yield return GetDefinition("AnimeTosho Torrents", GetSettings("https://animetosho.org", apiPath: @"/feed/nabapi", categories: new int[0], animeCategories: new[] { 5070 }));
                 yield return GetDefinition("Nyaa Pantsu", GetSettings("https://nyaa.pantsu.cat", apiPath: @"/feed/torznab", categories: new int[0], animeCategories: new[] { 5070 }));
+                yield return GetDefinition("Jackett", GetSettings("http://localhost:9117/torznab/YOURINDEXER", categories: new int[] { 5000, 5020, 5030, 5040, 5045, 5050, 5080 }, animeCategories: new[0]));
```
markus101: Is the aggregate torznab URL similar to this URL? If so should we use that instead, so when someone inevitably doesn't change the URL they get something that works.

Panderine: What would this do? Any documentation regarding this setting?

markus101: What do you mean? What setting?

Panderine: When adding a new torznab indexer, I get presented 2 options "custom" and "jackett". How do I use the Jackett option?. With custom I copy the link from jackett "torznab" button without issues, but what does the "jackett" option do? "http://localhost:9117/torznab/YOURINDEXER" What is "YOURINDEXER"?

markus101: `YOURINDEXER` would need to be replaced with the matching portion from the actual Jackett URL.

PR #2209, opened 2017-10-05
Made date format options more descriptive

franga2000:
#### Database Migration
NO

#### Description
Since the day and month numbers were both lower than 12 you couldn't tell which was which.  
I changed the day number to 25 because it very obviously couldn't be a month number and has the added bonus of actually being a tuesday (that was actually a coincidence :D).


rsunde:
I like this.

PR #2214, opened 2017-10-06
Better import error messages

rsunde:
#### Database Migration
NO

#### Description
Adds better import error messages, just saying Import Failed isnt enough.

#### Todos
- [ ] Tests
- [ ] Documentation

*


src/NzbDrone.Core/MediaFiles/EpisodeImport/ImportApprovedEpisodes.cs
```diff
@@ -122,9 +122,34 @@ public List<ImportResult> Import(List<ImportDecision> decisions, bool newDownloa
 
                     _eventAggregator.PublishEvent(new EpisodeImportedEvent(localEpisode, episodeFile, oldFiles, newDownload, downloadClientItem));
                 }
+                catch (DirectoryNotFoundException e)
```
markus101: It'd be better to throw a specific exception if the root folder can't be found instead of `DirectoryNotFoundException` and then check the exception message. Perhaps a `RootFolderNotFoundException` that inherits `DirectoryNotFoundException`.

src/NzbDrone.Core/MediaFiles/EpisodeImport/ImportApprovedEpisodes.cs
```diff
@@ -122,9 +122,34 @@ public List<ImportResult> Import(List<ImportDecision> decisions, bool newDownloa
...
+                    }
+
+                    importResults.Add(new ImportResult(importDecision, errorMessage));
+                }
+                catch (IOException e)
+                {
+                    _logger.Warn(e, "Couldn't import episode " + localEpisode);
+
+                    var errorMessage = "Failed to import episode";
+                    if (e.Message.Contains("Destination already exists."))
```
markus101: Similar to the above, can we throw a better exception here? Perhaps pre-check if the destination exists and throw our own exception. Parsing a message we're not controlling won't work if the language is not English.

src/NzbDrone.Common/Exceptions/DestinationAlreadyExistsIOException.cs
```diff
@@ -0,0 +1,29 @@
+using System;
+using System.IO;
+using System.Runtime.Serialization;
+
+namespace NzbDrone.Common
+{
+    public class DestinationAlreadyExistsIOException : IOException
```
markus101: One small nitpick, could you rename this to `DestinationAlreadyExistsException`, otherwise I think it's good.

markus101:
Made a few small tweaks and merged via https://github.com/Sonarr/Sonarr/commit/73cb789f59a63816228904ad7fc53b137e6865df
Thanks!

PR #2216, opened 2017-10-07
Add new codec HEVC with ID=5 for HDBits

Wyall:
According to LSL1337 on Radarr Discord there's a new codec option HEVC with ID=5

### DIFF
--- a/src/NzbDrone.Core/Indexers/HDBits/HDBitsSettings.cs
+++ b/src/NzbDrone.Core/Indexers/HDBits/HDBitsSettings.cs
@@ -59,7 +59,7 @@ namespace NzbDrone.Core.Indexers.HDBits
         Mpeg2 = 2,
         Vc1 = 3,
         Xvid = 4.
-        HEVC = 5
+        Hevc = 5
     }
 
     public enum HdBitsMedium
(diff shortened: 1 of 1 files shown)


src/NzbDrone.Core/Indexers/HDBits/HDBitsSettings.cs
```diff
@@ -58,7 +58,8 @@ public enum HdBitsCodec
         H264 = 1,
         Mpeg2 = 2,
         Vc1 = 3,
-        Xvid = 4
+        Xvid = 4.
+        HEVC = 5
```
markus101: Should be `Hevc` for code styling consistency.

src/NzbDrone.Core/Indexers/HDBits/HDBitsSettings.cs
```diff
@@ -58,7 +58,8 @@ public enum HdBitsCodec
         H264 = 1,
         Mpeg2 = 2,
         Vc1 = 3,
-        Xvid = 4
+        Xvid = 4.
```
thezoggy: shouldnt that be a trailing coma not a period. also squash and rebase your commits to keep history clean

markus101: Yes, should be a `,`.

During a review squashing and rebasing should be left to the very end so individual commits can be compared. (We can squash and merge right from GitHub now).

Wyall: Did I do that wrong (squashing and rebasing)? I'm sorry, github can be complicated if not used frequently

PR #2223, opened 2017-10-11
Add default calendar name for iCal

jamesmacwhite:
#### Database Migration
NO

#### Description

Provides a default name for the Sonarr.ics when added as its own calendar to an application that supports iCal. I thought this would be better than using the long URL path which is currently used because nothing else is set.

It appears both the `NAME` and `X-WR-CALNAME` are used for this purpose, they are however both non-standard and not in the RFC, which is why they have to be added as properties.

### DIFF
--- a/src/NzbDrone.Api/Calendar/CalendarFeedModule.cs
+++ b/src/NzbDrone.Api/Calendar/CalendarFeedModule.cs
@@ -4,6 +4,7 @@ using System.Collections.Generic;
 using System.Linq;
 using Ical.Net;
 using Ical.Net.DataTypes;
+using Ical.Net.General;
 using Ical.Net.Interfaces.Serialization;
 using Ical.Net.Serialization;
 using Ical.Net.Serialization.iCalendar.Factory;
@@ -91,6 +92,10 @@ namespace NzbDrone.Api.Calendar
             {
                 ProductId = "-//sonarr.tv//Sonarr//EN"
             };
+
+            var calendarName = "Sonarr TV Schedule";
+            calendar.AddProperty(new CalendarProperty("NAME", calendarName));
+            calendar.AddProperty(new CalendarProperty("X-WR-CALNAME", calendarName));
 
             foreach (var episode in episodes.OrderBy(v => v.AirDateUtc.Value))
             {


src/NzbDrone.Api/Calendar/CalendarFeedModule.cs
```diff
@@ -92,6 +93,10 @@ private Response GetCalendarFeed()
                 ProductId = "-//sonarr.tv//Sonarr//EN"
             };
 
+            var calendarName = "Sonarr TV Schedule Calendar";
+            calendar.AddProperty(new CalendarProperty("NAME", calendarName));
```
markus101: Is this used by a specific client? Looks like `X-WR-CALNAME` is fairly well used, but didn't find anything on name.

jamesmacwhite: It's a good question. I wasn't too sure either. I looked at a couple of iCal examples and a few had both.

Let me see if I can provide some samples and references.

jamesmacwhite: So X-WR-CALNAME does appear to be used widely and is classed as an extension, but the name property does appear to exist in some form.

https://tools.ietf.org/html/rfc7986#section-5.1

I can drop it, if you feel its not worth including, but both appear to be used in some cases, here's an example:

* https://www.formula1.com/sp/static/f1/2017/calendar/ical.ics

markus101: Thanks, leaving it is fine with me.

src/NzbDrone.Api/Calendar/CalendarFeedModule.cs
```diff
@@ -92,6 +93,10 @@ private Response GetCalendarFeed()
                 ProductId = "-//sonarr.tv//Sonarr//EN"
             };
 
+            var calendarName = "Sonarr TV Schedule Calendar";
```
markus101: Can you update this to `Sonarr TV Schedule`? Calendar is implied and best left off I think.

jamesmacwhite: No problem. Will update and ping when I've updated the PR.

jamesmacwhite: @markus101 Updated the default name. Should be good to merge.

markus101:
Thanks!

PR #2240, opened 2017-10-21
Change default port of uTorrent and qBittorrent

jamesmacwhite:
#### Database Migration
NO

#### Description

Changes the default port set when setting up uTorrent or qBittorrent as a download client.

* uTorrent sets `8080` as the alternative port WebUI port by default.
* qBittorrent uses `8080` as the default port.


markus101:
Thanks!

Issue #967, opened 2015-11-30
Symlink/Hardlink/Copy/Move


CochinoChingon:
Currently Only copy and hardlink are supported. I need symlink before switching over from sickrage. I'm pretty sure I'm not the only one holding out for symlink. the time I used Sonarr was great for adding shows quite simply, but over time my need for symlink trumps ease. 

I'm not the only one who needs this feature and I'm sure someone else could more eloquently explain a situations where it is needed.

kayone:
I'm curious what's the scenario that symlinks would be useful?

also, I assume the original file would stay in download directory and symlink would be in the series folder? or the other way around?

Megagator:
I'm running into this exact issue. I need torrents to continue to seed from the host machine, yet my collection lives on external drives/NAS. Previously, SickRage would copy the completed download to the destination drive and symlink that back to a seeding folder where it could continue to seed, keeping the actual bits off of my computer's primary drive. 

Hard links cannot be made between two different file systems, so Sonarr falls back to coping the data instead.

```
Mono.Unix.UnixIOException: Cross-device link [EXDEV].
  at Mono.Unix.UnixMarshal.ThrowExceptionForLastError () [0x00000] in /private/tmp/source-mono-mac-4.2.0-branch/bockbuild-mono-4.2.0-branch/profiles/mono-mac-xamarin/build-root/mono-4.2.1/mcs/class/Mono.Posix/Mono.Unix/UnixMarshal.cs:460 
  at Mono.Unix.UnixMarshal.ThrowExceptionForLastErrorIf (Int32 retval) [0x00007] in /private/tmp/source-mono-mac-4.2.0-branch/bockbuild-mono-4.2.0-branch/profiles/mono-mac-xamarin/build-root/mono-4.2.1/mcs/class/Mono.Posix/Mono.Unix/UnixMarshal.cs:473 
  at Mono.Unix.UnixFileSystemInfo.CreateLink (System.String path) [0x00000] in /private/tmp/source-mono-mac-4.2.0-branch/bockbuild-mono-4.2.0-branch/profiles/mono-mac-xamarin/build-root/mono-4.2.1/mcs/class/Mono.Posix/Mono.Unix/UnixFileSystemInfo.cs:277 
[...]

PR #2250, opened 2017-10-27
Mono: Try copying files with cp --reflink first.

bojanrajkovic:
#### Database Migration
No.

#### Description
This should save a bunch of space for people on compatible filesystems
(like Btrfs). `cp` with `--reflink=auto` will also just do a normal
copy of the file if reflinks aren't possible, so the fallback to the
base is more of a "just in case" than anything. A short (10s) timeout
is used if the source or destination mounts are network mounts, to avoid
cp hanging with network mounts where the server has gone down.

* Partial fix for #967 -- addresses https://github.com/Sonarr/Sonarr/issues/967#issuecomment-273346468, but not the overall issue.

### DIFF
--- a/src/NzbDrone.Common/Disk/DiskProviderBase.cs
+++ b/src/NzbDrone.Common/Disk/DiskProviderBase.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
@@ -190,7 +190,7 @@ namespace NzbDrone.Common.Disk
             File.Delete(path);
         }
 
-        public void CopyFile(string source, string destination, bool overwrite = false)
+        public virtual void CopyFile(string source, string destination, bool overwrite = false)
         {
             Ensure.That(source, () => source).IsValidPath();
             Ensure.That(destination, () => destination).IsValidPath();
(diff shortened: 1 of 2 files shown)


markus101:
I don't think `--reflink=auto` is supported in macOS as last I checked long options like that were not supported, which means it'd always throw an exception (or fail in another way).

I think @Taloth and I are in agreement that calling `cp` is not something we want to do, so I don't think this is a change we'll pull in, but I'll leave this for a bit to see if @Taloth has any further thoughts.

Taloth:
The biggest issue is isolation. At the very least one should check whether the source _and_ dest filesystems are both btrfs.
I still don't like calling 'cp', I _really_ don't, but I might allow it if it's properly isolated and the only viable solution. Personally I simply have a single zfs dataset for both the series and download directories, which works fine, that in combination with the potential maintenance issues makes me lean towards not accepting a 'cp' solution.
It would be better if the appropriate ioctl (btrfs_ioc_clone) is called (directly, or from mono). Right now we use the mono-specific UnixFileSystemInfo class to create a hardlink. Mono might be interested in a CreateRefLink method on that class, assuming it can be reliably implemented. Of course that's an order of magnitude more complicated than your PR.

Finally, if you want a truly generic solution, then you might want to consider writing (or using an existing one) FUSE filesystem wrapper. Such a wrapper could properly abstract multiple filesystems and would potentially be able to handle and manage symlinks as well, thus helping with the non-btrfs usecases.
I believe there are existing wrappers but I don't remember the name of the one I'm thinking of.
The advantage of such a wrapper is that any app would benefit from it and that the wrapper could solve the biggest issue with symlinks: track the links to prevent dead symlinks if files get renamed.

PR #2258, opened 2017-10-31
fixed moment.js deprecated zone

Fish2:
#### Database Migration
NO

#### Description
fixed moment.js deprecated zone and add

don't pull in yet got issues 

#### Todos
- [ ] Tests
- [ ] Documentation

* https://github.com/Radarr/Radarr/issues/2231

### DIFF
--- a/src/UI/Cells/RelativeDateCell.js
+++ b/src/UI/Cells/RelativeDateCell.js
@@ -12,7 +12,7 @@ module.exports = NzbDroneCell.extend({
 
         if (dateStr) {
             var date = moment(dateStr);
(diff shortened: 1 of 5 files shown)


markus101:
Why the two unrelated commits?
As mentioned previously images have already been compressed and aren't used by the UI https://github.com/Sonarr/Sonarr/pull/1959

Fish2:
I can remove them but why not Compression them fully its lossless so makes no difference

markus101:
They're unrelated to this PR, so shouldn't be included here. As for why not, there is little value in compressing the same thing multiple times, another pass of pingo would also remove some size, but not something we think is worthwhile.

Fish2:
Ok i have removed the other 2 commits. Yer true

PR #2268, opened 2017-11-04
Fixes/mobile ui

beschoenen:
#### Database Migration
NO

#### Description
This PR fixes a few UI quirks on mobile or smaller devices.


src/UI/Navbar/NavbarLayoutTemplate.hbs
```diff
@@ -34,10 +34,12 @@
     </div><!--/.container-fluid -->
 
     <div class="col-md-12 search">
-        <div class="col-md-6 col-md-offset-3">
-            <div class="input-group">
-                <span class="input-group-addon"><i class="fa fa-search"></i></span>
-                <input type="text" class="col-md-6 form-control x-series-search" placeholder="Search the series in your library">
+        <div class="row">
```
markus101: > Rows must be placed within a .container (fixed-width) or .container-fluid (full-width) for proper alignment and padding.

I believe that's why this fixes the width, because the padding is lost.

src/UI/Series/Index/FooterViewTemplate.hbs
```diff
@@ -1,10 +1,10 @@
 <div class="row">
     <div class="series-legend legend col-xs-6 col-sm-4">
         <ul class='legend-labels'>
-            <li><span class="progress-bar"></span>Continuing (All episodes downloaded)</li>
-            <li><span class="progress-bar-success"></span>Ended (All episodes downloaded)</li>
-            <li><span class="progress-bar-danger"></span>Missing Episodes (Series monitored)</li>
-            <li><span class="progress-bar-warning"></span>Missing Episodes (Series not monitored)</li>
+            <li><span class="progress-bar"></span>Continuing <i class="break-sm">(All episodes downloaded)</i></li>
```
markus101: Use a `span`, `i` should not be used outside of icons.

src/UI/Series/series.less
```diff
@@ -469,3 +469,14 @@
     list-style-type : none;
   }
 }
+
+//Series footer
+.break-sm {
+  font-style: normal;
+}
+
+@media only screen and (max-width : 600px) {
```
markus101: `max-width:`
Use `@screen-sm-max` or `@screen-xs-max` (for closer to 600px) instead of a custom width.

src/UI/Series/series.less
```diff
@@ -469,3 +469,14 @@
...
+  .break-sm {
```
markus101: I think you want `break-xs` instead of `sm` (small is 991px max)

src/UI/Settings/SettingsLayoutTemplate.hbs
```diff
@@ -11,7 +11,7 @@
 </ul>
 
 <div class="row settings-controls">
-    <div class="col-sm-4 col-sm-offset-7 col-md-3 col-md-offset-8">
+    <div class="col-xs-9 col-sm-5 col-sm-offset-6 col-md-4 col-md-offset-7 col-lg-3 col-lg-offset-8">
```
markus101: This causes the Nexus 5x screen (412x732) to overflow horizontally and require scrolling.

beschoenen: Can you sent me a screenshot of this, cause I can't reproduce it?
screen shot 2017-11-06 at 19 38 02

src/UI/Navbar/NavbarLayoutTemplate.hbs
beschoenen: So, this change is good, right?

src/UI/Settings/SettingsLayoutTemplate.hbs
beschoenen: I have to admit that this does not work a 100% on even smaller screens like the iPhone 5 or older android Devices. The text `Advanced Settings` gets put underneath the slider.

markus101: Leave this change out then, the v3 UI will change it anyways.

src/UI/Navbar/NavbarLayoutTemplate.hbs
markus101: Probably better to be more explicit, instead of relying on the side effect of `row`.

PR #2269, opened 2017-11-05
Feature/transmission seeding

beschoenen:
#### Database Migration
NO

#### Description
This PR adds the ability to set a custom seed ratio and seed idle time for transmission.

#### Todos
- [ ] Tests ?

### DIFF
--- a/src/NzbDrone.Core/Download/Clients/Transmission/TransmissionBase.cs
+++ b/src/NzbDrone.Core/Download/Clients/Transmission/TransmissionBase.cs
@@ -231,6 +231,23 @@ namespace NzbDrone.Core.Download.Clients.Transmission
             {
                 _proxy.MoveTorrentToTopInQueue(hash, Settings);
             }
+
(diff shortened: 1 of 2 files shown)


markus101:
Seeding ratios need to be configured per indexer, not per download client and as long as Sonarr can get the seeding time/ratio from the download client nothing needs to be set on each torrent. Sonarr would make a decision when processing to determine if it already met the seeding goals and then remove the item from the download client.

beschoenen:
Ok, I'm almost done with your way of implementing this feature, but could you tell me how I would get the indexer used to download a `TrackedDownload`. I'm trying to implement it [here](https://github.com/Sonarr/Sonarr/blob/develop/src/NzbDrone.Core/Download/TrackedDownloads/DownloadMonitoringService.cs#L116)

markus101:
You'd have to get the indexer from the `grabbed` history event using the `DownloadId` (torrent hash in this case).

beschoenen:
Continued in #2272

Issue #657, opened 2015-07-02
rTorrent Ratio Groups


timelf123:
![Screenshot](http://puu.sh/iKhDv/3f7f61c0fe.png)

timelf123:
Just had a brainfart and realized this is for rutorrent. Any plans to support the rutorrent webui?

tYGjQCsvVI:
I'm pretty sure this functionality exists in rTorrent:-

`'group.insert'
'group.insert_persistent_view'
'group.seeding.ratio.command'
'group.seeding.ratio.disable'
'group.seeding.ratio.enable'
'group.seeding.ratio.max'
'group.seeding.ratio.max.set'
'group.seeding.ratio.min'
'group.seeding.ratio.min.set'
'group.seeding.ratio.upload'
'group.seeding.ratio.upload.set'
'group.seeding.view'
'group.seeding.view.set'`

[Source](https://code.google.com/p/gi-torrent/wiki/rTorrent_XMLRPC_reference), although it may been removed in the latest version (Not sure why I believe this, but something in my head is telling me that that has occurred).

xelra:
This would probably become an obsolete feature with per-indexer seedratio and -time support.

sataris:
Given the recent release of the per-indexer seedratio  and -time support. 

Is this still on the agenda?

ZerithZA:
Hi, is there any plans to make this happen? I'm a bit more interested in the removing of torrents once done seeding though.

xelra:
@ZerithZA We have per-indexer seedration and -time support in Sonarr now. There's no need anymore to mock something up yourself in rTorrent.

ZerithZA:
The reason why I asked this is because I have a hosted seedbox from where I sync back the content locally to my synology nas.  During this time it continues being seeded on the seedbox, I set it so that hard links are made at the time of import so when the torrent gets removed from seeding / deleted from the disk, it also remove the link in the sync directory. 

I don't see any nice clean way of currently doing this. I'm somewhat excited for #282 as I'm hoping that it will take away some of the requirements from syncing and better handling of imports in season packs etc.

xelra:
@ZerithZA If you do custom file transfers, you can use the API endpoints to make Sonarr understand what you're doing. That way you can fully integrate any custom processing you're doing on your own into Sonarr's Completed Download Handling.

[...]

PR #2272, opened 2017-11-06
Indexer seed ratio

beschoenen:
#### Database Migration
NO

#### Description
Allows users to set a target seed ratio per indexer.

#### Todos
- [ ] Tests
- [ ] Documentation


LinusT:
Looks great, a much needed feature. And a way around the Remove torrent issues that are plaquing almost all clients. 

Some indexers has a minimum seed time limit as well so a suggestion is to add both a minimum ratio and minimum seed time limit.

src/NzbDrone.Core/Download/TrackedDownloads/DownloadMonitoringService.cs
```diff
@@ -113,12 +121,48 @@ private List<TrackedDownload> ProcessClientDownloads(IDownloadClient downloadCli
...
+
+            foreach (var trackedDownload in removableDownloads)
             {
                 _eventAggregator.PublishEvent(new DownloadCompletedEvent(trackedDownload));
             }
         }
 
+        private bool TorrentHasFinishedSeeding(TrackedDownload download)
+        {
+            var historyItem = _historyService.Find(download.DownloadItem.DownloadId, HistoryEventType.Grabbed).Last();
```
beschoenen: @markus101 I'm not sure about this line, is it okay like this?

markus101: Sorry! Missed this comment. Not sure what else we could do here instead, the last grabbed with that ID should be the one we're concerned about. Only possible issue would be torrents with the same hash from multiple trackers, but not sure we can do anything about that.

src/NzbDrone.Core/Download/TrackedDownloads/DownloadMonitoringService.cs
```diff
@@ -113,12 +121,48 @@ private List<TrackedDownload> ProcessClientDownloads(IDownloadClient downloadCli
...
+            if (historyItem == null)
+            {
+                return download.DownloadItem.CanBeRemoved && download.State == TrackedDownloadStage.Imported;
+            }
+
+            var indexer = _indexerFactory.All().First(i => i.Name == historyItem.Data["indexer"]);
```
markus101: I think we should store the indexer ID, so a simple change in name doesn't break this.

src/NzbDrone.Core/Indexers/BitMeTv/BitMeTvSettings.cs
```diff
@@ -47,6 +52,9 @@ public BitMeTvSettings()
         [FieldDefinition(4, Type = FieldType.Textbox, Label = "Minimum Seeders", HelpText = "Minimum number of seeders required.", Advanced = true)]
         public int MinimumSeeders { get; set; }
 
+        [FieldDefinition(5, Type = FieldType.Textbox, Label = "Seed Ratio", HelpText = "The ratio a torrent should reach before stopping, empty is download client's default")]
+        public string SeedRatio { get; set; }
```
markus101: This should be stored as a double instead of a string we convert to a double when needed.

src/NzbDrone.Core/Download/Clients/DownloadStation/TorrentDownloadStation.cs
```diff
@@ -276,6 +277,19 @@ protected long GetRemainingSize(DownloadStationTask torrent)
...
+        {
+            var couldConvertDownloaded = long.TryParse(torrent.Additional.Transfer["size_downloaded"], out var downloaded);
+            var couldConvertUploaded = long.TryParse(torrent.Additional.Transfer["size_uploaded"], out var uploaded);
+
+            if (!couldConvertDownloaded || !couldConvertUploaded)
+            {
+                return new Nullable<double>();
+            }
+
+            return (double) uploaded / downloaded;
```
Taloth: Potential div by zero.

src/NzbDrone.Core/Download/Clients/Hadouken/Hadouken.cs
```diff
@@ -62,7 +62,8 @@ public override IEnumerable<DownloadClientItem> GetItems()
                     RemainingSize = torrent.TotalSize - torrent.DownloadedBytes,
                     RemainingTime = eta,
                     Title = torrent.Name,
-                    TotalSize = torrent.TotalSize
+                    TotalSize = torrent.TotalSize,
+                    SeedRatio = (double) torrent.UploadedBytes / torrent.DownloadedBytes
```
Taloth: Potential div by zero.

src/NzbDrone.Core/Download/Clients/Transmission/TransmissionBase.cs
```diff
@@ -66,6 +66,8 @@ public override IEnumerable<DownloadClientItem> GetItems()
                 item.OutputPath = GetOutputPath(outputPath, torrent);
                 item.TotalSize = torrent.TotalSize;
                 item.RemainingSize = torrent.LeftUntilDone;
+                item.SeedRatio = (double) torrent.UploadedEver / torrent.DownloadedEver;
```
Taloth: Potential div by zero.

Taloth:
@markus101 According to the readme we're still on VS2015, so no C#7 language features allowed.
That said, I'm ok with switching, but maybe not till v3, we're going to up from net4.0 to net4.7 right? But we should set an explicit language version on the csproj so people get warnings/errors if they use newer features.
I didn't see any problematic c#7 features used. I don't want to see convoluted usage of the ?. operator for example. There are some really nice c#7 constructs.

But yes, till we officially up to vs2017, we should refrain from any c#7 features till we can properly switch.

Taloth:
My bigger concern is that a SeedRatio on the indexer isn't enough. I think I talked to you about this on hangouts.

The original purpose of [TorrentSeedConfiguration](https://github.com/Sonarr/Sonarr/search?utf8=%E2%9C%93&q=TorrentSeedConfiguration) was to provide a way to configure the torrent client on a per-torrent basis. We never tied up the supply end of that system, iirc there's an Github issue for that, but the idea was that either the indexer or the user could specify the minimum criteria.
The configuration would be determined either by the torznab attributes for it (which are tracker supplied), or some global settings. But ultimately it would _only_ be set when the torrent got added, as a default.
For Deluge, Sonarr only does a SeedRatio check because Deluge doesn't have a 'Finished Seeding' status instead simply 'Pauses' it and Sonarr checks _against_ the seedratio set in Deluge on the torrent level, not some value of it's own. it's a feature limited to the Deluge class.

Architecturally, Sonarr shouldn't check against it's own parameters. This is why the PR ends up trying to determine the indexer based on the downloadclient item in order to retrieve the SeedRatio parameter to check against it. IMO this is wrong.
Why not fix the download client to have a seed ratio property that can be set? Deluge already does, utorrent does. rtorrent has ratio_groups (which we don't support yet). I don't know if Hadouken does, but they're open to suggestion afaik.

The reason why I'm saying this is because then information keeps flowing from IndexerProxy->DecisionMaker->DownloadService->DownloadClientProxy, instead of the DownloadService calling back to the Indexer services to get some settings.
Implementing a check in Sonarr might seem inviting, because then it's 'fixed globally', but imho that's a poor excuse.

I'm ok with some features in Sonarr to set default seedratio/seedtime properties in the download client on a per torrent basis. But I just don't like Sonarr to be a SeedManager.

beschoenen:
@markus101 I fixed all your comments, I hope I changed the SeedRatio to double like you intended it be changed. 

Sorry that it took me so long.

beschoenen:
Any update for this?

Taloth:
My previous reply was mostly intended to begin a discussion, since I voiced several concerns. That wasn't just to markus, but also to you.

The general idea is that the settings in Sonarr should be used to set the download clients own ratio limits (most clients support it in one way or another). That avoids the DownloadService doing a call back to the history _and_ indexer subsystem to get the applicable seedratio.     
For example, you can see Sonarr using the seed ratio limit specified on the Deluge api to determine whether a torrent has finished seeding vs simply paused.

beschoenen:
Ok, @Taloth, I started with adding limits inside the clients, but looking at some of the documentation I don't think we can implement it like this for all clients.

- [x] Deluge
- [ ] Downloadstation
  Not supported? [API Docs](https://global.download.synology.com/download/Document/DeveloperGuide/Synology_Download_Station_Web_API.pdf)
- [ ] Hadouken
  Not supported? [RPC Docs](http://hadouken.readthedocs.io/en/latest/for-developers/jsonrpc.html)
- [ ] QBittorrent
- [ ] rTorrent
  Implement groups? [RPC Docs](https://github.com/mdevaev/emonoda/wiki/rTorrent-XMLRPC-Reference)
- [x] Transmission
  Implemented (Not tested)
- [ ] uTorrent
  No property to get the seed limit from?
- [x] Vuze
  Based on Transmission

beschoenen:
An option for hadouken would be to add tags to the torrent with the seed ratio in it that can be parsed back to a double.

`"sonarr-sr-2.0"` (?)

http://hadouken.readthedocs.io/en/latest/for-developers/jsonrpc-methods/webui-addTorrent.html

Taloth:
The Hadouken dev actually was involved when we first implemented support on it. So I would've suggested collaborating again to get support added. But hadouken hasn't been updated for years now... domain gone etc. Meaning it's basically dead. (@markus101 we should consider dropping support at one point, maybe add it to the v3 deprecation list?)

Supporting ratiogroups for rtorrent has been a open request for a while (https://github.com/Sonarr/Sonarr/issues/657). But I'm not sure how to get sonarr to pick or create the right group.

I think the focus should be to support the clients we can now, and progressively add others.

A few comments:
- Don't use `IIndexerFactory` to do the GetSeedConfiguration logic, delegate that to a `ISeedConfigurationProvider`, that way the source of the info and the determination logic (single ep vs season pack, for example) is isolated.   
- I'm also thinking that that logic should be in `DownloadService` or one of it's callers. DownloadService could pass the SeedConfiguration to the individual download client, instead of having the downloadclient call back to get it. Maybe even as a member property of RemoteEpisode.

Got a more things, but lets get the generic architecture in place first.

@markus101  How are we going to add the fields to all the settings? I'm thinking, maybe we should add nested settings, so we can add fields like MinSeeders and SeedConfiguration without having to add each individual field (and related metadata) separately. Like:
```
[FieldDefinition(4)]
public SeedConfigurationSettings SeedConfiguration { get; set; }
```
Otherwise we'll have to make it a first class citizen like delay profile.

markus101:
I added it to the list, I'm sure someone will mind...maybe.

The settings is an interesting problem, one thing that we could do is define the fields in a common location and set a higher order to them, say 100+. In v3 order is only used to order the items, unlike v2 where the UI relies on them being unique for different things. I'd be nice to not have to specify an order at all, but I don't think there is a reliable enough way to achieve that.

LinusT:
If we look a bit more holistically on this... many users of Sonarr will also be users of Radarr, and perhaps also Lidarr etc. Thus I would argue for the approach of configure the seeding rules once (which might be ration and/or time and perhaps other factors).

Would it be possible to instead of managing in Sonarr simply change the label of a download to reflect that it has been imported? Thus allowing the torrent client and/or a regular shell script automating the torrent client to take care of the stop seeding and/or delete rules?

(also registered as issue #2469)

src/NzbDrone.Core/Indexers/BitMeTv/BitMeTvSettings.cs
```diff
@@ -47,6 +54,13 @@ public BitMeTvSettings()
         [FieldDefinition(4, Type = FieldType.Textbox, Label = "Minimum Seeders", HelpText = "Minimum number of seeders required.", Advanced = true)]
         public int MinimumSeeders { get; set; }
 
+        [FieldDefinition(5, Type = FieldType.Textbox, Label = "Seed Ratio", HelpText = "The ratio a torrent should reach before stopping, empty is download client's default")]
+        public string UiSeedRatio
```
Taloth: I didn't comment on this earlier because there were bigger fish to fry. But why this workaround in every settings class? Shouldn't it simply be a double? The value conversion should be handled in the UI or the SchemaBuilder.

Taloth:
_Hint: You can actually check the build yourself via the guest link in TC._
Reason it's failing is because the build server doesn't support the latest language features yet. I pointed one out in a comment above.

As you may have read, I think we should introduce sub-settings so we can add various seed criteria. One limit is not going to be enough, especially since BTN has season vs episode criteria. And it needs a switch to turn on (So it only sets the seed config if the user explicitly specified so and otherwise should maintain the client defaults).
I'll have to take a look at how to do the sub-settings correctly.

Taloth:
@markus101 I now realize the reason why the c# 7 features are failing is because we switched to the older build agents since your server died. Maybe an idea to update the build tools on those agents.

beschoenen:
@Taloth I can remove the old `SeedRatio` logic right?

Taloth:
I don't see any of that remaining in the PR. Or you're talking about a different SeedRatio logic.

I think the old logic you had should be kept in a separate class so you can use it for clients that don't support it natively. But we can add that back in once we've implemented it for the clients with native support.

beschoenen:
@Taloth Ok, I'll leave it in then.

Could you maybe also have a look at the UI part you mentioned, I can't seem to figure out why, but Sonarr won't save the `SeedRatio` value if I use a `double` type for it, like this:

```cs
[FieldDefinition(5, Type = FieldType.Textbox, Label = "Seed Ratio", HelpText = "The ratio a torrent should reach before stopping, empty is download client's default")]
public double SeedRatio { get; set; }
```

Taloth:
That's because https://github.com/Sonarr/Sonarr/blob/ef03e9e9a7e8170d8166240d827e942362831c72/src/NzbDrone.Api/ClientSchema/SchemaBuilder.cs#L74 doesn't handle it.

beschoenen:
Aha! Thanks 😄

beschoenen:
What should be done before this can be merged?
The logic for clients that don't support setting a seed limit should be cleaned up.

And maybe we can put the subsettings aside for later.

beschoenen:
@Taloth I think that resolves all your points.

Taloth:
A bit hectic at work today, so I can't combine the two PRs yet, i'll try do that tomorrow.

In the meantime, can you do a full squash rebase on top of latest develop?

Taloth:
Ok, I squashed and rebased it, fixed a couple of bugs in the nested logic and added some other stuff.

https://github.com/Sonarr/Sonarr/tree/pr/n2272_beschoenen

I think all that remains is to actually test it with the supported clients.

beschoenen:
Ah, sorry, I didn't have to do it yet.

I know transmission works, but I'll do a testing round tomorrow.

beschoenen:
Also, should we mention the supported clients somewhere? In the help text maybe.

beschoenen:
@Taloth Just tested out Transmission, your new seed time settings don't seem to work.
I entered 1 in the field and it's saved as 1 second, which is then parsed back to 0 minutes.
```cs
Convert.ToInt32(seedConfiguration.SeedTime.Value.TotalMinutes)
```

But transmission also doesn't pick up the 0, it just sticks to my default of 30 minutes.

Taloth:
> Also, should we mention the supported clients somewhere? In the help text maybe.

Good idea, but do it via HelpLink and the wiki page. Maybe add a table to https://github.com/Sonarr/Sonarr/wiki/Supported-DownloadClients so we can say which clients support Removal, override seed time/seed ratio. etc

> Should be TimeSpan.FromMinutes obviously.

Yup it was seconds first, and I was kinda excited by adding the unit to the UI that I forgot to change to FromMinutes :smile: nice catch.

>I do think that we shouldn't restrict the values to greater than zero 1, 0 should be fine too, at least for seed ratio.

Did that intentionally. If people don't want to seed at all they're free to set their torrent client global value to 0 if they can. Anyway, then can set to 0.01 or something if they're so inclined. Although some clients might even reject values that low.
I really didn't want to make it too easy for ppl to never seed.

beschoenen:
@Taloth #2545 

>Did that intentionally. If people don't want to seed at all they're free to set their torrent client global value to 0 if they can. Anyway, then can set to 0.01 or something if they're so inclined. Although some clients might even reject values that low.
That would actually go against my use case, I want to seed everything, except torrent that Sonarr picks up from public trackers.

Taloth:
@beschoenen I merged in the work on develop.
It still needs a link to the appropriate wiki page though, but I don't know if it deserves it's own page or should be a header in the Indexer Integration page. The wiki, admittedly, is a bit of a mess.

markus101:
Closing this out, let's open an issue for any remaining work.

Issue #1866, opened 2017-04-22
Support for download client priority


RaphaelJenni:
It would be nice if you can set a main download client and a fallback one (prioritize clients).

In my case, I use my normal work machine during the day. It has way more power than my nas, where sonarr is running on. So I want, that during the day, sonarr downloads the series via sabnzbd on my PC and switches to the sabnzbd on my NAS during the night (when my PC isn't running).

markus101:
Perhaps something that could be useful in combination with disabling offline download clients, but we'd need to add the ability to prioritize them still. Definitely limited to a small number of users though.

rnyberg:
Either **1)** This or **2)** toggle for being enabled but not accepting new downloads would be really nice. 

Another use case for the feature would be migrating from one download client to another on the fly. Older secondary client can still churn through whatever is in the queue while new one gets the fresh stuff.

markus101:
Done in https://github.com/Sonarr/Sonarr/pull/2278

PR #2278, opened 2017-11-08
Round-robin over available Download Client

Taloth:
#### Database Migration
**YES** - 132
Adding Priority column to DownloadClients, defaults to 1.
Migration numbers enabled clients for backward compat preserving the old 'First' logic.

#### Description
The round-robin is tested only by unit-tests so far, never actually ran it.
But the problem is that once we do this, it'll round-robin instead of picking the first. Which may affect existing setup and does not allow people to have a 'primary' and 'fallback' mechanism till we have a priority setting.
So up for review, but otherwise parked till we can get that fleshed out.

#### Todos
- [x] Tests
- [x] Config option for prioritization instead of round-robin only.

* https://github.com/Sonarr/Sonarr/issues/1866


devbrian:
Just wanted to make a note on this, I have been using this in production on my server for a month now with no issues.

devbrian:
Any update on this being added to phantom possibly? @markus101

sachk:
Current status on this? I'd be nice to see this in master.

devbrian:
any chance on getting this added to V3?

fryfrog:
For this to be feasible, I'd think one would need to be able to toggle between round robin and *all*.

Taloth:
@fryfrog not really, 'all' has no meaning. Currently Sonarr picks the first, or if the first is down, it'll pick the second.

@devbrian Yeah, I'll look into getting it merged in, although 3.x basically means "3.0 or later".

fryfrog:
If you're going to the trouble of adding this "first" or "round-robin" option, you should go to the trouble of also adding an "all" option. I've seen it asked for as much as round robin.

Taloth:
But how do you envision 'All'? It's not like you want the same torrent to be submitted to all clients simultaneously.

fryfrog:
*I* certainly don't, nor do *I* even want round robin. But I've literally had multiple requests by people who want to send them to *all* their download clients. They both seem about as equally useful to me, which is not very.

All could be used by people w/ a seedbox *and* local torrents where the seedbox is used for seeding and the local is what they use for importing. It could also be used for multiple uncoordinated sites, kind of the opposite of round robin.

They're all dumb uses and honestly I think this will just complicate the UI for 99% of users. But if you're going to all this trouble to provide a super niche option... at least make the *other* two people happy too.

Taloth:
But that is a huge complication, because suddenly you have to ignore one client when considering imports. They should have a pp-script to sync the remote torrent to local.
The idea behind round-robin/fallbacks is that you can have have a main client that does most of the work, but a fallback client on a NAS that catches the slack. The current UI allows the user the add multiple clients, but literally the sequence at which you add them determines the priority.
It's definitely a niche feature. I'll probably have to dive into my irc chat logs to find out why I even started working on it, but here we are... 5 inches from the finish.

fryfrog:
Ah, nevermind then :)

Taloth:
@markus101 I opted to put it at the bottom because the hint text meant putting it next to Enable didn't make it look any better.

image

In the overview I also added a "Prio: x" badge if Prio > 1. It currently has kind.DISABLED because it looks best, but some tweaking there might be necessary.

image

devbrian:
Hopefully we get a merge soon :).  Great work Taloth!

fryfrog:
How about `Pri` instead of `Prio`? :)

Taloth:
Pri could be Primary.

it's waiting for a review from @markus101

fryfrog:
True, so just write the whole word. Don't abbreviate one and not the other. :)

Taloth:
I hate you... :smile:

src/NzbDrone.Core/Datastore/Migration/132_add_download_client_priority.cs
```diff
@@ -0,0 +1,45 @@
...
+                {
+                    int i = 1;
+                    while (reader.Read())
+                    {
+                        var id = reader.GetInt32(0);
+                        
+                        using (var updateCmd = conn.CreateCommand())
+                        {
+                            updateCmd.Transaction = tran;
+                            updateCmd.CommandText = "UPDATE DownloadClients SET Priority = ? WHERE Id = ?";
```
markus101: Should we be increasing the priority of clients of different types?

I have a torrent client and usenet client enabled, but my torrent client got priority 2 because I added it second. Instead I would expect to see both have a priority of 1.

A simple fix would be to have two counters, one for each protocol.

fryfrog:
:o

Taloth:
Less than 2 years. Got even older branches.

Taloth:
@devbrian Have you been able to test the latest v3 version? (Lidarr/Radarr wants to know if it's safe to port over :smile:)

PR #2282, opened 2017-11-12
Custom Specials Season Format

beschoenen:
#### Database Migration
YES

#### Description
This PR will allow users to set their own name for the "Specials" season.

#### Todos
- [ ] Tests
- [ ] Documentation


markus101:
This isn't something we want to support. They can either use the default `Specials` or they can add the season folder to the episode file name and customize it there.

Issue #1879, opened 2017-04-29
Conditionally adding absolute episode number


markus101:
Currently this doesn't work: `{.absolute:000}` as the period causes it to fail. This would require some changes to the number replacements we do and ensuring we handle ranges correctly.

We will also need to always use the anime format when renaming anime episodes, currently we don't if any of the episodes don't have an absolute episode number.

markus101:
The import will be delayed if it's missing (same as the episode title specification that delays it), which will better handle this case. No plans to make it optional any longer.

PR #2283, opened 2017-11-12
Use anime formatting when episode does not have absolute number

beschoenen:
#### Database Migration
NO

#### Description
The current behavior is kind of weird and unexpected. This PR tells Sonarr to still use anime formatting when an episode doesn't have an absolute number and the formatting doesn't use the absolute number in its formatting.

#### Todos
- [x] Tests
- [ ] Documentation


markus101:
Somewhat related: https://github.com/Sonarr/Sonarr/issues/1879

Falling back to standard naming when the the absolute number is missing is a bit unexpected regardless, making it conditional would be an improvement I think, though it'd require the season + episode number or the filename would be quite useless. Perhaps a good middle ground would be to block the automatic import and require manual intervention, though also unexpected.

beschoenen:
And what if I were to add a message in the `?` popup after the input field for the Anime naming schema, something along the lines of:

"If your scheme uses absolute numbering, but the absolute number is not available then the normal scheme will be used."

markus101:
I've implemented something similar to the episode title import specification to delay imports if the absolute episode number is missing and the episode recently aired, after which it'll fallback to the standard format as it does now.

PR #2295, opened 2017-11-18
Add missing error check when adding a magnet link to deluge

smmalis37:
#### Database Migration
NO

#### Description
Add missing error check when adding a magnet link to deluge. The same error check is already present for adding from a file, just not a magnet.

### DIFF
--- a/src/NzbDrone.Core/Download/Clients/Deluge/Deluge.cs
+++ b/src/NzbDrone.Core/Download/Clients/Deluge/Deluge.cs
@@ -35,6 +35,11 @@ namespace NzbDrone.Core.Download.Clients.Deluge
         {
             var actualHash = _proxy.AddTorrentFromMagnet(magnetLink, Settings);
 
+            if (actualHash.IsNullOrWhiteSpace())
+            {
+                throw new DownloadClientException("Deluge failed to add magnet " + magenetLink);
+            }
+
             if (!Settings.TvCategory.IsNullOrWhiteSpace())
             {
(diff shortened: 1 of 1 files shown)


markus101:
@Taloth I think you discussed this on IRC, what's the verdict?

Taloth:
No verdict exactly. I simply stated that we first need to know why deluge would reject it and how, so we can properly mock DelugeProxy.
The suggested change doesn't 'hurt', so I'm ok to have it merged in, it's just incomplete imho.

smmalis37:
Unfortunately I've been unable to reproduce the behavior this was meant to fix, so I'm unable to get logs with the details Taloth requested.

src/NzbDrone.Core/Download/Clients/Deluge/Deluge.cs
```diff
@@ -35,6 +35,11 @@ protected override string AddFromMagnetLink(RemoteEpisode remoteEpisode, string
         {
             var actualHash = _proxy.AddTorrentFromMagnet(magnetLink, Settings);
 
+            if (actualHash.IsNullOrWhiteSpace())
+            {
+                throw new DownloadClientException("Deluge failed to add magnet " + magenetLink);
```
Thirrian: magenetLink (typo) 😃

smmalis37: That's what I get for making this change directly in github's ui...

PR #2334, opened 2017-12-13
Mono internals does not properly copy/move symlinks

Taloth:
Based on the IRC chat with Ingram.

Mono doesn't properly move/copy symlinks if they're not on the same filesystem. And instead copies the content.
This PR patches that logic in the DiskProvider. It should also properly handle relative symlinks.

I've also added a check in TryCreateHardlink so that it doesn't try hardlink a symlink, I don't think that would actually ever hardlink but better safe than sorry.

This should make some scenarios work better for the Move/hardlink/symlink issue that's been lingering.

### DIFF
--- a/src/NzbDrone.Mono/Disk/DiskProvider.cs
+++ b/src/NzbDrone.Mono/Disk/DiskProvider.cs
@@ -1,4 +1,4 @@
-using System;
+﻿using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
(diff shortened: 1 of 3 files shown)


src/NzbDrone.Mono/Disk/DiskProvider.cs
```diff
@@ -96,11 +96,88 @@ public override List<IMount> GetMounts()
...
+            if (sourceInfo.IsSymbolicLink)
+            {
+                var isSameDir = UnixPath.GetDirectoryName(source) == UnixPath.GetDirectoryName(destination);
+                var symlinkInfo = (UnixSymbolicLinkInfo)sourceInfo;
+                var symlinkPath = symlinkInfo.ContentsPath;
+
+                var newFile = new UnixSymbolicLinkInfo(destination);
+
+                if (isSameDir)
+                { // We're in the same dir, so we can preserve relative symlinks.
```
markus101: Comment on it's own line.

src/NzbDrone.Mono/Disk/DiskProvider.cs
```diff
@@ -96,11 +96,88 @@ public override List<IMount> GetMounts()
...
+
+                var newFile = new UnixSymbolicLinkInfo(destination);
+
+                if (FileExists(destination) && overwrite)
+                {
+                    DeleteFile(destination);
+                }
+
+                if (isSameDir)
+                { // We're in the same dir, so we can preserve relative symlinks.
```
markus101: Comment on it's own line.

PR #2337, opened 2017-12-14
Discussion: Transfer providers

Taloth:
#### Database Migration
YES, but not added yet.

#### Description
Replace Remote Path Mapping with a flexible system to handle mounts, external copying mechanisms, unrar etc

Intended for implementation in v3, not v2.

#### Todos
- [ ] Download Client integration
- [ ] Import process integration
- [ ] Custom Transfer Implementation
- [ ] UI
- [ ] ...
- [ ] Tests
- [ ] Documentation

Not gonna list any since it's not a PR, but a discussion thread. Don't want to get ppl excited.


Taloth:
Use Case: Local union fs with multiple remote download clients. Transfers occur on the download servers and stay on that server.
Previously known usecase would be a remote download client uploading to a single cloud store directly, instead of first copying to the Sonarr host.

devbrian:
Let me see if I understand this pull request, as this might be exactly what I am doing on my setup, with a much worse solution.

My series path is `/mnt/unionfs/Media/TV/my_series_name`

I have download clients, which are remote servers, with completed directories like this:
`/mnt/remote1/completed/tv/`
`/mnt/remote2/completed/tv/`
`/mnt/remote3/completed/tv/`

When a download comes on `remote1`, it will be moved to - 

`/mnt/remote1/Media/TV/my_series_name`

Is this correct? 
@Taloth 
@markus101  

Would be great to see this in v3 if this is the case!

Issue #2325, opened 2017-12-11
Can’t connect to Download Client while it is downloading from filehosting


enchained:
### System Information/Logs

**Sonarr Version:** 2.0.0.5054

**Operating System:** Synology DSM 6.1.4-15217 Update 2

**.net Framework (Windows) or mono (macOS/Linux) Version:** 4.6.2

**Link to Log Files (debug or trace):** https://pastebin.com/2c6N8XPS

### Additional Information

I have Synology DS216+ii and I use its Download Station as my download client. It can download not only torrents, but files from filehostings too (the ones with countdowns and captchas). I added a few of filehosting downloads in its queue (NOT tagged as sonarr and NOT in sonarr folder) and it looks like it broke download client connection due to unrecognized statuses.

Sonarr System says: `All download clients are unavailable due to failures`

When I go to download clients and press Test it says:
`Failed to get the list of torrents: Error converting value "filehosting_waiting" to type 'NzbDrone.Core.Download.Clients.DownloadStation.DownloadStationTaskStatus'. Path 'data.tasks[154].status', line 1, position 123112.`

or
`Failed to get the list of torrents: Error converting value "captcha_needed" to type 'NzbDrone.Core.Download.Clients.DownloadStation.DownloadStationTaskStatus'. Path 'data.tasks[153].status', line 1, position 122570.`

I can provide a filehosting link to reproduce, but probably reproduceable with any synology-supported filehosting.

margaale:
Hi, just saw this issue. I'm looking into it. I should be really simple to solve...

margaale:
There it is...
Maybe @enchained could provide me a link so i can test if its works

markus101:
Fixed via https://github.com/Sonarr/Sonarr/commit/747f3e171c752a18148c1d115679ba2b9710d260

PR #2342, opened 2017-12-17
DownloadStation: Wrong status in enum

margaale:
#### Database Migration
NO

#### Description
Adding captcha_needed status as it wasn't listed in the docs. Changed the other two statuses to fix their names.

*  #2325


Taloth:
I actually was planning to do a JsonEnumConverter that properly converts the strings. Since download station uses a lowercase format with underscores, should be easy enough to do based on the Upper CamelCase enum values.

margaale:
Something like this? I never did it before...
I couldn't run the test thou, i'm at work now

Taloth:
Something like that yes, although I'll make it a bit more generic so it'll actually work on all enums instead for just that one.

Taloth:
Merged via https://github.com/Sonarr/Sonarr/commit/747f3e171c752a18148c1d115679ba2b9710d260

Issue #1872, opened 2017-04-25
Parser should consider E00 a placeholder for multiple specials


Taloth:
Scene releases with SxxE00 are in fact specials that do not have a specific episode number, and thus should trigger the Special Parser logic instead of returning x00 as number.
After E00 there's the special episode name and provides a meaningful delimiter similar to the '.Special.' or '.S01.Special.' that often appears.

Taloth:
Another edgecase with this is sense8 coz it actually has scene E00 but tvdb E01.

Taloth:
Fixed via #2351

PR #2351, opened 2017-12-26
Consider all scene SxxE00 releases Specials as well.

Taloth:
Pretty old issue about E00 actually being Specials.

@markus101  I'm not sure if there could be bad consequences of this change. So a PR first.

The Sense8 edge-case with episode shift seems to work because it doesn't actually manage to work out which special it is and thus fall back to the normal logic.

Issue https://github.com/Sonarr/Sonarr/issues/1872

### DIFF
--- a/src/NzbDrone.Core/Parser/ParsingService.cs
+++ b/src/NzbDrone.Core/Parser/ParsingService.cs
@@ -60,7 +60,11 @@ namespace NzbDrone.Core.Parser
                 parsedEpisodeInfo = Parser.ParsePath(filename);
             }
 
-            if (parsedEpisodeInfo == null || parsedEpisodeInfo.IsPossibleSpecialEpisode)
+            if (parsedEpisodeInfo != null && parsedEpisodeInfo.IsPossibleSceneSeasonSpecial && GetStandardEpisodes(series, parsedEpisodeInfo, true, null).Any())
+            {
+                // SxxE00 episodes are sometimes mapped via TheXEM, don't use episode title parsing in that case.
+            }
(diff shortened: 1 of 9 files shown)


Taloth:
@markus101 As mentioned on hangouts I was working on it a bit more, can you review the last commit?

I also queried skyhook. There are a lot of episodes named 'Christmas Special' so simply removing those keywords would be problematic. Once we run into specific mismatches we can add the appropriate test-case, ... although I think we've been pointing at the Advanced Parser issue for all those mismatches. :smile:

src/NzbDrone.Core/Parser/ParsingService.cs
```diff
@@ -60,7 +60,11 @@ public LocalEpisode GetLocalEpisode(string filename, Series series, ParsedEpisod
                 parsedEpisodeInfo = Parser.ParsePath(filename);
             }
 
-            if (parsedEpisodeInfo == null || parsedEpisodeInfo.IsPossibleSpecialEpisode)
+            if (parsedEpisodeInfo != null && parsedEpisodeInfo.IsPossibleSceneSeasonSpecial && GetStandardEpisodes(series, parsedEpisodeInfo, true, null).Any())
+            {
+                // SxxE00 episodes are sometimes mapped via TheXEM, don't use episode title parsing in that case.
```
markus101: What's the purpose of this `if`? It's not preventing falling into an `else` and it doesn't do anything on it's own.

Taloth: It prevents falling into the elseif, IF the E00 episode is mapped by TheXEM. This causes it to bypass the FindByTitle and go straight to the standard handling.

But if that's unclear based on reading the code, then it needs to be refactored. I'll see what I can do to clear it up.

PR #2355, opened 2017-12-30
Feature/manual scene numbering

beschoenen:
#### Database Migration
NO

#### Description
Adds a toggle to the shows edit UI to toggle scene numbering for a show.

#### Todos
- [ ] Tests
- [ ] Documentation


Taloth:
The most common scenario where scene mappings aren't correct is when scene and p2p don't 'agree' on a mapping, preventing TheXEM from mapping it.   
Thus, the only proper solution would be to improve TheXEM to be able to handle that scenario, something that's been discussed in TheXEM v2 requirements but afaik TheXEM doesn't have the development capacity atm to make it a reality yet.   
We're also doing work on the next version of Skyhook, with the intent to support foreign languages and possibly different metadata sources. That might also provide the necessary infrastructure to do this as well, but that's just a spot on the horizon and not available right now.

If you toggle this switch then suddenly WEB-DL releases might be imported properly but scene HDTV releases would not, so you end up with the same problem, just the other way around. This would mean the user would have to add restricted keywords to block release groups for that specific series.   
IMHO that would lead to an increase in support requests and confusion.

The one argument for this feature is that it puts the responsibility in the user's hand, but that's not where Sonarr was created for. Sonarr isn't Flexget and doesn't allow infinite customization.
That doesn't mean we don't want to deal with this particular problem, but we really want it done right.

beschoenen:
Is there a way we can hide this behind advanced settings? 
My current use case for this is the Pokemon Anime, I download the Japanese version which is followed by TheTVDB, but Scene numbering alters the seasons.

Taloth:
Even as an Advanced Option we'd have the same problem. People love toggling switches they're not supposed to and then call us with support questions coz something broke.

As for Pokemon, that's something I kinda disagree with TheXEM. They caved and switched to the US releases instead of keeping to the Japanese.
But it's still something that needs to be fixed on TheXEM. If their api exposes how Sonarr should differentiate between the two then we could handle it. For example by specifying that the mapping only applies for the 'Standard' series type. Although the season-specific series names could be a problem as well in that case.

Taloth:
@beschoenen I had a discussion with TheXEM mods about Pokemon. TheXEM uses the Pokemon US numbering. (as an exception coz normally they follow the original Japanese releases.)
This was done because there weren't any reliable Japanese sources available that could be used as a consistent mapping. I don't know specifics and tbh I don't want to get involved in those details.

I suggest you talk to them on the \#xem IRC channel on freenode.
If we can make a reliable mapping somehow on TheXEM then I would be willing to make Sonarr use a different mapping for the Anime series type vs the Standard series type. But I would only do that if TheXEM mods agree that it creates a workable solution.

xelra:
Disabling scene numbering is a much harder logic problem than one might think at first.

What exactly needs to be disabled? Only numbering? Numbering and aliases? Show wide aliases? Only season aliases? Or should everything be disabled?

If I had to vote, I'd say everything, because everything in between will just be an awkward state.

Now let's assume there is a feature to disable everything for a show, then that's just a single state of many possibilities. It would rather be a coincidence that the show then works, than a rule.

Usually there's a reason why a show is on XEM and that's because the tvdb map isn't identical with the scene map. The tvdb map is the rule. So someone already took action to say "the tvdb rule doesn't work". Now with the scene disable toggle the user is essentially saying "no, you're wrong, I want the tvdb map back".

Pokemon is really the only show that this applies to. We only have the English scene releases, because there are no well formated releases in the original language available.

If you have different releases, then you will have to use manual import. So the problem is in a general sense less related to Pokemon, but to importing releases with custom numbering. 
In my opinion that's on of the main use cases for a manual import. Unfortunately the manual import feature also applies scene rules.

My suggestion to fix this issue in a general way is to add a "disable scene rules" toggle to the manual import feature. And also add it to the appropriate endpoint on the api for automated manual imports.

PR #2361, opened 2018-01-05
Regex support in restrictions

Taloth:
Wanted to do this for a while. I thought I had some shelved code for it, but couldn't find it so started from scratch.
It's definitely not perfect. I simply use the perl /pattern/ format to differentiate it from standard case-insensitive terms, but the pattern itself is .net regex. (not that I expect people to create complicated regexes there, but worth noting that it isn't actually perlre.)

One of the concerns i have is that you'd want to give the user a way to easily specify anchors. Like the `[ ._-]` we use ourselves, but then shorter. I figured we could add a modifier like `/_WEB-DL_/v` that tells it to substitute _ with `[ ._-]`, but in most cases \b would suffice anyway.

Examples:
`NTb`: Simple case-insensitive term, existing behavior.
`/WEB-DL/`: Case sensitive regex
`/WEB-DL/i`: Case insensitive regex
`/\bWEB\b/`: Matches `Something.720p.WEB.x264` but not `WEBISODE` due to \b anchors.
`/-NTb$/`: Only match ntb release group.
`/\b[xh]265\b/i`: Either x265 or x265 codecs.

### DIFF
--- a/src/NzbDrone.Core/Restrictions/TermMatcher.cs
+++ b/src/NzbDrone.Core/Restrictions/TermMatcher.cs
@@ -1,0 +1,63 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Text.RegularExpressions;
+using NzbDrone.Common.Cache;
+
+namespace NzbDrone.Core.Restrictions
+{
+    public interface ITermMatcher
+    {
+        bool IsMatch(string term, string value);
+    }
+
+    public class TermMatcher : ITermMatcher
+    {
+        private ICached<Predicate<string>> _matcherCache;
+
+        public TermMatcher(ICacheManager cacheManager)
+        {
+            _matcherCache = cacheManager.GetCache<Predicate<string>>(GetType());
+        }
+
+        public bool IsMatch(string term, string value)
+        {
+            return GetMatcher(term)(value);
+        }
+
+        public Predicate<string> GetMatcher(string term)
+        {
+            return _matcherCache.Get(term, () => CreateMatcherInternal(term), TimeSpan.FromHours(24));
+        }
+
+        private Predicate<string> CreateMatcherInternal(string term)
+        {
+            Regex regex;
+            if (PerlRegexFactory.TryCreateRegex(term, out regex))
+            {
+                return regex.IsMatch;
+            }
+            else
+            {
+                return new CaseInsensitiveTermMatcher(term).IsMatch;
+
+            }
+        }
+
+        private sealed class CaseInsensitiveTermMatcher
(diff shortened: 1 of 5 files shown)


src/NzbDrone.Core/Restrictions/TermMatcher.cs
```diff
@@ -0,0 +1,63 @@
...
+        {
+            Regex regex;
+            if (PerlRegexFactory.TryCreateRegex(term, out regex))
+            {
+                return regex.IsMatch;
+            }
+            else
+            {
+                return new CaseInsensitiveTermMatcher(term).IsMatch;
+
```
markus101: Nit: Extra new line.

Jorman:
Hi, tnx for this.

I've one question, I'm very unconfortable with regex, so I try with some program to help in definition, the problem is, for example with this
`/\b[xh]265\b/i:` Either x265 or x265 codecs.

I got this
`(?i-x)[hx]265|hevc`

I don't know if Sonarr will interpreter in the right way this string.

Is possible to know which format I've to write the string? I have many languages on program
C++, PHP, Perl, Phyton ...

Tnx, Jo

Taloth:
I didn't bother writing a help/wiki page for it yet coz it's pretty advanced... okok, coz I'm lazy.

It's the .net regex flavor but uses perl encapsulation format for easier differentiation.
Perl encapsulation is `/{pattern}/{modifiers}` so in your case it'd be `/[hx]265|hevc/i` or `/(?i-x)([hx]265|hevc)/` (-x isn't needed, coz whitespace is not ignored by default)
but better would be `/\b([hx]\.?265|hevc)\b/i` so with \b anchors and optionally handles h.265 too (coz it happens).

Jorman:
Lol, I'm lazy too! For me regular expression is a kind of mistry!

For that I hope in some help in some program, out there thre's a lot of program that helps to make string!
Tnx for the info, you say that would be better
/\b([hx]\.?265|hevc)\b/i
Ok, I'm with you with this, but what format is that? Is not perl encapsulation, I can't find how to extract this code

Taloth:
https://perldoc.perl.org/perlrequick.html

That's where the slashes come from. Everything inside it is .net regex flavor, but that's largely irrelevant for basic patterns: https://regexr.com/3mi8i

Jorman:
Ok, so always perl but with
`/{pattern}/{modifiers}`
So now I'll try to find a perfect strings for me :D
Would be good, in future, have a collection of user regex!

Jorman:
Hi, tnx for you idea I successfully create
`/\b([hx]?[.-]?265|hevc)\b/i` Shared if some need it

Issue #2009, opened 2017-06-28
Drone Factory relics


demonbane:
I just disabled the Drone Factory per the warnings in the most recent version and I noticed a few things that are left over even after disabling it.

1. "Rescan Drone Factory Folder" on the Wanted tab is still there, even though I've disabled Drone Factory altogether.
2. Drone Factory still shows up under Settings -> Download Client after it's disabled. Since it's getting removed, it would make sense to just hide the setting once it's disabled.

Both of these are largely cosmetic issues, I just wanted to get them documented so that when DF does get killed off for good there's a record of details to be handled.

markus101:
Yeah, these will be removed in v3.

Taloth:
Actually markus, hiding the Rescan Drone Factory Folder button is a sound option, even for v2.
Not gonna remove the Drone Factory folder field, it's already an Advanced Setting and people need to option to have it enabled in v2, despite the warning.

PR #2363, opened 2018-01-08
Hide "Rescan Drone Factory Folder" if no Drone Factory folder is setup

JoshSchreuder:
#### Database Migration
NO

#### Description
Hides the "Rescan Drone Factory Folder" on the Wanted page if the user has no Drone Factory folder configured

#### Todos
- [x] Tests
- [n/a] Documentation

* #2009 (up-for-grabs)

### DIFF
--- a/src/NzbDrone.Automation.Test/NzbDrone.Automation.Test.csproj
+++ b/src/NzbDrone.Automation.Test/NzbDrone.Automation.Test.csproj
@@ -58,16 +58,17 @@
     <Reference Include="System.Xml" />
     <Reference Include="System.Xml.Linq" />
     <Reference Include="Microsoft.CSharp" />
-    <Reference Include="WebDriver, Version=3.2.0.0, Culture=neutral, processorArchitecture=MSIL">
-      <HintPath>..\packages\Selenium.WebDriver.3.2.0\lib\net40\WebDriver.dll</HintPath>
+    <Reference Include="WebDriver, Version=3.8.0.0, Culture=neutral, processorArchitecture=MSIL">
+      <HintPath>..\packages\Selenium.WebDriver.3.8.0\lib\net40\WebDriver.dll</HintPath>
     </Reference>
-    <Reference Include="WebDriver.Support, Version=3.2.0.0, Culture=neutral, processorArchitecture=MSIL">
-      <HintPath>..\packages\Selenium.Support.3.2.0\lib\net40\WebDriver.Support.dll</HintPath>
+    <Reference Include="WebDriver.Support, Version=3.8.0.0, Culture=neutral, processorArchitecture=MSIL">
+      <HintPath>..\packages\Selenium.Support.3.8.0\lib\net40\WebDriver.Support.dll</HintPath>
     </Reference>
   </ItemGroup>
   <ItemGroup>
     <Compile Include="AutomationTest.cs" />
     <Compile Include="AutomationTestAttribute.cs" />
+    <Compile Include="WantedPageTest.cs" />
     <Compile Include="MainPagesTest.cs" />
     <Compile Include="PageModel\PageBase.cs" />
     <Compile Include="Properties\AssemblyInfo.cs" />
--- a/src/NzbDrone.Automation.Test/PageModel/PageBase.cs
+++ b/src/NzbDrone.Automation.Test/PageModel/PageBase.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Threading;
 using OpenQA.Selenium;
 using OpenQA.Selenium.Remote;
@@ -25,6 +25,11 @@ namespace NzbDrone.Automation.Test.PageModel
         {
             var wait = new WebDriverWait(_driver, TimeSpan.FromSeconds(timeout));
             return wait.Until(d => d.FindElement(by));
+        }
+
+        public bool Displayed(By by)
+        {
+            return _driver.FindElement(by).Displayed;
         }
 
         public void WaitForNoSpinner(int timeout = 30)
--- a/src/NzbDrone.Automation.Test/WantedPageTest.cs
+++ b/src/NzbDrone.Automation.Test/WantedPageTest.cs
@@ -1,0 +1,46 @@
+using System.Threading;
+using FluentAssertions;
+using NUnit.Framework;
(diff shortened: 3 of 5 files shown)


src/NzbDrone.Automation.Test/NzbDrone.Automation.Test.csproj
```diff
@@ -58,16 +58,17 @@
...
     <Reference Include="Microsoft.CSharp" />
-    <Reference Include="WebDriver, Version=3.2.0.0, Culture=neutral, processorArchitecture=MSIL">
-      <HintPath>..\packages\Selenium.WebDriver.3.2.0\lib\net40\WebDriver.dll</HintPath>
+    <Reference Include="WebDriver, Version=3.8.0.0, Culture=neutral, processorArchitecture=MSIL">
+      <HintPath>..\packages\Selenium.WebDriver.3.8.0\lib\net40\WebDriver.dll</HintPath>
     </Reference>
-    <Reference Include="WebDriver.Support, Version=3.2.0.0, Culture=neutral, processorArchitecture=MSIL">
-      <HintPath>..\packages\Selenium.Support.3.2.0\lib\net40\WebDriver.Support.dll</HintPath>
+    <Reference Include="WebDriver.Support, Version=3.8.0.0, Culture=neutral, processorArchitecture=MSIL">
+      <HintPath>..\packages\Selenium.Support.3.8.0\lib\net40\WebDriver.Support.dll</HintPath>
```
JoshSchreuder: Apologies on this - I was unable to get WebDriver running locally with FF 57 unless I upgraded this. Happy to revert this change if you want.

src/UI/Wanted/Missing/MissingLayout.js
```diff
@@ -77,8 +79,16 @@ module.exports = Marionette.Layout.extend({
...
     onShow : function() {
+        var self = this;
+
+        this.downloadClientSettings = new DownloadClientSettingsModel();
         this.missing.show(new LoadingView());
-        this._showToolbar();
+
+        Backbone.$.when(this.downloadClientSettings.fetch()).done(function() {
+            self.droneFolder = self.downloadClientSettings.attributes.downloadedEpisodesFolder;
+            self._showToolbar();
```
JoshSchreuder: This does add a slight delay while the settings are fetched before rendering the toolbar - feedback appreciated on if this is the best way of going about it as I wasn't too sure.

src/NzbDrone.Automation.Test/WantedPageTest.cs
```diff
@@ -0,0 +1,46 @@
...
+
+        [Test]
+        public void show_drone_update_if_drone_folder_set()
+        {
+            page.SettingNavIcon.Click();
+            page.WaitForNoSpinner();
+
+            page.FindByClass("x-download-client-tab").Click();
+            page.FindByClass("x-advanced-settings").Click();
+            page.Find(By.Name("downloadedEpisodesFolder")).SendKeys("C:\\");
```
JoshSchreuder: Is there a drone folder value I can use that lets this test pass cross-platform?

src/NzbDrone.Automation.Test/PageModel/PageBase.cs
```diff
@@ -27,6 +27,11 @@ public IWebElement Find(By by, int timeout = 5)
             return wait.Until(d => d.FindElement(by));
         }
 
+        public bool Displayed(By by)
+        {
+            return _driver.FindElement(by).Displayed;
```
JoshSchreuder: Added this so I can assert the actual exception rather than the wait timeout exception from the above wrapper methods.

src/NzbDrone.Automation.Test/WantedPageTest.cs
Taloth: There's an `.AsOsAgnostic()` extension method, but I don't think that'll work coz the validator checks if the folder exists. Better just create a temporary folder somewhere. The NzbDroneRunner instance has an AppData property that contains a temp folder just for that test. So create a subfolder there and use it as fake drone factory folder.
But don't spend too much time on those automation tests. Atm they only test whether the UI shows up and the main pages are browsable, little more than a smoketest.

src/NzbDrone.Automation.Test/NzbDrone.Automation.Test.csproj
Taloth: Build agent seemed to be happy with 3.8 so 👍

src/UI/Wanted/Missing/MissingLayout.js
```diff
@@ -144,6 +148,16 @@ module.exports = Marionette.Layout.extend({
                 }
             ]
         };
+
+        if (this.droneFolder) {
+            leftSideButtons.items.splice(4, 0, {
```
Taloth: I don't like this magically splicing in the button. Options:
- Append Manual Import as well
- Keep as array, and remove the Drone Factory by findIndex.
- Add 'hidden' property and extend ToolbarLayout to skip it if hidden is or evaluates to true.

src/UI/Wanted/Missing/MissingLayout.js
Taloth: @markus101 This one's for you, I don't know what's the best way to deal with this.

markus101: A few ideas:

- Fetch download client settings when the app loads, much like UI settings and ensure it stays in sync when changes are made
- Fetch it when loading Wanted: Missing or settings and update it if the settings are changes. If it never changes it's cached, if it does, we can re-render and remove the button.
- Fetch during initialize then re-rendering after the request completes (if the display of the button would change)

There could be some re-rendering of the buttons if the visibility changes after the initial showing, but if we moved that button the the last position it'd just be added/removed (minus some minor styling changes)

Taloth: The caching approach is definitely the best, although I wonder if it's worth it considering v3 gets rid of the button entirely.

Two options that don't involve caching stuff:
1. Initial render with button disabled (not removed). Fetch and re-render (enabling the button) if needed.
2. Initial render with button removed. Fetch and re-render (adding the button) if needed.

1 has the least flicker, 2 is useful for most users since by now they shouldn't have the DF anymore.

Issue #480, opened 2015-02-04
Health check if mono is not running with the `--debug` switch


SonarrBot:
Without it stack traces are mostly useless, which sucks. We should recommend that its enabled and let users know to contact the package developer if they installed it via a package manager.

kayone:
we should include a `sonarr` script file with all the proper flags and get people to run that instead of asking them to run `mono /opt/nzbdrone/nzbdrone.exe`

markus101:
There are a number of scripts on the wiki for auto starting (init.d, upstart, etc), as well as a lot of other platforms that have plugins that aren't quite right, but yes, we should have a list of start up scripts that are viable and run mono the way Sonarr needs to.

kayone:
I'm not even talking about an autostart script, I'm talking about a simple script that just starts sonarr, then auto start scripts could call that.

markus101:
Ahhh, gotcha, makes sense!

Taloth:
Since we plan to provide the startup scripts in v3, this isn't really important, but it should be easy to do. I think it's possible to get the stackframe from mono so you can check if the sourcecode info is present, if it is then --debug was on.

margaale:
Is this still usefull? I was thinking I would give it a try, it actually seems very easy to do

markus101:
Yeah, still useful, without it stack traces don't have line numbers and as useful.

PR #2371, opened 2018-01-13
check if mono is running with --debug arg

margaale:
#### Database Migration
NO

#### Description
I Haven't test on a linux box. See issue #480

### DIFF
--- a/src/NzbDrone.Core/HealthCheck/Checks/MonoDebugCheck.cs
+++ b/src/NzbDrone.Core/HealthCheck/Checks/MonoDebugCheck.cs
@@ -9,6 +9,8 @@ namespace NzbDrone.Core.HealthCheck.Checks
         private readonly Logger _logger;
         private readonly StackFrameHelper _stackFrameHelper;
 
+        public override bool CheckOnSchedule => false;
+
         public MonoDebugCheck(Logger logger, StackFrameHelper stackFrameHelper)
         {
             _logger = logger;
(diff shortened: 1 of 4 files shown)


src/NzbDrone.Core/HealthCheck/Checks/MonoDebugCheck.cs
```diff
@@ -0,0 +1,44 @@
...
+        private readonly Logger _logger;
+        private readonly StackFrameHelper _stackFrameHelper;
+
+        public MonoDebugCheck(Logger logger, StackFrameHelper stackFrameHelper)
+        {
+            _logger = logger;
+            _stackFrameHelper = stackFrameHelper;
+        }
+
+        public class StackFrameHelper
```
margaale: Where shoud i put this class? I had to create it just for mocking purpose...

Taloth:
@markus101 Maybe we should do this only for new installs? Or just once during startup and log a warn instead of a health check.
I'm just not sure about it all coz ppl can't dismiss health checks.

markus101:
A warning during start would work. We can always add a health check later once we support dismissing them.

Taloth:
K, then i'll have this healthcheck log a warn, but return ok.

Taloth:
Merged via https://github.com/Sonarr/Sonarr/commit/13a259b47313100b53dfacd4423107da19a2cf9d

Issue #991, opened 2015-12-11
Don't overwrite metadata files after creation


markus101:
Emby watched statuses are nuked () and people that use something else to process them (convert MPAA ratings to another system), lose those changes.

Simple solution is to write once and never again, this would solve both issues and be similar to how Kodi/Plex behave in terms of getting metadata.

kayone:
Ideally we could even get smarter, where we surgically updated certain parts, episode name, number, description etc.

randomevents:
I have a suggestion because I (wrongly) assumed this is the way it already worked until the other day; If a show or season is flagged as Unmonitored, then don't update the metadata files. It's a simple check, and can be ignored if the user triggers a scan and refresh. DVD order issues aside (it's not a hard problem to work around), it's the retroactive adjustments from TheTVDB that are such a pain in the ass.

markus101:
I don't think we want to conflate the unmonitored flag with turning off other functionality. I don't think it will be obvious to the user and turning off metadata seems like could go a step further and stop getting data from TheTVDB, which seems really extreme.

Why keep the series in Sonarr if you don't want it alter anything? With in unmonitored it wouldn't upgrade any files either.

randomevents:
Couldn't (un)monitor upgrade files also read as (un)monitor metadata changes? Which is how I understood it; if I didn't want new files, why would I want new metadata? It's also seems like it would be a shame to have to clutter up the UI/

There's plenty of examples; manually editting NFOs from previous seasons (with another tool like media companion), another horrible act of violence messing up story order. The forum tiffs between the mods and editors at the TheTVDB that mess up the metadata but don't trigger file upgrades when the file is at the max, etc etc. It's especially horrible with cartoons, there's alot of fixes to be made between tvdb and the scene, and it doesn't help that theXEM isn't letting anyone sign up to do edits, so a lot of stuff doesn't get done.

Sure it's easy enough to remove dead shows, but I think like it or not, your team has made a great enough product that many people use it as a tv show media manager for all their stuff.

markus101:
Unmonitor stops season searching and RSS from automatically grabbing results, but you could search for one episode or do a manual search and Sonarr would import it even if it was unmonitored.

> There's plenty of examples; manually editting NFOs from previous seasons (with another tool like media companion), another horrible act of violence messing up story order.

Aired vs DVD ordering?

[...]

PR #2436, opened 2018-03-01
Improve Kodi Metadata provider handling of scene nfo files

### LINKED ISSUE #932: Preserved watched flag in emby nfo file
Emby modifies the nfo once it's watched. However, Sonarr apparently overwrites the file resetting it back to zero.

"Emby sets the watched flag in the .nfo files for the episodes, which Sonarr then overwrites later on, setting it back to 0, which Emby takes as if Kodi has set the watched status to 0, and updates itself showing that episode as unwatched"

To be confirmed.

### LINKED ISSUE #2398: nfo files deleted instead of recycled
### Bug Report & Feature Request

On upgrades old media files are moved to the recycle bin, but the metadata files are not. This is particularly problematic with .nfo since sonarr considers those kodi metadata files instead of scene .nfos since those are delete and thus the user cannot restore the original from the recycle bin.

So bug:
1. Sonarr should recycle all files associated with an episode, not just metadata files.

Feature:

2. It should scan the .nfo file for xml to determine whether it's kodi metadata or scene extrafile.
3. Sonarr should only rename .nfo to .nfo-orig if the kodi metadata provider wants to write the xml .nfo file.

This would allow Sonarr to treat the nfo properly. It's still desirable to recycle them anyway, but it'll make the system a bit more resilient if the user enables kodi metadata on an existing library with scene .nfo.

Taloth:
#### Database Migration
NO

#### Description
- Recycle Metadata files on episode removal, instead of permanently deleting them.
- Import nfo 'other' files as 'nfo' instead of 'nfo-orig'.
- On import existing the Kodi metadata provider will check if the .nfo is xml, and only mark if as EpisodeMetadata if that's the case. (leaving it to OtherFiles to pick up the scene nfo)
- On Kodi metadata generation, the existing .nfo OtherFile will be renamed to .nfo-orig.

- Bonus: Kodi metadata generation refresh will preserve 'watched' flag if possible, but not on episode upgrades.

The end result is that if kodi metadata is disabled, scene .nfo files will be treated as scene nfo and kept intact.

* closes #2398
* closes #932

#### Related but not solved:
* ref #991

### DIFF
--- a/src/NzbDrone.Core.Test/Extras/Metadata/Consumers/Roksbox/FindMetadataFileFixture.cs
+++ b/src/NzbDrone.Core.Test/Extras/Metadata/Consumers/Roksbox/FindMetadataFileFixture.cs
@@ -8,7 +8,7 @@ using NzbDrone.Core.Test.Framework;
 using NzbDrone.Core.Tv;
 using NzbDrone.Test.Common;
 
-namespace NzbDrone.Core.Test.Metadata.Consumers.Roksbox
+namespace NzbDrone.Core.Test.Extras.Metadata.Consumers.Roksbox
 {
     [TestFixture]
     public class FindMetadataFileFixture : CoreTest<RoksboxMetadata>
--- a/src/NzbDrone.Core.Test/Extras/Metadata/Consumers/Wdtv/FindMetadataFileFixture.cs
+++ b/src/NzbDrone.Core.Test/Extras/Metadata/Consumers/Wdtv/FindMetadataFileFixture.cs
@@ -8,7 +8,7 @@ using NzbDrone.Core.Test.Framework;
 using NzbDrone.Core.Tv;
 using NzbDrone.Test.Common;
 
-namespace NzbDrone.Core.Test.Metadata.Consumers.Wdtv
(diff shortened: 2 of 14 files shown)


Thirrian:
> Bonus: Kodi metadata generation refresh will preserve 'watched' flag if possible, but not on episode upgrades.

Pretty please configurable? :relaxed: 
I'd like to keep watched status even when quality upgrades happen...

Taloth:
Thirrian, it'snot that I don't want to. But preserving between upgrades is far more difficult, coz the old episode and metadata gets deleted before the upgrade is installed and thus would require a mechanism that deal with that.

Dnkhatri:
I think syncing with trakt would be the easiest option and for sonarr to keep that information in its database.  So when an episode is updated read status can be checked from trakt. Though everyone uses a different service to track watch list etc so adding 1 would result in people asking for many more.

Thirrian:
Ah no problem man :smile:
I've been wanting to get rid of trakt for a while now and I thought I saw a chance :grin:

Taloth:
that was the idea yes

PR #2440, opened 2018-03-03
changed: which event handler triggers movie search

ThiRaBrTNK:
Searching for a movie by name copy-pasted from somewhere looks like common scenario, and search is not triggered when text got pasted into input.
'Input' event works better then 'keyup' as it got triggered upon _any_ change of input's value (keypress, paste, text drag-n-drop).


Taloth:
@markus101 what do you think about this PR?

markus101:
Looks good in my tests. Thanks!

PR #2450, opened 2018-03-09
Fixed: sorting for series "A.P. Bio"

Thirrian:
https://forums.sonarr.tv/t/a-p-bio-sorting-under-p/17959

- Added sorting for "A.P. Bio"
- Added unit test
- Removed name normalization for series which was deleted on thetvdb


Taloth:
Can you add em to the unit tests too?

Thirrian:
I've added the test, and found a series that no longer exists so I removed it. Technically that is a separate issue, but given the size of the change I hope it's ok to include it here as well.

Taloth:
that's fine.

nova-techies:
Dunno if this is the right place for this but the forums led me here.

I just updated to 2.0.0.5163 and I'm still having A.P. Bio sorting under 'P' and not 'A'.  Did I miss a step here?

PR #2462, opened 2018-03-17
case-sensitive OS fix

delphiactual:
#### Database Migration
NO

#### Description
Filename normalization

* Allows correct compilation on OS's that use case-specific filenames


markus101:
Thanks!

Issue #755, opened 2015-08-22
Season Search for daily shows


markus101:
Doesn't appear to work and may be searching for the actual season number, which would be obviously wrong. Need to determine why its not working and see what we can do to fix it.

We may be able to search for a full season on newznab using partial episode matches, but mileage may vary.

thezoggy:
yeah, 'daily' shows are searching by season using the wrong data.. tries to do season # instead of the YYYY. 

auto/manual search works:
`https://api.dognzb.cr/api?t=tvsearch&cat=5030,5040&extended=1&apikey=<removed>&offset=0&limit=100&rid=5714&season=2014&ep=09/25`

season search doesnt:
`https://api.dognzb.cr/api?t=tvsearch&cat=5030,5040&extended=1&apikey=<removed>&offset=200&limit=100&rid=5714&season=19`

now there is the issue when a season spans across multiple years.. which year do you use/search for each year?

its been awhile since ive used sb with daily shows.. so im just going off reading the source code / memory.. will have to spin up an instance to confirm.. 
when sb does its season search, it does `&season=2010&q=2010.03` so you get stuff for that 'season' but limit the scope to the month.. thus it can fit within the 100 limit result (x limited pages needed to go through).

PR #2463, opened 2018-03-18
Daily season search

Taloth:
#### Database Migration
**YES** 120 - Replace/add db indexes

#### Description
A season search for a daily show used to simply do a 'standard' season search, which only works if the show uses years as seasons.
This change adds a DailySeasonSearchCriteria that groups by the year in the airdate and performs searches for those.

Possible side-effect I just realised is that Jackett might not handle the format.

This PR also includes half a dozen performance improvements:

Test method: With no download client I queried 1 daily show season that caused 1416 releases to be added to the pending queue.Subsequently all indexers were disabled, which causes the RssSync to only evaluate the pending releases.

The following measurements of cumulative improvements. The biggest improvement by far is sqlite initialization env vars.

Original: 47.5 sec total
Sqlite initialisation: 19.9 sec total
Command Progress debounce: 17.1 sec total
PendingRelease refactoring: 13.9 sec total
Database indexes: 12.2 sec total
EventAggregator caching: 10.1 sec total

I had some other more agressive caching changes, but they only saved around 1.2 sec outside of the profiler in release mode.
A bigger improvement was possible using a db transaction to keep the connection open... iirc around 3-4 sec improvment, but it led to other unstable to behavior and simply too risky.

* https://github.com/Sonarr/Sonarr/issues/755

### DIFF
--- a/src/NzbDrone.Common/Extensions/IEnumerableExtensions.cs
+++ b/src/NzbDrone.Common/Extensions/IEnumerableExtensions.cs
@@ -51,6 +51,34 @@ namespace NzbDrone.Common.Extensions
             }
         }
 
+        public static Dictionary<TKey, TItem> ToDictionaryIgnoreDuplicates<TItem, TKey>(this IEnumerable<TItem> src, Func<TItem, TKey> keySelector)
+        {
+            var result = new Dictionary<TKey, TItem>();
+            foreach (var item in src)
+            {
+                var key = keySelector(item);
+                if (!result.ContainsKey(key))
+                {
+                    result[key] = item;
+                }
+            }
+            return result;
+        }
+
+        public static Dictionary<TKey, TValue> ToDictionaryIgnoreDuplicates<TItem, TKey, TValue>(this IEnumerable<TItem> src, Func<TItem, TKey> keySelector, Func<TItem, TValue> valueSelector)
+        {
+            var result = new Dictionary<TKey, TValue>();
+            foreach (var item in src)
+            {
+                var key = keySelector(item);
+                if (!result.ContainsKey(key))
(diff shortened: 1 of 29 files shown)


src/NzbDrone.Common/Extensions/IEnumerableExtensions.cs
```diff
@@ -51,6 +51,30 @@ public static class EnumerableExtensions
             }
         }
 
+        public static Dictionary<TKey, TItem> ToDictionaryIgnoreDuplicates<TItem, TKey>(this IEnumerable<TItem> src, Func<TItem, TKey> keySelector)
+        {
+            var result = new Dictionary<TKey, TItem>();
+            foreach (var item in src)
+            {
+                var key = keySelector(item);
+                if (!result.ContainsKey(key))
```
markus101: Braces for multi-line `if`

src/NzbDrone.Common/Extensions/IEnumerableExtensions.cs
```diff
@@ -51,6 +51,30 @@ public static class EnumerableExtensions
...
+            return result;
+        }
+
+        public static Dictionary<TKey, TValue> ToDictionaryIgnoreDuplicates<TItem, TKey, TValue>(this IEnumerable<TItem> src, Func<TItem, TKey> keySelector, Func<TItem, TValue> valueSelector)
+        {
+            var result = new Dictionary<TKey, TValue>();
+            foreach (var item in src)
+            {
+                var key = keySelector(item);
+                if (!result.ContainsKey(key))
```
markus101: Braces for multi-line `if`

PR #2538, opened 2018-04-30
Added support for nested settings models so settings

Taloth:
I'm not entirely sure that this is the right way to go. The way it works make sense, but it's quite a bit of complexity. I can't think of an alternative yet though.


Taloth:
Merged in

PR #2548, opened 2018-05-05
Add Support for PushSafer Connections / Notifications 

### LINKED ISSUE #2471: Add Support for PushSafer Connections / Notifications
### What problem are you looking to solve?
Receive notifications via PushSafer, as you would with PushBullet or PushOver, when something is downloaded, snatched, upgraded, etc. 

### Other Information
PushSafer is very similar to PushOver with a few small differences

API: https://www.pushsafer.com/en/pushapi

BoKKeR:
#### Database Migration
Not sure

#### Description
Added Support for PushSafer Connections / Notifications. On top of that custom sound, device and icons are supported

#### Todos
- [ ] Tests
No unit tests have been made as the rest of the notifications have none. 
- [ ] Documentation
No documentation beyond the help in the connections tab. I cant find any extra documentation in the solution.

fixes #2471 
*


src/NzbDrone.Core/Notifications/Pushsafer/PushsaferSettings.cs
```diff
@@ -0,0 +1,39 @@
...
+
+    public class PushsaferSettings : IProviderConfig
+    {
+        private static readonly PushsaferSettingsValidator Validator = new PushsaferSettingsValidator();
+
+        [FieldDefinition(0, Label = "API Key", HelpLink = "https://www.pushsafer.com/dashboard")]
+        public string ApiKey { get; set; }
+
+        [FieldDefinition(1, Label = "Sound", Type = FieldType.Textbox, HelpText = "Notification sound, leave blank to use the default, set 0 for silent, set 1-50 for custom", HelpLink = "https://www.pushsafer.com/en/pushapi")]
+        public string Sound { get; set; }
```
Taloth: Sounds like it's supposed to be a nullable integer not a string. Same for Icon.

Given that they have special meaning, it might even be better to do a FieldType.Select

src/NzbDrone.Core/Notifications/Pushsafer/PushsaferSettings.cs
```diff
@@ -0,0 +1,39 @@
...
+
+        [FieldDefinition(2, Label = "Icon", Type = FieldType.Textbox, HelpText = "Notification icon, leave blank to use the default, set 1-176 for custom", HelpLink = "https://www.pushsafer.com/en/pushapi")]
+        public string Icon { get; set; }
+
+        [FieldDefinition(3, Label = "Device", Type = FieldType.Textbox, HelpText = "Leave blank to notify all devices, set Single Device ID or Device Group ID for custom devices", HelpLink = "https://www.pushsafer.com/en/pushapi")]
+        public string Device { get; set; }
+
+        public bool IsValid => !string.IsNullOrWhiteSpace(ApiKey);
```
Taloth: What is this for?

src/NzbDrone.Core/Notifications/Pushsafer/PushsaferSettings.cs
```diff
@@ -0,0 +1,39 @@
...
+        private static readonly PushsaferSettingsValidator Validator = new PushsaferSettingsValidator();
+
+        [FieldDefinition(0, Label = "API Key", HelpLink = "https://www.pushsafer.com/dashboard")]
+        public string ApiKey { get; set; }
+
+        [FieldDefinition(1, Label = "Sound", Type = FieldType.Textbox, HelpText = "Notification sound, leave blank to use the default, set 0 for silent, set 1-50 for custom", HelpLink = "https://www.pushsafer.com/en/pushapi")]
+        public string Sound { get; set; }
+
+        [FieldDefinition(2, Label = "Icon", Type = FieldType.Textbox, HelpText = "Notification icon, leave blank to use the default, set 1-176 for custom", HelpLink = "https://www.pushsafer.com/en/pushapi")]
+        public string Icon { get; set; }
```
Taloth: Reading the description, what does 0 mean?

src/NzbDrone.Core/Notifications/Pushsafer/PushsaferService.cs
```diff
@@ -0,0 +1,65 @@
...
+
+        public void SendNotification(string title, string message, PushsaferSettings settings)
+        {
+            var client = RestClientFactory.BuildClient(URL);
+            var request = new RestRequest(Method.POST);
+            request.AddParameter("k", settings.ApiKey);
+            request.AddParameter("t", title);
+            request.AddParameter("m", message);
+
+            if (!settings.Sound.IsNullOrWhiteSpace())
```
Taloth: `.IsNotNullOrWhiteSpace()` eliminates the negation. (twice)

src/NzbDrone.Core/Notifications/Pushsafer/PushsaferSettings.cs
BoKKeR: I will change it to Select

src/NzbDrone.Core/Notifications/Pushsafer/PushsaferSettings.cs
BoKKeR: The API allows small groups of devices to be individually notified, Or all devices can be notified.

src/NzbDrone.Core/Notifications/Pushsafer/PushsaferSettings.cs
BoKKeR: 0 makes have no effect The used icon will still be the default one (1)

src/NzbDrone.Core/Notifications/Pushsafer/PushsaferService.cs
BoKKeR: Will replace. thanks

src/NzbDrone.Core/Notifications/Pushsafer/PushsaferSettings.cs
Taloth: I meant the IsValid property, not the Device.

BoKKeR: Nothing actually I am removing it. :)

src/NzbDrone.Core/Notifications/Pushsafer/PushsaferService.cs
BoKKeR: This is the same approach that is used for a different notification: Pushover. I dont know if its worth changing there too or not.

src/NzbDrone.Core/Notifications/Pushsafer/PushsaferSettings.cs
BoKKeR: Should I list out all the icon options and sound names using Filetype.Select? Its tedious and the API owner can always change them or just add new ones thus we will have to keep it updated all the time. But I can do it if that is the best solution for the UI.

src/NzbDrone.Core/Notifications/Pushsafer/PushsaferService.cs
markus101: IIRC `IsNotNullOrWhiteSpace` was added later, would be good to fix those up as well as a separate commit.

src/NzbDrone.Core/Notifications/Pushsafer/PushsaferSettings.cs
markus101: I don't think a select is the best idea unless we can programatically get those and that's not something we currently do anywhere else. Personally I'd be okay with a simple text entry as long as we can point people at an official list somewhere.

src/NzbDrone.Core/Notifications/Pushsafer/InvalidResponseException.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System;
+
+namespace NzbDrone.Core.Notifications.Pushsafer
+{
+    public class InvalidResponseException : Exception
```
markus101: Not seeing this used anywhere.

src/NzbDrone.Core/Notifications/Pushsafer/PushsaferSettings.cs
```diff
@@ -0,0 +1,39 @@
...
+        {
+            RuleFor(c => c.ApiKey).NotEmpty();
+        }
+    }
+
+    public class PushsaferSettings : IProviderConfig
+    {
+        private static readonly PushsaferSettingsValidator Validator = new PushsaferSettingsValidator();
+
+        [FieldDefinition(0, Label = "API Key", HelpLink = "https://www.pushsafer.com/dashboard")]
```
markus101: `HelpText` to indicate they're looking for the `Your private key`?

src/NzbDrone.Core/Notifications/Pushsafer/PushsaferSettings.cs
BoKKeR: The list is in the HelpText. So I will leave this unchanged

src/NzbDrone.Core/Notifications/Pushsafer/InvalidResponseException.cs
BoKKeR: It can be safely removed as it's not used. PushSaferService actually reports to the logger every exception. That should be sufficient.

markus101:
The revert wiped out all the changes here.

Issue #2570, opened 2018-05-16
Preserve custom tags/suffixes when importing subtitles


jusher00:
### What problem are you looking to solve?
I have a lot of movie folders containing a structure like this:

`movie-2018.de.forced.idx`
`movie-2018.de.forced.sub`
`movie-2018.de.idx`
`movie-2018.de.sub`
`movie-2018.en.forced.idx`
`movie-2018.en.forced.sub`
`movie-2018.en.idx`
`movie-2018.en.sub`
`movie-2018.mkv`

[Plex supports this naming scheme](https://support.plex.tv/articles/200471133-adding-local-subtitles-to-your-media/), however Sonarr fails to import this correctly unfortunately.

When I import this, the subtitles flagged as `forced` neither keep that forced flag, nor is the language parsed.

It would be great if Sonarr was able to keep that `forced` flag and recognize the language as well.

markus101:
This should also include things like unknown language codes (`tc` (Traditional Chinese) or `sc` (Simplified Chinese)) or `sdh` (Subtitles for the Deaf or Hard-of-Hearing) or a specific audio track (director, etc).

bakerboy448:
Duplicate of #2639

wsegatto:
Let me add an important comment – at least for me and all Brazilians. 

For subtitles that contain the country attribute like pt-BR (Brazilian Portuguese) it is being removed when Sonarr renames the file.

Before:

The.Walking.Dead.S01E01.XXXXXXXX.mkv
The.Walking.Dead.S01E01.XXXXXXXX.pt-BR.srt
After:

The.Walking.Dead.S01E01.mkv
The.Walking.Dead.S01E01.srt

It should be renamed to The.Walking.Dead.S01E01.pt-BR.srt

For cases that don't contain the country attribute, it works fine.

The one that sets our subtitles to "pt-BR" is Bazarr, so it makes sense that Sonarr should maintain - and not lose - this critical information, specially when maintaining multiple language subtitles.

bakerboy448:
https://typecraft.org/tc2wiki/Brazilian_Portuguese_(BP)

Probably should add support and get all ISO languages covered as a stop gap

https://github.com/Sonarr/Sonarr/blob/0a2b109a3fe101e260b623d0768240ef8b7a47ae/src/NzbDrone.Core/Parser/IsoLanguages.cs

wsegatto:
Any updates?

gdmachado:
[...]

PR #2572, opened 2018-05-16
Preserve 'forced' tag when importing subtitles

jusher00:
#### Database Migration
NO

#### Description
Adds a forced property to `SubtitleService.cs`.

This is the first time with the C family for me, so not sure if I overlooked anything.

#### Todos
- [ ] Tests
- [ ] Documentation

#2570

### DIFF
--- a/src/NzbDrone.Core/Extras/Subtitles/SubtitleService.cs
+++ b/src/NzbDrone.Core/Extras/Subtitles/SubtitleService.cs
@@ -87,7 +87,8 @@ namespace NzbDrone.Core.Extras.Subtitles
             if (SubtitleFileExtensions.Extensions.Contains(Path.GetExtension(path)))
             {
                 var language = LanguageParser.ParseSubtitleLanguage(path);
-                var suffix = GetSuffix(language, 1, false);
+                var forced = LanguageParser.ParseForcedFlag(path);
+                var suffix = GetSuffix(language, forced, 1, false);
                 var subtitleFile = ImportFile(series, episodeFile, path, readOnly, extension, suffix);
                 subtitleFile.Language = language;
 
@@ -99,7 +100,7 @@ namespace NzbDrone.Core.Extras.Subtitles
             return null;
         }
 
-        private string GetSuffix(Language language, int copy, bool multipleCopies = false)
+        private string GetSuffix(Language language, bool forced, int copy, bool multipleCopies = false)
         {
             var suffixBuilder = new StringBuilder();
 
@@ -113,6 +114,12 @@ namespace NzbDrone.Core.Extras.Subtitles
             {
                 suffixBuilder.Append(".");
                 suffixBuilder.Append(IsoLanguages.Get(language).TwoLetterCode);
+            }
+
+            if (forced)
+            {
+                suffixBuilder.Append(".");
+                suffixBuilder.Append("forced");
             }
 
             return suffixBuilder.ToString();
--- a/src/NzbDrone.Core/Parser/LanguageParser.cs
+++ b/src/NzbDrone.Core/Parser/LanguageParser.cs
@@ -18,7 +18,9 @@ namespace NzbDrone.Core.Parser
                                                                 RegexOptions.Compiled);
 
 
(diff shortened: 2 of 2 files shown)


src/NzbDrone.Core/Extras/Subtitles/SubtitleService.cs
```diff
@@ -87,7 +87,8 @@ public override ExtraFile Import(Series series, EpisodeFile episodeFile, string
             if (SubtitleFileExtensions.Extensions.Contains(Path.GetExtension(path)))
             {
                 var language = LanguageParser.ParseSubtitleLanguage(path);
-                var suffix = GetSuffix(language, 1, false);
+                var forced = LanguageParser.ParseForcedFlag(path);
```
markus101: Parsing `forced` here makes sense, it's specific to subtitles, not general language parsing.

Tests for it should also be added.

src/NzbDrone.Core/Parser/LanguageParser.cs
```diff
@@ -18,7 +18,9 @@ public static class LanguageParser
                                                                 RegexOptions.Compiled);
 
 
-        private static readonly Regex SubtitleLanguageRegex = new Regex(".+?[-_. ](?<iso_code>[a-z]{2,3})$", RegexOptions.Compiled | RegexOptions.IgnoreCase);
+        private static readonly Regex SubtitleLanguageRegex = new Regex(".+?[-_. ](?<iso_code>[a-z]{2,3})([-_. ]forced)?$", RegexOptions.Compiled | RegexOptions.IgnoreCase);
```
markus101: Might be sufficient to just check for `(?:[-_. ]|$)`. Feels a bit weird to explicitly check for forced while not actually doing anything with it. Might be better to return the language and whether it's forced or not.

Tests should be added for this as well.

src/NzbDrone.Core/Parser/LanguageParser.cs
```diff
@@ -129,6 +131,23 @@ public static Language ParseSubtitleLanguage(string fileName)
...
+                var simpleFilename = Path.GetFileNameWithoutExtension(fileName);
+                var forcedFlagMatch = SubtitleForcedRegex.Match(simpleFilename);
+
+                return forcedFlagMatch.Success;
+            }
+            catch (Exception ex)
+            {
+                Logger.Debug(ex, "Failed parsing forced flag from subtitle file: {0}", fileName);
+            }
+        }
```
markus101: Needs a fallback return value, otherwise it won't compile.

src/NzbDrone.Core/Parser/LanguageParser.cs
jusher00: The current regex in `develop` only matches the language code at the end of the filename. I think that is a good method to avoid unwanted matches with movies like [*La Vie **en** rose*](https://www.imdb.com/title/tt0450188/). So with the `forced` in the regex, we make sure that the only thing that is allowed after a language code in the filename is the `forced` flag. Would you agree? Making it a non-capturing group, however, does makes sense of course.

As I'm fairly new to programming, I don't really know where to start with the tests for this. Could you point me to a file where I can add test cases?

Taloth:
I think the real question is: why two regexps?

Just have one, capture both and parse both into a ParsedSubtitleInfo.
Of course it wouldn't really fit in LanguageParser anymore, but that's probably a good thing.

`".+?[-_. ](?<iso_code>[a-z]{2,3})(?:[-_. ](?<flags>forced|sdh))?$"`

Also check other common suffixes. Yes, .forced is supported for plex, but what about others? like sdh and commentary.

As for the tests: https://github.com/Sonarr/Sonarr/blob/develop/src/NzbDrone.Core.Test/ParserTests/LanguageParserFixture.cs#L58-L67

Taloth:
The regex to do both isn't difficult, but would we want to? does plex handle both scenarios?

Add the Flags/Suffix to `SubtitleFile`. Then add a migration in .Core/Datastore/Migrations. It's a simple oneliner to add a column. Checkout migration 112 for an example. You'll probably be creating migration 122, we mark prs with a tag if they have migrations so check if there are existing ones with 122.
When fiddling with migrations be sure to make a backup of your database first.

I think the flag should be an enum too, but i'll let @markus101 to weigh in on that.

PR #2573, opened 2018-05-18
Improve code readability

Thirrian:
#### Database Migration
NO

#### Description
Replace double negation !string.IsNullOrWhiteSpace() with string.IsNotNullOrWhiteSpace()
Idea here: https://github.com/Sonarr/Sonarr/pull/2548#discussion_r189136824

#### ~~Todos~~
~~- [ ] Tests~~
~~- [ ] Documentation~~

### DIFF
--- a/src/Microsoft.AspNet.SignalR.Owin/Infrastructure/OwinEnvironmentExtensions.cs
+++ b/src/Microsoft.AspNet.SignalR.Owin/Infrastructure/OwinEnvironmentExtensions.cs
@@ -59,7 +59,7 @@ namespace Microsoft.AspNet.SignalR.Owin
             if (environment.TryGetValue(OwinConstants.HostAppModeKey, out value))
             {
                 var stringVal = value as string;
-                return !String.IsNullOrWhiteSpace(stringVal) &&
+                return String.IsNotNullOrWhiteSpace(stringVal) &&
                        OwinConstants.AppModeDevelopment.Equals(stringVal, StringComparison.OrdinalIgnoreCase);
             }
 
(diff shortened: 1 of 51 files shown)


src/Microsoft.AspNet.SignalR.Owin/Infrastructure/OwinEnvironmentExtensions.cs
```diff
@@ -59,7 +59,7 @@ internal static bool GetIsDebugEnabled(this IDictionary<string, object> environm
             if (environment.TryGetValue(OwinConstants.HostAppModeKey, out value))
             {
                 var stringVal = value as string;
-                return !String.IsNullOrWhiteSpace(stringVal) &&
+                return String.IsNotNullOrWhiteSpace(stringVal) &&
```
markus101: This isn't a method on `String`. Though we shouldn't make changes to this project.

src/NzbDrone.Api/Config/DownloadClientConfigModule.cs
```diff
@@ -18,7 +18,7 @@ public class DownloadClientConfigModule : NzbDroneConfigModule<DownloadClientCon
                            .SetValidator(rootFolderValidator)
                            .SetValidator(mappedNetworkDriveValidator)
                            .SetValidator(pathExistsValidator)
-                           .When(c => !string.IsNullOrWhiteSpace(c.DownloadedEpisodesFolder));
+                           .When(c => string.IsNotNullOrWhiteSpace(c.DownloadedEpisodesFolder));
```
markus101: Same here, not a method on `string`, should be converted to `c.DownloadedEpisodesFolder.IsNotNullOrWhiteSpace()`

Issue #2628, opened 2018-06-19
Sonarr Importing To Wrong Episode


IsamuDragon:
Have only seen this with Anime and when importing manually, but when importing a series with an alternative name (as based with XEM) it will import it to the wrong directory.

For example:
`[HorribleSubs] Shokugeki no Soma S3 - 23 [720p].mkv`
which through XEM should import it to S03E23, however it imported it to S01E23, and I had redownload both the original S01E23 and S03E23 files.

Before trying to re-import S01E23, I copied the redownloaded S03E23 and tried manual import again.(Thinking I had accidentally not noticed the S01 in the import screen) It stated S03 in the import screen and still imported to S01.

Edit: Running Win10PRO x64, Sonarr Version 2.0.0.5225 - 18 Jun 2018 
[sonarr.0.txt](https://github.com/Sonarr/Sonarr/files/2117892/sonarr.0.txt)

![Proof that it Imported to S01](https://user-images.githubusercontent.com/464964/41613999-742233f6-73b4-11e8-8beb-849f6c7551c2.png)

markus101:
Please include the information requested when filing a bug report, including version, OS, debug logs (and everything else relevant).

IsamuDragon:
Updated post

markus101:
Related PR has been merged.

PR #2630, opened 2018-06-20
Manual Import regression and improve handling of folders

Taloth:
Mainly fixing the Augmentation erroneously overriding the user selected values.

But I also addressed how it handles folders:
If the root folder doesn't contain a parse-able series, it used to simply fetch all files recursively and process them individually, preventing it from using folder information.
Instead it now processes direct files as individual files, and recurses into folders.

This allows you to point Manual Import at a `myFolder` like `myFolder/Series.Title.S01/1x02.Episode.1080p.mkv` and still have it process properly.


Taloth:
Btw. References:
- https://github.com/Sonarr/Sonarr/issues/2628
- https://www.reddit.com/r/sonarr/comments/8sdt68/2005225_ignores_manual_import_settings/

PR #2671, opened 2018-07-21
Add YggTorrent as Indexer

graniv:
#### Database Migration
NO

#### Description
Adding a YggTorrent as new indexer in Sonarr. This indexer is very popular in FR and CA.

#### Todos
- [x] Tests
- [x] Documentation


markus101:
We do not support indexers that do not have a proper API in Sonarr. You'll need to use Jackett for that.

https://github.com/Jackett/Jackett/issues/2852

PR #2728, opened 2018-09-10
Add compatibility w/ Hombrew-installed mono

jeffbyrnes:
#### Database Migration
NO

#### Description
By default, Homebrew installs mono to `/usr/local/bin`. However, when starting Sonarr, it does not appear to have this path as part of its `$PATH`, which results in Sonarr prompting the user to, incorrectly, install Mono via the “official” installer, even though it’s already installed.

Adding this check ensures that the right `$PATH` additions are made for a Homebrew-installed copy of mono.

#### Todos
- [x] Tests
- [x] Documentation

* None

### DIFF
--- a/osx/Sonarr
+++ b/osx/Sonarr
@@ -9,7 +9,11 @@ APPNAME="Sonarr"
  
 #set up environment
 if [[ -x '/opt/local/bin/mono' ]]; then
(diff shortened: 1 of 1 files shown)


markus101:
Thanks!

PR #2754, opened 2018-10-03
the date and time now show properly when hovering over the relative d…

geogolem:
format date/time in seriesOverviewIndex

#### Database Migration
NO

#### Description
Display a better formated date/time when hovering over the relative nextAirDate in SeriesIndexOverview

I'm not sure if this is the best fix.. but I just created the PR so that we could discuss...

Im not familiar enough with the code/new GUI to really know what im doing, but that date was annoying me...

### DIFF
--- a/frontend/src/Series/Index/Overview/SeriesIndexOverviewInfo.js
+++ b/frontend/src/Series/Index/Overview/SeriesIndexOverviewInfo.js
@@ -1,5 +1,6 @@
 import PropTypes from 'prop-types';
 import React from 'react';
+import formatDateTime from 'Utilities/Date/formatDateTime';
 import getRelativeDate from 'Utilities/Date/getRelativeDate';
 import formatBytes from 'Utilities/Number/formatBytes';
 import { icons } from 'Helpers/Props';
@@ -182,6 +183,7 @@ function SeriesIndexOverviewInfo(props) {
     nextAiring,
     showRelativeDates,
     shortDateFormat,
+	longDateFormat,
     timeFormat
   } = props;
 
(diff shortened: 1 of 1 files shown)


frontend/src/Series/Index/Overview/SeriesIndexOverviewInfo.js
```diff
@@ -254,6 +256,7 @@ SeriesIndexOverviewInfo.propTypes = {
   sizeOnDisk: PropTypes.number,
   sortKey: PropTypes.string.isRequired,
   showRelativeDates: PropTypes.bool.isRequired,
+  longDateFormat: PropTypes.string.isRequired,
```
geogolem: oops, i dont think this line is required.. this line needs to be removed... i think..

frontend/src/Series/Index/Overview/SeriesIndexOverviewInfo.js
```diff
@@ -182,6 +183,7 @@ function SeriesIndexOverviewInfo(props) {
     nextAiring,
     showRelativeDates,
     shortDateFormat,
+	longDateFormat,
```
geogolem: this line should be removed too, i think....

frontend/src/Series/Index/Overview/SeriesIndexOverviewInfo.js
markus101: Yeah, it's not being used, or passed in.

frontend/src/Series/Index/Overview/SeriesIndexOverviewInfo.js
markus101: Yes, it will fail the linting.

markus101:
I'll pull in the fix and fit up some of the other dates for the next release.

PR #2846, opened 2018-12-09
Removed .idea folder

Nyholm:
#### Database Migration
NO

#### Description
Removed `.idea` folder. Folders like that should be in a global .gitignore. The reason why this folder should be removed is because I don't want my IDE settings to be defined for me.


thezoggy:
but if you want to contribute to sonarr you should use a common set of settings (ex: code style) which is why you should include it.

Nyholm:
No, you don't need a common set of settings. If that was true nobody that is using non-jetbrains IDE's could not contribute. 

Sure you need coding style. But that is enforced with automatic PR reviews.

timdorr:
@thezoggy There is already [an .editorconfig file](https://github.com/Sonarr/Sonarr/blob/develop/.editorconfig) for that.

thezoggy:
@timdorr https://intellij-support.jetbrains.com/hc/en-us/articles/206544839

timdorr:
@thezoggy I'm not sure why you linked to that article. It doesn't say anything relevant to this discussion. 

For Resharper users, there is [an official plugin for the EditorConfig format](https://plugins.jetbrains.com/plugin/7294-editorconfig). But like @Nyholm said, non-JetBrains users won't see any of the configuration in the .idea folder, so having it doesn't necessarily enforce anything.

Taloth:
Checking in part of the .idea folder is recommended to sync certain styling rules between developers. .editorconfig most certainly didn't cover those rules back then and might not even today. Considering the core devs were all using JetBrains we wouldn't remove it at the request of just anybody checking out our repository. 'were' being the operative word, since we've switched away from JetBrains' IDE.
So for the sake of avoiding a length discussion, yes we'll remove the .idea folder, but we'll do that in v3 (phantom-develop).

I hope that satisfies your inquiry, and if you plan to contribute to Sonarr please know that most development takes place on v3 and should be the focus of any further improvements.

Nyholm:
Thank you. This small PR was meant for a quick merge since it is a no-brainier. 

I understand your reasoning and I'm happy that you have a plan to fix this. 
:+1:

PR #2852, opened 2018-12-18
Implemented Discord specific notification

samip5:
#### Database Migration
NO

#### Description
I used the existing code from Slack integration.
Now it uses, the username from Discord web-hook settings if it's not set on the Sonarr notification settings otherwise it uses the one from Notification settings.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,5703 @@
+{
+  "name": "Sonarr",
+  "version": "2.0.0",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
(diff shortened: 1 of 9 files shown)


markus101:
Sorry, this slipped through the cracks around the holidays and into the new year. I've taken a look, but a direct copy of slack isn't what we're looking for, but instead the intention was to use the implement it specifically for Discord. Given the source repo has been removed I've assumed this is abandoned and I've built out notifications using Discord's webhook API directly instead of the slack proxy.

Thanks for the PR and again I'm sorry this was forgotten about.

Issue #2898, opened 2019-01-19
Support series name & episode number in square brackets


whtsky:
**Describe the problem**

A lot of Chinese anime releases love to put series name & episode number in square brackets, like
```
【漫猫&爱恋字幕组】[五等分的新娘/​五等分的花嫁/​五等分の花嫁][Go-​Toubun_​no_​Hanayome][01][GB][1080P][HEVC]
【咕咕茶字幕组】★1月新番[天使降临到了我身边! / Watashi ni Tenshi ga Maiorita!][02][1080P][简体][MP4]
【動漫國字幕組】★01月新番[五等分的新娘/五等分的花嫁][02][1080P][繁體][MP4]
【DHR百合組】[天使降臨到我身邊！_Watashi ni Tenshi ga Maiorita!][02][繁體][720P][MP4]
[爱恋字幕社][盾之勇者成名录][Tate no Yuusha no Nariagari][02][720P][GB][MP4]
```

**Describe any solutions you think might work**

Support parsing series name & episode numer in square brackets.

**Additional context**

N/A

whtsky:
Thanks for the quick reply!

Looks like it’s failing to parse(title is `[YMDR][家有女友][Domestic na Kanojo][2019][02][1080p][AVC][JAP][BIG5][MP4-AAC][繁中]`)
```
19-1-20 02:50:03.1|Trace|DownloadDecisionMaker|Processing release 12/19
19-1-20 02:50:03.1|Debug|DownloadDecisionMaker|Processing release '[YMDR][][Domestic na Kanojo][2019][01][1080p][AVC][JAP][BIG5][MP4-AAC][]' from 'nyaasi - anime'
19-1-20 02:50:03.1|Debug|Parser|Parsing string '[YMDR][][Domestic na Kanojo][2019][01][1080p][AVC][JAP][BIG5][MP4-AAC][]'
19-1-20 02:50:03.1|Debug|Parser|Unable to parse [YMDR][][Domestic na Kanojo][2019][01][1080p][AVC][JAP][BIG5][MP4-AAC][]
19-1-20 02:50:03.1|Debug|Parser|Parsing string '[YMDR][][Domestic na Kanojo][2019][01][1080p][AVC][JAP][BIG5][MP4-AAC][]'
19-1-20 02:50:03.1|Debug|Parser|Unable to parse [YMDR][][Domestic na Kanojo][2019][01][1080p][AVC][JAP][BIG5][MP4-AAC][]
[...]

PR #2904, opened 2019-01-22
WIP: Support black lenticular bracketed or square brackted group, title and episode number in parser

whtsky:
#### Database Migration
NO

#### Description
Added a new regex for parsing black lenticular bracketed or square brackted group, title and episode number.
All unit test in `AbsoluteEpisodeNumberParserFixture` has passed.

#### Todos
- [x] Tests
- [ ] Documentation

* #2898

### DIFF
--- a/src/NzbDrone.Core.Test/ParserTests/AbsoluteEpisodeNumberParserFixture.cs
+++ b/src/NzbDrone.Core.Test/ParserTests/AbsoluteEpisodeNumberParserFixture.cs
@@ -94,6 +94,14 @@ namespace NzbDrone.Core.Test.ParserTests
         [TestCase("Tokyo Ghoul: RE S2 - Episode 4 VOSTFR (1080p)", "Tokyo Ghoul RE S2", 4, 0, 0)]
         [TestCase("To Aru Majutsu no Index III - Episode 5 VOSTFR (1080p)", "To Aru Majutsu no Index III", 5, 0, 0)]
         [TestCase("[Prout] Steins;Gate 0 - Episode 5 VOSTFR (BDRip 1920x1080 x264 FLAC)", "Steins;Gate 0", 5, 0, 0)]
+        [TestCase("【喵萌奶茶屋】★10月新番★[哥布林杀手/Goblin Slayer][12END][720p][繁体][招募翻译校对]", "Goblin Slayer", 12, 0, 0)]
+        [TestCase("[爱恋字幕社][盾之勇者成名录][Tate no Yuusha no Nariagari][02][720P][GB][MP4]", "Tate no Yuusha no Nariagari", 2, 0, 0)]
+        [TestCase("【幻樱字幕组】【10月新番】【欢迎光临，千岁酱Okoshiyasu, Chitose-chan】【13】【GB_MP4】【1280X720】", "Okoshiyasu, Chitose-chan", 13, 0, 0)]
+        [TestCase("★1月新番★【百慕达三角~多彩田园曲~/ Bermuda Triangle - Colorful Pastrale】【02】【BIG5】【1080P】【V2】", "Bermuda Triangle - Colorful Pastrale", 2, 0, 0)]
+        [TestCase("【极影字幕社】 ★一月新番【Fate /EXTRA Last Encore】【10】【BIG5】【720P】（字幕社招内详）", "Fate /EXTRA Last Encore", 10, 0, 0)]
+        [TestCase("【悠哈璃羽字幕社&诸神字幕组& JOJO奇妙冒险吧】[ JOJO的奇妙冒险黄金之风_ JoJo 's Bizarre Adventure - Golden Wind][15][繁日双语][ x264 1080p] ", "JoJo 's Bizarre Adventure - Golden Wind", 15, 0, 0)]
+        [TestCase("【风之圣殿字幕组】★04月新番[闪亮 美妙☆频道 / Kiratto_Pri-chan][041][简体][720P][MP4]", "Kiratto Pri-chan", 41, 0, 0)]
+        [TestCase("	[悠哈璃羽字幕社&LoliHouse]不吉波普不笑/ Boogiepop wa Warawanai - 04v2 [WebRip 1080p HEVC-10bit AAC][简繁内封字幕] ", "Boogiepop wa Warawanai", 4, 0, 0)]
         //[TestCase("", "", 0, 0, 0)]
         public void should_parse_absolute_numbers(string postTitle, string title, int absoluteEpisodeNumber, int seasonNumber, int episodeNumber)
         {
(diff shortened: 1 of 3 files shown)


thezoggy:
i assume these are only found on a torrent site? does their api serve up multibyte characters?

whtsky:
> does their api serve up multibyte characters?

They don't have any api; CJK charaters are used in RSS feed.
https://share.dmhy.org/topics/rss/rss.xml

Taloth:
Probably easier to replace the special characters with [ ] as cleanup step in the parser (we have several of those), instead of adding another regex. Provided that one of the existing regexes will parse the result properly.

src/NzbDrone.Core.Test/ParserTests/QualityParserFixture.cs
```diff
@@ -119,6 +119,7 @@ public void should_parse_webdl480p_quality(string title, bool proper)
         [TestCase("Survivorman.The.Lost.Pilots.Summer.HR.WS.PDTV.x264-DHD", false)]
         [TestCase("Victoria S01E07 - Motor zmen (CZ)[TvRip][HEVC][720p]", false)]
         [TestCase("flashpoint.S05E06.720p.HDTV.x264-FHD", false)]
+        [TestCase("【极影字幕社】 ★1月新番【巴哈姆特之怒：马纳利亚朋友】【Shingeki no Bahamut Manaria Friends】【01】GB MP4_720", false)]
```
Taloth: Do you have other examples of this? Coz to me it just looks like a bad release name. We can't accept just '720' which is season 7 episode 20 in some cases, or 720 abs nr for certain korean shows.

src/NzbDrone.Core.Test/ParserTests/AbsoluteEpisodeNumberParserFixture.cs
```diff
@@ -94,6 +94,15 @@ public class AbsoluteEpisodeNumberParserFixture : CoreTest
         [TestCase("Tokyo Ghoul: RE S2 - Episode 4 VOSTFR (1080p)", "Tokyo Ghoul RE S2", 4, 0, 0)]
         [TestCase("To Aru Majutsu no Index III - Episode 5 VOSTFR (1080p)", "To Aru Majutsu no Index III", 5, 0, 0)]
         [TestCase("[Prout] Steins;Gate 0 - Episode 5 VOSTFR (BDRip 1920x1080 x264 FLAC)", "Steins;Gate 0", 5, 0, 0)]
+        [TestCase("【喵萌奶茶屋】★10月新番★[哥布林杀手/Goblin Slayer][12END][720p][繁体][招募翻译校对]", "Goblin Slayer", 12, 0, 0)]
+        [TestCase("[爱恋字幕社][盾之勇者成名录][Tate no Yuusha no Nariagari][02][720P][GB][MP4]", "Tate no Yuusha no Nariagari", 2, 0, 0)]
```
Taloth: Which of these can Sonarr ACTUALLY process?

src/NzbDrone.Core/Parser/Parser.cs
```diff
@@ -239,7 +243,7 @@ public static class Parser
                           RegexOptions.IgnoreCase | RegexOptions.Compiled),
 
                 //Anime - Title Absolute Episode Number
-                new Regex(@"^(?:\[(?<subgroup>.+?)\][-_. ]?)?(?<title>.+?)(?:[-_. ]+(?<absoluteepisode>(?<!\d+)\d{2,3}(\.\d{1,2})?(?!\d+)))+(?:_|-|\s|\.)*?(?<hash>\[.{8}\])?(?:$|\.)?",
+                new Regex(@"^(?:\[(?<subgroup>.+?)\][-_. ]?)?.*?(?<title>[a-z][\x20-\x5A\x5E-\x7F]+?)(?:[-_. ]+(?<absoluteepisode>(?<!\d+)\d{2,3}(\.\d{1,2})?(?!\d+)))+(?:_|-|\s|\.)*?(?<hash>\[.{8}\])?(?:$|\.)?",
```
Taloth: This doesn't seem correct. `[a-z][\x20-\x5A\x5E-\x7F]+?` should only match a starting a-z.

Taloth:
Like half the regexes in the parser are for anime, just not Chinese ones.

As mentioned, I'd like to know the actually usability for Chinese releases, since a few of the parser testcases cannot be processed automatically anyway due to various reasons.

thezoggy:
seems like the [chinese] stuff should just stripped/ignored...  but then you have to deal with the other parts which just isnt supported naming.. so really its back to badly named releases that only exist on this one torrent site...

src/NzbDrone.Core.Test/ParserTests/QualityParserFixture.cs
whtsky: Looks like they are the only releasing group using this format.. I'll drop this commit

whtsky:
> I'd like to know the actually usability for Chinese releases

It's not usable for Chinese anime releases. I found 0 working example right now.

> so really its back to badly named releases that only exist on this one torrent site...

It's not about torrent site, it's about the publisher.. Take a look at https://share.dmhy.org/topics/list/sort_id/2 , https://bangumi.moe/ and https://share.acgnx.se/sort-1-1.html . 

> seems like the [chinese] stuff should just stripped/ignored...

There's only a small (20% maybe) set of rls group put Chinese title in separated brackets.. The more common case is `[Chinese / English]`, `[Chinese_English]` and `[ChineseEnglish]` ( no separator )

src/NzbDrone.Core.Test/ParserTests/AbsoluteEpisodeNumberParserFixture.cs
whtsky: Oh.. Just found that Sonarr will strip all non-ASCII charaters in release title

whtsky:
So Sonarr strips all non-ASCII characters in title, which should make the parsing a lot easier than I initially thought. 

HELP WANTED: could you point out where did Sonarr did the strip? I want to replace `【】` to `[]` but couldn't find where to tweak

whtsky:
I’m closing this PR since non-ASCII characters are striped before parseTitle ..

Taloth:
sry mate, my work week is long and chaotic right now.

I suspect you don't want to pursue this feature any further, but I wanted to comment on your questions and remarks regardless:

The unicode appears to be stripped in RssParser, calling XmlCleaner.ReplaceUnicode, it would be feasible to perform common substitutions there, particularly the brackets.

> It's not usable for Chinese anime releases. I found 0 working example right now.

That's what I was afraid of, which means you'd need mapping on TheXEM to make certain titles work. Which begs the question on whether it's worth it.
If you believe there's a market for releases like this, then it might be possible to perform more substitutions on the rss level.

> There's only a small (20% maybe) set of rls group put Chinese title in separated brackets.. The more common case is [Chinese / English], [Chinese_English] and [ChineseEnglish] ( no separator )

Sounds to me like the parser should then have a regex to handle `[/?(English)]` titles.

PR #2928, opened 2019-02-01
Support anime release with title and episode number in brackets

whtsky:
#### Database Migration
NO

#### Description
Follow up https://github.com/Sonarr/Sonarr/pull/2904#issuecomment-459064953

+ Replace black lenticular brackets to square brackets
+ Remove `END` postfix after anime's absolute episode number
+ Remove separators before English title

#### Todos
- [x] Tests
- [ ] Documentation

*  #2898

### DIFF
--- a/src/NzbDrone.App.Test/NzbDrone.Host.Test.csproj
+++ b/src/NzbDrone.App.Test/NzbDrone.Host.Test.csproj
@@ -126,12 +126,12 @@
   <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
   <PropertyGroup>
     <PostBuildEvent Condition="('$(OS)' == 'Windows_NT')">
-      xcopy /s /y "$(SolutionDir)\..\_output\NzbDrone.Mono.*"   "$(TargetDir)"
-      xcopy /s /y "$(SolutionDir)\..\_output\NzbDrone.Windows.*"   "$(TargetDir)"
+      xcopy /s /y "$(SolutionDir)\..\_output\Sonarr.Mono.*"   "$(TargetDir)"
+      xcopy /s /y "$(SolutionDir)\..\_output\Sonarr.Windows.*"   "$(TargetDir)"
     </PostBuildEvent>
     <PostBuildEvent Condition="('$(OS)' != 'Windows_NT')">
-      cp -rv $(SolutionDir)\..\_output\NzbDrone.Mono.* $(TargetDir) || true
-      cp -rv $(SolutionDir)\..\_output\NzbDrone.Windows.* $(TargetDir) || true
+      cp -rv $(SolutionDir)\..\_output\Sonarr.Mono.* $(TargetDir) || true
+      cp -rv $(SolutionDir)\..\_output\Sonarr.Windows.* $(TargetDir) || true
     </PostBuildEvent>
   </PropertyGroup>
   <!-- To modify your build process, add your task inside one of the targets below and uncomment it. 
--- a/src/NzbDrone.Core.Test/NzbDrone.Core.Test.csproj
+++ b/src/NzbDrone.Core.Test/NzbDrone.Core.Test.csproj
@@ -345,6 +345,7 @@
     <Compile Include="OrganizerTests\FileNameBuilderTests\TitleTheYearFixture.cs" />
     <Compile Include="OrganizerTests\FileNameBuilderTests\TitleTheFixture.cs" />
     <Compile Include="ParserTests\MiniSeriesEpisodeParserFixture.cs" />
+    <Compile Include="ParserTests\UnicodeReleaseParserFixture.cs" />
     <Compile Include="ParserTests\ValidateParsedEpisodeInfoFixture.cs" />
     <Compile Include="Profiles\Delay\DelayProfileServiceFixture.cs" />
     <Compile Include="Profiles\Qualities\QualityIndexCompareToFixture.cs" />
--- a/src/NzbDrone.Core.Test/ParserTests/UnicodeReleaseParserFixture.cs
+++ b/src/NzbDrone.Core.Test/ParserTests/UnicodeReleaseParserFixture.cs
@@ -1,0 +1,70 @@
+﻿using FluentAssertions;
+using NUnit.Framework;
+using NzbDrone.Core.Indexers;
+using NzbDrone.Core.Test.Framework;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace NzbDrone.Core.Test.ParserTests
+{
+    [TestFixture]
+    public class UnicodeReleaseParserFixture : CoreTest
+    {
+        [TestCase("【喵萌奶茶屋】★10月新番★[哥布林杀手/Goblin Slayer][12END][720p][繁体][招募翻译校对]", "Goblin Slayer", "喵萌奶茶屋", 12)]
+        [TestCase("[桜都字幕组][盾之勇者成名录/Tate no Yuusha no Nariagari][01][BIG5][720P]", "Tate no Yuusha no Nariagari", "桜都字幕组", 1)]
+        [TestCase("[YMDR][輝夜姬想讓人告白～天才們的戀愛頭腦戰～][Kaguya-sama wa Kokurasetai][2019][02][1080p][HEVC][JAP][BIG5][MP4-AAC][繁中]", "Kaguya-sama wa Kokurasetai", "YMDR", 2)]
+        [TestCase("【DHR百合組】[天使降臨到我身邊！_Watashi ni Tenshi ga Maiorita!][05][繁體][1080P10][WebRip][HEVC][MP4]", "Watashi ni Tenshi ga Maiorita!", "DHR百合組", 5)]
+        [TestCase("【傲娇零&自由字幕组】[刀剑神域III UnderWorld/ Sword Art Online - Alicization ][17][HEVC-10Bit-2160P AAC][外挂GB/BIG5][WEB-Rip][MKV+ass ]", "Sword Art Online - Alicization", "傲娇零&自由字幕组", 17)]
+        [TestCase("【悠哈璃羽字幕社＆拉斯观测组】[刀剑神域Alicization _ Sword Art Online - Alicization ][17][x264 1080p][CHS]", "Sword Art Online - Alicization", "悠哈璃羽字幕社＆拉斯观测组", 17)]
+        [TestCase("【极影字幕社】 ★04月新番 【巴哈姆特之怒 Virgin Soul】【Shingeki no Bahamut Virgin Soul】【24】 【END】GB MP4_720P", "Shingeki no Bahamut Virgin Soul", "极影字幕社", 24)]
+        [TestCase("[愛戀&漫貓字幕组][10月新番][關於我轉生後成爲史萊姆那件事][18][720P][BIG5][MP4]", "關於我轉生後成爲史萊姆那件事", "愛戀&漫貓字幕组", 18)]
+        [TestCase("【咕咕茶字幕組】★1月新番[天使降臨到了我身邊! / Watashi ni Tenshi ga Maiorita!][04][1080P][繁體][MP4]", "Watashi ni Tenshi ga Maiorita!", "咕咕茶字幕組", 4)]
+        [TestCase("【千夏字幕组】【天使降临到了我身边！_Watashi ni Tenshi ga Maiorita!】[第05话][1080p_HEVC][简繁外挂]", "Watashi ni Tenshi ga Maiorita!", "千夏字幕组", 5)]
+        [TestCase("【DHR动研字幕组】[多田君不恋爱_Tada-kun wa Koi wo Shinai][13完][繁体][720P][MP4]", "Tada-kun wa Koi wo Shinai", "DHR动研字幕组", 13)]
+        [TestCase("【动漫国字幕组】★01月新番[Endro～！][01][1080P][简体][MP4]", "Endro～！", "动漫国字幕组", 1)]
+        public void should_parse_chinese_anime_releases(string postTitle, string title, string subgroup, int absoluteEpisodeNumber)
+        {
+            postTitle = XmlCleaner.ReplaceUnicode(postTitle);
+
+            var result = Parser.Parser.ParseTitle(postTitle);
+            result.Should().NotBeNull();
+            result.ReleaseGroup.Should().Be(subgroup);
+            result.AbsoluteEpisodeNumbers.Single().Should().Be(absoluteEpisodeNumber);
+            result.SeriesTitle.Should().Be(title);
+            result.FullSeason.Should().BeFalse();
+        }
+
+        [TestCase("[喵萌奶茶屋&LoliHouse]玛娜利亚魔法学院/巴哈姆特之怒Manaria Friends - 03 [WebRip 1080p HEVC-10bit AAC][简繁内封字幕]", "巴哈姆特之怒Manaria Friends", "喵萌奶茶屋&LoliHouse", 3)]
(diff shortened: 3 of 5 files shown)


Taloth:
Ok, two things. (I reply slowly atm, coz my workday is rather hectic and long the last few weeks)

One, You need to rebase on (and merge to) phantom-develop (which is our v3 beta). We're not going to do any new features on develop (v2).

Two, I recently added '[PreSubstitutions](https://github.com/Sonarr/Sonarr/commit/1d862db7c94f2712109dedeb3de26c7dfbaf9df5#diff-23ff1526865b238134cc108b0cf2acd3R20)' to phantom-develop. The idea is that it can be used to perform specific replacements to clean up odd releases that don't need complete separate regexes.
I figured it should be doable to replace the multiple brackets with something that can be parsed by the normal regexes. Like `^\[\]\[\s*/?\s*(...English...)\](?:\[\d{4}\])?\[(\d)(?:END)?\]` -> 'title - abs - '
Can you look into that? It should allow you to deal with END and / more strictly, instead of a global replace.

A few potential problems.
- The initial [] contained the chinese release group, afaik, and the anime regexes might not like that disappearing. But the suggested substitution changing it to 'title - abs - ' should help.
- What does the '10' mean in the Goblin Slayer testcase? Just wondering coz it affects the regex.

Disregarding the potential series titles mismatches, I think we can make this work.

whtsky:
> The initial [] contained the chinese release group, afaik, and the anime regexes might not like that disappearing.

Exisiting anime regexes works well with `[]`.

> * What does the '10' mean in the Goblin Slayer testcase? Just wondering coz it affects the regex.
It was "10月新番"(New anime in October)

whtsky:
> I figured it should be doable to replace the multiple brackets with something that can be parsed by the normal regexes. Like `^\[\]\[\s*/?\s*(...English...)\](?:\[\d{4}\])?\[(\d)(?:END)?\]` -> 'title - abs - '

It's actually harder than adding a new regex due to the complex title format. 
e.g. `[]10[/Goblin Slayer][12END][720p][][]` -> `[]10 Goblin Slayer - 12[720p][][]` still can't be parsed correctly by normal regex.

src/NzbDrone.App.Test/NzbDrone.Host.Test.csproj
```diff
@@ -126,12 +126,12 @@
...
-      xcopy /s /y "$(SolutionDir)\..\_output\NzbDrone.Mono.*"   "$(TargetDir)"
-      xcopy /s /y "$(SolutionDir)\..\_output\NzbDrone.Windows.*"   "$(TargetDir)"
+      xcopy /s /y "$(SolutionDir)\..\_output\Sonarr.Mono.*"   "$(TargetDir)"
+      xcopy /s /y "$(SolutionDir)\..\_output\Sonarr.Windows.*"   "$(TargetDir)"
     </PostBuildEvent>
     <PostBuildEvent Condition="('$(OS)' != 'Windows_NT')">
-      cp -rv $(SolutionDir)\..\_output\NzbDrone.Mono.* $(TargetDir) || true
-      cp -rv $(SolutionDir)\..\_output\NzbDrone.Windows.* $(TargetDir) || true
+      cp -rv $(SolutionDir)\..\_output\Sonarr.Mono.* $(TargetDir) || true
+      cp -rv $(SolutionDir)\..\_output\Sonarr.Windows.* $(TargetDir) || true
```
whtsky: wrong filename aborts build on fish because `No matches for wildcard`

whtsky:
Hi @Taloth, thanks for help and ideas, updated as requested.

Taloth:
I moved the unittests to a separate file and added a few (including a multi-episode one). Also refactored the regex. You can view the changes here: https://github.com/Sonarr/Sonarr/compare/phantom-develop...pr/n2928_whtsky
I removed the second regex, but that might return later once we sort out LoliHouse.

My primary concern right now is with the LoliHouse releases. Problem is that I can find a few that are not going to be easy to deal with. Such as the last testcase, which fails.
Are there other release groups that also omit [...] for the release titles?
I'm also seeing some releases with multiple / slashes. Such as Seishun Buta Yarou / Aobuta.

Plan C involves moving the ReplaceUnicode call from the RssParser to the ParseTitle function, so we have access to the original title before stripping out unicode. But it would be preferred if we can do it without.

Lemme know what you think and whether there are other groups like LoliHouse.

whtsky:
> Problem is that I can find a few that are not going to be easy to deal with. Such as the last testcase, which fails.

https://github.com/Sonarr/Sonarr/compare/phantom-develop...pr/n2928_whtsky#diff-df53920a9dc7a1ff43350300d0795cfcR22

That's not a problem of LoliHouse release ... Because "Alicization" is a self-created word and can't be translated into Chinese, this series from other release groups can't be parsed as well(except for YMDR, they put Chinese title into a separated brackets):

> 【傲娇零&自由字幕组】[刀剑神域III UnderWorld/ Sword Art Online - Alicization ][17][HEVC-10Bit-2160P AAC][外挂GB/BIG5][WEB-Rip][MKV+ass ]
> 【悠哈璃羽字幕社＆拉斯观测组】[刀剑神域Alicization _ Sword Art Online - Alicization ][17][x264 1080p][CHS] 

https://share.dmhy.org/topics/list?keyword=Sword+Art+Online+Alicization

> Are there other release groups that also omit [...] for the release titles?

I found "ZERO字幕组" who also omits `[]`: https://share.dmhy.org/topics/list/team_id/391
And since LoliHouse is the only release group focus on high-quality WebRip HEVC 1080p, I really want Sonaar to be able to parse LoliHouse's releases. (I always download LoliHouse's version if that's available.)

> Plan C involves moving the ReplaceUnicode call from the RssParser to the ParseTitle function, so we have access to the original title before stripping out unicode. But it would be preferred if we can do it without.

In fact I prefer moving the ReplaceUnicode call to the ParseTitle function, so we can see the original title in the WebUI's Search interface.

Taloth:
I was moving the unicode replace stuff to the Parser when I noticed an error in the original regex. The XmlCleanup.ReplaceUnicode was meant to filter out invalid characters, it was never meant filter out chinese characters, after fixing that bug the chinese characters became available in the release title.
I've updated the testcases and parsing code to match and pushed it to the branch.

I've added a separate fixup for LoliHouse & Zero since it's otherwise hard to separate the / matching logic from our normal anime releases.

Since the chinese characters remain in the release title now, we're left with one failing testcase for "Manaria Friends" since that release contains chinese characters in the english title part. Can you look into that and let me know what's expected there?

Finally, can you build and run this version locally and see if it works at all on your indexer?

whtsky:
> Since the chinese characters remain in the release title now, we're left with one failing testcase for "Manaria Friends" since that release contains chinese characters in the english title part. Can you look into that and let me know what's expected there?

"Manaria Friends" is expected: http://anidb.net/perl-bin/animedb.pl?show=anime&aid=11544

Taloth:
I know, but I meant about the Chinese characters that come before it in that particular release. Is it part of the original title, or did the release group just make a typo?

whtsky:
It's part of the original title ...

It's very confusing here: the official Chinese title is "Mysteria Friends（巴哈姆特之怒） / 玛娜利亚魔法学院" while the official Romaji title is "Mysteria Friends".

> And seems to be related to extended episodes for a game?
Yes this is a spin-off of "Shingeki no Bahamut"（巴哈姆特之怒）, the game and the anime.

whtsky:
Looks like Chinese charaters are still stripped out in search view:
image

whtsky:
Works:

> [悠哈璃羽字幕社&LoliHouse] 天使降临到了我身边！ / Watashi ni Tenshi ga Maiorita! - 04 [WebRip 1080p HEVC-10bit AAC][简繁内封字幕]

Not working:

> 【咕咕茶字幕組】★1月新番[天使降臨到了我身邊! / Watashi ni Tenshi ga Maiorita!][04][1080P][繁體][MP4]

Taloth:
Try again.

whtsky:
It works very well now. Thanks for your great work 👍

whtsky:
Oh, `【漫貓&愛戀字幕組】[五等分的新娘/五等分的花嫁/五等分の花嫁][Go-Toubun_no_Hanayome][01_03][BIG5][720P][HEVC]` is rejected because "Unknown Series"

Taloth:
first one: slashes in chinese title, and episode number contains underscore instead of dash.
second one: that's what we talked about, it parses the title as "巴哈姆特之怒 Manaria Friends", which would need an alias on http://thexem.de/xem/show/4569 to deal with considering it's part of the title.

whtsky:
> slashes in chinese title, and episode number contains underscore instead of dash.

Slashes in chinese title shouldn't been the issue because `【漫貓&愛戀字幕組】[五等分的新娘/五等分的花嫁/五等分の花嫁][Go-Toubun_no_Hanayome][02][BIG5][1080P][HEVC]` can be parsed.

whtsky:
Since chinese characters has become available in the release title, we might need to support `第x话`(epx)` and `完`, `完结` (END) in regex as well

Taloth:
Yes, it's the underscore in the episode number. because even the normal regex can't deal with that.

... what can of worms did we just open. :smile:

But question, google says that 第 means 'First' does that mean that 第5 = actually 1-5?

whtsky:
> google says that 第 means 'First' does that mean that 第5 = actually 1-5?

In this context 第 means 'number'. e.g. `第1` means `first`

Taloth:
ok, and tnx for adding the testcases. I'll see if I can tweak the _ scenario. But first I need more samples of that one to be sure it's not a one-time fluke.

whtsky:
I can't find another meaningful example right now, so I guess we can just ignore the `_` case.

Taloth:
Ok, that leaves the language detection. Is there anything we can use to detect that these releases are Chinese. In some releases it says '繁中', '繁体', '简繁外挂' and '简繁内封字幕', but not all of them and there seems to be quite a bit of variation.
I could check the release group/title for chinese characters, but that's hardly conclusive.

whtsky:
There're Simplified Chinese(`zh-hans`) and Traditional Chinese(`zh-hant`).

'繁中' = '繁体' -> zh-hant
'简繁外挂' -> zh-hans & zh-hant, external subtitle
'简繁内封字幕' -> zh-hans & zh-hant, embedded subtitle

Also:
CHS -> zh-hans
CHT -> zh-hant
BIG5 -> zh-hant (BIG5 is an character encoding for zh-hant)
GB -> zh-hans (charater encoding for zh-hans)

---

My idea is if the release title contains `简` , `繁` or `[GB]` `[BIG5]` `[CHS]` `[CHT]` then it's Chinese

Taloth:
Tnx for the thorough explanation. Is it remotely possible for '简' or '繁' to occur in series/episode titles?

whtsky:
> I could check the release group/title for chinese characters, but that's hardly conclusive.

It occurred to me that you can check for `字幕组` (Chinese for "Fansub group"). They're very common in titles of Chinese non-anime series.

whtsky:
> Is it remotely possible for '简' or '繁' to occur in series/episode titles?

AFAIK Japanese Kanji & Korean Hanja don't use '简' & '繁'. And Chinese series don't use them too, so it's safe.

Taloth:
Great. Then the last commits should have that detection work correctly.

image

whtsky:
"Mandarin" is one of the Chinese spoken dialect; I think use "Chinese" is more accurate

Taloth:
I am aware. However, both Mandarin and Cantonese are on the language table in Sonarr right now while Chinese isn't. I'm somewhat reluctant to modify it.

Next you should look at the alternative titles for the series you want. TheXem can map alternative titles, but right now sonarr ignores any thexem alternative title with non-latin characters.
So ideally we should start with finding out what kind of alternative titles are needed and how many. Then we can discuss it with TheXem moderators.

whtsky:
Just found that `巴哈姆特之怒 Manaria Friends` is the official name in Taiwan: https://acg.gamer.com.tw/acgDetail.php?s=79337

> what kind of alternative titles are needed
The official names from different country/regions I guess?

Taloth:
I've talked to markus and we've changed mandarin to chinese and repurposed the original cantonese id for icelandic instead. Somebody is going to be unhappy anyway and we can't deal with all intricacies until switch to iso language codes.

I've pushed the generic changes straight to phantom-develop. I've also squashed and rebased all the chinese parsing code we collaborated on to a single commit and pushed it to the branch. Please rebase/checkout your own branch on that and push, so we can track further changes here in the PR.

Regarding the 'alternative titles': No we're not really interested in official names. What we need to find out is how many of the Chinese release titles match the title listed on TheTVDB (or an alternative on TheXEM). My suggestion is that you start with the series you are currently interested in, and see which release titles we need to be able to handle. This includes your preferred releases such as the ones from LoliHouse, but also other good release groups.
It's probably useful for you to know that Sonarr uses 'CleanSeriesTitle' to convert 'The Big Show: Day or Dusk' to 'thebigshowdaydusk' before comparison with our database. Therefore punctuation rarely is an issue. (CJK characters aren't modified, so they are compared as is.)

Can you keep testing the current code for the series you want to follow and report back how well it works? Then we can merge the pr to phantom-develop in a few days it still works as desired.

whtsky:
Updated regex to support release without Chinese anime ( https://github.com/Sonarr/Sonarr/pull/2928/commits/af04a692c63c9a751385be1c9335f83d9a6da700#diff-df53920a9dc7a1ff43350300d0795cfcR27 )

Taloth:
Hm, I already had a similar release https://github.com/Sonarr/Sonarr/commit/af04a692c63c9a751385be1c9335f83d9a6da700#diff-df53920a9dc7a1ff43350300d0795cfcR23 which parsed ok, but after closer inspection that's because the 10月新番 part is in brackets and gets picked up as chinese title.

PS: you can just use `(?<chinesetitle>..)?` no need to put another `(?:...)` around it.

whtsky:
> PS: you can just use `(?<chinesetitle>..)?` no need to put another `(?:...)` around it.

But `(?:\]\[|[_/·]\s*)` needs to optional too

Taloth:
you're totally right. (I can't even read my own regex... 😅 )

whtsky:
Weird Issue: I imported Sword Art Online: Alicization then tried to search:

image

`[悠哈璃羽字幕社&拉斯观测组&LoliHouse] 刀剑神域: Alicization / Sword Art Online: Alicization - 17 [WebRip 1080p HEVC-10bit AAC][简繁内封字幕]` -> Unknown Series

but test cases didn't fail: https://github.com/Sonarr/Sonarr/pull/2928/files#diff-df53920a9dc7a1ff43350300d0795cfcR41

Taloth:
That's the third season of 'Sword Art Online': https://www.thetvdb.com/series/sword-art-online/seasons/3
TheXEM maps the series title to that series: http://thexem.de/xem/show/1334
Due to TheXEM mapping, Sonarr knows it's another series and which tvdb id, but doesn't know the series itself unless you have it in your collection.

(TheTVDB doesn't allow separate series entries for individual cours or continuations, so any 'Alicization' entry on TheTVDB is prone to be removed by thetvdb mods)

whtsky:
OK, reimported as 'Sword Art Online' and search works fine.

But `[UHA-WINGS&RATH&LoliHouse] Sword Art Online꞉ Alicization - 04` is treated as 1x04 when importing while Sonarr know it's 3x04 when searching
image

Taloth:
Sonarr expects the series folder to already be in tvdb numbering, the series title is not evaluated there.
If imported from the downloader (or via Wanted->Manual Import) it'll rename it to 'Sword Art Online - 54'.
PS: If you're going to fiddle with that, I suggest you make a backup of your library files.

whtsky:
I can live with that; I manage my media files with Plex anyway.

I switched all my anime downloading task from flexget to Sonarr; everything works fine so far.

Taloth:
Awesome, want to wait a few days, see if further changes are needed?

whtsky:
Let's wait a week to see if auto download works as expected

whtsky:
I think it's ready to merge

Taloth:
Will do. Thank you for your efforts on this PR.

Tell us if there are issues with shows/releases in the future and we'll try deal with them.

PR #2930, opened 2019-02-01
Allow torrent removal if finished.

breakthestatic:
#### Database Migration
NO

#### Description
Currently, torrents in Transmission need to meet both seeding time AND ratio requirements in order to be considered for deletion.  This presents a problem, because Sonarr relies on Transmission's own internal completion mechanism to stop torrents once the seeding goals are met. This means that if a torrent was seeded for enough time, but didn't meet the ratio, it would be stopped and marked as "finished" in Transmission, but Sonarr would never remove it since the ratio was not met (and never will be due to Transmission stopping it).

#### Todos
- [ ] Tests
- [ ] Documentation

*

### DIFF
--- a/src/NzbDrone.Core/Download/Clients/Transmission/TransmissionBase.cs
+++ b/src/NzbDrone.Core/Download/Clients/Transmission/TransmissionBase.cs
@@ -100,7 +100,7 @@ namespace NzbDrone.Core.Download.Clients.Transmission
 
                 item.CanMoveFiles = item.CanBeRemoved =
                     torrent.Status == TransmissionTorrentStatus.Stopped &&
-                    item.SeedRatio >= torrent.SeedRatioLimit;
+                    (item.SeedRatio >= torrent.SeedRatioLimit || torrent.isFinished);
 
                 items.Add(item);
             }
--- a/src/NzbDrone.Core/Download/Clients/Transmission/TransmissionBase.cs
+++ b/src/NzbDrone.Core/Download/Clients/Transmission/TransmissionBase.cs
@@ -100,7 +100,7 @@ namespace NzbDrone.Core.Download.Clients.Transmission
 
                 item.CanMoveFiles = item.CanBeRemoved =
                     torrent.Status == TransmissionTorrentStatus.Stopped &&
-                    (item.SeedRatio >= torrent.SeedRatioLimit || torrent.isFinished);
+                    (item.SeedRatio >= torrent.SeedRatioLimit || item.Status == DownloadItemStatus.Completed);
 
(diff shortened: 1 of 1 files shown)


LSL1337:
This feature doesn't work "properly" in Deluge either.
Only ratio is supported in Deluge. If I use a Deluge Plug-in to stop seeding based on time, sonarr doesn't see it, cos the status is not set to stop by the plug-in, only paused.
the only client where it works, is utorrent, but the linux development is over for years now, and i couldn't get the webui to keep working for more than a few days. (until than it was great).

so I use deluge, and complete download handling is useless (in my case, with seed time requirements).

2 questions: is it possible to remove torrents, which are paused in deluge?
OR is it feasable to sonarr track the seed time requirements, and force remove the torrents, when they are met? (probably not) :(

Taloth:
@breakthestatic 'IsFinished' has multiple meanings in transmission, for example while it's doing a startup Check. Also the code is shared between the Vuze and Transmission clients in Sonarr, so it would need to be verified on both.
But your proposed change doesn't compile since 'isFinished' should have a upper case 'I'.

@LSL1337 Please don't hijack a Transmission-related Pull Request. You should've create a separate forum post regarding your issue. But to answer you anyway:
Sonarr expects auto_managed=true, stop_at_ratio=true, ratio>stop_ratio and state=paused. (No such thing as 'stopped' in deluge)
Deluge built in seed-time logic only lowers the priority of the torrent, it doesn't actually stop/pause it.
The lack of an explicit stopped state makes dealing with it difficult. The only way for sonarr to detect 'temporarily paused' vs 'stopped' by comparing the ratio to stop_ratio.
If you want Sonarr to see it, then you'll have to fix Deluge by introducing an appropriate state or property indicating the seeding criteria have been met; or reduce stop_ratio once your seed time limit has been reached.
If you have further questions, do not reply here. Post on the forum and include a link to my comment. TY.

breakthestatic:
@Taloth, is my update more appropriate?  Instead of checking for torrent.IsFinished, I check item.Status, which a few lines above seems to have the required extra checks in place.  I'm not familiar with Vuze, would anything else need to be done to account for this?  It seems that the object in question is abstract enough to handle both, but I'm not a c# developer, so this isn't a strong point for me.

iFredOS:
Finally I found the answer to this issue. I have the same issue and this is very annoying I have to manually remove the downloads from Transmission and manually import them. Can I know when this fix will be add to the develop branch?

Taloth:
Transmission seed criteria handling should be better in v3 once https://github.com/Sonarr/Sonarr/pull/3141 gets merged.

Issue #2887, opened 2019-01-14
qBittorrent Authenticate URL changed in 4.1+ (Removed in 4.2)


zero77:
When trying to test the connection or send torrents to qBittorrent it doesn't work and instead gets a 404 error.
I think this maybe because qBittorrent nightly has dropped support for their old API.

Also, i am using the most up to date nightly vershon of Sonarr V3

> HTTP request failed: [404:NotFound] [POST] at [http://163.172.39.74:9000/login]

API v2 Documentations:
https://github.com/qbittorrent/qBittorrent/wiki/Web-API-Documentation#general-information

markus101:
Which version of qBittorrent?

zero77:
@markus101 
I am using the newest nightly.

markus101:
Does it specify an exact version? 4.1.x?

Based on the docs they changed the URL in 4.1, but that's been out since May; trying to understand if they've only changed it recently or it's been broken for that long.

zero77:
It look like 3 days ago.

> Drop legacy WebAPI support PR: #10145
https://github.com/qbittorrent/qBittorrent/commits/master

I am using qBittorrent v4.2.0 alpha but, i think the next stable update will also  drop legacy API support.

itsbebbo:
qBittorrent has included v2 of the API since version 4.x of qBittorrent. The old API has been left enabled in qBittorrent until a commit a few weeks ago. I can only assume that Sonarr / Radarr etc still use the v1 API hence why they broke on removal of the old API.

itsbebbo:
As the Radarr code I updated for qBittorrent was identical (forked from) Sonarr I will PR this and update it.

itsbebbo:
Pull Request added and code updated to deal with V2 of the API

PR #2945, opened 2019-02-08
qBittorrent API V2 Implementation

itsbebbo:
#### Database Migration
NO

#### Description
qBittorrent from v4.0 onwards includes a new API (V2). The old API will be removed on the release of v4.2 of qBittorrent (due any day). The current build of Radarr will only work with the original API and therefore the update to qBittorrent will break Radarr.

#### Todos
- [ ] Tests
- [ ] Documentation

#2887

### DIFF
--- a/src/NzbDrone.Core/Download/Clients/QBittorrent/QBittorrentProxy.cs
+++ b/src/NzbDrone.Core/Download/Clients/QBittorrent/QBittorrentProxy.cs
@@ -52,7 +52,16 @@ namespace NzbDrone.Core.Download.Clients.QBittorrent
 
         public QBittorrentPreferences GetConfig(QBittorrentSettings settings)
         {
-            var request = BuildRequest(settings).Resource("/query/preferences");
+            HttpRequestBuilder request;
+            if (settings.UseApiV2)
+            {
+                request = BuildRequest(settings).Resource("/api/v2/app/preferences");
+            }
+            else
+            {
+                request = BuildRequest(settings).Resource("/query/preferences");
+            }
+
             var response = ProcessRequest<QBittorrentPreferences>(request, settings);
 
             return response;
@@ -60,9 +69,18 @@ namespace NzbDrone.Core.Download.Clients.QBittorrent
 
         public List<QBittorrentTorrent> GetTorrents(QBittorrentSettings settings)
         {
-            var request = BuildRequest(settings).Resource("/query/torrents")
-                                                .AddQueryParam("label", settings.TvCategory)
-                                                .AddQueryParam("category", settings.TvCategory);
+            HttpRequestBuilder request;
+            if (settings.UseApiV2)
+            {
+                request = BuildRequest(settings).Resource("/api/v2/torrents/info")
+                                                .AddQueryParam("category", settings.MovieCategory);
+            }
+            else
+            {
+                request = BuildRequest(settings).Resource("/query/torrents")
+                                                .AddQueryParam("label", settings.MovieCategory)
+                                                .AddQueryParam("category", settings.MovieCategory);
+            }
 
             var response = ProcessRequest<List<QBittorrentTorrent>>(request, settings);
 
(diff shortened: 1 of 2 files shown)


Taloth:
You shouldn't put this behind a setting, autodetect it.

We're also contemplating whether to drop the old api altogether (for Sonarr v2) and only support qbit 4.2 for Sonarr v3.

itsbebbo:
> You shouldn't put this behind a setting, autodetect it.
> 
> We're also contemplating whether to drop the old api altogether (for Sonarr v2) and only support qbit 4.2 for Sonarr v3.

I do agree with you on this one TBH, but for simple testing it was easier to implement V2 with the setting. I can then look at autodetecting once it is confirmed as working.

rebekah65:
@Taloth
Should this pr be moved to the v3 branch, to be approved.

itsbebbo:
I can code but not use GIT!

There is still one issue with the code in that it doesn’t get the v2 API version. It will need the version return from the proxy to be changed to string data type as it is now in the format 2.x rather than a single number. In my view we can preserve the old API version number by casting to a sting and prefixing with a “1.” so we have 1.x and 2.x?

This would also mean changing any code that relies on the version being returned. 

Thoughts?

Taloth:
We'll decide later on sonarr v2 vs v3, but currently leaning to v2 mainly because we're not ready yet to force ppl to v3.
However we do want auto-detection for qbit api v2, which should be doable if you incorporate it in the GetVersion call by calling the apiv2 first, and if that fails try apiv1 and cache the result.

Regarding GetVersion: Change the return type to System.Version instead of string, it supports comparison operators out of the box.
It means changing a few lines of code in `QbitTorrent` class too, but that's minimal.

Secondly, It might be a good idea to duplicate the QbitTorrentProxy, so we have a v1 and v2 api proxy instead of one with a bunch of if statements.
Next there are a few routes, but it all boils down to a few variations:
- Add a `QbitTorrentProxy` wrapper class that implements IQbitTorrentProxy and does the auto-detection and selects the right proxy to redirect the individual calls to. We do something similar with `FallbackHttpDispatcher`, which selects automatically between the built-in http client interface and the external curl http client interface (useful back when mono didn't support tls1.2)
- Add a `QbitTorrentProxySelector` class that does the auto-detect logic and transparently returns the right proxy, instead of redirecting individual calls.
- Add a `QbitTorrent.GetProxy()` method that does the autodetection and returns the right proxy. This is less preferred because it requires QbitTorrent to import multiple proxy interfaces which complicates the unit-tests.

itsbebbo:
@Taloth

In what module is the QBittorrent class created and proxy set? I’m trying to work my way back through the code to see where it is instantiated and the proxy class set so I can determine exactly what I need to change.

Taloth:
Sonarr uses so called Dependency Injection. The dependency injection framework ensures that all parameters of the constructor receive an instance of appropriate implementation, and by injecting interfaces it becomes easy to write unit tests. For example QbitTorrentFixture, which mocks all those interfaces and can control the [behavior](https://github.com/Sonarr/Sonarr/blob/37054673b73df0cae7a0693e29ee31907df85dc0/src/NzbDrone.Core.Test/Download/DownloadClientTests/QBittorrentTests/QBittorrentFixture.cs#L64-L69) for specific tests.
I can explain more about the process of Dependency Injection if necessary, so please ask if I need to clarify it further.

This is also why we have *Proxy classes, they are a thin layer tasked only with dealing with the api http calls. We don't have tests for the *Proxy classes because you would need an actual torrent client to test them. But the IQbitTorrentProxy allows us to mock/simulate Qbit's behavior and test our own business logic.

You can, for example, create `interface IQBittorrentProxySelector`, and `class QBittorrentProxySelector` with a ctor `QBittorrentProxySelector(QBittorrentProxyLegacy proxyV1, QBittorrentProxy proxyV2)`.
Then change `class QBittorrent` ctor to inject `IQBittorrentProxySelector proxySelector` instead of `IQBittorrentProxy proxy`.
The chain from QBittorrent -> QBittorrentProxySelector -> QBittorrentProxyLegacy -> HttpClient etc will be automatically instantiated for you.

itsbebbo:
@Taloth - That is very impressive, never seen that done before, similar things with MTS but not to this level. I understand pretty much what is required now. Thanks.

itsbebbo:
Working on #2951 instead owing to large number of changes required.

Issue #2824, opened 2018-11-20
Add {MediaInfo HDR} to Naming Options


Ashkaan:
**Describe the problem**
Just like we naming files based on quality and source, it would be helpful to be able to add the HDR tag to our files like we can in Radarr.

**Describe any solutions you think might work**
Implement the {MediaInfo HDR} feature as it's implemented in Radarr.

Prolesious:
Please add this!

nismanoke:
+1 Support

markus101:
If you want to support a feature add a thumbs up to the original post. If you have something to add to the conversation then add a comment. Comments with +1 and without adding anything to the conversation will be removed.

markus101:
At this time we're not actively merging changes into the v2 branches outside of bug fixes. It'd need to be based off of the v3 work.

Matty666:
ok, no problem. I can take a look at v3. Which branch should I look at?

markus101:
`phantom-develop`

Matty666:
@markus101 @Taloth I'm happy to help out with this, the HDR flag in the name is something I wanted now that more 4K HDR content is appearing, so that's why I wanted to help with this issue. If you explain in a bit more detail what you are both thinking I can have a go at it and submit more pull requests

Matty666:
Regarding the name of the token, `VideoColourRange` might be confusing, there is a `colour_range` element in the mediainfo output that has a value `Limited` in the files I have checked.

And if we changed it to `VideoColourRange` or similar to indicate the type, then I guess it would need to return `SDR` or `HDR`, rather than just a flag to indicate if the media is HDR or not

Taloth:
Ok, lemme know how far you want to take it after you've read this.

Radarr added support for additional audio codecs like Atmos. To do that they upped the CurVer to 5. They also removed some backward-compatible support for VideoCodec (which was deprecated by mediainfo and replaced with VideoFormat), something i'd like to keep.
Personally I don't see a reason to support all the exotic audio formats that apply only for movies, but there's something to be said for being somewhat compatible with Radarr with the mediainfo detection.
My biggest issue with the Radarr code was that it lacks sufficient test cases for the new audio formats.
So I'd like to have that ported back to Sonarr so we only have a single version increment. Which means that the code needs to be refactored properly and additional unittests added.

[...]

PR #2950, opened 2019-02-11
Added MediaInfo VideoDynamicRange token that updates mediainfo if not latest revision

Matty666:
Added the `{MediaInfo VideoDynamicRange}` token for the file name which checks the media info properties for HDR.

Also modified the file name builder to check if any of the matched tokens are media info tokens that require a specific revision, and if there are any and the schema revision is old, then update the media info in the episode file.

I'm not as familiar with the codebase as I would like, so if I've messed anything up let me know!

#### Database Migration
NO

#### Description

#### Todos
- [x] Tests
- [ ] Documentation

https://github.com/Sonarr/Sonarr/issues/2824


src/NzbDrone.Core/MediaFiles/MediaInfo/MediaInfoUpdater.cs
```diff
@@ -0,0 +1,63 @@
...
+        {
+            if (string.IsNullOrEmpty(series.Path) || string.IsNullOrEmpty(mediaFile.RelativePath)) return;
+
+            var path = Path.Combine(series.Path, mediaFile.RelativePath);
+            Update(mediaFile, path);
+        }
+
+        public EpisodeFile GetOrUpdate(EpisodeFile mediaFile, Series series)
+        {
+            if(mediaFile.MediaInfo.SchemaRevision < VideoFileInfoReader.CURRENT_MEDIA_INFO_SCHEMA_REVISION)
```
Taloth: I think you misunderstood this part of the exercise. We already do this based on MINIMUM_MEDIA_INFO_SCHEMA_REVISION in UpdateMediaInfoService. This is because we don't want to update mediainfo for the entire library unless it's absolutely necessary.
The idea is that for HDR (and _only_ HDR) we need at least 5.

So basically, if the user has {MediaInfo VideoDynamicRange} in their format, we need to ensure the file is v5.
We can do that two ways, either make the  lazy loaded and when called ensure that we have the right schema version. Or, do a second (initial) pass using the regex to determine the minimum schema version required.

Matty666: Yes. I extracted the update part of the code from `UpdateMediaInfoService` into that class, and it's used in the `{MediaInfo VideoDynamicRange}` token handler, as well as the `UpdateMediaInfoService`. As I understood it, the update service gets called for a full show scan and updates all episodes that don't have the current revision. The token handler will call GetOrUpdate and update the mediainfo if it's not the current revision also. I thought that the token handler will only run if the token is in the naming config.

Taloth: > I thought that the token handler will only run if the token is in the naming config.

It will, but then you should provide the expected version number in GetOrUpdate, not via CURRENT_*. If another unrelated change is made in the future as v6, then you don't want the entire library to be rescanned just because ppl have the HDR token.
It also doesn't need to run on AudioCodec... yes there's a difference due to the new audio codec support added, but forcing a refresh there would cause a full library scan for most users. Whereas using the HDR token is a new one that the user would have to add explicitly.

_PS: To explain why it's so important to avoid rescanning: There are quite a bit of users with cloud storage and several cloud providers do not support partial file reads. Rescanning all mediainfo would cause a lot of traffic and therefore be rather undesirable. And the acceptable trade-off is to only scan mediainfo for new imports._

Matty666: ahh ok, yes I was wondering if each individual media info formatter should know about its own revision. so yes, HDR is >=v5 and some other property is >=v6

Yep I understand now, thanks for explaining the issues around rescanning, I'll bear that in mind. I didn't consider cloud storage. Makes sense to remove the audio codec ones too considering that. It'll use the new audio stuff if its there.

Matty666: I put audio codec back to how it was and changed the video dynamic range to specify its minimum required revision. An alternative here would be to convert all the formatters to specify their minimum required revision.

Also it might be better to change the formatter from a static to an injectable class that can depend on the media info updater and manage updating itself, rather than passing in the func to each format method?

Matty666:
I guess the downside to updating inside the token handler is that the other tokens can't take advantage of the updated media info if they have already run.

Taloth:
> I guess the downside to updating inside the token handler is that the other tokens can't take advantage of the updated media info if they have already run.

Yup, but if you want to fix that you'll need an initial pass over the regex (not complicated) to determine if updating is needed. Heck, even `if (usedTokens.ContainsIgnoreCase("MediaInfo VideoDynamicRange") && schema < 5) Refresh()` would work

The MediaInfoFormatter should be responsible only for formatting, you don't want to give it the episode info which would be required to refresh it, and passing in a Func callback is just fooling yourself about trying to separate concerns.
It's possible to change it from a static to an instance, but it's not worth it imo.

I'll look at the rest of the changes tomorrow.

Matty666:
Yeah I was thinking that the formatter should care what version of the schema it needs... but you're making me rethink that now. Either the file name builder (or something else, it didn't quite sit with me that the filename builder should care about the revision either) can be responsible for getting the right revision to give to the formatter. Let me look at the regex pass and make some changes.

Matty666:
@Taloth I've changed it to use the regex match before it builds the file name. The media info formatter now just formats the tokens.

src/NzbDrone.Core/Organizer/FileNameBuilder.cs
```diff
@@ -632,6 +642,22 @@ private string GetLanguagesToken(string mediaInfoLanguages)
             return string.Join("+", tokens.Distinct());
         }
 
+        private void UpdateMediaInfoIfNeeded(string pattern, Dictionary<string, int> minimumMediaInfoSchemaRevisions, EpisodeFile episodeFile, Series series)
+        {
+            var matches = TitleRegex.Matches(pattern);
+
+            var shouldUpdateMediaInfo = matches.Cast<Match>()
+                .Select(m => minimumMediaInfoSchemaRevisions.GetValueOrDefault(m.Value, -1))
+                .Where(r => r > -1)
```
Taloth: Don't really need the Where, do we?
Could even combine it into a single .Any, but main point should be readability.

src/NzbDrone.Core/Organizer/FileNameBuilder.cs
```diff
@@ -632,6 +642,22 @@ private string GetLanguagesToken(string mediaInfoLanguages)
...
+                .Any(r => episodeFile.MediaInfo.SchemaRevision < r);
+
+
+            if (shouldUpdateMediaInfo)
+            {
+                _mediaInfoUpdater.Update(episodeFile, series);
```
Taloth: We should honor the EnableMediaInfo config flag that the original UpdateMediaInfoService used. Otherwise the mediainfo might be refreshed even though the user explicitly disabled mediainfo.
Can either add the check here or do it in the UpdateMediaInfoService (effectively ignoring the request for update).

src/NzbDrone.Core/MediaFiles/MediaInfo/MediaInfoUpdater.cs
```diff
@@ -0,0 +1,53 @@
...
+using NzbDrone.Core.Tv;
+
+namespace NzbDrone.Core.MediaFiles.MediaInfo
+{
+    public interface IUpdateMediaInfo
+    {
+        void Update(EpisodeFile mediaFile, Series series);
+    }
+
+    public class MediaInfoUpdater : IUpdateMediaInfo
```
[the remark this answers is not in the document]
Matty666: no its fine. I originally extracted it out to keep the event handling part separate, but I can merge it back and inject the whole service into the filename builder

src/NzbDrone.Core/Organizer/FileNameBuilder.cs
```diff
@@ -586,6 +592,10 @@ private void AddMediaInfoTokens(Dictionary<string, Func<TokenMatch, string>> tok
             tokenHandlers["{MediaInfo Simple}"] = m => $"{videoCodec} {audioCodec}";
 
             tokenHandlers["{MediaInfo Full}"] = m => $"{videoCodec} {audioCodec}{mediaInfoAudioLanguages} {mediaInfoSubtitleLanguages}";
+
+            tokenHandlers["{MediaInfo VideoDynamicRange}"] =
+                m => MediaInfoFormatter.FormatVideoDynamicRange(episodeFile.MediaInfo);
+            minimumMediaInfoSchemaRevisions["{MediaInfo VideoDynamicRange}"] = 5;
```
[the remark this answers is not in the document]
Matty666: Ahh, ok, yeah sorry, didn't consider the different ways of defining the tokens!

src/NzbDrone.Core/Organizer/FileNameBuilder.cs
Matty666: Yes no problem

src/NzbDrone.Core/Organizer/FileNameBuilder.cs
Matty666: no we don't 👍

src/NzbDrone.Core/MediaFiles/MediaInfo/MediaInfoUpdater.cs
```diff
@@ -0,0 +1,53 @@
...
+
+        private void Update(EpisodeFile mediaFile, string path)
+        {
+            if (string.IsNullOrEmpty(path) || !_diskProvider.FileExists(path))
+            {
+                _logger.Debug("Can't update MediaInfo because '{0}' does not exist", path);
+                return;
+            }
+
+            mediaFile.MediaInfo = _videoFileInfoReader.GetMediaInfo(path);
```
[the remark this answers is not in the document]
Matty666: Yep no problem.

Taloth:
I think that covers it, right? Are you happy with it?

Matty666:
Yes, I think it's a lot better than my first attempt.

Apart from one small thing, the tests that check FileNameBuilder updates MediaInfo are tied to the minimum schema revision of MediaInfo VideoDynamicRange... so if in the future we change the minimum revision for it, to say 8, it'll break them.

This is because of the static dictionary, but i'm wary of exposing it or just adding code to facilitate tests. I'd have to make it writable and public (or provide a setter) so the test could control the values. I guess a setter that isn't part of the interface isn't too bad?

Or is this a YAGNI thing... and if the revision changes in the future that's when it can be refactored?

Taloth:
> so if in the future we change the minimum revision for it, to say 8, it'll break them.

Yes, and that's a good thing. Tests should break if you change code. And as you mentioned, one would just refactor the test to reflect the revision changes. So that's fine as is.

I have to say that we're absolutely impressed with your work and determination to fine-tune the whole thing (both PRs). Thank you!

I'll squash and merge the whole thing. See if the build server likes it.

Matty666:
@Taloth thanks! I've enjoyed the challenge and learnt a few things along the way. I was going to ask if you don't mind more pull request reviews I'd be happy to take on more changes, esp to help v3 along to it's release.

Taloth:
For v3 it's mostly polish and UI work. There is a larger item I want to implement, but that one isn't easy to explain due to some architectural implications (talking about remote download clients/seedbox handling).
There are plenty of open issues in general, but most need more information before anyone can pick them up.

Which download clients do you use?
If you use torrents and deluge, there's an issue about the next deluge version authentication mechanism not being supported in Sonarr. That's an issue that's going to be relevant for v2 and v3.

I'd suggest you look through the issue list to see if there are items that touch something you use yourself, since such issues are easier to work on. If you wish to pick up a specific item, please post there and include a proposal on how to deal with it. That way we have a chance to point out the edge cases.

PR #2951, opened 2019-02-11
qBittorrent V1 / V2 API Autoselect

itsbebbo:
#### Database Migration
NO

#### Description

#### Todos
- [ ] Tests
- [ ] Documentation

#2887

### DIFF
--- a/src/NzbDrone.Core/Download/Clients/QBittorrent/QBittorrent.cs
+++ b/src/NzbDrone.Core/Download/Clients/QBittorrent/QBittorrent.cs
@@ -101,21 +101,22 @@ namespace NzbDrone.Core.Download.Clients.QBittorrent
 
             foreach (var torrent in torrents)
             {
-                var item = new DownloadClientItem();
-                item.DownloadId = torrent.Hash.ToUpper();
-                item.Category = torrent.Category.IsNotNullOrWhiteSpace() ? torrent.Category : torrent.Label;
-                item.Title = torrent.Name;
-                item.TotalSize = torrent.Size;
-                item.DownloadClient = Definition.Name;
-                item.RemainingSize = (long)(torrent.Size * (1.0 - torrent.Progress));
-                item.RemainingTime = GetRemainingTime(torrent);
-                item.SeedRatio = torrent.Ratio;
-
-                item.OutputPath = _remotePathMappingService.RemapRemoteToLocal(Settings.Host, new OsPath(torrent.SavePath));
-
+                var item = new DownloadClientItem()
+                {
+                    DownloadId = torrent.Hash.ToUpper(),
+                    Category = torrent.Category.IsNotNullOrWhiteSpace() ? torrent.Category : torrent.Label,
+                    Title = torrent.Name,
+                    TotalSize = torrent.Size,
+                    DownloadClient = Definition.Name,
+                    RemainingSize = (long)(torrent.Size * (1.0 - torrent.Progress)),
+                    RemainingTime = GetRemainingTime(torrent),
+                    SeedRatio = torrent.Ratio,
+                    OutputPath = _remotePathMappingService.RemapRemoteToLocal(Settings.Host, new OsPath(torrent.SavePath)),
+                };
                 // Avoid removing torrents that haven't reached the global max ratio.
                 // Removal also requires the torrent to be paused, in case a higher max ratio was set on the torrent itself (which is not exposed by the api).
                 item.CanMoveFiles = item.CanBeRemoved = (!config.MaxRatioEnabled || config.MaxRatio <= torrent.Ratio) && torrent.State == "pausedUP";
+                
 
                 if (!item.OutputPath.IsEmpty && item.OutputPath.FileName != torrent.Name)
                 {
--- a/src/NzbDrone.Core/Download/Clients/QBittorrent/QBittorrent.cs
+++ b/src/NzbDrone.Core/Download/Clients/QBittorrent/QBittorrent.cs
@@ -19,7 +19,7 @@ namespace NzbDrone.Core.Download.Clients.QBittorrent
     {
         private readonly IQBittorrentProxy _proxy;
 
-        public QBittorrent(IQBittorrentProxy proxy,
+        public QBittorrent(IQBittorrentProxySelector proxySelector,
                            ITorrentFileInfoReader torrentFileInfoReader,
                            IHttpClient httpClient,
                            IConfigService configService,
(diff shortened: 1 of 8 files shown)


itsbebbo:
@Taloth - I can compile on local machine but get an error with interface creation for some reason. This may be a little beyond me to fix in the short term but hopefully some of the code will assist (I have split the code into V1 and V2 and checked that they both work so it is only the Proxy Selector code that is having issues).

src/NzbDrone.Core/Download/Clients/QBittorrent/QBittorrent.cs
```diff
@@ -28,7 +28,7 @@ public class QBittorrent : TorrentClientBase<QBittorrentSettings>
                            Logger logger)
             : base(torrentFileInfoReader, httpClient, configService, diskProvider, remotePathMappingService, logger)
         {
-            _proxy = proxy;
+            _proxy = proxySelector.GetProxy(Settings);
```
Taloth: Don't call this in the ctor, the Settings won'be populated yet. replace `_proxy.` with `GetProxy(Settings).` in the rest of the class. (Or probably cleaner, create a private `IQbittorrentProxy Proxy => GetProxy(Settings);` helper property)

Taloth:
The ctor thing is what's causing the interface error, coz Settings => Definition.Settings and Definition isn't assigned yet in the ctor.
So move the GetProxy call as I mentioned and it will turn out fine.

The TestFixture needs to be adjusted too because Selector.GetProxy returns null by default on the mock.
Add to QbittorrentFixture:
```
            Mocker.GetMock<IQBittorrentProxySelector>()
                .Setup(s => s.GetProxy(It.IsAny<QBittorrentSettings>()))
                .Returns(Mocker.GetMock<IQBittorrentProxy>().Object);
```

itsbebbo:
@Taloth

Will do, I was just tidying some other bits up before I read your review comment. Makes perfect sense, annoyed I missed that!

I will sort out the unit tests as well - this will not be today though. 

Thanks for all the help and guidance, it will allow me to develop my skills and help out again (though I think you have practically coded the whole lot in your comments to me!)...

itsbebbo:
@Taloth - I seem to have gotten the auto select working now with your guidance. Just need to update the unit tests now and run Sonarr against different versions of qBittorrent to ensure that it works correctly. Thanks again for the help.

Taloth:
I kinda pushed you in the deep end of the pool with Sonarr's architectural principles, so you're doing great. And please take your time, my greatest concern is the v2 api behaving differently than what we expect and Sonarr not properly detecting the various torrent states such as download finished and seeding finished.

A few more notes: There is no guarantee about the lifetime of the QbitTorrent instance (and also the QbittorrentProxySelector and Proxy instances). That's one of the reasons why you don't see us use fields much. These service classes are generally 'stateless'.
Instead of storing the chosen proxy in a field, it's better to 'cache' the api version for a specific qbit connection for a period of time. So that we don't have to call getversion each minute when we query the qbit torrent queue.
You can use a cache, similar to what [RarbgTokenProvider](https://github.com/Sonarr/Sonarr/blob/9b9597093c69e632791d30df1897199911b0d32a/src/NzbDrone.Core/Indexers/Rarbg/RarbgTokenProvider.cs#L30) uses, to cache the api version. Btw the proxy uses a similar cache to store the authentication cookie.
I'd like you to consider what potentially happens when the user upgrades Qbit from v3 to v4.2 without restarting Sonarr. This might determine the ideal cache timeout periods. And if the user is on v4.x then he's unlikely to ever downgrade, so longer cache periods might be warranted in that case.

I'll do a more detailed code review later, but right now we're in 'get it working'-mode.

Taloth:
@itsbebbo Did you get around to testing the various combinations? Particularly the one with qbit 3.x

itsbebbo:
Sorry, have been away for a while and not managed to test yet. May be the weekend before I get chance but v2.1.5 works with the changes and defaults to using v2 of the API.

Taloth:
I've tested it against 4.1.5 and 3.3.7.
The following changes https://github.com/Sonarr/Sonarr/compare/qbittorrent-api-v2:

GetVersion throws when api v2 doesn't exist, which needed to be handled. After some testing against 3.3.7 directly, I decided to add a `IsApiSupported` method.
The reason is that 3.3.7 returns the 404 Not Found status code, and 4.1.5 returns OK or Forbidden. This meant that we don't need to complete the (re)authentication when trying to detect which api is supported.
It also allowed me to avoid dealing with a specific http error codes in the Selector.
Earlier on I already seperated GetApiVersion and GetVersion, it made sense to keep that. So there are 3 functions now.

Secondly I added a 'force' parameter to the GetProxy function so we can force it to recheck the api version if the user presses Test. That way, if they switch qbit version and notice an error in Sonarr, they'll be able to remedy it by hitting Test instead of having to wait 10min/restart sonarr.

The v2 api url changes you made were correct, except setForceStart, which had a copy-paste error.
The v2 api spec says the topPrio command will return 409 Conflict instead of Forbidden (which was returned in v1) if Torrent Queueing is disabled. I adjusted it accordingly.
Add by magnet, by file, topPrio, forceStart and remove all seem to function as intended.
I didn't download an actual torrent completely, but the api spec indicates that hasn't changed.

I also included a couple of unrelated fixes for Magnet Links which I ran into during testing.

itsbebbo:
Thanks for all your hard work. As you may be able to tell I am on my journey learning c# and have in no way started to master it yet...

Taloth:
As I said before, you've been doing a great job. 
Most changes I made are direct result of tests against 3.3.7, which you didn't get around to yet.
The clock was ticking though, coz these changes need to be tested on the develop branch for a while before it can be released to master (and Radarr), and I rather have that done before qbit 4.2 gets released.

That said, I haven't actually tested against a 4.2 alpha because I can't find a windows installer for it.

zero77:
@Taloth 
> That said, I haven't actually tested against a 4.2 alpha because I can't find a windows installer for it.

You could try compiling it from sauce:
https://github.com/qbittorrent/qBittorrent/wiki/Windows-compilation

But, if not the easiest way may be to use the linux version in a VM:
https://launchpad.net/~qbittorrent-team/+archive/ubuntu/qbittorrent-unstable

zero77:
@Taloth
Is there a beta/nightly update channel for v3 so we can test this out.
Thanks

Taloth:
There is only a beta channel for v3, phantom-develop, but this hasn't be merged in v3 yet, only in v2 develop.

zero77:
@Taloth 
Is a merge of this PR into the v3 branch something that's being planned.

Taloth:
Of course, that's kinda the meaning of 'yet'. Please be patient.

Issue #320, opened 2013-09-04
Indexer Prioritization


SonarrBot:
Its a reoccurring feature request to add indexer prioritization, which we have not implemented since the idea is pretty broken (what if indexer A has a better quality, but it wasn't searched).

Two possible solutions that came to me are:
- Disable searching for an indexer (advanced option) - option to disable indexer from any searching, would only be used for RSS Sync - that would come with its own limitations, but might be worth it. This would save API calls.
- Prioritize downloads from indexers - Allow users to choose which indexer would be used to download a release when two indexers have the desired same quality - this would save downloads, but not API calls.

In addition to, disabling searching we could disable the RSS sync portion, then they could be used for searching only.

omega1970:
Since Sonarr decides which file to push to the download client, I would like to see a priority for otherwise equal releases, since I'm using a mix of torrent and NZB, I would prefer the NZB indexer over the much slower Torrent downloader, a priority could give me that choice, or any other mechanism that makes sense to provide my preference of download client.

mr-bo-jangles:
Sorry to butt in, but I'd personally think it'd work best like this.

Sonarr checks the top priority indexer for all releases (typically done due to paid account for one indexer), 
1. If top quality releases are found, then queue up the best one and proceed as usual.
2. If quality is not top, then queue up the best release you have and add this file to a backlog of files to search for "upgrades" for on other indexers.
3. If release is not found, then immediately search the other providers in priority order.

This would work well if you could specify how many API calls you are allowed with each provider per month so the system could ration out the API calls to alternative indexers based upon urgency.

I'd say it was more urgent to find a release at all than upgrade from SD to HD.

mr-bo-jangles:
A useful side effect could be you could dequeue releases before they even start if better quality was found elsewhere (assuming you had a backlog of downloads)

markus101:
1. Queue in the download client?
2. There is not backlog search in Sonarr: https://github.com/Sonarr/Sonarr/wiki/FAQ#how-does-sonarr-find-episodes

We've discussed it before, not necessarily here, but we're not interested in searching one indexer at a time, but we are considering differentiating between manual and automatic searches and letting users opt out of one or the other: https://github.com/Sonarr/Sonarr/issues/253

[...]

PR #2999, opened 2019-03-16
Indexer priority

DrivingOn66:
#### Database Migration
NO

#### Description
Adds an index priority feature to resolve https://github.com/Sonarr/Sonarr/issues/320. Each indexer has a Priority property. When similar releases are found for a search, the results will be sorted by highest index priority first.

https://github.com/Sonarr/Sonarr/issues/320


src/NzbDrone.Core/Indexers/IndexerDefaults.cs
```diff
@@ -3,5 +3,6 @@
     public static class IndexerDefaults
     {
         public const int MINIMUM_SEEDERS = 1;
+        public const int PRIORITY = 1000; //default > 0 so it allows users to specify higher or lower prio indexers without having to go trough all of them
```
markus101: Why does it need to be 1000 for users to set higher/lower priority? Zero as neutral makes more sense to me than any other number and a negative number to lower the priority would work fine and is more obvious.

src/NzbDrone.Core/Indexers/BroadcastheNet/BroadcastheNetSettings.cs
```diff
@@ -37,6 +38,9 @@ public BroadcastheNetSettings()
         [FieldDefinition(3)]
         public SeedCriteriaSettings SeedCriteria { get; } = new SeedCriteriaSettings();
 
+        [FieldDefinition(4, Label = "Priority", HelpText = "Used when searching for releases, a higher priority indexer will be used if releases found have the same quality", Advanced = true)]
```
markus101: It's not limited to searching and I don't think we want to explicitly call out the quality.

`Used when processing releases, a higher priority indexer will be used if releases found are otherwise equal`

src/NzbDrone.Core/Indexers/IIndexerSettings.cs
```diff
@@ -5,5 +5,9 @@ namespace NzbDrone.Core.Indexers
     public interface IIndexerSettings : IProviderConfig
     {
         string BaseUrl { get; set; }
+        /// <summary>
```
markus101: summary comments are not necessary.

src/NzbDrone.Core/DecisionEngine/DownloadDecisionPriorizationService.cs
```diff
@@ -13,18 +15,23 @@ public interface IPrioritizeDownloadDecision
...
         }
 
         public List<DownloadDecision> PrioritizeDecisions(List<DownloadDecision> decisions)
         {
+            var indexers = _indexerFactory.All();
             return decisions.Where(c => c.RemoteEpisode.Series != null)
                             .GroupBy(c => c.RemoteEpisode.Series.Id, (seriesId, downloadDecisions) =>
                                 {
-                                    return downloadDecisions.OrderByDescending(decision => decision, new DownloadDecisionComparer(_delayProfileService));
+                                    return downloadDecisions.OrderByDescending(decision => decision, new DownloadDecisionComparer(_delayProfileService, _logger, indexers));
```
markus101: A logger should not be passed in, they are class specific. Take a look at how `Parser` gets a logger.

src/NzbDrone.Core/DecisionEngine/DownloadDecisionComparer.cs
```diff
@@ -74,6 +83,31 @@ private int ComparePreferredWordScore(DownloadDecision x, DownloadDecision y)
...
         }
 
+        private int CompareIndexerPriority(DownloadDecision x, DownloadDecision y)
+        {
+            return CompareBy(x.RemoteEpisode, y.RemoteEpisode, GetIndexerPriority);
+        }
+
+        private int GetIndexerPriority(RemoteEpisode remoteEpisode)
+        {
+            if (remoteEpisode == null  || remoteEpisode.Release == null || remoteEpisode.Release.IndexerId == 0)
```
markus101: We don't do these null checks anywhere else and things aren't blowing up, so these should be unnecessary.

src/NzbDrone.Core/DecisionEngine/DownloadDecisionComparer.cs
```diff
@@ -74,6 +83,31 @@ private int ComparePreferredWordScore(DownloadDecision x, DownloadDecision y)
...
+            {
+                return int.MinValue;
+            }
+
+            try
```
markus101: Use `_indexersById.TryGetValue` instead.

src/NzbDrone.Core/DecisionEngine/DownloadDecisionComparer.cs
```diff
@@ -74,6 +83,31 @@ private int ComparePreferredWordScore(DownloadDecision x, DownloadDecision y)
...
+            {
+                var indexer = _indexersById[remoteEpisode.Release.IndexerId];
+                var settings = indexer.Settings as IIndexerSettings;
+                return settings == null ? int.MinValue : settings.Priority;
```
markus101: Do we really want the default value to be `int.MinValue`?, I think a neutral value (`0`) would be better.

src/NzbDrone.Core/DecisionEngine/DownloadDecisionComparer.cs
```diff
@@ -11,12 +12,19 @@ namespace NzbDrone.Core.DecisionEngine
...
         public delegate int CompareDelegate(DownloadDecision x, DownloadDecision y);
         public delegate int CompareDelegate<TSubject, TValue>(DownloadDecision x, DownloadDecision y);
 
-        public DownloadDecisionComparer(IDelayProfileService delayProfileService)
+        public DownloadDecisionComparer(IDelayProfileService delayProfileService, Logger logger, IEnumerable<IndexerDefinition> allIndexers)
         {
             _delayProfileService = delayProfileService;
+            _logger = logger;
+
+            //Cache the indexers to prevent DB calls for each compare call
```
markus101: Nit: space after comment slashes.

src/NzbDrone.Core/Datastore/Migration/131_indexer_priority.cs
```diff
@@ -0,0 +1,82 @@
...
+using Newtonsoft.Json.Linq;
+using NzbDrone.Core.Datastore.Migration.Framework;
+using System;
+using System.Collections.Generic;
+using System.Data;
+
+namespace NzbDrone.Core.Datastore.Migration
+{
+    [Migration(131)]
+    public class indexer_priority : NzbDroneMigrationBase
```
markus101: Do we need to do a migration? If a value isn't set then the default should apply. I think if we ever wanted to change the default we might want to update the DB for existing records, but it should be fine for now.

src/NzbDrone.Core.Test/DecisionEngineTests/PrioritizeDownloadDecisionFixture.cs
```diff
@@ -51,6 +58,7 @@ private RemoteEpisode GivenRemoteEpisode(List<Episode> episodes, QualityModel qu
             remoteEpisode.Release.PublishDate = DateTime.Now.AddDays(-age);
             remoteEpisode.Release.Size = size;
             remoteEpisode.Release.DownloadProtocol = downloadProtocol;
+            remoteEpisode.Release.IndexerId = indexerDefinition == null? 0 : indexerDefinition.Id;
```
markus101: space after `null`

src/NzbDrone.Core/Indexers/IndexerDefaults.cs
DrivingOn66: IMO I don't expect many users to try negative numbers since it's counter intuitive. I took 1000 because it allows users to set other indexers to 900 800 etc. and leave a gap for them to fill in (801-899) if I want to add another indexer in the middle later on (for example when they get into a new private tracker). Tbh we could also pick 100. Just my opinion, if you disagree I'll change it to 0.

src/NzbDrone.Core/Indexers/BroadcastheNet/BroadcastheNetSettings.cs
DrivingOn66: Suggestion based on the remark in discord: `Used when processing releases, a higher priority indexer will be used if releases found are otherwise equal. Note: only applied on indexers with the same protocol`

src/NzbDrone.Core/Datastore/Migration/131_indexer_priority.cs
DrivingOn66: I prefer to write a migration instead of implementing a fallback to default value if the value is missing because it's more maintainable IMO. Knowing that the Priority field will always be available simplifies code and obseletes fallback code. Again my opinion so if you disagree I can remove the migration and test.

src/NzbDrone.Core/Indexers/IndexerDefaults.cs
markus101: Negative to lower priority seems pretty intuitive to me and also matches preferred words in terms of negative is lower than the default. I asked @Taloth what he thinks.

src/NzbDrone.Core/Datastore/Migration/131_indexer_priority.cs
markus101: You don't need to implement anything, if nothing is set in the settings JSON in the DB it'll use the default of the model.

src/NzbDrone.Core/Indexers/BroadcastheNet/BroadcastheNetSettings.cs
markus101: A bit shorter/straight forward I think? `Used when processing releases, a higher priority indexer will be used if releases found are otherwise equal. Applies to indexers with the same protocol`

DrivingOn66: Agreed

src/NzbDrone.Core/Datastore/Migration/131_indexer_priority.cs
DrivingOn66: Ok, I'll remove the migration

src/NzbDrone.Core/DecisionEngine/DownloadDecisionPriorizationService.cs
```diff
@@ -13,18 +13,21 @@ public interface IPrioritizeDownloadDecision
...
         }
 
         public List<DownloadDecision> PrioritizeDecisions(List<DownloadDecision> decisions)
         {
+            var indexers = _indexerFactory.All();
             return decisions.Where(c => c.RemoteEpisode.Series != null)
                             .GroupBy(c => c.RemoteEpisode.Series.Id, (seriesId, downloadDecisions) =>
                                 {
-                                    return downloadDecisions.OrderByDescending(decision => decision, new DownloadDecisionComparer(_delayProfileService));
+                                    return downloadDecisions.OrderByDescending(decision => decision, new DownloadDecisionComparer(_delayProfileService, indexers));
```
markus101: Taloth mentioned this to me in our discussion, but hasn't had a chance to mention it here:

"the Comparer should have no knowledge of the Indexer settings, nor of the Indexer Definitions. Meaning the Comparer must be provided either a Dictionary of Indexer ID to Priority, or better should get an interface to a class that will provide it."

Given the comparer isn't using DI I think we'll want to pass in an instance of the class that is responsible for looking up the priority, which can also be responsible for caching the priority for a short period of time (maybe 5 minutes? and possible cleared when indexers are saved?).

src/NzbDrone.Core/Indexers/Torznab/TorznabSettings.cs
```diff
@@ -57,6 +57,7 @@ public class TorznabSettings : NewznabSettings, ITorrentIndexerSettings
         public TorznabSettings()
         {
             MinimumSeeders = IndexerDefaults.MINIMUM_SEEDERS;
+            Priority = IndexerDefaults.PRIORITY;
         }
 
         [FieldDefinition(6, Type = FieldType.Number, Label = "Minimum Seeders", HelpText = "Minimum number of seeders required.", Advanced = true)]
```
markus101: This will conflict with the field definition for Priority in `NewznabSettings`.

src/NzbDrone.Core/DecisionEngine/DownloadDecisionComparer.cs
markus101: I think we want this to be `0` as well, but should come from the interface discussed in my other comment.

src/NzbDrone.Core/Indexers/IndexerDefaults.cs
```diff
@@ -3,5 +3,6 @@
     public static class IndexerDefaults
     {
         public const int MINIMUM_SEEDERS = 1;
+        public const int PRIORITY = 100; //default > 0 so it allows users to specify higher or lower prio indexers without having to go trough all of them
```
markus101: Taloth suggested 0 here as well.

src/NzbDrone.Core/Indexers/IIndexerSettings.cs
```diff
@@ -5,5 +5,6 @@ namespace NzbDrone.Core.Indexers
     public interface IIndexerSettings : IProviderConfig
     {
         string BaseUrl { get; set; }
+        int Priority { get; set; }
```
Taloth: @markus101 I didn't think about this a few days ago, but this should not be a Field at all! If it's global for all indexers, it belongs straight in IndexerDefinition (and yes, that would require a db migration and that also means it needs to be added to the UI separately.)

src/NzbDrone.Core/Indexers/Omgwtfnzbs/OmgwtfnzbsSettings.cs
```diff
@@ -35,10 +34,7 @@ public OmgwtfnzbsSettings()
         public string ApiKey { get; set; }
 
         [FieldDefinition(2, Label = "Delay", HelpText = "Time in minutes to delay new nzbs before they appear on the RSS feed", Advanced = true)]
-        public int Delay { get; set; }
-
-        [FieldDefinition(3, Label = "Priority", HelpText = "Used when processing releases, a higher priority indexer will be used if releases found are otherwise equal. Applies to indexers with the same protocol", Advanced = true)]
-        public int Priority { get; set; }
+        public int Delay { get; set; }        public int Priority { get; set; }
```
Taloth: Leftover Priority

src/NzbDrone.Core/Datastore/Migration/131_indexer_priority_column.cs
```diff
@@ -0,0 +1,18 @@
...
+namespace NzbDrone.Core.Datastore.Migration
+{
+    [Migration(131)]
+    public class indexer_priority_column : NzbDroneMigrationBase
+    {
+        protected override void MainDbUpgrade()
+        {
+            Alter.Table("Indexers").AddColumn("Priority").AsInt32().Nullable();
+
+            Execute.Sql("UPDATE Indexers SET Priority = 0");
```
Taloth: Use the IndexerDefaults.PRIORITY here for consistency?

src/NzbDrone.Core/Datastore/TableMapping.cs
```diff
@@ -60,7 +60,7 @@ public static void Map()
                   .Ignore(i => i.Enable)
                   .Ignore(i => i.Protocol)
                   .Ignore(i => i.SupportsRss)
-                  .Ignore(i => i.SupportsSearch)
+                  .Ignore(i => i.SupportsSearch)
```
Taloth: stray whitespace.

If you're using visual studio, install the EditorConfig extension. It'll read the .editorconfig in the root and configure the IDE for a few settings such as tab,spaces, eol etc

src/NzbDrone.Core.Test/DecisionEngineTests/PrioritizeDownloadDecisionFixture.cs
```diff
@@ -17,16 +17,21 @@
...
 using NzbDrone.Core.Profiles.Languages;
 using NzbDrone.Core.Test.Languages;
+using NzbDrone.Core.Indexers.Torznab;
 
 namespace NzbDrone.Core.Test.DecisionEngineTests
 {
     [TestFixture]
     public class PrioritizeDownloadDecisionFixture : CoreTest<DownloadDecisionPriorizationService>
     {
+        private static IndexerDefinition _TestIndexer1 = new IndexerDefinition { Id = 1, Settings = new TorznabSettings { Priority = 10 } };
```
Taloth: Move Priority to Definition the test too.

src/NzbDrone.Core/Indexers/Newznab/Newznab.cs
```diff
@@ -69,6 +69,7 @@ private IndexerDefinition GetDefinition(string name, NewznabSettings settings)
                        EnableRss = false,
                        EnableAutomaticSearch = false,
                        EnableInteractiveSearch = false,
+                       Priority = 0,
```
Taloth: IndexerDefaults.PRIORITY

src/NzbDrone.Core/Indexers/Newznab/NewznabSettings.cs
```diff
@@ -82,7 +82,7 @@ public NewznabSettings()
         public IEnumerable<int> AnimeCategories { get; set; }
 
         [FieldDefinition(5, Label = "Additional Parameters", HelpText = "Additional Newznab parameters", Advanced = true)]
-        public string AdditionalParameters { get; set; }
+        public string AdditionalParameters { get; set; }        public int Priority { get; set; }
```
Taloth: Leftover Priority

src/NzbDrone.Core/Indexers/Torznab/Torznab.cs
```diff
@@ -58,6 +58,7 @@ private IndexerDefinition GetDefinition(string name, TorznabSettings settings)
                        EnableRss = false,
                        EnableAutomaticSearch = false,
                        EnableInteractiveSearch = false,
+                       Priority = 0,
```
Taloth: IndexerDefaults.PRIORITY

src/NzbDrone.Core/Indexers/Rarbg/RarbgSettings.cs
```diff
@@ -22,7 +22,7 @@ public RarbgSettings()
         {
             BaseUrl = "https://torrentapi.org";
             RankedOnly = false;
-            MinimumSeeders = IndexerDefaults.MINIMUM_SEEDERS;
+            MinimumSeeders = IndexerDefaults.MINIMUM_SEEDERS;
```
Taloth: Nitpick: Stray whitespace, see the editorconfig comment I made elsewhere.

src/NzbDrone.Core/Datastore/TableMapping.cs
DrivingOn66: Thx for the tip, removed the whitespace

src/NzbDrone.Core/DecisionEngine/DownloadDecisionPriorizationService.cs
DrivingOn66: Agreed, except on the caching part. Looking at other bits of code I don't see any similar caching and I don't think it will make a noticable difference here. 'Premature optimization is the root of all evil' - it will add unnecessary complexity to the code which could contain bugs. We could add the caching part in the future if performance becomes an issue.

markus101: You were caching before and this seems like a pretty high impact area for caching the priority for a few minutes to drop the DB calls to pretty much 1 per indexer per RSS Sync/search.

`DelayProfileService` and `SceneMappingService` are both in the release processing path that use caching, with `DelayProfileService` wiping the cache when a delay profile is added/updated/deleted.

DrivingOn66: As I'm beginning to implement this I'm strongly feeling that this might be a bit overengineered. Since we moved the priority from the settings object to the IndexerDefinition class the code is simplified, Moving it to a new class seems kind of silly and makes the code less readable (code is spread out). What do you think @Taloth

Taloth: Code is already spread out: DownloadDecisionPriorizationService calls IndexerFactory to get the list of indexers, only to provide the Comparer with that result. The Comparer converts it into a Dictionary and stores that. Then later, the Comparer uses the dictionary to get the IndexerDefinition for a specific indexer id, pulls the Priority property from it and uses that during comparisons.

If you add a separate service (such as `IProvideIndexerPriority`), then the logic becomes localized in that one class. DownloadDecisionPriorizationService no longer has to fetch the list of indexers, and the Comparer won't have to fiddle with IndexerDefinitions either.

If I want to create unittests for the Comparer, I no longer have to create a bunch of fake IndexerDefinitions. Instead I can just mock up the method in IProvideIndexerPriority and return the priority directly.

I get why you think this is overengineering, but it has a specific purpose in mind: To hide the inner workings of one part of the application from another.
To name another example: IIndexerFactory has several methods to provide Enabled, or RssEnabled, or SearchEnabled indexers. It could've just as easily provided an All() method and have the caller filter the result, but that would mean the caller becomes aware of which settings cause an indexer to be 'Enabled' duplicating that particular code everywhere it's needed.
The Enabled logic is of course more complex than Priority, but it's the same principle.

I hope that clarifies our thought process a bit.

PR #3017, opened 2019-03-29
Phantom certificate validation

markus101:
#### Database Migration
NO

#### Description
Allows users to control whether certificate validation is enabled, disabled for local addresses or disabled for everything, defaulting to enabled (which is a breaking change).

Not sure how best to communicate this breaking change or if it really matters all that much given it's still an alpha.


jalmansor:
Add it to the changelog, document in the Wiki/FAQ, and if you're feeling super productive I suppose you could cross post to the various communications channels (forums, reddit for v3, discord)?

src/Sonarr.Api.V3/Indexers/ReleasePushModule.cs
```diff
@@ -1,50 +1,111 @@
...
         public ReleasePushModule(IMakeDownloadDecision downloadDecisionMaker,
                                  IProcessDownloadDecisions downloadDecisionProcessor,
+                                 IIndexerFactory indexerFactory,
                                  Logger logger)
         {
             _downloadDecisionMaker = downloadDecisionMaker;
             _downloadDecisionProcessor = downloadDecisionProcessor;
+            _indexerFactory = indexerFactory;
             _logger = logger;
+;
```
Taloth: stray ;

src/Sonarr.Api.V3/Indexers/ReleasePushModule.cs
```diff
@@ -1,50 +1,111 @@
...
+            {
+                try
+                {
+                    var indexer = _indexerFactory.Get(release.IndexerId);
+                    release.Indexer = indexer.Name;
+                    _logger.Debug("Push Release {0} associated with indexer {1} - {2}.", release.Title, release.IndexerId, release.Indexer);
+                }
+                catch (ModelNotFoundException)
+                {
+                    _logger.Debug("Push Release {0} not associated with unknown indexer {0}.", release.Title, release.IndexerId);
```
Taloth: 'not' 'unknown', sounds like a double negative.

src/Sonarr.Api.V3/Indexers/ReleasePushModule.cs
```diff
@@ -1,50 +1,111 @@
...
+
+            if (firstDecision?.RemoteEpisode.ParsedEpisodeInfo == null)
+            {
+                throw new ValidationException(new List<ValidationFailure>{ new ValidationFailure("Title", "Unable to parse", release.Title) });
+            }
+
+            return MapDecisions(new [] { firstDecision }).AsResponse();
+        }
+
+        private void ResolveIndexer(ReleaseInfo release)
```
Taloth: Maybe include this change in the v2 api implementation? Given that third-parties will likely use that api.

src/Sonarr.Api.V3/Indexers/ReleasePushModule.cs
```diff
@@ -1,50 +1,111 @@
...
+                                 IIndexerFactory indexerFactory,
                                  Logger logger)
         {
             _downloadDecisionMaker = downloadDecisionMaker;
             _downloadDecisionProcessor = downloadDecisionProcessor;
+            _indexerFactory = indexerFactory;
             _logger = logger;
+;
+            _releaseValidator = new ResourceValidator<ReleaseResource>();
+            _releaseValidator.RuleFor(s => s.Title).NotEmpty();
```
Taloth: What's the idea behind replacing PostValidator with this?

src/Sonarr.Api.V3/Indexers/ReleasePushModule.cs
markus101: Copy paste from the v2 implementation, I'll fix both (also didn't intend for this commit to land here, oops).

src/Sonarr.Api.V3/Indexers/ReleasePushModule.cs
markus101: PostValidator doesn't work unless you `CreateResource`, which would require a ID response that then gets the resource, which would be more complicated than we wanted here.

Taloth: Right, so the original code using the Post[..] = never even used the validator.

Shouldn't we use 'ReadResourceFromRequest' instead of this.Bind in that case? That would invoke the Validator while still allowing a custom result. Similar to what we do for Post["test"] and Post["action/{action}"].

markus101: Yeah, that's a better option, didn't think about. Now that that's out of the way 😄

Taloth:
Then, looks good to me.

I see no reason to document it for migration purposes. And if ppl already on v3 run into it unexpectedly then tough luck, it's beta.
But do include it in the changelog text.

PR #3031, opened 2019-04-05
New: Limit Custom Scripts to specified scripts folder

markus101:
#### Database Migration
NO

#### Description
Breaking change for scripts that converts the `path` to a drop down that lists the scripts available in the selected scripts folder (`SonarrAppData/Scripts` by default). The available scripts are fetched asynchronously from the API when the modal is opened and can be refetched by the user.

image

Do we also want to get rid of the `Arguments` setting? With these restrictions I don't think it adds too much value.


src/NzbDrone.Common/Processes/ProcessProvider.cs
```diff
@@ -361,9 +353,19 @@ private List<Process> GetProcessesByName(string name)
             return processes;
         }
 
-        private string GetMonoArgs(string path, string args)
+        private (string Path, string Args) GetPathAndArgs(string path, string args)
         {
-            return string.Format("--debug {0} {1}", path, args);
+            if (PlatformInfo.IsMono && path.EndsWith(".exe", StringComparison.InvariantCultureIgnoreCase))
```
markus101: More cases like these are a concern to me, I'm not sure it lead to the path + args we currently have, but it makes that approach make more sense.

src/NzbDrone.Common/NzbDrone.Common.csproj
```diff
@@ -72,6 +72,9 @@
     <Reference Include="ICSharpCode.SharpZipLib">
       <HintPath>..\packages\ICSharpCode.SharpZipLib.Patched.0.86.5\lib\net20\ICSharpCode.SharpZipLib.dll</HintPath>
     </Reference>
+    <Reference Include="System.ValueTuple, Version=4.0.3.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51, processorArchitecture=MSIL">
+      <HintPath>..\packages\System.ValueTuple.4.5.0\lib\net461\System.ValueTuple.dll</HintPath>
```
Taloth: Are we sure this ValueTuple assembly works on mono?

Be sure we have a unittest involving it.

src/NzbDrone.Core/Notifications/CustomScript/CustomScript.cs
```diff
@@ -127,6 +134,12 @@ public override void OnRename(Series series)
         public override ValidationResult Test()
         {
             var failures = new List<ValidationFailure>();
+            var scriptsPath = GetScriptsFolder();
```
Taloth: Since the user can change the scripts folder, wouldn't this make it possible for an attacker to change the scripts folder and still execute whatever they like? (`../../../usr/bin`)

markus101: That is a concern (but also allowed on other apps), that'd be my argument for removing `Arguments` (no pun intended).

Taloth:
I'm not going to comment on the UI code right now. it's too unfamiliar for me.

A few comments:

- How are we going to migrate existing users with scripts? It might be useful to allow executing existing scripts from any path and thus only validate the path when adding it. This would maintain some backward compatibility till the user has a chance to replace it.
- The configurable Global Script path kinda defeats the entire purpose of the PR. So either that needs to be hardcoded, or only in the configfile.
- It's not strictly necessary to keep the arguments, but I think it'll be useful to keep them.

Final concern is that I do think that there are still attack vectors. How far do we want to go, and which kind of attacks do we want to prevent?
Can we ensure that nothing gets downloaded into the scripts folder? 
1. If an attacker has access to sonarr, he would conceivably have access to sabnzbd api too and be able to change the download location. I don't think there's something we can do about this unlikely scenario, so it's out of scope.
2. iirc, Sonarr can't be configured to have a rootfolder in the AppData folder, and thus it's not possible to use extra files .bat to copy in malicious files.
3. ... any other scenario?

markus101:
- Only validating on save makes sense
- This is inline with every other app that I can think of, nzbget takes it a step further and requires a specific block of text in the script IIRC, I don't think making that a config-file only setting makes a difference and just reduces visibility, I'm not even sure limiting it really makes that much of a difference the more I think about it
- I think arguments exposes the bigger issue, you can run `cmd.exe /c C:\test.bat` or presumably execute anything you want

1. What's to stop them from downloading it somewhere else and having SAB execute a script that copies it into Sonarr's script folder? Or manually telling SAB to extract files there? There are way too many variables to try and control there I feel

If someone has access to SAB and Sonarr and they're running with wide open permissions for either I think that's going to be the biggest problem, not where the script executes from. I get that this is about minimizing the impact and I think dropping args is the best way to achieve that.

Perhaps instead of limiting where the scripts can be run from we prevent them from running from specific locations + eliminate the args?

Taloth:
> Perhaps instead of limiting where the scripts can be run from we prevent them from running from specific locations + eliminate the args?

The more I think about it, I have to agree with that assessment.

It's still going to be hard to migrate existing CustomScript installs because we won't be able to convert the path and arguments to a new 'script file' field reliably. So yes, we'd have to preserve Path and Arguments of existing entries and execute them without allowing the adding or modification of such entries.

src/NzbDrone.Common/Extensions/StringExtensions.cs
```diff
@@ -22,9 +22,14 @@ public static object NullSafe(this object target)
             return "[NULL]";
         }
 
+        public static string FirstCharToLower(this string input)
+        {
+            return input.First().ToString().ToLower() + input.Substring(1);
```
Taloth: Use ToLowerInvariant, otherwise you'll get bad conversions in some languages.
btw. char.ToLowerInvariant(input.First()) would work too.

Same at line 32.

src/NzbDrone.Common.Test/NzbDrone.Common.Test.csproj
```diff
@@ -85,6 +85,8 @@
     <Compile Include="ExtensionTests\IEnumerableExtensionTests\IntersectByFixture.cs" />
     <Compile Include="ExtensionTests\Int64ExtensionFixture.cs" />
     <Compile Include="ExtensionTests\IPAddressExtensionsFixture.cs" />
+    <Compile Include="ExtensionTests\StringExtensionTests\FirstCharcacterToLowerFixture.cs" />
```
Taloth: Typo in filename.

src/NzbDrone.Common.Test/ExtensionTests/StringExtensionTests/FirstCharcacterToUpperFixture.cs
```diff
@@ -0,0 +1,18 @@
+﻿using FluentAssertions;
+using NUnit.Framework;
+using NzbDrone.Common.Extensions;
+
+namespace NzbDrone.Common.Test.ExtensionTests.StringExtensionTests
+{
+    [TestFixture]
+    public class FirstCharcacterToUpperFixture
```
Taloth: Typo.

src/NzbDrone.Common.Test/ExtensionTests/StringExtensionTests/FirstCharcacterToUpperFixture.cs
```diff
@@ -0,0 +1,18 @@
...
+    {
+        [TestCase("hello", "Hello")]
+        [TestCase("camelCase", "CamelCase")]
+        [TestCase("a full sentence", "A full sentence")]
+        public void should_capitalize_first_character(string input, string expected)
```
Taloth: Doesn't happen, but what about an empty string protection?

Issue #3033, opened 2019-04-06
Add support for Gotify notifications


stephanrenggli:
**Describe the problem**

Receive notifications via Gotify, the same as you would with Pushover or Pushbullet.

**Additional context**

Gotify seems similar to Pushover, though you are able to host it yourself.
Gotify: https://gotify.net/

stephanrenggli:
I implemented this and it seems to work.

PR: https://github.com/Sonarr/Sonarr/pull/3034

evertonstz:
Came here to ask for this support too!

billysyrett:
@markus101 Can we get the Gotify notifications built by @stephanrenggli merged into a release?

stephanrenggli:
Gotify notifications have been in sonarr for around a year now. You can set it up by going to `Settings > Connect > Add > Gotify`
![image](https://user-images.githubusercontent.com/5135902/79338222-35b92680-7f27-11ea-86c8-a5ecbf5ca4d1.png)

Taloth:
Only in v3 beta, hence @billysyrett confusion. Sonarr v2 receives no new features.

billysyrett:
Ah that makes sense. Cheers, I've switched to v3 now.

PR #3034, opened 2019-04-07
New: Added support for Gotify notifications

stephanrenggli:
#### Database Migration
NO

#### Description
This adds support for Gotify (self-hosted notification service similar to pushover) to Sonarr.

#### Todos
- [ ] Tests
- [ ] Documentation

* https://github.com/Sonarr/Sonarr/issues/3033


src/NzbDrone.Core/Notifications/Gotify/GotifyService.cs
```diff
@@ -0,0 +1,55 @@
...
+        private readonly Logger _logger;
+
+        public GotifyProxy(Logger logger)
+        {
+            _logger = logger;
+        }
+
+        public void SendNotification(string title, string message, GotifySettings settings)
+        {
+            var URL = settings.GotifyServer + "/message?token=" + settings.AppToken;
```
markus101: Variables should be camel case, `url` not `URL`. Use `string.Format` better yet string interpolation.

src/NzbDrone.Core/Notifications/Gotify/GotifyService.cs
```diff
@@ -0,0 +1,55 @@
...
+            {
+                const string title = "Test Notification";
+                const string body = "This is a test message from Sonarr";
+
+                SendNotification(title, body, settings);
+            }
+            catch (Exception ex)
+            {
+                _logger.Error(ex, "Unable to send test message");
+                return new ValidationFailure("ApiKey", "Unable to send test message");
```
markus101: There isn't an `ApiKey` setting, but unless the error is specific to an invalid key it should either be on the server or no property.

src/NzbDrone.Core/Notifications/Gotify/GotifyService.cs
stephanrenggli: Changed this to use string interpolation to build the url.

src/NzbDrone.Core/Notifications/Gotify/GotifyService.cs
stephanrenggli: Removed the `ApiKey` property.

markus101:
Thanks for getting this started, I made some changes and landed this in https://github.com/Sonarr/Sonarr/commit/90626e353f4625798e68ac066f5038d8d4eaf5e5

Part of the cleanup was getting rid of the merge commit, but also fixed the naming of the proxy, removed test from the proxy and renamed a few things.

PR #3035, opened 2019-04-08
Add origin, source, container, and codec to title when determining release restriction matches

jtpavlock:
restriction matches

#### Database Migration
NO

#### Description
Right now release restrictions only check a release's title for "must contain" or "must not contain" matches. I added some optional release fields (may or may not be set by the indexer) that will be included when searching for a match to a restriction.

A good use case for this will be if an indexer specifies a scene release in its origin rather than requiring it in the filename (title) of the episode. With this change, a user can now reliable say "must not contain 'scene'" provided their indexer of choice has this info. 

#### Todos
- [ ] Tests
- [ ] Documentation (may require indexer specific documentation?)

* 

As a side note, I've never coded in C# and relied on stack overflow for how to check null property values and concatenate strings. Please forgive me if it's incorrect.

### DIFF
--- a/src/NzbDrone.Core/DecisionEngine/Specifications/ReleaseRestrictionsSpecification.cs
+++ b/src/NzbDrone.Core/DecisionEngine/Specifications/ReleaseRestrictionsSpecification.cs
@@ -30,10 +30,10 @@ namespace NzbDrone.Core.DecisionEngine.Specifications
             _logger.Debug("Checking if release meets restrictions: {0}", subject);
 
             var title = subject.Release.Title;
-            var origin = subject.Relase.Origin ?? "";
-            var source = subject.Relase.Source ?? "";
-            var container = subject.Relase.Container ?? "";
-            var codec = subject.Relase.Codec ?? "";
+            var origin = subject.Release.Origin ?? "";
+            var source = subject.Release.Source ?? "";
+            var container = subject.Release.Container ?? "";
+            var codec = subject.Release.Codec ?? "";
 
             title += origin + source + container + codec;           var title = subject.Release.Title;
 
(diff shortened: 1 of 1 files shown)


src/NzbDrone.Core/DecisionEngine/Specifications/ReleaseRestrictionsSpecification.cs
```diff
@@ -30,6 +30,13 @@ public virtual Decision IsSatisfiedBy(RemoteEpisode subject, SearchCriteriaBase
             _logger.Debug("Checking if release meets restrictions: {0}", subject);
 
             var title = subject.Release.Title;
+            var origin = subject.Release.Origin ?? "";
+            var source = subject.Release.Source ?? "";
+            var container = subject.Release.Container ?? "";
+            var codec = subject.Release.Codec ?? "";
+
+            title += origin + source + container + codec;           var title = subject.Release.Title;
```
markus101: `title` is already instantiated here (and you don't want two "lines" on one line).

I also don't think we'd want to combine all those variables into a single string to compare, if you for example you had a regex restriction that was `\bx265\b` it'd fail because you'd end up without any spaces/separation between the words. Checking each field individually would be more reliable.

markus101:
This will also need to be based off of `phantom-develop` (v3), we're not accepting PRs for v2 outside of major bug fixes.

src/NzbDrone.Core/DecisionEngine/Specifications/ReleaseRestrictionsSpecification.cs
jtpavlock: I agree on the regex I was just thinking about that. I think a simple delimiter like a “.” would work though to prevent issues

jtpavlock:
I messed up and pre-maturely deleted this fork to start from scratch off of v3 instead of rebasing. Request to move to the other pr instead?

markus101:
To this PR? I suppose that's the only option, given a PR is a conversation I wanted to keep the discussion and my comments in one place.

jtpavlock:
Right, sorry. And no this is the old one I deleted

Issue #2782, opened 2018-10-21
Accept/Reject releases based on indexer's release attributes


jordandrako:
**Describe the problem**
Some indexers, such as HDBits, support tags on their releases. It would be great if we could specify required tags in the indexer settings like Radarr does so we can avoid grabbing releases that will unnecessarily increase seed ratios. General support would be great, but I'm specifically looking for HDBits internal tags to be supported as a requirement before grabbing.

**Describe any solutions you think might work**
Here is some documentation on the tags Radarr supports: https://github.com/Radarr/Radarr/wiki/Indexer-Flags#1-required-flags

**Additional context**
These tags can also be used as a preference to grab freeleech or golden popcorn over other encodes. I've searched the issues and can't find anything related specifically to indexer tags, just sonarr tags in the restrictions which aren't the same: they require the required term to be in the release name. Indexer tags are part of the indexer's api as far as I know.

jordandrako:
Also, here is how Jackett handles it (I think): https://github.com/Jackett/Jackett/blob/060972475f13ffe59fcc09c51ffe91a547a29029/src/Jackett.Common/Definitions/hdbits.yml#L73

markus101:
Wouldn't `Internal` be part of the release name?

Something like additional parameters could be used to exclude releases if they can be filtered on HDBit's side, but I don't see a good way to translate "tags" on the indexer to specific restrictions within Sonarr and I don't think it makes a lot of sense to attempt it for a single indexer.

shauder:
This would be helpful for me also and is useful in Radarr.

2fst4u:
Please support this. Radarr handles this without issue and allows the flag to be shown on the search interface which is really helpful for making sure you grab a release that won't "cost" you.

Radarr can also prioritise entries with flags above ones without flags.

Same as OP, I'd like to be able to tell Sonarr to only grab torrents if they're freeleech.

2fst4u:
A further addition to why this would be great:

If you have multiple private trackers with different freeleech parameters, it would be preferable to prioritise one over the other if the same file is freeleech on one but not in another.

For example I have one tracker that makes all releases freeleech for a certain period of time, I would prefer these for brand new releases. I have another tracker that makes all releases over a certain file size freeleech, these would be preferable for older files.

Because my trackers aren't prioritised in any other way for the same file in a search, one of my indexers keeps being used more than the other when I'd like the one with freeleech releases to be used instead.

[...]

PR #3036, opened 2019-04-08
Add key-value type restriction matches & indexer granularity in release profile

jtpavlock:
#### Database Migration
Yes 135 - 8b5ad4024bf964f2dfdc75d43f089edaab1aba79

#### Description
Allows for key-value type restriction matches in release profiles. Keys may include things such as "origin" or "codec". 

Example for requiring scene releases: must contain: origin:scene

Also allows for specifying what indexer a release profile will apply to and enabling or disabling a release profile.

#### Todos
- [x] Attach new release profile fields to backend
- [ ] Indexer dropdown based on enabled indexers in release profile
- [ ] Autocomplete supported keys (possibly in future pr?)
- [x] Key-value logic
- [ ] Add all planned keys
- [ ] Add support for various indexers for keys
- [ ] Tests
- [ ] Documentation

*


markus101:
Please don't open a new PR, you need to rebase the on the `phantom-develop` branch and force push your changes.

jtpavlock:
> Please don't open a new PR, you need to rebase the on the `phantom-develop` branch and force push your changes. and update the branch for the PR.

I uh sorta got rid of that old fork to just start from scratch.... whoops. Okay let's see if I can figure this out

jtpavlock:
Can this one be re-opened?

src/NzbDrone.Core/DecisionEngine/Specifications/ReleaseRestrictionsSpecification.cs
```diff
@@ -30,6 +30,13 @@ public virtual Decision IsSatisfiedBy(RemoteEpisode subject, SearchCriteriaBase
             _logger.Debug("Checking if release meets restrictions: {0}", subject);
 
             var title = subject.Release.Title;
+            var origin = subject.Release.Origin ?? "";
+            var source = subject.Release.Source ?? "";
+            var container = subject.Release.Container ?? "";
+            var codec = subject.Release.Codec ?? "";
+
+            title += "." + origin + "." + source + "." + container + "." + codec;
```
markus101: If this information isn't part of release title, I'm not sure periods make the most sense since they're unlikely scene releases.

I still don't think appending it to the title is the best course of action, maybe checking each individually is too much, so testing them all together would be better.

Use string interpolation (`var str = $"{origin}.{source}"` etc) for combining these, it looks like that'll also handle null values.

markus101: Anything we're checking will need to be logged for troubleshooting purposes later, if a release is rejected or accepted for something not in the release name it'll be confusing to troubleshoot.

jtpavlock:
> @Taloth re-requesting your thoughts on this one.
> 
> There will need to be tests and I'm still unsure of the real world usage of this, what indexers/release names have release titles that don't contain this information, but have the extra tags that do?

Many torrent files don't actually use the word scene (check https://predb.me/?cats=tv) but rather just use their group name. Using a list of scene groups in must not contain may be _good enough_ but it won't always capture all groups and constant new groups popping up. 

Sure just checking the title will get us there 90% of the time, but I checking Origin, Codec, Container, etc. if set would mean 100% accurate restriction matching. 

BTN, a site I use and probably the most popular private tv tracker, sets these fields and often the info is not represented in the filename alone. I'm sure other trackers have similar set-ups as well.

src/NzbDrone.Core/DecisionEngine/Specifications/ReleaseRestrictionsSpecification.cs
jtpavlock: I didn't think about logging. Maybe checking individually would be the best approach and then you can log based on what matched the restriction.

Taloth:
@markus101 We only fill those fields for BTN. It's exclusively used for dvd/bluray detection.
So it's usecase is rather limited. And I would very much like to see examples where the Release Title on btn does not contain the required tags (other than 'scene').

I'm quite reluctant with matching on a concatenation of these fields. Primarily because they are visible nowhere, and thus the user would have no way of knowing how their restrictions behave.

If we want to consider custom criteria like this then it should be in the format of `origin:scene` (or `origin:/scene|ptp/i` for it's regex variant). It would also pave the way for criteria like the ['tags'](https://torznab.github.io/spec-1.3-draft/revisions/1.3-Torznab-Release-Tags.html) for certain newznab releases or any other future field we might expose.
But if we commit to fields like this, then we have to _commit_, meaning we'd have to determine what such a field might contain and how they would map for various indexers.

jtpavlock:
I didn't realize they were only set on BTN. I know on Jackett these types of fields are set quite often for many trackers into their "description" field, but I'm not sure how that gets translated on Sonarr's end when looking at releases. 

And you're right for BTN that the other fields (besides origin) are going to be in the filename as far as I can tell. I suppose the use case is fairly limited then, but would still be helpful as far as BTN users go being able to set a preferred origin. Perhaps given the availability, users from other sites may come forward to set these fields in their specific indexer's parser?

Adding indexer specific tags in the form of tag:value (e.g. origin:scene as you specified) for restrictions I think would be useful and the documentation could gradually grow to include more as they are implemented for each indexer.

jtpavlock:
So I started on some code for key:value type restrictions, but I'd like some help setting up the development environment if one of you guys has a sec.

If you guys don't think this would be a useful feature, I won't waste my time anymore, but I'd love to start testing it out when I can figure this dev env out.

markus101:
https://github.com/Sonarr/Sonarr/issues/2782 was opened previously for filtering based on release attributes from HDBits, which would be similar if not the same thing.

All these things seem very indexer specific though, so requiring one could result in rejecting every release from other indexers. Radarr does flags per indexer, which works to some degree, but has limited uses for the BTN tags ad likely the HDBits tags, unless the user always wanted internals like the original issue.

I'm not opposed to including the field based filtering, but I'm not sure how we best deal with the per-indexer fields and how we're going to expose these fields to the user.

jtpavlock:
Wouldn't Radarr's method of Indexer specific flags be sufficient? The only difference I'd like to implement would be to discern between Ignored and Required as currently Radarr only supports required indexer flags.

jtpavlock:
Another option would be to have prefixed key:value options in ignored and restricted settings such as

btn:origin:user
hdbits:origin:internal

releases not from the referenced indexer won't have an effect on the decision making

Taloth:
That's the tricky part, I don't want ambiguity or indexer specific fields. Any fields added will have to be global and clearly defined.

For torznab we had a preliminary spec to add 'tags' such as 'internal'. But it has to be clear what internal means in this case, because for torrent sites internal means releases by resident (p2p) release groups. But for scene it means an release internal to the release group itself.
Tags such as 'scene' or 'p2p' might be reasonable but the list of tags would get messy quickly. If you define a field 'origin' then you need to define the possible values and be strict about it.
I'm open to both approaches, 'origin:scene' vs 'tag:scene' both seems like an option that's likely to be reliable.

One edge-case is that irssi autodl announce channels will not contain these tags/fields, and thus be rejected (or erroneously approved).

jtpavlock:
Some indexer specific fields may be unavoidable (not sure of a good tv example, but I'm thinking of "golden popcorn" releases in PTP). 

@markus101 also made the point that not specifying a specific indexer in the tag could result in false rejections if all of a user's indexers do not support that tag. I would imagine it to be fairly difficult/tedious to find 1:1 mapping of tags between indexers.

I think it would be harder and more confusing to do it this way:

| Tags   | Values | Supported Indexers  |
|---|---|---|
| Origin  | Internal, Scene  | BTN, HDBits  |
| UploadFactor  | 0,1,2  | HDBits  | 

than to do it like this

| Indexer   | Supported Tags | Possible Values  |
|---|---|---|
| BTN  | Origin  | Internal, Scene, User  |
| HDBits  | Origin  | Internal, Other, Scene  | 
| HDBits  | UploadFactor  | Freeleech, Neutral, Half  | 

As for the announce channels, I'm not too familiar with the layout, but I'm curious how you currently grab other fields such as seeders, or earlier you said you needed them for dvd/bluray detection in BTN. I also wonder how Radarr would handle this same problem since they already have indexer tags implemented.

Taloth:
I've been giving this a little bit of thought.

I hope you understand why I prefer not to have a growing list of indexer specific tags or filters. I rather not support something that end up with a billboard filled with notes and instructions.
My point is that it should not matter what each indexer supports. The user needs to decide how they want other indexer results to be treated and adjust their release profile accordingly.

**'Origin'**

BTN has P2P, Internal, Scene and User as Origin. It'd be safe to classify that as P2P, Internal, Scene and Other. Matching HDBits. However, it'd be better to just have P2P, Scene and Other as option. Internal is a subclass of P2P and better suited via another tag.
Please note that you can 'prefer' scene in two ways: origin:Scene as preference, or origin:P2P & origin:Other as the opposite. The difference is how releases from other indexes (without the origin field) get treated.

'Internal' means the resident P2P group. It's possible to introduce tag:Internal for that, or origin:Internal, but it's just as easy to just use for example -NTb as preference in that particular case, so imo holds little value.

**'UploadFactor'**

Torznab specifies UploadFactor and DownloadFactor as [two decimal numbers](https://torznab.github.io/spec-1.3-draft/revisions/1.2-Torznab-Download-Upload-Factors.html) in order to incorporate any factor a tracker might have. It's my understanding that Jackett does emit these two attributes on their api when applicable, Sonarr just doesn't do anything with it atm.
Personally I'd prefer to treat this as a separate feature, being able to prioritize by the factor, but it sounds reasonable to me to automatically add 'tag:freeleech' if the downloadvolumefactor is 0 to make it easier for a user to filter on a commonly used value. And there are other tags that might be useful. I just prefer them all to be based on those two factors.

But this is an important edge-case: The user might likely want to prefer freeleech _on a specific site_ rather than globally. Is it likely that users want to do that? And how would we allow the user to configure that?
Prefixing the indexer name would work, meaning we're back to square one, but would that be the best choice in this case? How about jackett's 'all' feed?
How accessible indexer-based filtering must be depends on how likely it's going to be used.

> As for the announce channels, I'm not too familiar with the layout, but I'm curious how you currently grab other fields such as seeders, or earlier you said you needed them for dvd/bluray detection in BTN. 

Simple, we don't grab other fields when the user pushes releases to Sonarr via the ReleasePush api.

jtpavlock:
I support your stance and understand your concern with a growing wiki of notes. I agree that just P2P, Scene, and Other should cover most use-cases (really just P2P and Scene should be sufficient with restrictions). 

> Please note that you can 'prefer' scene in two ways: origin:Scene as preference, or origin:P2P & origin:Other as the opposite. The difference is how releases from other indexes (without the origin field) get treated.

Are you saying this is a current feature? Or just what you're envisioning how the feature would operate once implemented?

> 'Internal' means the resident P2P group. It's possible to introduce tag:Internal for that, or origin:Internal, but it's just as easy to just use for example -NTb as preference in that particular case, so imo holds little value.

Agree Internal should be easily implemented as a preference 

> 'UploadFactor'
Torznab specifies UploadFactor and DownloadFactor as two decimal numbers in order to incorporate any factor a tracker might have. It's my understanding that Jackett does emit these two attributes on their api when applicable, Sonarr just doesn't do anything with it atm.
Personally I'd prefer to treat this as a separate feature, being able to prioritize by the factor, but it sounds reasonable to me to automatically add 'tag:freeleech' if the downloadvolumefactor is 0 to make it easier for a user to filter on a commonly used value. And there are other tags that might be useful. I just prefer them all to be based on those two factors.

> But this is an important edge-case: The user might likely want to prefer freeleech on a specific site rather than globally. Is it likely that users want to do that? And how would we allow the user to configure that?
Prefixing the indexer name would work, meaning we're back to square one, but would that be the best choice in this case? How about jackett's 'all' feed?
How accessible indexer-based filtering must be depends on how likely it's going to be used.

I was just using uploadfactor as a potential use-case, and didn't really put much thought into it. I also think it should be treated as a separate feature.


So as it currently stands, I'd like to, at a minimum, implement a restriction based match for P2P and Scene that would be treated appropriately by all supported indexers if that is something that sounds good to you.

jtpavlock:
Another option, going in a different direction, would be to do something more in line with what #2782 suggests and have indexer-specific tag options presented in a clear manner on each specific indexer settings page. 

This would eliminate the need for a reference wiki and would allow more indexer-specific settings that may not align with another indexer's settings.

Taloth:
> Are you saying this is a current feature? Or just what you're envisioning how the feature would operate once implemented?

Just how it would operate in the future. I was pointing out how indexers without 'origin' would be treated. It's essentially 'origin:unknown'.
If you prefer scene, then you'd say 'origin:scene' +10, but that would exclude potential scene release from indexers not having the 'origin' field.
That leaves the option to either say 'origin:scene' +10 & 'origin:unknown' +10. Or say 'origin:p2p' -10.
My concern there is how to convey that to the user.

 
> Internal isn't just one group (at least with BTN) and could be just as variable as scene groups so using preferences might work if you know the name of all the resident internal groups, but it isn't ideal.

Fair point, although I didn't mean that it would be just one group. But aside of upload/download ratio bonuses, what would be the advantage of internal groups? They're still p2p, and their releases will still show up on other sites.

> I was just using uploadfactor as a potential use-case, and didn't really put much thought into it. I also think it should be treated as a separate feature.
> 
> So as it currently stands, I'd like to, at a minimum, implement a restriction based match for P2P and Scene that would be treated appropriately by all supported indexers if that is something that sounds good to you.
> 
> If possible, adding the ability to specify Internal or Other origins I think would be useful as well.

We need to have a broad view here, regardless of what we're going to implement.  If we come up with something that works for our usecase now, but later realize it's totally unworkable for other usecases, then we're going to have a problem.

1) BTN has 'Mixed' origin for season packs. How would you deal with that?
2) If a user wants freeleech, but only on a specific site. How would the user specify that in an intuitive manner?

@markus101 I fear we're going to dig ourselves into a hole if we try to incorporate tag/fields into preferred words since it lacks expressions. Especially since we've been desperately trying to avoid expressions in general.
I'd love to hear if you have suggestions on how to tackle that.

jtpavlock:
> But aside of upload/download ratio bonuses, what would be the advantage of internal groups? They're still p2p, and their releases will still show up on other sites.

I think the ratio bonuses are the main appeal with a minor use if you perhaps trust/favor your site's p2p groups over other sites for encoding.

> BTN has 'Mixed' origin for season packs. How would you deal with that?
If a user wants freeleech, but only on a specific site. How would the user specify that in an intuitive manner?

One thing we could do to perhaps achieve indexer-granularity while maintaining ease of use could be a dynamic drop down list?

Something like a tag/field section containing

| Indexer  | Restriction Type | Tag  | Value  | 
|---|---|---|---|
| BTN  | Must Contain  | Origin  | Internal  | 
| HDBits  | Must Contain  | Ratio  | Freeleech | 
| BTN  | Must Not Contain  | Origin  | Scene | 

or something like that where the drop-down menus for the tags and values would dynamically populate with supported types based on your chosen indexer. I'm sure we could incorporate preferred logic within this as well if wanted.

jtpavlock:
Any other thoughts on this?

Taloth:
I've been meaning to start working on a reply a couple of times now, but each time I feel a huge headache coming up :smile:

I suppose I'm looking for a 'minimal viable product' approach, where we can make a step in a good direction while still giving us some wiggle room.
Problem with 'customizable' stuff is that once implemented, you can never get rid of it without breaking backward compatibility. So I'd prefer something that gives the user a little bit of control, but only a little bit.

My concern about expressions stands. So lets take a step back and get our goals listed:

- "I'd like to, at a minimum, implement a restriction based match for P2P and Scene that would be treated appropriately by all supported indexers."
- "I think the ratio bonuses are the main appeal with a minor use if you perhaps trust/favor your site's p2p groups over other sites for encoding."

You suggested indexer-granularity. I think that's a good approach. We could simply limit a Release Profile to one or more indexers. I'm not sure whether we should do that with tags, or simply have a indexer dropdown. Either is fine by me.
Note that we'd have to deal with the edge-case when the indexer gets deleted. Maybe markus agrees with adding a 'Enabled' toggle, so we can set indexer=all & disable the profile if the chosen indexer gets deleted.
Alternative is to have a new 'Condition' field that determines whether the release profile applies which would allow you to say 'indexer:btn'. We'd need markus's input on that.

Next is the tag itself, given indexer granularity we could simply use `origin:scene` in the current restriction/preference fields.
We would likely end up having some autocomplete/hint when the user types `origin:`. Paving the way for doing specific matches on parser fields (codec:x264).
We can start small, with only 'origin' and work on things like 'freeleech' later.

The alternative is a table like you suggested. It's possible, but it puts a bit more emphasis on it being indexer tags, instead of release tags. Something i'd like to avoid since I can see it evolve to matching 'properties' of a release, which isn't indexer specific at all.

I suppose we just need a coin toss between these options. I'll try to bug markus about his opinion when he wakes up.

Taloth:
Okay, talked to markus.

- We prefer to have a Indexer dropdown over a table.
The dropdown would be 'All' by default, and should allow the user to select which indexer the release profile applies to. Multi-select isn't necessary but I'm ok with having it.
The advantage is that the indexer dropdown can safely be converted to a full 'Conditions' field later, if we so desire. So it's a safe first step.
- To deal with the deleted indexer scenario we indeed like to have an Enable checkbox. Which could also aid ppl experimenting with different release profiles, so it's not for just dealing with the edge-case.
- That means the actual feature has to be implemented using the prefix style, so 'origin:scene'.
- Auto-complete would be desirable, but having the rest of the functionality takes precedence. So yes to that too, but I suggest doing that later in a separate phase.

Is that enough to answer all your questions?

jtpavlock:
> * That means the actual feature has to be implemented using the prefix style, so 'origin:scene'

comma separated list like with restrictions for this?

Taloth:
it's space separated atm iirc but yes, instead of typing 'x265' you type 'origin:scene' but otherwise it's exactly the same.

jtpavlock:
Okay, makes sense to me. I'm sure I'll have questions along the way, but I can try and work on something for now. I release this pr as it stands has manifested into something entirely different that it's original concept... should I continue to use this branch/pr or would you prefer I open something new up?

Taloth:
You can reuse the branch & pr, it has the advantage of having the entire discussion.
Once you have something to be reviewed, simply squash and force push. Then nothing of the old commits would remain.

jtpavlock:
What is the plan for conveying the feature to the user? I know you said autocomplete once they reach “origin:” but is there anything before that to hint these keys exist?

You can probably tell for the ui, but my interpretation of what you guys want is:

1. User types normal restrictions and indexer restrictions e.g. origin:scene in the current “must contain” and “must not contain” text areas
2. User specifies indexer the entire release profile, including normal restrictions and preferred words, will apply to. Defaulting to all indexers
3. User chooses to enable or disable the feature via checkbox

Please let me know if this is incorrect

jtpavlock:
Hate to bump this since I know you guys are probably busy, but I got a free week coming up I can work on this might not get too much time after I just want to hash out the requirements and make sure I'm on the right track before I do too much coding

Taloth:
Yes, I saw the notification a few days ago but didn't want to mix thinking about both PRs. :smile:

> What is the plan for conveying the feature to the user? I know you said autocomplete once they reach “origin:” but is there anything before that to hint these keys exist?

Not sure yet. Sounds like some polish for later. Windows explorer gives no autocomplete for that either, the only feedback you get is that "ext:" turns blueish.
I'd really focus on the functionality at this moment.

1. correct
2. correct
3. not entirely. The user chooses to disable the ENTIRE release profile, so not 'this feature'. This allows users to experiment with various profiles without having to rebuild them all the time.
And allows us to disable a profile if the selected indexer gets deleted.

Taloth:
Regarding the UI:

The indexer lists has to come from the user's currently configured indexers. So it's essentially an integer field containing the indexer id.
It's probably best to ask @markus101 for suggestions on how to do that dropdown properly in the React/Redux UI code. But in the meantime just use a number input.
This allows you to focus on the backend functionality.

src/NzbDrone.Core/DecisionEngine/Specifications/ReleaseRestrictionsSpecification.cs
```diff
@@ -30,37 +30,45 @@ public virtual Decision IsSatisfiedBy(RemoteEpisode subject, SearchCriteriaBase
             _logger.Debug("Checking if release meets restrictions: {0}", subject);
 
             var title = subject.Release.Title;
-            var restrictions = _releaseProfileService.AllForTags(subject.Series.Tags);
+            List<ReleaseProfile> restrictions = _releaseProfileService.AllForTags(subject.Series.Tags);
```
jtpavlock: I noticed `var` is the standard here rather than explicit typing. I changed it for personal readability as I was trying to figure out what I was working with. If you want it changed back to var I can do that.

src/NzbDrone.Core/DecisionEngine/Specifications/ReleaseRestrictionsSpecification.cs
```diff
@@ -30,37 +30,45 @@ public virtual Decision IsSatisfiedBy(RemoteEpisode subject, SearchCriteriaBase
...
 
-            var required = restrictions.Where(r => r.Required.IsNotNullOrWhiteSpace());
-            var ignored = restrictions.Where(r => r.Ignored.IsNotNullOrWhiteSpace());
-
-            foreach (var r in required)
+            foreach (ReleaseProfile restriction in restrictions)
             {
-                var requiredTerms = r.Required.Split(new[] {','}, StringSplitOptions.RemoveEmptyEntries).ToList();
+                // TODO: attach Enabled and IndexerId fields to restriction
```
jtpavlock: I think this is where I might need some help... I took a stab at it, but quickly got overwhelmed.

Taloth: Better to filter that at the start. Probably even better to have releaseProfileService do it by adding a `EnabledForTags(..., subject.Release.IndexerId)` to ReleaseProfileService  or something like it, which would then do:
```
.AllForTags(...)
.Where(v => v.Enabled)
.Where(v => !v.IndexerId.HasValue || v.IndexerId == indexerId)
.ToList()
```

To add it to ReleaseProfile itself you need a database migration, check out the `rename_restrictions_to_release_profiles` migration, you see it rename the table and add a column. You obviously only need to add two columns, bool `Enabled` with default true and int `IndexerId` with default of null (or zero).
Give it migration number 135 and add that to the top of the PR text.

src/NzbDrone.Core/DecisionEngine/Specifications/ReleaseRestrictionsSpecification.cs
Taloth: Yes, the variable name could've been better, maybe 'profiles'.

src/NzbDrone.Core/DecisionEngine/Specifications/ReleaseRestrictionsSpecification.cs
```diff
@@ -30,37 +31,83 @@ public virtual Decision IsSatisfiedBy(RemoteEpisode subject, SearchCriteriaBase
...
+                /*if (!restriction.Enabled || restriction.IndexerId != subject.Release.IndexerId)
+                {
+                    continue;
+                }*/
+
+                var required = restrictions.Where(r => r.Required.IsNotNullOrWhiteSpace());
+                var ignored = restrictions.Where(r => r.Ignored.IsNotNullOrWhiteSpace());
+
+                var keyValueRegex = new Regex(@"\b\w+:\w+\b");
```
Taloth: - `private static readonly` please, so it can be reused.
- include RegexOptions.Compiled for the same purpose.
- Add the capture groups. (\w+):(\w+) (note that \w excludes _, so .* might be more appropriate)
 Avoids you having to do that Split(':') stuff later.
- Also just match it to the full string ^...$, no need for boundary anchors coz the terms are already split by `,`.

src/NzbDrone.Core/DecisionEngine/Specifications/ReleaseRestrictionsSpecification.cs
```diff
@@ -30,37 +31,83 @@ public virtual Decision IsSatisfiedBy(RemoteEpisode subject, SearchCriteriaBase
...
+                                        foundTerms.Add(kv);
+                                    }
+                                }
+                                else
+                                {
+                                    _logger.Debug("{0} not found in release", key);
+                                }
+                                break;
+                            default:
+                                _logger.Debug("{0} is not a supported key", key);
```
Taloth: So, it doesn't match, which could mean a rejection if it's a required term.

src/NzbDrone.Core/DecisionEngine/Specifications/ReleaseRestrictionsSpecification.cs
jtpavlock: Is there something else that needs to be done after specifying the migration to get the columns to be added?

saw this, but not sure if there's some in-process method
https://fluentmigrator.github.io/articles/quickstart.html?tabs=runner-dotnet-fm

jtpavlock:
Only done basic testing with BroadcasTheNet and origin:scene, but it's a working implementation.

jtpavlock:
So I've tried the indexer dropdown feature by just emulating what's done for qualityprofiles, but I could use some help with that still as I really am just shooting in the dark with that one. 

As far as autocomplete, I feel like that may be better as you said as a phase 2 thing once everything else is handled. 

The other backend stuff is a matter of implementing some of the keys for indexers (not sure what the plan is there) and with that what other keys and values we plan on supporting as well

markus101:
Remote Path Mappings do something really similar for autocomplete on the `Host` field. The UI will need to fetch the indexers at the same time it fetches the release restrictions and a field in the restriction should be able to select the indexers for use in the select box.

Taloth:
Here you go https://github.com/Sonarr/Sonarr/commit/c1f3da00a40545bd418e269274edccffdc115c7b

The card itself is a bit ugly layout but we can work on that later.

image

Please rebase on latest phantom-develop too.

jtpavlock:
Came back to say that I'd love to see this merged someday, but I haven't looked at it in a while, and I remember being severely hindered/overwhelmed by my lack of expertise with `.NET` considering the potential size of this PR. 

I'd love some help with this if anyone is willing to pick up the pieces! 

Also, as mentioned in #2782, maybe this would be better separated into smaller chunks?

Issue #2469, opened 2018-03-23
Change label in download client after import


LinusT:
## Feature Request

Option to change label in download client after episode has been imported by SONARR.
Example: set label "sonarr" on download and "sonarr-imported" after successful import.

### Description of request and what problem are you looking to solve?

The topic on how/if Lidarr, Radarr, Sonarr should be indexer aware and help manage seed ratio and seed time rules etc has been up numerous times in multiple forums (and PR's). As has the best way to setup download clients to manage different rules.

Purpose with this feature request is that with it the state of the grab/import process has been fully communicated to the download client, thus allowing it, it's plugins or scripts to manage the seeding and cleanup details. I also allows for proper detection of downloads completely seeded (and automatically paused) but failed to import to trigger manual intervention.

LinusT:
One year later... he discovers the original feature request. Works great.

PR #3049, opened 2019-04-14
Set post-import label for deluge

jtpavlock:
#### Database Migration
NO (not sure if anything needs to be done with the database tbh)

#### Description
Allows download clients to set a label after a successful import provided that the SetLabel method is overridden (currently only done for Deluge).

#### Todos
- [ ] Tests
- [ ] Documentation

* #2469 (kind-of... only implemented for deluge currently. Not sure what other download clients have the current supporting code). 

Right now it's only setting the label after a successful auto import post-download. I'm not sure how things work if Sonarr fails to import the release automatically, and the user later manually has to import the donwload or where in the code this happens.

### DIFF
--- a/src/NzbDrone.Core/Download/Clients/Deluge/Deluge.cs
+++ b/src/NzbDrone.Core/Download/Clients/Deluge/Deluge.cs
@@ -31,6 +31,24 @@ namespace NzbDrone.Core.Download.Clients.Deluge
             _proxy = proxy;
         }
 
+        public override void MarkItemAsImported(DownloadClientItem downloadClientItem)
+        {
+            // set post-import category
+            if (Settings.TvImportedCategory.IsNotNullOrWhiteSpace() &&
+                Settings.TvImportedCategory != Settings.TvCategory)
+            {
+                try
+                {
+                    _proxy.SetTorrentLabel(downloadClientItem.DownloadId.ToLower(), Settings.TvImportedCategory, Settings);
+                }
+                catch (DownloadClientUnavailableException)
+                {
+                    _logger.Warn("Failed to set torrent post-import label \"{0}\" for {1} in Deluge. Does the label exist?",
+                        Settings.TvImportedCategory, downloadClientItem.Title);
+                }
+            }
+        }
+
         protected override string AddFromMagnetLink(RemoteEpisode remoteEpisode, string hash, string magnetLink)
         {
             var actualHash = _proxy.AddTorrentFromMagnet(magnetLink, Settings);
@@ -40,12 +58,12 @@ namespace NzbDrone.Core.Download.Clients.Deluge
                 throw new DownloadClientException("Deluge failed to add magnet " + magnetLink);
             }
 
-            if (!Settings.TvCategory.IsNullOrWhiteSpace())
-            {
-                _proxy.SetLabel(actualHash, Settings.TvCategory, Settings);
-            }
-
             _proxy.SetTorrentSeedingConfiguration(actualHash, remoteEpisode.SeedConfiguration, Settings);
+
+            if (Settings.TvCategory.IsNotNullOrWhiteSpace())
+            {
+                _proxy.SetTorrentLabel(actualHash, Settings.TvCategory, Settings);
+            }
 
             var isRecentEpisode = remoteEpisode.IsRecentEpisode();
 
@@ -69,9 +87,9 @@ namespace NzbDrone.Core.Download.Clients.Deluge
 
             _proxy.SetTorrentSeedingConfiguration(actualHash, remoteEpisode.SeedConfiguration, Settings);
 
-            if (!Settings.TvCategory.IsNullOrWhiteSpace())
+            if (Settings.TvCategory.IsNotNullOrWhiteSpace())
             {
-                _proxy.SetLabel(actualHash, Settings.TvCategory, Settings);
+                _proxy.SetTorrentLabel(actualHash, Settings.TvCategory, Settings);
             }
 
             var isRecentEpisode = remoteEpisode.IsRecentEpisode();
@@ -91,7 +109,7 @@ namespace NzbDrone.Core.Download.Clients.Deluge
         {
             IEnumerable<DelugeTorrent> torrents;
 
-            if (!Settings.TvCategory.IsNullOrWhiteSpace())
+            if (Settings.TvCategory.IsNotNullOrWhiteSpace())
             {
                 torrents = _proxy.GetTorrentsByLabel(Settings.TvCategory, Settings);
             }
@@ -250,7 +268,7 @@ namespace NzbDrone.Core.Download.Clients.Deluge
 
         private ValidationFailure TestCategory()
         {
-            if (Settings.TvCategory.IsNullOrWhiteSpace())
+            if (Settings.TvCategory.IsNullOrWhiteSpace() && Settings.TvImportedCategory.IsNullOrWhiteSpace())
             {
                 return null;
             }
@@ -267,7 +285,7 @@ namespace NzbDrone.Core.Download.Clients.Deluge
 
             var labels = _proxy.GetAvailableLabels(Settings);
 
-            if (!labels.Contains(Settings.TvCategory))
+            if (Settings.TvCategory.IsNotNullOrWhiteSpace() && !labels.Contains(Settings.TvCategory))
             {
                 _proxy.AddLabel(Settings.TvCategory, Settings);
                 labels = _proxy.GetAvailableLabels(Settings);
@@ -276,7 +294,21 @@ namespace NzbDrone.Core.Download.Clients.Deluge
                 {
                     return new NzbDroneValidationFailure("TvCategory", "Configuration of label failed")
                     {
-                        DetailedDescription = "Sonarr as unable to add the label to Deluge."
+                        DetailedDescription = "Sonarr was unable to add the label to Deluge."
+                    };
+                }
+            }
+
+            if (Settings.TvImportedCategory.IsNotNullOrWhiteSpace() && !labels.Contains(Settings.TvImportedCategory))
+            {
+                _proxy.AddLabel(Settings.TvImportedCategory, Settings);
+                labels = _proxy.GetAvailableLabels(Settings);
+
+                if (!labels.Contains(Settings.TvImportedCategory))
+                {
+                    return new NzbDroneValidationFailure("TvImportedCategory", "Configuration of label failed")
+                    {
+                        DetailedDescription = "Sonarr was unable to add the label to Deluge."
                     };
                 }
             }
--- a/src/NzbDrone.Core/Download/Clients/Deluge/DelugeSettings.cs
+++ b/src/NzbDrone.Core/Download/Clients/Deluge/DelugeSettings.cs
@@ -13,6 +13,7 @@ namespace NzbDrone.Core.Download.Clients.Deluge
             RuleFor(c => c.Port).InclusiveBetween(1, 65535);
 
             RuleFor(c => c.TvCategory).Matches("^[-a-z]*$").WithMessage("Allowed characters a-z and -");
+            RuleFor(c => c.TvImportedCategory).Matches("^[-a-z]*$").WithMessage("Allowed characters a-z and -");
         }
     }
 
@@ -43,16 +44,19 @@ namespace NzbDrone.Core.Download.Clients.Deluge
         [FieldDefinition(4, Label = "Category", Type = FieldType.Textbox, HelpText = "Adding a category specific to Sonarr avoids conflicts with unrelated downloads, but it's optional")]
         public string TvCategory { get; set; }
 
-        [FieldDefinition(5, Label = "Recent Priority", Type = FieldType.Select, SelectOptions = typeof(DelugePriority), HelpText = "Priority to use when grabbing episodes that aired within the last 14 days")]
+        [FieldDefinition(5, Label = "Post-Import Category", Type = FieldType.Textbox, Advanced = true, HelpText = "Category for Sonarr to set after it has imported the download. Leave blank to disable this feature.")]
+        public string TvImportedCategory { get; set; }
+
+        [FieldDefinition(6, Label = "Recent Priority", Type = FieldType.Select, SelectOptions = typeof(DelugePriority), HelpText = "Priority to use when grabbing episodes that aired within the last 14 days")]
         public int RecentTvPriority { get; set; }
 
-        [FieldDefinition(6, Label = "Older Priority", Type = FieldType.Select, SelectOptions = typeof(DelugePriority), HelpText = "Priority to use when grabbing episodes that aired over 14 days ago")]
+        [FieldDefinition(7, Label = "Older Priority", Type = FieldType.Select, SelectOptions = typeof(DelugePriority), HelpText = "Priority to use when grabbing episodes that aired over 14 days ago")]
         public int OlderTvPriority { get; set; }
 
-        [FieldDefinition(7, Label = "Add Paused", Type = FieldType.Checkbox)]
+        [FieldDefinition(8, Label = "Add Paused", Type = FieldType.Checkbox)]
         public bool AddPaused { get; set; }
 
-        [FieldDefinition(8, Label = "Use SSL", Type = FieldType.Checkbox)]
+        [FieldDefinition(9, Label = "Use SSL", Type = FieldType.Checkbox)]
         public bool UseSsl { get; set; }
 
         public NzbDroneValidationResult Validate()
--- a/src/NzbDrone.Core/Download/Clients/QBittorrent/QBittorrent.cs
+++ b/src/NzbDrone.Core/Download/Clients/QBittorrent/QBittorrent.cs
@@ -33,6 +33,24 @@ namespace NzbDrone.Core.Download.Clients.QBittorrent
 
         private IQBittorrentProxy Proxy => _proxySelector.GetProxy(Settings);
 
+        public override void MarkItemAsImported(DownloadClientItem downloadClientItem)
+        {
+            // set post-import category
+            if (Settings.TvImportedCategory.IsNotNullOrWhiteSpace() &&
+                Settings.TvImportedCategory != Settings.TvCategory)
+            {
+                try
+                {
+                    Proxy.SetTorrentLabel(downloadClientItem.DownloadId.ToLower(), Settings.TvImportedCategory, Settings);
+                }
+                catch (DownloadClientException)
+                {
+                    _logger.Warn("Failed to set post-import torrent label \"{0}\" for {1} in qBittorrent. Does the label exist?",
+                        Settings.TvImportedCategory, downloadClientItem.Title);
+                }
+            }
+        }
+
         protected override string AddFromMagnetLink(RemoteEpisode remoteEpisode, string hash, string magnetLink)
         {
             if (!Proxy.GetConfig(Settings).DhtEnabled && !magnetLink.Contains("&tr="))
@@ -41,11 +59,6 @@ namespace NzbDrone.Core.Download.Clients.QBittorrent
             }
 
             Proxy.AddTorrentFromUrl(magnetLink, Settings);
-
-            if (Settings.TvCategory.IsNotNullOrWhiteSpace())
-            {
-                Proxy.SetTorrentLabel(hash.ToLower(), Settings.TvCategory, Settings);
-            }
 
             var isRecentEpisode = remoteEpisode.IsRecentEpisode();
 
@@ -68,18 +81,6 @@ namespace NzbDrone.Core.Download.Clients.QBittorrent
         protected override string AddFromTorrentFile(RemoteEpisode remoteEpisode, string hash, string filename, Byte[] fileContent)
         {
             Proxy.AddTorrentFromFile(filename, fileContent, Settings);
-
-            try
-            {
-                if (Settings.TvCategory.IsNotNullOrWhiteSpace())
-                {
-                    Proxy.SetTorrentLabel(hash.ToLower(), Settings.TvCategory, Settings);
-                }
-            }
-            catch (Exception ex)
-            {
-                _logger.Warn(ex, "Failed to set the torrent label for {0}.", filename);
-            }
 
             try
             {
@@ -216,6 +217,7 @@ namespace NzbDrone.Core.Download.Clients.QBittorrent
         {
             failures.AddIfNotNull(TestConnection());
             if (failures.HasErrors()) return;
+            failures.AddIfNotNull(TestCategory());
             failures.AddIfNotNull(TestPrioritySupport());
             failures.AddIfNotNull(TestGetTorrents());
         }
@@ -288,6 +290,53 @@ namespace NzbDrone.Core.Download.Clients.QBittorrent
             {
                 _logger.Error(ex, "Unable to test qBittorrent");
                 return new NzbDroneValidationFailure(String.Empty, "Unknown exception: " + ex.Message);
+            }
+
+            return null;
+        }
+
+        private ValidationFailure TestCategory()
+        {
+            if (Settings.TvCategory.IsNullOrWhiteSpace() && Settings.TvImportedCategory.IsNullOrWhiteSpace())
+            {
+                return null;
+            }
+
+            // api v1 doesn't need to check/add categories as it's done on set
+            var version = _proxySelector.GetProxy(Settings, true).GetApiVersion(Settings);
+            if (version < Version.Parse("2.0"))
+            {
+                return null;
+            }
+
+            Dictionary<string, QBittorrentLabel> labels = Proxy.GetLabels(Settings);
+
+            if (Settings.TvCategory.IsNotNullOrWhiteSpace() && !labels.ContainsKey(Settings.TvCategory))
+            {
+                Proxy.AddLabel(Settings.TvCategory, Settings);
+                labels = Proxy.GetLabels(Settings);
+
+                if (!labels.ContainsKey(Settings.TvCategory))
+                {
+                    return new NzbDroneValidationFailure("TvCategory", "Configuration of label failed")
+                    {
+                        DetailedDescription = "Sonarr was unable to add the label to qBittorrent."
+                    };
+                }
+            }
+
+            if (Settings.TvImportedCategory.IsNotNullOrWhiteSpace() && !labels.ContainsKey(Settings.TvImportedCategory))
+            {
+                Proxy.AddLabel(Settings.TvImportedCategory, Settings);
+                labels = Proxy.GetLabels(Settings);
+
+                if (!labels.ContainsKey(Settings.TvImportedCategory))
+                {
+                    return new NzbDroneValidationFailure("TvImportedCategory", "Configuration of label failed")
+                    {
+                        DetailedDescription = "Sonarr was unable to add the label to qBittorrent."
+                    };
+                }
             }
 
             return null;
--- a/src/NzbDrone.Core/Download/Clients/QBittorrent/QBittorrentLabel.cs
+++ b/src/NzbDrone.Core/Download/Clients/QBittorrent/QBittorrentLabel.cs
@@ -1,0 +1,10 @@
+﻿using Newtonsoft.Json;
+
+namespace NzbDrone.Core.Download.Clients.QBittorrent
+{
+    public class QBittorrentLabel
+    {
+        public string Name { get; set; }
+        public string SavePath { get; set; }
+    }
+}
--- a/src/NzbDrone.Core/Download/Clients/QBittorrent/QBittorrentProxySelector.cs
+++ b/src/NzbDrone.Core/Download/Clients/QBittorrent/QBittorrentProxySelector.cs
@@ -23,6 +23,8 @@ namespace NzbDrone.Core.Download.Clients.QBittorrent
 
         void RemoveTorrent(string hash, Boolean removeData, QBittorrentSettings settings);
         void SetTorrentLabel(string hash, string label, QBittorrentSettings settings);
+        void AddLabel(string label, QBittorrentSettings settings);
+        Dictionary<string, QBittorrentLabel> GetLabels(QBittorrentSettings settings);
         void SetTorrentSeedingConfiguration(string hash, TorrentSeedConfiguration seedConfiguration, QBittorrentSettings settings);
         void MoveTorrentToTopInQueue(string hash, QBittorrentSettings settings);
         void PauseTorrent(string hash, QBittorrentSettings settings);
--- a/src/NzbDrone.Core/Download/Clients/QBittorrent/QBittorrentSettings.cs
+++ b/src/NzbDrone.Core/Download/Clients/QBittorrent/QBittorrentSettings.cs
@@ -11,6 +11,9 @@ namespace NzbDrone.Core.Download.Clients.QBittorrent
         {
             RuleFor(c => c.Host).ValidHost();
             RuleFor(c => c.Port).InclusiveBetween(1, 65535);
+
+            RuleFor(c => c.TvCategory).Matches("^[^/\\\\](((?!//)[^\\\\])*[^/\\\\])?$").WithMessage("Can not contain '\\', '//', or start/end with '/'");
+            RuleFor(c => c.TvImportedCategory).Matches("^[^/\\\\](((?!//)[^\\\\])*[^/\\\\])?$").WithMessage("Can not contain '\\', '//', or start/end with '/'");
         }
     }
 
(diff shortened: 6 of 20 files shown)


jtpavlock:
From what I can tell, QBittorent and uTorrent are the only other download clients that currently have supporting code for setting a torrent label after the torrent has already been added so implementing this feature for them as well should be fairly easy. 

I'd like some feedback on how I currently did it with Deluge first though before trying anything else.

src/NzbDrone.Core/Download/DownloadClientItem.cs
```diff
@@ -10,6 +10,7 @@ public class DownloadClientItem
         public string DownloadClient { get; set; }
         public string DownloadId { get; set; }
         public string Category { get; set; }
+        public string ImportCategory { get; set; }
```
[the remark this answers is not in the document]
jtpavlock: not used, I'll remove it

jtpavlock:
Any other issues with this?

src/NzbDrone.Core/Download/Clients/Deluge/Deluge.cs
```diff
@@ -281,6 +286,20 @@ private ValidationFailure TestCategory()
...
+            if (!labels.Contains(Settings.TvImportedCategory))
+            {
+                _proxy.AddLabel(Settings.TvImportedCategory, Settings);
+                labels = _proxy.GetAvailableLabels(Settings);
+
+                if (!labels.Contains(Settings.TvImportedCategory))
+                {
+                    return new NzbDroneValidationFailure("TvImportCategory", "Configuration of label failed")
+                    {
+                        DetailedDescription = "Sonarr as unable to add the label to Deluge."
```
markus101: `Sonarr was` (and the original typo 😄)

src/NzbDrone.Core/Download/Clients/Deluge/Deluge.cs
```diff
@@ -31,6 +31,11 @@ public class Deluge : TorrentClientBase<DelugeSettings>
             _proxy = proxy;
         }
 
+        public override void SetImportedCategory(DownloadClientItem downloadClientItem)
+        {
+            _proxy.SetLabel(downloadClientItem.DownloadId.ToLower(), Settings.TvImportedCategory, Settings);
```
markus101: Should this no-op if `TvCategory` and `TvImportedCategory` are the same? Also is the expectation that an empty category will clear the label in Deluge?

src/NzbDrone.Core/Download/Clients/Deluge/DelugeSettings.cs
```diff
@@ -26,6 +27,7 @@ public DelugeSettings()
             Port = 8112;
             Password = "deluge";
             TvCategory = "tv-sonarr";
+            TvImportedCategory = "tv-sonarr";
```
markus101: This default is going to be an issue for existing installs that aren't using the default. It's a forced behaviour change and Sonarr will start changing the category which isn't what I'd expect. We'll need a migration to avoid that.

src/NzbDrone.Core/Download/Clients/Deluge/Deluge.cs
jtpavlock: I'll fix to check if the categories are the same before setting. I think clearing the label if blank is a reasonable expectation.

src/NzbDrone.Core/Download/Clients/Deluge/DelugeSettings.cs
jtpavlock: The default makes sense to me, but I don't know what goes into migrations so up to you what you want to do here.

markus101: The migration would look like this: https://github.com/Sonarr/Sonarr/blob/phantom-develop/src/NzbDrone.Core/Datastore/Migration/033_add_api_key_to_pushover.cs

A separate class needs to be made for the JSON field to ensure a future migration doesn't break things and then some tests for validating the behaviour.

jtpavlock: I can try to make a skeleton or something, but I really won't be much use here. I don't know databases at all

Taloth: No, imo it should default to empty which should mean "don't change the category", which also means you don't need a migration.

Taloth:
@markus101 On a side note, is it possible to have the UI display advanced settings if they have been changed from the default? Rational is that I'd like ppl to see modified settings like that. It's separate from this PR of course.

src/NzbDrone.Core/Download/Clients/Deluge/DelugeSettings.cs
```diff
@@ -43,16 +45,19 @@ public DelugeSettings()
         [FieldDefinition(4, Label = "Category", Type = FieldType.Textbox, HelpText = "Adding a category specific to Sonarr avoids conflicts with unrelated downloads, but it's optional")]
         public string TvCategory { get; set; }
 
-        [FieldDefinition(5, Label = "Recent Priority", Type = FieldType.Select, SelectOptions = typeof(DelugePriority), HelpText = "Priority to use when grabbing episodes that aired within the last 14 days")]
+        [FieldDefinition(5, Label = "Post-Import Category", Type = FieldType.Textbox, HelpText = "Category for Sonarr to set after it has imported the download")]
```
Taloth: Sounds like an advanced setting to me.

src/NzbDrone.Core/Download/CompletedDownloadService.cs
```diff
@@ -115,6 +120,7 @@ private void Import(TrackedDownload trackedDownload)
             {
                 trackedDownload.State = TrackedDownloadStage.Imported;
                 _eventAggregator.PublishEvent(new DownloadCompletedEvent(trackedDownload));
+                downloadClient.SetImportedCategory(trackedDownload.DownloadItem);
```
Taloth: Architecturally, CompletedDownloadService doesn't know about download clients.
Might be good idea to move it to DownloadEventHub, which is tasked to handle the DownloadCompletedEvent (amongst others), and already has access to the download client in order to remove downloads.
It's a fairly trivial change, all pieces are already in place.

src/NzbDrone.Core/Download/IDownloadClient.cs
```diff
@@ -13,5 +13,6 @@ public interface IDownloadClient : IProvider
         IEnumerable<DownloadClientItem> GetItems();
         void RemoveItem(string downloadId, bool deleteData);
         DownloadClientInfo GetStatus();
+        void SetImportedCategory(DownloadClientItem downloadClientItem);
```
Taloth: How about renaming 'SetImportCategory' to something like 'MarkItemAsImported'. This allows various download clients to behave differently depending on their individual capabilities.

src/NzbDrone.Core/Download/Clients/Deluge/DelugeSettings.cs
jtpavlock: what if you just want to remove the label post import?

src/NzbDrone.Core/Download/CompletedDownloadService.cs
jtpavlock: I'm assuming you mean within DownloadCompletedEvent... does this method imply the release has been imported already?

src/NzbDrone.Core/Download/IDownloadClient.cs
jtpavlock: Doesn't that imply the scope of the method is broader/different than just setting the imported category? Unless the plan is to have something like 'SetImportedCategory' called under 'MarkItemAsImported' among other future features

src/NzbDrone.Core/Download/Clients/Deluge/DelugeSettings.cs
jtpavlock: fair

src/NzbDrone.Core/Download/IDownloadClient.cs
Taloth: > Unless the plan is to have something like 'SetImportedCategory' called under 'MarkItemAsImported' among other future features

Yes. The rest of Sonarr is completely unaware of the presence of Categories.
It's the download client that decides to use a Category to filter 'sonarr' downloads. So it should also be the download client that decides to use an Import Category after successful imports. 
'MarkItemAsImported' is a good generic name for the concept, which just happens to be implemented as ImportedCategory for deluge.

src/NzbDrone.Core/Download/CompletedDownloadService.cs
Taloth: No, within the [event handler in DownloadEventHub](https://github.com/Sonarr/Sonarr/blob/phantom-develop/src/NzbDrone.Core/Download/DownloadEventHub.cs#L38-L46).
DownloadCompletedEvent is a 'message' that can be raised by CompletedDownloadService and ManualImportService (Manual/Forced Import).

src/NzbDrone.Core/Download/Clients/Deluge/DelugeSettings.cs
Taloth: Originally we wanted to disallow having no category in download clients, because it offers better segregation. But we couldn't for backward compat reasons.
iirc for some download clients Sonarr actually spits out a warning if you don't have a category.
So changing from a category to none after import seems odd, and something I rather not support.

The alternative is to have a checkbox as well to enable the ImportedCategory.

The reason is that you want to avoid that non-advanced users have to change two fields just to configure their normal category. So you need a way to have a default of 'don't change category on import'

jtpavlock: good point, I'll remove the default

jtpavlock: Should I change the help text to add something like "Leaving this empty means Sonarr won't change the category" ?

Taloth: Good idea. but maybe "Set to empty to disable changing the category" or something even shorter. Just go with what reads well on the UI :smile:

src/NzbDrone.Core/Download/DownloadEventHub.cs
```diff
@@ -40,6 +40,9 @@ public void Handle(DownloadCompletedEvent message)
                 !message.TrackedDownload.DownloadItem.CanBeRemoved ||
                 message.TrackedDownload.DownloadItem.Status == DownloadItemStatus.Downloading)
             {
+                var downloadClient = _downloadClientProvider.Get(message.TrackedDownload.DownloadClient);
+                downloadClient.MarkItemAsImported(message.TrackedDownload.DownloadItem);
```
jtpavlock: Not sure if this is exactly where this should be. I figured there is not "import marking" if the item is removed from the download client

jtpavlock: Not sure if this is exactly where this should be. I figured there is not "import marking" if the item is removed from the download client

Taloth: It's where it should be, but it's probably better to invert the if. So If (Should be removed) { Remove } else { MarkItemAsImported }

But, you should put these two lines in a separate function and add the same try catch logic as with RemoveFromDownloadClient, we don't want this to throw an exception up.
Also add the Debug line, so ppl can see in the debug log that this action is taken.

src/NzbDrone.Core/Download/Clients/Deluge/Deluge.cs
```diff
@@ -31,6 +31,16 @@ public class Deluge : TorrentClientBase<DelugeSettings>
             _proxy = proxy;
         }
 
+        public override void MarkItemAsImported(DownloadClientItem downloadClientItem)
+        {
+            // set post-import category
+            if (!Settings.TvImportedCategory.Equals(Settings.TvCategory) &&
+                !Settings.TvImportedCategory.IsNullOrWhiteSpace())
```
Taloth: PS: we have a IsNotNullOrWhiteSpace.

You might want to swap the condition for readability. 
```
if (Settings.TvImportedCategory.IsNotNullOrWhiteSpace() &&
    Settings.TvImportedCategory != Settings.TvCategory)
```
Note that this changes the if to "is the function active AND is the category different", which is a bit more readable.

src/NzbDrone.Core/Download/DownloadEventHub.cs
jtpavlock: What action exactly should be specified here. That we are "marking an item as imported" ?

jtpavlock: something like ba5c28b ?

Taloth: yup, fine by me.
Users will probably not understand what's happening exactly, lacking in-depth knowledge, but at least they know some post-import action was taken.

Taloth:
I assume you tested it in production, right?

Do you have plans to implement this for the other clients that support it? Qbittorrent, rtorrent, utorrent.
Not sure if nzbget supports it, sabnzbd certainly doesn't (i wish it did).

jtpavlock:
Last tested as of 3aeb8b4cfe49694c423c7dcd7d9a115c44cd6096 but I'll test again

As far as I can tell, only UTorrent and QBittorrent have code to set a category/label after torrent creation so implementing those two won't be terribly hard I can do that. As far as the others, I'm not sure on _ability_, but regardless I don't really have plans on adding supporting label creation code if it's not already there.

jtpavlock:
So if you delete the post-import label from deluge after testing the download client, it will throw an exception on import as it's trying to set a label that doesn't exist. I'm not sure why deleting sonarr-tv (non post-import) label doesn't have the same effect though when adding a torrent as I can't find where that's handled.

Taloth:
> As far as I can tell, only UTorrent and QBittorrent have code to set a category/label after torrent creation so implementing those two won't be terribly hard I can do that.

If you're in a position to test with utorrent & qbit, then please yes.

> ... but regardless, I don't really have plans on adding supporting label setting code if it's not already there.

Understandable. Especially rtorrent is a bit tricky.
But, if you're willing to test it, then I can provide you with the necessary code in RtorrentProxy to set the label.

> I'm not sure why deleting sonarr-tv (non post-import) label doesn't have the same effect though when adding a torrent 

[Here](https://github.com/Sonarr/Sonarr/blob/phantom-develop/src/NzbDrone.Core/Download/Clients/Deluge/Deluge.cs#L74) but I'd expect an exception in both cases.

jtpavlock:
I'll have to set up the clients, but I don't see why I wouldn't be able to test it. If you give me the rtorrentproxy code I'll look at it too.

> Here but I'd expect an exception in both cases.

Right I saw that, but I was also expecting an exception. Maybe I missed it when debugging. Either way, is this something you want to handle? Could re-check/add the label in both cases before setting it

Taloth:
I haven't tested it yet, but the rtorrentproxy call should be like [this](https://gist.github.com/Taloth/55786df75d3ea475e44e7bb73a7c787c).

As for the label, I didn't want Sonarr to go add labels to deluge randomly. That's why it's only done during the test. I'm not sure what hte best behavior is while grabbing release, but for PortImport i'd just try catch and log a warning.

jtpavlock:
Let me know what you think of 310ad5c

I could also do the same for setting the label initially, and end up wrapping setting labels into a method. I also wasn't sure if you want to re-throw the exception being passed from the proxy. 

Also, I apologize if I'm being too pedantic in asking for review/assistance

jtpavlock:
So the qBittorrent works for every case I've tested _except_ if the post-import category doesn't already exist in qBittorrent. It's weird to me since setting the normal category works fine if it doesn't already exist. I can't quite figure out what's causing the difference in behavior here.

src/NzbDrone.Core/Download/Clients/Deluge/Deluge.cs
```diff
@@ -37,7 +37,15 @@ public override void MarkItemAsImported(DownloadClientItem downloadClientItem)
...
             {
-                _proxy.SetLabel(downloadClientItem.DownloadId.ToLower(), Settings.TvImportedCategory, Settings);
+                try
+                {
+                    _logger.Debug("Setting post-import label {0} in deluge", Settings.TvImportedCategory);
+                    _proxy.SetLabel(downloadClientItem.DownloadId.ToLower(), Settings.TvImportedCategory, Settings);
+                }
+                catch (Exception e)
+                {
+                    _logger.Warn(e, "Deluge failed to set label {0}", Settings.TvImportedCategory);
```
Taloth: This shows up in System->Events, so better `Failed to set Post-Import label {0} in deluge` so the user knows what it's about. Otherwise it's exactly what I meant.

Taloth:
Sonarr actually sets the category while adding the torrent to qbit, and then sets the category again (leftover from when we didn't on add). So that's probably a factor. We'll have to check the qbit documentation, fortunately they're open source too.

jtpavlock:
> Sonarr actually sets the category while adding the torrent to qbit

So including a category when add a new torrent must create it if it already exists. 

Looks like what we want then is [this](https://github.com/qbittorrent/qBittorrent/wiki/Web-API-Documentation#add-new-category) to add a new category to qbittorrent. 

I could either test if the category exists, then add when setting the category, or I could do it like in Deluge where we have a testCategory when saving the settings that creates the categories before the fact.

jtpavlock:
A word on 3390a09...

Due to the differences in qBittorrent api v1 and v2, I couldn't really implement what we do with deluge of creating the categories when initializing the download client settings.[ v1 will create a new category if it doesn't already exist when setting a category to a torrent](https://github.com/qbittorrent/qBittorrent/wiki/WebUI-API-Documentation#set-torrent-category). However, [v2 will not](https://github.com/qbittorrent/qBittorrent/wiki/Web-API-Documentation#set-torrent-category). So we don't have to worry about v1, but we do have to look at now checking/adding a new category for v2 if needed. This difference, combined with the abstraction of the qBit proxy handling both v1 and v2 meant I didn't feel I could handle the issue on a higher (i.e. not in the proxies) level.

Thus you see my attempt at implementing a check within the proxy method. It works, but I'm still seeing a `[Warn] HttpClient: HTTP Error - Res: [POST] http://localhost:8080/api/v2/torrents/setCategory: 409.Conflict` being thrown not sure if we should just ignore it since we are expecting it.

src/NzbDrone.Core/Download/Clients/QBittorrent/QBittorrent.cs
```diff
@@ -43,6 +43,19 @@ public override void MarkItemAsImported(DownloadClientItem downloadClientItem)
...
+                    {
+                        // category doesn't exist; add and try again
+                        Proxy.AddTorrentLabel(Settings.TvImportedCategory, Settings);
+                        Proxy.SetTorrentLabel(downloadClientItem.DownloadId.ToLower(), Settings.TvImportedCategory, Settings);
+                    }
+                    else
+                    {
+                        throw;
+                    }
+                }
```
jtpavlock: This outputs `[Warn] HttpClient: HTTP Error - Res: [POST] http://localhost:8080/api/v2/torrents/setCategory: 409.Conflict` since we are allowing the conflict to happen. Not sure if we should just ignore it since we are expecting it or try a different route.

Taloth:
I think you found a bug in our qbit api v2 implementation. 
But it sounds like we need to duplicate the logic that's in deluge. Creating the label during the Test.
It looks like that's doable for both v1 and v2 apis. I can take care of that if you prefer.
Either way we just need to keep the same try catch for the MarkAsImported logic like with deluge.

jtpavlock:
> But it sounds like we need to duplicate the logic that's in deluge. Creating the label during the Test.
> It looks like that's doable for both v1 and v2 apis. I can take care of that if you prefer.

Not necessarily in v1. v1 doesn't have a way to get all the torrent categories to check if one exists or not. We could just implement an empty placeholder method? Ultimately it's still fine since v1 doesn't need to do the checking, etc. to begin with since it creates a category if it doesn't exist while setting.

jtpavlock:
In the meantime, I got uTorrent working with the caveat that I can't seem to figure out how to (if possible) change/remove a label. Right now, uTorrent supports multiple labels and it simply adds the post-import label to the current label so each torrent would have 2 labels.

Taloth:
Yes, the GetCategories proxy function would throw NotSupportedException for v1, but ultimately you'd just skip the category check altogether for that api version.

Taloth:
for utorrent: the `s=label&v=[value]` query params in the proxy adds the label to the torrent, or moves it to the front. `s=label&v=`  removes the first label in the list.
So:
AddTorrentLabel: `s=label&v=[newlabel]` (add to front)
RemoveTorrentLabel: `s=label&v=[oldlabel]&s=label&v=` (move old label to front, then remove... yes multiple query params of the same name, the order of them matters in utorrent)

Call AddTorrentLabel if TvImportedcategory is not empty, to add it.
then call RemoveTorrentLabel if TvCategory is not empty, to remove it.
(In that order ensures that a crash doesn't leave the torrent in no-label)

jtpavlock:
I'll have a look at qBittorrent and uTorrent again, thanks for the help. For rtorrent, I just pushed untested code as I'm not really in a position to test if you can take that one.

jtpavlock:
Any tips to properly parse [the qbittorrent v2 get all categories call](https://github.com/qbittorrent/qBittorrent/wiki/Web-API-Documentation#get-all-categories)? I'm thinking to handle the parsing in the proxy and return a string[] of just the label names

Taloth:
It's probably better to create a QBittorrentLabel class with the 'name' and 'savePath' members. Then deserialize to Dictionary<string, QBittorrentLabel> and return that in the proxy.
Benefit is that we get the savepath for the label, which could be useful at a later date.

regarding rtorrent, give me some time and I'll spin up a docker container for it.

jtpavlock:
Currently we have

Deluge: working
qBittorrent: working
uTorrent: working
rTorrent: untested

At this point I don't have any plans to implement any other download clients

src/NzbDrone.Core/Download/Clients/QBittorrent/QBittorrentLabel.cs
```diff
@@ -0,0 +1,13 @@
+﻿using Newtonsoft.Json;
+
+namespace NzbDrone.Core.Download.Clients.QBittorrent
+{
+    [JsonObject]
+    public class QBittorrentLabel
+    {
+        [JsonProperty("name")]
```
Taloth: PS: Json serializer automatically handles 'name' vs 'Name', so the JsonProperty attribute is redundant. (JsonObject isn't needed either)

src/NzbDrone.Core/Download/Clients/QBittorrent/QBittorrentProxySelector.cs
```diff
@@ -23,6 +23,8 @@ public interface IQBittorrentProxy
 
         void RemoveTorrent(string hash, Boolean removeData, QBittorrentSettings settings);
         void SetTorrentLabel(string hash, string label, QBittorrentSettings settings);
+        void AddTorrentLabel(string label, QBittorrentSettings settings);
```
Taloth: nitpick: Just AddLabel & GetLabels since they don't apply to specific torrents.

src/NzbDrone.Core/Download/DownloadClientBase.cs
```diff
@@ -147,5 +147,7 @@ protected ValidationFailure TestFolder(string folder, string propertyName, bool
 
             return null;
         }
+
+        public virtual void MarkItemAsImported(DownloadClientItem downloadClientItem) { }
```
Taloth: Maybe we should add a _logger.Debug here to indicate that this client does not have a Post-Import action?

src/NzbDrone.Core/Download/DownloadEventHub.cs
```diff
@@ -77,5 +79,23 @@ private void RemoveFromDownloadClient(TrackedDownload trackedDownload)
...
+        {
+            var downloadClient = _downloadClientProvider.Get(trackedDownload.DownloadClient);
+            try
+            {
+                _logger.Debug("[{0}] Marking download as imported from {1}", trackedDownload.DownloadItem.Title, trackedDownload.DownloadItem.DownloadClient);
+                downloadClient.MarkItemAsImported(trackedDownload.DownloadItem);
+            }
+            catch (NotSupportedException)
+            {
+                _logger.Warn("Marking item as imported is not supported by your download client ({0}).", downloadClient.Definition.Name);
```
Taloth: Hm, I missing this in an earlier review.
We could just change this to _logger.Debug and throw new NotSupportedException in DownloadClientBase, instead of adding a _logger.Debug there.

src/NzbDrone.Core/Download/Clients/QBittorrent/QBittorrent.cs
jtpavlock: I think this stems from https://github.com/Sonarr/Sonarr/blob/52aa84e9f959108ef2c2deefa73a92a7b0ae1531/src/NzbDrone.Common/Http/HttpClient.cs#L86-L101

Really to fix this it seems we could suppress errors for calls that we know the api will return errors, but then we don't get the exception thrown. Or we could just let the logging happen further up the stack and be dealt with on a case-by-case.

src/NzbDrone.Core/Download/Clients/QBittorrent/QBittorrent.cs
```diff
@@ -33,6 +33,24 @@ public class QBittorrent : TorrentClientBase<QBittorrentSettings>
...
+            if (Settings.TvImportedCategory.IsNotNullOrWhiteSpace() &&
+                Settings.TvImportedCategory != Settings.TvCategory)
+            {
+                try
+                {
+                    Proxy.SetTorrentLabel(downloadClientItem.DownloadId.ToLower(), Settings.TvImportedCategory, Settings);
+                }
+                catch (DownloadClientException ex)
+                {
+                    _logger.Warn(ex, "Failed to set torrent label \"{0}\" for {1} in qBittorrent",
```
jtpavlock: So I am trying to clean up some of the exceptions being dealt with, and I am thinking maybe this isn't too useful? We already have the exception handled in DownloadEventHub, and there doesn't seem to be a point in handling it earlier here.

Taloth: DownloadEventHub has no knowledge about what we were trying to do: setting the import category. So you'd lose that information if you didn't catch the exception here.

jtpavlock:
Is there a reason why uTorrent specifically enforces a category?

https://github.com/Sonarr/Sonarr/blob/52aa84e9f959108ef2c2deefa73a92a7b0ae1531/src/NzbDrone.Core/Download/Clients/uTorrent/UTorrentSettings.cs#L14

I noticed some issues with setting qBittorrent categories with special characters and I think just standardizing the validation rules for the categories may be useful

src/NzbDrone.Core/Download/Clients/uTorrent/UTorrent.cs
```diff
@@ -42,29 +42,26 @@ public override void MarkItemAsImported(DownloadClientItem downloadClientItem)
...
-                {
-                    _proxy.SetTorrentLabel(downloadClientItem.DownloadId.ToLower(), Settings.TvImportedCategory, Settings);
+                _proxy.SetTorrentLabel(downloadClientItem.DownloadId.ToLower(), Settings.TvImportedCategory, Settings);
 
-                    if (Settings.TvCategory.IsNotNullOrWhiteSpace())
-                    {
-                        _proxy.RemoveTorrentLabel(downloadClientItem.DownloadId.ToLower(), Settings.TvCategory, Settings);
-                    }
-                }
-                catch (Exception ex)
```
jtpavlock: I didn't want to just catch any exception here, but I couldn't actually find a scenario where setting a label would throw an exception as uTorrent will just create a label on set if it doesn't exist. If anything goes wrong here, it's likely to be a connection to uTorrent issue that may be better served being labeled appropriately as an error in DownloadEventHub

Taloth:
Clients have varying restrictions to category names, so the validation can't be generic. Easiest is just to add the appropriate regex for qbit.

jtpavlock:
> Easiest is just to add the appropriate regex for qbit.

That was way more annoying than I anticipated

Taloth:
Ok, i've tested rtorrent. There was an error in the api call I provided. The return value of 'SetLabel' is a string rather than an integer success code.

```  
line: 49
-       int SetLabel(string hash, string label);
+       string SetLabel(string hash, string label);

line 171:
-       if (response != 0)
+       if (response != label)
```

src/NzbDrone.Core/Download/Clients/QBittorrent/QBittorrentSettings.cs
```diff
@@ -11,6 +11,9 @@ public QBittorrentSettingsValidator()
         {
             RuleFor(c => c.Host).ValidHost();
             RuleFor(c => c.Port).InclusiveBetween(1, 65535);
+
+            RuleFor(c => c.TvCategory).Matches("^[^/\\\\](((?!//)[^\\\\])*[^/\\\\])?$").WithMessage("Can not contain '\\', '//', or start/end with '/'");
```
Taloth: Fyi: you can use @"..."  to avoid some of the escapes

Taloth:
I suppose that's it, can you apply the change for rtorrent and maybe the Warn msg? Squash and rebase on latest phantom-develop and i'll merge it in.

src/NzbDrone.Core/Download/Clients/QBittorrent/QBittorrent.cs
```diff
@@ -43,22 +43,9 @@ public override void MarkItemAsImported(DownloadClientItem downloadClientItem)
...
-                    }
-                    else
-                    {
-                        throw;
-                    }
-                }
-                catch (Exception ex)
-                {
-                    _logger.Warn(ex, "Failed to set torrent label \"{0}\" for {1} in qBittorrent", 
+                    _logger.Warn("Failed to set post-import torrent label \"{0}\" for {1} in qBittorrent. Does the label exist?",
```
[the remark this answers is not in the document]
jtpavlock: My thought process is we are kind of expecting this as it happens when a user removes the label from their client after initially setting it. Stacktrace won't tell us anything we don't already know here and since it's just the warning I didn't want to clutter the log with unnecessary info. If it's an exception we aren't expecting, it will get thrown to DownloadEventHub as an error where the stacktrace will get shown.

jtpavlock:
Did you see my comments on the http code throwing the warning in qbit?

Taloth:
Had to scroll back quite a bit :smile: That only happens if the user removed the category. I'm ok with having an extra warning on that.
At one time I considered adding a SuppressHttpErrorCodes list property so the proxy can say in the httprequest that it'll handing 409. But it's too much of an hassle atm.

jtpavlock:
okay should be good to go

Dnkhatri:
> okay should be good to go

Deluge 2.0 got out of beta and just got released is this compatible with deluge 2.0

Taloth:
Please rebase on latest phantom-develop to get rid of the merge conflict.

jtpavlock:
Oops, accidentally rebased off my fork's phantom-develop XD. Should be good now

Taloth:
@jtpavlock I just realized that there's a scenario that we completely overlooked when implementing this.
Some users were talking about Post-Import and how it vastly improved their memory usages. So I wanted to warn them that Post-Import category download are not deleted by Sonarr after seeding finishes. That was the first time I realized that was actually a scenario that we didn't consider.

There are some options for that:
- Add a warning to the Post-Import field.
- For supported torrent clients, set the Remove After Ratio Limit flag on a per torrent basis.
- Periodically include the Post-Import downloads in the normal results, so they get evaluated by Sonarr.

jtpavlock:
> * Add a warning to the Post-Import field.

Easy, but now it's a compromise of functionality.

> * For supported torrent clients, set the Remove After Ratio Limit flag on a per torrent basis.

Good option depending on how many clients support this. Also whether or not seed time is supported so that functionality isn't "lost" either. I would actually prefer if this was the default option instead of Sonarr doing it even for the regular category tag if it's supported. 

> * Periodically include the Post-Import downloads in the normal results, so they get evaluated by Sonarr.

I know I'm not the common use-case, but I initially wanted this feature to "remove" torrents from Sonarr's eyes once it was done with them as I have hundreds of torrents seeding at a time and didn't want Sonarr to have to scan through them. Maybe this could be opt-in? Or just make sure Sonarr doesn't scan if the feature isn't enabled on the specified indexer.

Taloth:
> I would actually prefer if this was the default option instead of Sonarr doing it even for the regular category tag if it's supported.

And that's why we called it MarkAsImported, so we could do fancy stuff like this regardless of ImportedCategory usage. :smile:
But often there is no setting for it. Deluge has a remove_at_ratio, but not remove_at_seed_time, instead is deprioritizes such torrents.
Still it'd be good to set remove_at_ratio and ImportedCategory and then forget about it (for deluge). But other clients will need different behavior.

> I know I'm not the common use-case, but I initially wanted this feature to "remove" torrents from Sonarr's eyes once it was done with them as I have hundreds of torrents seeding at a time and didn't want Sonarr to have to scan through them.

I think we can work something out. for Deluge we're able to determine if the user's label/global setting is configured in such a fashion that Sonarr need not actively observe it. And if it is we need not query the ImportedCategory, otherwise we might be able to get away with querying it periodically.

We should probably create a new github issue for this, describe how the 4 clients currently behave and what scenario's we need to cover. Then we can see if the current architecture supports it.

jtpavlock:
> We should probably create a new github issue for this, describe how the 4 clients currently behave and what scenario's we need to cover. Then we can see if the current architecture supports it.

Good idea. Might be worth putting in the warming message while a better solution gets fleshed out as well.

Issue #2699, opened 2018-08-03
Sonarr trying to rename file over 255 characters


Adrian-at-CrimsonAuzre:
**Describe the bug**
I have a file that encompasses multiple episodes of a show, specifically the Battlestar Galactica 2005 Razor miniseries. With the naming options I have set, Sonarr fails to rename the file as it tries to put all the episode titles in the file name.
This doesn't crash Sonarr, but there's no error in the UI when the rename fails so I had trouble figuring out why it wasn't moving/renaming the file.

**Screenshots**
![image](https://user-images.githubusercontent.com/1719461/43646030-b88ea22a-9701-11e8-9071-57d4f1a70b5e.png)

**Logs**
Important part
```
18-8-3 09:42:28.2|Info|RenameEpisodeFileService|Selected episode files renamed for Battlestar Galactica (2003)
18-8-3 09:42:31.7|Info|RenameEpisodeFileService|Renaming 1 files for Battlestar Galactica (2003)
18-8-3 09:42:31.7|Error|RenameEpisodeFileService|Failed to rename file /media/TV Shows/Battlestar Galactica Complete Bluray 1080p x265/tmp/Battlestar.Galactica.Razor.Flashbacks.Complete.v3.1080p.6ch.x265.HEVC.mkv

[v2.0.0.5228] System.IO.PathTooLongException: Path is too long. Path: 
  at System.IO.File.Move (System.String sourceFileName, System.String destFileName) [0x00116] in <c5bcd0ec45b240acb20cfcfa5eee2246>:0 
  at NzbDrone.Common.Disk.DiskProviderBase.MoveFileInternal (System.String source, System.String destination) [0x00000] in <8faeb593f49341d6a7a6d2c3c281887c>:0 
  at NzbDrone.Mono.Disk.DiskProvider.MoveFileInternal (System.String source, System.String destination) [0x00076] in <cd50cb4f3a50490d80a10df172583553>:0 
  at NzbDrone.Common.Disk.DiskProviderBase.MoveFile (System.String source, System.String destination, System.Boolean overwrite) [0x000e3] in <8faeb593f49341d6a7a6d2c3c281887c>:0 
[...]

PR #3061, opened 2019-04-18
New: Limit filenames to a maximum of 255 characters

markus101:
#### Database Migration
NO

#### Description
Limits file name sections to 255 characters (accounting for season or episode folders being part of the episode naming format). 

- If the title is acceptable it is returned as-is
- If there are more than 2 episode titles, attempt to return the first and last titles with an ellipsis between them
- Otherwise try to return the first title
- Finally return a partial title

There is an edge case where only an ellipsis could be returned, but the odds of that seem pretty low.

* #2699

### DIFF
--- a/src/NzbDrone.Core.Test/NzbDrone.Core.Test.csproj
+++ b/src/NzbDrone.Core.Test/NzbDrone.Core.Test.csproj
@@ -343,6 +343,7 @@
     <Compile Include="MetadataSource\SkyHook\SkyHookProxyFixture.cs" />
     <Compile Include="NotificationTests\NotificationBaseFixture.cs" />
     <Compile Include="NotificationTests\SynologyIndexerFixture.cs" />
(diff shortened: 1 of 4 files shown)


fryfrog:
Does it matter that in ZFS the limit is actually 255 *bytes*? It equates to characters in English (maybe even UTF-8?), but I know *some* characters use more and thus the limit could be 255/2 characters.

lps-rocks:
I think this needs to be more intelligent than just a hard-coded limit since a filename / file path length limitation depends upon the filesystem.

https://en.wikipedia.org/wiki/Comparison_of_file_systems#Limits

markus101:
.net and mono both enforce the lower limit, if Sonarr tried to be smart and allow longer limits it’d get rejected later on. We’re not going to needlessly be smart about it.

Yes, the limit for ZFS is bytes not characters, but this is a good first step and we can look at cases beyond it later if the need arises.

PR #3065, opened 2019-04-19
Add airs before info for specials in metadata export

netpok:
#### Database Migration
YES

#### Description
Currently special episodes doesn't have the airs before/after information in the metadata export although once it was planned (based on the comment in the code) and it is provided by TheTVDB.

This pull request implements this feature based on the current export mechanism of Kodi:
 - If airs before/after is not provided it defaults to -1 (this is the default behaviour of Kodi)
 - If the episode has an airs before episode on tvdb it is set to that
 - If the episode has an airs after season on tvdb it is set to the provided season and episode 4096 (this is the default behaviour of Kodi)

To support this feature it is needed to be implemented on the SkyHook tvdb proxy (which is closed source), the new fields are to be added for episodes:
 - `displaySeasonNumber`: The season number 
 - `displayEpisodeNumber`: The episode number if it's an airs before record or null if airs after type

Of course if any of those fields are not present they are handled as null.

I made this pull request against the current version as it is compatible with both the v2 and v3 version of Sonarr.

#### Todos
- [ ] Implement on server side
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/src/NzbDrone.Core/Datastore/Migration/131_add_displayseasonnumber_and_displayepisodenumber_to_episodes.cs
+++ b/src/NzbDrone.Core/Datastore/Migration/131_add_displayseasonnumber_and_displayepisodenumber_to_episodes.cs
@@ -1,0 +1,15 @@
+using FluentMigrator;
+using NzbDrone.Core.Datastore.Migration.Framework;
+
+namespace NzbDrone.Core.Datastore.Migration
+{
+    [Migration(131)]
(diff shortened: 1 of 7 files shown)


netpok:
Maybe this can be also used to displaying specials in the seasons on the frontend, so one can see if it is closely coupled with season and should be acquired and watched before the following episode. But that I think is definitely a v3 feature.

markus101:
We are not accepting PRs to `develop` outside of major bugs, this will need to be based off of `phantom-develop` please.

netpok:
Opened a new one: #3073

PR #3105, opened 2019-05-08
Add Ability to search by downloadId

lps-rocks:
#### Database Migration
NO

#### Description
Add Ability to V3 history API to search by downloadId


markus101:
Is this intended for external API usage only? Can you explain the use case?

lps-rocks:
> Is this intended for external API usage only? Can you explain the use case?

This is intended for external API use only. The use case I've been able to come up with is integrating an external automation script that needs a way to correlate an infohash (downloadId) from a download client back to the activity and therefore the season/episode.

My current work-around is to list the history page by page and matching on the downloadId field. The functionality already exists in the download history code but it's just not exposed in the API. 

IMHO the API endpoints probably need a huge re-write to allow for dynamic searching of the entire structure of the model. It would enable much more flexibility for any external API access. Right now it's very heavily limited on what fields are searchable.

Issue #738, opened 2015-08-11
Add ability to move&continue seeding torrents instead of copy/hardlink


gurabli:
I recently migrated from Flexget to Sonarr, and now with Jackett, it is doing the job very nice.
Flexget is really powerful when it comes to customization, but now Sonarr is keeping up nicely with major functions already supported. What is really missing is the move downloaded torrent and continue seeding, without copy or hardlinking. 

As series name and season number are available for Sonarr with adding a tv show, it could be done the way like this, and pass these to the download client (I think Deluge is the best choice here, it handles these things exceptionally good).

An option like: Move downloaded torrents to 'download location{Series_name (from rage or tvdb)}{Season ## (from rage or tvdb)}

Example.Show.S02E02 would be moved and continued seeding at 'custom path'\Example Show\Season 2'

It is working excellent in flexget with Deluge the best, and in Transmission too. It doesn't need to be available for all download clients, if Deluge and Transmission is supported, it is already great. 

Lot's of us would like have this implemented, I'm sure, and woul also let us have proper folder structures (without hardlinks) and keeping torrents alive as they could be left to seed infinite time.

markus101:
I don't see this being something we want to support, it only works for people that aren't renaming files, which I suspect is a lower percentage of our overall users, a 1% feature as we call it as well as people with a remote torrent client (seedboxes).

Obviously copying isn't ideal for long term seeding, but what is the concern with hard linking? Why isn't it sufficient?

gurabli:
Hardlinking is fine, no problems with that, and on Linux it is working really nice. 
What I suggested is a great feature imho, but really not necessary one. I though it is rather easy to implement, as getting the series name and season number from rage/tvdb is something actually Sonarr already does, the move when done (downloading) could be just a slider to turn on or off, Deluge is very flexible with this.

Interesting to see that renaming is so frequent, as most of us I believe are using scene releases for tv shows, in that case I don't see the point of renaming, as Kodi, Plex, Emby, etc fully support proper scene naming. Obviously, I was wrong with my assumption. 

Just curios: if Deluge stops seeding the torrent, as it reached the seeding goal/time, and hardlinking is used, will it move after seeding is over, and do this automatically?

markus101:
> Just curios: if Deluge stops seeding the torrent, as it reached the seeding goal/time, and hardlinking is used, will it move after seeding is over, and do this automatically?

[...]

PR #3123, opened 2019-05-16
Add option for passing a download dir to transmission

julakali:
#### Database Migration
NO

#### Description
This adds an (advanced) option for transmission to allow downloading directly into the correct directory.
The local series directory will be mapped to remote directory (if such a mapping exists).
The season subdirectory will be generated by filenamebuilder with respect to the configured season folder name pattern.

I need this because i don't want to extract the files i download (.rar support would be helpful, too) and I don't want anyone to move them around after downloading. Just make transmission download them into the correct directories is fine and i guess this could be of use for others aswell.

I am in no way a C# expert, so I'm expecting your feedback, of course.

#### Todos
- [ ] Tests
- [ ] Documentation

* This MAY be a solution to issue #738 at least for some people

### DIFF
--- a/src/NzbDrone.Core/Download/Clients/Transmission/Transmission.cs
+++ b/src/NzbDrone.Core/Download/Clients/Transmission/Transmission.cs
@@ -7,6 +7,7 @@ using NLog;
 using FluentValidation.Results;
 using NzbDrone.Core.MediaFiles.TorrentInfo;
 using NzbDrone.Core.RemotePathMappings;
+using NzbDrone.Core.Organizer;
 
 namespace NzbDrone.Core.Download.Clients.Transmission
 {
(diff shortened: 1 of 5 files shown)


markus101:
Thanks for your contribution, but due the complications of supporting this for all clients as well as dealing with RAR releases/upgrades/etc this is not something we're planning to support. We'd be interested in a root folder based download directory for seeding files and allowing hard links into the series folder.

If you want this exact behaviour a Custom Script when a release is sent to the download client would be your best path forward.

julakali:
I can only partly understand your motivation - Isn't it better to have this option (maybe for all clients then)?

Currently, I hacked in .rar support by adding .rar in MediaFileExtensions.cs, which obviously causes other issues (such as whole seasons getting deleted).
With rar support and this download directory option, i can download the files right to the place where they belong, i don't need to move them around or make hardlinks and i can seed forever.
With Kodi (supports rar) or Plex/Emby/... with rar2fs, unpacking is not required.
I think this is the simplest approach which in consequence should be supported.

Anyway I'm okay with the decision, please keep up the good work, I'm amazed something like this even exists!

Issue #1750, opened 2017-03-10
Dynamically Alter Check For Finished Downloads interval


markus101:
With faster internet speeds and downloads finishing in seconds instead of minutes the `Check For Finished Downloads` task could run more frequently when there is work to do and slow down when the download client is idle/not as busy.

Right now it runs between 60 and 90 seconds apart, depending on when the timer ticked and how long it took to execute, so something that downloads in 20 seconds sits there for a while in addition to the queue never updating the progress (it's likely it will be imported before the queue actually updates in that case).

I'd like to see the timer tick every 5 seconds and the `Check For Finished Downloads` task would run by default every minute, if a download was sent from Sonarr to the download client we'd run it every 5 or 10 seconds, if it takes a long time to run it will be running a lot, but won't run multiple times, but won't run immediately back to back, the same thing would happen if Sonarr detected a new item in the download client queue, it would speed up until the item was complete or was removed and slow back down.

If the task takes a while to run (lot's of torrents maybe) we could increase the interval to average it out over the last X number of runs so it's running at most around 50% of the time (which would also help keep CPU usage down).

Taloth:
I think you need to be real careful with this, we already have a noticeable impact on the api of certain clients, especially those with a large amount of torrents.

LSL1337:
yeah, but if my deluge only runs a few torrents, than it can't be that big deal.
I'd love this feature, but I think still this is the most 'first world problem' I've ever seen :)

maybe make it download client setting, and if somebody wants to opt in, they can
just my 2cents

ebalders:
I would like to add that this would not only speed up addition of completed downloads, but could potentially save lots of time when downloading episodes that have several bad releases. You could have a situation where every failed download takes up to 90 seconds to update in sonarr. With some older episodes, this greatly increases the amount of time required to successfully pull the episode down.

jalmansor:
If the download client has a websocket, that seems like it would be a much better way of getting the completed status (I'm not sure how many do). Alternately if the download client has webhooks that could be another option. I personally just dislike scheduled tasks for things that could do better.

This is something that's been on my mine since upgrading to gigabit fiber last year, so I'm just tossing things out and seeing if anything sticks.

markus101:
Unfortunately there has been maybe one or two clients that use websockets and one that was purely websockets so polling is the only real way forward here and aggressively polling will cripple some clients.

Throwing this into the someday maybe pile.

winklevos:
[...]

PR #3127, opened 2019-05-19
Allow CheckForFinishedDownload and RefreshSeries to be set from the UI

tkdrob:
#### Database Migration
NO

#### Description
Allow CheckForFinishedDownloadInterval and RefreshSeriesInterval to be set from the UI. I use Sonarr with synology. Refreshing every 12 hours and checking for completed downloads every 30 seconds or a minute is excessive IMO considering I use deluge which using libtorrent (a single threaded process). I believe these are good options for people. Perhaps they should be classified as advances settings.

#### Todos
- [ ] Tests
- [ ] UI Integration

*


markus101:
Neither of these are things we intend to have configurable. Keeping series information up to date is important even if it seems useless a lot of the time and checking for completed downloads is important for keeping Sonarr in the know about what downloads are there. There is https://github.com/Sonarr/Sonarr/issues/1750 but we're not sure it's something we'll implement.

PR #3141, opened 2019-05-29
Transmission Seed Idle Limit handling

### LINKED ISSUE #2807: Sonarr ignoring idle time remove of Transmission
**Describe the bug**
In transmission you can set two OR condition to stop a torrent: seeding over limit **or** idle time over max defined.

Before remonving torrents from transmission Sonarr check that seeding is over limit. But only seeding; making idle time limit in transmission useless.

parameter in transmission is:
"idle-seeding-limit": 15,
"idle-seeding-limit-enabled": true,

This avoid torrents with no demand to remain indefinately in download queue.

**Logs**
Problem seems to be here: https://github.com/Sonarr/Sonarr/blob/0d95873a05b640eaff02deaa1f4071cde4fa0b86/src/NzbDrone.Core/Download/Clients/Transmission/TransmissionBase.cs#L101

only seedLimit is checked. I think an || should be added for Idle Time limit?

item.CanMoveFiles = item.CanBeRemoved =
                    torrent.Status == TransmissionTorrentStatus.Stopped &&
                    (item.SeedRatio >= torrent.SeedRatioLimit || item./idleTime/ >= torrent.idleTimeLimit;

Or if not possible add an option to remove as soon as stopped in client like before commit 
https://github.com/Sonarr/Sonarr/commit/2d86e44c63a3cb588663415ff110a805befee4ab   

**System Information**

 - Sonarr Version: 2.0.0.5252
 - Operating System: Unbuntu

**Additional context**
 - Transmission 2.92

Taloth:
#### Database Migration
NO

#### Description
Revises the Transmission Seed Criteria detection.

The torrent will be eligible for removal if:
- If a per-torrent or global ratio is set and reached. And the Torrent status is 'Stopped'.
- If a per-torrent idle-limit is set, and seed time (not idle-time) is larger than that. Torrent status must be 'Stopped' or 'Seeding'.
- If a per-torrent idle-limit is not set, but a global one is. Then if torrent status is 'Stopped'.

The is largely a workaround to transmission not providing the idle-limit and/or not having a seed time limit.
The per-torrent idle-limit is likely set _by_ sonarr and treated as a seed time limit. The global idle limit is not set by sonarr and treated as idle limit, but since it cannot be verified the only criteria is that it must be Stopped.

#### Todos
- [x] Tests
- [x] Documentation

Fixes #2807

### DIFF
--- a/src/NzbDrone.Core/Download/Clients/Transmission/TransmissionBase.cs
+++ b/src/NzbDrone.Core/Download/Clients/Transmission/TransmissionBase.cs
@@ -32,6 +32,7 @@ namespace NzbDrone.Core.Download.Clients.Transmission
 
         public override IEnumerable<DownloadClientItem> GetItems()
         {
+            var configFunc = new Lazy<TransmissionConfig>(() => _proxy.GetConfig(Settings));
             var torrents = _proxy.GetTorrents(Settings);
 
             var items = new List<DownloadClientItem>();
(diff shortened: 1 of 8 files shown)


src/NzbDrone.Core/Download/Clients/Transmission/TransmissionProxy.cs
```diff
@@ -164,15 +160,20 @@ private TransmissionResponse GetTorrentStatus(IEnumerable<string> hashStrings, T
...
+                "status",
+                "secondsDownloading",
+                "secondsSeeding",
                 "errorString",
                 "uploadedEver",
                 "downloadedEver",
                 "seedRatioLimit",
+                "seedRatioMode",
+                "seedIdleLimit",
+                "seedidleMode",
```
xelra: This needs to be `"seedIdleMode"` with a capital `I`.

Taloth: Fixed. Tnx for testing, of course.

src/NzbDrone.Core/Download/Clients/Transmission/TransmissionBase.cs
```diff
@@ -98,16 +99,43 @@ public override IEnumerable<DownloadClientItem> GetItems()
...
         }
 
+        protected bool HasReachedSeedLimit(TransmissionTorrent torrent, double? ratio, Lazy<TransmissionConfig> config)
+        {
+            var isStopped = torrent.Status == TransmissionTorrentStatus.Stopped;
+            var isSeeding = torrent.Status == TransmissionTorrentStatus.Seeding;
+            
+            if (torrent.SeedRatioMode == 1)
+            {
+                if (isStopped && ratio.HasValue && ratio >= torrent.SeedRatioLimit) return true;
```
markus101: Nit: Consider wrapping the return in braces. Since the lines (this and the ones below) are getting pretty long with the return on the same line.

PR #3166, opened 2019-06-13
New: Remove some more retagging groups from filenames.

desimaniac:
#### Database Migration
NO

#### Description
- Remove some more groups who re-tag/re-upload releases from filenames.
- Added test for `-RP` which was already being removed.


desimaniac:
Thank You!

PR #3167, opened 2019-06-14
Added UHD, X265, Sport and Anime categories

Fossil01:
We can probably delete NZBs.org too?

#### Database Migration
NO

#### Description
Added existing and future UHD, X265, Sport and Anime categories to NZB Finder indexer.

* Sonarr not searching in all categories


markus101:
Any PRs need to go against the v3 branch `phantom-develop` we're no longer merging commits to v2.

Sports should not be included by default, TheTVDB doesn't support them and people can add them if they want them directly in the UI. Anime has it's own configuration setting and should not be part of the normal categories.

Fossil01:
Roger. Will submit a new PR when I have time. Thanks for the feedback.

Taloth:
Anime categories don't go in the tv categories list. And I don't think we should be adding them by default.
Adding UHD & x265 is ok.

PR #3173, opened 2019-06-16
Hinted EnhancedSelectInput instead of SelectInput and Remote Path Mapping Host

Taloth:
For Remote Path Mappings i wanted to show which download clients a particular host was used for. Instead of a simply auto-complete or dropdown.

Most select inputs use the SelectInput browser control, rather than the more stylish EnhancedSelectInput.
There are some reasons to prefer SelectInput, mostly the visualisation on mobile, but I felt that switching to Enhanced globally looks good.

Before and after on desktop:

image

After for mobile:

image

Some fixes along the way:
- On Chrome you had to click twice on a option for it to be selected. (not on firefox)
- Tap outside of the modal content on mobile didn't cancel it
- hover color overrode the selected color which looked weird
- overflow wasn't handled properly in nested flex components (fix is min-width: 0)

### DIFF
--- a/frontend/src/Components/Form/EnhancedSelectInput.js
+++ b/frontend/src/Components/Form/EnhancedSelectInput.js
@@ -6,7 +6,7 @@ import classNames from 'classnames';
 import getUniqueElememtId from 'Utilities/getUniqueElementId';
 import isMobileUtil from 'Utilities/isMobile';
 import * as keyCodes from 'Utilities/Constants/keyCodes';
-import { icons, scrollDirections } from 'Helpers/Props';
+import { icons, sizes, scrollDirections } from 'Helpers/Props';
 import Icon from 'Components/Icon';
 import Portal from 'Components/Portal';
 import Link from 'Components/Link/Link';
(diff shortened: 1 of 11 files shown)


frontend/src/Components/Form/EnhancedSelectInput.js
```diff
@@ -6,16 +6,16 @@ import classNames from 'classnames';
...
+import { icons, sizes, scrollDirections } from 'Helpers/Props';
 import Icon from 'Components/Icon';
 import Portal from 'Components/Portal';
 import Link from 'Components/Link/Link';
 import Measure from 'Components/Measure';
 import Modal from 'Components/Modal/Modal';
 import ModalBody from 'Components/Modal/ModalBody';
 import Scroller from 'Components/Scroller/Scroller';
-import EnhancedSelectInputSelectedValue from './EnhancedSelectInputSelectedValue';
-import EnhancedSelectInputOption from './EnhancedSelectInputOption';
```
Qstick: Root folder select component is the only other component that uses these two, would it make sense to switch it to the new hinted value/option components and purge these? Or just build the hinted functionality into the existing components instead of making new named components?

Taloth: The Hinted one creates a specific layout that can't be overridden since it has a div with two items in it with a specific flex and style. If a future component wants to do a different layout. For example an icon. Then it wouldn't be able to properly layout the dropdown component, particularly on mobile.
That's why the base components are separate.

Issue #1886, opened 2017-05-01
SQLite on Network Share


lokkju:
Sonarr currently uses WAL mode for journals with SQLite.  WAL mode has some advantages, but one major disadvantage is that it *can not* safely be used over non-local filesystems (https://sqlite.org/wal.html); docker for windows and other virtualization systems using CIFS mounted host paths often fail with sqlite locking or corruption errors when using WAL with sqlite file on host shared paths.

Providing an option to disable WAL mode (perhaps using standard DELETE mode) for transactions would be very useful for virtualizing Sonarr, or other cases where the config files and sqlite databases need to live on a SMB/CIFS/NFS path.

Could we have some sort of config file option or command line option that disables WAL mode journaling throughout the program?

This was already addressed for OSX in #167 - may as well just make it an advanced option, so we can set it when needed, and by default on OSX.

Taloth:
> docker and other virtualization systems often fail with sqlite locking or corruption errors when using WAL with sqlite file on host shared paths.

Do you have a link for that with some evidence? I'd expect a linux docker to behave largely the same. A linux docker in a windows hyper-v that's different.
I'm not saying it can't go wrong, docker is a special animal, but I need more info to be able to determine the best course of action.

> sqlite databases need to live on a SMB/CIFS/NFS path.

It shouldn't, like never. Even other synchronization modes aren't reliable over networks, it's horrible for performance too.

> This was already addressed for OSX in #167 - may as well just make it an option.

Euh, no, rule 1 of the Fool Proof Handbook is to never add an option that the user must set to avoid breaking stuff. Either detect the edge-case and deal with it automatically, or throw a big fat warning saying it's unsupported. :smile:
For example, we might be able to detect if the db is on a known reliable fs and use wal in those cases. Or detect it's on a network or cloud drive and simply refuse to start. Or inside a docker, and force non-wal mode. Things like that.

But as I said, need more info.... please

lokkju:
Updated the description a bit.  

To be clear, as the title of the issue said, I'm only discussing docker *for windows*; which uses CIFS/SMB to mount host paths.  It mounts them with the "nobrl" option, which causes lock requests not to be sent to the server (https://github.com/docker/for-win/issues/11).  This is unique to docker for windows, though similar problems arise on docker for osx.

If your solution is that network paths are not supported for the database files, then that's fine; it just means that anyone using docker for windows will have massive problems; and perhaps a startup warning that the appdata filesystem must be local would be nice.

[...]

PR #3180, opened 2019-06-24
Add disablewal startup arg

onedr0p:
#### Database Migration
NO

#### Description
Add disablewal startup arg (useful for network filesystems) more detail in https://github.com/Sonarr/Sonarr/issues/1886

#### Todos
- [x] Tests
- [ ] Documentation

* https://github.com/Sonarr/Sonarr/issues/1886

I don't know if the PR should be closed if this is merged, however it will make all the people using network file systems happy

### DIFF
--- a/src/NzbDrone.Common.Test/EnvironmentTests/StartupArgumentsFixture.cs
+++ b/src/NzbDrone.Common.Test/EnvironmentTests/StartupArgumentsFixture.cs
@@ -55,7 +55,16 @@ namespace NzbDrone.Common.Test.EnvironmentTests
             args.PreservedArguments.Should().Be(preserved);
         }
 
-     
+        [TestCase("/disablewal", "/disablewal")]
+        [TestCase("/Disablewal", "/disablewal")]
+        [TestCase("-Disablewal", "/disablewal")]
+        public void should_preserver_disable_wal(string arg, string preserved)
+        {
+            var args = new StartupContext(new[] { arg });
+            args.PreservedArguments.Should().Be(preserved);
+        }
+
+
         [Test]
         public void should_preserver_both()
         {
--- a/src/NzbDrone.Common/ConsoleService.cs
+++ b/src/NzbDrone.Common/ConsoleService.cs
@@ -30,6 +30,7 @@ namespace NzbDrone.Common
             }
 
             Console.WriteLine("                 /{0} Don't open Sonarr in a browser", StartupContext.NO_BROWSER);
+            Console.WriteLine("                 /{0} Disable WAL", StartupContext.DISABLE_WAL);
             Console.WriteLine("                 /{0} Start Sonarr terminating any other instances", StartupContext.TERMINATE);
             Console.WriteLine("                 /{0}=path Path to use as the AppData location (stores database, config, logs, etc)", StartupContext.APPDATA);
             Console.WriteLine("                 <No Arguments>  Run application in console mode.");
--- a/src/NzbDrone.Common/EnvironmentInfo/StartupContext.cs
+++ b/src/NzbDrone.Common/EnvironmentInfo/StartupContext.cs
@@ -18,6 +18,7 @@ namespace NzbDrone.Common.EnvironmentInfo
     {
         public const string APPDATA = "data";
         public const string NO_BROWSER = "nobrowser";
+        public const string DISABLE_WAL = "disablewal";
         internal const string INSTALL_SERVICE = "i";
         internal const string UNINSTALL_SERVICE = "u";
         public const string HELP = "?";
@@ -69,6 +70,11 @@ namespace NzbDrone.Common.EnvironmentInfo
                 if (Flags.Contains(NO_BROWSER))
                 {
                     args += " /" + NO_BROWSER;
+                }
+
+                if (Flags.Contains(DISABLE_WAL))
+                {
+                    args += " /" + DISABLE_WAL;
                 }
 
                 return args.Trim();
--- a/src/NzbDrone.Core/Datastore/ConnectionStringFactory.cs
+++ b/src/NzbDrone.Core/Datastore/ConnectionStringFactory.cs
@@ -48,9 +48,22 @@ namespace NzbDrone.Core.Datastore
             connectionBuilder.CacheSize = (int)-10000;
             connectionBuilder.DateTimeKind = DateTimeKind.Utc;
 
-            connectionBuilder.JournalMode = appDataDriveType == DriveType.Network ? SQLiteJournalModeEnum.Truncate : SQLiteJournalModeEnum.Wal;
-            connectionBuilder.JournalMode = disableWal ? SQLiteJournalModeEnum.Truncate : SQLiteJournalModeEnum.Wal;
-            connectionBuilder.JournalMode = OsInfo.IsOsx ? SQLiteJournalModeEnum.Truncate : SQLiteJournalModeEnum.Wal;
+            connectionBuilder.JournalMode = SQLiteJournalModeEnum.Wal;
+
+            if (appDataDriveType == DriveType.Network)
+            {
+                connectionBuilder.JournalMode = SQLiteJournalModeEnum.Truncate;
+            }
+
+            if (disableWal)
+            {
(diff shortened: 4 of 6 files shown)


Taloth:
Can you add a IsWalSupported method to ConnectionStringFactory and add the IDiskProvider.GetMount check as suggested in the issue?

If 9P isn't detected properly as network FS then we can deal with that in ProcMountProvider globally. As well as any other edge-case.
I'd love to know what `/proc/mounts` reports for the 9P scenario (and `/proc/filesystems`), it might be handled properly out of the box.

Detecting locking capabilities as Andy suggested is a good idea but more work and doing a network FS check in IsWalSupported should get the ball rolling.

Please note that a docker container on windows _IS_ a HyperV VM. And any disk IO to the Windows host filesystem is using a network filesystem because it communicates via a HyperV vSwitch and not a HyperV attached virtual harddisk. Either CIFS, 9P or NFS or w/e, it's a remote filesystem.   
So detecting that covers most if not all of the cases mentioned in the issue.

onedr0p:
I can try that out, but I also think it would be nice to have the arg just in case it cannot determine the FS type.

Taloth:
If it comes to that then I'd prefer a config.xml option. But let's cross that bridge when we have to, ok?

onedr0p:
Sounds good, I have updated my commits with the network fs detection.

I can also try to add the options to `config.xml` more options the better :)

onedr0p:
Obviously there are some conditional problems in the logic. I would think the command line arg would take precedence. Let me know what you think. Thanks!

Taloth:
Drop the cmdline option in favor of the config.xml option, it's far less code overall since it carries over in restarts automatically. It's also something that's not specific to individual startups so doesn't belong in the cmdline.
Check out ConsoleLogLevel for an example of another 'hidden' config.xml option, maybe DatabaseJournalMode defaulting to 'Auto'. (but I'm telling ya, you won't need it, if the user runs into an old network fs we don't know about _and_ isn't listed in /proc/filesystems, we can throw it in the list)

And might I suggest making GetConnectionString non static. That gives it access to the instance. 
Store _diskProvider like you did earlier, also add IConfigFileProvider, you'll need it later.
Add GetJournalMode(string path) and call it from GetConnectionString.
Check for OsInfo.IsOsx, return Journal.
Check GetMount(path) for network fs, return Journal. (Also maybe log it (fs name and type) at debug level)
else return Wal.
(Somewhere in GetJournalMode you'd also check the config.xml option as provided by IConfigFileProvider.)
That isolates the entire logic in GetJournalMode.

Please lemme know if the logic works properly with the docker setup you had issues with.

markus101:
Users time and time again have proven they toggle things on or off, often forgetting and then wondering why something doesn't work days/weeks/months later.

Less is definitely more with config settings, that's not the power user way, but power users make up a tiny portion of the user base and any setting that has the potential to cause issues we  try to think of better ways to deal with, such as automatic detection and any time we can avoid a user having to make a change (especially an intentionally hidden setting) and have Sonarr just work is a big win. The mere 16 participants in the issue point to this particular issue having a small overall impact to users, albeit a bigger impact to the users that are affected.

I think we should avoid the option until it's absolutely required, but as Taloth said, adding additional drives to be treated as network drives can be done and will improve other aspects of the Sonarr.

onedr0p:
Makes sense to me, I also like the `config.xml` method better because we don't need to modify the `Dockerfile` at all.

I have removed the command line arg and going that way, with the network fs detection.

@Taloth 
> And might I suggest making GetConnectionString non static. That gives it access to the instance.

I have tried this but get a nasty error from TinyIOT, I am not really familiar with what it is complaining about here

```[Fatal] ConsoleApp: EPIC FAIL!

[v10.0.0.33351] TinyIoC.TinyIoCResolutionException: Unable to resolve type: NzbDrone.Core.Datastore.IDbFactory ---> TinyIoC.TinyIoCResolutionException: Unable to resolve type: NzbDrone.Core.Datastore.DbFactory ---> TinyIoC.TinyIoCResolutionException: Unable to resolve type: NzbDrone.Core.Datastore.IConnectionStringFactory ---> TinyIoC.TinyIoCResolutionException: Unable to resolve type: NzbDrone.Core.Datastore.ConnectionStringFactory ---> System.NullReferenceException: Object reference not set to an instance of an object.
   at NzbDrone.Core.Datastore.ConnectionStringFactory.IsWalSupported(String path) in C:\Users\devin\source\repos\Sonarr\src\NzbDrone.Core\Datastore\ConnectionStringFactory.cs:line 49
   at NzbDrone.Core.Datastore.ConnectionStringFactory.GetConnectionString(String dbPath) in C:\Users\devin\source\repos\Sonarr\src\NzbDrone.Core\Datastore\ConnectionStringFactory.cs:line 74
   at NzbDrone.Core.Datastore.ConnectionStringFactory..ctor(IAppFolderInfo appFolderInfo, IDiskProvider diskProvider, IConfigFileProvider configFileProvider, Logger logger) in C:\Users\devin\source\repos\Sonarr\src\NzbDrone.Core\Datastore\ConnectionStringFactory.cs:line 28
   at lambda_method(Closure , Object[] )
   at TinyIoC.TinyIoCContainer.ConstructType(Type requestedType, Type implementationType, ConstructorInfo constructor, NamedParameterOverloads parameters, ResolveOptions options) in C:\Users\devin\source\repos\Sonarr\src\NzbDrone.Common\TinyIoC.cs:line 3561```

Snippet:

```c#
namespace NzbDrone.Core.Datastore
{
    public interface IConnectionStringFactory
    {
        string MainDbConnectionString { get; }
        string LogDbConnectionString { get; }
        string GetDatabasePath(string connectionString);
        bool IsWalSupported(string path);
    }

    public class ConnectionStringFactory : IConnectionStringFactory
    {
        private readonly IConfigFileProvider _configFileProvider;

        public ConnectionStringFactory(IAppFolderInfo appFolderInfo, IDiskProvider diskProvider, IConfigFileProvider configFileProvider, Logger logger)
        {
            MainDbConnectionString = GetConnectionString(appFolderInfo.GetDatabase());
            LogDbConnectionString = GetConnectionString(appFolderInfo.GetLogDatabase());

            _diskProvider = diskProvider;
            _logger = logger;
        }

        public string MainDbConnectionString { get; private set; }
        public string LogDbConnectionString { get; private set; }


        public string GetDatabasePath(string connectionString)
        {
            var connectionBuilder = new SQLiteConnectionStringBuilder(connectionString);

            return connectionBuilder.DataSource;
        }

        public bool IsWalSupported(string path)
        {
            var driveType = _diskProvider.GetMount(path).DriveType;

            if (driveType == DriveType.Network || driveType == DriveType.Unknown)
            {
                return false;
            }

            return false;
        }

        private string GetConnectionString(string dbPath)
        {
            var connectionBuilder = new SQLiteConnectionStringBuilder();

            connectionBuilder.DataSource = dbPath;
            connectionBuilder.CacheSize = (int)-10000;
            connectionBuilder.DateTimeKind = DateTimeKind.Utc;

            if(OsInfo.IsOsx)
            {
                _logger.Debug("Mac OSX operating system detected, disabling WAL mode for SQLite");
                connectionBuilder.JournalMode = SQLiteJournalModeEnum.Truncate;
                connectionBuilder.Add("Full FSync", true);
            }
            else
            {
                if(this.IsWalSupported(dbPath) != true)
                {
                    _logger.Debug("Network filesystem store for application data detected, disabling WAL mode for SQLite");
                    connectionBuilder.JournalMode = SQLiteJournalModeEnum.Truncate;
                    connectionBuilder.Add("Full FSync", true);
                }
                else
                {
                    connectionBuilder.JournalMode = SQLiteJournalModeEnum.Wal;
                }
            }

            if(_configFileProvider.DatabaseJournalMode != "Auto")
            {
                connectionBuilder.JournalMode = SQLiteJournalModeEnum.Truncate;
                connectionBuilder.Add("Full FSync", true);
            }
            else
            {
                connectionBuilder.JournalMode = SQLiteJournalModeEnum.Wal;
            }

            connectionBuilder.Pooling = true;
            connectionBuilder.Version = 3;
           
            return connectionBuilder.ConnectionString;
        }
    }
}```

PR #3181, opened 2019-06-24
Disabling WAL on network/unknown FS, or in config.xml

onedr0p:
#### Database Migration
NO

#### Description
- Disable WAL on Network File Systems
- Disable WAL on Unknown File Systems
- Disable if `DatabaseJournalMode` is set to anything other than `Auto` in `config.xml`

#### Todos
- [x] Tests

I have tested this on my windows machine and can verify by adding:

```
<DatabaseJournalMode>AnythingButWAL</DatabaseJournalMode>
```

to config.xml, the `JournalMode` is set to `Truncate`

* #1886

I don't know if the PR should be closed if this is merged, however it will make all the people using network file systems happy

### DIFF
--- a/src/NzbDrone.Api/Config/HostConfigModule.cs
+++ b/src/NzbDrone.Api/Config/HostConfigModule.cs
@@ -1,91 +1,83 @@
-using System.IO;
-using System.Linq;
-using System.Reflection;
-using FluentValidation;
-using NzbDrone.Common.EnvironmentInfo;
-using NzbDrone.Common.Extensions;
+using Sonarr.Http.REST;
 using NzbDrone.Core.Authentication;
 using NzbDrone.Core.Configuration;
 using NzbDrone.Core.Update;
-using NzbDrone.Core.Validation;
-using NzbDrone.Core.Validation.Paths;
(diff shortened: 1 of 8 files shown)


onedr0p:
@Taloth @markus101 new PR here, let me know what you think!

onedr0p:
> Why a new PR? You don't need to push commits one by one, that triggers an alert for each commit (and they're all going to get squashed to a single commit).

Sorry about that... I re-installed windows 10 a few days ago and don't have Git set up, I was using Github web app :/

onedr0p:
I think having the setting there is a good idea, it is an alternative way of setting the journal mode. In case of `Network` and `Unknown` failing to catch whether or not to use `Truncate`, won't it make Sonarr a bit more agile on how it can be deployed and where? More and more people are using Docker and might fight this issue in the future if they want to store their Container data in a central place.

onedr0p:
Working on getting my dev env set up, sorry for the last commits.

onedr0p:
Ugh, I really messed things up. Sorry mate. Let me get my dev env set up and do it the right way....

Issue #3233, opened 2019-08-02
User configurable free space check


KenMurphy:
**Describe the problem**
Sonarr will sometimes fill up my staging drive before I can offload media to its permanent location. The free space check only fails if there is less than 100MB available. I would like to set this to leave a couple gigabytes free.

**Describe any solutions you think might work**
A size field just below the 'free space check' toggle which allows the user to set the amount of free space available before passing the check. I'm picturing something like:
Minimum free space: [1000] Megabytes
located below the 'Skip Free Space Check' option on the Media Management settings page

**Additional context**
I'm running a Cloudbox instance. My download drive is on a different partition than my local media dir. When downloading lots of media at once the local media directory can fill up before being able to offload to Google Drive (if I have already hit the daily upload limit for example). Because my local media drive is on my root filesystem this results in my system temporarily running out of disk.

I'm willing to implement this feature if a pull request is likely to be accepted for it but I wanted to check before doing the work. Are there any reasons this feature would be rejected given that it operates correctly?

markus101:
Yeah, currently the limit leaves at least 100MB free, I've discussed with Taloth and we're both open to an option to control that and definitely open for a PR with it.

Thanks for reaching out!

KenMurphy:
I've added a pull request for this issue, comments welcome. Thanks!

PR #3240, opened 2019-08-04
User configurable minimum free disk space

KenMurphy:
#### Database Migration
NO

#### Description
Allow user to configure the minimum amount of disk space available after import instead of using a fixed 100 megabyte value.

#### Todos
- [x] Tests

#3233

### DIFF
--- a/frontend/src/Settings/MediaManagement/MediaManagement.js
+++ b/frontend/src/Settings/MediaManagement/MediaManagement.js
@@ -169,6 +169,22 @@ class MediaManagement extends Component {
                         isAdvanced={true}
                         size={sizes.MEDIUM}
                       >
+                        <FormLabel>Minimum Free Space (in MB)</FormLabel>
+
+                        <FormInputGroup
+                          type={inputTypes.NUMBER}
+                          name="minimumFreeSpaceWhenImporting"
+                          helpText="Prevent import if it would leave less than this amount of disk space available"
+                          onChange={onInputChange}
+                          {...settings.minimumFreeSpaceWhenImporting}
+                        />
+                      </FormGroup>
+
+                      <FormGroup
+                        advancedSettings={advancedSettings}
+                        isAdvanced={true}
+                        size={sizes.MEDIUM}
+                      >
                         <FormLabel>Use Hardlinks instead of Copy</FormLabel>
 
                         <FormInputGroup
--- a/frontend/src/Settings/MediaManagement/MediaManagement.js
+++ b/frontend/src/Settings/MediaManagement/MediaManagement.js
@@ -169,10 +169,11 @@ class MediaManagement extends Component {
                         isAdvanced={true}
                         size={sizes.MEDIUM}
                       >
-                        <FormLabel>Minimum Free Space (in MB)</FormLabel>
+                        <FormLabel>Minimum Free Space</FormLabel>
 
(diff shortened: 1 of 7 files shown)


frontend/src/Settings/MediaManagement/MediaManagement.js
```diff
@@ -164,6 +164,22 @@ class MediaManagement extends Component {
...
+                        isAdvanced={true}
+                        size={sizes.MEDIUM}
+                      >
+                        <FormLabel>Minimum Free Space (in MB)</FormLabel>
+
+                        <FormInputGroup
+                          type={inputTypes.NUMBER}
+                          name="minimumFreeSpaceWhenImporting"
+                          helpText="Prevent import if it would leave less than this amount of disk space available"
+                          onChange={onInputChange}
```
markus101: add `unit='MB'` and remove `(in MB)` from the label.

src/NzbDrone.Core.Test/MediaFiles/EpisodeImport/Specifications/FreeSpaceSpecificationFixture.cs
```diff
@@ -68,7 +68,7 @@ public void should_reject_when_there_isnt_enough_disk_space()
         }
 
         [Test]
-        public void should_reject_when_there_isnt_enough_space_for_file_plus_100mb_padding()
+        public void should_reject_when_there_isnt_enough_space_for_file_plus_min_free_space()
```
markus101: This test fails because the default mocked result is `0`. and the 100MB file fits in the 150MB space.

thezoggy:
shouldnt your free space check be before import starts.. not after import?  since your assuming that during the xfer there wont be more space used up. which could easily lead to an issue if the user has an actively used machine/downloading on that same os..

markus101:
It is before the import starts. It's possible that more space is consumed before the file is actually completely transferred, but there is no way to deal with that and the copy/move would fail in that case.

KenMurphy:
Thanks @markus101 , I've updated my pull request with the requested fixes and one I didn't catch originally

src/Sonarr.Api.V3/Config/MediaManagementConfigModule.cs
```diff
@@ -12,7 +12,7 @@ public MediaManagementConfigModule(IConfigService configService, PathExistsValid
             SharedValidator.RuleFor(c => c.FileChmod).NotEmpty();
             SharedValidator.RuleFor(c => c.FolderChmod).NotEmpty();
             SharedValidator.RuleFor(c => c.RecycleBin).IsValidPath().SetValidator(pathExistsValidator).When(c => !string.IsNullOrWhiteSpace(c.RecycleBin));
-            SharedValidator.RuleFor(c => c.MinimumFreeSpaceWhenImporting).GreaterThan(100);
+            SharedValidator.RuleFor(c => c.MinimumFreeSpaceWhenImporting).GreaterThan(99);
```
markus101: I think `GreaterThanOrEqualTo` would be better, albeit longer.

KenMurphy: You're right, that one makes more sense. Updated my PR

markus101:
Thanks for contributing!

PR #3249, opened 2019-08-06
Added: X265 & TV Documentary categories to NZB Finder

Fossil01:
#### Database Migration
NO

#### Description
Added existing and future X265 and TV Documentary categories to NZB Finder indexer.

* Sonarr not searching in all categories

### DIFF
--- a/src/NzbDrone.Core/Indexers/Newznab/Newznab.cs
+++ b/src/NzbDrone.Core/Indexers/Newznab/Newznab.cs
@@ -44,7 +44,7 @@ namespace NzbDrone.Core.Indexers.Newznab
                 yield return GetDefinition("DrunkenSlug", GetSettings("https://api.drunkenslug.com"));
                 yield return GetDefinition("Nzb.su", GetSettings("https://api.nzb.su"));
                 yield return GetDefinition("NZBCat", GetSettings("https://nzb.cat"));
-                yield return GetDefinition("NZBFinder.ws", GetSettings("https://nzbfinder.ws", categories: new[] { 5010, 5030, 5040, 5045, 5080, 5090 }));
+                yield return GetDefinition("NZBFinder.ws", GetSettings("https://nzbfinder.ws", categories: new[] { 5010, 5030, 5040, 5045, 5090 }));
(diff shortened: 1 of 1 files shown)


markus101:
Why should these be added as defaults as opposed to users adding them if they want them?

x265 is still pretty niche and documentaries are typically named in a way that Sonarr can’t parse them anyways.

Fossil01:
Because there is a ton of stuff out there and users can always block X265 by either removing the category or in their settings.

I'd say that the majority of the users have no clue about adding categories themselves. 

Documentary names:
Hatayasot.2017.DOCU.E02.720p.HDTV.x264-Sweet-Star
Dayan.The.First.Family.DOCU.S01E04.720p.HDTV.x264-HebTV
National.Geographic.Life.On.Mars.WS.PDTV.XviD-hV

jalmansor:
> Because there is a ton of stuff out there and users can always block X265 by either removing the category or in their settings.
> 
> I'd say that the majority of the users have no clue about adding categories themselves.
> 
> Documentary names:
> Hatayasot.2017.DOCU.E02.720p.HDTV.x264-Sweet-Star
> Dayan.The.First.Family.DOCU.S01E04.720p.HDTV.x264-HebTV
> National.Geographic.Life.On.Mars.WS.PDTV.XviD-hV

I agree with @Fossil01 on this, as I've also seen where categories aren't necessarily perfect. Plus, x265 seems to have improved significantly in its support and should be worth adding.

Though, for the sake of argument, could a better solution for documentaries be a new series type? I haven't looked at the code to see exactly how viable that is, so I could be completely off base.

Fossil01:
Truth be told Documentaries as less of concern. So if you like X265 only thats fine too.

thezoggy:
and most of those docu wont be pulled by sonarr in the first place due to naming conflicts with scrapers/unsupported format.

thezoggy:
additionally.. more categories you add means you will hit max rss feed entries. which will roll stuff off quicker.. which means you need to be either more aggressive with the hits so you dont miss stuff. this is why the 'TV' category just isnt used (also you want to eliminate the noise of stuff you just would never want).

PR #3253, opened 2019-08-08
New: OnHealthIssue Notification

### LINKED ISSUE #1798: Notifications on health check warnings/errors
This feature would add notification about like proxy, download client and Indexers errors to connection notifications.

healthcheck

Qstick:
#### Database Migration
YES (Migration 135: Add OnHealthIssue column to notification table)

#### Description
Adds notification option for onHealthIssue, also adds option for toggle warnings in addition to errors (can remove this if wanted).

Warning are only triggered once, not at every health check task. Once a warning/error is cleared it will get triggered again if it returns. 

Pulled from Lidarr implementation

#### Todos
- [x] Tests

Fixes #1798

### DIFF
--- a/frontend/src/Settings/Notifications/Notifications/EditNotificationModalContent.js
+++ b/frontend/src/Settings/Notifications/Notifications/EditNotificationModalContent.js
@@ -42,10 +42,13 @@ function EditNotificationModalContent(props) {
     onDownload,
     onUpgrade,
     onRename,
+    onHealthIssue,
     supportsOnGrab,
     supportsOnDownload,
     supportsOnUpgrade,
     supportsOnRename,
+    supportsOnHealthIssue,
+    includeHealthWarnings,
     tags,
     fields,
     message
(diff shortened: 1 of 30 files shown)


Qstick:
Where do you guys want to go with this now that there are 3 types (notice, warning, error). Just let me know how you want to proceed.

Taloth:
How about put the status of the last notifications or 'active warnings' in logs.db to prevent re-sending them on restart? Could potentially be used to aggregate regular notifications in the future.
I'm ok with having it in the main db too, of course.

Taloth:
I also talked with markus about that if we're adding all these event/notification types, then maybe they no longer should be primary fields. Having a grayed out 'On Health Check' doesn't make much sense for the kodi and plex connectors as they're primarily meant to update libraries. (and hopefully sync watched states somewhere in the future).   

I suppose it'd become desirable to refactor those events to a list of events for which notifications should be sent to the particular connector. That way any connector that supports notifications would naturally support all of them.     
It's not something that needs to be done now, but some food for thought.

Qstick:
Thanks for fixing this up and pushing @Taloth

PR #3254, opened 2019-08-08
 New: Replace SharpRaven with new Sentry SDK

Qstick:
#### Database Migration
NO

#### Description
Remove obsolete SharpRaven package and replace with new Sentry Dot Net package 
URL: https://github.com/getsentry/sentry-dotnet

This also pulls over a few preference items from Lidarr, feel free to request any changes to setup
- 200 Breadcrumbs (last 200 lines of log file)
image

- Filtering of certain events (we are filtering many "User Error" events related to permissions, corrupt DBs, Jackett, etc.

- User hash for UI events (same as back end, so that events can be identified that come from same user )

#### Todos
- [x] Tests

None


Taloth:
`System.Collections.Immutable, Version=1.2.3.0`. netstandard stuff.

I'm dealing with that right now with the switch to PackageReference and it's affecting a few things, particularly for the bindingRedirects.
I actually went further, switching to Microsoft.NET.Sdk style csproj files, that seems to fix the bindingRedirect issues.

If you can hold off for the weekend then I can see whether it all works.

I do like having a different sentry lib, coz SharpRaven had some nasty dependencies that required hacks on mono. Will need to check if the new sdk doesn't have the same issue.

Qstick:
Just let me know on anything changed. PackageRef format is nice, we popped over about a month ago in nightly, had to update a few back end dependencies.

Qstick:
Just let me know on anything changed. PackageRef format is nice, we popped over about a month ago in nightly, had to update a few back end dependencies.

src/NzbDrone.Common/Instrumentation/Sentry/SentryTarget.cs
```diff
@@ -4,52 +4,123 @@
...
-        {
-            {LogLevel.Debug, ErrorLevel.Debug},
-            {LogLevel.Error, ErrorLevel.Error},
-            {LogLevel.Fatal, ErrorLevel.Fatal},
-            {LogLevel.Info, ErrorLevel.Info},
-            {LogLevel.Trace, ErrorLevel.Debug},
-            {LogLevel.Warn, ErrorLevel.Warning},
+        // use string and not Type so we don't need a reference to the project
+        // where these are defined
+        private static readonly HashSet<string> FilteredExceptionTypeNames = new HashSet<string> {
```
Taloth: @markus101 Do we really want to block any log reporting for these?
Iirc we already rate-limited boot loops a bit.

Things like TinyIoCResolutionException could be a deployment issue we'd want to know about.

It might be better if we could make something survive the reboot and block the same message from being reported again. file in appdata, tmp, anything really.

markus101: Good point, `TinyIoCResolutionException` is a great example of something that could have gone wrong in a deploy (missed binary or something missing on the platform), it's unlikely it'd get deployed with the testing in place, but we don't cover every base.

We might be able to do rate limiting on the sentry side, I know you did some work to avoid constant restarts if the app fails to load, would that cover us and anything else we rate limit? Not opposed to your suggested approach to avoid reporting the same thing over and over.

Qstick:
Closing for #3268

Issue #2973, opened 2019-02-26
Replacement search fails for some failed downloads


hr1232:
**Describe the bug**
If Sonarr downloads a multi-episode release (or a season release) which is broken, the search for an alternative release fails. The error happens as follows:

1) Sonarr downloads a release (e.g. "Star Trek The Next Generation S07E25E26 720p HDTV x264 F1rebladerunner").
2) Sonarr lists the two contained episodes (Start Trek TNG S07E25 and S07E26) seperately in its queue while the download is in progress. Both queue entries are linked to the same download.
3a) (automatic post-processing) "Check For Finished Download" task finds the download is failed. Sonarr post-processes S07E25 first, removes it from the queue and blacklists the release. S07E26 stays in the queue.
3b) (manual post-processing) User removes and blacklists the release manually because it has been marked for manual post-processing (for whatever reasen). S07E26 stays in the queue.
4) In both cases, the blacklisting of the release starts a search for alternatives, which ignores all results because "release in queue is of equal or higher quality". This is because S07E26 is still listed in the queue, even though the download has already been removed.
5) After 1 minute, Sonarr starts the "Check For Finished Download" task again and silently removes S07E26 because the release has been removed from the downloader.
6) No alternative release will ever be searched automatically. Sonarr will needlesly wait around until the user searches manually or a new release comes up through RSS.

The bug happens always, if a release has been marked for manual post-processing and the user handles it manually. User can evade it by manually post-processing the second release before the "Check For Finished Download" task starts again. For multi-episode releases this might be a workaround, for season-releases its impossible to process 20-30 episodes before they are silently.

For releases that can be handled automatically it is highly dependend on the processing speed of the server and the number of releases the indexer returns for the first episode. If post-processing of the second episode can start before it is silently removed, an alternative release will be downloaded. Espiecially with season-releases of popular searies with many releases, Sonarr has no chance to get to the last episode before it is silently removed.

**Expected behaviour**
Remove all queue entries related to one release during the first post-processing (should be easy because the release name is the same). After that, the search for alternative releases for the first episode will not fail anymore because there are no zombie entries left in the queue.

**System Information**
 - Sonarr Version: 2.0.0.5301
 - Operating System: Linux 4.15.0-43 in Docker (linuxserver/sonarr)
 - mono Version: 5.18.0.240

hr1232:
[...]

Issue #1710, opened 2017-02-23
Ability to ignore episodes from queue page in the download client (will not be imported)


beeblebrox:
I noticed some torrents were remaining in the Activity list.  It turns out that I already have resulting episodes imported of equal or better quality.  This results in this log `07-2-23 03:48:10.7|Debug|UpgradeSpecification|This file isn't an upgrade for all episodes. Skipping <...>`

At this point the torrent seems to get stuck under activity and I don't think it ever gets cleaned up when seeding is done either.  Shouldn't the import logic go an extra step and mark it completed so it is removed from activity and move on?  In this situation, it seems like my only option is to click remove from download client on the activity list?  That isn't desirable.

What are your thoughts?

markus101:
Since they aren't imported Sonarr doesn't see them as done and wouldn't do anything with them. We may need a way to manually set it as complete in that case, what do you think @Taloth? Would either need something new or a way to add it to history hidden (could also tie into our clearing of history, but really it's just hidden).

Taloth:
It's basically an exceptional situation that requires user-intervention. I dunno if it offer a manual/force import, it probably should if it doesn't. The option is to force import (overwriting the non-upgrade file), or to remove the download. I don't think Sonarr should make a decision for the user.

markus101:
I believe it will show manual import there, but there isn't a way to not import and ignore it, leaving the only option to delete it from the download client, change the category or import it and overwrite whatever quality it there.

beeblebrox:
Changing the category was a very simple workaround, thanks @markus101!

The manual import is not so easy. I had over 80 items trigger this at once, and you can only do the import for a single item at a time.  ff it really was lower quality I don't want the import anyway.  

If I could make a request it would be marking the item in activity to make it clear that it is failing to import and requires manual intervention. Perhaps even a tooltip with the workaround in it? 

Here is my usecase in which this will probably happen again:  I did a search for missing and sonarr grabbed a ton of things thinking it was better quality (I guess?) the importer disagreed. I do this if my server has a disruption for a period of time which prevents Sonarr from operating.

markus101:
> If I could make a request it would be marking the item in activity to make it clear that it is failing to import and requires manual intervention. Perhaps even a tooltip with the workaround in it?

The icon on the left should have the information why it's not importing, it doesn't explain how to fix it, but should point you in the correct direction.

[...]

Issue #907, opened 2015-10-30
New Downloads are not tracked when CDH is importing from a remote location


markus101:
This issue/borderline bug causes issues when CDH is importing a download and a new release is queued, Sonarr has the history event for it, but the queue is not updated to reflect its existence because CheckForFinishedDownloads has paused the debouncer and has asked CDH to import a release. This would also cause delays if FDH is searching for a better release.

The proposed solution is to separate checking and processing to separate commands, if no check has completed the processor would have nothing to do, but would try again the next time the task executes, using a concurrent queue similar to the Command queue we should be able to only queue releases that we aren't already tracking, we'll need to make sure that an item thats removed from the queue for processing isn't reprocessed the next time it runs and causes issues (if its already been imported, but the tracked download doesn't reflect that).

markus101:
The task queue improvements work is also going to impact this, the result being a manual import will prevent Check for Finished Downloads from executing. Another reason to separate the queue syncing aspect and the import/failed download handling aspect.

Issue #668, opened 2015-07-08
CDH: Don't mark as imported if there are other valid files


markus101:
If there is more than one file and only one is imported the whole release shouldn't be considered imported and subsequently nuked from Activity: Queue (this applies to automatic and manual imports).

PR #3262, opened 2019-08-15
Improve download processing

markus101:
#### Database Migration
NO

#### Description
This improves our download handing in a few different ways:
- Allows communication with download clients while imports are in progress
- Doesn't treat a download as imported until _all_ expected files have been imported, this could be taken a bit further and not show already imported downloads in the queue as well, but this is getting big already
- Failed downloads that are searched for shouldn't result in failures because the queue hasn't been updated (a race between the search results being processed and the queue updating) as failed items are marked as failed and will be skipped.

In addition users can now ignore downloads and block Sonarr from processing them further, which ties in with the second point. Ignoring results in a history event being added so the user can see it happened and we have something to filter the tracked downloads with. I'm not in love with this approach, but I felt something was necessary to allow removal of completed downloads that are partially imported for a known reason.

#### Todos
- [X] Tests
- [X] Documentation

* #907
* #2973
* #1710
* #668


src/NzbDrone.Core/Download/TrackedDownloads/TrackedDownloadAlreadyImportService.cs
```diff
@@ -0,0 +1,37 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+using NzbDrone.Common.Extensions;
+using NzbDrone.Core.History;
+
+namespace NzbDrone.Core.Download.TrackedDownloads
+{
+    public interface ITrackedDownloadAlreadyImportService
```
Taloth: Don't think this is really a 'service'. ITrackedDownloadAlreadyImported should do it right?

src/NzbDrone.Core/Download/TrackedDownloads/TrackedDownloadService.cs
```diff
@@ -158,7 +174,7 @@ public void UpdateTrackable(List<TrackedDownload> trackedDownloads)
         }
 
         private void LogItemChange(TrackedDownload trackedDownload, DownloadClientItem existingItem, DownloadClientItem downloadItem)
-	{
+	    {
```
Taloth: stray tab in original code

src/NzbDrone.Core/Download/ProcessMonitoredDownloadsCommand.cs
```diff
@@ -2,7 +2,7 @@
 
 namespace NzbDrone.Core.Download
 {
-    public class CheckForFinishedDownloadCommand : Command
+    public class ProcessMonitoredDownloadsCommand : Command
```
Taloth: This breaks the api.

src/Sonarr.Api.V3/Queue/QueueResource.cs
```diff
@@ -55,8 +57,9 @@ public static QueueResource ToResource(this NzbDrone.Core.Queue.Queue model, boo
                 Sizeleft = model.Sizeleft,
                 Timeleft = model.Timeleft,
                 EstimatedCompletionTime = model.EstimatedCompletionTime,
-                Status = model.Status,
+                Status = model.Status.FirstCharToLower(),
                 TrackedDownloadStatus = model.TrackedDownloadStatus,
+                TrackedDownloadState = model.TrackedDownloadState,
```
Taloth: wouldn't we need FirstCharToLower here too, and secondly, does this break the api? (clients might be case sensitive)

src/NzbDrone.Core/Download/ProcessMonitoredDownloadsCommand.cs
markus101: In what sense? The rename is for clarity, which makes sense I think.

Taloth: Coz third parties can publish the command via the api. Granted it's not very likely in this particular case, but I think we should maintain backward compat.

src/Sonarr.Api.V3/Queue/QueueResource.cs
markus101: Not in this case because enums are handled automatically, but status is a string.

Just the v3 beta if they are case sensitive, which shouldn't be a problem.

src/NzbDrone.Core/Download/ProcessMonitoredDownloadsCommand.cs
markus101: Re-added the old command and will log a deprecated message (warning) is someone actually calls it.

markus101:
I've finally addressed the comments already provided, another review so we can get this landed would be great.

src/NzbDrone.Core/DecisionEngine/Specifications/QueueSpecification.cs
```diff
@@ -42,6 +43,16 @@ public Decision IsSatisfiedBy(RemoteEpisode subject, SearchCriteriaBase searchCr
                 var qualityProfile = subject.Series.QualityProfile.Value;
                 var languageProfile = subject.Series.LanguageProfile.Value;
 
+                // If the item has already failed ignore it, it'll be cleaned up automatically later.
+                // To avoid a race make it's it's not FailedPending (failed awaiting removal/search)
+                // or Failed (removed/searching) though the latter should be removed automatically.
+
+                if (queueItem.TrackedDownloadState == TrackedDownloadState.FailedPending ||
+                    queueItem.TrackedDownloadState == TrackedDownloadState.Failed)
```
Taloth: 'Failed' doesn't happen since DownloadIsTrackable function already rejects those. 
FailedPending are listed, but I don't think that this check is needed.

Taloth:
Didn't check the tests, but I think I made it through the code itself.

src/Sonarr.Api.V3/Queue/QueueStatusModule.cs
```diff
@@ -47,10 +48,10 @@ private QueueStatusResource GetQueueStatus()
                 TotalCount = queue.Count + pending.Count,
                 Count = queue.Count(q => q.Series != null) + pending.Count,
                 UnknownCount = queue.Count(q => q.Series == null),
-                Errors = queue.Any(q => q.Series != null && q.TrackedDownloadStatus.Equals("Error", StringComparison.InvariantCultureIgnoreCase)),
-                Warnings = queue.Any(q => q.Series != null && q.TrackedDownloadStatus.Equals("Warning", StringComparison.InvariantCultureIgnoreCase)),
-                UnknownErrors = queue.Any(q => q.Series == null && q.TrackedDownloadStatus.Equals("Error", StringComparison.InvariantCultureIgnoreCase)),
-                UnknownWarnings = queue.Any(q => q.Series == null && q.TrackedDownloadStatus.Equals("Warning", StringComparison.InvariantCultureIgnoreCase))
+                Errors = queue.Any(q => q.Series != null && q.TrackedDownloadStatus.HasValue && q.TrackedDownloadStatus.Value == TrackedDownloadStatus.Error),
```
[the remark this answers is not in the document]
markus101: I'm not sure I did, but makes sense, I'll fix. Thanks!

src/NzbDrone.Core/History/HistoryService.cs
```diff
@@ -307,6 +303,29 @@ public void Handle(EpisodeFileRenamedEvent message)
...
+                                  SeriesId = message.SeriesId,
+                                  EpisodeId = episodeId,
+                                  DownloadId = message.DownloadId,
+                                  Language = message.Language
+                              };
+
+                history.Data.Add("DownloadClient", message.DownloadClient);
+                history.Data.Add("Message", message.Message);
+
+                _historyRepository.Insert(history);
```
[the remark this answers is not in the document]
markus101: Good call, will also prevent against a partial update (along the lines of your next comment).

src/NzbDrone.Core/DecisionEngine/Specifications/QueueSpecification.cs
markus101: This is is for https://github.com/Sonarr/Sonarr/issues/2973 it's rare, but does happen.

Taloth: Failed items never appear in the queue coz you actively filter those out here https://github.com/Sonarr/Sonarr/pull/3262/files#diff-5ab3194831f90f6cba9326ea9dd65a2aR174
Thing is that the queue is a copy, so when you set TrackedDownload.State == FailedPending, those items will still be as Downloading in the queue till he event is raised (https://github.com/Sonarr/Sonarr/pull/3262/files#diff-5ab3194831f90f6cba9326ea9dd65a2aR76).
In ProcessFailed you set the Failed state, but that isn't propagated yet either. You immediately raise the (async) event that'll lead to the search.
Only at the end of ProcessMonitoredDownloadsCommand  you propagate those states to the queue.
So yes, it's very likely that the queue is updated before the search finishes, but it isn't guaranteed.

I think the QueueSpecification should be 'Temporary', this would cause those releases to be added to the pending/fallback queue instead of being dropped entirely. Then the timing doesn't matter much anymore since the next rsssync would pick them up.

markus101: `FailedPending` is applied to the queue when https://github.com/Sonarr/Sonarr/pull/3262/files#diff-5ab3194831f90f6cba9326ea9dd65a2aL75 executes (and the queue updates). That happens before the items are removed from the queue entirely (and replacements searched for) and happens synchronously so the search doesn't happen before the internal queue state is applied.

You're right about `Failed` not making it this far though.

> I think the QueueSpecification should be 'Temporary'

I can see how it'd help here, but worried that doing so would result in something being removed from the queue manually by the user and suddenly another release shows up, maybe it's not an issue with our history spec. I won't worry about it here, but we can discuss and decide one way or another.

Taloth:
I'd say, ship it. it should work fine as is

hergonoway:
Just THANK YOU, from a seedbox user.

PR #3270, opened 2019-08-19
Phantom diagnostic script

Taloth:
_base set to csproj-migration for easier comparison, should be changed to phantom-develop later._

#### Description
A bit of a hobby project. I wanted to be able to run some tests remotely, which usually involves spitting out custom binaries. With CSharpScript I could run simple or complex scripts remotely via the api.
With monaco-editor I could expose it to the ui even with an editor that can at least show compile errors inline.

The feature is hidden behind the ./diag UI route. Which is hidden by default.
The route exposes a 'status' page, which atm shows cpu, memory and sqlite db sizes.
And a second route that exposes the script console.

The script console is disabled by default for security reasons, a directory 'debugscripts' must exist in the appdata folder. (The idea was to later save user-defined scripts there.)
I'm also thinking about the idea to define a couple of maintenance scripts, either via skyhook/services or embedded, that can be called 'safely'.

The first commit revises how webpack generates the bundles. It properly splits up chunks between index & vendor now (without duplication).
And most importantly, it has a separate lazy-loaded monaco-editor bundle that doesn't get loaded till it's needed, for performance reasons.

The whole react/redux stack is new territory for me, so please review that.


frontend/src/preload.js
```diff
@@ -1,2 +0,0 @@
-/* eslint no-undef: 0 */
-__webpack_public_path__ = `${window.Sonarr.urlBase}/`;
```
ta264, line 2: I think you need to keep preload to get the fuse worker to load properly when `urlBase` is defined.  IIRC it's something to do with the `__webpack_public_path__` not getting picked up correctly otherwise due to the js load order.

I had to restore it in #3430 after it was dropped in the previous webpack PR.

Taloth: I recall needing this to make monaco worker load properly. But I'm not planning on merging this in anytime soon. I just need it for debugging something at catalina :)

Taloth: Yeah, I used the inline option in the worker-loader so it doesn't need to pull in a separate worker js, it just loads from the bundle.

PR #3276, opened 2019-08-25
Removed curlsharp and revised webpack and sentry logic

Taloth:
#### Description
I discovered that Sentry would not be able to send events on certain mono setups. I included the RuntimeInformation Facade from mono 5.4, which still works on 6.0
Sentry was also still being used with analytics off, I fixed that.
The rest there is changes in the tags that are submitted.

I also included the webpack bundling revision (from another branch).


src/NzbDrone.Core.Test/HealthCheck/Checks/DotnetVersionCheckFixture.cs
```diff
@@ -0,0 +1,57 @@
...
+        [TestCase("4.7")]
+        [TestCase("4.7.1")]
+        public void should_return_notice(string version)
+        {
+            GivenOutput(version);
+
+            Subject.Check().ShouldBeNotice();
+        }
+
+        public void should_return_warning(string version)
```
ta264: Only minor but this test has no cases. I guess you're keeping the warning option around in case you want to use it later? Not sure it's used at the moment?

Taloth: yup, at first I produced warnings for it, but the notice shows as blue in the UI so draws enough attention. I kept the function around for later use. It's mostly that I want to ramp up the pressure to upgrade, notice now, warning later.

Taloth:
ReconfigureLogging.HandleAsync(ConfigFileSavedEvent) handles that part. Otherwise changing the log level wouldn't apply immediately either :)

ta264:
But as far as I can tell `ReconfigureLogging.HandleAsync(ConfigFileSavedEvent)` never calls `UnregisterRemoteLoggers` - that only ever happens on `ApplicationStartedEvent`.  So the log level changes but not the sentry part I don't think.

PR #3288, opened 2019-08-29
Upgrade Nancy to 2.0.0

ta264:
#### Database Migration
NO

#### Description
Upgrading Nancy to 2.0.0 was the main chore for making Lidarr dotnet core compatible.  I have a similar PR ready to go on Lidarr once we get 0.7 out but it'd be great if you guys could review it to make sure I've not done something stupid with passing all the `INancyEnvironment` objects around.  Unfortunately it seemed unavoidable.  

As a quid pro quo for the review I thought I'd port the change across to Sonarr :)

The changes in `Sonarr.Api.V3` and `NzbDrone.Api` are almost all generated by sed and aren't very exiciting:
- Add in a `using Nancy.Configuration;`
- Get an instance of `INancyEnvironment` and pass it through to the base constructor
- Update the routing format

#### Todos
- [ ] Tests


src/Sonarr.Http/TinyIoCNancyBootstrapper.cs
```diff
@@ -52,6 +53,47 @@ protected override sealed INancyEngine GetEngineInternal()
             return this.ApplicationContainer.Resolve<INancyEngine>();
         }
 
+                //
```
Taloth: Don't think the comments are useful here.

src/Sonarr.Http/Extensions/ReqResExtensions.cs
```diff
@@ -32,9 +32,9 @@ public static object FromJson(this Stream body, Type type)
             return Json.Deserialize(value, type);
         }
 
-        public static JsonResponse<TModel> AsResponse<TModel>(this TModel model, HttpStatusCode statusCode = HttpStatusCode.OK)
+        public static JsonResponse<TModel> AsResponse<TModel>(this TModel model, NancyContext context, HttpStatusCode statusCode = HttpStatusCode.OK)
```
Taloth: I don't think we need this anymore. The other place appears ToErrorModel, and there we can just return the ErrorModel instead of as response.

src/Sonarr.Http/Extensions/NancyJsonSerializer.cs
```diff
@@ -1,22 +1,23 @@
...
 using NzbDrone.Common.Serializer;
 
 namespace Sonarr.Http.Extensions
 {
     public class NancyJsonSerializer : ISerializer
     {
-        public bool CanSerialize(string contentType)
+        public bool CanSerialize(MediaRange contentType)
         {
             return true;
```
Taloth: We need to check the content type here, coz in my tests we return json when xml is explicitly requested, and even return Content-Type: application/xml... with a json body.

Taloth:
Closed via 90fb1646e0021ad56d10ca863ecfea0e1a4ff65a

A few additional changes, but squashed.

PR #3292, opened 2019-09-04
Allow digits in Deluge categories

rbraunschweig:
#### Database Migration
NO

#### Description
Deluge's Label plugin allows digits, but the client's category fields in Sonarr do not.


Taloth:
Looks good, merged it in. Not sure why ppl need digits, but can't hurt right?

PR #3298, opened 2019-09-09
Update unity

ta264:
#### Database Migration
NO

#### Description
Update unity from 2.1 to 5.11.  It's only used in tests so hopefully not too controversial.

#### Todos
- [x] Tests


Taloth:
We should just use a different injection framework there.

Issue #2296, opened 2017-11-19
Memory usage climbs over time under mono


smcpeck:
## Bug Report
I'm running all of my home media automation apps on a 2GB VM and get memory alerts from it about every other day.  NzbDrone starts getting to the point of eating up a lot of memory that isn't getting released back to the system.

### System Information/Logs

**Sonarr Version:** 2.0.0.5054
**Operating System:** Ubuntu 16.04.3
**mono Version:** 5.4.1.6

### Log File Snippet
```
17-11-18 06:52:49.2|Info|RssSyncService|RSS Sync Completed. Reports found: 250, Reports grabbed: 0
17-11-18 06:55:47.5|Warn|DownloadMonitoringService|Unable to retrieve queue and history items from NZB Get

[v2.0.0.5054] System.ObjectDisposedException: Cannot access a disposed object.
Object name: 'System.Net.Sockets.NetworkStream'.
  at System.Net.WebConnectionStream.EndWrite (System.IAsyncResult r) [0x00097] in <50d80b08c1a5449282b22aedf03ce925>:0 
  at System.Net.WebConnectionStream.Write (System.Byte[] buffer, System.Int32 offset, System.Int32 size) [0x00053] in <50d80b08c1a5449282b22aedf03ce925>:0 
  at NzbDrone.Common.Http.Dispatchers.ManagedHttpDispatcher.GetResponse (NzbDrone.Common.Http.HttpRequest request, System.Net.CookieContainer cookies) [0x000d2] in M:\BuildAgent\work\b69c1fe19bfc2c38\src\NzbDrone.Common\Http\Dispatchers\ManagedHttpDispatcher.cs:59 
  at NzbDrone.Common.Http.Dispatchers.FallbackHttpDispatcher.GetResponse (NzbDrone.Common.Http.HttpRequest request, System.Net.CookieContainer cookies) [0x000b5] in M:\BuildAgent\work\b69c1fe19bfc2c38\src\NzbDrone.Common\Http\Dispatchers\FallbackHttpDispatcher.cs:53 
  at NzbDrone.Common.Http.HttpClient.ExecuteRequest (NzbDrone.Common.Http.HttpRequest request) [0x0007e] in M:\BuildAgent\work\b69c1fe19bfc2c38\src\NzbDrone.Common\Http\HttpClient.cs:119 
[...]

PR #3299, opened 2019-09-09
Swap to ImageSharp for image resizing to avoid leaks

ta264:
#### Database Migration
NO

#### Description
Swap from ImageResizer to ImageSharp (a pure c# solution) for image resizing.  Helps with #2296 and removes dependency on `System.Drawing` which is missing from .NET core.

#### Todos
- [x] Tests


src/NzbDrone.Core/MediaCover/ImageResizer.cs
```diff
@@ -15,25 +18,20 @@ public class ImageResizer : IImageResizer
         public ImageResizer(IDiskProvider diskProvider)
         {
             _diskProvider = diskProvider;
+
+            // More conservative memory allocation
+            SixLabors.ImageSharp.Configuration.Default.MemoryAllocator = new SimpleGcMemoryAllocator();
         }
```
Taloth:
```
            // Thumbnails don't need super high quality
            SixLabors.ImageSharp.Configuration.Default.ImageFormatsManager.SetEncoder(JpegFormat.Instance, new JpegEncoder()
            {
                Quality = 92
            });
```

Series: The 100
- poster.jpg 728k (source image)
- poster-500.jpg 43.9k original sonarr (q90 i think) vs 55.8k imagesharp (q92)
- poster-250.jpg 12.7k original sonarr (q90 i think) vs 16.2k imagesharp (q92)

q90 produces like 44.5k for poster-500.jpg so slightly larger than gdiplus, at comparable quality (visually inspected).
q92 reduces artifacts by 30% or so, not sure if it's worth the 20% bump in size.

I still dunno if we should go for q90 or q92, but it's better than the 120k file at default q.
If you really want good quality then you probably have to go for q95, but remember that retina screens are going to use the poster.jpg anyway since the browser selects the right image based in ppi.

ta264:
In your example it's a 15kb increase in size for q92 so even with 10,000 series that's only an extra 150mb. And at 10000 series you'd already have 7.8gb of images in total (since we're storing original too) so I'd say the extra 150mb is just rounding error and go for the higher quality, especially if it's noticeable by eye.

Taloth:
Closed via 3ac3dd3ca5d3cd9fcada7b94ed3738f40a50bb78

PR #3346, opened 2019-10-20
Fixed: Get actual file names from QBittorrent API

### LINKED ISSUE #3968: qBittorrent 4.3 won't import finished downloads
Xref from Radarr https://github.com/Radarr/Radarr/issues/5032

Confirmed issue with qBt 4.3 (alpha)  torrent sub folder no longer created which breaks importing as ARR expects the subfolder

Sonarr Code ref https://github.com/Sonarr/Sonarr/blob/phantom-develop/src/NzbDrone.Core/Download/Clients/QBittorrent/QBittorrent.cs#L151

ta264:
#### Database Migration
 NO

#### Description
Previously we were assuming that the output filename was the category
output dir + the torrent name.  This isn't true if the torrent has
been renamed or sanitised.

#### Todos
- [ ] Tests

#### Issues fixed
- Fixes #3968

### DIFF
--- a/src/NzbDrone.Core/Download/Clients/QBittorrent/QBittorrent.cs
+++ b/src/NzbDrone.Core/Download/Clients/QBittorrent/QBittorrent.cs
@@ -1,18 +1,18 @@
 using System;
 using System.Linq;
 using System.Collections.Generic;
+using System.Net;
+using FluentValidation.Results;
+using NLog;
+using NzbDrone.Common.Cache;
 using NzbDrone.Common.Disk;
 using NzbDrone.Common.Extensions;
 using NzbDrone.Common.Http;
 using NzbDrone.Core.Configuration;
 using NzbDrone.Core.MediaFiles.TorrentInfo;
-using NLog;
 using NzbDrone.Core.Validation;
-using FluentValidation.Results;
-using System.Net;
 using NzbDrone.Core.Parser.Model;
 using NzbDrone.Core.RemotePathMappings;
-using NzbDrone.Common.Cache;
 
 namespace NzbDrone.Core.Download.Clients.QBittorrent
 {
@@ -122,6 +122,7 @@ namespace NzbDrone.Core.Download.Clients.QBittorrent
 
         public override IEnumerable<DownloadClientItem> GetItems()
         {
+            var version = Proxy.GetApiVersion(Settings);
             var config = Proxy.GetConfig(Settings);
             var torrents = Proxy.GetTorrents(Settings);
 
@@ -138,18 +139,17 @@ namespace NzbDrone.Core.Download.Clients.QBittorrent
                     DownloadClientInfo = DownloadClientItemClientInfo.FromDownloadClient(this),
                     RemainingSize = (long)(torrent.Size * (1.0 - torrent.Progress)),
                     RemainingTime = GetRemainingTime(torrent),
-                    SeedRatio = torrent.Ratio,
-                    OutputPath = _remotePathMappingService.RemapRemoteToLocal(Settings.Host, new OsPath(torrent.SavePath)),
+                    SeedRatio = torrent.Ratio
                 };
+
+                if (version >= new Version("2.6.1"))
+                {
+                    item.OutputPath = _remotePathMappingService.RemapRemoteToLocal(Settings.Host, new OsPath(torrent.ContentPath));
+                }
 
                 // Avoid removing torrents that haven't reached the global max ratio.
                 // Removal also requires the torrent to be paused, in case a higher max ratio was set on the torrent itself (which is not exposed by the api).
                 item.CanMoveFiles = item.CanBeRemoved = (torrent.State == "pausedUP" && HasReachedSeedLimit(torrent, config));
-
-                if (!item.OutputPath.IsEmpty && item.OutputPath.FileName != torrent.Name)
-                {
-                    item.OutputPath += torrent.Name;
-                }
 
                 switch (torrent.State)
                 {
@@ -216,6 +216,49 @@ namespace NzbDrone.Core.Download.Clients.QBittorrent
         public override void RemoveItem(string hash, bool deleteData)
         {
             Proxy.RemoveTorrent(hash.ToLower(), deleteData, Settings);
+        }
+
+        public override DownloadClientItem GetImportItem(DownloadClientItem item, DownloadClientItem previousImportAttempt)
+        {
+            // On API version >= 2.6.1 this is already set correctly
+            if (!item.OutputPath.IsEmpty)
+            {
+                return item;
+            }
+
+            var files = Proxy.GetTorrentFiles(item.DownloadId.ToLower(), Settings);
+            if (!files.Any())
+            {
+                _logger.Debug($"No files found for torrent {item.Title} in qBittorrent");
+                return item;
+            }
+
+            var properties = Proxy.GetTorrentProperties(item.DownloadId.ToLower(), Settings);
+            var savePath = new OsPath(properties.SavePath);
+
+            var result = item.Clone();
+
+            OsPath outputPath;
+            if (files.Count == 1)
+            {
+                outputPath = savePath + files[0].Name;
+            }
+            else
+            {
+                // we have multiple files in the torrent so just get
+                // the first subdirectory
+                var relativePath = new OsPath(files[0].Name);
+                while (!relativePath.Directory.IsEmpty)
+                {
+                    relativePath = relativePath.Directory;
+                }
+
+                outputPath = savePath + relativePath;
+            }
+
+            result.OutputPath = _remotePathMappingService.RemapRemoteToLocal(Settings.Host, outputPath);
+
+            return result;
         }
 
         public override DownloadClientInfo GetStatus()
--- a/src/NzbDrone.Core.Test/Download/CompletedDownloadServiceTests/ImportFixture.cs
+++ b/src/NzbDrone.Core.Test/Download/CompletedDownloadServiceTests/ImportFixture.cs
@@ -44,6 +44,7 @@ namespace NzbDrone.Core.Test.Download.CompletedDownloadServiceTests
 
             _trackedDownload = Builder<TrackedDownload>.CreateNew()
                     .With(c => c.State = TrackedDownloadState.Downloading)
+                    .With(c => c.ImportItem = completed)
                     .With(c => c.DownloadItem = completed)
                     .With(c => c.RemoteEpisode = remoteEpisode)
                     .Build();
--- a/src/NzbDrone.Core.Test/Download/CompletedDownloadServiceTests/ProcessFixture.cs
+++ b/src/NzbDrone.Core.Test/Download/CompletedDownloadServiceTests/ProcessFixture.cs
@@ -50,6 +50,10 @@ namespace NzbDrone.Core.Test.Download.CompletedDownloadServiceTests
                   .Setup(c => c.Get(It.IsAny<int>()))
                   .Returns(Mocker.GetMock<IDownloadClient>().Object);
 
+            Mocker.GetMock<IProvideImportItemService>()
+                  .Setup(c => c.ProvideImportItem(It.IsAny<DownloadClientItem>(), It.IsAny<DownloadClientItem>()))
+                  .Returns((DownloadClientItem item, DownloadClientItem previous) => item);
+
             Mocker.GetMock<IHistoryService>()
                   .Setup(s => s.MostRecentForDownloadId(_trackedDownload.DownloadItem.DownloadId))
                   .Returns(new EpisodeHistory());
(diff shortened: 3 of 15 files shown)


Taloth:
One thing is that you can save an api call by keeping the download path blank till the torrent has finished downloading. I'd also suggest doing both the filename comparison check, which would save you the api call in many cases.

Alternative solution to this problem is to check whether the name of the torrent ends with an extension. It's not 99% proof, but likely to be mostly reliable.

Basically you have 5 scenarios:

- Single file torrent. path points to 'completed', torrent name is 'file.mkv'
- Multi file torrent. path points to 'completed/name', torrent name is 'name'
- Single file torrent. path points to 'completed', torrent name is 'file-renamed.mkv'
- Multi file torrent. path points to 'completed/name', torrent name is 'name-renamed'
- Multi file torrent. path points to 'completed/name-renamed', torrent name is 'name-renamed'

Based on that:

1. If the path ends with the torrent name, it's multi file, guaranteed.
2. If the name ends with an extension, it's very very likely single file.
3. Otherwise it might be one of those renamed ones, but likely a multi file torrent.

What you can also do is do CleanTitle on the folder and file name, and compare those. That strips out most of the special characters.

I'm not opposed to getting the list of torrent files, but it doesn't seem strictly necessary. I suppose it would be useful to deal with the 3rd option more reliably. But using step 1 and 2 would avoid the api call in 99% of the cases.

ta264:
Thanks, useful comments as always!

I'm not sure you can avoid getting the files if you want it to be foolproof but I might well be missing something.

In my experimentation with the qbit API, `SavePath` is always just `/data/sonarr` and doesn't include the torrent name so I'm not sure your 5 scenarios are quite right.

The title you get back has no sanitisation, and you're free to change it to anything you want in the qbit UI. If you change it in qbit, the title in the API changes.  Neither qbit sanitisation or user renaming seems to have an impact on `SavePath` and the qbit API gives no indication at all as to where the files are actually saved within the completed directory unless you query the files endpoint. Essentially title is unrelated to file output location apart from as the default in the case where no sanitisation is required.

At the moment, we just assume the files will be in `SavePath + title`. This can go wrong in several ways:
1) title looks like an OsPath from different OS to SavePath so the + operation throws and sonarr and every further attempt to fetch items from qbit will throw.
2) qbit has sanitised the filename so sonarr guesses the wrong file location and so cannot import the download
3) the user has renamed the torrent so sonarr can't find the files.

1) is pretty major - the only way to reinstate communication between sonarr and qbit is to go into qbit and rename the offending torrent.

Unless I'm missing a trick with the qbit API I think it's safer to get the file listing. But you're definitely, we might as well wait until download has finished which should also spread the load a bit.

Taloth:
> Neither qbit sanitisation or user renaming seems to have an impact on SavePath and the qbit API gives no indication at all as to where the files are actually saved within the completed directory unless you query the files endpoint.

Ok, clear.

> But you're definitely, we might as well wait until download has finished which should also spread the load a bit.

Yup, it's my biggest concern. But a larger issue is ppl seeding a thousand torrent. Anything we can do to avoid us getting the files would be a plus.
It might be time to implement a separate api for that so the Completed Download handling logic can actively request it.
Like `bool OutputPathDeferred` and then CDH that can actively ask the client to populate the OutputPath on demand. Or just do it without the boolean, leave OutputPath empty and have CDH call the client to populate the deferred details. I see some options there. Like a 'I want to do something with the files, give me more details' method which populates additional properties including, possibly, a list of files.
Btw, we're already calling `FetchTorrentDetails` on-demand to determine the seeding status and only done when the torrent state is pausedUP.

Not sure if it's worth the effort, Maybe emulate the 1000 torrents by calling the api a 1000 times in a loop, see how much cpu and time it takes?

ta264:
I think this is good for another look now. When a tracked download has a completed download state, it'll query the qbittorrent API to get an accurate file listing. Until then `OutputPath` is not populated. I think it should therefore only query the file listing one per torrent (until sonarr restarted).

src/NzbDrone.Core/Download/DownloadEventHub.cs
```diff
@@ -35,6 +36,14 @@ public void Handle(DownloadFailedEvent message)
             RemoveFromDownloadClient(trackedDownload);
         }
 
+        public void Handle(DownloadImportingEvent message)
+        {
+            if (message.TrackedDownload.DownloadItem.OutputPath.IsEmpty)
```
Taloth, line 41: I'm thinking that maybe using the boolean I mentioned earlier is still a good idea. Like `TorrentDetailsDeferred`.
Then you have a `downloadClient.PopulateTorrentDetails(trackedDownload.DownloadItem)` method doing the actual work.

`downloadClient.PopulateTorrentDetails(trackedDownload.DownloadItem)`

That way it's higher abstraction and in the future you can do more 'deferred' stuff. Similar to the `MarkAsImported` function which atm just sets the PostImportCategory.

src/NzbDrone.Core/Download/Clients/QBittorrent/QBittorrent.cs
```diff
@@ -218,6 +212,39 @@ public override void RemoveItem(string hash, bool deleteData)
             Proxy.RemoveTorrent(hash.ToLower(), deleteData, Settings);
         }
 
+        public override OsPath GetOutputPath(string hash)
+        {
+            var torrent = Proxy.GetTorrentProperties(hash, Settings);
+            var savePath = new OsPath(torrent.SavePath);
+            _logger.Trace($"Got save path {savePath}");
```
Taloth, line 219: That's a developer trace line.

src/NzbDrone.Core/Download/Clients/QBittorrent/QBittorrent.cs
```diff
@@ -138,19 +138,13 @@ public override IEnumerable<DownloadClientItem> GetItems()
...
                     RemainingSize = (long)(torrent.Size * (1.0 - torrent.Progress)),
                     RemainingTime = GetRemainingTime(torrent),
-                    SeedRatio = torrent.Ratio,
-                    OutputPath = _remotePathMappingService.RemapRemoteToLocal(Settings.Host, new OsPath(torrent.SavePath)),
+                    SeedRatio = torrent.Ratio
                 };
 
                 // Avoid removing torrents that haven't reached the global max ratio.
                 // Removal also requires the torrent to be paused, in case a higher max ratio was set on the torrent itself (which is not exposed by the api).
                 item.CanMoveFiles = item.CanBeRemoved = (torrent.State == "pausedUP" && HasReachedSeedLimit(torrent, config));
```
Taloth, line 147: We should cache the SavePath provided by GetItems somewhere, because if SavePath changes then the user moved the files/output dir. Which means you have to force it to refresh the Deferred Torrent details.
Normally this was handled implicitly.

Alternatively we can have a timer that allows PopulateTorrentDetails to refresh it every 5-15 minutes or so.

ShadwDrgn:
Having an issue with this PR on the following data:
/info endpoint: (note that the movie name has been replaced with "Movie" here but the original name was also just a single word.
https://pastebin.com/QX2fxgxr
/api/v2/torrents/files endpoint (Same replacement)
https://pastebin.com/241LHXGD
radarr log: (same replacement)
https://pastebin.com/PAVzrteh

austinwbest:
I still think checking if a directory exists at {save_path} + {name} first before hitting the API would work. It has worked that way and only breaks when people/sites rename things, in which case the directory doesn't exist and it would then look at the contents. Would save un-necessary API requests (just my thought anyways)

krootee:
At this moment pretty big issue for some sites - for example, Sonarr would pull '<release>[rartv]' torrent, but the actual folder would be '<release>[rarbg]', and now it's unable to import it.

ta264:
Yes, if you want to use those sites you'll need to downgrade to qbittorrent 4.2.5 for now

rix1337:
Thanks for looking into this and providing a valid fix. Looking at qBittorrent's thread, a fix from their side is unlikely.

Taloth:
@ta264  Wait a sec, based on that other thread, is this only affecting magnet links? If that's the case then we can fudge (like glassez suggests) those for specific qbit versions.
And use content_path for the next qbit version, assuming it properly points to single file or folder.

winklevos:
@Taloth it does look like it points correctly to specific files / sub dir for content. Also includes the parent dir
`"content_path": "/downloads/file.mkv"`
`"content_path": "/downloads/folder"`

austinwbest:
So on a single file, it points at the file & on a multi file it points at the directory? If so that makes life alot easier for >= 4.3

winklevos:
@austinwbest Yep that is correct, tested with the  linuxserver/qbittorrent:unstable docker image.
Checking for content_path would be the best way rather than version comparing

FranciscoPombal:
> @austinwbest Yep that is correct, tested with the linuxserver/qbittorrent:unstable docker image.
> Checking for content_path would be the best way rather than version comparing

That commit also bumps the WebAPI version to 2.6.1, so you can also safely compare against that.

19wolf:
Forgive me if I missed something, but it looks like this PR is ready to go? Why is it not merged yet?

austinwbest:
Because it's not :)

winklevos:
I think the current PR is the work around using the file API not the new content_path which is the better solution

austinwbest:
Absolutely pointless to hit the api like that when content path is exposed. It is literally the reason they added it 😳

winklevos:
Exactly, in fact it would be extra CPU cycles on both ends just for the same info. I'd offer to make the changes but this isn't my language

viggy96:
It appears that #4068 properly uses the content_path that is exposed by qBittorrent. Pretty sure that merge request is ready to be merged.

ta264:
@ShadwDrgn hopefully that error is fixed now in the recent changes.  This PR will now use `content_path` when it becomes available i.e. qbit push a release with API version >= 2.6.1

Taloth (commented):
I'd say it looks good.

Gotta be tested though on enough qbit versions:
- [ ] 3.2.4 (proxy v1, minimum supported version)
- [x] 4.2.5 (proxy v2, would now use GetTorrentFiles)
- [x] 4.3.0.1 (proxy v2, requires GetTorrentFiles)
- [x] 4.3.x nightly (content_path)

ta264:
Hopefully comments addressed.  Tested on:
 - [ ]  3.2.4 (proxy v1, minimum supported version)
- [x] 4.2.5 (proxy v2, would now use GetTorrentFiles)
- [x] 4.3.0.1 (proxy v2, requires GetTorrentFiles)
- [ ] 4.3.x nightly (content_path)

viggy96:
Is testing what is holding up this merge request?

bakerboy448:
> Is testing what is holding up this merge request?

If you are able to help test any of the versions not yet tested, I'm sure the help would be much appreciated

austinwbest:
> Is testing what is holding up this merge request?

Literally just tested this yesterday lol.. Patience grasshopper

Taloth:
A thank you to all those that participated in testing with various qbit versions.

nebbywan:
@ta264 Tried your branch out on an offending rarbg torrent with qBit 4.3.0.1. It failed to import but didn't throw an error in the UI. It looks like, for some reason, qBit added a forward slash at the end of the output string and this caused the problem referenced at https://github.com/Radarr/Radarr/issues/1761

Here's the relevant log
[Log.txt](https://github.com/Sonarr/Sonarr/files/5539555/Log.txt)

bakerboy448:
@nebbywan it was tested the last few days and everyone confirmed it was working correctly. Were you using the latest commits?

Regardless it was just merged, so once it is pushed out try that release and see

nebbywan:
@bakerboy448 Sorry been studying for exams and only found time to slack off this afternoon. But yes, I was using c82db75. Anyway, I guess I'll wait for it to get pushed out and open a separate issue if it still occurs.

micahmo:
Hey all, I didn't look at the code super closely, but can you help me to understand the compatibility of the fixes with qBit?

Looking at the tests, they seem to imply that 4.2.5 and 4.3.0.1 use the same method, but of course 4.3+ is what broke the existing assumptions about the torrent vs. file name.

> 4.2.5 (proxy v2, would now use GetTorrentFiles)
> 4.3.0.1 (proxy v2, requires GetTorrentFiles)

So does this PR really fix badly named torrents with qBit 4.3.0(.1)? (With a workaround?)

Maybe it would be better to wait until `content_path` makes its way into a qBit stable release. (Looks like it's targeted to 4.3.1, although the issue (qbittorrent/qBittorrent#13389) isn't closed, for some reason.) What do you think?

Taloth:
For qbit version 4.3.0.1 and below we fetch the torrent filelist and determine what `content_path` actually, more api calls but more accurate.
For qbit version 4.3.1 (not released) we use `content_path` as provided.

The various tests were simply to check the different qbit api's that sonarr officially still supports.

So this fix should indeed deal with 4.3.0.1 properly. We don't have to wait for content_path to be released, we tested it already against the nightly.

There could be remaining issues with torrent names, but those aren't related to the changes that were introduced after qbit 4.2.5

micahmo:
@Taloth Thank you for the explanation and for your work on this!!

PR #3357, opened 2019-10-25
Auto blacklist failed torrents

Vearaxx:
A very helpful feature would be to have radarr communicate with a torrenting software to automatically blacklist a torrent which failed multiple times. This could be very helpful in minimizing the need for users to check their torrenting software every week to cleanup bad torrents. This could be implemented in profiles with a time specification and amount of download attempts allowed before being blacklisted.


markus101:
You’ve opened a PR instead of a feature request and it it looks like you’re talking about Radarr. There is an existing feature request for handling stalled torrents...

PR #3359, opened 2019-10-28
Added qBittorent state 'moving'

hijxf:
#### Database Migration

NO

#### Description

The state 'moving' wasn't being recording in Radarr, so it would show up as a warning.

image


hijxf:
Anyone?

markus101:
Thanks for the fix!

Issue #3403, opened 2019-11-16
Post import labels not being set for multi-episode releases


l3uddz:
**Describe the bug**
Post-import label for multi-episode releases are not being set (on Deluge).

**Screenshots**
![alt text](https://i.imgur.com/h6s8xo6.png)

![alt text](https://i.imgur.com/ttPo5YM.png)

**System Information**

 - Sonarr Version: 3.0.3.655
 - Operating System: Ubuntu 18.04
 - mono (Linux) Version: 5.20.1.34

**Additional context**
Works perfectly for seasons / single episodes.

markus101:
What do the debug logs show when that file was imported?

l3uddz:
So for a valid import, the following occurs:

```
19-11-16 13:29:41.2|Debug|EpisodeFileMovingService|Hardlinking episode file: /mnt/unionfs/downloads/torrents/deluge/Zapped.S03.1080p.AMZN.WEB-DL.DDP5.1.H.264-ETHiCS/Zapped.S03E06.1080p.AMZN.WEB-DL.DDP5.1.H.264-ETHiCS.mkv to /mnt/unionfs/Media/TV/Zapped!/Season 3/Zapped!.S03E06.WEBDL-1080p.AMZN.h264-ETHICS.mkv
19-11-16 13:29:41.2|Debug|DiskTransferService|HardLinkOrCopy [/mnt/unionfs/downloads/torrents/deluge/Zapped.S03.1080p.AMZN.WEB-DL.DDP5.1.H.264-ETHiCS/Zapped.S03E06.1080p.AMZN.WEB-DL.DDP5.1.H.264-ETHiCS.mkv] > [/mnt/unionfs/Media/TV/Zapped!/Season 3/Zapped!.S03E06.WEBDL-1080p.AMZN.h264-ETHICS.mkv]
19-11-16 13:29:41.2|Debug|EpisodeService|Linking [Season 3/Zapped!.S03E06.WEBDL-1080p.AMZN.h264-ETHICS.mkv] > [[77541]Amulet]
19-11-16 13:29:41.2|Debug|NotificationService|No tags set for this notification.
19-11-16 13:29:41.2|Debug|NotificationService|No tags set for this notification.
19-11-16 13:29:41.2|Debug|CustomScript|Executing external script: /scripts/torrent/TorrentCleanup.py
[...]

PR #3404, opened 2019-11-16
Multi-episode files didn't trigger the DownloadCompletedEvent

Taloth:
#### Description

DownloadCompletedEvent was only raised if the number of files imports was equal or greater than the number of Episodes in the grab event (based on the release title).
Instead it should've compared the number of episodes in the ImportResult rather than the number of ImportResults.

* #3403


Taloth:
Merged

PR #3419, opened 2019-11-23
Increase session expiry time

2omsnF:
#### Database Migration
NO

#### Description
Increase the session expiry time.

There's a compelling reason to have a short expiry on sites with
senstive information like a bank, but if I forget to lock my computer
and someone starts messing with sonarr, there is very little damage they
can do.

#### Todos
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/src/Sonarr.Http/Authentication/AuthenticationModule.cs
+++ b/src/Sonarr.Http/Authentication/AuthenticationModule.cs
@@ -34,7 +34,7 @@ namespace Sonarr.Http.Authentication
 
             if (resource.RememberMe)
             {
-                expiry = DateTime.UtcNow.AddDays(7);
+                expiry = DateTime.UtcNow.AddYears(1);
             }
 
             return this.LoginAndRedirect(user.Identifier, expiry, _configFileProvider.UrlBase + "/");


markus101:
In v3 it’s a sliding window, as long as you access it once in those 7 days it extends another week.

Without a way to force logout on a device you forgot to logout on a year is far too long. I’m not opposed to extending to 2 weeks which is pretty standard for non-banking things.

2omsnF:
The timeout on the log-out provides no additional security, since it seems to be purely client-side: https://github.com/NancyFx/Nancy/blob/de458a9b42db6478e0c2bb8adef0f9fa342a2674/src/Nancy.Authentication.Forms/FormsAuthentication.cs#L277-L279 note how the expiry is not included in the signed portion of the cookie.

However, I think that a "log out everywhere" button would be nice to have, and I have some idea on how to accomplish that.

Taloth:
@markus101 He's not wrong about Nancy not enforcing the auth token expiration. Either the expiration time needs to be included in the encrypted cookie, or it needs a session token to be stored in the database. Both would require us to replace part of Nancy.

Better would be to do the check in the AuthenticationService, which has a direct route to the database so can deal with extending expiry periods.

The biggest issue is that when you change the password, the guid doesn't change. So the cookie remains valid anyway. It's very counter-intuitive. And the only way I see us fix that is by using session guid rather than user guid.

My concern here is effort vs benefit, with Form auth we can do a few things but not with basic auth (which is essentially 'forever' anyway, until the pwd changes)

PR #3430, opened 2019-11-29
Sonarr ui speedup

ta264:
#### Database Migration
NO

#### Description
Sonarr UI speedup for large libraries.

- Don't call `_movieRepository.All()` for each RSS report.
- Fix up scrolling so we only render new items each scroll
- Use a worker for frontend searches
- Assorted other minor speedups
- Option for production profiling build in yarn `yarn run build --production --profile

#### Todos
- [x] Tests


hijxf:
Good work here. Hopefully a maintainer can take a more detailed look and get this merged!

frontend/src/AddSeries/ImportSeries/Import/ImportSeriesRow.js
```diff
@@ -26,7 +24,7 @@ function ImportSeriesRow(props) {
   } = props;
 
   return (
-    <VirtualTableRow style={style}>
+    <>
```
hijxf: Is this shorthand?

markus101:
I need to look closer, but looking at just the commit messages there is a lot of noise that users don't care about. The New and Fixed should be removed and a single `New: Improved performance of series list` added in it's place.

ta264:
Yes I agree it wants squashing down before merging. I thought it might be helpful to leave a bit more context for review but happy to squash now if you prefer

markus101:
Naw, the context is good, was more thinking if we want to land via multiple commits for the context in the future.

frontend/src/Store/Actions/Creators/createHandleActions.js
```diff
@@ -42,6 +42,7 @@ export default function createHandleActions(handlers, defaultState, section) {
 
         if (_.isArray(payload.data)) {
           newState.items = payload.data;
+          newState.itemMap = _.zipObject(_.map(payload.data, 'id'), _.range(payload.data.length));
```
markus101: I think this could be replaced by:

```
newState.itemMap = payload.data.reduce((acc, d, index) => {
  acc[d.id] = index;
  return acc;
}, {});
```

As a helper function in this file (since I don't see a huge need outside of here) it'd eliminate the need for 3 different lodash functions and whatever iterating they need to do.

frontend/src/Components/Page/Header/SeriesSearchInput.js
```diff
@@ -82,6 +72,14 @@ class SeriesSearchInput extends Component {
       );
     }
 
+    if (item.type === LOADING_TYPE) {
+      return (
+        <LoadingIndicator
+          rippleClassName={styles.ripple}
```
markus101: This loader looks a little big and out of place and vertically it's not centered (perhaps leading to the out of place part of it).

image

frontend/src/AddSeries/ImportSeries/Import/ImportSeriesRow.js
markus101: Yeah, it's shorthand for a fragment.

frontend/src/Components/Table/VirtualTable.js
```diff
@@ -44,28 +42,39 @@ class VirtualTable extends Component {
...
-  componentDidUpdate(prevProps, preState) {
-    const { scrollIndex, rowHeight } = this.props;
+    const {
+      width
+    } = this.state;
 
-    if (scrollIndex != null && scrollIndex !== prevProps.scrollIndex) {
-      const scrollTop = (scrollIndex + 1) * rowHeight + 20;
+    if (this._grid &&
+        (prevState.width !== width ||
```
markus101: The indentation here looks wrong, looks like 4 instead of 2.

markus101:
It looks like this will be required to cherry pick the changes in https://github.com/Radarr/Radarr/pull/3982 over which would fix https://github.com/Sonarr/Sonarr/issues/3480 perhaps we can sneak those commits into this PR as well? The `Trigger fewer signalr broadcasts` commit is fine, it's the latter that relies on changes in this PR.

frontend/src/Components/Page/Header/SeriesSearchInput.js
ta264: OK I've shrunk the loader from 50 to 30 pixels and added some padding to the container so that the loader takes up the same space as a single series result.  That way if you get an exact match the `Add New Series` title doesn't jump around.

markus101:
Again, awesome job with this, it's working great.

PR #3466, opened 2019-12-19
Update indexer presets

Fossil01:
#### Database Migration
NO

#### Description
Removed PFMonkey from presets as the indexer closed in 2018. Added x265 category to NZB Finder as it will recategorize those into a new separate category soon.


Fossil01:
The NZB Finder RSS URLs will be different in a soon to be released version. Is this something that needs changing in Sonarr or is only the /api endpoint used?  /api remains the same.

Taloth:
@Fossil01 Only the /api endpoint is used in Sonarr, so that's fine.

PR #3471, opened 2019-12-21
Fixed: Replace duplicate slashes from file names when importing

### LINKED ISSUE #3470: double slash from nzbget prevent importing of files
**Describe the bug**
if the nzbget path of the completed files contains a double dash (eg: /downloads/nzbet//completed) the import will fail.
Sonarr should probably parse the path and find common errors like this.
**Logs**
```
Import failed, path does not exist or is not accessible by Sonarr: /downloads/nzbs/nzbget//completed/tv/Lost.S03E05.1080p.BluRay.x264-aAF en. Ensure the path exists and the user running Sonarr has the correct permissions to access this file/folder
```
**System Information**

 - Sonarr Version: 3.0.3.654
 - Operating System: Docker version 19.03.5, build 633a0ea838
 - Mono 5.20.1.34

gl3nni3:
#### Database Migration
NO

#### Description
Fixes the problem with double / characters not parsing correctly.

#3470

### DIFF
--- a/src/NzbDrone.Common/Disk/OsPath.cs
+++ b/src/NzbDrone.Common/Disk/OsPath.cs
@@ -71,6 +71,11 @@ namespace NzbDrone.Common.Disk
                 case OsPathKind.Windows:
                     return path.Replace('/', '\\');
                 case OsPathKind.Unix:
+                    path = path.Replace('\\', '/');
+                    while (path.Contains("//"))
+                    {
+                        path = path.Replace("//", "/");
+                    }
                     return path.Replace('\\', '/');
             }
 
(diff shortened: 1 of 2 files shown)


src/NzbDrone.Common/Disk/OsPath.cs
```diff
@@ -71,6 +71,11 @@ private static string FixSlashes(string path, OsPathKind kind)
                 case OsPathKind.Windows:
                     return path.Replace('/', '\\');
                 case OsPathKind.Unix:
+                    path = path.Replace('\\', '/');
+                    while (path.Contains("//"))
```
markus101: The while is unnecessary.

The return line also doesn’t make sense given the previous line.

gl3nni3: I added the requested test and fixed the return vallue.
I use the while loop for the case /a///path//////somewhere. In the case such a path is provided.

PR #3484, opened 2020-01-02
Add ability to filter languages in MediaInfo formatter

Taloth:
#### Database Migration
NO

#### Description

Definitely a 1% feature.

@markus101 We talked about this and I think it would still be better to filter by language profile or something, but I feel like there's more edgecases there and harder to implement.
Allow `{MediaInfo SubtitleLanguages:EN+DE}` custom format to influence formatter:
- `:EN+DE` filters out all except EN and DE (the order here determines the output order too)
- `:-DE-NO` filters out DE and NO
- `:EN+DE+` (trailing + ) will replace any unlisted languages with `--`, so a file with `EN+NL+NO` would become `[EN+--]` (not sure about this one)

Formats work for AudioLanguages, AudioLanguagesAll and even Full. For Full it'll eliminates any possible future formatting for the codec/etc stuff that's in Full, but that's not a big issue imo.

Finally I'm also not sure about how to document it.

#### Todos
- [x] Tests
- [ ] Documentation

### DIFF
--- a/frontend/src/Helpers/Props/icons.js
+++ b/frontend/src/Helpers/Props/icons.js
@@ -24,6 +24,7 @@ import {
 import {
   faArrowCircleLeft as fasArrowCircleLeft,
   faArrowCircleRight as fasArrowCircleRight,
+  faAsterisk as fasAsterisk,
   faBackward as fasBackward,
   faBars as fasBars,
   faBolt as fasBolt,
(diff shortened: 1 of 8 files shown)


Taloth:
Problem with that is what would you do if the languages are EN & NL and the format is `:DE+`?
In the current format it'd be `[--]`, leaving the hyphens out would mean either `[+]` or just nothing.

PR #3486, opened 2020-01-04
Show size in mass editor

### LINKED ISSUE #3273: Show size in Mass Editor
**Describe the problem**
Mass move in v3 doesn't show the size of the folder. This would be useful if you are moving large numbers of TV Shows it'll be good to have a rough idea of how much you are moving.
Moving series' to the cloud you can work out how long it'll take to upload etc. (Useful for people on slow uploads) 

Add Size Collum in Mass editor

**Additional context**

Meridian1701:
#### Database Migration
NO

#### Description
Added a size column on the Mass Editor page. Is the column label and column position ok?

alt text

Also noted that the title sorting doesn't appear to be working. 

image

In this case when I'm sorting in ascending order, Chernobyl should be on the bottom. If it's incorrect behaviour, I can fix it in a seperate issue/PR. 

#### Todos
- [X] Tests (Couldn't find any front end test?)
- [X] Documentation 

Fixes #3273

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -140,7 +140,5 @@ output/*
 
 _start
 
-src/.idea/
-
 #JetBrains
 .idea
--- a/.gitignore
+++ b/.gitignore
@@ -141,3 +141,6 @@ output/*
 _start
 
 src/.idea/
+
+#JetBrains Rider
+.idea
--- a/.gitignore
+++ b/.gitignore
@@ -142,5 +142,5 @@ _start
 
 src/.idea/
 
-#JetBrains Rider
+#JetBrains
 .idea
--- a/frontend/src/Series/Editor/SeriesEditor.js
+++ b/frontend/src/Series/Editor/SeriesEditor.js
@@ -62,6 +62,12 @@ function getColumns(showLanguageProfile) {
       isVisible: true
     },
     {
+      name: 'sizeOnDisk',
+      label: 'Size',
+      isSortable: true,
+      isVisible: true
+    },
+    {
       name: 'tags',
       label: 'Tags',
       isSortable: false,
--- a/frontend/src/Series/Editor/SeriesEditorRow.js
+++ b/frontend/src/Series/Editor/SeriesEditorRow.js
@@ -10,6 +10,7 @@ import TableSelectCell from 'Components/Table/Cells/TableSelectCell';
 import SeriesTitleLink from 'Series/SeriesTitleLink';
 import SeriesStatusCell from 'Series/Index/Table/SeriesStatusCell';
 import styles from './SeriesEditorRow.css';
+import formatBytes from 'Utilities/Number/formatBytes';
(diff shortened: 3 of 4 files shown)


Taloth:
Sorting seems ok. articles such as 'A'/'The' are ignored in the sort order. So 'Discovery of Witches' goes last.

.gitignore
```diff
@@ -141,3 +141,6 @@ output/*
 _start
 
 src/.idea/
+
+#JetBrains Rider
```
markus101: Nit: `# Jetbrains` since it applies to everything (and consistent with other comments).

frontend/src/Series/Editor/SeriesEditor.js
```diff
@@ -61,6 +61,12 @@ function getColumns(showLanguageProfile) {
       isSortable: true,
       isVisible: true
     },
+    {
+      name: 'sizeOnDisk',
+      label: 'Size',
+      isSortable: true,
+      isVisible: true
```
markus101: I was thinking this would be optional (personally it's unnecessary information to me here), like it is in the series list. That'd mean this PR requires the UI around makingit optional though. What do you think @Taloth?

Meridian1701:
> Sorting seems ok. articles such as 'A'/'The' are ignored in the sort order. So 'Discovery of Witches' goes last.

Ok, I guess it makes sense from an "english perspective". I'm not a native speaker.

frontend/src/Series/Editor/SeriesEditor.js
Meridian1701: I would be happy to implement this feature.

Taloth: @markus101 That table has not settings yet though. But I actually looked at the PR earlier and thought the same.

markus101: Yeah, it's not there, otherwise it'd be trivial,. It should be pretty straight forward following another table as an example.

Meridian1701: How about implementing an equivalent table options dialog like the one on the history page? I don't think it would be too hard. As for persistence of the column settinngs, I guess it would involve an update of the bank end and db schema. That could perhaps be done in a seperate PR, if it's something wanted?

markus101: That's what I've been referring to with this comment chain.

It wouldn't require any backend changes, it's purely frontend. I think it makes sense to do it as part of this PR, since it changes the UX.

Meridian1701: Got it! Hopefully I will be able to implement it this weekend.

markus101:
Size on Disk and column reordering implemented in 069fc5cd33de8129960bfb5fb889e6783a01f74b

PR #3499, opened 2020-01-09
BTN: Update name

ThatNerdyPikachu:
#### Database Migration
No

#### Description
``BroadcastheNet`` -> ``BroadcasTheNet``, as per site.

Please review class namespace, and folder / file name changes, to make sure I got everything. (Used Rider's refactoring, so *should* be good)


ThatNerdyPikachu:
Should I just limit this to the first commit (and keep the migration string change), is changing everything a bit too much?

Taloth:
There are a few things wrong here, but I gotta ask: Why fix it? It seems such a trivial capitalization difference that's hardly worth the time.

Real problem is that you changed the class names, and changed an old migration. Users upgrading from old databases aren't going to match since 'Implementation' is different for them. It'd need a new migration to fix those up, rather than revise old ones.

So some options:
1) Only change the BroadcasTheNet.Name property, boom done. Leaving the namespaces&classes unchanged.
2) Add a new migration to fix up the Implementation and ConfigContract columns in the database.

Issue #3521, opened 2020-01-15
Telegram unable to send test message


pshurgal:
**Describe the bug**
I am trying to enable Telegram integration at my server, but it doesn't work.
I have correct bot id and chat id, I can send message to the chat by entering request URL at my browser (with proxy enabled), but when I try enable Telegram integration I always get error message.

Telegram is blocked in Russian Federation, but it works using proxy server. I have tried to enable proxy at general setting but it seems that TelegramProxy doesn't use global proxy settings.

**Logs**
 > 20-1-15 17:37:26.0|Error|RestSharpExtensions|Error communicating with server
 > 
 > [v3.0.3.688] System.Net.WebException: Error: ConnectFailure (Connection refused) ---> System.Net.Sockets.SocketException: Connection refused
 >   at System.Net.Sockets.SocketAsyncResult.CheckIfThrowDelayedException () [0x00014] in <91935ad653254a93b9d73a9f8f2f7a2d>:0 
 >   at System.Net.Sockets.Socket.EndConnect (System.IAsyncResult asyncResult) [0x0002c] in <91935ad653254a93b9d73a9f8f2f7a2d>:0 
 >   at System.Net.Sockets.SocketTaskExtensions+<>c.<ConnectAsync>b__2_1 (System.IAsyncResult asyncResult) [0x00006] in <91935ad653254a93b9d73a9f8f2f7a2d>:0 
 >   at System.Threading.Tasks.TaskFactory`1[TResult].FromAsyncCoreLogic (System.IAsyncResult iar, System.Func`2[T,TResult] endFunction, System.Action`1[T] endAction, System.Threading.Tasks.Task`1[TResult] promise, System.Boolean requiresSynchronization) [0x00019] in <254335e8c4aa42e3923a8ba0d5ce8650>:0 
 > --- End of stack trace from previous location where exception was thrown ---
 > 
 >   at System.Net.WebConnection.Connect (System.Net.WebOperation operation, System.Threading.CancellationToken cancellationToken) [0x0019b] in <91935ad653254a93b9d73a9f8f2f7a2d>:0 
 >    --- End of inner exception stack trace ---
 >   at System.Net.WebConnection.Connect (System.Net.WebOperation operation, System.Threading.CancellationToken cancellationToken) [0x00217] in <91935ad653254a93b9d73a9f8f2f7a2d>:0 
[...]

PR #3523, opened 2020-01-16
Fixed: RestClient does not use global proxy settings

pshurgal:
#### Database Migration
NO

#### Description
Telegram is blocked in Russian Federation, but it works using proxy server. But `TelegramProxy` wasn't using global proxy settings.

#### TODOS

Test that connection goes through proxy server.

* https://github.com/Sonarr/Sonarr/issues/3521


Taloth:
We actually want to get rid of RestSharp usage, but this change looks solid. ty

Did you manage to test it?

pshurgal:
Yes, I have tested Telegram connection using some public proxies, and it works now. And I suppose it will work with other services which use RestSharp.

Taloth:
excellent, thank you.

PR #3597, opened 2020-03-02
Fixed: Tag inputs respect non-QWERTY layouts

twinpixxx:
#### Database Migration
NO

#### Description
While in QWERTY `keyCode` for comma is 188, in Cyrillic layout
same `keyCode` represents regular letter. It's better idea to look
for a `key` instead, which is consistent between keyboard layouts


markus101:
Thanks for the fix on this one.

PR #3610, opened 2020-03-08
Mono 6 patch, updated framework checks to 4.7.2/mono 5.18 and post-install update check

Taloth:
#### Database Migration
NO

#### Description
- Mono 6 workaround for the copy/filesystem issue. untested
- Copied over the framework version checks from the phantom-x64 branch to ensure ppl start moving over
- Increased debian package dependency for mono to 5.18
- Now actually doing something with the `update_required` file that was created by the debian package to force an early update check. (necessary to avoid database schema downgrade issues)

#### Todos
- [ ] Tests

### DIFF
--- a/src/NzbDrone.Core/Update/InstallUpdateService.cs
+++ b/src/NzbDrone.Core/Update/InstallUpdateService.cs
@@ -1,6 +1,8 @@
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.IO;
+using System.Threading;
 using NLog;
 using NzbDrone.Common;
 using NzbDrone.Common.Disk;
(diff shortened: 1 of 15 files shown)


src/NzbDrone.Core/Update/InstallUpdateService.cs
```diff
@@ -252,5 +258,65 @@ public void Execute(ApplicationUpdateCommand message)
...
+                _logger.Info("Installing post-install update from {0} to {1}", BuildInfo.Version, latestAvailable.Version);
+                _diskProvider.DeleteFile(updateMarker);
+
+                var installing = InstallUpdate(latestAvailable);
+
+                if (installing)
+                {
+                    _logger.Debug("Install in progress, giving installer 30 seconds.");
+
+                    Stopwatch watch = Stopwatch.StartNew();
```
markus101: Nit: `var watch`

PR #3620, opened 2020-03-12
Don't auto monitor and search new episodes on tvdb that aired long time ago

Taloth:
1) Don't monitor newly added episodes with airtime < 14 days ago
2) Don't monitor any newly added episodes when no episodes existed previously
3) Don't auto-search newly added episodes with airtime < 14 days ago

The 3rd check isn't really needed coz it's unmonitored anyway, but well.... that's the first piece of code I added.

### DIFF
--- a/src/NzbDrone.Core/Tv/RefreshEpisodeService.cs
+++ b/src/NzbDrone.Core/Tv/RefreshEpisodeService.cs
@@ -34,6 +34,7 @@ namespace NzbDrone.Core.Tv
 
             var existingEpisodes = _episodeService.GetEpisodeBySeries(series.Id);
             var seasons = series.Seasons;
+            var hasExisting = existingEpisodes.Any();
 
             var updateList = new List<Episode>();
             var newList = new List<Episode>();
@@ -85,6 +86,8 @@ namespace NzbDrone.Core.Tv
                 }
             }
 
+            UnmonitorReaddedEpisodes(series, newList, hasExisting);
+
             var allEpisodes = new List<Episode>();
             allEpisodes.AddRange(newList);
             allEpisodes.AddRange(updateList);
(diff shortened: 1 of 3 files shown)


src/NzbDrone.Core/Tv/RefreshEpisodeService.cs
```diff
@@ -34,6 +34,7 @@ public void RefreshEpisodeInfo(Series series, IEnumerable<Episode> remoteEpisode
 
             var existingEpisodes = _episodeService.GetEpisodeBySeries(series.Id);
             var seasons = series.Seasons;
+            var hasExisting = existingEpisodes.Any();
```
markus101: Will this be a problem for new series that didn't initially have episodes? It's rare with TVDB, but other metadata providers could handle that differently.

Perhaps we should treat season 1 differently? I think the worst case there is it is older than 14 days and gets unmonitored anyways.

Taloth: valid point, i was going to say that new episodes air in the future.
How about: If any new episode aired < 14. We unmonitor all new episodes unless they air today/future?
Should cover both cases, new series get monitored properly unless they were seriously behind with adding it. And any se

markus101: Make it 1 day to deal with the TVDB caching/delays in being added for whatever reason? Otherwise I think that sounds good.

src/NzbDrone.Core/Tv/RefreshEpisodeService.cs
```diff
@@ -120,6 +123,41 @@ private bool GetMonitoredStatus(Episode episode, IEnumerable<Season> seasons)
...
+                return;
+            }
+
+            var threshold = DateTime.UtcNow.AddDays(-14);
+
+            var oldEpisodes = episodes.Where(e => e.AirDateUtc.HasValue && e.AirDateUtc.Value.Before(threshold)).ToList();
+
+            if (oldEpisodes.Any())
+            {
+                if (!hasExisting)
```
markus101: Nit: Since we're handling the else I'd flip this. `if (hasExisting)... else`

PR #3634, opened 2020-03-20
Download History

### LINKED ISSUE #3554: Separate Table for Tracking Grabbed and Imported Downloads
Currently we rely on history to determine whether a release has been imported and that requires 1 on n files to be imported. When we changed this to be n of n we ran into issues with episode IDs not matching and history not being as accurate as it was expected to be.

Instead of relying on each episode's history to determine if a download was completely imported we should track the download as a whole in a new table, which will house both imported and grabbed events. We'll add to this table when a download is grabbed as well as when it is completely imported so it can be used to automatically import a download that was grabbed, imported and later re-grabbed because we know it was grabbed after the previous import.

tracking each file that we import in this table will also allow us to check later to see which files were imported, without solely relying on logs and episode history as both can be incomplete.

Ultimately we want:

- Ability to re-grab the same release and properly import files
- Only treat a download as imported if all files were actually imported
- Hide imported episodes from queue (this might be a combination both histories) if all episodes haven't been imported yet

Table name: `DownloadHistory`
Events: `DownloadGrabbed`, `FileImported`, `DownloadImported`

- When a release is grabbed each episode in the release will get an row in `History` and 1 row in `DownloadHistory`
- When a double episode is imported `History` will have 2 rows and `DownloadHistory` will have one for the `FileImported` and one for the `DownloadImported`

### LINKED ISSUE #3557: Pending Import still in import queue after removed from download client
**Describe the bug**
If an complete automatic import fails (one cause is because it was replacing an existing season and I had a file with 2 episodes in 1 and was replacing with 2 separate files) then it will import all the other files and report a stalled import in the queue as intended.

If I manually import the remaining files myself (not through the queue as Sonarr has a habit of deleting additional files in a season if trying to import just 1 at a time) then it will spam the logs once per minute with "Import failed, path does not exist or is not accessible by Sonarr".

The error is identical to all the reports involving insufficient permissions or remote mapping, but in this case the perms and mapping are fine.

It simply can't locate the leftover files that were removed from the downloads directory and will continue looking for them even after I clear it from Transmission and the queue. It's apparently been doing it for the better part of the day, so no working timeout is evident.

The errors only stopped after I restarted Sonarr. If I recreated the folders it complained about, the errors stopped but Sonarr did nothing with them and the errors immediately started again if the folders were deleted.

Note, this is only happening on Sonarr v3, Sonarr v2 never had this log issue after doing the same action.

**Logs**
Just thousands of:
`20-2-1 21:17:02.3|Error|DownloadedEpisodesImportService|Import failed, path does not exist or is not accessible by Sonarr: \\HOMENAS\Videos\Downloads\[HorribleSubs] Sword Art Online - Alicization - War of Underworld - 11 [1080p].mkv. Ensure the user running Sonarr has access to the network share`
No other references in the logs.

**System Information**

 - Sonarr Version: 3.0.3.688
 - Operating System: Windows 10
 - .net Framework (Windows) Version: 3.5/4.8

markus101:
#### Database Migration
YES

#### Description
Implements a new history designed to persist through episode changes, based on the `downloadId` instead of `episodeId` + `downloadId`.

* Fixes #3557
* Closes #3554

### DIFF
--- a/src/NzbDrone.Core.Test/Datastore/DatabaseRelationshipFixture.cs
+++ b/src/NzbDrone.Core.Test/Datastore/DatabaseRelationshipFixture.cs
@@ -2,6 +2,7 @@
 using FizzWare.NBuilder;
 using FluentAssertions;
 using NUnit.Framework;
+using NzbDrone.Core.History;
 using NzbDrone.Core.MediaFiles;
 using NzbDrone.Core.Qualities;
 using NzbDrone.Core.Test.Framework;
@@ -59,21 +60,21 @@ namespace NzbDrone.Core.Test.Datastore
         {
             var quality = new QualityModel { Quality = Quality.Bluray720p, Revision = new Revision(version: 2 )};
 
-            var history = Builder<History.History>.CreateNew()
+            var history = Builder<EpisodeHistory>.CreateNew()
                                                   .With(c => c.Id = 0)
                                                   .With(c => c.Quality = quality)
                                                   .Build();
 
             Db.Insert(history);
 
-            var loadedQuality = Db.Single<History.History>().Quality;
+            var loadedQuality = Db.Single<EpisodeHistory>().Quality;
             loadedQuality.Should().Be(quality);
         }
 
         [Test]
         public void embedded_list_of_document_with_json()
         {
-            var history = Builder<History.History>.CreateListOfSize(2)
+            var history = Builder<EpisodeHistory>.CreateListOfSize(2)
                                                   .All().With(c => c.Id = 0)
                                                   .Build().ToList();
 
@@ -83,7 +84,7 @@ namespace NzbDrone.Core.Test.Datastore
 
             Db.InsertMany(history);
 
-            var returnedHistory = Db.All<History.History>();
+            var returnedHistory = Db.All<EpisodeHistory>();
 
             returnedHistory[0].Quality.Quality.Should().Be(Quality.HDTV1080p);
         }
--- a/src/NzbDrone.Core/Datastore/Migration/139_add_download_history.cs
+++ b/src/NzbDrone.Core/Datastore/Migration/139_add_download_history.cs
@@ -1,0 +1,98 @@
+using System;
+using System.Collections.Generic;
+using System.Data;
+using FluentMigrator;
+using NzbDrone.Common.Serializer;
+using NzbDrone.Core.Datastore.Migration.Framework;
+
+namespace NzbDrone.Core.Datastore.Migration
+{
+    [Migration(139)]
+    public class add_download_history : NzbDroneMigrationBase
+    {
+        protected override void MainDbUpgrade()
+        {
+            Create.TableForModel("DownloadHistory")
+                  .WithColumn("EventType").AsInt32().NotNullable()
+                  .WithColumn("SeriesId").AsInt32().NotNullable()
+                  .WithColumn("DownloadId").AsString().NotNullable()
+                  .WithColumn("SourceTitle").AsString().NotNullable()
+                  .WithColumn("Date").AsDateTime().NotNullable()
+                  .WithColumn("Protocol").AsInt32().Nullable()
+                  .WithColumn("IndexerId").AsInt32().Nullable()
+                  .WithColumn("DownloadClientId").AsInt32().Nullable()
+                  .WithColumn("Release").AsString().Nullable()
+                  .WithColumn("Data").AsString().Nullable();
+
+            Create.Index().OnTable("DownloadHistory").OnColumn("EventType");
+            Create.Index().OnTable("DownloadHistory").OnColumn("SeriesId");
+            Create.Index().OnTable("DownloadHistory").OnColumn("DownloadId");
+
+            Execute.WithConnection(InitialImportedDownloadHistory);
+        }
+
+        private static readonly Dictionary<int, int> EventTypeMap = new Dictionary<int, int>()
+        {
+            // EpisodeHistoryType.Grabbed -> DownloadHistoryType.Grabbed
+            {1, 1},
+            // EpisodeHistoryType.DownloadFolderImported -> DownloadHistoryType.DownloadImported
+            {3, 2},
+            // EpisodeHistoryType.DownloadFailed -> DownloadHistoryType.DownloadFailed
+            {4, 3},
+            // EpisodeHistoryType.DownloadIgnored -> DownloadHistoryType.DownloadIgnored
+            {7, 4}
+        };
+
+        private void InitialImportedDownloadHistory(IDbConnection conn, IDbTransaction tran)
+        {
+            using (var cmd = conn.CreateCommand())
+            {
+                cmd.Transaction = tran;
+                cmd.CommandText = "SELECT SeriesId, DownloadId, EventType, SourceTitle, Date, Data FROM History WHERE DownloadId IS NOT NULL AND EventType IN (1, 3, 4, 7) GROUP BY EventType, DownloadId";
+
+                using (var reader = cmd.ExecuteReader())
+                {
+                    while (reader.Read())
+                    {
+                        var seriesId = reader.GetInt32(0);
+                        var downloadId = reader.GetString(1);
+                        var eventType = reader.GetInt32(2);
+                        var sourceTitle = reader.GetString(3);
+                        var date = reader.GetDateTime(4);
+                        var rawData = reader.GetString(5);
+                        var data = Json.Deserialize<Dictionary<string, string>>(rawData);
+
+                        var downloadHistoryEventType = EventTypeMap[eventType];
+                        var protocol = data.ContainsKey("protocol") ? Convert.ToInt32(data["protocol"]) : (int?)null;
+                        var downloadHistoryData = new Dictionary<string, string>();
+
+                        if (data.ContainsKey("indexer"))
+                        {
+                            downloadHistoryData.Add("indexer", data["indexer"]);
+                        }
+
+                        if (data.ContainsKey("downloadClient"))
+                        {
+                            downloadHistoryData.Add("downloadClient", data["downloadClient"]);
+                        }
+
+                        using (var updateCmd = conn.CreateCommand())
+                        {
+                            updateCmd.Transaction = tran;
+                            updateCmd.CommandText = @"INSERT INTO DownloadHistory (EventType, SeriesId, DownloadId, SourceTitle, Date, Protocol, Data) VALUES (?, ?, ?, ?, ?, ?, ?)";
+                            updateCmd.AddParameter(downloadHistoryEventType);
+                            updateCmd.AddParameter(seriesId);
+                            updateCmd.AddParameter(downloadId);
+                            updateCmd.AddParameter(sourceTitle);
+                            updateCmd.AddParameter(date);
+                            updateCmd.AddParameter(protocol);
+                            updateCmd.AddParameter(downloadHistoryData.ToJson());
+
+                            updateCmd.ExecuteNonQuery();
+                        }
+                    }
+                }
+            }
+        }
+    }
+}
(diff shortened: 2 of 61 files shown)


src/NzbDrone.Core/Datastore/Migration/139_add_download_history.cs
```diff
@@ -0,0 +1,58 @@
...
+
+            Execute.WithConnection(InitialImportedDownloadHistory);
+        }
+
+        private void InitialImportedDownloadHistory(IDbConnection conn, IDbTransaction tran)
+        {
+            using (var cmd = conn.CreateCommand())
+            {
+                cmd.Transaction = tran;
+                cmd.CommandText = "SELECT SeriesId, DownloadId, SourceTitle, Date FROM History WHERE DownloadId IS NOT NULL AND EventType = 3 GROUP BY DownloadId";
```
Taloth: Shouldn't EventType = 1 (Grabbed) be processed too?

src/NzbDrone.Core/Download/FailedDownloadService.cs
```diff
@@ -34,7 +34,7 @@ public void MarkAsFailed(int historyId)
             var downloadId = history.DownloadId;
             if (downloadId.IsNullOrWhiteSpace())
             {
-                PublishDownloadFailedEvent(new List<History.History> { history }, "Manually marked as failed");
+                PublishDownloadFailedEvent(new List<Core.History.History> { history }, "Manually marked as failed");
```
Taloth: How about we rename `History.History` to `History.EpisodeHistory`? Avoids the namespace clash and seems more applicable now we have a separate history for downloads.

src/NzbDrone.Core/Download/History/DownloadHistory.cs
```diff
@@ -0,0 +1,23 @@
...
+namespace NzbDrone.Core.Download.History
+{
+    public class DownloadHistory : ModelBase
+    {
+        public DownloadHistoryEventType EventType { get; set; }
+        public int SeriesId { get; set; }
+        public string DownloadId { get; set; }
+        public string SourceTitle { get; set; }
+        public DateTime Date { get; set; }
+    }
```
Taloth: I feel like we should include more details while were at it:
- Protocol
- IndexerId (maybe Indexer name too?)
- DownloadClientId (maybe Download Client name too?)
- Data json with additional details, but in particular:
  - ReleaseInfo (kinda like PendingReleases), it could be useful in calculating upgrade word scores
  - Original Preferred Word score

src/NzbDrone.Core/Download/History/DownloadHistory.cs
```diff
@@ -0,0 +1,23 @@
+﻿using System;
+using NzbDrone.Core.Datastore;
+
+namespace NzbDrone.Core.Download.History
+{
+    public class DownloadHistory : ModelBase
+    {
+        public DownloadHistoryEventType EventType { get; set; }
+        public int SeriesId { get; set; }
```
Taloth: What are we doing to do when a series gets deleted and readded? (esp when people 'panic' re-add a series to try refresh metadata)   
Previously, history would get purged in housekeeping. I'm wondering if we'd want the same here or be a bit more 'safe'.

src/NzbDrone.Core/Download/History/DownloadHistoryService.cs
```diff
@@ -0,0 +1,164 @@
...
+                SeriesId = message.SeriesId,
+                DownloadId = message.DownloadId,
+                SourceTitle = message.SourceTitle,
+                Date = DateTime.UtcNow
+            });
+        }
+
+        public void Handle(SeriesDeletedEvent message)
+        {
+            _repository.DeleteBySeriesId(message.Series.Id);
```
Taloth: As mentioned elsewhere, do we want to delete downloadHistory in this case? if so, probably better as Housekeeping job a la CleanupOrphanedHistoryItems.

src/NzbDrone.Core/Download/History/DownloadHistoryService.cs
```diff
@@ -0,0 +1,164 @@
...
+        {
+            _repository = repository;
+            _historyService = historyService;
+        }
+
+        public bool DownloadAlreadyImported(string downloadId)
+        {
+            var events = _repository.FindByDownloadId(downloadId);
+
+            // Events are ordered by date descending, if a grabbed event comes before an imported event then it was never imported
```
Taloth: Nice, but that could cause issues with partial imports that never finished, and was later regrabbed. Those FileImported events would still be there and potentially picked up. (At least that's what I assume FileImported would eventually be used for)
We could deal with that by comparing the Data of the FileImported events to the grab event and ignore them.

src/NzbDrone.Core/Download/History/DownloadHistoryService.cs
```diff
@@ -0,0 +1,164 @@
...
+
+            return false;
+        }
+
+        public DownloadHistory GetLatestDownloadHistoryItem(string downloadId)
+        {
+            var events = _repository.FindByDownloadId(downloadId);
+
+            // Events are ordered by date descending. We'll return the most recent expected event.
+            foreach (var e in events)
```
Taloth: What you're doing is just a `FirstOrDefault(e => e.EventType == .. || ..)`

src/NzbDrone.Core/Download/TrackedDownloads/TrackedDownloadAlreadyImported.cs
```diff
@@ -7,12 +7,12 @@ namespace NzbDrone.Core.Download.TrackedDownloads
 {
     public interface ITrackedDownloadAlreadyImported
     {
-        bool IsImported(TrackedDownload trackedDownload, List<History.History> historyItems);
+        bool IsImported(TrackedDownload trackedDownload, List<Core.History.History> historyItems);
     }
 
     public class TrackedDownloadAlreadyImported : ITrackedDownloadAlreadyImported
```
Taloth: No longer used.

src/NzbDrone.Core/Download/History/DownloadHistoryService.cs
```diff
@@ -0,0 +1,164 @@
...
+            if (!message.NewDownload)
+            {
+                return;
+            }
+
+            var downloadId = message.DownloadId;
+
+            if (downloadId.IsNullOrWhiteSpace())
+            {
+                downloadId = _historyService.FindDownloadId(message);
```
Taloth: Is this to deal with manual imports? If so, it's worth a code comment.

src/NzbDrone.Core/Download/History/DownloadHistory.cs
markus101: What would we use this for? Unlike History this is meant as an internal tracking of grabs/imports and I'm not opposed to adding more information if we have a goal for it, but right now this is a moving target and delays fixing of the known issues.

src/NzbDrone.Core/Datastore/Migration/139_add_download_history.cs
markus101: I was more concerned with the imports, I can add the grabs as well.

src/NzbDrone.Core/Download/History/DownloadHistoryService.cs
markus101: Unless we're not going to link it to the series, relink it to a series (making history mutable or series IDs consistent) I think it makes sense to clear it. Not deleting it here would be there is a point where we're inconsistent, the house keeping task for history was meant to ensure eventual consistency if handling the series delete failed, not as the only means. If we don't purge it here why clean it up at all?

src/NzbDrone.Core/Download/History/DownloadHistory.cs
Taloth: The existing Grab/Import have the Protocol, indexer & Download client details. I believe it would be prudent to duplicate that here so we have all information about a download here.

ReleaseInfo isn't present anywhere yet, and the usecase is to be able to use all info in the ReleaseInfo/TorrentInfo to calculate upgrade scores that don't rely on title alone, such as the Indexer filter in the profile or other tags (freeleech).   
It's not something we currently have (except the indexer filter), but I worked on it a bit and the full ReleaseInfo was something I missed. It's only applicable for the Grab event, obviously.

src/NzbDrone.Core/Download/History/DownloadHistoryService.cs
Taloth: Hm, I thought we didn't clear History immediately on series deletions. But I just checked, we do, so I was wrong about that.

Reason for keeping downloadHistory is to be able to keep ignoring downloads for series that were already imported, but still seeding.
So that would be an argument to keep the DownloadHistory permanently.

Hm... If we tracked DownloadRemoved separately, then we could purge downloadhistory more selectively. That'd deal with the 'seeding' edge-case, while removing unused records.
I'm ok with doing that another time though.

src/NzbDrone.Core/Download/History/DownloadHistoryService.cs
markus101: True, though I think this might be easier to grok and reduces the need to recheck the event type.

src/NzbDrone.Core/Download/TrackedDownloads/TrackedDownloadAlreadyImported.cs
markus101: Still used in `CompletedDownloadService`

Taloth: In that case, shouldn't this be updated to check DownloadHistory instead?

src/NzbDrone.Core/Download/History/DownloadHistoryService.cs
markus101: Yeah, it'd be nice to keep that all around, same as we talked about being able to map between episodes that change numbering so we can keep the linked episode files.

src/NzbDrone.Core/Datastore/Migration/139_add_download_history.cs
```diff
@@ -0,0 +1,94 @@
...
+            Create.Index().OnTable("DownloadHistory").OnColumn("EventType");
+            Create.Index().OnTable("DownloadHistory").OnColumn("SeriesId");
+            Create.Index().OnTable("DownloadHistory").OnColumn("DownloadId");
+
+            Execute.WithConnection(InitialImportedDownloadHistory);
+        }
+
+        private static readonly Dictionary<int, int> EventTypeMap = new Dictionary<int, int>()
+        {
+            {1, 1},
```
Taloth, line 36: Nitpick: add comments like `// EpisodeHistoryType.Grabbed -> DownloadHistoryType.Grabbed`

src/NzbDrone.Core.Test/Datastore/DatabaseRelationshipFixture.cs
```diff
@@ -59,21 +59,21 @@ public void embedded_document_as_json()
         {
             var quality = new QualityModel { Quality = Quality.Bluray720p, Revision = new Revision(version: 2 )};
 
-            var history = Builder<History.History>.CreateNew()
+            var history = Builder<History.EpisodeHistory>.CreateNew()
```
Taloth, line 62: Can actually get rid of `History` now, here and in other places. Might be worth to just do a search&replace for `History.EpisodeHistory`.

markus101: Yeah, I though ti got them, but I think those were the `Core.History.EpisodeHistory` ones, replaced them now.

Taloth:
Was already approved, ship it!

PR #3684, opened 2020-04-16
Phantom search improvements

Taloth:
#### Database Migration
NO

#### Description

Mixed a few changes together

- nzedb can return a newznab api error result when downloading .nzb, detect that and produce readable error
- Newznab/Torznab will now _also_ search by title and mapped season number for SceneMappings where both SeasonNumber and SceneSeasonNumber is specified. Examples are Spartacus and Cosmos.
  Parser will no longer produce Wrong season in that scenario either
  This means that such episodes are now also searchable, rather than only parseable on rss
- Display the SceneSeasonNumber on the UI when specified
- Bonus feature for Jackett: If they specify `title` as supported query parameter, we'll use that instead of `q` for single/season/daily searches with the original series title (or scene title), rather than the newznab cleaned one. I didn't discuss this with them yet, but they'll probably be happy since it allows them handling apostrophe themselves.

Example: Searching Cosmos 2x01 results in the following queries:
```
/api?t=tvsearch&...&tvdbid=260586&rid=29336&tvmazeid=1128&season=2&ep=1
 - Yielded one result, so fallback to regular title search is not triggered

/api?t=tvsearch&...&q=Cosmos Possible Worlds&season=1&ep=1
 - Yielded a dozen results

/api?t=tvsearch&...&q=Cosmos - Otros mundos&season=1&ep=1 
 - Yield no results, probably in foreign language category, if any
```

We should probably add a language token to the naming exceptions...

### DIFF
--- a/src/NzbDrone.Core/DataAugmentation/Scene/SceneMappingService.cs
+++ b/src/NzbDrone.Core/DataAugmentation/Scene/SceneMappingService.cs
@@ -15,11 +15,13 @@ namespace NzbDrone.Core.DataAugmentation.Scene
     public interface ISceneMappingService
     {
         List<string> GetSceneNames(int tvdbId, List<int> seasonNumbers, List<int> sceneSeasonNumbers);
+        List<SceneMapping> GetSceneMappings(int tvdbId, List<int> seasonNumbers);
         int? FindTvdbId(string sceneTitle, string releaseTitle);
         List<SceneMapping> FindByTvdbId(int tvdbId);
         SceneMapping FindSceneMapping(string sceneTitle, string releaseTitle);
         int? GetSceneSeasonNumber(string seriesTitle, string releaseTitle);
         int? GetTvdbSeasonNumber(string seriesTitle, string releaseTitle);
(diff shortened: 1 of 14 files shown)


src/NzbDrone.Core/DataAugmentation/Scene/SceneMappingService.cs
```diff
@@ -66,6 +68,24 @@ public List<string> GetSceneNames(int tvdbId, List<int> seasonNumbers, List<int>
...
+        public List<SceneMapping> GetSceneMappings(int tvdbId, List<int> seasonNumbers)
+        {
+            var mappings = FindByTvdbId(tvdbId);
+
+            if (mappings == null)
+            {
+                return new List<SceneMapping>();
+            }
+
+            seasonNumbers = seasonNumbers.ToList();
```
markus101, line 80: `seasonNumbers` is already a list.

Taloth: Yes, but I want to add one to it without modifying a list the caller might have.
But yeah, it's ugly code. I'll change it.

PR #3708, opened 2020-04-26
New: Use release quality source if not in downloaded file and resolution

markus101:
#### Database Migration
NO

#### Description
Allows Sonarr to augment the quality source of the folder/file with the quality source of the original release to better handle cases where the indexer's release name indicates the source is WEB-DL/WebRip, but the file name does not include the quality source. It will not override the quality if the indexer indicates it's WEB-DL, but the folder/file indicate it's WEBRip, though it would be possible to do that if we later decided to.

#### Todos
- [x] Tests
- [x] Documentation


src/NzbDrone.Core/MediaFiles/EpisodeImport/Aggregation/Aggregators/Augmenters/Quality/AugmentQualityFromReleaseName.cs
```diff
@@ -0,0 +1,58 @@
...
+            {
+                return null;
+            }
+
+            var fileQuality = localEpisode.FileEpisodeInfo?.Quality.Quality;
+            var folderQuality = localEpisode.FolderEpisodeInfo?.Quality.Quality;
+            var localQuality = folderQuality ?? fileQuality;
+
+            // Return early if the file or folder quality is not a television source (preferring the folder over the file)
+            if (localQuality?.Source != QualitySource.Television)
```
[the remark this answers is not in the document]
markus101: > If the filename contains HDTV then it must be HDTV. If the filename contains WEB-DL then it's WEB-DL. Both is Confidence=Tag and handled by those augmenters.
They should win, even if the indexer says otherwise. The indexer should only be used if the source isn't specified in the filename.

The problem is we don't have a differentiation there. `QualityParser` will return `HDTV720p` even if it only has 720p in the filename.

> Previously this would've been HDTV, and blocked on import. Now it'll be imported as bluray. I feel that might be a bit too eager.

Previously (before last week) it'd import as whatever quality the filename said it was, then if the same release with a different name was grabbed as an upgrade it'd reject that as having the same file size as existing.

> Example: [CR] Sailor Moon - 004 [480p][48CE2D0F] testcase. Resolution clearly is 480p. QualityDetectionSource will be 'Name' because of that, even though it's entirely unknown what the Source really is. We can guess, but we'd trust the release title over that guess.

For a lot of anime the source of this would actually be WEB too unfortunately.

> I don't think any of this is truly fixable without splitting up Source & Resolution in QualityModel into Source, Resolution, SourceConfidence, ResolutionConfidence and EffectiveQuality. This would allow QualityParser to specify what it parsed explicitly and what it inferred.
Leaving it to the aggregators to decide what information to trust the most.

Yeah, that's the crux of it.

> PS: A temporary workaround could be to have the aggregators do a secondary parse of the actual name, if information between indexer & filename/mediainfo conflicts. Saving us from having to change QualityModel throughout the program and the existing items in the history table.

This seems like the easiest way to clear this up and not block future improvements.

src/NzbDrone.Core.Test/ParserTests/QualityParserFixture.cs
```diff
@@ -381,7 +401,10 @@ public void should_parse_quality_from_name(string title)
         [TestCase("[CR] Sailor Moon - 004 [48CE2D0F].avi")]
         public void should_parse_quality_from_extension(string title)
         {
-            QualityParser.ParseQuality(title).QualityDetectionSource.Should().Be(QualityDetectionSource.Extension);
+            var result = QualityParser.ParseQuality(title);
+
+            result.SourceDetectionSource.Should().Be(QualityDetectionSource.Extension);
```
Taloth, line 406: Maybe a test with resolution + extension. Which should result in Source=Extension & Resolution=Name. combo of `should_parse_resolution_from_name` and `should_parse_quality_from_extension`.

src/NzbDrone.Core/MediaFiles/EpisodeImport/Aggregation/Aggregators/Augmenters/Quality/AugmentQualityFromFolder.cs
```diff
@@ -13,10 +15,18 @@ public AugmentQualityResult AugmentQuality(LocalEpisode localEpisode)
                 return null;
             }
 
+            var sourceConfidence = quality.SourceDetectionSource == QualityDetectionSource.Unknown
```
Taloth, line 18: Maybe checking against `Name` explicitly is better (returning Tag). In case the folder name contains an extension (returning Fallback).

src/NzbDrone.Core/MediaFiles/EpisodeImport/Aggregation/Aggregators/Augmenters/Quality/AugmentQualityFromReleaseName.cs
```diff
@@ -0,0 +1,58 @@
...
+            {
+                return null;
+            }
+
+            var fileQuality = localEpisode.FileEpisodeInfo?.Quality;
+            var folderQuality = localEpisode.FolderEpisodeInfo?.Quality;
+            var localQuality = folderQuality ?? fileQuality;
+
+            // Return early if the file or folder quality source was parsed.
+            if (localQuality?.SourceDetectionSource != QualityDetectionSource.Unknown)
```
Taloth, line 31: Same thing, this would consider `Extension`  trustworthy, which it isn't.

src/NzbDrone.Core/MediaFiles/EpisodeImport/Aggregation/Aggregators/Augmenters/Quality/AugmentQualityFromReleaseName.cs
```diff
@@ -0,0 +1,58 @@
...
+            if (history == null)
+            {
+                return null;
+            }
+
+            var historyQuality = QualityParser.ParseQuality(history.SourceTitle);
+
+            // If the resolution was parsed from the name instead of unknown or the extension and it matches the grab history use the source from the grabbed history.
+
+            if (localQuality.SourceDetectionSource != QualityDetectionSource.Name &&
```
Taloth, line 47: Can't releasename provide both source & resolution?

Note that this is essentially the same behavior as 'folder name' (which can be absent for single file torrents).

src/NzbDrone.Core/Parser/QualityParser.cs
```diff
@@ -72,7 +72,8 @@ public static QualityModel ParseQuality(string name)
                 try
                 {
                     result.Quality = MediaFileExtensions.GetQualityForExtension(Path.GetExtension(name));
-                    result.QualityDetectionSource = QualityDetectionSource.Extension;
+                    result.SourceDetectionSource = QualityDetectionSource.Extension;
+                    result.ResolutionDetectionSource= QualityDetectionSource.Extension;
```
Taloth, line 76: nitpick: missing space before =

src/NzbDrone.Core/MediaFiles/EpisodeImport/Aggregation/Aggregators/Augmenters/Quality/AugmentQualityFromReleaseName.cs
```diff
@@ -0,0 +1,58 @@
...
+            }
+
+            var history = _downloadHistoryService.GetLatestGrab(downloadClientItem.DownloadId);
+
+            if (history == null)
+            {
+                return null;
+            }
+
+            var historyQuality = QualityParser.ParseQuality(history.SourceTitle);
```
Taloth, line 43: If you compare `history.Quality.Quality == localQuality.Quality` then you can bail out immediately, saving a ParseQuality.
That way the following code is only executed in case of conflicts.

src/NzbDrone.Core/MediaFiles/EpisodeImport/Aggregation/Aggregators/Augmenters/Quality/AugmentQualityFromReleaseName.cs
```diff
@@ -0,0 +1,58 @@
...
+
+            // If the resolution was parsed from the name instead of unknown or the extension and it matches the grab history use the source from the grabbed history.
+
+            if (localQuality.SourceDetectionSource != QualityDetectionSource.Name &&
+                localQuality.ResolutionDetectionSource == QualityDetectionSource.Name &&
+                historyQuality.ResolutionDetectionSource == QualityDetectionSource.Name &&
+                localQuality.Quality.Resolution == historyQuality.Quality.Resolution)
```
Taloth, line 50: Why compare the resolution?

Scenario:
Grabbed Release is `Tom.S01E01.720p.WEB-DL-SNR`
Single file torrent containing `snr-tom101.avi`
No folder, so FolderInfo will be null.
FileInfo will say `SDTV`, with Source=Extension & Resolution=Extension.

history.Quality.Quality = WEBDL 720p, which doesn't match SDTV so triggers the ParseQuality.

This should result in WEBDL 720p Source=Name & Resolution=Name with Confidences=Tag.

src/NzbDrone.Core/Parser/QualityParser.cs
```diff
@@ -324,83 +332,131 @@ public static QualityModel ParseQualityName(string name)
...
             if (normalizedName.Contains("848x480"))
             {
+                result.ResolutionDetectionSource = QualityDetectionSource.Name;
+                
                 if (normalizedName.Contains("dvd"))
                 {
+                    result.SourceDetectionSource = QualityDetectionSource.Name;
                     result.Quality = Quality.DVD;
                 }
+                else if (normalizedName.Contains("bluray"))
```
Taloth, line 381: ContainsIgnoreCase (I removed the ToLower on normalizedName earlier today to allow a regex to do a case-sensitive match)

src/NzbDrone.Core/MediaFiles/EpisodeImport/Aggregation/Aggregators/Augmenters/Quality/AugmentQualityFromReleaseName.cs
markus101: Unfortunately, download history doesn't have quality stored for it.

src/NzbDrone.Core/MediaFiles/EpisodeImport/Aggregation/Aggregators/Augmenters/Quality/AugmentQualityFromReleaseName.cs
```diff
@@ -0,0 +1,89 @@
...
+                              historyQuality.ResolutionDetectionSource == QualityDetectionSource.Name
+                ? historyQuality.Quality.Resolution
+                : (int?)null;
+
+            var revision = historyQuality.SourceDetectionSource == QualityDetectionSource.Name &&
+                           historyQuality.ResolutionDetectionSource == QualityDetectionSource.Name
+                ? historyQuality.Revision
+                : null;
+
+            if (source != QualitySource.Unknown && resolution.HasValue)
```
markus101, line 71: Instead of deciding what we should return should we add a new confidence level and let `AggregateQuality` make the correct decision? That way we're not over complicating things here, we're just returning what we know from the release name.

Taloth: Adding a confidence level is not ideal because then you'd be combining confidence with priority (something we're already doing in AugmentQualityFromReleaseName to a limited degree).

Perhaps we should have AggregateQuality handle the 'equal confidence' scenario by preferring Folder over File over Release. Meaning that AugmentQualityFromReleaseName won't have to sort that out.

Technically, a conflicting Folder & File at the same confidence should produce an error. We don't now, and i'm ok with that. But it also means we're not going for the ideal solution anyway.

In practice, FolderInfo already contains the release title as provided by the indexer, so AugmentQualityFromReleaseName is only there to cover the scenario where it isn't available. So perhaps we should simplify it to deal solely with that scenario.
The separation of Source & Resolution already deals with a few of the other edge cases that we were trying to solve.

So I'm seeing the following options:
- either keep AugmentQualityFromReleaseName simple, based only on releasename and ignoring folder&file entirely
- or; adjust AugmentQualityFromReleaseName to return a result only if FolderInfo is absent
- or; adjust AugmentQualityFromReleaseName to return a result if Folder/File quality doesn't match the Grab quality as proposed earlier

Perhaps it's easier to simply sort the Augment* instances by a constant priority rather than SourceConfidence. If we always run through them by Folder, File, MediaInfo?, ReleaseName, Extension then the existing comparison based on confidence then ensures the right source/resolution is selected. Regardless of the chosen option above.

markus101: > In practice, FolderInfo already contains the release title as provided by the indexer

For usenet yes, but not not always for torrents. We sometimes see torrent indexers change the name to include the source in the release’s name, even BTN does this for some web/webrip releases in this case the file and folder name different.

1. This could override the webrip/web DL naming from the file if the release name altered that, but feels like the best way to do it, we just return what we know and let something else figure out what to do with it.

2. Not feasible given release name doesn’t always equal folder name.

3. Similar to 1, we could end up overriding what the file/folder explicitly says, but I feel like taking in other values and using them to drive what return muddies the water.

A specific order would help so we don’t override something we prefer more, coupled with (1) I think it solves the problems with overriding something explicit.

markus101:
Updated per our latest discussions.

src/NzbDrone.Core/MediaFiles/EpisodeImport/Aggregation/Aggregators/AggregateQuality.cs
```diff
@@ -19,55 +20,66 @@ public class AggregateQuality : IAggregateLocalEpisode
...
 
-        public LocalEpisode Aggregate(LocalEpisode localEpisode, bool otherFiles)
+        public LocalEpisode Aggregate(LocalEpisode localEpisode, DownloadClientItem downloadClientItem, bool otherFiles)
         {
-            var augmentedQualities = _augmentQualities.Select(a => a.AugmentQuality(localEpisode))
+            var augmentedQualities = _augmentQualities.OrderBy(a => a.Order)
+                                                      .Select(a => a.AugmentQuality(localEpisode, downloadClientItem))
                                                       .Where(a => a != null)
-                                                      .OrderBy(a => a.SourceConfidence);
+                                                      .OrderBy(a => a.SourceConfidence)
```
Taloth, line 28: Remove this. You can't sort by SourceConfidence and Order both. Sort by Order only and let the loop later sort out what the highest confidence is (it already is doing that).

Taloth:
ship it

PR #3735, opened 2020-05-06
Fixed: Don't lock command queue if updating is disabled

Qstick:
#### Database Migration
NO

#### Description
Splits the update command to ApplicationUpdate and ApplicationUpdateCheck. This will prevent the command queue from being locked down due to the isExclusive ApplicationUpdateCommand if the update command is awaiting something else to finish... especially for cases when we skip updates anyway.


Taloth:
I know I proposed the name... but ApplicationCheckUpdateCommand becomes "Application Check Update" in the UI.
Maybe swap that around to ApplicationUpdateCheckCommand as you mentioned in the initial comment. Or set `Name` in ctor to something like "Check For Application Update"?

Issue #3752, opened 2020-05-15
Tear out chown and merge chmod flags


Taloth:
**Describe the bug**
chown never works. We should remove it for v3.
Produces support question from ppl that expect the app to magically getting root and fix their permission issues.

File & Folder probably don't need separate masks either.
If a file should be accessible by some group then so should the folder.
When applying chmod to files , simply clear out 'x'.

xelra:
Chown is very much needed for everyone who set up permissions in his library correctly. Today my setup broke over this.
We need this back asap.

The reason why you believed that chown isn't working is probably because only the superuser can change the owner of a file. But chown also does group.

So you would just leave owner blank and specify the group. As soon as you specify a user, chown throws an error. If you want to make it more user-friendly, just remove the field to enter an owner and set it to default to blank.

xelra:
There's more. I also saw that the modes for files and directories have been compressed. This is not common practice on any posix system.
I agree that the separation doesn't make sense, but it should now ask for `umask` instead of arbitrarily asking for either file or directory mode and then adding/removing execution bits for the other.
That would be expected behavior within the POSIX toolset.

Taloth:
I agree on the umask, but ppl shouldn't be needing these options. Setup the umask and uid/gid of the sonarr process any any new file/folder will get created properly. Do the same on the downloader.
Relying on the app to fiddle around with chown is a bad practice in my opinion.

I wanted the settings gone now while v3 is in beta, precisely because it's a breaking change.

If there's a scenario in which this breaks where setting up process umask/uid/gid doesn't work, let me know.

Taloth:
> In my own setup it's a little bit different, since the download location is on a remotely mapped drive, which is mounted with the media gid. So Sonarr gets access to these files that way.

Anyone with a local storage and separate services will not have that 'workaround'. Remote network shares that don't use unix permissions the same way are kinda out of scope because imo.
How would it work in your case if that network share was just a local drive?

Without being owner, sonarr can't change permissions nor the group.
Sonarr would be required to copy the file first, to gain ownership.

So your approach works, ONLY because you use separate filesystems for your download and library locations. Forcing the copy and thus giving ownership of the moved file to sonarr. Which is then capable of adjusting the permissions and group.
How would it work if both download and library locations are on the same filesystem?

Imo there are still several approaches the user can take that does not sacrifice security:
[...]

Issue #3161, opened 2019-06-12
"Folder chmod mode" setting not applied when moving series


jobrien2001:
**Describe the bug**
I have "Folder chmod mode" set to 777, and works as intended in automation.  But if i decide to edit a series, change its folder, and choose to let Sonarr move the files, folder permissions of the new folders will be 755.

**Logs**
19-6-12 12:34:46.5|Info|MoveSeriesService|Moving Test Series from '/media.tv/Test Series' to '/media.tv.old/Test Series'

**System Information**
 - Sonarr Version: 3.0.1.513
 - Operating System: UNRAID OS

markus101:
Please fill out the entire bug template.

jobrien2001:
Done, thank you.

jobrien2001:
Also, want to add that this bug also shows on files going to the recycle bin.  Folders in the recycle bin default to 755 and files to 644.

markus101:
Not sure that's really a bug, the intent was only to apply it when importing to the series folder, moving whole series is new, but the recycle bin has never had that functionality.

jobrien2001:
I was expecting the original permissions of the folder and files would be retained when sonarr is moving files around.

As it stands, if i move a series through the interface then i have to run a terminal and chmod after.

Taloth:
Don't set chmod/chown in Sonarr it's only there to workaround bad setups, just use umask. A process umask specifies the default permission for new files&folders.
I have no idea what how the unraid package is configured so you'll have to figure out where to change the umask. Our systemd [unit defaults to 002](https://github.com/Sonarr/Sonarr/blob/phantom-develop/distribution/debian/sonarr.service#L8), which means 775. So unraid probably uses a different startup config.

jobrien2001:
I see, i run the linuxserver docker so perhaps i will look for more info there.  Thanks for your explanation.

PR #3760, opened 2020-05-17
Remove chown and merge chmod flags

principis:
#### Database Migration
NO

#### Description
Removed all `chown` operations because they never work.
I haven't touched `chmod` because I doubt the masks should be merged.

I edited both Sonarr.Api.V3 and NzbDrone.Api because I don't know if it's still used for V3, let me know if I should remove my edits.

This is my first PR here, so let me know what I need to change. :)

#### Todos
- [ ] Tests
- [ ] Documentation

* #3752

### DIFF
--- a/src/NzbDrone.Api/Config/MediaManagementConfigModule.cs
+++ b/src/NzbDrone.Api/Config/MediaManagementConfigModule.cs
@@ -7,10 +7,10 @@ namespace NzbDrone.Api.Config
 {
     public class MediaManagementConfigModule : NzbDroneConfigModule<MediaManagementConfigResource>
     {
-        public MediaManagementConfigModule(IConfigService configService, PathExistsValidator pathExistsValidator)
+        public MediaManagementConfigModule(IConfigService configService, PathExistsValidator pathExistsValidator, FileChmodValidator fileChmodValidator)
             : base(configService)
         {
-            SharedValidator.RuleFor(c => c.FileChmod).NotEmpty().ValidFileChmod();
+            SharedValidator.RuleFor(c => c.FileChmod).SetValidator(fileChmodValidator).When(c => !string.IsNullOrEmpty(c.FileChmod));
             SharedValidator.RuleFor(c => c.RecycleBin).IsValidPath().SetValidator(pathExistsValidator).When(c => !string.IsNullOrWhiteSpace(c.RecycleBin));
         }
 
--- a/src/NzbDrone.Api/Config/MediaManagementConfigModule.cs
+++ b/src/NzbDrone.Api/Config/MediaManagementConfigModule.cs
@@ -1,5 +1,6 @@
 ﻿using FluentValidation;
 using NzbDrone.Core.Configuration;
+using NzbDrone.Core.Validation;
 using NzbDrone.Core.Validation.Paths;
 
 namespace NzbDrone.Api.Config
@@ -9,7 +10,7 @@ namespace NzbDrone.Api.Config
         public MediaManagementConfigModule(IConfigService configService, PathExistsValidator pathExistsValidator)
             : base(configService)
         {
-            SharedValidator.RuleFor(c => c.FileChmod).NotEmpty();
+            SharedValidator.RuleFor(c => c.FileChmod).NotEmpty().ValidFileChmod();
             SharedValidator.RuleFor(c => c.RecycleBin).IsValidPath().SetValidator(pathExistsValidator).When(c => !string.IsNullOrWhiteSpace(c.RecycleBin));
         }
 
--- a/src/NzbDrone.Api/Config/MediaManagementConfigModule.cs
+++ b/src/NzbDrone.Api/Config/MediaManagementConfigModule.cs
@@ -1,4 +1,5 @@
 ﻿using FluentValidation;
+using NzbDrone.Common.EnvironmentInfo;
 using NzbDrone.Core.Configuration;
 using NzbDrone.Core.Validation;
 using NzbDrone.Core.Validation.Paths;
@@ -10,7 +11,7 @@ namespace NzbDrone.Api.Config
         public MediaManagementConfigModule(IConfigService configService, PathExistsValidator pathExistsValidator, FileChmodValidator fileChmodValidator)
             : base(configService)
         {
-            SharedValidator.RuleFor(c => c.FileChmod).SetValidator(fileChmodValidator).When(c => !string.IsNullOrEmpty(c.FileChmod));
+            SharedValidator.RuleFor(c => c.FileChmod).SetValidator(fileChmodValidator).When(c => !string.IsNullOrEmpty(c.FileChmod) && PlatformInfo.IsMono);
             SharedValidator.RuleFor(c => c.RecycleBin).IsValidPath().SetValidator(pathExistsValidator).When(c => !string.IsNullOrWhiteSpace(c.RecycleBin));
         }
 
--- a/src/NzbDrone.Api/Config/MediaManagementConfigModule.cs
+++ b/src/NzbDrone.Api/Config/MediaManagementConfigModule.cs
@@ -10,7 +10,6 @@ namespace NzbDrone.Api.Config
             : base(configService)
         {
             SharedValidator.RuleFor(c => c.FileChmod).NotEmpty();
-            SharedValidator.RuleFor(c => c.FolderChmod).NotEmpty();
             SharedValidator.RuleFor(c => c.RecycleBin).IsValidPath().SetValidator(pathExistsValidator).When(c => !string.IsNullOrWhiteSpace(c.RecycleBin));
         }
 
--- a/src/NzbDrone.Common/Disk/DiskProviderBase.cs
+++ b/src/NzbDrone.Common/Disk/DiskProviderBase.cs
@@ -5,6 +5,7 @@ using System.Linq;
 using System.Security.AccessControl;
 using System.Security.Principal;
 using System.Text;
+using Mono.Unix.Native;
 using NLog;
 using NzbDrone.Common.EnsureThat;
 using NzbDrone.Common.EnvironmentInfo;
@@ -361,29 +362,11 @@ namespace NzbDrone.Common.Disk
 
         }
 
-        protected static string GetFolderPermissions(string filePermissions)
+        protected static FilePermissions GetFolderPermissions(FilePermissions permissions)
         {
-            var builder = new StringBuilder();
+            permissions |= (FilePermissions) ((int) (permissions & (FilePermissions.S_IRUSR | FilePermissions.S_IRGRP | FilePermissions.S_IROTH)) >> 2);
 
-            if (filePermissions.Length == 4)
-            {
-                filePermissions = filePermissions.Substring(1);
-                builder.Append("0");
-            }
-
-            foreach (char c in filePermissions)
-            {
-                int value = (int) char.GetNumericValue(c);
-
-                if (value > 0 && value % 2 == 0)
-                {
-                    value++;
-                }
-
-                builder.Append(value);
-            }
-
-            return builder.ToString();
+            return permissions;
         }
 
         private static void RemoveReadOnly(string path)
--- a/src/NzbDrone.Common/Disk/DiskProviderBase.cs
+++ b/src/NzbDrone.Common/Disk/DiskProviderBase.cs
@@ -31,8 +31,8 @@ namespace NzbDrone.Common.Disk
 
         public abstract long? GetAvailableSpace(string path);
         public abstract void InheritFolderPermissions(string filename);
-        public abstract void SetPermissions(string path, string mask, string user, string group);
-        public abstract void CopyPermissions(string sourcePath, string targetPath, bool includeOwner);
+        public abstract void SetPermissions(string path, string mask);
+        public abstract void CopyPermissions(string sourcePath, string targetPath);
         public abstract long? GetTotalSize(string path);
 
         public DateTime FolderGetCreationTime(string path)
--- a/src/NzbDrone.Common/Disk/DiskProviderBase.cs
+++ b/src/NzbDrone.Common/Disk/DiskProviderBase.cs
@@ -4,7 +4,6 @@ using System.IO;
 using System.Linq;
 using System.Security.AccessControl;
 using System.Security.Principal;
-using System.Text;
 using Mono.Unix.Native;
 using NLog;
 using NzbDrone.Common.EnsureThat;
--- a/src/NzbDrone.Common/Disk/DiskProviderBase.cs
+++ b/src/NzbDrone.Common/Disk/DiskProviderBase.cs
@@ -4,7 +4,6 @@ using System.IO;
 using System.Linq;
 using System.Security.AccessControl;
 using System.Security.Principal;
-using Mono.Unix.Native;
 using NLog;
 using NzbDrone.Common.EnsureThat;
 using NzbDrone.Common.EnvironmentInfo;
@@ -361,13 +360,6 @@ namespace NzbDrone.Common.Disk
 
         }
 
-        protected static FilePermissions GetFolderPermissions(FilePermissions permissions)
-        {
-            permissions |= (FilePermissions) ((int) (permissions & (FilePermissions.S_IRUSR | FilePermissions.S_IRGRP | FilePermissions.S_IROTH)) >> 2);
-
-            return permissions;
-        }
-
         private static void RemoveReadOnly(string path)
         {
             if (File.Exists(path))
@@ -522,6 +514,11 @@ namespace NzbDrone.Common.Disk
             {
                 stream.CopyTo(fileStream);
             }
+        }
+
+        public virtual bool IsValidFilePermissionMask(string mask)
+        {
+            throw new NotSupportedException();
         }
     }
 }
--- a/src/NzbDrone.Common/Disk/DiskProviderBase.cs
+++ b/src/NzbDrone.Common/Disk/DiskProviderBase.cs
@@ -4,6 +4,7 @@ using System.IO;
 using System.Linq;
 using System.Security.AccessControl;
 using System.Security.Principal;
+using System.Text;
 using NLog;
 using NzbDrone.Common.EnsureThat;
 using NzbDrone.Common.EnvironmentInfo;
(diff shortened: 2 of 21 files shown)


src/NzbDrone.Mono/Disk/DiskProvider.cs
```diff
@@ -89,11 +97,6 @@ public override void CopyPermissions(string sourcePath, string targetPath, bool
                 {
                     Syscall.chmod(targetPath, srcStat.st_mode);
                 }
-
-                if (includeOwner && (srcStat.st_uid != tgtStat.st_uid || srcStat.st_gid != tgtStat.st_gid))
```
Taloth, line 93: Keep this in. I think it was added to deal with a particular edgecase when copying folders.
It'll still not be effective in much scenarios, lacking appropriate permissions, but it's not user facing.

Taloth: Actually... this was the issue https://github.com/Sonarr/Sonarr/issues/3161
'Implemented' in v3 last year.

I'm torn a bit between tearing it out and leaving it in .... 😕 

... yup, tear it out.

principis:
I didn't see the benefit because it removes a feature and adds validation, but I understand the reasoning. I'll update the pr.

principis:
@Taloth I've updated the pr with the chmod changes. I've also added some tests.

I'm not sure if I should change `FileChmod` to just `Chmod` (in the code, not ui) it could be better, but I left it for now.

src/NzbDrone.Common/Disk/DiskProviderBase.cs
```diff
@@ -360,6 +361,31 @@ public void SetPermissions(string filename, WellKnownSidType accountSid, FileSys
 
         }
 
+        protected static string GetFolderPermissions(string filePermissions)
+        {
+            var builder = new StringBuilder();
```
Taloth, line 366: This works. But I think it's cleaner to manipulate the output of `FilePermissions fp = NativeConvert.FromOctalPermissionString(mask)` that we're already getting.
you can even just mask out the 3 'r' (see FilePermissions), shift them down 2 bits and or.

src/NzbDrone.Mono.Test/DiskProviderTests/DiskProviderFixture.cs
```diff
@@ -151,11 +151,73 @@ public void should_copy_folder_permissions()
...
+
+            Subject.SetPermissions(tempPath, "775");
+            Syscall.stat(tempPath, out fileStat);
+            NativeConvert.ToOctalPermissionString(fileStat.st_mode).Should().Be("0775");
+
+            Subject.SetPermissions(tempPath, "640");
+            Syscall.stat(tempPath, out fileStat);
+            NativeConvert.ToOctalPermissionString(fileStat.st_mode).Should().Be("0750");
+
+            Subject.SetPermissions(tempPath, "0041");
```
Taloth, line 218: Doesn't this remove owner permissions? Might need a bit of code to cleanup the test folder afterwards.

principis: Yes, but it should be fine because it's a temp file and the file can be removed because the parent folder should have enough permissions.
I added that test, although it's unrealistic, to make sure the function works. I'll remove the file afterwards.

src/NzbDrone.Common/Disk/DiskProviderBase.cs
principis: Sure, I'll change it.

Taloth: forgot to mention that you also need to make sure the user doesn't add executable flag to filechmod. Either through validation, or masking them out later.

principis: I'll improve the validation then, because currently it only checks if it's not empty.

src/NzbDrone.Mono.Test/DiskProviderTests/DiskProviderFixture.cs
Taloth: Paths retrieved via GetTempFilePath are automatically removed after the test. So if the permissions don't prevent removal then no further changes are needed.

principis:
Sorry for the delay, I'm quite busy with upcoming exams.

I've changed `GetFolderPermissions` to use `FilePermissions` and also added proper validation. I changed the default value of `0644` to `644` because the 0 isn't necessary and may be confusing for some people (new unix users which aren't aware of the special bits). Alongside prohibiting the executable bit, I also prohibited the Setuid, Setgid, and Sticky Bits, because I think media files should never have those set. Let me know if I'm wrong! :)

src/NzbDrone.Core/Configuration/ConfigService.cs
```diff
@@ -254,32 +254,11 @@ public bool SetPermissionsLinux
 
         public string FileChmod
         {
-            get { return GetValue("FileChmod", "0644"); }
+            get { return GetValue("FileChmod", "644"); }
```
Taloth, line 257: Note: `0` prefix is used in some programming languages to indicate 'octal notation'. So that's likely why it was there.

src/NzbDrone.Core/Datastore/Migration/140_remove_folderchmod_config.cs
```diff
@@ -0,0 +1,15 @@
...
+using NzbDrone.Core.Datastore.Migration.Framework;
+
+namespace NzbDrone.Core.Datastore.Migration
+{
+    [Migration(140)]
+    public class remove_folderchmod_config : NzbDroneMigrationBase
+    {
+        protected override void MainDbUpgrade()
+        {
+            Execute.Sql("DELETE FROM Config WHERE Key = 'folderchmod'");
```
Taloth, line 12: Note sure if we should merge it with filechmod. Like `filechmod = (filechmod | folderchmod) & ~(execute flags)`.
I don't expect anyone to have them configured differently, so it's probably not needed.

src/NzbDrone.Core/Sonarr.Core.csproj
```diff
@@ -28,6 +28,9 @@
     <Reference Include="Growl.CoreLibrary">
       <HintPath>..\Libraries\Growl.CoreLibrary.dll</HintPath>
     </Reference>
+    <Reference Include="Mono.Posix">
```
Taloth, line 31: No Mono.Posix in Sonarr.Core, hide that in Sonarr.Mono

src/NzbDrone.Core/Datastore/Migration/140_remove_folderchmod_config.cs
Taloth: Also delete chown user&group

src/NzbDrone.Core/Validation/FileChmodValidator.cs
```diff
@@ -0,0 +1,27 @@
...
+{
+    public static class FileChmodValidator
+    {
+        public static IRuleBuilderOptions<T, string> ValidFileChmod<T>(this IRuleBuilder<T, string> ruleBuilder)
+        {
+            return ruleBuilder.Must(x =>
+            {
+                try
+                {
+                    var permissions = NativeConvert.FromOctalPermissionString(x);
```
Taloth, line 15: I understand the conundrum, the only thing I can think of is to either add a IsValidFilePermissionMask function to DiskProvider or copy the flags and do it ourselves. First option is easier I think, DiskProviderBase would simply return throw new NotSupportedException(), with the real implementation in Sonarr.Mono. 
You'd need a non-static validator, similar to pathExistsValidator, to access IDiskProvider.

principis:
> A bit of a review delay, but was a bit busy at work.
> 
> Biggest issue is mono assembly leaking out of containment :)

I'm quite inexperienced with this scale of C# project, I thought it may be a problem, but didn't know how to deal with it. Thank you! :) (I actually forgot to commit the csproj in the first instance.)

I have an exam coming up, so I'll fix the requested changes on Friday.

Taloth:
Thank you for your effort in this PR. I realize it grew from 1 small up-for-grabs to something a bit more complicated.

Good luck with the exam.

src/NzbDrone.Core/Datastore/Migration/140_remove_folderchmod_config.cs
principis: I think they shouldn't be merged, it may cause unexpected issues.
Sorry forgot the chown :)

src/NzbDrone.Core/Configuration/ConfigService.cs
principis: I changed it back.

principis:
No problem, I learned things and enjoyed doing it. :)

I've improved the validation by also checking if the user can read and write to the file. I think that should always be the case for Sonarr to function properly. (?)
Please let me know if I overlooked something.

src/NzbDrone.Core/Validation/FileChmodValidator.cs
principis: Should be fixed now. :)

src/NzbDrone.Core/Datastore/Migration/140_remove_chown_and_folderchmod_config.cs
```diff
@@ -0,0 +1,17 @@
...
+namespace NzbDrone.Core.Datastore.Migration
+{
+    [Migration(140)]
+    public class remove_chown_and_folderchmod_config : NzbDroneMigrationBase
+    {
+        protected override void MainDbUpgrade()
+        {
+            Execute.Sql("DELETE FROM Config WHERE Key = 'folderchmod'");
+            Execute.Sql("DELETE FROM Config WHERE Key = 'chownuser'");
+            Execute.Sql("DELETE FROM Config WHERE KEY = 'chowngroup'");
```
Taloth, line 14: Nitpick: 'KEY' -> 'Key'...
Also consider `Key IN ('folderchmod', 'chownuser', 'chowngroup')` turns it into a single query

src/NzbDrone.Mono/Disk/DiskProvider.cs
```diff
@@ -72,13 +72,49 @@ public override void InheritFolderPermissions(string filename)
...
+
+            return permissions;
+        }
+
+        public override bool IsValidFilePermissionMask(string mask)
+        {
+            try
+            {
+                var permissions = NativeConvert.FromOctalPermissionString(mask);
+                return (permissions & (FilePermissions.S_ISGID | FilePermissions.S_ISUID | FilePermissions.S_ISVTX |
```
Taloth, line 106: I had to read it twice to understand what you were really checking, so probably better to split this up.
Separately check for setuid/setuid/sticky, return false otherwise.
Separately check no execute flags, return false otherwise.
Then check for rw for user.
That should make it easier to follow.

src/NzbDrone.Api/Config/MediaManagementConfigModule.cs
```diff
@@ -1,16 +1,16 @@
...
 {
     public class MediaManagementConfigModule : NzbDroneConfigModule<MediaManagementConfigResource>
     {
-        public MediaManagementConfigModule(IConfigService configService, PathExistsValidator pathExistsValidator)
+        public MediaManagementConfigModule(IConfigService configService, PathExistsValidator pathExistsValidator, FileChmodValidator fileChmodValidator)
             : base(configService)
         {
-            SharedValidator.RuleFor(c => c.FileChmod).NotEmpty();
-            SharedValidator.RuleFor(c => c.FolderChmod).NotEmpty();
+            SharedValidator.RuleFor(c => c.FileChmod).SetValidator(fileChmodValidator).When(c => !string.IsNullOrEmpty(c.FileChmod));
```
Taloth, line 13: I was thinking about this and on windows the validator would still run and thus throw a NotSupportedException. probably better to add an OsInfo.IsMono in the When condition.

src/Sonarr.Api.V3/Config/MediaManagementConfigModule.cs
```diff
@@ -1,17 +1,17 @@
...
     public class MediaManagementConfigModule : SonarrConfigModule<MediaManagementConfigResource>
     {
-        public MediaManagementConfigModule(IConfigService configService, PathExistsValidator pathExistsValidator)
+        public MediaManagementConfigModule(IConfigService configService, PathExistsValidator pathExistsValidator, FileChmodValidator fileChmodValidator)
             : base(configService)
         {
             SharedValidator.RuleFor(c => c.RecycleBinCleanupDays).GreaterThanOrEqualTo(0);
-            SharedValidator.RuleFor(c => c.FileChmod).NotEmpty();
-            SharedValidator.RuleFor(c => c.FolderChmod).NotEmpty();
+            SharedValidator.RuleFor(c => c.FileChmod).SetValidator(fileChmodValidator).When(c => !string.IsNullOrEmpty(c.FileChmod));
```
Taloth, line 14: Same, OsInfo.IsMono guard.

src/NzbDrone.Core/Datastore/Migration/140_remove_chown_and_folderchmod_config.cs
principis: Sorry typo :/

principis:
Ok added the changes. I'll rebase to solve conflicts.

Taloth:
Thank you for your contribution.

imathew:
Hi all, I understand that chown wasn't working well so its removal makes sense, but it was useful to have chmod be separate for files and directories. In my case, for example, I always want directories to be 2775, and files 0664.

I would usually leave the directory perms alone (they're created with correct permissions at the OS level) and have Sonarr set the file permissions alone. I can't use this new merged setting, though, as it sets new directories to 0775 (overriding my 2775). This has flow-on effects to groups on my system, as subsequent season folders are then created with incorrect permissions, locking them out.

Thanks, appreciate your work!

Taloth:
> though, as it sets new directories to 0775 (overriding my 2775).

That was fixed like yesterday.

imathew:
> 
> 
> > though, as it sets new directories to 0775 (overriding my 2775).
> 
> That was fixed like yesterday.

Ha, sorry! I've been troubleshooting my file issues today and just figured it out, and found this. Sorry I should have searched better.

Taloth:
It's ok, the timing was just incredible :smile: Be sure to update Sonarr to at least 3.0.4.1016

d00nicus:
As someone who the chown functionality was working well for, the total removal of it is a poor outcome - I've gone from a working setup to needing to execute chown on my library on a regular basis if I want to stay on the current version.

I was under the impression the point of advanced options being hidden by default was to hide them from new/inexperienced users, a user pull request that unilaterally removes something already hidden under advanced simply because it doesn't work for *some* users (which I'm assuming is a group that @principis falls into) seems sub-optimal at best. 

For the time being I've rolled back to 3.0.3.750 (as the only Linux build I have saved from before this was merged) but I'm glad to at least have found the specific commit responsible for this so that I can look at undoing it, but it'd be nice to not have to do that every time I need/want to update Sonarr however.

markus101:
Sonarr still supports changing the group, just not the user.

> because it doesn't work for some user

It only works for _some_ users, specifically users that are running Sonarr as root, which is not recommended on any system. Removing is removes the foot gun since it doesn't work in more cases than it does work. If you have a need to change the owner and you're running as root you can use a custom script to achieve the same thing.

d00nicus:
Scripts (or forking) it is then I guess, when I next update to a version newer than this commit anyhow - unless anything vital gets fixed that affects me, or a feature I really need is added I don't see a rush to update from 3.0.3.750 for the time being.

Thanks for responding, even if not the response I was hoping for - I appreciate it.

principis:
> I was under the impression the point of advanced options being hidden by default was to hide them from new/inexperienced users, a user pull request that unilaterally removes something already hidden under advanced simply because it doesn't work for _some_ users (which I'm assuming is a group that @principis falls into) seems sub-optimal at best.

I'm sorry this didn't work out as expected. I like Sonarr, wanted to contribute and saw the issue so I implemented it as requested there.
_I am confused what was wrong with chmod tho, I've implemented it such that 2775 should work fine... (I think)_

As @markus101 already said, chown only works when Sonarr is run as root, so I still stand behind removing it.

d00nicus:
I appreciate your position on this, although I do think at minimum any patch that alters the configured behaviour in regards to file permissions/ownership in this way without an opt-out needs to display some sort of notification in the GUI that those settings have been changed/reverted when the config is set to a non-default value rather than waiting for affected users to discover the mis-set files on their own.

I can respect *why* you did it, but I strongly believe it shouldn't have been a silent change to the running configuration.

Taloth:
> I like Sonarr, wanted to contribute and saw the issue so I implemented it as requested there.

@principis You did, the contribution worked exactly as it was supposed to.

We did get quite a few support questions from people not understanding what a file mask of 644 meant, and that if they wanted 755 for the folder they would need to set perms to 644 and automatically get 755 for folders. So in a later change i swapped the file-only mask to folder-only and added a lot of ui sugar to show the various masks and permissions as well as the common presets. Functionally the same, just presented differently to the user. I haven't seen support questions about it since.

> I am confused what was wrong with chmod tho

As for 2775, that was an pre-existing bug in sonarr, as far back as v2. Sonarr would remove the special bits (sticky/setgid bits) from existing folders when the mask was set to '775' preventing setgid propagation to new subfolders. One of our regular users pointed that out and I fixed it as well. In v2 some users set 644 to files and 2755 for folders to work around that, such as imathew posted above. (FYI the cmdline `chmod 0775` and `chmod 755` are functionally different since the former clears the special bits, where the later does not)
We no longer allow setting setgid bits in sonarr, but simultaneously we no longer discard it from existing folders thus properly propagating the setgid on the root folder to series->season folders. So people can now just use the cmdline to set setgid on the appropriate library folders once, and have them propagate properly as sonarr creates new folders.

I have to note that this change was made in Sonarr v3 beta, specifically because it's a potentially breaking change. And beta users are explicitly notified on the sonarr.tv website that we can and will be make breaking changes during the v3 beta phase.

@d00nicus If you're considering forking or running your own sonarr version, I strongly encourage you to first reevaluate how you're doing permissions.
Talk to us on discord if you want to clarify what your particular workflow is, we have plenty of people willing to discuss it and offer advise. It's entirely possible you have an unique workflow, but I'm sure an alternative approach is available that doesn't require you running your own sonarr version.

d00nicus:
> I have to note that this change was made in Sonarr v3 beta, specifically because it's a potentially breaking change. And beta users are explicitly notified on the sonarr.tv website that we can and will be make breaking changes during the v3 beta phase.

I understand (and respect the why - breaking changes happen, and they're often unavoidable) - having not had my server running for a while and recently updating it, it would just have been nice to *know* that a breaking change had altered my configuration so that I could change my workflow accordingly, the update screen in Sonarr doesn't show changelogs that far back (would definitely nice to be able to see more there), so the first I knew was when permissions issues started appearing.

I'll likely pop onto the Discord in a few days, as I'm not going to be around much over the weekend/Monday, although regardless of outcome there, as a developer forking a copy to tinker with for curiosity's sake still holds an interest

principis:
@Taloth Thank you for your explanation :smile:

PR #3398, opened 2019-11-15
New: List Support

Qstick:
#### Database Migration
YES [ List Table, List Status Table, List Exclusions Table ]

#### Description
Adds basic functionality for list support to backend and frontend, no list options as of yet

Discussion points:
- Which lists options do we want (TVDB, Trakt, IMDB, SickRage Import, Second Sonarr Instance?)
- Any Skyhook integration needed for any list options? Mapping from IMDB, TVDB Caching, Trakt Auth API key hiding, etc.. 
- Should we take monitor status from the list in certain cases? (SickRage Import, or Sonarr Sync)
- Are you guys content with naming? Does something like Integrations make more sense?
- Should List Sync Job Interval be User Adjustable?

Current State:
image

#### Todos
- [x] Tests

#### Issues Fixed or Closed by this PR
Relate to #276 
*

PR #3899, opened 2020-08-05
New: Index priority

Dtaggart:
#### Database Migration
Yes

#### Description
Add indexer priority setting that allows users to set a priority on specific indexers.

#### Implementation 
- Priority setting is a integer from 1 to 50.
- Default is 25
- Already defined indexers will get a set priority of 25, this will essential make the indexers behave the same way it has always.
- Priority Tag will only appear if one or more indexers priority is not set to default 25

Added to download decision comparer
- CompareQuality
- CompareLanguage
- ComparePreferredWordScore
- CompareProtocol
- **_CompareIndexerPriority_** <-*New*
- CompareEpisodeCount
- CompareEpisodeNumber
- ComparePeersIfTorrent
- CompareAgeIfUsenet
- CompareSize

Added the setting to the indexer card listed tags 
image

image

#### To dos
- [x] Tests
Partially addresses but not fully
* https://github.com/Sonarr/Sonarr/issues/320


src/NzbDrone.Core/Indexers/IIndexerSettings.cs
```diff
@@ -5,5 +5,7 @@ namespace NzbDrone.Core.Indexers
     public interface IIndexerSettings : IProviderConfig
     {
         string BaseUrl { get; set; }
+
+        int SearchPriority { get; set; }
```
markus101, line 9: This should be part of the definition, not part of settings, same as download clients. It should also just be called priority, since it doesn't strictly apply to searching.

src/NzbDrone.Core/DecisionEngine/DownloadDecisionComparer.cs
```diff
@@ -28,6 +28,7 @@ public int Compare(DownloadDecision x, DownloadDecision y)
         {
             var comparers = new List<CompareDelegate>
             {
+                CompareSearchPriority,
```
markus101, line 31: Indexer priority is not more important than quality, it should be just after protocol I think.

frontend/src/Utilities/Array/sortBySearchPriorityThenName.js
```diff
@@ -0,0 +1,5 @@
+function sortBySearchPriorityThenName(a, b) {
+    return (a.fields.find(({name}) => name === 'searchPriority').value - b.fields.find(({name}) => name === 'searchPriority').value) || a.name.localeCompare(b.name);
+  }
+  
+  export default sortBySearchPriorityThenName;
```
markus101, line 5: Nit: missing new line.

src/NzbDrone.Core/Parser/Model/ReleaseInfo.cs
```diff
@@ -14,6 +14,7 @@ public class ReleaseInfo
         public string CommentUrl { get; set; }
         public int IndexerId { get; set; }
         public string Indexer { get; set; }
+        public int SearchPriority { get; set; }
```
markus101, line 17: `IndexerPriority` or link the indexer in another way.

src/NzbDrone.Core/Indexers/BroadcastheNet/BroadcastheNetSettings.cs
```diff
@@ -23,6 +25,7 @@ public BroadcastheNetSettings()
         {
             BaseUrl = "http://api.broadcasthe.net/";
             MinimumSeeders = IndexerDefaults.MINIMUM_SEEDERS;
+            SearchPriority = IndexerDefaults.SEARCH_PRIORITY;
```
markus101, line 28: Moving this off of indexer settings and into the definition (because it applies to all indexers regardless of protocol, unlike minimum seeders), means you don't need to copy paste all this code to each indexer.

src/NzbDrone.Core/Indexers/IIndexerSettings.cs
Dtaggart: Didnt even know DC had a priority, will have a look and change accordingly

Dtaggart:
@markus101 Sorry it took me a few days, had to sit down and understand how you had tests setup. Is there any documentation around search priority that needs to be updated? I didn't see any in the wiki. Thanks

src/NzbDrone.Core/Indexers/IndexerDefinition.cs
```diff
@@ -10,6 +10,7 @@ public class IndexerDefinition : ProviderDefinition
         public DownloadProtocol Protocol { get; set; }
         public bool SupportsRss { get; set; }
         public bool SupportsSearch { get; set; }
+        public int Priority { get; set; } = IndexerDefaults.PRIORITY;
```
markus101, line 13: Set the priority directly here.
I don't know that it makes sense to set to the lowest by default.

frontend/src/Settings/Indexers/Indexers/Indexer.js
```diff
@@ -113,6 +114,10 @@ class Indexer extends Component {
               </Label>
           }
 
+          <Label kind={kinds.SUCCESS}>
```
markus101, line 117: This shouldn't be green, it's not on/off like the other settings being enabled. `DEFAULT` is probably good here.

frontend/src/Settings/Indexers/Indexers/EditIndexerModalContent.js
```diff
@@ -131,7 +132,20 @@ function EditIndexerModalContent(props) {
                   );
                 })
               }
+              
+              <FormGroup>
+                <FormLabel>Indexer Priority</FormLabel>
 
+                <FormInputGroup
```
markus101, line 139: This should be treated as an advanced option.

src/NzbDrone.Core/Indexers/IndexerDefinition.cs
Dtaggart: Well I was thinking in terms that if someone wants to use it in a existing environment, lets say they have 6. If they want to make 1 indexer higher priority you only need to change that one indexer. If you were to say set them all to 1, to make one indexer a priority you need to change the other 5 indexers to something higher than 1.

src/NzbDrone.Core/Indexers/IndexerDefaults.cs
```diff
@@ -3,5 +3,7 @@
     public static class IndexerDefaults
     {
         public const int MINIMUM_SEEDERS = 1;
+
+        public const int PRIORITY = 100;
```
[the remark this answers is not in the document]
Dtaggart: Also used in the migration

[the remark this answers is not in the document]
Dtaggart: I guess thats not a good idea, seeing how future migrations would operate on the field in a linear fashion and expect that it would never change

frontend/src/Settings/Indexers/Indexers/IndexersConnector.js
```diff
@@ -2,14 +2,14 @@ import PropTypes from 'prop-types';
...
 import { createSelector } from 'reselect';
-import sortByName from 'Utilities/Array/sortByName';
+import sortBySearchPriorityThenName from 'Utilities/Array/sortBySearchPriorityThenName';
 import createSortedSectionSelector from 'Store/Selectors/createSortedSectionSelector';
 import { fetchIndexers, deleteIndexer, cloneIndexer } from 'Store/Actions/settingsActions';
 import Indexers from './Indexers';
 
 function createMapStateToProps() {
   return createSelector(
-    createSortedSectionSelector('settings.indexers', sortByName),
```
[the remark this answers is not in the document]
Taloth: Not sure, I feel like there's an argument for both. I do somewhat prefer sort by name but that's quite probably bias since I don't have 20 indexers in the first place. Not exactly the hill I want to die on.
I do agree that it should be consistent with download clients.
So... verdict, yeah sort by priority then name would be ok. Assuming it resorts properly on changes.

Also 'sortByPriorityThenName' would be even better since there's nothing 'indexey' in the function. (can we get it down to 'sortByMagic'?)

PS: This function needs to change if we change to the +1 -1 scheme.

src/NzbDrone.Core/DecisionEngine/DownloadDecisionComparer.cs
```diff
@@ -32,6 +32,7 @@ public int Compare(DownloadDecision x, DownloadDecision y)
                 CompareLanguage,
                 ComparePreferredWordScore,
                 CompareProtocol,
+                CompareIndexerPriority,
```
Taloth, line 35: I think you need to drop this below CompareEpisodeNumber... CompareEpisodeCount is what ensures that season packs are grabbed first (if all else is equal). Otherwise the user gets into the situation that if not all episodes are available as single downloads, one single episode will prevent the rest from ever being grabbed (since the season pack wouldn't be an upgrade for all episodes).
It'll still prefer a pack from the indexer with a higher priority over a pack from a lower prio. But it won't prefer a single episode from a high prio indexer over a pack from a lower prio.

src/NzbDrone.Core/Download/ProcessDownloadDecisions.cs
```diff
@@ -73,6 +73,7 @@ public ProcessedDecisions ProcessDecisions(List<DownloadDecision> decisions)
 
                 try
                 {
+                    _logger.Info("Downloading release {0} at priority {1}.", remoteEpisode.Release.Indexer, remoteEpisode.Release.IndexerPriority);
```
Taloth, line 76: Remove. DownloadReport already logs at Info.
Also {0} = Indexer name. so "Downloading release MyIndexer at priority x." whoopsie.
If you still want to log prio, log it at debug/trace level.

src/NzbDrone.Core/Indexers/IndexerDefinition.cs
```diff
@@ -10,6 +10,7 @@ public class IndexerDefinition : ProviderDefinition
         public DownloadProtocol Protocol { get; set; }
         public bool SupportsRss { get; set; }
         public bool SupportsSearch { get; set; }
+        public int Priority { get; set; } = 100;
```
Taloth, line 13: You have two usecases that are competing for a particular default: 
- Users with fallback indexer with lower grab-limits or generally more junk releases.
- Users with premium indexers that they prefer.

The DownloadClients priority was more of the Fallback variety and thus a default of 1 (and 1 hidden on the UI), 2 being the fallback.

I don't like the idea of an arbitrarily large 100 number. I also don't want to see it on the UI unless the user is actually using priorities, and definitely not a "Priority: 100" on every indexer by default.

I see two approaches:
- We ignore the premium indexer usecase, anyone wanting it can simply set all others to a higher number. Default would be 1, only values > 1 shown on the ui.
- We switch to 0 as default and allow +1 (higher prio) and -1 (lower prio). Then we can hide 0 on the ui and won't have an arbitrary default. This also means that DownloadClient prio should be updated to that scheme as well for consistency (and a sql in the migration `UPDATE DownloadClients SET Priority = (1 - Priority)`)

Dtaggart: Well I actually became interested in working on this because of my case that differs.
I personally much rather download from my junk indexers first before my premium ones. I am trying to build my ratios for my more premium indexers and would rather use them as a last resort. This is ok by me because sonarr has so many filters for junk indexers that it is rarely a issue that it gets it wrong.

If we are comparing it to Download Client Priority, that is a 1-50 range option. 1 being default. So it defaults a bit differently but that is because normally you don't have more then one, that might be a assumption on my part but I don't think there and many users with more then one. Indexers is where I see people having a lot. I personally have 15 set up and plan on adding more as my private tracker collection grows.

Now the reason I think priority should be shown for each indexer is because we are now sorting the the cards by this value so I think it should always be shown. I do see it wouldn't be needed for people not using it so I do agree that we could hide it if all are set to 100. 

I personally don't like the -1 +1 approach. When talking about the priority the easiest thing to understand is using it to describe positive numbers. I think you would get more questions if you try introducing negatives here.  

I understand that setting the default priority to the lowest available differs from the Download client priority but I do think this is the best way to handle this. Another argument for this way is if someone set priorities, went to create a new one they would need to remember they need to set the priority for the new indexer. Defaulting to the highest I think would just cause more issues. You should have to purposely set something for it to be considered a higher priority not default into it. 

I would be much more inclined to set the max priority to 50 to mach the download client and keep the current implementation.

src/NzbDrone.Core/Download/ProcessDownloadDecisions.cs
Dtaggart: Ok yeah that makes sense

src/NzbDrone.Core/Indexers/IndexerDefinition.cs
markus101: > So it defaults a bit differently but that is because normally you don't have more then one, that might be a assumption on my part but I don't think there and many users with more then one.

I think that's a fair assumption, 1 for torrents and 1 for newsgroups.

> Now the reason I think priority should be shown for each indexer is because we are now sorting the the cards by this value so I think it should always be shown.

I still don't think sorting by priority in this list makes sense, it'd make sense in a dragable list, but less so in the card layout used.

> I do see it wouldn't be needed for people not using it so I do agree that we could hide it if all are set to 100.

I'm in favour of this as well, if you're not using priority at all it's just noise.

> Defaulting to the highest I think would just cause more issues. You should have to purposely set something for it to be considered a higher priority not default into it.

Same thing for defaulting to the lowest, if you have 10 indexers and want one to have a lower priority then you're forced to edit every indexer to fix that. A default of either the highest or the lowest doesn't make sense to me, set it right in the middle and following download clients it should be 25.


TLDR:

- Range from 1-50
- Default priority to 25 (users can go higher and lower without editing every other indexer)
- Hide priority if they're all the default
- Don't sort by priority, just name

Dtaggart: > Same thing for defaulting to the lowest, if you have 10 indexers and want one to have a lower priority then you're forced to edit every indexer to fix that. A default of either the highest or the lowest doesn't make sense to me, set it right in the middle and following download clients it should be 25.

Fair point, I could get behind that.

I'll work on fixing this up based on you're comments

Dtaggart:
@markus101 @Taloth Alrighty, looks like I hit it all. Srry it took a week+. I am not a front end js guy was putting off but think I figured out the right way to hide the priority. If there is a better way let me know. Thanks a lot for the input.

frontend/src/Settings/Indexers/Indexers/Indexers.js
```diff
@@ -76,6 +76,7 @@ class Indexers extends Component {
                   <Indexer
                     key={item.id}
                     {...item}
+                    showPriority={items.some((index) => index.priority != 25)}
```
markus101, line 79: Compute and store this outside of the map so it isn't recalculated for each item.

```
const showPriority = items.some((index) => index.priority != 25)
...
showPriority={showPriority}
```

src/NzbDrone.Core.Test/DecisionEngineTests/PrioritizeDownloadDecisionFixture.cs
```diff
@@ -38,7 +38,7 @@ private Episode GivenEpisode(int id)
                             .Build();
         }
 
-        private RemoteEpisode GivenRemoteEpisode(List<Episode> episodes, QualityModel quality, Language language, int age = 0, long size = 0, DownloadProtocol downloadProtocol = DownloadProtocol.Usenet)
+        private RemoteEpisode GivenRemoteEpisode(List<Episode> episodes, QualityModel quality, Language language, int age = 0, long size = 0, DownloadProtocol downloadProtocol = DownloadProtocol.Usenet, int indexerPriority = 100)
```
markus101, line 41: Set the default to `25`

src/NzbDrone.Core/Download/ProcessDownloadDecisions.cs
```diff
@@ -73,6 +73,7 @@ public ProcessedDecisions ProcessDecisions(List<DownloadDecision> decisions)
 
                 try
                 {
+                    _logger.Trace("Downloading from Indexer {0} at priority {1}.", remoteEpisode.Release.Indexer, remoteEpisode.Release.IndexerPriority);
```
markus101, line 76: `Grabbing` instead of `Downloading`? Ties in with the naming + I think helps differentiate between the download client downloading and Sonarr passing the release to the client.

src/NzbDrone.Core/Datastore/Migration/142_add_priority_to_indexers.cs
```diff
@@ -0,0 +1,15 @@
+﻿using FluentMigrator;
+using NzbDrone.Core.Datastore.Migration.Framework;
+using NzbDrone.Core.Indexers;
+
+namespace NzbDrone.Core.Datastore.Migration
+{
+    [Migration(142)]
```
markus101, line 7: This conflicts with the migration in import lists: https://github.com/Sonarr/Sonarr/pull/3398

Please change to 143.

markus101 (changes requested):
Thanks for the fixups, just a few more little things.

src/NzbDrone.Core/Datastore/Migration/142_add_priority_to_indexers.cs
```diff
@@ -4,7 +4,7 @@
 
 namespace NzbDrone.Core.Datastore.Migration
 {
-    [Migration(142)]
+    [Migration(143)]
```
markus101, line 7: Please rename the file as well.

markus101 (changes requested):
One last thing.

src/NzbDrone.Core/Datastore/Migration/142_add_priority_to_indexers.cs
Dtaggart: oh yeah my fault for missing will do

markus101 (approved):
👍

PR #3923, opened 2020-08-22
Clustering

Matty666:
#### Database Migration
NO

#### Description
These commits are to introduce cluster analysis for the download decisions. It currently will cluster decisions by 200mb size and order by the average size of the cluster.

#### Todos
- [ ] Documentation

Couldn't find the feature in the issues, if there is one for this please point it out?


Matty666:
> Btw. it doesn't need a config option. It can be useful atm for testing, but not when we release it.

No worries, I removed the config setting.

src/NzbDrone.Core/DecisionEngine/ClusterAnalysis/Ordered/OrderedEnumerable.cs
```diff
@@ -0,0 +1,38 @@
...
+        }
+
+        public override IOrderedEnumerable<ClusteredElement<TElement>> ApplyOrdering(IEnumerable<ClusteredElement<TElement>> source)
+        {
+            if (Parent == null)
+            {
+                return _descending ? source.OrderByDescending(_keySelector, _comparer) : source.OrderBy(_keySelector, _comparer);
+            }
+
+            var parentOrdering = Parent.ApplyOrdering(source);
```
Taloth, line 29: This is not correct, forget about clustering for a sec: To recursively sort you'd first sort by yourself, then the parent. This automatically ensures that the groups behave properly because sorts are stable.

Proof: sort `[ 101, 204, 203, 105, 103 ]`
.OrderBy(v => v/100)
.ThenBy(v => v % 100)
First sort by our criteria ( % 100): `[ 101, 203, 103, 204, 105 ]`
Then sort by our parent criteria (/100): `[ 101, 103, 105, 203, 204 ]`
voila properly sorted.
If instead i use your logic then we'd end up with `[ 101, 103, 203, 204, 105 ]`

In clustering this is different, because clustering/grouping is not stable. You have to deal with individual groups/clusters, rather than the entire list.

Imagine the following:
sort `[ 101, 204, 203, 105, 103 ]`
.OrderBy(v => v/100)
.ThenByCluster(v => v % 100, 2)

OrderBy effectively produces several groups: `[101, 105, 103], [204, 203]`
ThenByCluster (cluster by 2 distance inclusive) should operate on those two groups individually, so: `[101, 105, 103]` becomes `[101, 103], [105]` or `[101], [105, 103]` depending on your implementation. And [203, 204] remains `[203, 204]`.
the final result becomes either [101, 103, 105, 203, 204] or [101, 105, 103, 203, 204].

I haven't tested it, but I think in your current implementation you end up with the wrong result.

src/NzbDrone.Core/DecisionEngine/ClusterAnalysis/Ordered/ClusterableOrderableEnumerable.cs
```diff
@@ -0,0 +1,39 @@
...
+            return GetEnumerator();
+        }
+
+        public IOrderedClusteredEnumerable<TElement> CreateOrderedEnumerable<TKey>(
+            Func<TElement, TKey> keySelector, IComparer<TKey> comparer, bool descending)
+        {
+            return new OrderedEnumerable<TElement, TKey>(_source, null, keySelector, comparer, descending);
+        }
+
+        public IOrderedClusteredEnumerable<TElement> CreateClusterOrderedEnumerable(
```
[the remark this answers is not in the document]
Matty666: Yes I see what you're getting at here, I'll take a look at refactoring this to simplify it

src/NzbDrone.Core/DecisionEngine/ClusterAnalysis/Ordered/OrderedEnumerable.cs
```diff
@@ -0,0 +1,38 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace NzbDrone.Core.DecisionEngine.ClusterAnalysis.Ordered
+{
+    public sealed class OrderedEnumerable<TElement, TKey> : OrderedClusteredEnumerableBase<TElement>
```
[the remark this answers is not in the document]
Matty666: Yes, that was my dilemma when naming it. Yes I agree, hopefully simplifying it and removing the class hierarchy will remove the need for this class

src/NzbDrone.Core/DecisionEngine/DownloadDecisionPrioritizationService.cs
```diff
@@ -24,13 +25,18 @@ public DownloadDecisionPriorizationService(IConfigService configService, IDelayP
...
                             .Union(decisions.Where(c => c.RemoteEpisode.Series == null))
                             .ToList();
         }
+
+        private IEnumerable<DownloadDecision> PrioritizeDecisionsWithClusterAnalysis(int seriesId, IEnumerable<DownloadDecision> downloadDecisions)
+        {
+            return downloadDecisions
+                .Cluster()
+                .OrderByDescending(decision => decision, new DownloadDecisionComparer(_configService, _delayProfileService))
+                .ThenByClusterDescending(d => d.RemoteEpisode.Release.Size, 200.Megabytes());
```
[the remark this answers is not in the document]
Matty666: ok i can add that in if the PreferredSize is available (even if it's zero)

src/NzbDrone.Core/DecisionEngine/DownloadDecisionComparer.cs
```diff
@@ -36,7 +36,6 @@ public int Compare(DownloadDecision x, DownloadDecision y)
                 CompareEpisodeNumber,
                 ComparePeersIfTorrent,
```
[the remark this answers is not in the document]
Matty666: Yes the extension methods at the moment hide the distance function, but we can add extension methods that expose the distance function rather than just a distance value selector. When we were talking about this originally we discussed `log`, specifically around peer counts. There is a test here that is a simple example:

https://github.com/Matty666/Sonarr/blob/5ce59602a34aac64d4aaec2ede63b0c80a13815c/src/NzbDrone.Core.Test/DecisionEngineTests/ClusterAnalysis/HierarchicalClusteringFixture.cs#L103

src/NzbDrone.Core/DecisionEngine/ClusterAnalysis/Ordered/OrderedEnumerable.cs
Matty666: > This is not correct, forget about clustering for a sec: To recursively sort you'd first sort by yourself, then the parent. This automatically ensures that the groups behave properly because sorts are stable.
>
>Proof: sort [ 101, 204, 203, 105, 103 ]
>.OrderBy(v => v/100)
>.ThenBy(v => v % 100)
>First sort by our criteria ( % 100): [ 101, 203, 103, 204, 105 ]
>Then sort by our parent criteria (/100): [ 101, 103, 105, 203, 204 ]
>voila properly sorted.
>If instead i use your logic then we'd end up with [ 101, 103, 203, 204, 105 ]

The implementation doesn't actually do a sort, it uses the framework sorting, all it does is apply the `OrderBy` orderings in the order that they were called when GetEnumerator is called.

I added a test to check this: https://github.com/Matty666/Sonarr/blob/59503e824f05a566c95cf379a8c24199268ae467/src/NzbDrone.Core.Test/DecisionEngineTests/ClusterAnalysis/ClusterOrderedEnumerableFixture.cs#L84

Matty666: >In clustering this is different, because clustering/grouping is not stable. You have to deal with individual groups/clusters, rather than the entire list.
>
>Imagine the following:
>sort [ 101, 204, 203, 105, 103 ]
>.OrderBy(v => v/100)
>.ThenByCluster(v => v % 100, 2)
>
>OrderBy effectively produces several groups: [101, 105, 103], [204, 203]
>ThenByCluster (cluster by 2 distance inclusive) should operate on those two groups individually, so: [101, 105, 103] >becomes [101, 103], [105] or [101], [105, 103] depending on your implementation. And [203, 204] remains [203, 204].
>the final result becomes either [101, 103, 105, 203, 204] or [101, 105, 103, 203, 204].
>
>I haven't tested it, but I think in your current implementation you end up with the wrong result.

Yes let me take a look into this and add some more tests to see. The implementation applies each clustering first to extend each list item with a value which is the average value of the cluster that the item is in, then once that is done it orders by the orderings, either the key selector for non-clustered orders, or the average cluster value of each item for the cluster that was specified

Matty666: Here is an example of how the implementation works for:

`{v: 101}, {v: 204}, {v: 203}, {v: 105}, {v: 103}`

```
OrderBy(v => v / 100)
.ThenByCluster(v => v % 100, 2)
```

### Step 1 - apply clustering averages - `ClusterBy(v => v % 100, 2)`
results in 2 clusters: `[{v: 204}, {v: 105}]` and `[{v: 101}, {v: 203}, {v: 103}]`

applying the cluster average to each instance you end up with:
```
{v: 101, cavg: 135.6}
{v: 204, cavg: 154.5}
{v: 203, cavg: 135.6}
{v: 105, cavg: 154.5}
{v: 103, cavg: 135.6}
```

_now sort by `v / 100` then by cluster average_

### Step 2 - `OrderedEnumerable` calculates the sorting keys for `v / 100`
and also the keys for cavg which are just the same values
```
{v: 101, skey: 1, cavg: 135.6}
{v: 204, skey: 2, cavg: 154.5}
{v: 203, skey: 2, cavg: 135.6}
{v: 105, skey: 1, cavg: 154.5}
{v: 103, skey: 1, cavg: 135.6}
```

### Step 3 - `OrderedEnumerable` performs quick sort
when sorting, the key comparison compares the keys of two instances and if they are the same then compares the next keys and so on. If they are all the same it uses index order

so in order of skey then cavg:
```
{v: 101, skey: 1, cavg: 135.6}
{v: 103, skey: 1, cavg: 135.6}
{v: 105, skey: 1, cavg: 154.5}
{v: 203, skey: 2, cavg: 135.6}
{v: 204, skey: 2, cavg: 154.5}
```

Is this what you were expecting, sorting by the cluster averages by applying all the clusterings first to calculate each items sorting key?

Taloth: Hey Matty, thanks for sticking with it.

> Is this what you were expecting, sorting by the cluster averages by applying all the clusterings first to calculate each items sorting key?

I don't think so, it's a simple example but it starts failing with more complicated scenarios. Particularly when multiple subclusters occur or suborderings.

> Step 1 - apply clustering averages - ClusterBy(v => v % 100, 2)
results in 2 clusters: [{v: 204}, {v: 105}] and [{v: 101}, {v: 203}, {v: 103}]

No, there's no cluster of 204 and 105, because they were already grouped differently by OrderBy. This causes issues further down the line.

Also, cavg appears wrong because the selected value is modulo % 100, so the 204, 105 cluster would be cavg: 4.5.   
It's probably clearer to split up the modulo into two separate variables: `{a: 1, b: 1}, {a: 2, b: 4}, {a: 2, b: 3}, {a: 1, b: 5}, {a: 1, b: 3}`, then `list.OrderBy(o => o.a).ThenClusterBy(o => o.b, 2)`. But the modulo was a shorter/easier notation.

Your clusters are created differently because of the fact that your clustering all items, rather than intermediate groups.
Take the values 9,6,5,4,2,1. `OrderBy(v => v >= 5).ThenClusterBy(v => v, 2)`
Do you agree that the clusters determined by ThenClusterBy would be [9] + [6,5,4] + [2,1]?
(Clusters of [9] + [6,5] + [4,2,1] is technically valid but not logical since 4 is closer to the middle group.)

This is problematic, because the [2,1] group is lower than the group 4 belongs to, and thus gets sorted as such.
First consider the OrderBy: Performing it manually it results in [4,2,1] + [9,6,5] as groups (in that order).
Then perform ThenClusterBy manually, which would keep [4,2,1] as group, and split up the other one into [6,5] + [9].
Logically this means the final order _should_ be 4,2,1,6,5,9... correct?

Consider your scoring:
```
{v: 9, skey: true, cavg: 9}
{v: 6, skey: true, cavg: 5}
{v: 5, skey: true, cavg: 5}
{v: 4, skey: false, cavg: 5}
{v: 2, skey: false, cavg: 1.5}
{v: 1, skey: false, cavg: 1.5}
```
First you sort by cavg: (ThenClusterBy)
```
{v: 2, skey: false, cavg: 1.5}
{v: 1, skey: false, cavg: 1.5}
{v: 6, skey: true, cavg: 5}
{v: 5, skey: true, cavg: 5}
{v: 4, skey: false, cavg: 5}
{v: 9, skey: true, cavg: 9}
```
then you sort by skey: (OrderBy)
```
{v: 2, skey: false, cavg: 1.5}
{v: 1, skey: false, cavg: 1.5}
{v: 4, skey: false, cavg: 5}
{v: 6, skey: true, cavg: 5}
{v: 5, skey: true, cavg: 5}
{v: 9, skey: true, cavg: 9}
```
so the final result becomes 2,1,4,6,5,9 which is different than our manually determined order, simply because '4' ended up in the wrong cluster.

I confirmed this with a testcase:
```
        [Test]
        public void should_order_recursively_2()
        {
            var source = new List<int> { 9, 6, 5, 4, 2, 1 };

            var result = source
                .Cluster()
                .OrderBy(v => v >= 5)
                .ThenByCluster(v => v, 2)
                .ToList();

            result.Should().Equal(new List<int> { 4, 2, 1, 6, 5, 9 });
            // Actual: {2, 1, 4, 6, 5, 9} 
        }
```

The 'sort in reverse order' approach ONLY works with regular stable sort, not with clustering.    
In my opinion, the only solution is to make OrderBy and ClusterBy to create a sorted list of groups to be fed into the next Then*. Clustering must be applied to those individual groups. That way the clustering algorithm can ignore values that are already part of another group, despite being close to the cluster.   
This also means you need less housekeeping because you no longer need to track the scores of the individual stages.

I hope this explains the problem. Let me know if further explanation is necessary or when you need help to update the algorithm.

Matty666: > Also, cavg appears wrong because the selected value is modulo % 100, so the 204, 105 cluster would be cavg: 4.5.
Yes you're right, my mistake, 4.5 and 2.3. Doh!

Matty666: > Do you agree that the clusters determined by ThenClusterBy would be [9] + [6,5,4] + [2,1]?

Yes, thats the clustering if you do `source.ClusterBy(v => v, 2).ToList()` 
`[2,1] + [9] + [5,6,4]`

Matty666: I see what you're getting at, I think! :)

Something like this?

`OrderBy(v => v >= 5)` will sort the list by the key selector, then group that sorted list by the key selector, enumerate each group and call the next ThenBy 

`ThenByCluster(v => v, 2)` will accept the grouping from the previous, cluster that individual group, enumerate each cluster and call the next ThenBy, and so on

Matty666: `9, 6, 5, 4, 2, 1`

I tried this manually, clustering each group:
```
var source = new List<int> { 9, 6, 5, 4, 2, 1 };

var ordering = source.OrderBy(v => v >= 5).ToList();

var grouping = ordering.GroupBy(v => v >= 5, v => v).ToList();

var clusterGroup1 = grouping[0].ClusterBy(v => v, 2).ToList();
var clusterGroup2 = grouping[1].ClusterBy(v => v, 2).ToList();
```

order by `v >= 5` results in all the `false` results first in index order, then the `true` results

```
{v: 4, sk: false}
{v: 2, sk: false}
{v: 1, sk: false}
{v: 9, sk: true}
{v: 6, sk: true}
{v: 5, sk: true}
```

then by cluster `v => v, 2`

first group: `false`
```
{v: 4, sk: false, cavg: 4}

{v: 2, sk: false, cavg: 1.5}
{v: 1, sk: false, cavg: 1.5}
```

second group: `true`
```
{v: 9, sk: true, cavg: 9}

{v: 6, sk: true, cavg: 5.5}
{v: 5, sk: true, cavg: 5.5}
```

I'm not sure now how to sort these clusters, because the clustering only clusters the results.

Taloth: For OrderBy: Group first, sort by group key. So `.GroupBy(keySelector).OrderBy(v => v.Key).ToList()`

For clustering you want the same, make a group of each cluster and sort those groups by the cavg. Note that you only sort the groups themselves, not the items within (that way you retain somewhat of a stable sort).

Note that once all OrderBy and ClusterBy is done and someone calls the ClusteredEnumerable.GetEnumerator function, then you can do a `sourceGroups.SelectMany(v => v).GetEnumerator()`, flattening the already sorted groups.

Matty666: Hey @Taloth , i've re-worked the enumerable now.

It now groups the elements, orders each group, then applies any subsequent group/order to each group. I modified the clustering to implement IGrouping with the key being the cluster average, which makes it simpler to then do the enumerable.

I added `should_order_recursively_and_by_cluster` to test the `9, 6, 5, 4, 2, 1` grouping, and it results in:

`2, 1, 4, 6, 5, 9`

The `2, 1, 4` ordering is due to 4 ending up in it's own cluster. This is because the algorithm takes the max distance between each element in the cluster it's merging, it merges 2, 1 first, then 4 into that cluster, and so the max between 4 and that cluster is 3, which is over the cut point.

PR #3979, opened 2020-09-24
Dataloss when moving series folder to root folder with only different casing

Taloth:
With series folder C:\media\movies\Series.Title
Moving to root folder C:\media\Movies\Series.Title

These folders are the same, but TransferFolder doesn't detect it as such, and switches to a recursive move+delete.

This PR adjusts the logic to detect it and throw an exception.
It'll also handle series.title -> Series.Title renames and should handle case-sensitive filesystems as well.

I think we still should add a protection mechanism to the DeleteFolder function, to prevent deletion if any large file remains to avoid Dataloss if there's another edgecase we failed to handle.

#### Todos
- [x] Detect casing
- [x] Detect large files before DeleteFolder and cancel delete

### DIFF
--- a/src/NzbDrone.Common.Test/DiskTests/DiskTransferServiceFixture.cs
+++ b/src/NzbDrone.Common.Test/DiskTests/DiskTransferServiceFixture.cs
@@ -452,6 +452,38 @@ namespace NzbDrone.Common.Test.DiskTests
         }
 
         [Test]
+        public void MoveFolder_should_detect_caseinsensitive_parents()
+        {
+            WithRealDiskProvider();
+
+            var original = GetFilledTempFolder();
+            var root = new DirectoryInfo(GetTempFilePath());
+            var source = new DirectoryInfo(root.FullName + "A/series");
+            var destination = new DirectoryInfo(root.FullName + "a/series");
+
+            Subject.TransferFolder(original.FullName, source.FullName, TransferMode.Copy);
+
+            Assert.Throws<IOException>(() => Subject.TransferFolder(source.FullName, destination.FullName, TransferMode.Move));
+        }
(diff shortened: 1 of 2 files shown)


Taloth:
@Qstick second commit handles the TransferFile specific stuff. But the change is quite a few lines of code for handling the obscure symlink edgecase. So I'm thinking I should revert the commit and just accept the risk of that edgecase. Lemme know what you think.

Qstick:
Yea, I know its a possibility but don't think we have seen a case of it. Someone would have to try hard to create that case I think. I'd be in agreement with that. You could leave the comment if need be as a reminder that its a possible edge case.

cocokola:
If I am following along, this is what happened to mem lost 10 years of .torrent files (not a major loss but nonetheless) and 500gig of videos downloaded that needed to be redownloaded.  It is only likely to happen during misconfiguration, the app stopped mid-task, etc where the use-case becomes valid.

Taloth:
I actually merged in the first commit a couple of days ago, so it's closed. That one handles the casing problem when moving/renaming folders.

@Qstick You probably know, but if you pull that commit be sure to also pull the fix that deals with that broken updater (and everything else).

Qstick:
> I actually merged in the first commit a couple of days ago, so it's closed. That one handles the casing problem when moving/renaming folders.
> 
> @Qstick You probably know, but if you pull that commit be sure to also pull the fix that deals with that broken updater (and everything else).

Yep, thanks @Taloth, was giving it a few days to make sure nothing else came up before we pulled both over 😁 We'll likely do it tomorrow

Taloth:
Probably better to just give it 1-2 weeks tbh

PR #3980, opened 2020-09-25
Added Filelist.io indexer support

Taloth:
#### Database Migration
NO

#### Description
Primarily adds FileList.io indexer api support for recent and searches. No specials and daily at this time.
Note that Filelist.io has no paging.

Additionally, this PR adds a mode to the select input to turn it into a Multi-Select input, where the dropdown items have checkboxes allowing multiple selections.
I'm not fond of the implementation since it uses a MultiSelectInputConnector as wrapper, but EnhancedSelectInput still has some logic to make it aware of multi-select. Perhaps, instead of using a seperate class we need to include the checkbox and multi-select support directly into EnhancedSelectInput. So any advice there is welcome.

Unrelated I also added PrivacyLevel to all fields, the idea behind this is to be able to do something with it in the UI. I'm considering the option to blur usernames/apikeys to facilitate safer screenshots. There's no functionality behind this yet, just annotations.

#### Todos
- [x] Implement rss and search
- [ ] Test actual grab
- [ ] Ask Filelist staff about daily series


frontend/src/Components/Form/MultiSelectInputSelectedValue.js
```diff
@@ -0,0 +1,44 @@
...
+
+  return (
+    <EnhancedSelectInputSelectedValue
+      className={styles.selectedValue}
+      {...otherProps}
+    >
+      <div className={styles.valueText}>
+        {
+          value.map((key, index) => {
+            const v = values.find((i) => i.key === key);
```
markus101, line 22: Since this will run every time it's rendered if values is going to have a lot of values having an object with the key being the key and the value being the entire object would be better for performance, not sure it'll get to that point, but something to consider.

frontend/src/Components/Form/MultiSelectInputOption.js
```diff
@@ -0,0 +1,76 @@
...
+import EnhancedSelectInputOption from './EnhancedSelectInputOption';
+import CheckInput from './CheckInput';
+import styles from './MultiSelectInputOption.css';
+
+class MultiSelectInputOption extends Component {
+
+  //
+  // Listeners
+
+  onPress = () => {
```
markus101, line 13: Nit: not needed.

frontend/src/Components/Form/MultiSelectInputConnector.js
```diff
@@ -0,0 +1,45 @@
...
+  return {};
+}
+
+const mapDispatchToProps = {
+};
+
+class MultiSelectInputConnector extends Component {
+
+  //
+  // Lifecycle
```
markus101, line 18: Nit: unneeded

markus101 (approved):
Looking really good, took it for a quick spin and it works great. A couple nits and a possible perf improvement in same cases, but otherwise 👍

frontend/src/Components/Form/MultiSelectInputOption.js
Taloth: CheckInput requires an event handler, and normally it's absolutely required to be an effective control. But we kinda abuse it to show a check symbol, so I didn't want to remove the isRequired from props there.

Issue #3984, opened 2020-09-29
DTS-HD in file name results in false positive group of "-HD"


bakerboy448:
confirmed bug; opening issue for tracking purposes.

"DTS-HD" needs to be added to the exclude list

Adding details from #Development

Release Example: Shameless US (2010) S04 (1080p BDRip x265 10bit DTS-HD MA 5 1 - WEM)[TAoE]

bakerboy448:
should also account for DTS-X, DTS-MA, and DTS-ES

PR #3986, opened 2020-09-30
Fix Parser False Positive Group Matches

bakerboy448:
#### Database Migration
NO

#### Description
Fix: Add various DTS- audio formats to exclusion so parser does not detect it as a group
Fix: Add Blu-Ray to exclusion so parser does not detect it as a group

#### Todos
- [x] Tests
- [ ] Documentation

* #3984 

Regex is the bane of my existence, but I think I did this right...testing indicates it did not pick up the -HD of DTS-HD as expected.

It may make sense to split the audio into it's own group?

### DIFF
--- a/src/NzbDrone.Core.Test/ParserTests/ReleaseGroupParserFixture.cs
+++ b/src/NzbDrone.Core.Test/ParserTests/ReleaseGroupParserFixture.cs
@@ -36,8 +36,7 @@ namespace NzbDrone.Core.Test.ParserTests
         [TestCase("SomeShow.S20E13.1080p.BluRay.DTS-X.MA.5.1.x264-RlsGroup","RlsGroup")]
         [TestCase("SomeShow.S20E13.1080p.BluRay.DTS-MA.5.1.x264-RlsGroup","RlsGroup")]
         [TestCase("SomeShow.S20E13.1080p.BluRay.DTS-ES.5.1.x264-RlsGroup","RlsGroup")]
-        
-       //[TestCase("", "")]
+        //[TestCase("", "")]
         public void should_parse_release_group(string title, string expected)
         {
             Parser.Parser.ParseReleaseGroup(title).Should().Be(expected);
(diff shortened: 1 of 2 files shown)


markus101 (changes requested):
Needs to have test cases added to `ReleaseGroupParserFixture` to verify the behaviour.

bakerboy448:
> Needs to have test cases added to `ReleaseGroupParserFixture` to verify the behaviour.

added 1 for the moment....need to dig up a few examples so we're covering each DTS...or atleast one that has an expected group name 👍 
 (unless you think 1 is enough).

Hopefully will update sometime tomorrow.

bakerboy448:
@markus101 should be all set here.

bakerboy448:
Added Blu-ray as well. Based on discussions with RG it is an issue that is experienced.

src/NzbDrone.Core.Test/ParserTests/ReleaseGroupParserFixture.cs
```diff
@@ -31,6 +31,12 @@ public class ReleaseGroupParserFixture : CoreTest
         [TestCase("Black Monday S01 E01-E02 1080p AMZN WEBRip DDP5.1 x264 monkee", null)]
         [TestCase("Series.Title.S01E05.The-Aniversary.WEBDL-1080p.mkv", null)]
         [TestCase("Series.Title.S01E05.The-Aniversary.HDTV-1080p.mkv", null)]
+        [TestCase("Shameless US (2010) S04 (1080p BDRip x265 10bit DTS-HD MA 5 1 - WEM)[TAoE]",null)]
```
Taloth, line 34: Only this one fails, the other ones already pass because it'd simply match TrollUHD & RlsGroup already in the original regex. Drop the trailing -RlsGroup in the tests.

bakerboy448: done.

PR #3992, opened 2020-10-02
Improve use of All() and speedup Unmapped Folder matching

Qstick:
#### Database Migration
NO

#### Description

- Add `AllSeriesPaths` Method
- Use `AllSeriesPaths` in lieu of `All()` in HealthChecks, Diskspace, RootFolderService
- Only call All() (AllSeriesPaths) once for `AllWithUnmappedFolders` instead of for each folder

#### Todos
- [x] Tests

### DIFF
--- a/src/NzbDrone.Core/Tv/SeriesService.cs
+++ b/src/NzbDrone.Core/Tv/SeriesService.cs
@@ -24,6 +24,7 @@ namespace NzbDrone.Core.Tv
         Series FindByPath(string path);
         void DeleteSeries(int seriesId, bool deleteFiles, bool addImportListExclusion);
         List<Series> GetAllSeries();
+        List<string> GetAllSeriesPaths();
         List<Series> AllForTag(int tagId);
         Series UpdateSeries(Series series, bool updateEpisodesToMatchSeason = true, bool publishUpdatedEvent = true);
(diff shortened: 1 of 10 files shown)


Qstick:
A few stats on a couple cases I ran for this:

Commit 1
```
With 85 Series (small case) - GetAll vs GetAllPaths
[Info] RootFolderCheck: GetAll time: 23ms
[Info] RootFolderCheck: GetAllPaths time: 7ms
```

Commit 2
```
27 Root Folders, ~450 Series

Before
[Debug] Api: [GET] /api/v3/rootFolder: 200.OK (8040 ms)

After
[Debug] Api: [GET] /api/v3/rootFolder: 200.OK (195 ms)
```

src/NzbDrone.Core/Tv/SeriesService.cs
```diff
@@ -157,6 +158,11 @@ public List<Series> GetAllSeries()
             return _seriesRepository.All().ToList();
         }
 
+        public List<string> GetAllSeriesPaths()
+        {
+            return _seriesRepository.AllSeriesPaths().ToList();
```
markus101, line 163: `_seriesRepository.AllSeriesPaths()` already returns a list.

markus101 (changes requested):
Just some small things to clean up, then I think we're good.

Qstick:
@markus101 Fixed

PR #3994, opened 2020-10-03
Newznab/Torznab categories dropdown with indexer provided category names

Taloth:
#### Description
Changed Newznab/Category from comma separated text box to multi select.
The categories are pulled from the indexer caps dynamically.

- If selected categories don't exist then the number is shown instead.
- If the selected category doesn't exist in the indexr, the number is shown instead. Those are removed the next change the user makes.
- If the indexer isn't reachable or requires an apikey but is invalid, then the default newznab categories are shown instead.

image

### DIFF
--- a/frontend/src/Components/Form/DeviceInputConnector.js
+++ b/frontend/src/Components/Form/DeviceInputConnector.js
@@ -2,13 +2,13 @@ import PropTypes from 'prop-types';
 import React, { Component } from 'react';
 import { connect } from 'react-redux';
 import { createSelector } from 'reselect';
-import { fetchOptions, clearOptions } from 'Store/Actions/providerOptionActions';
+import { fetchOptions, clearOptions, defaultState } from 'Store/Actions/providerOptionActions';
 import DeviceInput from './DeviceInput';
 
 function createMapStateToProps() {
   return createSelector(
     (state, { value }) => value,
-    (state) => state.providerOptions,
+    (state) => state.providerOptions.devices || defaultState,
     (value, devices) => {
 
       return {
@@ -51,7 +51,7 @@ class DeviceInputConnector extends Component {
   }
 
   componentWillUnmount = () => {
-    this.props.dispatchClearOptions();
+    this.props.dispatchClearOptions({ section: 'devices' });
   }
 
   //
@@ -65,6 +65,7 @@ class DeviceInputConnector extends Component {
     } = this.props;
 
     dispatchFetchOptions({
+      section: 'devices',
       action: 'getDevices',
       provider,
       providerData
--- a/src/NzbDrone.Core/Indexers/Newznab/NewznabCategoryFieldOptionsConverter.cs
+++ b/src/NzbDrone.Core/Indexers/Newznab/NewznabCategoryFieldOptionsConverter.cs
@@ -1,0 +1,75 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Web.UI.WebControls;
+using NzbDrone.Core.Annotations;
+
+namespace NzbDrone.Core.Indexers.Newznab
+{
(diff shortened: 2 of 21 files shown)


Qstick:
Nice... with nesting. Looking good.

frontend/src/Components/Form/DeviceInputConnector.js
```diff
@@ -2,13 +2,13 @@ import PropTypes from 'prop-types';
...
 import { createSelector } from 'reselect';
-import { fetchOptions, clearOptions } from 'Store/Actions/providerOptionActions';
+import { fetchOptions, clearOptions, defaultState } from 'Store/Actions/providerOptionActions';
 import DeviceInput from './DeviceInput';
 
 function createMapStateToProps() {
   return createSelector(
     (state, { value }) => value,
-    (state) => state.providerOptions,
+    (state) => state.providerOptions.getDevices || defaultState,
```
markus101, line 11: Not really following how `getDevices` is used here. I'd also expect it to be `devices` as a property instead of `getDevices` which looks like a function.

markus101 (approved):
Just one question about `getDevices`, otherwise this is awesome, nice work!

frontend/src/Components/Form/DeviceInputConnector.js
Taloth: Currently we only have one dynamic field in indexers. But I considered the possibility of multiple. Which would interfere with eachother since the options would be stored in providerOptions state directly.
by specifying a subsection we can keep them separate.

For devices I can specify the subsection in the UI code. But for newznab the action 'getNewznabCategories' is specified by the field definition.
So I didn't want to have `devices` as subsection and `getNewznabCategories`. 
But I can strip the `get` from the action when generating the section name.

src/NzbDrone.Core/Indexers/Newznab/NewznabCategoryFieldOptionsConverter.cs
```diff
@@ -0,0 +1,75 @@
...
+using NzbDrone.Core.Annotations;
+
+namespace NzbDrone.Core.Indexers.Newznab
+{
+    public static class NewznabCategoryFieldOptionsConverter
+    {
+        public static List<FieldSelectOption> GetFieldSelectOptions(List<NewznabCategory> categories)
+        {
+            // Ignore categories not relevant for Sonarr
+            var ignoreCategories = new[] { 0, 1000, 2000, 3000, 4000, 6000, 7000 };
```
ilike2burnthing, line 14: @Taloth can 7000 be _unignored_? Some indexers only have this, while others are a mess and mark 90% of uploads as 'Other'.

garfield69: Hang on a moment.
The Jackett torznab caps category table has 8000 as Books and 7000 as Other, which seems to contradict the current docs at https://newznab.readthedocs.io/en/latest/misc/api/#predefined-categories
This has been this way since Jackett was created over 5 years ago
It may be we need to amend Jackett to correct its table!

Taloth: Yeah, 7000 is books for pretty much every indexer I know of. There are some differences in the special categories between default newznab and nzedb, but the primary ones are all the same.

garfield69: https://github.com/Jackett/Jackett/pull/9986

Taloth: When I implemented this I was actually considering to return the full list of categories if the filtered list was empty, but I kinda forgot.

PR #4007, opened 2020-10-09
treat manual BAD mark in NZBGet as download failure

atryzen:
#### Database Migration
NO

#### Description
When marking item download as bad in NZBGet, Sonarr doesn't treat it as a failure and therefore doesn't preform a new search for this item.
#### Todos
- [V] Tests
- [Not needed] Documentation
*

### DIFF
--- a/src/NzbDrone.Core/Download/Clients/Nzbget/Nzbget.cs
+++ b/src/NzbDrone.Core/Download/Clients/Nzbget/Nzbget.cs
@@ -131,16 +131,16 @@ namespace NzbDrone.Core.Download.Clients.Nzbget
                 historyItem.CanBeRemoved = true;
 
                 if (item.DeleteStatus == "MANUAL")
-				{
-					if (item.MarkStatus == "BAD")
-					{
-						historyItem.Status = DownloadItemStatus.Failed;
-					}
-					else
-					{
-						continue;
-					}
-				}
+		{
+		    if (item.MarkStatus == "BAD")
+		    {
+		        historyItem.Status = DownloadItemStatus.Failed;
+		    }
+		    else
+		    {
+		        continue;
+		    }
+		}
 
                 if (!_successStatus.Contains(item.ParStatus))
                 {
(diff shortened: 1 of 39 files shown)


atryzen:
@markus101 what do you think?

src/NzbDrone.Core/Download/Clients/Nzbget/Nzbget.cs
```diff
@@ -161,7 +161,7 @@ private IEnumerable<DownloadClientItem> GetHistory()
 
                 if (!_successStatus.Contains(item.DeleteStatus) && item.DeleteStatus.IsNotNullOrWhiteSpace())
                 {
-                    if (_deleteFailedStatus.Contains(item.DeleteStatus))
+                    if (_deleteFailedStatus.Contains(item.DeleteStatus) || _deleteFailedStatus.Contains(item.MarkStatus))
```
markus101, line 164: This block has nothing to do with `MarkStatus`, it should not be responsible for marking a manually deleted item with a mark status of `BAD` as failed, same as it doesn't mark failures due to unpacking or scripts as failed here.

src/NzbDrone.Core/Download/Clients/Nzbget/Nzbget.cs
```diff
@@ -130,7 +130,7 @@ private IEnumerable<DownloadClientItem> GetHistory()
                 historyItem.CanMoveFiles = true;
                 historyItem.CanBeRemoved = true;
 
-                if (item.DeleteStatus == "MANUAL")
+                if (item.DeleteStatus == "MANUAL" && item.MarkStatus != "BAD")
```
markus101, line 133: This should be rewritten as:

```
if (item.DeleteStatus == "MANUAL")
{
    if (item.MarkStatus == "BAD")
    {
        historyItem.Status = DownloadItemStatus.Failed;
    }
    else
    {
        continue;
    }
}
```

markus101 (changes requested):
This needs a test case added to `NzbgetFixture` and one for the original behaviour of `MANUAL` since it looks like one was never added (and behaviour for both cases needs to be tested).

src/NzbDrone.Core/Download/Clients/Nzbget/Nzbget.cs
atryzen: Requested changes have been made. Regarding tests, it might take me a few more days to find the time to do that.

markus101:
Closes via ab478fd64bdf2b710fb865006858a1a7dbdbad21

PR #4061, opened 2020-11-03
Set correct theme color for Android

agneevX:
#### Database Migration
NO

#### Description
This PR sets the correct theme color based on the Sonarr frontend primary color. 

#### Todos
- [ ] Tests
- [ ] Documentation

*

### DIFF
--- a/frontend/src/Content/Images/Icons/manifest.json
+++ b/frontend/src/Content/Images/Icons/manifest.json
@@ -1,5 +1,6 @@
 {
-    "name": "",
+    "name": "Sonarr",
+	"description": "PVR for Usenet and BitTorrent users",
     "icons": [
         {
             "src": "/Content/Images/Icons/android-chrome-192x192.png",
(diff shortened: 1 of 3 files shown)


frontend/src/Content/Images/Icons/manifest.json
```diff
@@ -14,5 +15,29 @@
...
-    "display": "standalone"
-}
\ No newline at end of file
+    "start_url": "/",
+    "scope": "/",
+    "display": "standalone",
+    "shortcuts": [
+        {
+            "name": "Add new...",
+            "url": "/add/new",
```
markus101, line 24: Are these properly handled is URL base is set? I suspect that'll be an issue because this isn't a file we manipulate (like `index.html`).

agneevX: Fair point. With Base URL set to `/radarr`, I visited `radarr.home/add/new` and was immediately redirected to `radarr.home/radarr/add/new`. Is there something else you meant?

Taloth: Yes, that only works when talking directly with sonarr (it redirects), not with som reverse proxy.

You need to add something like a placeholder `{urlbase}` and add c# code to sonarr to do the substitution when serving the json to the user.

agneevX: If the user is using `/sonarr` as the base and `location /sonarr`, in nginx for e.g., then when they add the PWA, as (e.g. `sonarr.url/sonarr`), shouldn't `/add/new` automatically get appended to this URL?

Taloth: starting with `/` means root of domain. Which is why [/test](/test) points to `github.com/test` rather than `github.com/Sonarr/Sonarr/pull/4061/test`.
I'm expecting PWA to behave the same, but you should definitely test that first.

As for nginx, no. 
`sonarr.taloth.me -> 123.123.123.123:8989/sonarr` is invalid
`sonarr.taloth.me -> 123.123.123.123:8989`, urlbase = empty
`taloth.me/sonarr -> 123.123.123.123:8989/sonarr`, urlbase = 'sonarr'

So setup a reverse proxy like the last one, and test pwa.

frontend/src/Content/Images/Icons/manifest.json
```diff
@@ -12,7 +13,8 @@
             "type": "image/png"
         }
     ],
-    "theme_color": "#3a3f51",
```
markus101, line 15: Is `theme_color` no longer valid?

frontend/src/login.html
```diff
@@ -7,9 +7,9 @@
     <meta name="apple-mobile-web-app-capable" content="yes" />
 
     <!-- Chrome, Opera, and Firefox OS -->
-    <meta name="theme-color" content="#3a3f51" />
+    <meta name="theme-color" content="#2193B5" />
```
markus101, line 10: This was intentionally dark, instead of using the brighter colour, I don't recall the original discussion, but the darker colour is going to be less abrasive for most people. What's the reason to change it?

Issue #3991, opened 2020-10-02
Series Monitoring Option On Series Details Page


markus101:
**Describe the problem**
Only Series Editor (Mass Edit) and add/import series have the option to specify which episodes are monitored.

**Describe any solutions you think might work**
Add a similar option to the series details page.

RobinDadswell:
@markus101 I've started looking at this, just figuring my way around the Java script! Do you assign things here?

markus101:
Yeah, makes sense to link it to you.

PR #4067, opened 2020-11-08
New: Added Series Monitoring Toggle to Series Details

RobinDadswell:
#### Database Migration
NO

#### Description
Added Series monitoring toggle to Series detail similar to Season pass page and when adding a new season 

#### Todos
- [ ] Tests
- [ ] Documentation

* #3991 
Update button is currently below the dropdown box in the footer, I cannot seem to see why though - picture below for reference. 

image
 
I would appreciate feedback on this if anything needs changing etc.

### DIFF
--- a/frontend/src/Series/MonitoringOptions/MonitoringOptionsModalContent.js
+++ b/frontend/src/Series/MonitoringOptions/MonitoringOptionsModalContent.js
@@ -1,0 +1,140 @@
+import PropTypes from 'prop-types';
+import React, { Component } from 'react';
+import { inputTypes } from 'Helpers/Props';
+import Button from 'Components/Link/Button';
+import SpinnerButton from 'Components/Link/SpinnerButton';
+import ModalContent from 'Components/Modal/ModalContent';
+import ModalHeader from 'Components/Modal/ModalHeader';
+import ModalBody from 'Components/Modal/ModalBody';
+import ModalFooter from 'Components/Modal/ModalFooter';
+import Form from 'Components/Form/Form';
+import FormGroup from 'Components/Form/FormGroup';
+import FormLabel from 'Components/Form/FormLabel';
+import FormInputGroup from 'Components/Form/FormInputGroup';
+
+const NO_CHANGE = 'noChange';
+
+class MonitoringOptionsModalContent extends Component {
+
+  //
+  // Lifecycle
+
+  constructor(props, context) {
+    super(props, context);
+
+    this.state = {
+      monitor: NO_CHANGE
+    };
+  }
+
+  componentDidUpdate(prevProps) {
+    const {
+      isSaving,
+      saveError
+    } = prevProps;
+
+    if (prevProps.isSaving && !isSaving && !saveError) {
+      this.setState({
+        monitor: NO_CHANGE
+      });
+    }
+  }
+
+  onInputChange = ({ name, value }) => {
+    this.setState({ [name]: value });
+  }
+
+  //
+  // Listeners
+
+  onSavePress = () => {
+    const {
+      onSavePress
+    } = this.props;
+    const {
+      monitor
+    } = this.state;
+
+    const changes = {};
+
+    if (monitor !== NO_CHANGE) {
+      changes.monitor = monitor;
+      onSavePress(changes);
+    }
+  }
+
+  //
+  // Render
+
+  render() {
+    const {
+      title,
+      item,
+      isSaving,
+      onInputChange,
+      onModalClose,
+      ...otherProps
+    } = this.props;
+
+    const {
+      monitor
+    } = this.state;
+
+    return (
+      <ModalContent onModalClose={onModalClose}>
+        <ModalHeader>
+          Monitor Series - {title}
+        </ModalHeader>
+
+        <ModalBody>
+          <Form {...otherProps}>
+            <FormGroup>
+              <FormLabel>Monitoring</FormLabel>
+
+              <FormInputGroup
+                type={inputTypes.MONITOR_EPISODES_SELECT}
+                name="monitor"
+                value={monitor}
+                includeNoChange={true}
+                onChange={this.onInputChange}
+              />
+            </FormGroup>
+          </Form>
+        </ModalBody>
+
+        <ModalFooter>
+          <Button
+            onPress={onModalClose}
+          >
+            Cancel
+          </Button>
+
+          <SpinnerButton
+            isSpinning={isSaving}
+            onPress={this.onSavePress}
+          >
+            Save
+          </SpinnerButton>
+        </ModalFooter>
+      </ModalContent>
+    );
+  }
+}
+
+MonitoringOptionsModalContent.propTypes = {
+  seriesId: PropTypes.number.isRequired,
+  title: PropTypes.string,
+  item: PropTypes.object,
+  saveError: PropTypes.object,
+  monitor: PropTypes.string,
+  isSaving: PropTypes.bool.isRequired,
+  onInputChange: PropTypes.func.isRequired,
+  onSavePress: PropTypes.func.isRequired,
+  onModalClose: PropTypes.func.isRequired
+};
+
+MonitoringOptionsModalContent.defaultProps = {
+  isSaving: false
+};
+
+export default MonitoringOptionsModalContent;
--- a/frontend/src/Series/MonitoringOptions/MonitoringOptionsModalContent.js
+++ b/frontend/src/Series/MonitoringOptions/MonitoringOptionsModalContent.js
@@ -55,11 +55,8 @@ class MonitoringOptionsModalContent extends Component {
       monitor
     } = this.state;
 
-    const changes = {};
-
     if (monitor !== NO_CHANGE) {
-      changes.monitor = monitor;
-      onSavePress(changes);
+      onSavePress({ monitor });
     }
   }
 
@@ -68,8 +65,6 @@ class MonitoringOptionsModalContent extends Component {
 
   render() {
     const {
-      title,
-      item,
       isSaving,
       onInputChange,
       onModalClose,
@@ -83,7 +78,7 @@ class MonitoringOptionsModalContent extends Component {
     return (
       <ModalContent onModalClose={onModalClose}>
         <ModalHeader>
-          Monitor Series - {title}
+          Monitor Series
         </ModalHeader>
 
         <ModalBody>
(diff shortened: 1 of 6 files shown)


Thirrian:
This is something I'd put behind the Edit button, it feels like a lot of screen real estate for something I'd probably never use. Just my 2 cents, really not meant in a negative way! 🙂

markus101:
Personally I was envisioning a button in the toolbar that opened a modal with the dropdown. The footer is a lot of screen space lost for something that is going to be used occasionally. I don't think edit is the right spot since it isn't a value stored with the series.

RobinDadswell:
I will look at moving it over the next couple of weeks to be a button as you've mentioned. I only went with a footer as it was similar to where this is used in Season Pass.

I initially started looking at it in the edit modal and agreed that it wasn't the right place for this.

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

RobinDadswell:
I have updated this to be here 
image

and a modal that now looks like this
image

frontend/src/Series/MonitoringOptions/MonitoringOptionsModalContent.js
```diff
@@ -0,0 +1,140 @@
...
+    } = this.props;
+    const {
+      monitor
+    } = this.state;
+
+    const changes = {};
+
+    if (monitor !== NO_CHANGE) {
+      changes.monitor = monitor;
+      onSavePress(changes);
```
markus101, line 62: `onSavePress({ monitor });` and drop the `changes` constant.

frontend/src/Series/MonitoringOptions/MonitoringOptionsModalContentConnector.js
```diff
@@ -0,0 +1,60 @@
...
+}
+
+MonitoringOptionsModalContentConnector.propTypes = {
+  seriesId: PropTypes.number.isRequired,
+  isSaving: PropTypes.bool.isRequired,
+  saveError: PropTypes.object,
+  dispatchUpdateMonitoringOptions: PropTypes.func.isRequired,
+  onModalClose: PropTypes.func.isRequired,
+  onSavePress: PropTypes.func,
+  monitor: PropTypes.string
```
markus101, line 57: Not a prop and functions should always be last.

frontend/src/Store/Actions/seriesActions.js
```diff
@@ -377,8 +379,56 @@ export const actionHandlers = handleThunks({
...
+  [UPDATE_SERIES_MONITOR]: function(getState, payload, dispatch) {
+    const {
+      id,
+      monitored,
+      monitor
+    } = payload;
+
+    const series = [];
+    const seriesToUpdate = { id };
+    if (payload.hasOwnProperty('monitored')) {
```
markus101, line 393: New line before if block

frontend/src/Store/Actions/seriesActions.js
```diff
@@ -377,8 +379,56 @@ export const actionHandlers = handleThunks({
...
+    dispatch(set({
+      section,
+      isSaving: true
+    }));
+
+    const promise = createAjaxRequest({
+      url: '/seasonPass',
+      method: 'POST',
+      data: JSON.stringify({
+        series,
```
markus101, line 407: No need to store the series to update as a variable then add it to an array, instead just do:

```
series: [
  seriesToUpdate
]
```

frontend/src/Store/Actions/seriesActions.js
```diff
@@ -377,8 +379,56 @@ export const actionHandlers = handleThunks({
...
+    });
+
+    promise.fail((xhr) => {
+      dispatch(set({
+        section,
+        isSaving: false,
+        saveError: xhr
+      }));
+    });
+    location.reload();
```
markus101, line 430: Left over from debugging? Or a way to update the page? If the saving the modal should trigger an episode fetch after the saving finishes.

frontend/src/Series/MonitoringOptions/MonitoringOptionsModalContent.js
```diff
@@ -0,0 +1,140 @@
...
+    );
+  }
+}
+
+MonitoringOptionsModalContent.propTypes = {
+  seriesId: PropTypes.number.isRequired,
+  title: PropTypes.string,
+  item: PropTypes.object,
+  saveError: PropTypes.object,
+  monitor: PropTypes.string,
```
markus101, line 129: not a prop that is passed in.

frontend/src/Store/Actions/seriesActions.js
markus101: `if (monitored != null)` will pass if true or false, but I don't see a case where this will ever be set.

markus101 (changes requested):
A few little things and looks like one bigger piece in refetching the series/episodes after the save.

frontend/src/Store/Actions/seriesActions.js
RobinDadswell: @markus101 apologies I am completely new to JS, to be clear if I were to do this then I could get rid of both the if block and the push correct?

frontend/src/Store/Actions/seriesActions.js
RobinDadswell: It was a way to both update the page and dismiss the modal (couldn't work out why the modal wasn't closing), I will look at this bit again

frontend/src/Store/Actions/seriesActions.js
RobinDadswell: answered this myself, yes you can - just working on the other fixes before I commit again

frontend/src/Store/Actions/seriesActions.js
markus101: Typically that's done when `isSaving` completes in a `componentDidUpdate` function (at least in our pre-React Hook world).

frontend/src/Series/MonitoringOptions/MonitoringOptionsModalContent.js
RobinDadswell: This is sorted

frontend/src/Series/MonitoringOptions/MonitoringOptionsModalContentConnector.js
RobinDadswell: Done

frontend/src/Store/Actions/seriesActions.js
RobinDadswell: Done

frontend/src/Store/Actions/seriesActions.js
RobinDadswell: Done

frontend/src/Series/MonitoringOptions/MonitoringOptionsModalContent.js
RobinDadswell: resovled

frontend/src/Store/Actions/seriesActions.js
RobinDadswell: I have removed the location.reload() and got the modal to close, it doesn't however update the page still - I had a look at the toggle monitor/unmonitor for an episode and this also doesn't update the page so I believe this is in keeping with other similar features of Sonarr. 

Ideally I want it to update the page so it's more obvious that it has done what you have asked it to but unsure how to make that work as fetchEpisodes broke the page!

RobinDadswell:
Just noticed that the "None" option isn't working currently (only one I hadn't tried) so will get that working before this can be implemented.

frontend/src/Store/Actions/seriesActions.js
markus101: > I had a look at the toggle monitor/unmonitor for an episode and this also doesn't update the page so I believe this is in keeping with other similar features of Sonarr.

That's because there isn't anything to update. Toggling monitored at the series level doesn't change episodes. Toggling it at the season level does and the episode are updated, though that's locally.

> Ideally I want it to update the page so it's more obvious that it has done what you have asked it to but unsure how to make that work as fetchEpisodes broke the page!

Yeah it definitely needs to be updated. Take a look how season pass works, after the changes are applied the affected series are fetched, which is part of it, since both the series and the episodes in the series need to be fetched.

markus101:
> Just noticed that the "None" option isn't working currently (only one I hadn't tried) so will get that working before this can be implemented.

What's wrong with `None`?

RobinDadswell:
> > Just noticed that the "None" option isn't working currently (only one I hadn't tried) so will get that working before this can be implemented.
> 
> What's wrong with `None`?

It actually doesn't do anything when none is selected within the dropdown in the new modal - I think it's because it needs to make "monitored" to false maybe but I need to test that, would mean I need to put monitored back into the mix for the update action though

frontend/src/Store/Actions/seriesActions.js
RobinDadswell: This is now working

RobinDadswell:
@markus101 Can I get this reviewed again post changes please

frontend/src/Store/Actions/seriesActions.js
```diff
@@ -377,8 +380,89 @@ export const actionHandlers = handleThunks({
...
+      data: JSON.stringify({
+        series: [
+          seriesToUpdate
+        ],
+        monitoringOptions: { monitor }
+      }),
+      dataType: 'json'
+    }).request;
+    promise.done((data) => {
+
```
markus101, line 415: Nit: Remove empty line

frontend/src/Store/Actions/seriesActions.js
```diff
@@ -377,8 +380,89 @@ export const actionHandlers = handleThunks({
...
+      dispatch(fetchEpisodes({ seriesId: id }));
 
+      const episodesToUpdate = getState().episodes.items.reduce((acc, episode) => {
```
markus101, line 418: Since we're fetching all the episodes again we shouldn't have to update individual episodes (so 418-450 inclusive can be deleted).

markus101 (changes requested):
Looking great! Just a couple things to clean up and it'll be ready to land.

markus101 (changes requested):
Looking great! Just a couple things to clean up and it'll be ready to land.

frontend/src/Store/Actions/seriesActions.js
RobinDadswell: Removed

frontend/src/Store/Actions/seriesActions.js
RobinDadswell: Ahh sweet, I didn't realise that - I have removed it, much tidier!

RobinDadswell:
@markus101 all done mate

PR #4068, opened 2020-11-08
Qbittorrent - use content path property instead of save path

vinimk:
#### Database Migration
NO

#### Description
Since qbittorent 4.3.0.1 the save path is not consistent anymore, with new release of qbittorent (not released yet, but is already merged on master) (https://github.com/qbittorrent/qBittorrent/issues/13389#issuecomment-716126144 ) we have a new content_path property that is exactly what we need. This pull request changes to use this new property and it still has the fallback to the old one. Since it is a new property in a method that we already call, just a check to see if it is null should be sufficient to be backwards compatible

* #3968


rrrevin:
How long before this makes it into official v2 release? My system has been broken for over a month now :(

fryfrog:
This is already being worked on in pull request #3346, which seems to be quite a bit bigger. May want to give it a review and see what is going on.

PR #4101, opened 2020-11-17
Fixed: List Import no longer fails due to duplicates

### LINKED ISSUE #4100: V3 Import Lists Failing
**Describe the bug**
I have been trying to use Trakt lists in Sonarr V3, they do work with Limit of 100 +/- few but the lists ive been trying to use have entries in 1000 and above range.  But when setting the limit to 250 it seems the import crashes and fails to add any new entries.  I assumed that using low limits which work like 100 would allow me to use these large lists as I assumed that Sonarr would import 100 new shows every run but its not its just importing the first 100 and then the first 100 every run after that??? So that wont work for my needs I need to be able to run Lists with limits 1000 up???

**Screenshots**
image

**Logs**
https://gist.github.com/johnkennyuk/eb2702e793d6f18435e3b6909a4fd464
https://gist.github.com/johnkennyuk/eb050e6ce5a95f2ae8977dc651f5f8a4
**System Information**

 - Sonarr Version: 3.0.4.1006
Package Version 3.0.4.1006-ls77 by linuxserver.io

 - Operating System: Ubuntu server 18.04 LTS
 - Mono Version 5.20.1.34

**UI Bugs:**
NA

**Additional context**

bakerboy448:
Co-Authored-By: Qstick <376117+Qstick(at)users.noreply.github.com>

#### Database Migration
NO

#### Description
Title.

Fixes #4100

### DIFF
--- a/src/NzbDrone.Core/Tv/AddSeriesService.cs
+++ b/src/NzbDrone.Core/Tv/AddSeriesService.cs
@@ -53,10 +53,12 @@ namespace NzbDrone.Core.Tv
             return newSeries;
         }
 
-        public List<Series> AddSeries(List<Series> newSeries, bool ignoreErrors = false)
+        public List<Series> AddSeries(List<Series> newSeries, bool ignoreErrors = false) 
         {
+
             var added = DateTime.UtcNow;
             var seriesToAdd = new List<Series>();
+            var existingSeries = _seriesService.GetAllSeries();
 
             foreach (var s in newSeries)
             {
@@ -67,6 +69,22 @@ namespace NzbDrone.Core.Tv
                     var series = AddSkyhookData(s);
                     series = SetPropertiesAndValidate(series);
                     series.Added = added;
+                    if (existingSeries.Any(f => f.TvdbId == series.TvdbId))
+                    {
+                        _logger.Debug("TVDB ID {0} was not added due to validation failure: Series already exists in database", s.TvdbId);
+                        continue;
+                    }
+                    if (seriesToAdd.Any(f => f.TvdbId == series.TvdbId))
+                    {
+                        _logger.Debug("TVDB ID {0} was not added due to validation failure: Series already exists on list", s.TvdbId);
+                        continue;
+                    }
+                    var duplicateSlug = seriesToAdd.FirstOrDefault(f => f.TitleSlug == series.TitleSlug);
+                    if (duplicateSlug != null)
+                    {
+                        _logger.Debug("TVDB ID {0} was not added due to validation failure: Duplicate Slug {1} used by series {2}", s.TvdbId, s.TitleSlug, duplicateSlug.TvdbId);
+                        continue;
+                    }
                     seriesToAdd.Add(series);
                 }
                 catch (ValidationException ex)
@@ -76,7 +94,7 @@ namespace NzbDrone.Core.Tv
                         throw;
                     }
 
-                    _logger.Debug("tvdbid {0} was not added due to validation failures. {1}", s.TvdbId, ex.Message);
+                    _logger.Debug("TVDB ID {0} was not added due to validation failures. {1}", s.TvdbId, ex.Message);
                 }
             }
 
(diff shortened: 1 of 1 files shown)


src/NzbDrone.Core/Tv/AddSeriesService.cs
```diff
@@ -67,7 +67,21 @@ public List<Series> AddSeries(List<Series> newSeries, bool ignoreErrors = false)
                     var series = AddSkyhookData(s);
                     series = SetPropertiesAndValidate(series);
                     series.Added = added;
-                    seriesToAdd.Add(series);
+                    if (!seriesToAdd.Any(f => f.TitleSlug == series.TitleSlug))
```
markus101, line 70: This is only going to check against series added from this list, but this series could already exist in the DB and would fail for the same reason.

src/NzbDrone.Core/Tv/AddSeriesService.cs
```diff
@@ -67,7 +67,21 @@ public List<Series> AddSeries(List<Series> newSeries, bool ignoreErrors = false)
...
+                    {
+                        if (!seriesToAdd.Any(f => f.TvdbId == series.TvdbId))
+                        {
+                            seriesToAdd.Add(series);
+                        }
+                        else
+                        {
+                            _logger.Trace("tvdbid {0} was not added due to duplicate TVDBid in list.", s.TvdbId);
```
markus101, line 78: `TVDB ID` instead of the two mixed cases.

src/NzbDrone.Core/Tv/AddSeriesService.cs
```diff
@@ -67,7 +67,21 @@ public List<Series> AddSeries(List<Series> newSeries, bool ignoreErrors = false)
...
+                        }
+                    }
+                    else
+                    { 
+                        _logger.Debug("tvdbid {0} was not added due to duplicate title slug in list.", s.TvdbId);
```
markus101, line 83: `TVDB ID` instead of `tvdbid`

markus101 (changes requested):
I think this needs to check the existing series (already in the DB) as well as series in the list.

src/NzbDrone.Core/Tv/AddSeriesService.cs
Qstick: @markus101 Doesn't SeriesTitleSlugValidator which is run from SetPropertiesAndValidate do that already?

markus101: Yeah, you're right, still relevant for ID filtering after a redirect though.

bakerboy448:
Sounds good, I'll have that handled today!

src/NzbDrone.Core/Tv/AddSeriesService.cs
```diff
@@ -67,7 +69,13 @@ public List<Series> AddSeries(List<Series> newSeries, bool ignoreErrors = false)
                     var series = AddSkyhookData(s);
                     series = SetPropertiesAndValidate(series);
                     series.Added = added;
-                    seriesToAdd.Add(series);
+                    if (existingSeries.Any(f => f.TvdbId == series.TvdbId) || seriesToAdd.Any(f => f.TitleSlug == series.TitleSlug) || seriesToAdd.Any(f => f.TvdbId == series.TvdbId))                    {
```
markus101, line 72: Do these in separate if blocks. Something like.

```
if (existingSeries.Any(f => f.TvdbId == series.TvdbId))
{
  _logger.Debug("TVDB ID {0} was not added due to validation failure: Series already exists", s.TvdbId);
  continue;
}

var duplicateSlug = seriesToAdd.FirstOrDefault(f => f.TitleSlug == series.TitleSlug)

if (duplicateSlug != null)
{
  _logger.Debug("TVDB ID {0} was not added due to validation failure: Duplicate slug {1} used by series {2}", s.TvdbId, s.TitleSlug, duplicateSlug.TvdbId);
  continue;
}
...
```

(also fixes the spelling of encountered.

src/NzbDrone.Core/Tv/AddSeriesService.cs
```diff
@@ -67,7 +69,13 @@ public List<Series> AddSeries(List<Series> newSeries, bool ignoreErrors = false)
...
+                        _logger.Debug("TVDB ID {0} was not added due to validation failure: duplicate encounted within list or existing series.", s.TvdbId);
+                    }
+                    else
+                    {
+                        seriesToAdd.Add(series);
```
markus101, line 77: Doesn't need to be in an `else` block if continue is used in the if blocks.

markus101 (approved):
Have a couple more suggestions.

PR #4111, opened 2020-11-21
New: sequential download option for qBittorrent

mickael9:
#### Database Migration
NO

#### Description
Adds an advanced option to use sequential downloads for qBittorrent

### DIFF
--- a/src/NzbDrone.Core/Download/Clients/QBittorrent/QBittorrentProxyV1.cs
+++ b/src/NzbDrone.Core/Download/Clients/QBittorrent/QBittorrentProxyV1.cs
@@ -130,6 +130,11 @@ namespace NzbDrone.Core.Download.Clients.QBittorrent
                 request.AddFormParameter("paused", true);
             }
 
+            if (settings.SequentialDownload)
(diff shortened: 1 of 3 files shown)


markus101:
What's the advantage of downloading sequentially/why should Sonarr support it?

mickael9:
Hi, the advantage is that you can start watching while the download is still in progress.

The option is shown as a check box in qBittorrent when manually adding a new torrent, so I think it's reasonable to have the option to enable it for Sonarr-initiated downloads.

image

bakerboy448:
Sonarr won't import it until its complete though and you really shouldn't be printing plex at your downloads folder

mickael9:
I agree, but I don't use Plex. I just point mpv directly to the file I want to watch.
It's probably not the most common use case which is why I put it as an advanced option but I don't think there is any harm in having it

markus101:
We’ve discussed this and it’s not something we plan to support on qbit or other clients. A custom script that talks to the qbit API after its added by Sonarr is your best bet to achieve Hugo’s.

PR #4121, opened 2020-11-24
Update GitHub templates

bakerboy448:
-remove other issue template
-update support and add wiki
-add lock yml (Likely need to config GH still)


.github/config.yml
```diff
@@ -0,0 +1,12 @@
+blank_issues_enabled: false
+contact_links:
+  - name: Support via Discord
+    url: https://discord.gg/M6BvZn5
+    about: Chat with users and devs on support and setup related topics.
+  - name: Support via Reddit
+    url: https://reddit.com/r/Sonarr
+    about: Discuss and search thru support topics.
```
markus101, line 8: `through`

.github/config.yml
```diff
@@ -0,0 +1,12 @@
...
+  - name: Support via Forums
+    url: https://forums.sonarr.tv/
```
markus101, line 10: `about` should be the same as the one for reddit.

.github/config.yml
```diff
@@ -0,0 +1,12 @@
...
+  - name: Support via IRC
+    url: http://webchat.freenode.net/?channels=#sonarr
```
markus101, line 12: It'd be good to include `about` here too, similar to `Discord`

markus101 (changes requested):
Just minor things.

markus101 (approved):
:shipit:

Issue #4148, opened 2020-12-05
"series" URL base breaks series pages in v3


danepowell:
**Describe the bug**
When Sonarr runs under a "series" URL base, trying to access any series page (i.e. for a particular TV show) from the web UI results in a 404 in Sonarr v3. This wasn't a problem in v2.

For instance, if my server is localhost and URL base is "series", and I'm trying to access American Ninja Warrior, the UI takes me to https://localhost/series/american-ninja-warrior, which results in a 404 because the link _should_ be https://localhost/series/series/american-ninja-warrior

**Logs**
https://pastebin.com/tiX89Gn7

**System Information**

 - Sonarr Version: 3.0.4.1023
 - Operating System: Ubuntu 20.04
 - .net Framework (Windows) or mono (macOS/Linux) Version: [e.g. .net 4.5 or mono 5.12]
 - Branch: phantom-develop

**UI Bugs:**
 - OS: MacOS
 - Browser: Chrome
 - Version: [e.g. 22]

RobinDadswell:
Hi @danepowell,

I have just replicated this in my dev environment however I have a couple of questions: 
- Are you using a a reverse proxy solution? 
- Is using "series" a requirement or could you utilise another URL base (if you need a URL base at all)? 

URL base is meant to be used in reverse proxy scenarios to proxy a single domain to multiple different webpages. When setting a URL base all this actually does is enable a secondary listener address as per the following screenshot
![image](https://user-images.githubusercontent.com/19610103/101654911-ef29b680-3a38-11eb-9052-8d4e9b43cea6.png)

as such "https://localhost/series/american-ninja-warrior" would be an acceptable URL for the standard localhost binding.

danepowell:
I use an Apache reverse proxy:
```bash
/etc/apache2$ cat conf-enabled/sonarr.conf
<Location /series>
  ProxyPass http://localhost:8989/series
  ProxyPassReverse http://localhost:8989/series
</Location>
```

I don't actually access it from localhost, but from elsewhere on the network via the hostname.

Of course I could choose any other URL base, I just chose "series" to be consistent with how Sonarr describes itself.

thezoggy:
and did you set the needed setting in sonarr v3: settings > general > url base ?

danepowell:
Yep I set the URL base:
[...]

PR #4156, opened 2020-12-11
Fixed: '/series' URL Base breaking UI navigation

RobinDadswell:
#### Database Migration
NO

#### Description
Currently if the URL base matches the root path of an existing Sonarr page this causes issues in with opening said page (reported about series assumed for all other similar root paths). 

I have tested each link on each page manually and this appears to be working - the API has not been tested from outside of the Sonarr webpage however this should not cause any issues due to the webpage working. 

The logo in the Page Header was the only link that seemed to cause an issue; in testing setting the "to" value to be '/' still works as expected. 

* #4148


markus101:
Thanks!

PR #4159, opened 2020-12-13
New: Flood Download Client

jesec:
#### Database Migration
NO

#### Description
Add support for [Flood](https://flood.js.org), a modern Web UI for various torrent clients.

### DIFF
--- a/src/NzbDrone.Core/Download/Clients/Flood/Flood.cs
+++ b/src/NzbDrone.Core/Download/Clients/Flood/Flood.cs
@@ -1,0 +1,233 @@
+using System;
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Linq;
+using FluentValidation.Results;
+using NLog;
+using NzbDrone.Common.Disk;
+using NzbDrone.Common.Http;
+using NzbDrone.Core.Configuration;
+using NzbDrone.Core.Download.Clients.Flood.Models;
+using NzbDrone.Core.MediaFiles.TorrentInfo;
+using NzbDrone.Core.Parser.Model;
+using NzbDrone.Core.RemotePathMappings;
+using NzbDrone.Core.ThingiProvider;
+
+namespace NzbDrone.Core.Download.Clients.Flood
+{
+    public class Flood : TorrentClientBase<FloodSettings>
+    {
+        private readonly IFloodProxy _proxy;
+
+        public Flood(IFloodProxy proxy,
+                        ITorrentFileInfoReader torrentFileInfoReader,
+                        IHttpClient httpClient,
+                        IConfigService configService,
+                        IDiskProvider diskProvider,
+                        IRemotePathMappingService remotePathMappingService,
+                        Logger logger)
+            : base(torrentFileInfoReader, httpClient, configService, diskProvider, remotePathMappingService, logger)
+        {
+            _proxy = proxy;
+        }
+
+        private static IEnumerable<string> HandleTags(RemoteEpisode remoteEpisode, FloodSettings settings)
+        {
+            var result = new HashSet<string>();
+
+            if (settings.Tags.Any())
+            {
+                result.UnionWith(settings.Tags);
+            }
+
+            if (settings.AdditionalTags.Any())
+            {
+                foreach (var additionalTag in settings.AdditionalTags)
+                {
+                    switch (additionalTag)
+                    {
+                        case (int)AdditionalTags.TitleSlug:
+                            result.Add(remoteEpisode.Series.TitleSlug);
+                            break;
+                        case (int)AdditionalTags.Quality:
+                            result.Add(remoteEpisode.ParsedEpisodeInfo.Quality.Quality.ToString());
+                            break;
+                        case (int)AdditionalTags.Language:
+                            result.Add(remoteEpisode.ParsedEpisodeInfo.Language.ToString());
+                            break;
+                        case (int)AdditionalTags.ReleaseGroup:
+                            result.Add(remoteEpisode.ParsedEpisodeInfo.ReleaseGroup);
+                            break;
+                        case (int)AdditionalTags.Year:
+                            result.Add(remoteEpisode.Series.Year.ToString());
+                            break;
+                        case (int)AdditionalTags.Indexer:
+                            result.Add(remoteEpisode.Release.Indexer);
+                            break;
+                        case (int)AdditionalTags.Network:
+                            result.Add(remoteEpisode.Series.Network);
+                            break;
+                        default:
+                            throw new DownloadClientException("Unexpected additional tag ID");
+                    }
+                }
+            }
+
+            return result;
+        }
+
+        public override string Name => "Flood";
+        public override ProviderMessage Message => new ProviderMessage("Sonarr is unable to remove torrents that have finished seeding when using Flood", ProviderMessageType.Warning);
+
+        protected override string AddFromTorrentFile(RemoteEpisode remoteEpisode, string hash, string filename, byte[] fileContent)
+        {
+            _proxy.AddTorrentByFile(Convert.ToBase64String(fileContent), HandleTags(remoteEpisode, Settings), Settings);
+
+            return hash;
+        }
+
+        protected override string AddFromMagnetLink(RemoteEpisode remoteEpisode, string hash, string magnetLink)
+        {
+            _proxy.AddTorrentByUrl(magnetLink, HandleTags(remoteEpisode, Settings), Settings);
+
+            return hash;
+        }
+
+        public override IEnumerable<DownloadClientItem> GetItems()
+        {
+            var items = new List<DownloadClientItem>();
+
+            var list = _proxy.GetTorrents(Settings);
+
+            foreach (var torrent in list)
+            {
+                var properties = torrent.Value;
+
+                if (!Settings.Tags.All(tag => properties.Tags.Contains(tag)))
+                {
+                    continue;
+                }
+
+                var item = new DownloadClientItem
+                {
+                    DownloadClientInfo = DownloadClientItemClientInfo.FromDownloadClient(this),
+                    DownloadId = torrent.Key,
+                    Title = properties.Name,
+                    OutputPath = _remotePathMappingService.RemapRemoteToLocal(Settings.Host, new OsPath(properties.Directory)),
+                    Category = properties.Tags.Count > 0 ? properties.Tags[0] : null,
+                    RemainingSize = properties.SizeBytes - properties.BytesDone,
+                    TotalSize = properties.SizeBytes,
+                    SeedRatio = properties.Ratio,
+                    Message = properties.Message,
+                };
+
+                if (properties.Eta > 0)
+                {
+                    item.RemainingTime = TimeSpan.FromSeconds(properties.Eta);
+                }
+
+                if (properties.Status.Contains("error"))
+                {
+                    item.Status = DownloadItemStatus.Warning;
+                }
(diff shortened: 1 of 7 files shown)


jesec:
Ping @markus101.

Happy holidays!

I would be appreciated if you can give some preliminary reviews. I deployed the PR on my Sonarr/Radarr instances, and the PR is working perfectly fine so far.

Thanks.

markus101 (changes requested):
Haven't tested, but left some preliminary comments.

src/NzbDrone.Core/Download/Clients/Flood/Flood.cs
```diff
@@ -0,0 +1,221 @@
...
+                        IConfigService configService,
+                        IDiskProvider diskProvider,
+                        IRemotePathMappingService remotePathMappingService,
+                        Logger logger)
+            : base(torrentFileInfoReader, httpClient, configService, diskProvider, remotePathMappingService, logger)
+        {
+            _proxy = proxy;
+        }
+
+        private static IEnumerable<string> HandleTags(RemoteEpisode media, FloodSettings settings)
```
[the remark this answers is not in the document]
jesec: Done.

src/NzbDrone.Core/Download/Clients/Flood/FloodSettings.cs
```diff
@@ -0,0 +1,86 @@
...
+        public IEnumerable<int> AdditionalTags { get; set; }
+
+        [FieldDefinition(10, Label = "Target Ratio", Type = FieldType.Textbox, HelpText = "Ratio to reach before the torrent can be removed, -1 means infinite")]
+        public double TargetRatio { get; set; }
+
+        [FieldDefinition(11, Label = "Start on Add", Type = FieldType.Checkbox)]
+        public bool StartOnAdd { get; set; }
+
+        [FieldDefinition(12, Label = "Sequential Download", Type = FieldType.Checkbox, Advanced = true)]
+        public bool SequentialDownload { get; set; }
```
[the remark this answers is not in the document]
jesec: `SequentialDownload` removed.

src/NzbDrone.Core/Download/Clients/Flood/FloodSettings.cs
```diff
@@ -0,0 +1,86 @@
...
+namespace NzbDrone.Core.Download.Clients.Flood
+{
+    public class FloodSettingsValidator : AbstractValidator<FloodSettings>
+    {
+        public FloodSettingsValidator()
+        {
+            RuleFor(c => c.Host).ValidHost();
+            RuleFor(c => c.Port).InclusiveBetween(1, 65535);
+
+            RuleFor(c => c.TargetRatio).GreaterThanOrEqualTo(0).Unless(c => c.TargetRatio == -1);
```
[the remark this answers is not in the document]
jesec: `TargetRatio` removed.

src/NzbDrone.Core/Download/Clients/Flood/Flood.cs
```diff
@@ -0,0 +1,221 @@
...
+            }
+
+            return items;
+        }
+
+        public override DownloadClientItem GetImportItem(DownloadClientItem item, DownloadClientItem previousImportAttempt)
+        {
+            var result = item.Clone();
+
+            var contentPaths = _proxy.GetTorrentContentPaths(item.DownloadId, Settings);
```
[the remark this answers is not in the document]
jesec: Unfortunately, no. Get torrent API call provides directory path only. There are many cases: single-file vs. multi-file, base directory vs. sub directory.

When the torrent is single-file, Sonarr expects path to the file. Otherwise, it will try to import the whole directory, which can cause problems if there are unrelated files in the directory. For multi-file torrents, Sonarr expects path to the base directory of the torrent.

As such, an additional API call to fetch the list of contents of a torrent is required. It is more efficient (resource saving) to fetch when the import is requested.

src/NzbDrone.Core/Download/Clients/Flood/Flood.cs
```diff
@@ -0,0 +1,221 @@
...
+                else if (properties.Status.Contains("downloading"))
+                {
+                    item.Status = DownloadItemStatus.Downloading;
+                }
+                else if (properties.Status.Contains("stopped"))
+                {
+                    item.Status = DownloadItemStatus.Paused;
+                }
+
+                if (Settings.TargetRatio != -1 && item.SeedRatio > Settings.TargetRatio)
```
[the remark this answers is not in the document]
jesec: `TargetRatio` removed.

src/NzbDrone.Core/Download/Clients/Flood/Flood.cs
```diff
@@ -0,0 +1,214 @@
...
+
+                var item = new DownloadClientItem
+                {
+                    DownloadClientInfo = DownloadClientItemClientInfo.FromDownloadClient(this),
+                    DownloadId = torrent.Key,
+                    Title = properties.Name,
+                    OutputPath = _remotePathMappingService.RemapRemoteToLocal(Settings.Host, new OsPath(properties.Directory)),
+                    Category = properties.Tags.Count > 0 ? properties.Tags[0] : null,
+                    RemainingSize = properties.SizeBytes - properties.BytesDone,
+                    RemainingTime = TimeSpan.FromSeconds(properties.Eta),
```
markus101, line 117: This is returning `-1` for paused torrents, instead we should return null (like we do for other clients): `properties.Eta > 0 ? (TimeSpan?)(TimeSpan.FromSeconds(properties.Eta)) : null,` will do it.

src/NzbDrone.Core/Download/Clients/Flood/Flood.cs
```diff
@@ -0,0 +1,214 @@
...
+            if (contentPaths.Count < 2)
+            {
+                // For single-file torrent, OutputPath should be the path of file.
+                result.OutputPath = item.OutputPath + new OsPath(contentPaths[0]);
+            }
+            else
+            {
+                // For multi-file torrent, OutputPath should be the path of base directroy of torrent.
+                // isBasePath is false by default. Presumably base directory would be first fragment of path of content.
+                result.OutputPath = item.OutputPath + new OsPath(contentPaths[0].Split(new char[] { '\\', '/' }, StringSplitOptions.RemoveEmptyEntries)[0]);
```
markus101, line 163: This is not working for multi-file torrents. It currently returns the path to the first file, but instead I believe it should just be returning `item.OutputPath`.

markus101 (changes requested):
Found a couple issues while testing, will be able to test more once the path issue is resolved.

src/NzbDrone.Core/Download/Clients/Flood/Flood.cs
jesec: `item.OutputPath` is the `directory` of `TorrentProperties`, which, by definition, is always a directory. This directory may or may not be the base directory. This causes problems as Sonarr can grab contents of unrelated torrents which reside in the same destination directory.

This line of code tries to combine the directory with the first segment (directory) of a content when the torrent is multi-file:

TorrentProperties.directory + first path segment of first content of a multi-file torrent (path -> split -> pathSegments[0])

`/mnt/data0/downloads` + **`Big.Buck.Bunny.1080P`**/~Soundtracks/Bunny.wav~

I tested with a multi-file torrent. It seems that this approach works fine as torrent clients always create a sub-directory for multi-file torrents by default. 

For single-file torrents, Sonarr expects path to a single file, which is handled by another case (above). 

What's not working? Did this line of code fail to get the first segment, combine paths, or...?

markus101: > It currently returns the path to the first file.

It gave me the path to a text file, not a folder. Even if it happened to give me a video file it'd be wrong for a multi-video file download.

image
`result.OutputPath` is set to `C:\downloads\torrents\test\The.Rookie.S03E02.WEBRip.x264-ION10\RARBG.txt`.
`item.OutputPath` is set to `C:\timebomb\downloads\torrents\test\The.Rookie.S03E02.WEBRip.x264-ION10`

jesec: In your case, `directory` is the base directory, which is indeed a possible case. This case is not unhandled here. I assume there are some differences in how clients handle root directory of a multi-file torrent. I will work out a more robust method for detection of base directory.

markus101: Yup, they pretty much universally suck at it, we're sometimes left guessing which has lead to issues in the past and somewhat recently with qbit.

jesec: It is done. Please try it.

jesec: Sonarr can probably utilize a method that returns an array of paths (of contents of torrents) instead. That allows things to be much more precise and eliminates any interferences.

markus101: Working much better now. In the future please don't squash fixes immediately, we can do that when merging the PR.

jesec: Ack. It is a shame that Github PR does not have the capability to compare different versions of change sets directly. I typically use more professional Code Review tools like Gerrit, which does have the capability (and as result, pushing additional commits can be _"stupid"_).

jesec: Fortunately, though, there is a commit compare function: https://github.com/jesec/Sonarr/compare/17237a0e..0cae35f3.

src/NzbDrone.Core/Download/Clients/Flood/Flood.cs
```diff
@@ -0,0 +1,231 @@
...
+                            throw new DownloadClientException("Unexpected additional tag ID");
+                    }
+                }
+            }
+
+            return result;
+        }
+
+        public override string Name => "Flood";
+
```
markus101, line 79: Until we have remove support we need to add:
```
public override ProviderMessage Message => new ProviderMessage("Sonarr is unable to remove torrents that have finished seeding when using Flood", ProviderMessageType.Warning);
```

markus101 (changes requested):
Working great, just the warning to add then we'll get it landed.

src/NzbDrone.Core/Download/Clients/Flood/Flood.cs
jesec: Actually the current behavior is that Sonarr defaults to instant removal of completed torrents, which is not a desired behavior in BitTorrent community. 

If we want to add the warning, I think it would be a good idea to make `item.CanMoveFiles` and `item.CanBeRemoved` always false so Sonarr won't instant remove completed torrents. Otherwise the warning can be misleading and needs to be reworded.

markus101: Yeah you’re absolutely correct, I missed that part. That + the warning then, please.

jesec:
@markus101 Can you check what happened with CI? I don't have access to TeamCity.

bakerboy448:
@jesec there’s the little login as guest button on the page! (Not obvious, I missed it too!)

jesec:
> @jesec there’s the little login as guest button on the page! (Not obvious, I missed it too!)

Thanks.

I think it is a flaky test which does not appear to be related to this PR.

markus101:
Thanks for all the work on this @jesec!

PR #4160, opened 2020-12-13
Safety for moving files when destination already exists

beschoenen:
#### Database Migration
NO

#### Description

#### Todos
- [ ] Tests
- [ ] Wiki Updates

I lost about 20 episodes when using Sonarr's rename function.

An episode was added to a season in between other episodes, Sonarr mapped the episode files to the correct new episode, but when trying to rename the files to their new filename, I lost some files, because the destination file already existed on disk.

Example:

 - I have a season of 20 episodes (with correctly named files, `Series Name - 1x10.mp4`).
 - Now, a new episode is added for episode 10 (causing Sonarr to map episodes 11 to 21 to files 1x10 to 1x20 on disk).
 - When I try to rename the files, sonarr will try to rename episode 1x10 to 1x11, but 1x11 already exists on disk (if your media format causes a conflict, of course).

This PR aims to prevent loses in this scenario by checking if the new destination is available, otherwise queueing it for a later rename.

A fix for v2 would have my preference.

Any input would be greatly appreciated!

### DIFF
--- a/src/NzbDrone.Core/MediaFiles/RenameEpisodeFileService.cs
+++ b/src/NzbDrone.Core/MediaFiles/RenameEpisodeFileService.cs
@@ -110,11 +110,39 @@ namespace NzbDrone.Core.MediaFiles
 
         private void RenameFiles(List<EpisodeFile> episodeFiles, Series series)
         {
-            var renamed = new List<EpisodeFile>();
+            var notRenamed = TryRenameFiles(episodeFiles, series);
+
+            while (notRenamed.Any())
+            {
+                var newNotRenamed = TryRenameFiles(notRenamed, series);
+
+                if (newNotRenamed.Count == notRenamed.Count) break;
+
+                notRenamed = newNotRenamed;
+            }
+
+            if (notRenamed.Empty() || episodeFiles.Count != notRenamed.Count)
+            {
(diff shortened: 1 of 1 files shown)


src/NzbDrone.Core/MediaFiles/RenameEpisodeFileService.cs
```diff
@@ -110,19 +110,46 @@ private IEnumerable<RenameEpisodeFilePreview> GetPreviews(Series series, List<Ep
...
         private void RenameFiles(List<EpisodeFile> episodeFiles, Series series)
         {
-            var renamed = new List<EpisodeFile>();
+            var notRenamed = TryRenameFiles(episodeFiles, series);
+
+            while (notRenamed.Any())
+            {
+                var newNotRenamed = TryRenameFiles(notRenamed, series);
+
+                if (newNotRenamed.Count == notRenamed.Count) break;
```
markus101, line 119: I think we need to throw/do something for this case, otherwise it's a silent failure. Also, we either need to roll back every rename or still broadcast that the rename completed, as we'd leave files out of sync in media players.

markus101 (changes requested):
I think the case where things aren't completely renamed needs to be handled better.

PR #4192, opened 2020-12-24
Only allow one of the two checkboxes around empty folders to be checked at time.

RobinDadswell:
#### Database Migration
NO

#### Description
Both "Create empty series folders" and "Delete empty folders" can be active at the same time - this will- disable them both from being active. 

image

### DIFF
--- a/frontend/src/Settings/MediaManagement/MediaManagement.js
+++ b/frontend/src/Settings/MediaManagement/MediaManagement.js
@@ -99,6 +99,7 @@ class MediaManagement extends Component {
 
                         <FormInputGroup
                           type={inputTypes.CHECK}
+                          isDisabled={settings.deleteEmptyFolders.value && !settings.createEmptySeriesFolders.value}
                           name="createEmptySeriesFolders"
                           helpText="Create missing series folders during disk scan"
                           onChange={onInputChange}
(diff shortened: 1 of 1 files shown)


markus101 (changes requested):
We should add validation in the API side as well, that’ll direct users to fix their settings if they attempt to save them, even if they haven’t altered them.

Taloth:
As mentioned on discord, the scenario of both of them being checked it already handled internally. So we do not need to enforce it.

markus101:
Based on the conversations on discord it doesn't seem like we really need this, but I'm wondering what the original driving factor was.

RobinDadswell:
The driving factor was that I noticed there was the ability to have both settings selected which could prove confusing to users as the default behaviour isn't clear if both are selected (although it is more clear after Taloth explained it in Discord) - the consensus based on our conversations in Discord (As I am reading it) was that the tick box guard makes sense - the API validation is the bit that isn't needed

[Link to last message from Taloth on the subject](https://discord.com/channels/383686866005917708/383686866475548674/791755176339832872)

thezoggy:
these only invalidate each other for new->future state but can be useful to normalize previous folder states.
aka you had empty folder before using sonarr, having the option check to delete empty would clean it up.

Issue #4203, opened 2020-12-30
Add number of episodes next to file size inside series


Fringe09:
**Describe the problem**
Number of episodes are displayed when hovering over file size inside a series. This is not possible on a mobile phone.

**Describe any solutions you think might work**
Add number of episodes right next to the file size.

PR #4227, opened 2021-01-06
New: Show number of files as tooltip over size on disk

RobinDadswell:
#### Database Migration
NO

#### Description
Adds a box on series info to show the number of files
image

* #4203

### DIFF
--- a/frontend/src/Series/Details/SeriesDetails.css
+++ b/frontend/src/Series/Details/SeriesDetails.css
@@ -121,6 +121,7 @@
 
 .path,
 .sizeOnDisk,
+.fileCountMessage,
 .qualityProfileName,
 .network,
 .links,
--- a/frontend/src/Series/Details/SeriesDetails.css
+++ b/frontend/src/Series/Details/SeriesDetails.css
@@ -121,8 +121,10 @@
 
 .fileCountMessage {
   font-weight: 300;
-  font-size: 12px;
-  color: $iconButtonHoverLightColor;
+  font-size: 15px;
+  color: #e1e2e3;
+  white-space: nowrap;
+  padding: 5px;
 }
 
 .path,
--- a/frontend/src/Series/Details/SeriesDetails.css
+++ b/frontend/src/Series/Details/SeriesDetails.css
@@ -119,9 +119,14 @@
   margin: 5px 10px 5px 0;
 }
 
+.fileCountMessage {
+  font-weight: 300;
+  font-size: 12px;
+  color: $iconButtonHoverLightColor;
+}
+
 .path,
 .sizeOnDisk,
-.fileCountMessage,
 .qualityProfileName,
 .network,
 .links,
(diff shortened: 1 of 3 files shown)


RobinDadswell:
Updated to now look like
image

markus101 (changes requested):
Code looks good, but I find the tooltip blends in too much with the labels below. I'm not sure if we want to inverse the tooltip (dark text or lighter background) or more the tooltip above. @Taloth thoughts?

Taloth:
Afaik Robin is still moving the code to inside Label.js so it's more global.

It blends a bit in the background, but it doesn't need to 'pop'. The original tooltip is inverted. However this is more in line with the other tooltips like Links. I'm ok with a different color scheme.

RobinDadswell:
> Afaik Robin is still moving the code to inside Label.js so it's more global.
> 
> It blends a bit in the background, but it doesn't need to 'pop'. The original tooltip is inverted. However this is more in line with the other tooltips like Links. I'm ok with a different color scheme.

I have paused on that temporarily as work has started to get busy which means I am relaxing heavily in the evenings. 

The colour scheme is the same as the links tooltip as per the comment from Taloth

thezoggy:
is this media files or files in the folder? (which could include a bit more  because of subs/metadata/etc)

markus101:
Media files (same as the size is based on the number of media files).

thezoggy:
> 
> 
> Media files (same as the size is based on the number of media files).

then, just show the number of files with the size to group common context ?

RobinDadswell:
> then, just show the number of files with the size to group common context ? (which it actually does show it already on the size as a tooltip) - that's what I am doing; but it's about converting the logic from tooltip and making it work across all other popups like that.

thezoggy:
quick mockup of what I was talking about:
sonarr

Taloth:
We don't want to add more tags or make existing ones bigger. The point here is to make sure the tooltips work on touch devices, and that's it.

austinwbest:
Still think this should be n/n and not just the amount of files. 50/100 episodes (same concept as the index)

frontend/src/Series/Details/SeriesDetails.css
```diff
@@ -119,6 +119,14 @@
   margin: 5px 10px 5px 0;
 }
 
+.fileCountMessage {
+  font-weight: 300;
+  font-size: 15px;
+  color: #e1e2e3;
```
markus101, line 125: I think `color: $white` works a bit better, though I think it's better in `Tooltip.css` in the `&.inverse` section:

```
&.inverse {
  background-color: $themeDarkColor;
  box-shadow: 0 5px 10px $popoverShadowInverseColor;
  color: $white;
}
```

Also looks like these are in the wrong order and failing in CI.

markus101 (changes requested):
After playing with the other options I think the styling mostly works, just a change to the text.

I think only showing episode files makes sense in this case given episodes without files don't contribute to the size on disk.

Let's get this landed after the small fixups and we can look at refactoring the rest at another time.

RobinDadswell:
@markus101  ready for another go!

PR #4248, opened 2021-01-12
Fixed: Parse standalone UHD as 4K if no other resolution info is present

bakerboy448:
#### Database Migration
NO

#### Description
Title

#### Todos
- [X] Tests
- Wiki Updates-N/A

N/A

Reference: https://www.reddit.com/r/sonarr/comments/kvsr4n/sonarr_think_uhd_release_is_fhd/


bakerboy448:
For posterity:
https://discord.com/channels/383686866005917708/383687854057979906/798646691997417492

An edge case that may cause an import loop and suboptimal behavior exists if the following are true:
- mediainfo is not available
- file name is garbage
- folder name has resolution of UHD AND file name has a different resolution OR  or folder name has an actual resolution  AND file name has a different resolution

thezoggy:
Why the movie tests for sonarr?

bakerboy448:
it's what we found that matched the criteria needed for test cases rather than mocking up fake releases....so really just what was easiest

PR #4249, opened 2021-01-13
Update BTN API URL to use HTTPS instead of HTTP

merqwury:
#### Database Migration
NO

#### Description
It seems overnight the Sonarr <-> BTN API Integration has stopped working giving an error `Invalid torrent file specified —> MonoTorrent.BEncoding.BEncodingException: Could not find what value to decode`

Changing the URL to use the HTTPS URL of the API instead of the HTTP, alleviates the issue.


Taloth:
Add a migration to fixup existing ones. There are some migrations that do this for other indexers, for inspiration.
Be sure to use a migration number that's not yet reserved in existing Pull Requests.

PR #4288, opened 2021-01-28
Updated Nyaa Indexer Default Settings

KennyStryker:
#### Database Migration
YES

#### Description
1. Added base URL.

2. Nyaa recently updated their parameters, so updated their default parameters in NyaaSettings since old parameters no longer work.

#### Todos
- [X] Tests
- [X] Wiki Updates


markus101 (changes requested):
Needs a migration to fix the old cats if they no longer work

PR #4290, opened 2021-01-29
New: Add name field to release profiles

rcdailey:
This adds a new, non-functional field to release profiles for a friendly
name. This name may be used, for example, by scripts invoking the API to
look for specifically named profiles in order to find their ID, which is
required to perform further operations on it. Additionally it adds an
additional organization tool for users to manage their release profiles.

#### Database Migration
YES

#### Description
See above

### DIFF
--- a/frontend/src/Settings/Profiles/Release/EditReleaseProfileModalContent.js
+++ b/frontend/src/Settings/Profiles/Release/EditReleaseProfileModalContent.js
@@ -29,6 +29,7 @@ function EditReleaseProfileModalContent(props) {
 
   const {
     id,
+    name,
     enabled,
     required,
     ignored,
@@ -46,6 +47,20 @@ function EditReleaseProfileModalContent(props) {
 
       <ModalBody>
         <Form {...otherProps}>
+
+          <FormGroup>
+            <FormLabel>Name</FormLabel>
+
+            <FormInputGroup
+              type={inputTypes.TEXT}
+              name="name"
+              {...name}
+              placeholder="Optional name"
+              canEdit={true}
+              onChange={onInputChange}
+            />
+          </FormGroup>
+
           <FormGroup>
             <FormLabel>Enable Profile</FormLabel>
 
--- a/frontend/src/Settings/Profiles/Release/ReleaseProfile.js
+++ b/frontend/src/Settings/Profiles/Release/ReleaseProfile.js
@@ -56,6 +56,7 @@ class ReleaseProfile extends Component {
   render() {
     const {
       id,
+      name,
       enabled,
       required,
       ignored,
@@ -79,6 +80,14 @@ class ReleaseProfile extends Component {
         overlayContent={true}
         onPress={this.onEditReleaseProfilePress}
       >
+        {
+          name ?
+            <div className={styles.name}>
+              {name}
+            </div> :
+            null
+        }
+
         <div>
           {
             split(required).map((item) => {
@@ -184,6 +193,7 @@ class ReleaseProfile extends Component {
 
 ReleaseProfile.propTypes = {
   id: PropTypes.number.isRequired,
+  name: PropTypes.string,
   enabled: PropTypes.bool.isRequired,
   required: PropTypes.string.isRequired,
   ignored: PropTypes.string.isRequired,
--- a/frontend/src/Settings/Profiles/Release/ReleaseProfile.css
+++ b/frontend/src/Settings/Profiles/Release/ReleaseProfile.css
@@ -9,3 +9,11 @@
   flex-wrap: wrap;
   margin-top: 5px;
 }
+
+.name {
+  @add-mixin truncate;
+
+  margin-bottom: 20px;
+  font-weight: 300;
+  font-size: 24px;
+}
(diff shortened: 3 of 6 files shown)


rcdailey:
Screenshots of the changes:

image

image

New field in API response:

```json
[
  {
    "name": "Friendly Name",
    "enabled": true,
    "required": "asdf",
    "ignored": "asdfdddd",
    "preferred": [
      {
        "key": "111",
        "value": 100
      }
    ],
    "includePreferredWhenRenaming": true,
    "indexerId": 0,
    "tags": [],
    "id": 1
  }
]
```

bakerboy448:
Wiki updates will be needed as well

rcdailey:
Do you mean here? https://wiki.servarr.com/Sonarr

What pages should get updates and how may I update them?

Thank you for the feedback.

rcdailey:
My latest push amends in changes to address the eslint problems

Drazzilb08:
https://wiki.servarr.com/Settings_Profiles
Here is the page you would update with the information regarding your PR. Find the section for release profiles then edit Sonarr's section as needed. The language is MediaWiki. If you're unfamiliar with MediaWiki Syntax https://www.mediawiki.org/wiki/Help:Formatting will help you out.

rcdailey:
Since the wiki is not versioned with this change, I will update the wiki only after this merges. Until then, I feel it would be misleading to other readers. Thank you for pointing it out, I will make sure it gets done.

bakerboy448:
Yup my thoughts exactly!

either update the wiki once it merges or just before once Markus said it looks alright

markus101 (changes requested):
Thanks for the screenshots! Just the one comment.

frontend/src/Settings/Profiles/Release/EditReleaseProfileModalContent.js
```diff
@@ -46,6 +47,21 @@ function EditReleaseProfileModalContent(props) {
...
+
+          <FormGroup>
+            <FormLabel>Name</FormLabel>
+
+            <FormInputGroup
+              type={inputTypes.TEXT}
+              name="name"
+              {...name}
+              helpText="An optional friendly name for this release profile"
+              placeholder="Optional name"
```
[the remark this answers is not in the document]
rcdailey: I will go ahead and remove `helpText` per your suggestion, which I agree with. My original goal was to keep it consistent with other fields on this form; I don't recall any that were missing help text. Thank you for the suggestion. Look forward to the fix in my next push.

frontend/src/Settings/Profiles/Release/ReleaseProfile.js
```diff
@@ -79,6 +80,9 @@ class ReleaseProfile extends Component {
         overlayContent={true}
         onPress={this.onEditReleaseProfilePress}
       >
+        <div className={styles.name}>
```
markus101, line 83: After playing with it for a moment we should conditionally render this so we don't have a margin without  a name

```
{
  name ?
    <div className={styles.name}>
      {name}
    </div> :
  null
}
```

Please also add a blank line between the closing brace and the next `<div>`.

markus101 (changes requested):
A small styling change then we're good to land it.

frontend/src/Settings/Profiles/Release/ReleaseProfile.js
rcdailey: I've made the requested change. Note that on the `null` line, the linter was not satisfied with the indentation you chose, so I aligned it with the `</div>` and it was happy with that.

Let me know if there's anything else you'd like changed. Thank you for the quick turnaround.

rcdailey:
Do not merge yet, I'm hitting an issue I need to debug:

```
{
  "message": "constraint failed\r\nNOT NULL constraint failed: ReleaseProfiles.Name"
}
```

rcdailey:
Ok I resolved the previous issue, it will be in my next push but I wanted to comment first. This time I did the following testing:

1. Delete all sqlite databases under `C:\ProgramData\Sonarr`
1. Run `phantom-develop` version of Sonarr
1. Created a release profile
1. Switch to my branch, rebuild, relaunch Sonarr
1. Create another release profile without a name
1. Create a third release profile *with* a name

I wanted to make sure I properly tested the upgrade path this time. Here are the results as screenshots. I used "DB Browser" to inspect the sqlite database before and after the upgrade:

Before upgrade, this is what the schema and ReleaseProfiles table looked like after step 3:

image

image

After the upgrade, I created 2 more release profiles (one with and without name). Note the following for all 3 release profiles:

1. The before_upgrade one has its default name set to `NULL` (expected)
2. The after_upgrade without name is set to `NULL` (different code path, but also expected)
3. The after_upgrade WITH name is properly set to a non-null string

image

image

I hope this time I've got it working pretty good. I apologize for not having tested this thoroughly enough when I started. If you see any other potential issues please let me know. Thank you again for your patience.

rcdailey:
@markus101 Markus, is there anything else I can do to help move this one forward? Thank you again for your time and feedback.

markus101:
Thanks for contributing!

PR #4308, opened 2021-02-05
Update feature request template

bakerboy448:
git

### DIFF
--- a/.github/ISSUE_TEMPLATE/feature_request.md
+++ b/.github/ISSUE_TEMPLATE/feature_request.md
@@ -1,14 +1,20 @@
 ---
 name: Feature request
 about: Suggest an idea for Sonarr
+title: ''
+labels: ''
+assignees: ''
 
 ---
 
-**Describe the problem**
-A clear and concise description of the problem you're looking to solve.
+**Is your feature request related to a problem? Please describe.**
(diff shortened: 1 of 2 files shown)


.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -1,14 +1,20 @@
 ---
 name: Feature request
 about: Suggest an idea for Sonarr
+title: ''
+labels: enhancement
```
markus101, line 5: I actually hate this about the bug template, half the bugs that get filed aren't bugs and new issues that need to be triaged get labeled as bugs straight away. IMO we shouldn't set a label so we can triage properly.

markus101 (changes requested):
Personally I hate the auto tagging of issues.

.github/ISSUE_TEMPLATE/feature_request.md
bakerboy448: I'll update the bug template as well then, incoming pushes

Might I suggest a Support label then so we don't have any non-labeled issues and can reserve non-labeled for items needing triage?  That'll allow us to tag the support GHIs accordingly.

markus101: Not opposed to a label, but they get insta-closed most of the time.

PR #4311, opened 2021-02-06
New: Add logo to loading page

Snarkenfaugister:
#### Database Migration
NO

#### Description
Add 48px logo to loading screen, above message.
logo

#### Issues Fixed or Closed by this PR

* #4304

PR #4316, opened 2021-02-09
Fixed: Use SVG for loading page icon

Snarkenfaugister:
Use SVG image for loading page icon instead of base64 png

#### Database Migration
NO

#### Description
Address issues raised in #4311

As above


Taloth:
You have to embed the svg just like the png was. We don't want additional resource _loads_ for what's just a _loading screen_. The irony... :smile:

Snarkenfaugister:
> 
> 
> You have to embed the svg just like the png was. We don't want additional resource _loads_ for what's just a _loading screen_. The irony... 😄

And here I was thinking that was the wrong way to do things. Reverted to the original. 😄

Taloth:
Normally you'd be right, but since it's a loading screen you don't want to slow down the loading just to get a friggin icon loaded... for the loading screen.

Taloth:
Isn't it possible to embed it via the base64?

Issue #4167, opened 2020-12-15
Add Recommended to the List types for Trakt


lsaadeh:
**Describe the problem**
Add Recommended to the List types for Trakt

**Describe any solutions you think might work**
Obtain the recommended list similarly to how to other lists types are obtained

**Additional context**
https://trakt.tv/shows/recommended/monthly

PR #4317, opened 2021-02-09
New: Add Recommended to the List types for Trakt

Snarkenfaugister:
#### Database Migration
NO

#### Description
Adds recommended trakt lists for week/month/year/all, similar to top watched.
Are all four wanted in this list?
recommended

* #4167

### DIFF
--- a/src/NzbDrone.Core/ImportLists/Trakt/Popular/TraktPopularListType.cs
+++ b/src/NzbDrone.Core/ImportLists/Trakt/Popular/TraktPopularListType.cs
@@ -18,6 +18,15 @@ namespace NzbDrone.Core.ImportLists.Trakt.Popular
         [EnumMember(Value = "Top Watched Shows By Year")]
         TopWatchedByYear = 5,
         [EnumMember(Value = "Top Watched Shows Of All Time")]
-        TopWatchedByAllTime = 6
+        TopWatchedByAllTime = 6,
+
+        [EnumMember(Value = "Recommended Shows By Week")]
+        RecommendedByWeek = 7,
+        [EnumMember(Value = "Recommended Shows By Month")]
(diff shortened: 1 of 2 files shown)


src/NzbDrone.Core/ImportLists/Trakt/Popular/TraktPopularListType.cs
```diff
@@ -18,6 +18,15 @@ public enum TraktPopularListType
         [EnumMember(Value = "Top Watched Shows By Year")]
         TopWatchedByYear = 5,
         [EnumMember(Value = "Top Watched Shows Of All Time")]
-        TopWatchedByAllTime = 6
+        TopWatchedByAllTime = 6,
+
```
markus101, line 22: Nit: Unnecessary new line

src/NzbDrone.Core/ImportLists/Trakt/Popular/TraktPopularRequestGenerator.cs
```diff
@@ -52,6 +52,18 @@ private IEnumerable<ImportListRequest> GetSeriesRequest()
                 case (int)TraktPopularListType.TopWatchedByAllTime:
                     link += "/shows/watched/all" + filtersAndLimit;
                     break;
+                case (int)TraktPopularListType.RecommendedByWeek:
```
markus101, line 55: I think a small refactor here will go a long way, we don't need to add `filtersAndLimit` to each case, we can add it at the end, each case would change to something like: `link += "/shows/recommended/weekly"` and the outside the switch, `link += filtersAndLimit`

markus101 (changes requested):
One nit and a small refactor to clear things up a bit.

Issue #4323, opened 2021-02-13
Ensure Consistency of SSL option across download Clients


Paulus88:
I have all my web solutions behind a webproxy and this has been setup to only allow HTTPS/443 traffic.
It would be awsome if SSL will be added in Sonarr for all Download Clients.

bakerboy448:
Please use one of the support channels: [forums](https://forums.sonarr.tv/), [subreddit](https://www.reddit.com/r/sonarr/), [discord ](https://discord.gg/Ex7FmFK), or [IRC ](http://webchat.freenode.net/?channels=#sonarr)for support/questions.

Closing- template not followed

and SSL already exists for download clients

Paulus88:
An example, this client has no SSL option:
![image](https://user-images.githubusercontent.com/25663775/107853364-c990a300-6e15-11eb-9757-608cb5cf1dca.png)

It exists for some download clients, not all.

bakerboy448:
Just looked:

only Utorrent is missing an SSL option

however, SSL is also not consistent across download down clients:
- some have it as advanced
- Some have help test, most don’t

fryfrog:
@Paulus88: A *support* suggestion for you, running all this software through your reverse proxy is adding a single point of failure w/o benefit if your services are all on the same host or trusted lan. If they're crossing the wide open internet or an untrusted lan, it does make sense though.

Paulus88:
> @Paulus88: A _support_ suggestion for you, running all this software through your reverse proxy is adding a single point of failure w/o benefit if your services are all on the same host or trusted lan. If they're crossing the wide open internet or an untrusted lan, it does make sense though.

Hi @fryfrog, thanks for the reply, I know this already but its not a corporate solution, its just a hobby solution that I like to keep secure from LAN and WAN. All environments also have their own host that automatically keep them selves up to date and clean with cronjobs/scheduled tasks and the entrance to these hosts is the webproxy very tightly configured with the firewalls on the hosts. This also allows me to turn an envirement off and replace it if ever needed without effecting other solutions.
Even if the webproxy fails it will have very little impact as it is only used to administrator applications through the browser and as you mentioned I also like to access it from other locations so webproxy seems the safest solution.

fryfrog:
I'm absolutely not suggesting you stop using a reverse proxy, they're the bee's knees and I <3 it too.

[...]

PR #4326, opened 2021-02-15
Fix: Consistent SSL option for Download Clients

RobinDadswell:
#### Database Migration
YES

#### Description
To enable consistent ability to utilise SSL for all relevant download clients

#### Todos
- [ ] Tests
- [ ] Wiki Updates

* #4323


RobinDadswell:
@bakerboy448  is there a wiki article to update for this?

bakerboy448:
@RobinDadswell technically speaking no... but the download clients wiki entries could use some work... maybe at least a settings section that explains the options?

https://wiki.servarr.com/Settings_Download_Clients is the page to edit

RobinDadswell:
> @RobinDadswell technically speaking no... but the download clients wiki entries could use some work... maybe at least a settings section that explains the options?
> 
> https://wiki.servarr.com/Settings_Download_Clients is the page to edit

I will have a look at it later; each client has slightly different settings though based on what it needs etc. so it may be a large Wiki piece - maybe not one for that page but as a subpage!

src/NzbDrone.Core/Download/Clients/NzbVortex/NzbVortexSettings.cs
```diff
@@ -42,19 +43,22 @@ public NzbVortexSettings()
         [FieldDefinition(1, Label = "Port", Type = FieldType.Textbox)]
         public int Port { get; set; }
 
-        [FieldDefinition(2, Label = "Url Base", Type = FieldType.Textbox, Advanced = true, HelpText = "Adds a prefix to the NZBVortex url, e.g. http://[host]:[port]/[urlBase]/api")]
+        [FieldDefinition(2, Label = "Use SSL", Type = FieldType.Checkbox, HelpText = "Use secure connection when connecting to NZBVortex")]
```
markus101, line 46: I think NZBVortex is the odd man out, where they only support SSL connections and using a reversing proxy to connect insecurely is indirectly. I think sticking with no setting makes sense.

src/NzbDrone.Core/Datastore/Migration/156_update_download_client_ssl_options.cs
```diff
@@ -0,0 +1,84 @@
...
+using FluentMigrator;
+using Newtonsoft.Json.Linq;
+using NzbDrone.Common.Extensions;
+using NzbDrone.Common.Serializer;
+using NzbDrone.Core.Datastore.Migration.Framework;
+
+namespace NzbDrone.Core.Datastore.Migration
+{
+    [Migration(156)]
+    public class update_download_client_ssl_options : NzbDroneMigrationBase
```
markus101, line 12: I don't think this migration is necessary, it should inherit the default from settings when parsed from the DB.

markus101 (changes requested):
I think NZBVortex should not have an option (they enforce SSL) and the migrations aren't necessary.

src/NzbDrone.Core/Datastore/Migration/156_update_download_client_ssl_options.cs
RobinDadswell: removed

src/NzbDrone.Core/Download/Clients/NzbVortex/NzbVortexSettings.cs
RobinDadswell: Updated :)

RobinDadswell:
Current failing test is not related to the changes made

markus101 (approved):

PR #4397, opened 2021-03-20
Add Tags to existing Series from multiple Import Lists

danimart1991:
#### Database Migration
NO

#### Description

Add Tags to existing Series if available in multiple Import Lists.

#### Todos
- [ ] Tests
- [ ] Wiki Updates

### DIFF
--- a/src/NzbDrone.Core/ImportLists/ImportListSyncService.cs
+++ b/src/NzbDrone.Core/ImportLists/ImportListSyncService.cs
@@ -101,6 +101,9 @@ namespace NzbDrone.Core.ImportLists
                 if (existingSeries != null)
                 {
                     _logger.Debug("{0} [{1}] Rejected, Series Exists in DB", report.TvdbId, report.Title);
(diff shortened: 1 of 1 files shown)


bakerboy448:
Movie? 👀 😉

danimart1991:
> Movie? 👀 😉

Fixed. 😅 Thanks for review.

markus101:
We’ve discussed an this isn’t something we are going to support. Tags will be added when Sonarr adds the series, otherwise they will not be tagged. This allows people to tag the series with a unique tag for the list and not have it alter tags on existing series.

danimart1991:
> We’ve discussed an this isn’t something we are going to support. Tags will be added when Sonarr adds the series, otherwise they will not be tagged. This allows people to tag the series with a unique tag for the list and not have it alter tags on existing series.

I think this problem could be fixed if the "add multiple tags" is an option, but ok... Thanks for review it. 😉

PR #4398, opened 2021-03-20
Fixed: Parsing RSS with null values

cicomalieran:
#### Database Migration
NO

#### Description
Given that [XElement.Attribute()](https://docs.microsoft.com/en-us/dotnet/api/system.xml.linq.xelement.attribute?view=net-5.0) will return a null value when it cannot find the attribute, defensive programing is required.

Given that [XAttribute](https://docs.microsoft.com/en-us/dotnet/api/system.xml.linq.xattribute?view=net-5.0) is a class that has nothing to do with long/Int64, the Value property should be accessed and parsed instead of using a cast.

#### Todos
- [x] Tests - unit tests passing.
- [ ] Wiki Updates - Not applicable? This is my first contribution here.

* https://github.com/lidarr/Lidarr/issues/1730

### DIFF
--- a/src/NzbDrone.Core/Indexers/RssParser.cs
+++ b/src/NzbDrone.Core/Indexers/RssParser.cs
@@ -257,12 +257,11 @@ namespace NzbDrone.Core.Indexers
                                  {
                                      try
                                      {
-                                         var lengthValue = v.Attribute("length")?.Value;
                                          return new RssEnclosure
                                                 {
                                                     Url = v.Attribute("url")?.Value,
                                                     Type = v.Attribute("type")?.Value,
-                                                    Length = string.IsNullOrEmpty(lengthValue) ? default(long) : XmlConvert.ToInt64(lengthValue)
+                                                    Length = v.Attribute("length")?.Value?.ParseInt64() ?? default
                                                 };
                                      }
                                      catch (Exception e)
--- a/src/NzbDrone.Core/Indexers/RssParser.cs
+++ b/src/NzbDrone.Core/Indexers/RssParser.cs
@@ -261,7 +261,7 @@ namespace NzbDrone.Core.Indexers
                                                 {
                                                     Url = v.Attribute("url")?.Value,
                                                     Type = v.Attribute("type")?.Value,
-                                                    Length = v.Attribute("length")?.Value?.ParseInt64() ?? default
+                                                    Length = v.Attribute("length")?.Value?.ParseInt64() ?? 0
                                                 };
                                      }
                                      catch (Exception e)
--- a/src/NzbDrone.Core/Indexers/RssParser.cs
+++ b/src/NzbDrone.Core/Indexers/RssParser.cs
@@ -257,11 +257,12 @@ namespace NzbDrone.Core.Indexers
                                  {
                                      try
                                      {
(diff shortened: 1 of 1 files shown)


src/NzbDrone.Core/Indexers/RssParser.cs
```diff
@@ -257,11 +257,12 @@ protected virtual RssEnclosure[] GetEnclosures(XElement item)
                                  {
                                      try
                                      {
+                                         var lengthValue = v.Attribute("length")?.Value;
```
markus101, line 260: ~~Nit: new line before return~~ Eliminated via next comment.

src/NzbDrone.Core/Indexers/RssParser.cs
```diff
@@ -257,11 +257,12 @@ protected virtual RssEnclosure[] GetEnclosures(XElement item)
...
                                          return new RssEnclosure
                                                 {
-                                                    Url = v.Attribute("url").Value,
-                                                    Type = v.Attribute("type").Value,
-                                                    Length = (long)v.Attribute("length")
+                                                    Url = v.Attribute("url")?.Value,
+                                                    Type = v.Attribute("type")?.Value,
+                                                    Length = string.IsNullOrEmpty(lengthValue) ? default(long) : XmlConvert.ToInt64(lengthValue)
```
markus101, line 265: We have an extension that makes this cleaner:

`Length = v.Attribute("length")?.Value.ParseInt64() ?? 0`;

No need for the separate `lengthValue` variable and `0` is more obvious than `default(long)` I think.

markus101 (changes requested):
Just a small tweak needed.

src/NzbDrone.Core/Indexers/RssParser.cs
cicomalieran: thanks! solved. default is beneficial compared to 0 because you can search for references with default. (should you ever have the question "What are the places where I'm setting the default value?". I simplified it to the literal expression instead of using the operator, hopefully that makes it easier to read.

src/NzbDrone.Core/Indexers/RssParser.cs
```diff
@@ -259,9 +259,9 @@ protected virtual RssEnclosure[] GetEnclosures(XElement item)
                                      {
                                          return new RssEnclosure
                                                 {
-                                                    Url = v.Attribute("url").Value,
-                                                    Type = v.Attribute("type").Value,
-                                                    Length = (long)v.Attribute("length")
+                                                    Url = v.Attribute("url")?.Value,
+                                                    Type = v.Attribute("type")?.Value,
+                                                    Length = v.Attribute("length")?.Value?.ParseInt64() ?? default
```
markus101, line 264: `error CS8107: Feature 'default literal' is not available in C# 7.0. Please use language version 7.1 or greater.`

At this time we're sticking with C# 7, changing to `0` will resolve.

markus101 (changes requested):
Failing on CI.

src/NzbDrone.Core/Indexers/RssParser.cs
cicomalieran:
```suggestion
                                                    Length = v.Attribute("length")?.Value?.ParseInt64() ?? 0
```

PR #4406, opened 2021-03-25
New: Add option to import from any user's personal list

danimart1991:
#### Database Migration
NO

#### Description
Actually, Trakt Import List feature only allows personal list from authenticated user. This new option allow the user to set any user to import his lists.


markus101:
@Qstick do you foresee any issues with this?

danimart1991:
> @Qstick do you foresee any issues with this?

Hello @Qstick. This is not a fix. I want to import series from my wife's Trakt and my Trakt too. The problem is that the Import Lists -> List option only allows to import manual created Lists from a user, not its personal lists.

Qstick:
What happens when you don't have access to the user's list that you are trying to access from the authed user?

danimart1991:
> What happens when you don't have access to the user's list that you are trying to access from the authed user?

Hi @Qstick. I just test to change a public profile to private to check this.

Sonarr brings an error and in log you could see a HTTP 401 error.
image
image

Qstick:
Seems ok to me @markus101, as long as you are good with above error handling.

Issue #3366, opened 2019-10-29
Reanalyze media files for proper media management rename


nismanoku:
The problem is that i download an episode and have it imported to my media directory. The file is the gets a new filename, with audiolanguage and subtitlelanguage tags, as set in sonarr, this works good. But then i noticed that some episodes have their language tags not set so the episode file is named incorrectly because the language tags where missing from the original file (mkv). So i manually update the tags of the file to the proper ones. But then i have no way to have sonarr reanalyze the file and rename it with the proper language tags.

Would like to see a button to manually reanalyze a file and to rename it with the reanalyzed tags. Down under a sample.

* Animal Kingdom (2016) S04E01 AMZN 1080P Web-DL EAC3 5.1 h264 [NL] -NTb

Changed the audio tags to proper language as it was not set. and press button in sonarr ;) to reanalyze and rename and voila the result down under.

* Animal Kingdom (2016) S04E01 AMZN 1080P Web-DL EAC3 5.1 **[EN]** h264 [NL] -NTb

Now i rename them manually, but sometimes there are a lot and i batch tag them and reimport a season or complete tv-show so they get reanalyzed and named properly. 

By the way love the prefer name tags have AMZN automatically set with it as you can see in the sample.

markus101:
I don't see the problem. The audio language tags come from the media info, not the file name. If there is a new file name Sonarr will re-analyze it unless you have `Analyse video files` disabled.

nismanoku:
I change the audio language tags of the file manually (batch) then i have to manually edit or reimport the file for the language tags to be updated or manually change filename, because i can't reanalyse the file in sonarr if the filename doesn't change. Rescan & Refresh doesn't update media info of the file. Would be handy to have a option to reanalyse a file so the renamer of sonarr can set the updated media info tags of the file to the filename automatic without having to change filename manually or to have the file reimported.

markus101:
> I change the audio language tags of the file manually 

Changed them how?

nismanoku:
I use mkvpropedit to batch multiple and mkvtoolnix for single files

nickp85:
[...]

PR #4418, opened 2021-03-30
New: Setting SceneName and ReleaseGroup for EpisodeFiles via API

mdhiggins:
Include releasegroup in the API output
allow sceneName and releaseGroup to be modified via PUT request

#### Database Migration
NO

#### Description
Allow releaseGroup to be read via episodeFile API request similar to how movieFile functions on Radarr

Allow both releaseGroup and sceneName to be modified by the API

Uses here would be primarily for post processing scripts to include this information if its lost or missing from Sonarr. My primary use case is that I have a script that rewraps MKV files in and MP4 container. After this file extension change the information about the sceneName and releaseGroup is lost and it makes finding subtitles and other release specific things more difficult. This would allow that information to be restored after a rescan is completed.

I have created a similar pull request for Radarr for sceneName though Radarr already supports reading and writing to the releaseGroup property.

#### Todos
- [ ] Tests
- [ ] Wiki Updates

### DIFF
--- a/src/Sonarr.Api.V3/EpisodeFiles/EpisodeFileModule.cs
+++ b/src/Sonarr.Api.V3/EpisodeFiles/EpisodeFileModule.cs
@@ -94,8 +94,14 @@ namespace Sonarr.Api.V3.EpisodeFiles
         {
             var episodeFile = _mediaFileService.Get(episodeFileResource.Id);
             episodeFile.Quality = episodeFileResource.Quality;
-            episodeFile.SceneName = episodeFileResource.SceneName;
-            episodeFile.ReleaseGroup = episodeFileResource.ReleaseGroup;
+            if (episodeFileResource.SceneName != null)
+            {
+                episodeFile.SceneName = episodeFileResource.SceneName;
+            }
+            if (episodeFileResource.ReleaseGroup != null)
+            {
+                episodeFile.ReleaseGroup = episodeFileResource.ReleaseGroup;
+            }
             _mediaFileService.Update(episodeFile);
         }
 
--- a/src/Sonarr.Api.V3/EpisodeFiles/EpisodeFileModule.cs
+++ b/src/Sonarr.Api.V3/EpisodeFiles/EpisodeFileModule.cs
@@ -94,6 +94,8 @@ namespace Sonarr.Api.V3.EpisodeFiles
         {
             var episodeFile = _mediaFileService.Get(episodeFileResource.Id);
             episodeFile.Quality = episodeFileResource.Quality;
+            episodeFile.SceneName = episodeFileResource.SceneName;
+            episodeFile.ReleaseGroup = episodeFileResource.ReleaseGroup;
             _mediaFileService.Update(episodeFile);
         }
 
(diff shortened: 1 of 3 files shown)


TheCatLady:
+1

As a workaround for #3366, I am (1) renaming/moving the original media file, (2) refreshing, (3) writing a modified media file using the original file name (and deleting the renamed/moved original file), and (4) refreshing a second time.

This is currently the only way to get Sonarr to re-analyze a file when the filename is unchanged, so allowing the scene name and release group to be set via the API would be very helpful.

AFAIK, there is currently no other way to do any processing/modification of files and have the changes picked up while preserving the original release group metadata.

Same goes for Radarr/Radarr#6132.

src/Sonarr.Api.V3/EpisodeFiles/EpisodeFileModule.cs
```diff
@@ -113,6 +115,16 @@ private object SetQuality()
                 {
                     episodeFile.Quality = resource.Quality;
                 }
+
+                if (resource.SceneName != null)
+                {
+                    episodeFile.SceneName = resource.SceneName;
```
markus101, line 121: `SceneName` isn't a freeform field, this will need to be validated to ensure it conforms with it.

How it's calculated during import is here: https://github.com/Sonarr/Sonarr/blob/develop/src/NzbDrone.Core/MediaFiles/EpisodeImport/SceneNameCalculator.cs

There is also `OriginalFilePath` which is the job folder (if present) and the filename.

markus101 (changes requested):
I think we need validation around `SceneName`, but also want to know what @Taloth @Qstick and @ta264 think.

src/Sonarr.Api.V3/EpisodeFiles/EpisodeFileModule.cs
mdhiggins: The `isScenePath` method https://github.com/Sonarr/Sonarr/blob/d6e5d9c424b39680d169bced723b3615043b0f97/src/NzbDrone.Core/Parser/SceneChecker.cs#L7 appears to be designed to determine if a file name is also a `SceneName` during import since it may be ambiguous. Not sure if this check needs to be performed in a situation where it was explicitly being set by the API in a less ambiguous situation

markus101: It's not the for the ambiguity, it's to determine that it is actually a scene name based on those rules. Unless @Taloth disagrees with this being validated I think it makes sense that it conforms with those checks.

Qstick:
Id agree with validation on Scene name given what’s it’s used for. Also to @TheCatLady point it would be nice if we re-analyze on size change

TheCatLady:
> Also to @TheCatLady point it would be nice if we re-analyze on size change

@Qstick that would be amazing. But allowing setting the scene name via API would be nice since #3366 has been open for a long time, and this would help greatly in the interim. ❤️

mdhiggins:
Updated to include isSceneTitle validation on both Radarr/Sonarr pull requests

markus101 (approved):
Thanks!

PR #4478, opened 2021-05-06
Fix scroll restore

ta264:
#### Database Migration
NO

#### Description
Pressing 'back' from a series page to go back to the index restores the scroll position.

Updates the `Link` component so that clicking on the sonarr icon top left will also restore the scroll position.

#### Todos
- [x] Tests

### DIFF
--- a/frontend/src/Components/Page/Header/PageHeader.js
+++ b/frontend/src/Components/Page/Header/PageHeader.js
@@ -53,7 +53,10 @@ class PageHeader extends Component {
         <div className={styles.logoContainer}>
           <Link
             className={styles.logoLink}
-            to={'/'}
+            to={{
+              pathname: '/',
+              state: { restoreScrollPosition: true }
+            }}
           >
             <img
               className={styles.logo}
--- a/frontend/src/Components/Table/VirtualTable.js
+++ b/frontend/src/Components/Table/VirtualTable.js
@@ -39,7 +39,8 @@ class VirtualTable extends Component {
     super(props, context);
 
     this.state = {
-      width: 0
+      width: 0,
+      scrollRestored: false
     };
 
     this._grid = null;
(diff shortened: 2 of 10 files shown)


frontend/src/Components/Page/Header/PageHeader.js
```diff
@@ -53,7 +53,10 @@ class PageHeader extends Component {
         <div className={styles.logoContainer}>
           <Link
             className={styles.logoLink}
-            to={'/'}
+            to={{
```
markus101, line 56: I don't think returning to Home via a link should restore scroll position, back makes sense because you're returning, but on fresh nav you're going somewhere new.

frontend/src/Components/Table/VirtualTable.js
```diff
@@ -141,11 +153,11 @@ class VirtualTable extends Component {
                     width={width}
                     height={height}
                     headerHeight={height - headerHeight}
-                    rowHeight={rowHeight}
+                    rowHeight={ROW_HEIGHT}
```
markus101, line 156: This will break taller rows with the banner won't it?

frontend/src/Components/Table/VirtualTable.js
```diff
@@ -83,11 +101,10 @@ class VirtualTable extends Component {
       className,
       items,
       scroller,
+      scrollTop: ignored,
```
markus101, line 104: Instead of this we should pass `{...otherProps}` at the top instead of the bottom so other props don't override explicitly set props.

markus101 (changes requested):
Haven't tested, but I don't think clicking the Sonarr icon should restore scroll position.

markus101:
Fixed in 99843d28768f96275a20d4afb1526e15ac0e24fa

PR #4491, opened 2021-05-16
Send signalr message for episode monitored flag changes

Taloth:
#### Database Migration
NO

#### Description
Changing monitored flags for episodes don't result in a signalr message.
This means that second browser tabs won't update, but also third-party apps using signalr won't get notified (Bazarr specifically).

note that I could do this in EpisodeModule handling the SetMonitored api calls and send broadcasts. However this means we wouldn't send updates for monitored flag changes triggered internally.

Some concerns:

- Added `GetResourceByIdForBroadcast` since I didn't want the series to be included in the notification. I have to check if that causes problems in the UI.
- It sends out 4 notifications for the calendar, episode, cutoff, missing endpoints. Which is a bit excessive.
- I do wonder what other things lack notifications to the UI.

### DIFF
--- a/src/NzbDrone.Core/Tv/EpisodeRepository.cs
+++ b/src/NzbDrone.Core/Tv/EpisodeRepository.cs
@@ -30,7 +30,8 @@ namespace NzbDrone.Core.Tv
         void SetMonitoredFlat(Episode episode, bool monitored);
         void SetMonitoredBySeason(int seriesId, int seasonNumber, bool monitored);
         void SetMonitored(IEnumerable<int> ids, bool monitored);
-        void SetFileId(int episodeId, int fileId);
+        void SetFileId(Episode episode, int fileId);
+        void ClearFileId(Episode episode, bool unmonitor);
     }
 
     public class EpisodeRepository : BasicRepository<Episode>, IEpisodeRepository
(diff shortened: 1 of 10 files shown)


src/NzbDrone.Core/Tv/EpisodeRepository.cs
```diff
@@ -173,25 +175,44 @@ public void SetMonitoredBySeason(int seriesId, int seasonNumber, bool monitored)
...
+            mapper.AddParameter("seriesId", seriesId);
+            mapper.AddParameter("seasonNumber", seasonNumber);
+            mapper.AddParameter("monitored", monitored);
+
+            var sqlUpdate = $"UPDATE Episodes SET Monitored = @monitored WHERE {sqlWhere}";
 
-            mapper.ExecuteNonQuery(sql);
+            mapper.ExecuteNonQuery(sqlUpdate);
+
+            episodes.ForEach(ModelUpdated);
```
markus101, line 193: Could avoid looping 2x and set monitored + send model updated here, but not that big of a concern.

markus101 (approved):
Just the one comment, I think the trade-offs are reasonable to achieve this functionality at least for now.

src/NzbDrone.Core/Tv/EpisodeRepository.cs
Taloth: I can also just add ModelUpdate(id) coz it's gonna call GetResourceById anyway.

Taloth:
Hmm, there's actually another problem. If I toggle a season monitored flag. It steps through the episodes surprisingly slowly.

PR #4495, opened 2021-05-18
Flood: explicitly cast DateFinished long? to long

jesec:
This is still somehow required even with `properties.DateFinished != null`. 

Fixes:

Sonarr\src\NzbDrone.Core\Download\Clients\Flood\Flood.cs(166,90,166,113):

error CS1503: Argument 1: cannot convert from 'long?' to 'long'


jesec:
cc: @Taloth

PR #4511, opened 2021-05-25
New: Added Running Years into the shows details

RobinDadswell:
#### Database Migration
NO

#### Description
Shows what years the show ran between, this is useful when some shows have remakes etc. as you can easily see the start and finish dates when looking at your shows. 

Show that has ended: 
image

Show that is continuing: 
image

Show that is upcoming: 
image

### DIFF
--- a/frontend/src/Components/HeartRating.css
+++ b/frontend/src/Components/HeartRating.css
@@ -2,3 +2,7 @@
   margin-right: 5px;
   color: $themeRed;
 }
+
+.rating {
+  margin-right: 15px;
+}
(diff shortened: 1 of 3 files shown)


thezoggy:
i would say just show just the YYYY instead of YYYY- if its really just the single year. the 'status' of continuing/upcoming is being implied which to some may just look like a bug and not understand the context of -  let alone the status may not be correct

RobinDadswell:
> i would say just show just the YYYY instead of YYYY- if its really just the single year. the 'status' of continuing/upcoming is being implied which to some may just look like a bug and not understand the context of - let alone the status may not be correct

image
I think there is an accepted standard to have a hyphen after the year for ranges that aren't finished

thezoggy:
alright you convinced me :P

PR #4525, opened 2021-06-04
Fixed: Parsing of some anime releases with year in title

DevYukine:
#### Database Migration
NO

#### Description
This is intended to fix an edge case which The Xem can't handle alone, the year after the episode breaks all current RegEx when a Seasonal Title mapping is used. It would just drop the Seasonal Title and use the base title which makes it map to a wrong Season.

Also i'm unsure if this is how you would do it assuming TheXem would map the Title `Anime Title the Final` to a specific season of `Anime Title`, would i still use absolute episodes there and then xem change that to seasonal episodes? if this is the wrong implementation please enlighten me how i would add a seasonal episode without season in the name cause this is the only way i got the unit test to work so far

#### Todos
- [x] Tests

### DIFF
--- a/src/NzbDrone.Core.Test/ParserTests/AbsoluteEpisodeNumberParserFixture.cs
+++ b/src/NzbDrone.Core.Test/ParserTests/AbsoluteEpisodeNumberParserFixture.cs
@@ -94,7 +94,7 @@ namespace NzbDrone.Core.Test.ParserTests
         [TestCase("Abc x Abc (2011) - 141 - Magician [KaiDubs] [1080p]", "Abc x Abc (2011)", 141, 0, 0)]
         [TestCase("Abc Abc 484 VOSTFR par Abc-Abc (1280*720) - version MQ", "Abc Abc", 484, 0, 0)]
         [TestCase("Abc - Abc Abc Abc - 107 VOSTFR par Fansub-Miracle Sharingan (1920x1080) - HQ_Draft", "Abc - Abc Abc Abc", 107, 0, 0)]
-        [TestCase("Abc Abc Abc Abc Episode 10 VOSTFR (1920x1080) Miracle Sharingan Fansub.MKV - Team - (À suivre)", "Abc Abc Abc Abc", 10, 0, 0)]
+        [TestCase("Abc Abc Abc Abc Episode 10 VOSTFR (1920x1080) Miracle Sharingan Fansub.MKV - Team - (ï¿½ suivre)", "Abc Abc Abc Abc", 10, 0, 0)]
         [TestCase("[Glenn] Series! 3 - 11 (1080p AAC)[C34B2B3B].mkv", "Series! 3", 11, 0, 0)]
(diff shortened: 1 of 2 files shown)


src/NzbDrone.Core.Test/ParserTests/AbsoluteEpisodeNumberParserFixture.cs
```diff
@@ -94,7 +94,7 @@ public class AbsoluteEpisodeNumberParserFixture : CoreTest
         [TestCase("Abc x Abc (2011) - 141 - Magician [KaiDubs] [1080p]", "Abc x Abc (2011)", 141, 0, 0)]
         [TestCase("Abc Abc 484 VOSTFR par Abc-Abc (1280*720) - version MQ", "Abc Abc", 484, 0, 0)]
         [TestCase("Abc - Abc Abc Abc - 107 VOSTFR par Fansub-Miracle Sharingan (1920x1080) - HQ_Draft", "Abc - Abc Abc Abc", 107, 0, 0)]
-        [TestCase("Abc Abc Abc Abc Episode 10 VOSTFR (1920x1080) Miracle Sharingan Fansub.MKV - Team - (� suivre)", "Abc Abc Abc Abc", 10, 0, 0)]
+        [TestCase("Abc Abc Abc Abc Episode 10 VOSTFR (1920x1080) Miracle Sharingan Fansub.MKV - Team - (À suivre)", "Abc Abc Abc Abc", 10, 0, 0)]
```
markus101, line 97: Looks like this line can be reverted

src/NzbDrone.Core.Test/ParserTests/AbsoluteEpisodeNumberParserFixture.cs
```diff
@@ -106,7 +106,7 @@ public class AbsoluteEpisodeNumberParserFixture : CoreTest
         [TestCase("[SubsPlease] Anime Title 300-nen, With Even More Title - 01 (1080p) [8DE44442]", "Anime Title 300-nen, With Even More Title", 1, 0, 0)]
         [TestCase("[Chihiro] Anime Title 300-nen, With Even More Title 02 [720p Hi10P AAC][031FA533]", "Anime Title 300-nen, With Even More Title", 2, 0, 0)]
         [TestCase("[BakeSubs] 86 - 01 [1080p][D40A9E55].mkv", "86", 1, 0, 0)]
-        //[TestCase("", "", 0, 0, 0)]
```
markus101, line 109: Please leave the commented out test case as well.

markus101 (changes requested):

src/NzbDrone.Core.Test/ParserTests/AbsoluteEpisodeNumberParserFixture.cs
DevYukine: Sadly i don't even know why this was changed in the first place, looks like my IDE doesn't like the file encoding in there, does sonarr not use UTF8 by any chance? (i copy pasted the code over again and i don't see any changes in my git so i assume it just auto converts it in my IDE)

markus101:
Landed in 14b551b02749ff03f64e12b11a99c235f7212604

PR #4535, opened 2021-06-13
New: Add Size column to Activity: Queue

### LINKED ISSUE #4527: Activity Queue: Size Column
**Describe the solution you'd like**
It would be nice to be able to see the size of each download in the queue as a column. It should be the size of the torrent (and whatever usenet calls their downloads) downloading which should be clear somehow through wording ("Download Size"?), not the size of the release as that can't always be gotten for RAR releases.

**Describe alternatives you've considered**
Manually looking in qBittorrent/clients which can be slow if you have 700 torrents seeding and UI performance isn't great.

napei:
#### Database Migration
NO

#### Description
Adds a `Size` column to the activity queue

* Fixes #4527

### DIFF
--- a/frontend/src/Activity/Queue/QueueRow.js
+++ b/frontend/src/Activity/Queue/QueueRow.js
@@ -19,6 +19,7 @@ import QueueStatusCell from './QueueStatusCell';
 import TimeleftCell from './TimeleftCell';
 import RemoveQueueItemModal from './RemoveQueueItemModal';
 import styles from './QueueRow.css';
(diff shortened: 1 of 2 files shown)


frontend/src/Activity/Queue/QueueRow.js
```diff
@@ -150,6 +151,12 @@ class QueueRow extends Component {
               );
             }
 
+            if (name === 'size') {
```
markus101, line 154: Please add the new column after the release title column.

markus101 (changes requested):

Issue #4562, opened 2021-06-30
DownloadStationTaskProxyV2 issues | Downloads cannot be added to Download Station


Prime-Omega:
**Description**

Potentially related to [#4388](https://github.com/Sonarr/Sonarr/issues/4388). I upgraded my Synology from DSM 6 to DSM 7. Unfortunately that automatically caused my Download station to upgrade from version 3.8.16-3563 to 3.9.0-4600. Everything works up until the point where Sonarr has to forward the torrent to Download station. Also weirdly enough, Radarr is not affected by this and is still able to relay torrents to download station.

**To Reproduce**
1. Go to a show & look for an episode
2. Sonarr returns a bunch of torrents
3. Download a torrent
4. Sonarr returns Reason: 120 - Unknown error

**Platform Information**

Syno version > DSM 7.0-41890
DS version > 3.9.0-4600
Docker version > 20.10.3-1233
Sonarr version > dev 3.0.6.1272

**Trace Logs**
[sonarr.trace.txt](https://github.com/Sonarr/Sonarr/files/6743988/sonarr.trace.txt)

### **EDIT: You can manually fix this by adding a destination folder in Sonarr under settings > download clients > download station config.**

bakerboy448:
Diagnosed here https://github.com/Radarr/Radarr/pull/6294#issuecomment-871640833

@twentynine78 will review and work on it when he can.

markus101:
Based on that thread it sounds like Sonarr and Radarr are configured differently, which would explain "Radarr is not affected by this" can you confirm?

Prime-Omega:
Running Radarr in the same environment as mentioned above by the way, nightly ver 4.0.0.5120.

To0wnn:
I have the exact same issue. Radarr is working but Sonarr is not. Reinstalled DS and Sonarr but the same result somehow..

To0wnn:
Difference between Radarr uses DownloadStationTaskProxy and Sonarr DownloadStationTaskProxy**V2** do not know is this is the issue or how you can switchback

bakerboy448:
So now you're saying the fix that was merged for DS now breaks DS on only Sonarr yet the same fix that is merged into Radarr Nightly works fine

that makes no sense at all.

To0wnn:
Maybe i do not understand then indeed. With version nightly-3.0.5.1146 of sonarr it is working. I use it in a docker. DS i use torrents

bakerboy448:
[...]

PR #4567, opened 2021-07-05
Fixed: issue with DownloadStationTaskProxyV2 if no destination folder…

twentynine78:
… specified in settings

#### Database Migration
NO

#### Description
Use the new DownloadStation2 API to retrieve the default destination folder if it is not configured in the download client settings, as the setting is now mandatory with the new API.

#### Todos
- [X] Tests

#4562


markus101:
Was testing this and ran into an issue where `DownloadClientRootFolderCheck` (health check) was failing because the download path was not known if both category and directory were blank. Instead of this separate lookup I've implemented a fix in 4bf3ab1511b4ea25642476bf9df13f91b6f73d76 that health check works correctly as well.

PR #4572, opened 2021-07-08
New: Renamed Blacklist to Blocklist

RobinDadswell:
#### Database Migration
YES - 160

#### Description
As part of the move to a more inclusive society, words like whitelist/blacklist are being moved away from in favour of the likes of allow list and blocklist. This replaces any references to the phrase blacklist with blocklist in the following areas: 
- Frontend User Visible
- Frontend Developer Visible
- Backend User Visible (e.g. logs) 
- Backend Developer Visible (with the exception of previous database migrations) 
- SQLLite Database

Blocklist was chosen in part via the Google Style Guide which discusses this matter and in tandem with both the Sonarr and Radarr/Lidarr/Readarr teams within the Sonarr Discord collaboration channel (https://discord.com/channels/383686866005917708/542785748768915456/859534462839685160)

I have tested this on my development machine, it took a refresh with a clear cache however this then worked as it currently does. 

#### Todos
- [x] Tests
- [x] Wiki Updates - https://github.com/Servarr/Wiki/pull/161


Taloth:
Where's the backward compat api?

RobinDadswell:
It is a breaking change for 3rd parties using the API. I haven't put this to be backwards compatible.

Taloth:
Imo it has to be backward compat. Put in a BlacklistModule for both api versions. We don't break backward api compat unless absolutely necessary.

RobinDadswell:
I have added a BlacklistModule for both API versions - it is the only code that still has reference to blacklist and it is in name only (it references the Blocklist namespace otherwise) - I have left it in it's own folder but happy to move it into the other folder if that is what you guys want. 

I think the goal should be to remove the blacklist API endpoint in the future so that there is only one place to maintain blocklist API code if changes are to be made in the future

markus101:
> I think the goal should be to remove the blacklist API endpoint in the future so that there is only one place to maintain blocklist API code if changes are to be made in the future

It wouldn't be added to future API versions, but it also shouldn't need to be updated, if we add a new feature to blocklist we wouldn't need to backport that to blacklist, API consumers can update to use the new endpoint instead.

frontend/src/Store/Actions/Creators/createHandleActions.js
```diff
@@ -11,7 +11,7 @@ import {
   REMOVE_ITEM
 } from 'Store/Actions/baseActions';
 
-const blacklistedProperties = [
+const blocklistedProperties = [
```
markus101, line 14: I think this would be better as `omittedProperties`

src/NzbDrone.Api/Blacklist/BlacklistModule.cs
```diff
@@ -1,30 +1,31 @@
-﻿using NzbDrone.Core.Blacklisting;
```
markus101, line 1: Probably worth a comment at the top that it's deprecated.

src/Sonarr.Api.V3/Blacklist/BlacklistModule.cs
```diff
@@ -1,42 +1,43 @@
-﻿using NzbDrone.Core.Blacklisting;
```
markus101, line 1: Same here, probably worth a comment at the top that it's deprecated.

markus101 (changes requested):
Just a few small comments, otherwise it looks good.

src/NzbDrone.Api/Queue/QueueActionModule.cs
```diff
@@ -43,12 +43,12 @@ public class QueueActionModule : SonarrRestModule<QueueResource>
 
         private object Remove(int id)
         {
-            var blacklist = false;
-            var blacklistQuery = Request.Query.blacklist;
+            var blocklist = false;
```
markus101, line 46: For backwards compatibility this should handle `blacklist`

src/Sonarr.Api.V3/Queue/QueueActionModule.cs
```diff
@@ -81,9 +81,9 @@ private object Grab()
         private object Remove(int id)
         {
             var removeFromClient = Request.GetBooleanQueryParameter("removeFromClient", true);
-            var blacklist = Request.GetBooleanQueryParameter("blacklist");
+            var blocklist = Request.GetBooleanQueryParameter("blocklist");
```
markus101, line 84: For backwards compatibility this should handle `blacklist`

src/Sonarr.Api.V3/Queue/QueueActionModule.cs
```diff
@@ -96,14 +96,14 @@ private object Remove(int id)
         private object Remove()
         {
             var removeFromClient = Request.GetBooleanQueryParameter("removeFromClient", true);
-            var blacklist = Request.GetBooleanQueryParameter("blacklist");
+            var blocklist = Request.GetBooleanQueryParameter("blocklist");
```
markus101, line 99: Same here

markus101 (commented):
A couple more backwards compatibility comments.

src/NzbDrone.Core.Test/HealthCheck/Checks/DownloadClientRootFolderCheckFixture.cs
```diff
@@ -68,7 +68,7 @@ public void should_return_downloads_in_root_folder_if_downloading_to_root_folder
         {
             GivenRootFolder(_downloadRootPath);
 
-            Subject.Check().ShouldBeWarning(wikiFragment: "downloads_in_root_folder");
+            Subject.Check().ShouldBeWarning(wikiFragment: "downloads-in-root-folder");
```
markus101, line 71: We don't need to check the fragment, we only care that there is a warning as the result. I've removed from develop.

RobinDadswell: removed this commit and rebased as you've fixed in develop

PR #4606, opened 2021-08-03
New: Change Today color in calendar for better visibility

kevinwlau:
#### Database Migration
NO 

#### Description
I found it hard to see the highlighted `today` block in the calendar, so I adjusted the color for more contrast.
This is the smallest of changes, but it makes a big difference in visibility.

Let me know if there are any problems in regards to text contrast for accessibility.

### DIFF
--- a/frontend/src/Styles/Variables/colors.js
+++ b/frontend/src/Styles/Variables/colors.js
@@ -178,7 +178,7 @@ module.exports = {
   //
   // Calendar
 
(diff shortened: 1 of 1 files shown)


markus101:
Please post a screenshot as well.

kevinwlau:
Before:
image

After: 
image

kevinwlau:
Looked into the WCAG accessibility guidelines for color contrast, and found that `#c5c5c5` passes while still making the Today more prominent
image

Issue #4153, opened 2020-12-10
Search for anime episodes with absolute AND seasonal episode formats


reconman:
**Describe the problem**
Some anime releases like Attack on Titan have different subbing groups with different naming schemes. Some use `Shingeki no Kyojin 60`, some `Shingeki no Kyojin S04E01`, and others `Shingeki no Kyojin S04E01 - 60`.

https://github.com/Sonarr/Sonarr/blob/phantom-develop/src/NzbDrone.Core/IndexerSearch/NzbSearchService.cs#L84 disables the seasonal search when single episodes are searched for.

The second naming scheme will never be found in the indexers, only the ones using the absolute episode numbers.

**Describe any solutions you think might work**
I would suggest doing both a seasonal and non-seasonal search for that episode and combine the results. Of course, duplicates should be eliminated before returning the list. The third naming scheme above will be found by both search methods.

**Additional context**

reconman:
I debugged the source code a bit and guess I was wrong about how the code needs to be changed.

Most anime torrent sites don't have season and episode search support, which the "Standard" search relies on. So you would have to then put the SXXEXX string into the q= parameters instead of the absolute episode number.

thezoggy:
Anime uses abs. Standard uses SxxExx. So just change series type and it will also change how to search/what matches

reconman:
> Anime uses abs. Standard uses SxxExx. So just change series type and it will also change how to search/what matches

Except that Sonarr can't search for SxxExx on most anime indexers. It will simply skip them. Only animetosho via Jackett seems to work.

thezoggy:
if its in standard naming format then it usually means your talking about scene releases rather than p2p stuff. and for standard naming just let it search normal indexers (rather than trying to force it on anime only). example of this is like pokemon.  works just fine as 'standard' (dont treat it as anime)

reconman:
Afaik "scene" IS p2p for anime.

Name me one popular non-anime indexer which offers `Tatoeba Last Dungeon Mae no Mura no Shounen ga Joban no Machi de Kurasu Youna Monogatari`. (I picked a random show from this season)

thezoggy:
I'd be shocked if there was any scene releases for a non mainstream anime. yeah nothing in predb for "Suppose a Kid From the Last Dungeon Boonies Moved to a Starter Town"  but anyways, again if you want sonarr to use standard SxxExx then use standard. if you want it to do absolute then use anime.

[...]

PR #4728, opened 2021-10-31
Add optionnal standard numbering anime search

Bakeneko:
#### Database Migration
NO

#### Description
Add an indexer setting to enable additionnal standard numbering format searches for anime.

#### Todos
- [x] Tests

Related to #4153

### DIFF
--- a/src/NzbDrone.Core/Indexers/Fanzub/FanzubSettings.cs
+++ b/src/NzbDrone.Core/Indexers/Fanzub/FanzubSettings.cs
@@ -25,6 +25,9 @@ namespace NzbDrone.Core.Indexers.Fanzub
         [FieldDefinition(0, Label = "Rss URL", HelpText = "Enter to URL to an Fanzub compatible RSS feed")]
         public string BaseUrl { get; set; }
 
+        [FieldDefinition(1, Label = "Anime Standard Format Search", Type = FieldType.Checkbox, HelpText = "Also search for anime using the standard numbering")]
+        public bool AnimeStandardFormatSearch { get; set; }
+
         public NzbDroneValidationResult Validate()
         {
             return new NzbDroneValidationResult(Validator.Validate(this));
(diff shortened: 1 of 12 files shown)


src/NzbDrone.Core/Indexers/Fanzub/FanzubSettings.cs
```diff
@@ -25,6 +25,9 @@ public FanzubSettings()
         [FieldDefinition(0, Label = "Rss URL", HelpText = "Enter to URL to an Fanzub compatible RSS feed")]
         public string BaseUrl { get; set; }
 
+        [FieldDefinition(1, Label = "Anime seasonal search", Type = FieldType.Checkbox, HelpText = "Add seasonal format search for Anime")]
```
markus101, line 28: This format is referred to as `Standard` everywhere else in Sonarr, it should be referred to that way in the setting as well, Season/Seasonal is far too easily confused with season pack searches.

Maybe: `Anime Standard Format Search` with the help text of `Also search for anime using the standard numbering format`

markus101 (changes requested):
Made a suggestion to improve the naming, but season anime searches should also search for standard season packs if that option is enabled.

Bakeneko:
Added @markus101 suggestions and tests for request generators.

markus101:
Made a small change and landed this in cee17483d97c1414b05b9f19729d612ef43934c2

reloxx13:
Heya,

when this checkbox is checked, shouldn't it also use t=TVSEARCH and not t=SEARCH for animes?

Standard uses tvsearch, while anime uses search, even when `Anime Standard Format Search` is checked.
When using t=SEARCH with anime as type, the results are ignoring the searched season and episode and returning alot of unwanted/unasked episodes. while changing t=TVSEARCH (i did it manual with the previos SEARCH api call) is returning the correct episodes.

if i set the anime serie to standard and add the anime category to the search categories of standard, im getting the correct episodes returned.

Issue #4538, opened 2021-06-14
German releases detected as English instead of German


virusperfect:
**Describe the bug**
The releases of one of my German trackers don't get detected as German but as English instead. I think the problem is that they use only "Ger" in the name of their releases instead of "German". Releases with "German" are detected correctly.

**To Reproduce**
Releases that gets detected as English:
"Attack.on.Titan.S04E16.Ger.Sub.AAC.1080p.WebRip.x264-Mindus"
Releases that gets detected as German:
"Attack.on.Titan.S01.German.DL.DTS.1080p.BluRay.x265-ABJ"

**Expected behavior**
I'm not sure if this even unexpected behavior. But I would like releases with "Ger" to also be detected as German and not English.

**Screenshots**

**Platform Information (please complete the following information):**
 - OS: Ubuntu 20.04
 - Docker: Yes
 - .net Framework (Windows) or mono (macOS/Linux) (System -> Status): Mono 6.12.0.122
 - Browser and Version (Only needed for UI issues):
 - Sonarr Version: 3.0.6.1196
 - Sonarr Branch: main

**Trace Logs**
Turn on Trace logs under Settings -> General and wait for the bug to occur again.
**Upload the full log file here (or another site (e.g. pastebin) and link it). Issues will be closed, if they do not include this!**
[sonarr.trace.txt](https://github.com/Sonarr/Sonarr/files/6649209/sonarr.trace.txt)

thezoggy:
ger.sub would imply that it only has german subs.. so still english release

virusperfect:
Sorry that was an unlucky example. This also gets detected as being English even though it has German audio:
"To.Your.Eternity.S01E03.Ger.Dub.AAC.1080p.WebDL.x264-TKP21"

thezoggy:
usually the language like "French" / "Spanish" / "German" is used. the three letter abbrv is not standard which is why its never been supported. i assume this is from some torrent site by the name of the p2p releases?

virusperfect:
Yes this is from a private tracker. If you don't plan to support three letter abbreviations I guess can close the issue.

markus101:
If it's not something we're going to support we'll close it.

reloxx13:
there is another somehow related issue. 
after sonarr renames the files, there is no way to get the origin language again if you have to reimport the files. you can add `{MediaInfo AudioLanguages}` to the file name which i "DE" or PL", but sonarr is not able to recognize it as it is not looking for iso codes.

[...]

Issue #4240, opened 2021-01-10
Automatically set language based on audio language


viggy96:
**Describe the problem**
Anime is often tagged with the incorrect language in various indexers like Nyaa. They are often identified as being in English when in fact they are not. 

**Describe any solutions you think might work**
Sonarr already analyses the video files after importing and recognises the language of the audio tracks. Sonarr could use this information to properly set the language property for each episode.

thezoggy:
and follow scene rules, if two audio 'dual' if more than two 'multi' tag

NuGundam:
Yes I like this to happen to so then we use the Language Profile as it is attended and users can get the Japanese with subs and when available it will upgrade to English which is most of the time dual audio.

geropan:
Sorry to hijack but this could also solve and issue that I am seeing:

I have a language profile containing two languages: Spanish and English (in this priority order). I have allowed upgrades and requested to the cutoff to Spanish.
![image](https://user-images.githubusercontent.com/57835069/105513094-cd734e80-5cc9-11eb-9a54-07ac6c6fc55a.png)

When an release is grabbed including Spanish language, upon import of the **downloaded file it is labelled as English (even though the title is in Spanish)**. This causes for the same released to be grabbed again in a periodic loop (due to my Language upgrade and cutoff settings):
![image](https://user-images.githubusercontent.com/57835069/105513530-568a8580-5cca-11eb-8c84-5441d72c3a74.png)

As the OP stated, **Sonarr already knows that the imported release has Spanish audio language**, but this information is not used in this context:
![image](https://user-images.githubusercontent.com/57835069/105513698-8a65ab00-5cca-11eb-98e4-1c1e12ad434d.png)

At the moment the only way to stop this is to manually unmonitor the episode once downloaded, although this defeats the purpose of automatic upgrades and so on (e.g. in a use case with a profile with 3 languages).

viggy96:
@geropan Exactly the issue I'm seeing with my anime. I created a language profile with Japanese and English, preferring Japanese. Sonarr often downloads episodes again to get the "preferred" language, even though the existing file _already_ has Japanese audio.

geropan:
I know that Radarr implementation is different (it relies on user defined custom format for upgrades), but I suspect it might be using MediaInfo to determine languages.

In the example below there are no hints on the filename to determine the languages, but they are still being identified correctly:
[...]

PR #4737, opened 2021-11-05
Language Augmenter

NicolasFerec:
#### Database Migration
NO

#### Description
Implementation of a Language Augmenter based on the Radarr one. 
Allow to use MediaInfo to detect episode language.

#### Todos
- [ ] Tests
- [ ] Wiki Updates

#4240 : Solves this issue with the exact proposed solution.
~~#4538 : If files contains correct audio track language tagging, it solve the issue.~~


bakerboy448:
You're going to need and want test cases for this as well

it also looks like the existing tests failed

NicolasFerec:
Hi @bakerboy448, thank you for your quick reply.

> You're going to need and want test cases for this as well
I don't know how to make those :/

> it also looks like the existing tests failed
Yes I saw that, again, no idea what to do about it

> assuming this is even something that @markus101 and @Taloth are game for
I'm not sure, I spoke to Markus on Discord. For him it's not a complete solution. For me it's better than nothing. 
At least it solves the #4240 issue.
If it really isn't something you want, I let you close this PR I guess.

> also commits should be meaningful, not just file updates ideally
What do you mean by meaningful ? Isn't all commits file updates ?

NicolasFerec:
Ok I think I found something about the test not passing, but it's weird, it's nothing that I've touch in my commits.
In the NzbDrone.Core.Test.NotificationTests.ProwlProviderTest there is two string to 'NzbDrone.Core.Notifications.Prowl.ProwlSettings' conversions. 

My guess is to modify this file with this code below :
```cs
using NUnit.Framework;
using NzbDrone.Core.Notifications.Prowl;
using NzbDrone.Core.Test.Framework;
using NzbDrone.Test.Common;
using NzbDrone.Test.Common.Categories;

namespace NzbDrone.Core.Test.NotificationTests
{
    [Explicit]
    [ManualTest]
    [TestFixture]
    public class ProwlProviderTest : CoreTest<ProwlProxy>
    {
        private const string _apiKey = "66e9f688b512152eb2688f0486ae542c76e564a2";

        private const string _badApiKey = "1234567890abcdefghijklmnopqrstuvwxyz1234";

        private ProwlSettings _settings = new ProwlSettings { ApiKey = _apiKey };

        private ProwlSettings _badSettings = new ProwlSettings { ApiKey = _badApiKey };

        [Test]
        public void Verify_should_not_throw_for_a_valid_apiKey()
        {
            Subject.Test(_settings);
            ExceptionVerification.ExpectedWarns(0);
        }

        [Test]
        public void Verify_should_throw_for_an_invalid_apiKey()
        {
            _settings.ApiKey = _badApiKey;

            Assert.Throws<ProwlException>(() => Subject.Test(_settings));
            
            ExceptionVerification.ExpectedWarns(1);
        }

        [Test]
        public void SendNotification_should_not_throw_for_a_valid_apiKey()
        {
            Subject.SendNotification("NzbDrone Test", "This is a test message from NzbDrone", _settings);
            ExceptionVerification.ExpectedWarns(0);
        }

        [Test]
        public void SendNotification_should_log_a_warning_for_an_invalid_apiKey()
        {
            Subject.SendNotification("NzbDrone Test", "This is a test message from NzbDrone", _badSettings);

            ExceptionVerification.ExpectedWarns(1);
        }
    }
}```

File: src/NzbDrone.Core/MediaFiles/EpisodeImport/Aggregation/Aggregators/AggregateLanguage.cs [hunk 1]
```diff
@@ -1,71 +1,54 @@
...
-
-            var language = languages.FirstOrDefault(l => l != Language.English) ?? Language.English;
-
-            _logger.Debug("Using language: {0}", language);
-
-            localEpisode.Language = language;
+            foreach (var augmentLanguage in _augmentLanguages)
+            {
+                var augmentedLanguage = augmentLanguage.AugmentLanguage(localEpisode, downloadClientItem);
+                if (augmentedLanguage == null)
```
markus101, line 30: New line before `if`

src/NzbDrone.Core/MediaFiles/EpisodeImport/Aggregation/Aggregators/AggregateLanguage.cs
```diff
@@ -1,71 +1,54 @@
...
-            return localEpisode;
-        }
+                _logger.Trace("Considering Languages {0} ({1}) from {2}", string.Join(", ", augmentedLanguage.Languages ?? new List<Language>()), augmentedLanguage.Confidence, augmentLanguage.Name);
 
-        private Language GetLanguage(ParsedEpisodeInfo parsedEpisodeInfo, List<Episode> episodes)
-        {
-            if (parsedEpisodeInfo == null)
-            {
-                return Language.English;
+                if (augmentedLanguage?.Languages != null && augmentedLanguage.Languages.Count > 0 && !(augmentedLanguage.Languages.Count == 1 && augmentedLanguage.Languages.Contains(Language.Unknown)))
```
markus101, line 37: Here you're using: `augmentedLanguage?.Languages`, but previously in the logging line `augmentedLanguage.Languages ?? new List<Language>()` is used`, if `augmentedLanguage` wasn't previously defined it fails logging, so `augmentedLanguage?.Language` here doesn't matter and it's already continuing if `augmentedLanguage == null`. The `Languages` property should never be null (set a default value of an empty list in the constructor of that class).

src/NzbDrone.Core/MediaFiles/EpisodeImport/Aggregation/Aggregators/AggregateLanguage.cs
```diff
@@ -1,71 +1,54 @@
...
+                    languagesConfidence = augmentedLanguage.Confidence;
+                }
             }
 
-            var normalizedReleaseTitle = Parser.Parser.NormalizeEpisodeTitle(parsedEpisodeInfo.ReleaseTitle);
 
-            foreach (var episode in episodes)
-            {
-                var episodeTitleLanguage = LanguageParser.ParseLanguage(episode.Title, false);
+            var language = languages.FirstOrDefault(l => l != Language.English) ?? Language.English;
```
markus101, line 45: This is going to cause issues some of the time. If you have a release that has 3 language tracks in the following order:

1. English
2. French
3. German

you're going to have issues because the language returned here is always going to be French, if you want German or English, it doesn't matter because this only sees French.

If the first non-Enlish track is `Unknown` then it'll come back as unknown.

Ideally this logic needs to be expanded so multiple audio tracks can be stored or in the mean time the first audio track should be used instead of some arbitrary one.

src/NzbDrone.Core/MediaFiles/EpisodeImport/Aggregation/Aggregators/Augmenters/Language/AugmentLanguageFromFileName.cs
```diff
@@ -0,0 +1,24 @@
...
+using NzbDrone.Core.Parser.Model;
+
+namespace NzbDrone.Core.MediaFiles.EpisodeImport.Aggregation.Aggregators.Augmenters.Language
+{
+    public class AugmentLanguageFromFileName : IAugmentLanguage
+    {
+        public int Order => 1;
+        public string Name => "FileName";
+
+        public AugmentLanguageResult AugmentLanguage(LocalEpisode localMovie, DownloadClientItem downloadClientItem)
```
markus101, line 12: `localEpisode`

src/NzbDrone.Core/MediaFiles/EpisodeImport/Aggregation/Aggregators/Augmenters/Language/AugmentLanguageFromFileName.cs
```diff
@@ -0,0 +1,24 @@
...
+        {
+            var languages = new List<Languages.Language> { localMovie.FileEpisodeInfo?.Language };
+
+            if (languages == null)
```
markus101, line 16: If `localMovie.FileEpisodeInfo?.Language` is null this will create an array with a null value, instead:

```
if (localMovie.FileEpisodeInfo?.Language == null)
{
  return null;
}

return new AugmentLanguageResult(localMovie.FileEpisodeInfo?.Language, Confidence.Filename);
```

Is better.

src/NzbDrone.Core/MediaFiles/EpisodeImport/Aggregation/Aggregators/Augmenters/Language/AugmentLanguageFromFolder.cs
```diff
@@ -0,0 +1,24 @@
...
+    public class AugmentLanguageFromFolder : IAugmentLanguage
+    {
+        public int Order => 2;
+        public string Name => "FolderName";
+
+        public AugmentLanguageResult AugmentLanguage(LocalEpisode localEpisode, DownloadClientItem downloadClientItem)
+        {
+            var languages = new List<Languages.Language> { localEpisode.FolderEpisodeInfo?.Language };
+
+            if (languages == null)
```
markus101, line 16: Same as previous comment.

markus101 (changes requested):
As discussed previously on Discord this needs to address the multiple audio track issue in a different way.

This PR does not address #4538, that issue is purely a parsing issue.

src/NzbDrone.Core/MediaFiles/EpisodeImport/Aggregation/Aggregators/AggregateLanguage.cs
NicolasFerec: Thank you for the review! For this particular one, although I can see your point, I didn't make any modification to that line. The behaviour you described already is what's happening on the released app.

As I mentioned in a comment above:

>  If you check the AggragateLanguage.cs at the line 48 (line 35 in the original file), returning the first non-english language we found was already what was done. I just "upgraded" the sources from where we fetch possible languages. I didn't change the decision about the selected language at the end.

So with this information, what do you want me to do? I know that ideally the real multi-language capability would have been the best but I honestly don't have that kind of skill. Does that mean this PR will never be accepted for it is? Just to know whether I should try to fix all the other point or should I cancel this PR right now.

markus101:
This PR will need to be rebased on develop and have the merge commits removed.

Qstick:
This should be in v4 (develop) already

PR #4278, opened 2021-01-23
Detect Dolby Vision as HDR and MediaInfo Update

Matty666:
#### Database Migration
NO

#### Description
This PR is to address issue #4276 where Dolby Vision HDR was not being picked up by sonarr in the media info formatting for VideoDynamicRange

I updated mediainfo library windows and mac to 20.09 and added `VideoHdrFormat` and `VideoHdrFormatCompatibility` properties to the `MediaInfoModel`.

Adjusted `MediaInfoFormatter.FormatVideoDynamicRange` to return HDR if the VideoHdrFormat field is not empty.

This should handle older media info schemas by defaulting the properties to empty, and the formatter will just assume SDR

#### Issues Fixed or Closed by this PR
#4276

PR #4765, opened 2021-11-27
New: Add {MediaInfo VideoDynamicRangeType} token for renaming

Alanoll:
#### Database Migration
NO

#### Description
This is based off the work done by @Matty666 in #4278 and #2950

With the multiple hdr formats and some shows getting released as both DolbyVision and HDR10, Sonarr was renaming then all HDR. It is possible to use Preferred Words renaming to keep the extra metadata when it exists in the release name, but it's not always included (e.g. I believe Scene standards say it's optional if the HDR version was released first).

I added a new token instead of changing the logic of the original one. It adds a token that could be considered a dupe, but also means that existing users won't start seeing new stuff they're not expecting in their names.

The new token attempts to parse what type of HDR the media is based on the existing metadata pulls (so a rescan isn't needed and the model version doesn't change so no DB serializers need altering).

(if I'm off base on this, please comment and reject :) )

#### Todos
- [X] Tests
- [ ] Wiki Updates

*


Alanoll:
Sorry about the double PR, but meant this to be a draft since I haven't completed the tests yet and was looking for comments before putting in some more time.

markus101 (approved):
Looks good, the second token was a good call.

Issue #4743, opened 2021-11-10
Jellyfin/Emby Missing Support On Delete Notifications


neo-neo1:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Delete notifications for Emby/Jellyfin do not exist. This is a cross-post from Radarr where @RobinDadswell [made a pull request](https://github.com/Radarr/Radarr/pull/6699) to solve issue https://github.com/Radarr/Radarr/issues/4181. It has been tested to function as expected. Please implement this on Sonarr as well.

**For the non-familiar: Why this functionality is crucial:** 
On Jellyfin/Emby if media library is connected via NFS/SMB/etc.. real time monitoring doesn't work. Solution involves Radarr to send notifications upon media imports & deletes to Jellyfin/Emby to update it's library. However, delete notifications don't function as the [checkboxes are greyed out](https://i.imgur.com/QIyXE3J.png) since this implementation was not added.

### Expected Behavior

Please see above.

### Environment

```markdown
- OS: Agnostic
- Sonarr: Agnostic
- Docker Install: Agnostic
- Using Reverse Proxy: Agnostic
- Browser: Agnostic
```

### What branch are you running?

Develop

### Anything else?

Thanks in advance.

PR #4767, opened 2021-11-29
Implement OnDelete notification for emby/jellyfin

pixxon:
#### Database Migration
YES ( auto enable delete updates )

#### Description
Like for plex, notify emby and jellyfin for delete operations.

#### Todos
- [ ] Tests
Tested it locally, found no existing tests for emby connection.
```
Emby.Server.Implementations.Session.SessionWebSocketListener: Watching 1 WebSockets.
Jellyfin.Server.Middleware.LegacyEmbyRouteRewriteMiddleware: Removing "/mediabrowser" from route.
Jellyfin.Api.Auth.CustomAuthenticationHandler: AuthenticationScheme: "CustomAuthentication" was successfully authenticated.
Emby.Server.Implementations.IO.LibraryMonitor: New file refresher created for "Y:\XXX"
Emby.Server.Implementations.IO.LibraryMonitor: "XXX" ("Y:\XXX") will be refreshed.
MediaBrowser.Controller.Entities.BaseItem: Removed item: Y:\XXX\Season 1\XXX - S01E10.mkv
```
- [ ] Wiki Updates
Found nothing relevant for it, so no update needed.

#4743

### DIFF
--- a/src/NzbDrone.Core/Datastore/Migration/163_add_on_delete_to_emby_notifications.cs
+++ b/src/NzbDrone.Core/Datastore/Migration/163_add_on_delete_to_emby_notifications.cs
@@ -1,0 +1,16 @@
+using FluentMigrator;
+using NzbDrone.Core.Datastore.Migration.Framework;
+
+namespace NzbDrone.Core.Datastore.Migration
+{
+    [Migration(163)]
+    public class add_on_delete_to_emby_notifications : NzbDroneMigrationBase
+    {
+        protected override void MainDbUpgrade()
+        {
+            Execute.Sql("UPDATE Notifications SET OnSeriesDelete = 1 WHERE Implementation = 'MediaBrowser'");
+            Execute.Sql("UPDATE Notifications SET OnEpisodeFileDelete = 1 WHERE Implementation = 'MediaBrowser'");
+            Execute.Sql("UPDATE Notifications SET OnEpisodeFileDeleteForUpgrade = 1 WHERE Implementation = 'MediaBrowser'");
+        }
+    }
+}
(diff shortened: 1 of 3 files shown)


src/NzbDrone.Core/Datastore/Migration/163_add_on_delete_to_emby_notifications.cs
```diff
@@ -0,0 +1,16 @@
...
+namespace NzbDrone.Core.Datastore.Migration
+{
+    [Migration(163)]
+    public class add_on_delete_to_emby_notifications : NzbDroneMigrationBase
+    {
+        protected override void MainDbUpgrade()
+        {
+            Execute.Sql("UPDATE Notifications SET OnSeriesDelete = 1 WHERE Implementation = 'MediaBrowser'");
+            Execute.Sql("UPDATE Notifications SET OnEpisodeFileDelete = 1 WHERE Implementation = 'MediaBrowser'");
+            Execute.Sql("UPDATE Notifications SET OnEpisodeFileDeleteForUpgrade = 1 WHERE Implementation = 'MediaBrowser'");
```
markus101, line 13: I think this is unnecessary if `On Upgrade` events are being sent, as this would cause multiple events to be sent back to back for the same series.

Though overall I think people can opt into this behaviour instead of forcing it.

markus101 (changes requested):
I don't think we need to enable on delete events for existing connections (we didn't in other cases either).

markus101:
Dropped the migration and committed in a27984c0328210bca4854d1a0926899c3f4361b9

neo-neo1:
Thanks for this @pixxon

Issue #4416, opened 2021-03-29
URL Base for Connections


pixxon:
**Describe the bug**

Most of the connections ( or at least the ones I use/checked ) only allow host to be specified. This limits the options especially when the services are running behind a reverse proxy under different baseurls.

**To Reproduce**
1. Go to 'Settings -> Connect'
2. Click on 'New Connection'
3. Select 'Emby'
4. Set Name to anything and Host to '192.168.0.129/emby'
5. Click 'Test'
4. See error --> "must be valid Host without http://"

**Expected behavior**
Services running under baseurl should be connectable by Sonarr.

**Screenshots**

![image](https://user-images.githubusercontent.com/12481929/112828787-76ef0b80-9090-11eb-9c1b-cf6accc2971c.png)

**Platform Information (please complete the following information):**
 - OS: Ubuntu 20.04 ( 3.0.4.1021-ls8 by linuxserver.io )
 - Docker: Yes
 - .net Framework (Windows) or mono (macOS/Linux) (System -> Status): Mono 5.20.1.34
 - Browser and Version (Only needed for UI issues): N/A
 - Sonarr Version: 3.0.4.1021
 - Sonarr Branch: phantom-develop

pixxon:
This is very similar to https://github.com/Sonarr/Sonarr/issues/1747

pixxon:
I assume either allow full url here: https://github.com/Sonarr/Sonarr/blob/phantom-develop/src/NzbDrone.Core/Notifications/MediaBrowser/MediaBrowserSettings.cs#L13

Or introduce a baseurl like in case of download clients.

PR #4769, opened 2021-11-29
Allow urlbase to be specified for mediaservers

pixxon:
#### Database Migration
NO ( Default value is empty for urlbase )

#### Description
Allows people to run emby/jellyfin/plex behind reverse proxy using subdirectory and still connect them to Sonarr.

#### Todos
- [ ] Tests
The old ones are running, I didn't see any meaningful test for urls, but can add if needed.
- [ ] Wiki Updates
Found no wiki for connection specific settings.

#4416

### DIFF
--- a/src/NzbDrone.Core/Notifications/MediaBrowser/MediaBrowserProxy.cs
+++ b/src/NzbDrone.Core/Notifications/MediaBrowser/MediaBrowserProxy.cs
@@ -66,10 +66,7 @@ namespace NzbDrone.Core.Notifications.Emby
 
         private HttpRequest BuildRequest(string path, MediaBrowserSettings settings)
         {
-            var scheme = settings.UseSsl ? "https" : "http";
-            var url = $@"{scheme}://{settings.Address}/mediabrowser";
-            
-            return new HttpRequestBuilder(url).Resource(path).Build();
+            return new HttpRequestBuilder($@"{settings.Address}/mediabrowser").Resource(path).Build();
         }
 
         private void CheckForError(HttpResponse response)
--- a/src/NzbDrone.Core/Notifications/MediaBrowser/MediaBrowserSettings.cs
+++ b/src/NzbDrone.Core/Notifications/MediaBrowser/MediaBrowserSettings.cs
@@ -3,6 +3,7 @@ using Newtonsoft.Json;
 using NzbDrone.Core.Annotations;
 using NzbDrone.Core.ThingiProvider;
 using NzbDrone.Core.Validation;
+using System;
 
 namespace NzbDrone.Core.Notifications.Emby
 {
@@ -12,6 +13,7 @@ namespace NzbDrone.Core.Notifications.Emby
         {
             RuleFor(c => c.Host).ValidHost();
             RuleFor(c => c.ApiKey).NotEmpty();
+            RuleFor(c => c.UrlBase).ActuallyValidUrlBase();
         }
     }
 
@@ -19,31 +21,29 @@ namespace NzbDrone.Core.Notifications.Emby
     {
         private static readonly MediaBrowserSettingsValidator Validator = new MediaBrowserSettingsValidator();
 
-        public MediaBrowserSettings()
-        {
-            Port = 8096;
-        }
-
         [FieldDefinition(0, Label = "Host")]
         public string Host { get; set; }
 
         [FieldDefinition(1, Label = "Port")]
-        public int Port { get; set; }
+        public int Port { get; set; } = 8096;
 
         [FieldDefinition(2, Label = "Use SSL", Type = FieldType.Checkbox, HelpText = "Connect to Emby over HTTPS instead of HTTP")]
         public bool UseSsl { get; set; }
 
-        [FieldDefinition(3, Label = "API Key", Privacy = PrivacyLevel.ApiKey)]
+        [FieldDefinition(3, Label = "Url Base", Type = FieldType.Textbox, Advanced = true, HelpText = "Adds a prefix to the Emby server url, see http://[host]:[port]/[urlBase]")]
+        public string UrlBase { get; set; } = "";
+
+        [FieldDefinition(4, Label = "API Key", Privacy = PrivacyLevel.ApiKey)]
         public string ApiKey { get; set; }
 
-        [FieldDefinition(4, Label = "Send Notifications", HelpText = "Have MediaBrowser send notfications to configured providers", Type = FieldType.Checkbox)]
+        [FieldDefinition(5, Label = "Send Notifications", HelpText = "Have MediaBrowser send notfications to configured providers", Type = FieldType.Checkbox)]
         public bool Notify { get; set; }
 
-        [FieldDefinition(5, Label = "Update Library", HelpText = "Update Library on Import & Rename?", Type = FieldType.Checkbox)]
+        [FieldDefinition(6, Label = "Update Library", HelpText = "Update Library on Import & Rename?", Type = FieldType.Checkbox)]
         public bool UpdateLibrary { get; set; }
 
         [JsonIgnore]
-        public string Address => $"{Host}:{Port}";
+        public string Address => $"{(UseSsl ? "https" : "http")}://{Host}:{Port}{(String.IsNullOrEmpty(UrlBase) ? "" : $"/{UrlBase}")}";
 
         public bool IsValid => !string.IsNullOrWhiteSpace(Host) && Port > 0;
 
--- a/src/NzbDrone.Core/Validation/RuleBuilderExtensions.cs
+++ b/src/NzbDrone.Core/Validation/RuleBuilderExtensions.cs
@@ -38,6 +38,10 @@ namespace NzbDrone.Core.Validation
         {
             return ruleBuilder.SetValidator(new RegularExpressionValidator(@"^(?!\/?https?://[-_a-z0-9.]+)", RegexOptions.IgnoreCase)).WithMessage($"Must be a valid URL path (ie: '{example}')");
         }
+        public static IRuleBuilderOptions<T, string> ActuallyValidUrlBase<T>(this IRuleBuilder<T, string> ruleBuilder, string example = "sonarr/base")
+        {
+            return ruleBuilder.SetValidator(new RegularExpressionValidator(@"([-_a-z0-9.]+(/[-_a-z0-9.]+)*)?", RegexOptions.IgnoreCase)).WithMessage($"Must be a valid URL path (ie: '{example}')");
+        }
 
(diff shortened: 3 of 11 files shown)


bakerboy448:
> Found no wiki for connection specific settings.

needs work and the settings listed out for each, but they'd go here

https://wiki.servarr.com/sonarr/supported#notifications

format would be

```markdown
-  Emby {#embry}
  - {Setting Label} - {Description}
  - (Advanced Option) {Setting Label} - {Description}

**Comment by pixxon** [comment]
> needs work

Can I somehow get access to the build logs? It compiled for me locally and the tests I should have impacted ran fine as well.

> https://wiki.servarr.com/sonarr/supported#notifications

I can add all the missing fields for the connections I modified. Do I just log in there and edit or there is a repository somewhere for it?

File: src/NzbDrone.Core/Notifications/MediaBrowser/MediaBrowserProxy.cs [hunk 1]
```diff
@@ -66,10 +66,7 @@ private string ProcessRequest(HttpRequest request, MediaBrowserSettings settings
 
         private HttpRequest BuildRequest(string path, MediaBrowserSettings settings)
         {
-            var scheme = settings.UseSsl ? "https" : "http";
-            var url = $@"{scheme}://{settings.Address}/mediabrowser";
-            
-            return new HttpRequestBuilder(url).Resource(path).Build();
+            return new HttpRequestBuilder($@"{settings.Address}/mediabrowser").Resource(path).Build();
```
markus101, line 69: `@` is unnecessary.

src/NzbDrone.Core/Notifications/MediaBrowser/MediaBrowserSettings.cs
```diff
@@ -12,38 +13,37 @@ public MediaBrowserSettingsValidator()
...
 
         [FieldDefinition(1, Label = "Port")]
-        public int Port { get; set; }
+        public int Port { get; set; } = 8096;
 
         [FieldDefinition(2, Label = "Use SSL", Type = FieldType.Checkbox, HelpText = "Connect to Emby over HTTPS instead of HTTP")]
         public bool UseSsl { get; set; }
 
-        [FieldDefinition(3, Label = "API Key", Privacy = PrivacyLevel.ApiKey)]
+        [FieldDefinition(3, Label = "Url Base", Type = FieldType.Textbox, Advanced = true, HelpText = "Adds a prefix to the Emby server url, see http://[host]:[port]/[urlBase]")]
```
markus101, line 33: `see x` would refer to a link, instead use the same wording we use for SABnzbd: `Adds a prefix to the Sabnzbd url, e.g. http://[host]:[port]/[urlBase]/api`

src/NzbDrone.Core/Notifications/MediaBrowser/MediaBrowserSettings.cs
```diff
@@ -12,38 +13,37 @@ public MediaBrowserSettingsValidator()
...
+        [FieldDefinition(5, Label = "Send Notifications", HelpText = "Have MediaBrowser send notfications to configured providers", Type = FieldType.Checkbox)]
         public bool Notify { get; set; }
 
-        [FieldDefinition(5, Label = "Update Library", HelpText = "Update Library on Import & Rename?", Type = FieldType.Checkbox)]
+        [FieldDefinition(6, Label = "Update Library", HelpText = "Update Library on Import & Rename?", Type = FieldType.Checkbox)]
         public bool UpdateLibrary { get; set; }
 
         [JsonIgnore]
-        public string Address => $"{Host}:{Port}";
+        public string Address => $"{(UseSsl ? "https" : "http")}://{Host}:{Port}{(String.IsNullOrEmpty(UrlBase) ? "" : $"/{UrlBase}")}";
```
markus101, line 46: This should be built in the proxy. Instead of passing individual values to those proxy functions the settings object can be passed in, follow what is already being done in SabnzbdProxy: https://github.com/Sonarr/Sonarr/blob/7f079ac8f29ae5972e85fc53c4951c1882329595/src/NzbDrone.Core/Download/Clients/Sabnzbd/SabnzbdProxy.cs#L38

Same applies for the other media servers.

src/NzbDrone.Core/Validation/RuleBuilderExtensions.cs
```diff
@@ -38,6 +38,10 @@ public static class RuleBuilderExtensions
         {
             return ruleBuilder.SetValidator(new RegularExpressionValidator(@"^(?!\/?https?://[-_a-z0-9.]+)", RegexOptions.IgnoreCase)).WithMessage($"Must be a valid URL path (ie: '{example}')");
         }
+        public static IRuleBuilderOptions<T, string> ActuallyValidUrlBase<T>(this IRuleBuilder<T, string> ruleBuilder, string example = "sonarr/base")
```
markus101, line 41: Why do we need another validator? Why is this one actually valid and the other isn't? The other requires a leading slash which is valid and already in use for download clients.

markus101 (changes requested):
Several suggestions to follow the existing patterns we use for download clients so things are consistent.

src/NzbDrone.Core/Validation/RuleBuilderExtensions.cs
pixxon: The 'Actually' part is there to catch attention in case I forget to mention that the original Validator is buggy.

Saying that anything is a valid url base, that is not a root url is a faulty logic by itself. If we include the fact that the root url regex itself is buggy, we raise the bug potential to even higher level.

image
image
Both picture above show invalid root urls, which are being accepted by the current validator. 

I've decided to first create a new good validator, before wandering into other components, to retro fit the old urls with database migration.

src/NzbDrone.Core/Notifications/MediaBrowser/MediaBrowserSettings.cs
pixxon: Should I fix the "see x" phrase in other mentions as well? I've used that one, because Deluge actually uses it. ( And without shame, I copy pasted it. )

src/NzbDrone.Core/Notifications/MediaBrowser/MediaBrowserProxy.cs
pixxon: Copy paste strikes again, will fix it!

src/NzbDrone.Core/Notifications/MediaBrowser/MediaBrowserSettings.cs
pixxon: Reason I have it in the 'Settings' is that it can be referenced infinite times. For example Xmbc uses it 16 times, which would require me to write the following line 16 times into the proxy.

Can I keep the Address in the Settings, but use either the HttpRequestBuilder or another common utility, to format the string?

src/NzbDrone.Core/Validation/RuleBuilderExtensions.cs
markus101: Other characters are valid, they just need to be URL encoded though.

> I've decided to first create a new good validator

I think we can just fix the one that already exists and if there are issues we fix them, though tests should be added.

> before wandering into other components, to retro fit the old urls with database migration.

Presumably these URL bases are already working, so I don't think there is anything to migrate.

src/NzbDrone.Core/Notifications/MediaBrowser/MediaBrowserSettings.cs
markus101: No, I don't think this belongs in settings. XBMC was already using `HttpRequestBuilder.BuildBaseUrl` in `SendCommand` / `ProcessRequest`. Why would it have to be written 16 times in the proxy?

src/NzbDrone.Core/Notifications/MediaBrowser/MediaBrowserSettings.cs
markus101: Ahh, yeah, it's be good to fix that one too

src/NzbDrone.Core/Validation/RuleBuilderExtensions.cs
pixxon: > Other characters are valid, they just need to be URL encoded though.

This implies that the content in the Url Base is *before* pct encode, which means anything could be added there, as they will be encoded properly.

However, if we say that here must be *after* pct encode the path must be match the [specification](https://datatracker.ietf.org/doc/html/rfc3986#section-3.3).

This whole sort of unspecified state is why I rather didn't touch the older validations, as it might be totally possible that the user put something there, that works *by magic* and the validator would complain about it after upgrade.

src/NzbDrone.Core/Notifications/MediaBrowser/MediaBrowserSettings.cs
pixxon: There are 16 references to `XbmcSettings.Address` in 13 functions, where the settings come as parameters. This would mean, I have to write `var baseUrl = HttpRequestBuilder.BuildBaseUrl(settings.UseSsl, settings.Host, settings.Port, settings.UrlBase);` 13 times at those locations.

`Address` is a "hidden" member of many settings, including download clients, connections. If you want to change settings to include only pure members, then it will either result in code duplication or requires structural changes for proxies to be able cacheing these internally.

src/NzbDrone.Core/Validation/RuleBuilderExtensions.cs
pixxon: Then I will just implement Url Base to comply with rfc3986 paths and if it cases validation errors upon edits, users will have to manually migrate it for themselves? That's acceptable for me.

src/NzbDrone.Core/Notifications/MediaBrowser/MediaBrowserSettings.cs
pixxon: Then I will remove the `Address` property from the settings of these connections. ( It will cause some mismatch with others as they have this `Address` field, but I rather not refactor everything in a single pr. )

src/NzbDrone.Core/Validation/RuleBuilderExtensions.cs
markus101: Yeah, I think that's safe enough.

src/NzbDrone.Core/Notifications/MediaBrowser/MediaBrowserSettings.cs
markus101: Yeah, that's reasonable, more so-not make it worse and we can clean up the remaining afterwards.

markus101:
Closing as this hasn't seen any activity in more than a year.

PR #4779, opened 2021-12-06
Fixed: Convert Trakt list name to URL slug

stevietv:
#### Database Migration
NO

#### Description
Matching the functionality available in Radarr - this now the ToUrlSlug function to the parser.
The this is used to clean the user entered listname for trakt lists to match to how Trakt generates the URL for a list.

#### Todos
- [ ] Tests
- [ ] Wiki Updates

*


src/NzbDrone.Core/Parser/Parser.cs
```diff
@@ -761,6 +761,29 @@ public static string RemoveFileExtension(string title)
             return title;
         }
 
+        public static string ToUrlSlug(string value)
```
markus101, line 764: I don't think this should be part of the parser, `StringExtensions` would be a better place.

src/NzbDrone.Core/ImportLists/Trakt/List/TraktListRequestGenerator.cs
```diff
@@ -26,7 +26,7 @@ private IEnumerable<ImportListRequest> GetSeriesRequest()
         {
             var link = Settings.BaseUrl.Trim();
 
-            var listName = Settings.Listname.Trim();
+            var listName = Parser.Parser.ToUrlSlug(Settings.Listname.Trim());
```
markus101, line 29: The trim here is unnecessary.

markus101 (changes requested):

src/NzbDrone.Core/Parser/Parser.cs
stevietv: I would agree. I've added it in Parser as thats where the same function is in Radarr. Let me know if you still want me to move it.

src/NzbDrone.Core/ImportLists/Trakt/List/TraktListRequestGenerator.cs
stevietv: are you sure, is listname getting trimmed elsewhere?

markus101: The spaces get replaced with `-` then trimmed within `ToUrlSlug`, if we want to be more explicit we could `Trim()` within that function, but I think letting them be replaced and cleaned up is fine.

src/NzbDrone.Core/Parser/Parser.cs
markus101: Yeah, let's move it.

src/NzbDrone.Common/Extensions/StringExtensions.cs
```diff
@@ -165,5 +165,28 @@ public static bool ContainsIgnoreCase(this IEnumerable<string> source, string va
         {
             return source.Contains(value, StringComparer.InvariantCultureIgnoreCase);
         }
+                
+        public static string ToUrlSlug(string value)
```
markus101, line 169: `this string value`

src/NzbDrone.Core/ImportLists/Trakt/List/TraktListRequestGenerator.cs
```diff
@@ -26,7 +26,7 @@ private IEnumerable<ImportListRequest> GetSeriesRequest()
         {
             var link = Settings.BaseUrl.Trim();
 
-            var listName = Settings.Listname.Trim();
+            var listName = StringExtensions.ToUrlSlug(Settings.Listname);
```
markus101, line 29: With the other change, this will change to `Settings.Listname.ToUrlSlug()`

markus101 (changes requested):
Just need to convert it to an extension method, but otherwise it looks good.

.gitignore
```diff
@@ -1,144 +0,0 @@
-# Build Folders (you can keep bin if you'd like, to store dlls and pdbs)
```
markus101, line 1: This needs to be reverted.

stevietv: there, I think I fixed it. stil new to Visual Studio and GitHub

markus101 (approved):

PR #4791, opened 2021-12-10
Fixed: Improved messaging when episode file was detected as deleted from disk

bakerboy448:
#### Database Migration
NO

#### Description
Due to the confusing message, users frequently incorrectly believe Sonarr deleted the file when it simply could no longer be seen.

#### Todos
-  Tests
-  Wiki Updates

https://forums.sonarr.tv/t/unable-to-find-file-on-disk-so-it-was-deleted/29282

### DIFF
--- a/frontend/src/Activity/History/Details/HistoryDetails.js
+++ b/frontend/src/Activity/History/Details/HistoryDetails.js
@@ -207,7 +207,7 @@ function HistoryDetails(props) {
         reasonMessage = 'File was deleted by via UI';
         break;
       case 'MissingFromDisk':
-        reasonMessage = 'Sonarr was unable to find the file on disk so it was removed';
+        reasonMessage = 'Sonarr was unable to find the file on disk so the file was unlinked from the episode in the database';
         break;
       case 'Upgrade':
         reasonMessage = 'File was deleted to import an upgrade';


stevietv:
would 'entry was removed from the database' be even less confusing?

bakerboy448:
> would 'entry was removed from the database' be even less confusing?

could be;  open to see what markus/taloth think

frontend/src/Activity/History/Details/HistoryDetails.js
```diff
@@ -207,7 +207,7 @@ function HistoryDetails(props) {
         reasonMessage = 'File was deleted by via UI';
         break;
       case 'MissingFromDisk':
-        reasonMessage = 'Sonarr was unable to find the file on disk so it was removed';
+        reasonMessage = 'Sonarr was unable to find the file on disk so the file was removed from the database';
```
markus101, line 210: I think: 
`Sonarr was unable to find the file on disk so the file was unlinked from the episode in the database`

is the most accurate we can be, since the file itself was never in the database

bakerboy448: agreed; pushed

PR #4820, opened 2021-12-27
Don't lowercase UrlBase in ConfigFileProvider

evilmarty:
UrlBase should honour the case it is given.

#### Database Migration
NO

#### Description
When setting `UrlBase` with a value the characters are changed to lowercase. This causes issues where reverse proxies are case-sensitive thus fail to proxy the requests because of case differences.

### DIFF
--- a/src/NzbDrone.Core/Configuration/ConfigFileProvider.cs
+++ b/src/NzbDrone.Core/Configuration/ConfigFileProvider.cs
@@ -195,7 +195,7 @@ namespace NzbDrone.Core.Configuration
                     return urlBase;
                 }
 
-                return "/" + urlBase.Trim('/').ToLower();
+                return "/" + urlBase;
             }
         }
 


src/NzbDrone.Core/Configuration/ConfigFileProvider.cs
```diff
@@ -195,7 +195,7 @@ public string UrlBase
                     return urlBase;
                 }
 
-                return "/" + urlBase.Trim('/').ToLower();
+                return "/" + urlBase;
```
markus101, line 198: Leave the `.Trim()`

markus101 (changes requested):

src/NzbDrone.Core/Configuration/ConfigFileProvider.cs
evilmarty: If you notice on line [191](https://github.com/Sonarr/Sonarr/pull/4820/files#diff-bb1f86bcc03982676d45aa15a713f388209474ee00645e920cc358780566b454R191) `urlBase` is already trimmed thus it is redundant to `Trim` the string again.

markus101: `Trim('/')` trims leading and trailing `/`s in that case, `Trim()` here trims leading and trailing spaces, though changing `Trim('/')` to `Trim('/', ' ')` would prevent unexpected URL bases with just slashes and spaces from being set.

evilmarty: I get what you're saying but like I said it's superfluous because it's already being applied when `urlBase` is set. Thus, if I were to re-add it:
```c#
var urlBase = GetValue("UrlBase", "").Trim('/');

if (urlBase.IsNullOrWhiteSpace())
{
    return urlBase;
}

return "/" + urlBase.Trim('/');
```

Would be equivalent to (ignoring the if statement):
```c#
return "/" + GetValue("UrlBase", "").Trim('/').Trim('/');
```

markus101: Doh, kept reading the removed one as trim without arguments, which would legitimately remove extra spaces. Nothing to see here.

markus101:
One issue we've been discussing on Discord is the fact that anyone that by mistake or otherwise used `/Sonarr` instead of `/sonarr`, but it's been working may be in for a rude awakening after upgrading to a version that has this change, so it's not something we can land in the existing v3 builds of Sonarr and would have to wait for v4.

evilmarty:
That's fair. Thanks @markus101.

markus101:
Pulled this into the v4 WIP branch.

thiagomoreirac:
Hello, 

I have the same issue with an nginx reverse proxy. Is it possible to create a new option case_sensitivity which could enable it explicitly for those who need it?  the default behavior can continue to be the case insensitivity. 

Thank you.

PR #4821, opened 2021-12-28
Use BuildInfo.AppName for RARBG appId instead of hardcoded value

Ga22be:
* Added app_id to captcha check to avoid 403 forbidden error
* Migrated app_id from hard coded to BuildInfo.AppName

#### Database Migration
NO

#### Description
Previously the captcha check threw an exception in the logs since RARBG's API responds with 403 forbidden if `app_id` is skipped/overlooked in the query. I also took the opportunity to update `app_id` from hard coded to use the `BuildInfo.AppName` value instead.

### DIFF
--- a/src/NzbDrone.Core.Test/IndexerTests/RarbgTests/RarbgFixture.cs
+++ b/src/NzbDrone.Core.Test/IndexerTests/RarbgTests/RarbgFixture.cs
@@ -3,6 +3,7 @@ using System.Linq;
 using FluentAssertions;
 using Moq;
 using NUnit.Framework;
+using NzbDrone.Common.EnvironmentInfo;
 using NzbDrone.Common.Http;
 using NzbDrone.Core.Indexers;
 using NzbDrone.Core.Indexers.Rarbg;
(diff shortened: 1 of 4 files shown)


Taloth:
iirc the Captcha stuff hasn't worked for ages due to changes to CF Captcha implementation.

Did you run into CF captcha? And were you able to get a valid captcha token?

Ga22be:
@Taloth, I did not. But I pressed the button in the UI and was greeted with a long stack trace in the logs. Since I don't like those I decided to track down the origin and amend it. 😅 

I discussed it shortly in the Prowlarr discord last night with @bakerboy448, @ta264, and "PearsonFlyer". They suggested I create a PR there and here as it supposedly could flow down-stream from here.

Taloth:
I can merge it, it just won't help anyone since CF Captcha changed... I basically should remove the whole feature, but never bothered. :smile:

Ga22be:
Well, this at least ensures it is a more bug free feature that you remove ^^

Ga22be:
@Taloth, will you merge this or disregard it and remove the feature all together? Because if one clicks the captcha button the log will be filled with exceptions as of now :)

markus101:
This slipped through the cracks, but the changes are fine, so I've finally landed them, thanks for contributing.

Ga22be:
Thanks, happy to have contributed

PR #4833, opened 2022-01-03
Set frontend `theme-color` to logo color

agneevX:
#### Database Migration
NO

#### Description
Set `theme-color` to Sonarr's logo color.

09894E05-CCD5-4E5A-9BA2-043A68661B37

#### Todos
- [ ] Tests
- [ ] Wiki Updates

*

### DIFF
--- a/frontend/src/index.ejs
+++ b/frontend/src/index.ejs
@@ -6,8 +6,8 @@
     <meta name="mobile-web-app-capable" content="yes" />
     <meta name="apple-mobile-web-app-capable" content="yes" />
 
-    <!-- Chrome, Opera, and Firefox OS -->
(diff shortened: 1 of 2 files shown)


thezoggy:
What about the line below it?

Stanzilla:
Yeah I think it would look better if you'd try to match the header color

markus101:
IIRC the original reason this was changed from matching the header or logo colour was because it was super bright on some devices, not sure if this is something we want to change as it will disrupt existing users.

agneevX:
I see. I've had a chance to review this across devices and I think the current implementation is fine and works well, even with custom CSS themes.

PR #4846, opened 2022-01-08
Fixed: Clarify Indexer Priority Helptext

bakerboy448:
#### Database Migration
NO

#### Description
Currently many users commonly are mislead by the existing helptext that implies indexer priority will change how indexers are searched or any other possibilities other than simply being a tiebreaker.

This clarifies the currently confusing helptext

#### Todos
-  Tests
- Wiki Updates - Already done `(Advanced Option) Indexer Priority - Priority of this indexer to prefer one indexer over another in release tiebreaker scenarios. 1 is highest priority and 50 is lowest priority.`

*

### DIFF
--- a/frontend/src/Settings/Indexers/Indexers/EditIndexerModalContent.js
+++ b/frontend/src/Settings/Indexers/Indexers/EditIndexerModalContent.js
@@ -143,7 +143,7 @@ function EditIndexerModalContent(props) {
                 <FormInputGroup
                   type={inputTypes.NUMBER}
                   name="priority"
-                  helpText="Indexer Priority from 1 (Highest) to 50 (Lowest). Default: 25."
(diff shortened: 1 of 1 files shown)


frontend/src/Settings/Indexers/Indexers/EditIndexerModalContent.js
```diff
@@ -143,7 +143,7 @@ function EditIndexerModalContent(props) {
                 <FormInputGroup
                   type={inputTypes.NUMBER}
                   name="priority"
-                  helpText="Indexer Priority from 1 (Highest) to 50 (Lowest). Default: 25."
+                  helpText="Indexer Priority to be used for release ranking tiebreaking from 1 (Highest) to 50 (Lowest). Default: 25."
```
markus101, line 146: Not a fan of this wording. Maybe something like:

`Indexer Priority from 1 (Highest) to 50 (Lowest). Default: 25. Used when grabbing releases as a tiebreaker for otherwise equal releases, Sonarr will still use all enabled indexers for RSS Sync and Searching`

bakerboy448: agreed

Issue #4850, opened 2022-01-12
SeriesRepository.FindByTitle Cannot Handle Numeric Series Name


bejhan:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

A release for a series with a numeric name that includes the series year is rejected with error `Unknown Series`.
e.g. `1883 (2021) S01E01 1883 (1080p AMZN Webrip x265 10bit EAC3 5 1 - Goki)[TAoE]`

### Expected Behavior

I took this to the Discord support channel before reporting this bug and was told that this is expected behavior since the year is not actually part of the series name. However, after tracing through the code, I am convinced that there is code in place to handle this case and someone there encouraged me to the post a bug in case I am correct in my investigation.

The `Unknown Series` error originates in `DownloadDecisionMaker`:
```
var remoteEpisode = _parsingService.Map(parsedEpisodeInfo, report.TvdbId, report.TvRageId, searchCriteria);
remoteEpisode.Release = report;

if (remoteEpisode.Series == null)
{
	var reason = "Unknown Series";
```

When the release name contains the series year, `_parsingService.Map()` ends up calling `SeriesRepository.FindByTitle()` via the following stack of calls:
```
Sonarr.Core.dll!NzbDrone.Core.Tv.SeriesRepository.FindByTitle(string cleanTitle, int year) Line 55	C#
Sonarr.Core.dll!NzbDrone.Core.Tv.SeriesService.FindByTitle(string title, int year) Line 146	C#
Sonarr.Core.dll!NzbDrone.Core.Parser.ParsingService.GetSeries(NzbDrone.Core.Parser.Model.ParsedEpisodeInfo parsedEpisodeInfo, int tvdbId, int tvRageId, NzbDrone.Core.DataAugmentation.Scene.SceneMapping sceneMapping, NzbDrone.Core.IndexerSearch.Definitions.SearchCriteriaBase searchCriteria) Line 390	C#
Sonarr.Core.dll!NzbDrone.Core.Parser.ParsingService.Map(NzbDrone.Core.Parser.Model.ParsedEpisodeInfo parsedEpisodeInfo, int tvdbId, int tvRageId, NzbDrone.Core.Tv.Series series, NzbDrone.Core.IndexerSearch.Definitions.SearchCriteriaBase searchCriteria) Line 167	C#
[...]

PR #4851, opened 2022-01-12
Fixed: Parsing of numeric only titles that include a year

bejhan:
#### Database Migration
NO

#### Description
There is logic to handle parsing when series year is included in release name by matching against `CleanTitle` and `Year` in `Series` table. This logic previously failed to find a match for numeric series name due to extraneous whitespace. Therefore, this PR is simply trimming numeric series names to prevent extraneous whitespace from preventing query match. See the issue description and comments for a much more in-depth explanation/rationale.

#### Todos
- [x] Tests
- [ ] Wiki Updates

* #4850

### DIFF
--- a/src/NzbDrone.Core/Parser/Parser.cs
+++ b/src/NzbDrone.Core/Parser/Parser.cs
@@ -455,7 +455,7 @@ namespace NzbDrone.Core.Parser
         private static readonly Regex LanguageRegex = new Regex(@"(?:\W|_)(?<italian>\b(?:ita|italian)\b)|(?<german>german\b|videomann)|(?<flemish>flemish)|(?<greek>greek)|(?<french>(?:\W|_)(?:FR|VOSTFR)(?:\W|_))|(?<russian>\brus\b)|(?<dutch>nl\W?subs?)|(?<hungarian>\b(?:HUNDUB|HUN)\b)|(?<spanish>\b(?:español|castellano)\b)",
                                                                 RegexOptions.IgnoreCase | RegexOptions.Compiled);
 
-        private static readonly Regex YearInTitleRegex = new Regex(@"^(?<title>.+?)(?:\W|_)?(?<year>\d{4})",
+        private static readonly Regex YearInTitleRegex = new Regex(@"^(?<title>.+?)[-_. ]+?\(?(?<year>\d{4})\)?",
                                                                 RegexOptions.IgnoreCase | RegexOptions.Compiled);
 
         private static readonly Regex TitleComponentsRegex = new Regex(@"^(?:(?<title>.+?) \((?<title>.+?)\)|(?<title>.+?) \| (?<title>.+?))$",
--- a/src/NzbDrone.Core.Test/ParserTests/ParserFixture.cs
+++ b/src/NzbDrone.Core.Test/ParserTests/ParserFixture.cs
@@ -35,6 +35,31 @@ namespace NzbDrone.Core.Test.ParserTests
             result.Should().Be(title.CleanSeriesTitle());
         }
 
+        [TestCase("Series S03E14 720p HDTV X264-DIMENSION", "Series")]
+        [TestCase("Series.S03E14.720p.HDTV.X264-DIMENSION", "Series")]
+        [TestCase("Series-S03E14-720p-HDTV-X264-DIMENSION", "Series")]
+        [TestCase("Series_S03E14_720p_HDTV_X264-DIMENSION", "Series")]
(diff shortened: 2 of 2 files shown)


bejhan:
@bakerboy448, I know you said #4850 is not a bug but there are even tests to cover this case (see `should_fallback_to_title_without_year_and_year_when_title_lookup_fails` in `GetSeriesFixture.cs` and `should_FindByTitle_using_year_when_FindByTitle_matching_fails` in `MapFixture.cs`), so I do believe this case was intended to be covered.

markus101 (changes requested):
This needs a test added to ParserFixture to verify the change and ensure it doesn't regress in the future.

bejhan:
After further investigation, I determined that trimming the series name in `CleanSeriesTitle()` is simply a band-aid solution.
The extraneous whitespace actually originates in the `GetSeriesTitleInfo()`, where `YearInTitleRegex` is including the trailing space in `seriesTitleInfo.TitleWithoutYear`.

src/NzbDrone.Core/Parser/Parser.cs
```diff
@@ -455,7 +455,7 @@ public static class Parser
         private static readonly Regex LanguageRegex = new Regex(@"(?:\W|_)(?<italian>\b(?:ita|italian)\b)|(?<german>german\b|videomann)|(?<flemish>flemish)|(?<greek>greek)|(?<french>(?:\W|_)(?:FR|VOSTFR)(?:\W|_))|(?<russian>\brus\b)|(?<dutch>nl\W?subs?)|(?<hungarian>\b(?:HUNDUB|HUN)\b)|(?<spanish>\b(?:español|castellano)\b)",
                                                                 RegexOptions.IgnoreCase | RegexOptions.Compiled);
 
-        private static readonly Regex YearInTitleRegex = new Regex(@"^(?<title>.+?)(?:\W|_)?(?<year>\d{4})",
+        private static readonly Regex YearInTitleRegex = new Regex(@"^(?<title>.+?)(?:\W|_)*?(?<year>\d{4})",
```
bejhan, line 458: This regex will also trim any trailing non-word characters from the series name, which would affect something like `American Dad! (2005) S01E01`, though I don't think any releases actually include the exclamation mark.
Still, it might be safer to explicitly check for whitespace instead: `^(?<title>.+?)\s*(?:\W|_)?(?<year>\d{4})`

Thoughts, @markus101?

markus101: Generally they don't, but could probably do something like `^(?<title>.+?)[-_. ]+?(?<year>\d{4})` which would only capture it on characters that are actually used as separators (we've converted a bunch of `\W|_` to those 4 characters in other regex). You want to use `+` instead of `*` because we want one or more, instead of zero or more.

bejhan: I used `*` to handle a case like `Series2022` but I guess that's considered invalid, right?

markus101: I think it’s safe to treat that as invalid and let it fail.

src/NzbDrone.Core/Parser/Parser.cs
```diff
@@ -455,7 +455,7 @@ public static class Parser
         private static readonly Regex LanguageRegex = new Regex(@"(?:\W|_)(?<italian>\b(?:ita|italian)\b)|(?<german>german\b|videomann)|(?<flemish>flemish)|(?<greek>greek)|(?<french>(?:\W|_)(?:FR|VOSTFR)(?:\W|_))|(?<russian>\brus\b)|(?<dutch>nl\W?subs?)|(?<hungarian>\b(?:HUNDUB|HUN)\b)|(?<spanish>\b(?:español|castellano)\b)",
                                                                 RegexOptions.IgnoreCase | RegexOptions.Compiled);
 
-        private static readonly Regex YearInTitleRegex = new Regex(@"^(?<title>.+?)(?:\W|_)?(?<year>\d{4})",
+        private static readonly Regex YearInTitleRegex = new Regex(@"^(?<title>.+?)[-_. ]+?\(?(?<year>\d{4})\)?",
```
bejhan, line 458: Now that the regex is more precise, it will only handle separator then year (e.g. `Series 2021` or `Series.2021`) or separator then parenthesis then year (e.g. `Series (2021)`). 
@markus101, are there any other characters that might enclose year (e.g `[`) that we need to cover?

markus101: I'm sure someone will come up with some wild naming for some release, but not concerned with that.

PR #4855, opened 2022-01-16
Fix issues when setting up Emby in the Connect settings

LASER-Yi:
#### Database Migration
NO

#### Description
Fix issues when using Notification with Emby, the documentation can be found in [Emby's document](https://github.com/MediaBrowser/Emby/wiki/User-Authentication#authenticating-a-user)
image

#### Todos
- [ ] Tests
- [ ] Wiki Updates

NO

### DIFF
--- a/src/NzbDrone.Core/Notifications/MediaBrowser/MediaBrowserProxy.cs
+++ b/src/NzbDrone.Core/Notifications/MediaBrowser/MediaBrowserProxy.cs
@@ -25,7 +25,7 @@ namespace NzbDrone.Core.Notifications.Emby
             {
                 Name = title,
                 Description = message,
-                ImageUrl = "https://raw.github.com/NzbDrone/NzbDrone/develop/Logo/64.png"
+                ImageUrl = "https://raw.github.com/Sonarr/Sonarr/develop/Logo/64.png"
             }.ToJson());
 
             ProcessRequest(request, settings);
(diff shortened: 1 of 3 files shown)


src/NzbDrone.Core/Notifications/MediaBrowser/MediaBrowserProxy.cs
```diff
@@ -22,11 +22,11 @@ public void Notify(MediaBrowserSettings settings, string title, string message)
...
             request.SetContent(new
-                           {
-                               Name = title,
-                               Description = message,
-                               ImageUrl = "https://raw.github.com/NzbDrone/NzbDrone/develop/Logo/64.png"
-                           }.ToJson());
+            {
+                Name = title,
+                Description = message,
+                ImageUrl = "https://raw.github.com/NzbDrone/NzbDrone/develop/Logo/64.png"
```
bakerboy448, line 28: Sonarr/Sonarr

bakerboy448 (changes requested):

src/NzbDrone.Core/Notifications/MediaBrowser/MediaBrowserProxy.cs
LASER-Yi: Updated

bakerboy448:
Note: likely breaking API changes for previous legacy versions

markus101:
How long ago did this changes get implemented on Emby's side, does Jellyfin also support this new header and URL Base as well?

Qstick:
Based on what I could dig up, this happened before the JF fork (back in 2018 at some point). Though it was not well documented at all. It seems like JF supports it, but that was my biggest worry in letting this through. I think someone needs to test with JF

PR #4856, opened 2022-01-16
Fixed: Warning for search all

PearsonFlyer:
#### Database Migration
NO

#### Description
Changed the warning text on Search All to properly indicate how to stop it once started. Restarting Sonarr does not work.

#### Todos
- [ ] Tests
- [ ] Wiki Updates

*

### DIFF
--- a/frontend/src/Wanted/CutoffUnmet/CutoffUnmet.js
+++ b/frontend/src/Wanted/CutoffUnmet/CutoffUnmet.js
@@ -252,7 +252,7 @@ class CutoffUnmet extends Component {
                         Are you sure you want to search for all {totalRecords} Cutoff Unmet episodes?
                       </div>
                       <div>
-                        This cannot be cancelled once started without restarting Sonarr.
+                        This cannot be cancelled once started without disabling all of your indexers.
                       </div>
                     </div>
                   }
--- a/frontend/src/Wanted/Missing/Missing.js
+++ b/frontend/src/Wanted/Missing/Missing.js
@@ -265,7 +265,7 @@ class Missing extends Component {
                         Are you sure you want to search for all {totalRecords} missing episodes?
                       </div>
                       <div>
-                        This cannot be cancelled once started without restarting Sonarr.
+                        This cannot be cancelled once started without disabling all of your indexers.
                       </div>
                     </div>
                   }


bakerboy448 (approved):

frontend/src/Wanted/CutoffUnmet/CutoffUnmet.js
```diff
@@ -252,7 +252,7 @@ class CutoffUnmet extends Component {
                         Are you sure you want to search for all {totalRecords} Cutoff Unmet episodes?
                       </div>
                       <div>
-                        This cannot be cancelled once started without restarting Sonarr.
+                        This cannot be cancelled once started without disabling all of your indexers.
```
markus101, line 255: Restarting Sonarr is still a valid way to cancel, why not list both?

PearsonFlyer: Is it? As far as I'm aware (and what everyone tells us in support channels), it continues on startup again and a restart is not a valid way to stop a search-all.

PearsonFlyer: I just tested it with Cutoff Unmet, and it does seem to interrupt the process. This is different behavior than the other arrs, so if you would like to decline this PR, I'm fine with that.

markus101: That's news to me they don't behave the same way, in Sonarr any in progress task that is not completed when Sonarr starts is tossed out, scheduled tasks would be re-queued by the scheduler, user queued tasks are just lost. I'm curious why this was changed downstream, maybe @Qstick knows?

Qstick: I'm not sure it was intentional if it did change

Qstick: So I think this will only be different in Radarr and not the others. It's because we don't have a wanted page and we always use normal search commands when firing UI searches, never the singular cutoff unmet or missing search commands

markus101: Ahh that makes sense, in which case, I think restarting Sonarr or disabling all indexers is what we want to say here.

markus101, line 255: `This cannot be cancelled once started without restarting Sonarr or disabling all of your indexers.`

markus101 (changes requested):
Restarting is still a valid way to cancel it.

markus101:
Landed the proposed verbiage in v4.

PR #4870, opened 2022-01-24
New: Add qBittorrent sequential order and first and last piece priority options

mathispesch:
#### Database Migration

NO

#### Description

This PR adds two settings to the download client qBittorrent: download in sequential order and prioritise first and last pieces.

Those settings were introduced in the qBittorrent API in 4.0.0 but until api v2 (4.1.0) they only work if the form is submitted as a `form-data` and not as a `x-www-form-urlencoded` (couldn't find the reason why but their doc says to use `form-data` after qBittorrent 3.3.1 and the tests I did showed this). As earlier versions of the qBittorrent API are only compatible with `x-www-form-urlencoded` and as I don't want to introduce changes to the HttpRequestBuilder to change the ContentType, this is only available for qBittorrent 4.1.0+.

image

#### Todos

- [x] Tests -> none added
- [x] Wiki Updates -> not needed

### DIFF
--- a/src/NzbDrone.Core/Download/Clients/QBittorrent/QBittorrentProxyV2.cs
+++ b/src/NzbDrone.Core/Download/Clients/QBittorrent/QBittorrentProxyV2.cs
@@ -143,20 +143,7 @@ namespace NzbDrone.Core.Download.Clients.QBittorrent
                                                 .Post()
                                                 .AddFormParameter("urls", torrentUrl);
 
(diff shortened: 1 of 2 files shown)


bakerboy448:
Needs Unit Tests

https://github.com/Sonarr/Sonarr/blob/develop/src/NzbDrone.Core.Test/Download/DownloadClientTests/QBittorrentTests/QBittorrentFixture.cs

PR #4871, opened 2022-01-24
New: Add more groups that do not follow -Group format

Tamberlox:
#### Database Migration
NO

#### Description
Adds a few more groups that do not follow the conventional -Group format

#### Todos
/

/

### DIFF
--- a/src/NzbDrone.Core/Parser/Parser.cs
+++ b/src/NzbDrone.Core/Parser/Parser.cs
@@ -457,7 +457,7 @@ namespace NzbDrone.Core.Parser
         private static readonly Regex ExceptionReleaseGroupRegexExact = new Regex(@"(?<releasegroup>(?:D\-Z0N3|Fight-BB)\b)", RegexOptions.IgnoreCase | RegexOptions.Compiled);
 
         //groups whose releases end with RlsGroup) or RlsGroup]
-        private static readonly Regex ExceptionReleaseGroupRegex = new Regex(@"(?<releasegroup>(Tigole|Joy|ImE|UTR|t3nzin|Anime Time|Project Angel|Hakata Ramen)(?=\]|\)))", RegexOptions.IgnoreCase | RegexOptions.Compiled);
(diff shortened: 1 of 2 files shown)


bakerboy448:
Tests fail

> Expected string to be "Ghost" with a length of 5, but "QxR" has a length of 3, differs near "QxR" (index 0).

src/NzbDrone.Core/Parser/Parser.cs
```diff
@@ -457,7 +457,7 @@ public static class Parser
         private static readonly Regex ExceptionReleaseGroupRegexExact = new Regex(@"(?<releasegroup>(?:D\-Z0N3|Fight-BB)\b)", RegexOptions.IgnoreCase | RegexOptions.Compiled);
 
         //groups whose releases end with RlsGroup) or RlsGroup]
-        private static readonly Regex ExceptionReleaseGroupRegex = new Regex(@"(?<releasegroup>(Tigole|Joy|ImE|UTR|t3nzin|Anime Time|Project Angel|Hakata Ramen)(?=\]|\)))", RegexOptions.IgnoreCase | RegexOptions.Compiled);
+        private static readonly Regex ExceptionReleaseGroupRegex = new Regex(@"(?<releasegroup>(QxR|Silence|afm72|Panda|Ghost|MONOLITH|Tigole|Joy|ImE|UTR|t3nzin|Anime Time|Project Angel|Hakata Ramen)(?=\]|\)))", RegexOptions.IgnoreCase | RegexOptions.Compiled);
```
bakerboy448, line 460: anything with QXR will be detected - are they the actual group? it seems they're the organization and then the various other groups are part of it?

QxR being here also causes the tests containing QxR and not expecting UxR to fail

bakerboy448 (changes requested):

src/NzbDrone.Core/Parser/Parser.cs
Tamberlox: It's the same case as with UTR actually. I'm not sure how you guys managed that use case though.

Tamberlox:
I've gone and removed QxR from the code.

PR #4874, opened 2022-01-26
Fixed: Updated ruTorrent stopped state helptext

PearsonFlyer:
#### Database Migration
NO

#### Description
Updated helptext for option to add ruTorrent torrents in a stopped state.

#### Todos
- [ ] Tests
- [ ] Wiki Updates

*

### DIFF
--- a/src/NzbDrone.Core/Download/Clients/rTorrent/RTorrentSettings.cs
+++ b/src/NzbDrone.Core/Download/Clients/rTorrent/RTorrentSettings.cs
@@ -64,7 +64,7 @@ namespace NzbDrone.Core.Download.Clients.RTorrent
         [FieldDefinition(10, Label = "Older Priority", Type = FieldType.Select, SelectOptions = typeof(RTorrentPriority), HelpText = "Priority to use when grabbing episodes that aired over 14 days ago")]
         public int OlderTvPriority { get; set; }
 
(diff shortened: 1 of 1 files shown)


bakerboy448 (approved):

src/NzbDrone.Core/Download/Clients/rTorrent/RTorrentSettings.cs
```diff
@@ -64,7 +64,7 @@ public RTorrentSettings()
         [FieldDefinition(10, Label = "Older Priority", Type = FieldType.Select, SelectOptions = typeof(RTorrentPriority), HelpText = "Priority to use when grabbing episodes that aired over 14 days ago")]
         public int OlderTvPriority { get; set; }
 
-        [FieldDefinition(11, Label = "Add Stopped", Type = FieldType.Checkbox, HelpText = "Enabling will prevent magnets from downloading before downloading")]
+        [FieldDefinition(11, Label = "Add Stopped", Type = FieldType.Checkbox, HelpText = "Enabling will add torrents and magnets to ruTorrent in a stopped state")]
```
markus101, line 67: The old wording is clearly wrong, but it looks like it might cause issues with resolving magnets before the download starts, if that's the case we might want to be more specific about that.

PearsonFlyer: I have added some text specifying that it may break magnet links.

PR #4876, opened 2022-01-26
Fixed: Movies with Umlauts are now correctly matched and have correct CleanTitles.

JBou:
#### Database Migration
NO

#### Description
This fixes series with umlauts not getting mapped correctly. 
For example:
`Die.Pfefferkoerner.S17E13.Lebensretter.German.HDTVRip.x264-ATAX` would not get mapped to the Series `Die Pfefferkörner`.
Currently you have to make a request on the Scene Mapping Form to make it work. This should fix especially a lot of german releases getting parsed correctly and should therefore also decrease the [Scene naming requests](https://forums.sonarr.tv/t/scene-naming-tvdb-naming-conflicts/137)
This fix is being ported from [Radarr](https://github.com/Radarr/Radarr/commit/10091b94541cf3b1f6a891cb12c57765de9a1724).

#### Todos
- [x] Tests
- [ ] Wiki Updates

*

### DIFF
--- a/src/NzbDrone.Core/Parser/Parser.cs
+++ b/src/NzbDrone.Core/Parser/Parser.cs
@@ -674,8 +674,10 @@ namespace NzbDrone.Core.Parser
 
             // Replace `%` with `percent` to deal with the 3% case
             title = PercentRegex.Replace(title, "percent");
+            title = NormalizeRegex.Replace(title).ToLower();
+            title = ReplaceGermanUmlauts(title);
 
-            return NormalizeRegex.Replace(title).ToLower().RemoveAccent();
+            return title.RemoveAccent();
         }
 
         public static string NormalizeEpisodeTitle(string title)
@@ -707,6 +709,19 @@ namespace NzbDrone.Core.Parser
             title = DuplicateSpacesRegex.Replace(title, " ");
 
             return title.Trim().ToLower();
+        }
+
+        public static string ReplaceGermanUmlauts(string title)
+        {
+            title = title.Replace("ä", "ae");
+            title = title.Replace("ö", "oe");
+            title = title.Replace("ü", "ue");
+            title = title.Replace("Ä", "Ae");
+            title = title.Replace("Ö", "Oe");
+            title = title.Replace("Ü", "Ue");
+            title = title.Replace("ß", "ss");
+            
+            return title;
         }
 
         public static string ParseReleaseGroup(string title)
--- a/src/NzbDrone.Core.Test/ParserTests/ParserFixture.cs
+++ b/src/NzbDrone.Core.Test/ParserTests/ParserFixture.cs
@@ -68,6 +68,14 @@ namespace NzbDrone.Core.Test.ParserTests
             title.CleanSeriesTitle().Should().Be("seriaes");
         }
 
+        [Test]
+        public void should_replace_umlauts_in_title()
+        {
+            const string title = "Tote Mädchen lügen nicht";
+
+            title.CleanSeriesTitle().Should().Be("totemaedchenluegennicht");
+        }
+
         [TestCase("Sonar TV - Series Title : 02 Road From Code [S04].mp4")]
         public void should_clean_up_invalid_path_characters(string postTitle)
         {


src/NzbDrone.Core/Parser/Parser.cs
```diff
@@ -675,7 +675,7 @@ public static string CleanSeriesTitle(this string title)
             // Replace `%` with `percent` to deal with the 3% case
             title = PercentRegex.Replace(title, "percent");
 
-            return NormalizeRegex.Replace(title).ToLower().RemoveAccent();
+            return ReplaceGermanUmlauts(NormalizeRegex.Replace(title).ToLower()).RemoveAccent();
```
markus101, line 678: Like the `PercentRegex` above these should be separated into their own lines.

src/NzbDrone.Core.Test/ParserTests/ParsingServiceTests/MapFixture.cs
```diff
@@ -20,14 +20,15 @@ public class MapFixture : TestBase<ParsingService>
...
+        private ParsedEpisodeInfo _umlautInfo;
         private SingleEpisodeSearchCriteria _singleEpisodeSearchCriteria;
 
         [SetUp]
         public void Setup()
         {
             _series = Builder<Series>.CreateNew()
-                .With(s => s.Title = "30 Stone")
-                .With(s => s.CleanTitle = "stone")
+                .With(s => s.Title = "Die Pfefferkörner")
```
markus101, line 30: This is not the correct place to to test this change. Tests for this functionality should be in `ParserFixture`, same as is done with `should_remove_accents_from_title`

markus101 (changes requested):
Is there a real world example where this change would actually eliminate the need for a mapping? `Die Pfefferkörner` is a bad example because there wouldn't be an English title with that name (or at least one that would last), possibly valid with metadata in other languages.

One concern with this change is if different languages treat diacritics differently it'll result in mismatches that would need a more confusing alias, but I'm not sure if this concern is a real world issue.

src/NzbDrone.Core/Parser/Parser.cs
JBou: Is it okay like this?
```
title = NormalizeRegex.Replace(title).ToLower();
return ReplaceGermanUmlauts(title).RemoveAccent();
```

src/NzbDrone.Core.Test/ParserTests/ParsingServiceTests/MapFixture.cs
JBou: Should I completely revert MapFixture then?
I just put them there because Radarr also had a Test for umlauts there.

JBou:
> Is there a real world example where this change would actually eliminate the need for a mapping? `Die Pfefferkörner` is a bad example because there wouldn't be an English title with that name (or at least one that would last), possibly valid with metadata in other languages.

Maybe there aren't as many releases as I thought, but another example is:
13 Reasons why - [Tote Mädchen lügen nicht](https://thetvdb.com/series/13-reasons-why)

> One concern with this change is if different languages treat diacritics differently it'll result in mismatches that would need a more confusing alias, but I'm not sure if this concern is a real world issue.

It shouldn't be many mismatches in my opinion, but better check it again. I don't have much experience with release titles in different languages.

I will push a fix for your comments the next days.

src/NzbDrone.Core.Test/ParserTests/ParsingServiceTests/MapFixture.cs
markus101: Yeah I think so.

src/NzbDrone.Core/Parser/Parser.cs
markus101: I think I'd go one step further and go with:

```
title = NormalizeRegex.Replace(title).ToLower();
title = ReplaceGermanUmlauts(title);

return title.RemoveAccent();
```

Makes it a bit more readable and ideally better shows off the preferred way to implement another similar feature in the future.

thezoggy:
seems like this should be a generic fixup rather than a 'german' specific. since more than german users umlauts and the goal is to get it to an ascii equal to make searching more fruitful since most release names dont actually use umulats (as its against accepted charmap)

thezoggy:
providing list of accented/umlaut characters from another project of how we handled them..
```
		// used when replacing accented characters during sorting
		characterEquivalents : {
			'a' : '\u00e1\u00e0\u00e2\u00e3\u00e4\u0105\u00e5', // áàâãäąå
			'A' : '\u00c1\u00c0\u00c2\u00c3\u00c4\u0104\u00c5', // ÁÀÂÃÄĄÅ
			'c' : '\u00e7\u0107\u010d', // çćč
			'C' : '\u00c7\u0106\u010c', // ÇĆČ
			'e' : '\u00e9\u00e8\u00ea\u00eb\u011b\u0119', // éèêëěę
			'E' : '\u00c9\u00c8\u00ca\u00cb\u011a\u0118', // ÉÈÊËĚĘ
			'i' : '\u00ed\u00ec\u0130\u00ee\u00ef\u0131', // íìİîïı
			'I' : '\u00cd\u00cc\u0130\u00ce\u00cf', // ÍÌİÎÏ
			'o' : '\u00f3\u00f2\u00f4\u00f5\u00f6\u014d', // óòôõöō
			'O' : '\u00d3\u00d2\u00d4\u00d5\u00d6\u014c', // ÓÒÔÕÖŌ
			'ss': '\u00df', // ß (s sharp)
			'SS': '\u1e9e', // ẞ (Capital sharp s)
			'u' : '\u00fa\u00f9\u00fb\u00fc\u016f', // úùûüů
			'U' : '\u00da\u00d9\u00db\u00dc\u016e' // ÚÙÛÜŮ
		},
```

JBou:
The accents in your table are already removed with the [RemoveAccent()](https://github.com/Sonarr/Sonarr/blob/d80565f6a9b045f2e8de594eda0d2ce51ea497bd/src/NzbDrone.Common/Extensions/StringExtensions.cs#L59) method.
The only problem is, that right now `ä` becomes `a`, `ü` -> `u`, ... and so on. This is not correct in german.
In German, `ä` is replaced by `ae`, `ü` by `ue`. This PR fixes it.
German releases don't use umlauts, but they use `ae`, `ue`, ...

thezoggy:
but shouldn't sonarr use ReplaceGermanUmlauts only on gemeran shows?

Issue #4877, opened 2022-01-28
Windows Long Path Support - Paths not allowed to be greater than 259 characters


Alanoll:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

At the moment on Windows, if your directory structure is deep, then portions of the EpisodeTitle are being truncated to account for the MAX_PATH of 255 from legacy Windows API.
Support was adding a couple years ago for Windows 10 1607's LongPath support, but the path limit in Sonarr isn't increasing correctly.

### Expected Behavior

Paths should be allowed to exceed 255 characters, but individual path segments are still limited to 255 characters.

### Environment

```markdown
- OS: Windows 10 21H2
- Sonarr: 3.0.6.1451
- Docker Install: No
- Using Reverse Proxy: No
```

### What branch are you running?

Develop

### Trace Logs?

I didn't capture trace because no "errors" are reported.

### Anything else?

It looks like this may be the source: https://github.com/Sonarr/Sonarr/blob/5a9df521addfa4eb00d4ca87956b97b7fabe0965/src/NzbDrone.Common/Disk/LongPathSupport.cs#L34

The generated path to check is 300+ characters, but the individual segment exceeds the 255 limit, so an exception is always thrown.

This is probably comes up with Anime series names and episode title combinations, but could also arise with multi-episode files when renamed.

Alanoll:
For others that come across this issue, the path limit could be overwritten by setting the MAX_PATH environment variable to work around the check itself.

HTPCplus:
Is there an option somewhere to truncate multi-episode titles to BlahBlah S00E01-E08 - Title 1...Title 8?
Although the way Sonarr names it works correctly within Emby and with VLC in Windows, my primary Windows viewer (MPC-HC) cannot handle the long file names.
If I rename it something simple, it plays fine.

ZackaryH8:
> For others that come across this issue, the path limit can be overwritten by setting the MAX_PATH environment variable to work around the check itself.

I believe Python does something like this when you install it to your system

markus101:
Closed via https://github.com/Sonarr/Sonarr/commit/52c6bc5549ab998ccc018d138c55f8f924eed6d3

PR #4878, opened 2022-01-28
Fix: Windows Long Path Support - Paths not allowed to be greater than 259 characters

Alanoll:
#### Database Migration
NO

#### Description
I changed the crafted string so it meets the segment limitations be still exceeded the old 255 character maximum.

#4877

### DIFF
--- a/src/NzbDrone.Common/Disk/LongPathSupport.cs
+++ b/src/NzbDrone.Common/Disk/LongPathSupport.cs
@@ -31,7 +31,10 @@ namespace NzbDrone.Common.Disk
                 {
                     try
                     {
-                        Path.GetDirectoryName($@"C:\{new string('a', 300)}\ab");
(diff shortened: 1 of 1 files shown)


thezoggy:
the test was to check if long support is supported, so now your test is to see if its under the limit to know if its supported? this seems backwards

Taloth:
No, he checks MAX_PATH properly now instead of doing a MAX_NAME test like we erroneously did.

thezoggy:
he's changing from root + dir  to root + dir + dir and decreased each dir to less than max. i guess it should still trigger because root+ is still going to push it over the limit but just seemed odd of the change of 300->254

Taloth:
It's coz on windows _with_ long path support. each path segment is still limited to 255. just the full path limit of 260 is removed.
So checking a path that's root+254+254+seperator pushes it well beyond 260. So we check if long path support is enabled, without running into the still existing 255 segment limit.

PR #4881, opened 2022-02-01
New: Option to show release group column on series list

danimart1991:
#### Database Migration
NO

#### Description
- Add Release Groups for Seasons and Series to Statistics entities.
- Include Release Groups column in Index Page (Table).
- Include Release Groups filter in Index Page (Table).

image

image

#### Todos
- [ ] Tests
- [ ] Wiki Updates

### DIFF
--- a/frontend/src/Series/Index/Table/SeriesIndexRow.js
+++ b/frontend/src/Series/Index/Table/SeriesIndexRow.js
@@ -416,7 +416,9 @@ class SeriesIndexRow extends Component {
 
             if (name === 'releaseGroups') {
               const joinedReleaseGroups = releaseGroups.join(', ');
-              const truncatedReleaseGroups = releaseGroups.slice(0, 4).join(', ');
+              const truncatedReleaseGroups = releaseGroups.length > 3 ?
+                `${releaseGroups.slice(0, 3).join(', ')}...` :
+                joinedReleaseGroups;
 
               return (
                 <VirtualTableRowCell
--- a/frontend/src/Series/Index/Table/SeriesIndexRow.js
+++ b/frontend/src/Series/Index/Table/SeriesIndexRow.js
@@ -415,7 +415,8 @@ class SeriesIndexRow extends Component {
             }
 
             if (name === 'releaseGroups') {
-              const joinedReleaseGroups = releaseGroups.sort((a, b) => a.localeCompare(b)).join(', ');
+              const joinedReleaseGroups = releaseGroups.join(', ');
+              const truncatedReleaseGroups = releaseGroups.slice(0, 4).join(', ');
 
               return (
                 <VirtualTableRowCell
@@ -423,7 +424,7 @@ class SeriesIndexRow extends Component {
                   className={styles[name]}
                 >
                   <span title={joinedReleaseGroups}>
-                    {joinedReleaseGroups}
+                    {truncatedReleaseGroups}
                   </span>
                 </VirtualTableRowCell>
               );
--- a/frontend/src/Series/Index/Table/SeriesIndexRow.js
+++ b/frontend/src/Series/Index/Table/SeriesIndexRow.js
@@ -114,6 +114,7 @@ class SeriesIndexRow extends Component {
       episodeCount,
       episodeFileCount,
       totalEpisodeCount,
+      releaseGroups,
       sizeOnDisk
     } = statistics;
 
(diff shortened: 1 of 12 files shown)


bakerboy448:
This seems like it may be an issue to scale.  In theory and it is possible to have a different release group for every single episode.

On series can have thousands of epiodes.

danimart1991:
> This seems like it may be an issue to scale. In theory and it is possible to have a different release group for every single episode.
> 
> One series can have thousands of episodes.

Maybe a solution could be order by Release Group Count for series and show the first 5-6 items with ... if there are more?

src/NzbDrone.Core/SeriesStats/SeasonStatistics.cs
```diff
@@ -54,5 +58,20 @@ public class SeasonStatistics : ResultSet
                 return previousAiring;
             }
         }
+
+        public List<string> ReleaseGroups
+        {
+            get
+            {
+                List<string> releaseGroups = null;
```
markus101, line 66: a list should never be null, `var releasegroups = new List<string>()`;

Or convert to a ternary instead:

```
return ReleaseGroupsString.IsNullOrWhiteSpace() ? new List<string>() : ReleaseGroupsString.Split('|').Distinct().ToList()`
```

(also add the sort here).

src/NzbDrone.Core/SeriesStats/SeriesStatisticsService.cs
```diff
@@ -43,7 +43,8 @@ private SeriesStatistics MapSeriesStatistics(List<SeasonStatistics> seasonStatis
                                        EpisodeFileCount = seasonStatistics.Sum(s => s.EpisodeFileCount),
                                        EpisodeCount = seasonStatistics.Sum(s => s.EpisodeCount),
                                        TotalEpisodeCount = seasonStatistics.Sum(s => s.TotalEpisodeCount),
-                                       SizeOnDisk = seasonStatistics.Sum(s => s.SizeOnDisk)
+                                       SizeOnDisk = seasonStatistics.Sum(s => s.SizeOnDisk),
+                                       ReleaseGroups = seasonStatistics.Where(s => s.ReleaseGroups != null).SelectMany(s => s.ReleaseGroups).Distinct().ToList()
```
markus101, line 47: Not returning null avoids this check for not null

frontend/src/Store/Actions/seriesActions.js
```diff
@@ -261,6 +273,31 @@ export const filterBuilderProps = [
       return tagList.sort(sortByName);
     }
   },
+  {
+    name: 'releaseGroups',
+    label: 'Release Groups',
+    type: filterBuilderTypes.ARRAY,
+    optionsSelector: function(items) {
+      const releaseGroupsList = items.reduce((acc, series) => {
```
markus101, line 281: I think no suggestions is better, otherwise this is going to be a lot of processing for large lists of series, people can freeform it.

frontend/src/Series/Index/Table/SeriesIndexRow.js
```diff
@@ -413,6 +414,21 @@ class SeriesIndexRow extends Component {
...
+            if (name === 'releaseGroups') {
+              const joinedReleaseGroups = releaseGroups.sort((a, b) => a.localeCompare(b)).join(', ');
+
+              return (
+                <VirtualTableRowCell
+                  key={name}
+                  className={styles[name]}
+                >
+                  <span title={joinedReleaseGroups}>
+                    {joinedReleaseGroups}
```
markus101, line 426: Like baker said this is going to overflow, definitely needs to truncate nicely. The title will have the full list.

markus101 (changes requested):
Untested, but a bunch of little things to address.

frontend/src/Series/Index/Table/SeriesIndexRow.js
danimart1991: I did my best here. I'm not a frontend developer instead. 😅

image

frontend/src/Series/Index/Table/SeriesIndexRow.js
```diff
@@ -413,6 +414,22 @@ class SeriesIndexRow extends Component {
               );
             }
 
+            if (name === 'releaseGroups') {
+              const joinedReleaseGroups = releaseGroups.join(', ');
+              const truncatedReleaseGroups = releaseGroups.slice(0, 4).join(', ');
```
markus101, line 419: I think if this truncates we should show an ellipsis as well and I think we'd want to cap at 3 to deal with longer release groups:

```
const truncatedReleaseGroups = releaseGroups.length > 3 ?
  `${releaseGroups.slice(0, 3).join(', ')}...` : 
  joinedReleaseGroups;
```

markus101 (changes requested):
Looking good, just one more thing to show an ellipsis then we should be good.

markus101:
Thanks!

PR #4905, opened 2022-02-17
Better upgrade flow with included tags.

JNWolff:
Better upgrade flow with included tags.
The ablility to mix and match resolutions and tags in an organized upgrade flow. 

It this example the tag is nordic, but it could be an audio preference or release group. 

Get first release ie web 1080p.
Upgrade web to webdl 720p (if someone prefers to have archive only 720p, but still wants the first release that comes out).
Upgrade webdl with new webdl with included tag "nordic".
Upgrade to bluray ONLY if prior tag is included in bluray aswell (bluray nordic).
Or go directly to bluray nordic, if webdl nordic never exists.
Upgrade bluray nordic with a new tag "nordic + dts-hd ma"
And so on.

Tags would play as big of a part as source and resolution.


bakerboy448:
- why are you trying to merge Sonarr/develop into Sonarr/main?
- All Pull Requests go to Sonarr/develop from a branch on your fork
- These are not tags. Tags have nothing to do with what is or is not grabbed by themselves.  Those are release profiles with preferred words / must contain / must not contain that impact those.  Where or what indicates that tags have any impact by themselves for what is or is not grabbed?

PR #4912, opened 2022-02-22
Fixed: SABnzbd assume develop version is 3.0.0

### LINKED ISSUE #4911: current 3.6-develop version of sabnzbd is assumed to be 1.1.0 and can't test successfully

### Current Behavior

Sonarr does not work with sabnzbd develop. When testing a connection to current sabnzb develop the following is shown:
```
Test was aborted due to an error: Cannot combine OsPaths of different platforms ('' + 'downloads/complete')
Sabnzbd develop version, assuming version 1.1.0 or higher.
```

the logs show ``

### Expected Behavior

Sonarr should assume sabnzbd develop to be at least version 2.0 and be able to connect to it

### Steps To Reproduce

* run sabnzbd develop
* configure it
* test the connection

### Environment

```markdown
- OS: Ubuntu 20.04
- Sonarr: 3.0.6.1342
- Docker Install: No, using Snap
- Using Reverse Proxy: yes
```

Main

### Trace Logs?

https://www.toptal.com/developers/hastebin/xuwidodinu.yaml

### Anything else?

The reason seems to be that Sonarr maps "develop" to version 1.1.0: https://github.com/Sonarr/Sonarr/blob/0a2b109a3fe101e260b623d0768240ef8b7a47ae/src/NzbDrone.Core/Download/Clients/Sabnzbd/Sabnzbd.cs#L338-L346

The code has a version gate of 2.0 to retrieve the correct paths for completed downloads
https://github.com/Sonarr/Sonarr/blob/0a2b109a3fe101e260b623d0768240ef8b7a47ae/src/NzbDrone.Core/Download/Clients/Sabnzbd/Sabnzbd.cs#L219-L230

Since 1.1.0 is less than 2.0 sonarr expects the download folder configuration at the wrong place.

bakerboy448:
#### Database Migration
NO

#### Description

> that code bit was back form 2017, as sab 2.x and older is defunct in various ways.. safe to assume anyone using sab develop is 3.x

_Originally posted by @thezoggy in https://github.com/Sonarr/Sonarr/issues/4911#issuecomment-1048220658_

#### Todos
- [ ] Tests
- [ ] Wiki Updates

* Fixes #4911

### DIFF
--- a/src/NzbDrone.Core/Download/Clients/Sabnzbd/Sabnzbd.cs
+++ b/src/NzbDrone.Core/Download/Clients/Sabnzbd/Sabnzbd.cs
@@ -362,7 +362,7 @@ namespace NzbDrone.Core.Download.Clients.Sabnzbd
 
                 if (rawVersion.Equals("develop", StringComparison.InvariantCultureIgnoreCase))
                 {
-                    return new NzbDroneValidationFailure("Version", "Sabnzbd develop version, assuming version 2.0.0 or higher.")
+                    return new NzbDroneValidationFailure("Version", "Sabnzbd develop version, assuming version 3.0.0 or higher.")
                     {
                         IsWarning = true,
                         DetailedDescription = "Sonarr may not be able to support new features added to SABnzbd when running develop versions."
--- a/src/NzbDrone.Core/Download/Clients/Sabnzbd/Sabnzbd.cs
+++ b/src/NzbDrone.Core/Download/Clients/Sabnzbd/Sabnzbd.cs
@@ -340,8 +340,8 @@ namespace NzbDrone.Core.Download.Clients.Sabnzbd
                     return null;
                 }
 
-                major = 3;
-                minor = 5;
+                major = 2;
+                minor = 0;
                 patch = 0;
             }
 
@@ -362,7 +362,7 @@ namespace NzbDrone.Core.Download.Clients.Sabnzbd
 
                 if (rawVersion.Equals("develop", StringComparison.InvariantCultureIgnoreCase))
                 {
-                    return new NzbDroneValidationFailure("Version", "Sabnzbd develop version, assuming version 3.5.0 or higher.")
+                    return new NzbDroneValidationFailure("Version", "Sabnzbd develop version, assuming version 2.0.0 or higher.")
                     {
                         IsWarning = true,
                         DetailedDescription = "Sonarr may not be able to support new features added to SABnzbd when running develop versions."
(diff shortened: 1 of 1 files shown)


markus101:
Given the number of people running ancient versions of Sonarr it's not a stretch to assume someone is running something equally ancient on the SABnzbd side, would it be possible to have the version appended/prepended to `develop`, something like `3.5.0-develop` or `develop-3.5.0`? We could then parse the develop version out and treat anything that is just `develop` as an safe guess. Right now if there was a feature in development we couldn't know that a particular `develop` version actually contains that functionality and to err on the safe side wait for an official release, but if it's not feasible we can make due with using a safe fallback.

Taloth:
I specifically asked baker to set it to 2.0, because it doesn't matter. Sonarr has no api checks above sabnzbd 2.0.

> Given the number of people running ancient versions of Sonarr it's not a stretch to assume someone is running something equally ancient on the SABnzbd side

Maybe, but that means that such a person runs the newest Sonarr, and a Sabnzbd develop version of years ago. They deserve to be burned.
Having a version in there would be nice, but that means the version would have to be in Sabnzbd sources rather than release pipeline. Also, neither does Sonarr. Our dev builds are 10.x and therefore always assumed newest regardless of how old the code is.

Safihre:
If I let SABnzbd report `3.5.0-develop`, it will parse it correctly even on older versions of Sonarr?
Because I used to do that (set the version like that), but stopped at some point because I didn't think it was useful. If it fixes the problem, then it's an easy fix for us.

Taloth:
> If I let SABnzbd report `3.5.0-develop`, it will parse it correctly even on older versions of Sonarr? 

Yes, we parse it with a regex that isn't anchored. So it just searches for `1.1.1` and ignores the rest. Atm it does expect 3 digits, where the 3rd can be `.x`, but `-develop` is better. So `3.5.x`, `3.5.0-develop` is fine. but `3.5-develop` isn't.
https://github.com/Sonarr/Sonarr/blob/0d99c87d8739be8fdd5891bcbd9fd4c5c05542f4/src/NzbDrone.Core/Download/Clients/Sabnzbd/Sabnzbd.cs#L36
Been like that for 5+ years, so should be fine for any version of nzbdrone/sonarr. Support for `1.1.x` 'git' version was added 6 years ago.

Safihre:
Done! Set to `3.6.0-develop`.
Still probably good to up this check to at least `3.0.0`.

markus101:
> Done! Set to `3.6.0-develop`.

Thanks!

> Still probably good to up this check to at least 3.0.0.

Yeah, I think that is a good compromise, if someone ends up being on something much older they can upgrade finally.

bakerboy448:
3.0.0 it is

Taloth:
Closed via https://github.com/Sonarr/Sonarr/commit/77412f23766e8bb553e83d87f1e599e795a67f8b

PR #4768, opened 2021-11-29
New: Import subtitles from sub folders

Bakeneko:
#### Database Migration
NO

#### Description
Improve extra files import by searching for filename match and season/episode match.

For subtitles, if no match is found and the folder contains only one episode, use any subtitle file available.

For subtitle filename parsing, ignore case for iso code search, fix language search and add support for common tags (full, forced, foreign, default, cc, psdh, sdh).

Add partial test coverage for extras.

#### Todos
- [x] Tests

#### Issues Fixed or Closed by this PR

Related to #2513

Issue #3278, opened 2019-08-26
Import multiple subtitles of the same language


Jelmerb:
**Describe the problem**
I have multiple subtitle files of the same language (English)
* `Pennyworth.S01E06.WEBRip.x264-ION10.en.srt`
* `Pennyworth.S01E06.WEBRip.x264-ION10.2.en.srt`

Only `Pennyworth.S01E06.WEBRip.x264-ION10.en.srt` is imported, but I would like to have both imported by Sonarr.

**Describe any solutions you think might work**
Check if the name of the video file is in the subtitle file and then copy/move it to the designated folder along with the video file.

Jelmerb:
`Pennyworth.S01E06.WEBRip.x264-ION10.en.srt` is a SDH subtitle and `Pennyworth.S01E06.WEBRip.x264-ION10.2.en.srt` is the normal subtitle. The download didn't add the SDH subtitle properly so there is now way to know which is which.

Jas0n99:
I just noticed that sonarr only imported my `file.en.srt` and not the `file.eng.srt`. I name my subs as follows:

- `file.en.forced.srt` = Forced
- `file.en.srt` = Non-SDH (or only one sub file)
- `file.eng.srt` = SDH

Plex doesn't recognize sdh naming so using 2 & 3 letter codes allows a consistent sort-order in the plex menus.

I couldn't find any documentation on subtitles in the wiki, just kind of gathered bits & pieces from the issues tracker.

Until there is more elaborate configuration, or even documentation, perhaps it would just make sense to copy _all_ existing files with the specified extension(s) in the "import extra files" box? Leaving it up to the user to have their own file-handling process to rename / move whatever files to the root folder with the video for the time being.

Issue #2513, opened 2018-04-17
Import subtitle files from sub/subs/subtitles folder


tinefin:
Sonarr doesn't detect subtitles in subfolders like for instance Rarbg uses.

As an English Second Language person, subtitles are necessary. Sonarr is pretty great at detecting and renaming subtitles not in subfolders including proper language extension, so I'd rather not reinvent that part.
Therefore I would like to write a post processing script to move subtitles from subfolders named [Sub / Subs / sub / subs / Subtitles / subtitles] to the same level as the episode file, so Sonarr will detect it, move it and rename it appropriately.

However, the current "Download" event looks like it is triggered _after_ the files have already been moved, which is too late. The description for "sonarr_episodefile_sourcefolder" says "Full path to the folder the episode file was imported from".
https://github.com/Sonarr/Sonarr/wiki/Custom-Post-Processing-Scripts

Would it be possible to implement an event that runs before moving files?

markus101:
Not a straight dupe because we could handle well named subs in a consistently named folder.

We'll need some examples of the folder/file layout to determine the best way to do this.

A directory listing of the completed download with the video file and any other files would be helpful, for example:

```
C:\Test\Completed\TV\Series.title.s01e01.x264-RlsGrp
C:\Test\Completed\TV\Series.title.s01e01.x264-RlsGrp\Series.title.s01e01.x264-RlsGrp.mkv
C:\Test\Completed\TV\Series.title.s01e01.x264-RlsGrp\subs\Series.title.s01e01.x264-RlsGrp.srt
```

tinefin:
@markus101 Folder layout and contents follows this pattern:

# Rarbg

### Basic form:
Completed\TV\Series.s02e04.WEBRip.x264-RlsGrp\Series.s02e04.WEBRip.x264-RlsGrp.mp4
Completed\TV\Series.s02e04.WEBRip.x264-RlsGrp\RlsGrp.txt
Completed\TV\Series.s02e04.WEBRip.x264-RlsGrp\Subs\2_Eng.srt

### Sometimes there is a .nfo file:
Completed\TV\Series.s02e04.WEBRip.x264-RlsGrp\Series.s02e04.WEBRip.x264-RlsGrp.nfo

### Sometimes there might be an extra english sub:
Completed\TV\Series.s02e04.WEBRip.x264-RlsGrp\Subs\3_Eng.srt

### Sometimes there are more languages, like:
[...]

PR #4924, opened 2022-03-01
New: Preserve language tags when importing subtitle files

lyarenei:
#### Database Migration
YES

#### Description
I noticed too late in #2513 that while the import for multiple subtitle files for same language was fixed, the language tags (forced, sdh, default, ...) were still dropped. So these changes attempt to address that; preserve the tags when renaming/importing the subtitle files.

The recognized tags are in the [language regex](https://github.com/Sonarr/Sonarr/blob/0521fc5681e845030ac17a9ebf4664792b24e27f/src/NzbDrone.Core/Parser/LanguageParser.cs#L27).
I don't know if a support for more/custom tags is needed/wanted or not, for the sake of implementation simplicity, I assumed the language regex is enough.

#### Todos
- [x] Tests
- [ ] Wiki Updates

* https://github.com/Sonarr/Sonarr/issues/2570
    - Resolves the issue itself, but does not fix the language issue discussed later

* https://github.com/Sonarr/Sonarr/issues/3278 
    - This is already fixed by changes in #4768, but relates to the language tags topic.


src/NzbDrone.Core/Extras/Subtitles/SubtitleService.cs
```diff
@@ -68,11 +68,18 @@ public override IEnumerable<ExtraFile> MoveFilesAfterRename(Series series, List<
             var subtitleFiles = _subtitleFileService.GetFilesBySeries(series.Id);
 
             var movedFiles = new List<SubtitleFile>();
+            
+            string GetSubtitleFileIdentifier(SubtitleFile subtitleFile)
+            {
+                var languageTags = LanguageParser.ParseLanguageTags(subtitleFile.RelativePath);
+                return subtitleFile.Language + subtitleFile.Extension +
+                       languageTags.Aggregate(string.Empty, (tags, tag) => tags + tag);
+            }
```
lyarenei, line 77: Maybe it would be better to store the parsed tags in the database than repeatedly parsing them from the file name?

src/NzbDrone.Core/Parser/LanguageParser.cs
```diff
@@ -146,6 +148,23 @@ public static Language ParseSubtitleLanguage(string fileName)
...
             return Language.Unknown;
         }
+        
+        public static IEnumerable<string> ParseLanguageTags(string fileName)
+        {
+            try
+            {
+                var simpleFilename = Path.GetFileNameWithoutExtension(fileName);
+                var tagMatches = SubtitleLanguageRegex.Matches(simpleFilename);
+                var languageTags = from Match tagMatch in tagMatches select tagMatch.Groups["tags"].Value.ToLower();
```
lyarenei, line 158: I got some weird issue here as the code stopped returning multiple matches. The `.Matches()` just started behaving exactly like `.Match`, only giving the last match. Yesterday, it worked just fine.

I am not aware of any changes which could change the behavior, tried to clean and rebuild Sonarr, but it's still happening. Do you have any ideas?

src/NzbDrone.Core/Extras/Subtitles/SubtitleService.cs
```diff
@@ -175,16 +183,18 @@ public override IEnumerable<ExtraFile> ImportFiles(LocalEpisode localEpisode, Ep
                 }
             }
 
-            var subtitleFiles = new List<Tuple<string, Language, string>>();
+            var subtitleFiles = new List<Tuple<string, Language, string, IEnumerable<string>>>();
```
lyarenei, line 186: I have been pondering about if the tags should be just put in the single string just after they are parsed (delimit them with . for example), but on the other hand having it as enumberable is more flexible.

Any opinion on that?

src/NzbDrone.Core/Extras/Subtitles/SubtitleService.cs
markus101: Pros and cons to both, if additional parsing is added later reparsing the filename will add more value, but not doing the same thing over and over is also valuable.

src/NzbDrone.Core/Extras/Subtitles/SubtitleService.cs
markus101: List/enumerable would be nicer to deal with, I'd also get away from the Tuple and move the return value to a class.

src/NzbDrone.Core/Parser/LanguageParser.cs
```diff
@@ -146,6 +148,23 @@ public static Language ParseSubtitleLanguage(string fileName)
...
+                var match = SubtitleLanguageRegex.Match(simpleFilename);
+                var languageTags = from Capture tag in match.Groups["tags"].Captures select tag.Value.ToLower();
+                return languageTags.Where(tag => !tag.Empty());
+            }
+            catch (Exception ex)
+            {
+                Logger.Debug(ex, "Failed parsing language tags from subtitle file: {0}", fileName);
+            }
+
+            return new string[] {};
```
markus101, line 166: `IEnumerable.Empty` (not sure the exact syntax offhand)

src/NzbDrone.Core/Parser/LanguageParser.cs
```diff
@@ -146,6 +148,23 @@ public static Language ParseSubtitleLanguage(string fileName)
...
         }
+        
+        public static IEnumerable<string> ParseLanguageTags(string fileName)
+        {
+            try
+            {
+                var simpleFilename = Path.GetFileNameWithoutExtension(fileName);
+                var match = SubtitleLanguageRegex.Match(simpleFilename);
+                var languageTags = from Capture tag in match.Groups["tags"].Captures select tag.Value.ToLower();
+                return languageTags.Where(tag => !tag.Empty());
```
markus101, line 159: Can empty tags be filtered before they're added to the enumerable?

src/NzbDrone.Core/Extras/Subtitles/SubtitleService.cs
```diff
@@ -175,16 +183,18 @@ public override IEnumerable<ExtraFile> ImportFiles(LocalEpisode localEpisode, Ep
...
                 var language = LanguageParser.ParseSubtitleLanguage(file);
                 var extension = Path.GetExtension(file);
-                subtitleFiles.Add(new Tuple<string, Language, string>(file, language, extension));
+                var languageTags = LanguageParser.ParseLanguageTags(file);
+
+                subtitleFiles.Add(new Tuple<string, Language, string, IEnumerable<string>>(file, language, extension, languageTags));
             }
 
-            var groupedSubtitleFiles = subtitleFiles.GroupBy(s => s.Item2 + s.Item3).ToList();
+            var groupedSubtitleFiles = subtitleFiles.GroupBy(s => s.Item2 + s.Item3 + s.Item4.Aggregate(string.Empty, (tags, tag) => tags + tag)).ToList();
```
markus101, line 197: This group-by is pretty rough as a Tuple, as mentioned above, convert to the return value to a class.

src/NzbDrone.Core/Extras/Subtitles/SubtitleService.cs
```diff
@@ -235,6 +246,15 @@ private string GetSuffix(Language language, int copy, bool multipleCopies = fals
                 suffixBuilder.Append(IsoLanguages.Get(language).TwoLetterCode);
             }
 
+            if (languageTags.Any())
+            {
+                foreach (var tag in languageTags)
```
markus101, line 251: I think you can shorten to:

```
suffixBuilder.Append(".");
suffixBuilder.Append(string.Join('.', languageTags));
```

markus101 (changes requested):
Haven't tested but added some comments and responded to your questions.

src/NzbDrone.Core/Extras/Subtitles/SubtitleService.cs
lyarenei: I've left that as-is, since additional parsing is more likely to be added than not, imo. It will also leave the code "prepared" - minimizing possible future changes size, should someone implement something like that.

However, I'm still open to this and don't mind implementing this the other way.

src/NzbDrone.Core/Extras/Subtitles/SubtitleService.cs
lyarenei: Changed to class - f2df494f42333ff5d57baf3c5e5d620cf045c3d8
Tags are stored as enumerable and there's a shorthand for getting them as a string.

src/NzbDrone.Core/Parser/LanguageParser.cs
lyarenei: Fixed in 6f950a9.

src/NzbDrone.Core/Parser/LanguageParser.cs
lyarenei: Yes - [b3ce4b9](https://github.com/Sonarr/Sonarr/pull/4924/commits/b3ce4b93fb91d0e3b5e8afac1c18b0f846257f64)

src/NzbDrone.Core/Extras/Subtitles/SubtitleService.cs
lyarenei: Changed to class - https://github.com/Sonarr/Sonarr/commit/f2df494f42333ff5d57baf3c5e5d620cf045c3d8

src/NzbDrone.Core/Extras/Subtitles/SubtitleService.cs
lyarenei: Indeed, refactored in [3b8ce68](https://github.com/Sonarr/Sonarr/pull/4924/commits/3b8ce68c56e17d8f09dda428c9bdc499e31e99a3)

lyarenei:
Looking at the failing build - the integration tests for ubuntu are failing - more exactly there are 44 ignored test cases due to `ignore reason: OneTimeSetUp: httpbin is bugged`.

lyarenei:
@markus101 Thanks for the review and answers. I've addressed the comments, so you can take a look again.

Since this feature now contains a database migration, are ther any additional steps I need to do? I've looked into the contribution guidelines and didn't find anything, so I'm asking just in case.

lyarenei:
Hi @markus101, can you please take a look when you'll have some time?

markus101 (changes requested):
Storing the FullPath is a non-starter, if you need to pass the full path for import purposes passing it as a separate parameter or a property that isn't stored (named `Path` to match `EpisodeFile`) would be much better.

src/NzbDrone.Core/Datastore/Migration/170_add_fields_to_subtitle_file.cs
```diff
@@ -0,0 +1,15 @@
...
+
+namespace NzbDrone.Core.Datastore.Migration
+{
+    [Migration(170)]
+    public class add_fields_to_subtitle_file : NzbDroneMigrationBase
+    {
+        protected override void MainDbUpgrade()
+        {
+            Alter.Table("SubtitleFiles").AddColumn("LanguageTags").AsString().Nullable();
+            Alter.Table("SubtitleFiles").AddColumn("FullPath").AsString().Nullable();
```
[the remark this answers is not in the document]
lyarenei: Refactored in 8c22325

src/NzbDrone.Core/Datastore/Migration/170_add_fields_to_subtitle_file.cs
```diff
@@ -0,0 +1,15 @@
+using FluentMigrator;
+using NzbDrone.Core.Datastore.Migration.Framework;
+
+namespace NzbDrone.Core.Datastore.Migration
+{
+    [Migration(170)]
+    public class add_fields_to_subtitle_file : NzbDroneMigrationBase
```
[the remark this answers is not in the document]
lyarenei: Renamed in 0fad6dc

src/NzbDrone.Core/Parser/LanguageParser.cs
```diff
@@ -152,6 +154,25 @@ public static Language ParseSubtitleLanguage(string fileName)
...
             return Language.Unknown;
         }
+        
+        public static IEnumerable<string> ParseLanguageTags(string fileName)
+        {
+            try
+            {
+                var simpleFilename = Path.GetFileNameWithoutExtension(fileName);
+                var match = SubtitleLanguageRegex.Match(simpleFilename);
+                var languageTags = from Capture tag in match.Groups["tags"].Captures
```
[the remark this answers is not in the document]
lyarenei: Rewritten in 57656c1

src/NzbDrone.Core/Extras/Subtitles/SubtitleService.cs
```diff
@@ -235,6 +245,12 @@ private string GetSuffix(Language language, int copy, bool multipleCopies = fals
                 suffixBuilder.Append(IsoLanguages.Get(language).TwoLetterCode);
             }
 
+            if (!string.IsNullOrEmpty(languageTags))
```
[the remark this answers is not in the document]
lyarenei: Cool, didn't notice that, still new to C#. Thanks, fixed in cb36976.

src/NzbDrone.Core/Extras/Subtitles/SubtitleFile.cs
```diff
@@ -1,10 +1,28 @@
...
+        {
+            LanguageTags = string.Join(".", newTags);
+        }
+
+        public void SetRelativePath(string fullPath, string seriesPath)
+        {
+            var endsWithSeparator = seriesPath.EndsWith(Path.PathSeparator.ToString());
+            var startPos = endsWithSeparator ? seriesPath.Length : seriesPath.Length + 1;
+            RelativePath = fullPath.Substring(startPos);
+        }
```
lyarenei, line 26: Was thinking about moving this to ExtraFile, but that might be better to move it there in the future if needed?

lyarenei:
> Storing the FullPath is a non-starter, if you need to pass the full path for import purposes passing it as a separate parameter or a property that isn't stored (named `Path` to match `EpisodeFile`) would be much better.

I've noticed `RelativePath` int the `ExtraFile` base class, so I ended up using that.

All comments should be resolved, please have a look again. Thanks.

src/NzbDrone.Core/Extras/Subtitles/SubtitleFile.cs
markus101: `PathExtensions.GetRelativePath()` already exists to handle this logic, `subFile.RelativePath = ...`

src/NzbDrone.Core/Extras/Subtitles/SubtitleFile.cs
```diff
@@ -1,10 +1,28 @@
...
 {
     public class SubtitleFile : ExtraFile
     {
         public Language Language { get; set; }
+
+        public string AggregateString => Language + Extension + LanguageTags;
+
+        public string LanguageTags { get; set; }
+
+        public void SetLanguageTags(IEnumerable<string> newTags)
```
markus101, line 16: Language tags can be set directly, no need for a function to set it.

src/NzbDrone.Core/Extras/Subtitles/SubtitleFile.cs
```diff
@@ -1,10 +1,28 @@
...
 
 namespace NzbDrone.Core.Extras.Subtitles
 {
     public class SubtitleFile : ExtraFile
     {
         public Language Language { get; set; }
+
+        public string AggregateString => Language + Extension + LanguageTags;
+
+        public string LanguageTags { get; set; }
```
markus101, line 14: Thinking about this a bit more, I think it'd make sense to make this an array and have the value stored in the DB as such (serialized as a JSON string), the logic to automatically serialize and deserialize already exists.

src/NzbDrone.Core/Extras/Subtitles/SubtitleFile.cs
```diff
@@ -1,10 +1,28 @@
...
+using NzbDrone.Core.Extras.Files;
 using NzbDrone.Core.Languages;
 
 namespace NzbDrone.Core.Extras.Subtitles
 {
     public class SubtitleFile : ExtraFile
     {
         public Language Language { get; set; }
+
+        public string AggregateString => Language + Extension + LanguageTags;
```
markus101, line 12: Make it `Language + LanguageTags + Extension` to match naming?

markus101 (changes requested):
Thanks for the fixes, a few more things that will make things a bit more flexible in the future.

src/NzbDrone.Core/Extras/Subtitles/SubtitleFile.cs
lyarenei: Removed in 24334c4.

src/NzbDrone.Core/Extras/Subtitles/SubtitleFile.cs
lyarenei: Refactored in 24334c4.

src/NzbDrone.Core/Extras/Subtitles/SubtitleFile.cs
lyarenei: Reordered in 49d89e2.

Qstick (changes requested):

src/NzbDrone.Core/Datastore/Migration/add_language_tags_to_subtitle_files.cs
```diff
@@ -0,0 +1,14 @@
+using FluentMigrator;
```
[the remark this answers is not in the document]
lyarenei: Oh I see, looks like I dropped it in 96a7c73e6a9b7934a5aa2c154ec06ea30e462f3c. Fixed in 139417e824521f287661c14f32c68acf0e3ed1b7.

lyarenei:
I'm not sure what to do with the failing tests. When I tried to take a look and run them locally, I got over 100 failures and all of them seem not even remotely related to these changes - for example I have ~70 failures in API tests. 

Any ideas what might be wrong?

lyarenei:
> Sorry, those tests were broken in develop and went unnoticed. If you rebase on develop they'll be resolved. As for the local tests I'm not sure why they'd be failing.

Thanks for the fix. There seems to be only one test failing, but that one passes locally, so maybe some instability at play here? Everything else seems to look good.

markus101:
Thanks for all your fixes on this!

lyarenei:
> Thanks for all your fixes on this!

@markus101 No problem. What about the wiki update, do you want to mention this change there? Should I edit the page (settings/importing) now or wait until a new release? Or is this handled by someone else?

markus101:
I think it's worth adding there with a note that it's in develop only at the moment.

lyarenei:
> I think it's worth adding there with a note that it's in develop only at the moment.

Ok, I updated the wiki page. https://wiki.servarr.com/en/sonarr/settings#importing.

PR #4940, opened 2022-03-10
New: Add optional Source Title column to history

stevietv:
#### Database Migration
NO

#### Description
I've had it a few times that a show has bulk imported, and due to mismatches between scene and TVDB they don't quite get matched up properly to an episode in Sonarr.

What I like to do here is then look at the history page and see if the releases somewhat match up with the episodes. An example here is RuPauls Drag Race Untucked. The show didn't have a season 1 (as it began as a sideshow to season 2 of the main show). Some groups released it back then as season 1 and some release it now as season 2. It can often be determined from the release title if it was the correct one as epiosde title is in there.

Looking down the history list it'd be nice to see also the release name at a glance without having to go to details. Hence this PR. Hope you find it OK.

#### Todos
- [NA] Tests
- [NA] Wiki Updates

*

### DIFF
--- a/frontend/src/Activity/History/HistoryRow.js
+++ b/frontend/src/Activity/History/HistoryRow.js
@@ -217,6 +217,16 @@ class HistoryRow extends Component {
               );
             }
 
+            if (name === 'sourceTitle') {
+              return (
+                <TableRowCell
+                  key={name}
+                >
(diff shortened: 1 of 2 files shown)


markus101:
Thanks!

PR #4961, opened 2022-03-29
Properly handle 119 Errorcode on Diskstation client

margaale:
#### Database Migration
NO

#### Description
Try to authenticate on the next request when sid not found is received.


markus101:
Thanks!

Issue #4373, opened 2021-03-11
MySQL Option for k8 users!


darkcloud784:
**Is your feature request related to a problem? Please describe.**
Yes, lots of bug reports have been opened for the sqlite database being locked due to sqlite not liking external storage/network storage.

**Describe the solution you'd like**
Implement a mysql external db option, I know this has been rejected before but it would REALLY REALLY be beneficial for the k8s community as well as others to be able to not have to worry about sqlite db corruption.

**Describe alternatives you've considered**
force local storage - issue is that those like me that run multi-node k8 cluster this is obviously not an option.
improve external storage speeds - The issue here is not the speed but the iops. I as well as many others can't spend $923740297492347 on NVME drives for my nas.
find some way to rsync the db between nodes - problem with this is that it still poses a problem with db corruption while copying the db between nodes.

**Additional context**
I completely understand that this has been rejected multiple times before but I really think it would be a benefit for not only myself  but the growing population of homelabers that use k8s as their container orchestration. This should also cut down on the bug reports for db corruption issues as well as the issues seen with the db being locked.

bakerboy448:
Duplicate of #1669
Duplicate of #4372 

Please use the search function and do not post duplicates.
Please use one of the support channels: [forums](https://forums.sonarr.tv/), [subreddit](https://www.reddit.com/r/sonarr/), [discord ](https://discord.gg/Ex7FmFK), or [IRC ](http://webchat.freenode.net/?channels=#sonarr)for questions.

acjohnson:
mysql and postgres support would be awesome in the Servarr apps. Thanks for these great tools, keep up the good work!

Issue #4175, opened 2020-12-18
Allow use of an external database 


19wolf:
**Describe the problem**
Sqlite doesn't work well or at all on network/distributed filesystems, and I think being able to use an external database like mysql or similar would solve that issue. 

**Describe any solutions you think might work**
Don't use a network filesystem

**Additional context**
None

markus101:
We do not have plans to support other database engines in the foreseeable future.

cyberjacob:
This has been a highly requested feature for several years now, what's the reasoning behind this decision?

bakerboy448:
#1669
#4372 
#4373

>We don't have plans to support another DB engine at this time, it's a significant change that would require validating all migrations on every DB engine supported (adding MySQL would only open the doors to add more) and deal with differences in various versions (which we've run into with SQLite a few times).

Issue #3764, opened 2020-05-20
Support using networked database


caewok:
When using Sonarr in k8s or docker swarm, database corruption is a significant issue because the typical solution for storage is to use NFS or GlusterFS or something similar to ensure the data is saved when the container is stopped and then started on a new node. Most of these solutions are incompatible with the sqlite setup currently in place for Sonarr. 

It would be great if the user could give Sonarr information on startup (maybe with environmental variables?) to use a networked database. This would make Sonarr much less dependent on the local storage device, and should alleviate database corruption issues. Related to #1669 and #1886.

markus101:
We understand why this is wanted for some users, but it's not something we plan to support, at least in the foreseeable future.

Issue #1669, opened 2017-02-09
MySQL support


drashna:
It would be nice to have the option to use MySQL/mariaDB/etc rather than SQLite.  

I already run mariaDB for MySQL support on a system on my network, so it would be nice/useful to be able to use that db server instead.

markus101:
We don't have plans to support another DB engine at this time, it's a significant change that would require validating all migrations on every DB engine supported (adding MySQL would only open the doors to add more) and deal with differences in various versions (which we've run into with SQLite a few times).

PR #4999, opened 2022-04-22
New: Support for Postgres database

3744111:
#### Database Migration
YES | NO

#### Description
Starts of Postgres support for Sonarr using Prowlarr/Radarr commits as a base:

https://github.com/Prowlarr/Prowlarr/pull/515
https://github.com/Prowlarr/Prowlarr/pull/633
https://github.com/Prowlarr/Prowlarr/pull/865
https://github.com/Radarr/Radarr/pull/6873

#### Todos
- [ ] Tests
- [ ] Wiki Updates

#4175 
#1669
#4373
#3764


bakerboy448:
You'll probably wanna swing by discord and chat this one out before spending too much more time and effort on it.   Previously Markus and taloth have indicated they have no plans whatsoever to support any other database besides SQLite.

3744111:
Alright, I'll swing by discord and see if we can figure something out. I did see that there are no plans for supporting other databases due to differences between version, but having implemented Postgres it really not that bad. There are about 3 places where the queries are different between SQLite and Postgres.

teambvd:
Would be grateful to see this implemented. I've gotten around some of the growing performance issues I've had with Sonarr over the last 4-6 months (15+ seconds to view download history, and again each time paging through, among others) or so by tuning the SQLite DB a bit. Maxing out the page size (along with a vacuum to take effect - vacuum alone made no noticeable impact), and matching that to the block size on the storage side brought it back down to ~3-ish seconds, much closer to 'normal'. But I fear even that might just be a bit of a band-aid.

I don't really need anything more than SQLite for Radarr, as even 10k items should be fine (was with Sonarr anyway). But for Sonarr (and probably Lidarr as well, but that's a different matter), it seems about the time I ticked over ~23-25k items, it started getting pretty sluggish even on it's mirrored NVME pool.

anthr76:
Any chance of seeing this merged anytime soon?

nwilging:
+1 to this. Recently integrated postgres with radarr and the performance boost is like a night and day difference. I am bound to using network shares for my servarr volume mounts, and so I encounter _frequent_ issues with sqlite.

With the growing popularity of containerization and cloud-based deployments, it would be great to see database support in Sonarr that isn't limited by network-mounted storage devices.

arestifo:
Any updates on this?

feld:
I've been using Radarr with Postgres and it has been noticeably faster. I hope to see Postgres support make its way into Sonarr too...

wsidl:
Seeing comments like this from @bakerboy448 on #4175:

> We don't have plans to support another DB engine at this time, it's a significant change that would require validating all migrations on every DB engine supported (adding MySQL would only open the doors to add more) and deal with differences in various versions (which we've run into with SQLite a few times).

Maybe this is my lack of familiarity with .Net, but I know in any other language, using any decent ORM with a database migration plan would resolve all of this. I deal with multiple database types on a regular basis and only use SQLite for local development specifically for all the reasons people mention above: it's faster, it's fault tolerant, it's naturally backed up.

It's for this very reason I'd be happy to find another platform to support my needs.

bakerboy448:
@wsidl postgres is on the table;  nothing beyond that for the Starrs

the Radarr comments markus noted have been resolved.  But as you can see from OPs attempt here and from the other Starr postgres migrations - it's a lot of effort rewriting effectively every migration and db query to be postgres compliant / conditionalized for postgres if needed.

wsidl:
Thanks @bakerboy448, but my point is there are tools and libraries (ORMs) that handle the database types for you. Quick search through the code base and looks like the project's using Fluent Migrator and Dapper, both of which all support all database types users in your previous issues have been requesting.

Knowing this now and my lack (and un-willingness to touch .Net), why can't we use these libraries to their full extent?

markus101:
Using an ORM doesn't remove the need to support the underlying DB engine, if we're going to support something then we're going to have an environment to test and debug it in, we're not going to do that for multiple dedicated DB engines.

onedr0p:
I agree with @markus101 ORM doesn't really buy too much here if the primary focus is sqlite with postgresql as an option, in fact it might get in the way more if it's not fully compliant with the needs of Sonarr. Postgres is by far the best RDBMS database option out there and I do not think anything else needs to be supported for an external database here.

Qstick:
Replaced by #5559

wsidl:
Don't know what kind of work is involved to make integration and e2e tests work for this app, but I've typically written automated test strategies to iterate through a series of connection strings to all database types I've wanted to support. This would then be run through a container in a CI pipeline. Since this is an OS project, you can get this functionality for free on most platforms.

From the sounds of it, doing anything in .Net is more complicated than it should be.

Issue #4778, opened 2021-12-06
UI: Allow custom text in HTML title bar


KarmicIT:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your feature request related to a problem? Please describe

No

### Describe the solution you'd like

A common question lately for the *arr's is to allow a single instance of the application to support multiple versions of a download (i.e. 1080p and 4k versions). This is not currently possible and the workaround is to run multiple instances instead, one per download type.

I'd like to request the ability to "name" an instance and have that name shown in the HTML title bar, either next to - or replacing - the existing text of "Sonarr". This would let people easily see which instance they are currently browsing without needing to remember that the Sonarr instance on port 1234 is the 4k version versus the instance on port 4321 which is for 1080p.

This feature could be added within the Settings>General>Host area, with a freeform text box that is only visible when showing advanced options is enabled.

### Describe alternatives you've considered

Raw HTML files are not available to edit and would be overwritten by any updates.

### Anything else?

This capability already exists for hard-coded menus and TV names when you browse the GUI. I.e. if you go to the Settings menu the HTML title bar shows "General Settings - Sonarr". Likewise, if you are looking at a TV show it will say "Archer (2009) - Sonarr".

This request would enable the HTML title bar to show something similar to "Indexer Settings - 1080p - Sonarr" or "Archer (2009) - All Your Base Are Belong To Us - Sonarr", etc.

Birdheezy:
I'd like to 2nd this

markus101:
This was closed via the linked PR.

PR #5025, opened 2022-05-15
New: UDP Syslog support and Instance Name

RobinDadswell:
#### Database Migration
NO

#### Description
Allows for UDP Syslogging and adds instance name so you can see what logs come from which instance when running multiple instances. Instance name is used in both syslog and within the tab; and requires Sonarr to be in the instance name somewhere

This is already in Radarr, Readarr, Lidarr, Prowlarr and Whisparr; commits have been loosely cherry-picked from there (sonarr-ified where required and updated to make a bit more logical sense)

Syslog is a config file only parameter set.

#### Todos
- [ ] Tests
- [ ] Wiki Updates

* #4778

### DIFF
--- a/frontend/src/Settings/General/HostSettings.js
+++ b/frontend/src/Settings/General/HostSettings.js
@@ -19,6 +19,7 @@ function HostSettings(props) {
     bindAddress,
     port,
     urlBase,
+    instanceName,
     enableSsl,
     sslPort,
     sslCertHash,
@@ -68,6 +69,22 @@ function HostSettings(props) {
           helpTextWarning="Requires restart to take effect"
           onChange={onInputChange}
           {...urlBase}
+        />
+      </FormGroup>
+
+      <FormGroup
+        advancedSettings={advancedSettings}
+        isAdvanced={true}
+      >
+        <FormLabel>Instance Name</FormLabel>
+
+        <FormInputGroup
+          type={inputTypes.TEXT}
+          name="instanceName"
+          helpText="Instance name in tab and for Syslog app name"
+          helpTextWarning="Requires restart to take effect"
+          onChange={onInputChange}
+          {...instanceName}
         />
       </FormGroup>
 
--- a/src/NzbDrone.Core/Configuration/ConfigFileProvider.cs
+++ b/src/NzbDrone.Core/Configuration/ConfigFileProvider.cs
@@ -39,6 +39,7 @@ namespace NzbDrone.Core.Configuration
         string SslCertHash { get; }
         string UrlBase { get; }
         string UiFolder { get; }
+        string InstanceName { get; }
         bool UpdateAutomatically { get; }
         UpdateMechanism UpdateMechanism { get; }
         string UpdateScriptPath { get; }
@@ -205,6 +206,19 @@ namespace NzbDrone.Core.Configuration
         // public string UiFolder => GetValue("UiFolder", "UI", false);GetValue("UiFolder", "UI", false);
         public string UiFolder => "UI";
 
+        public string InstanceName
+        {
+            get
+            {
+                var instanceName = GetValue("InstanceName", BuildInfo.AppName);
+
+                if (instanceName.StartsWith(BuildInfo.AppName) || instanceName.EndsWith(BuildInfo.AppName) )
+                {
+                    return instanceName;
+                }
+                return BuildInfo.AppName;
+            }
+        }
 
         public bool UpdateAutomatically => GetValueBoolean("UpdateAutomatically", false, false);
 
@@ -217,7 +231,6 @@ namespace NzbDrone.Core.Configuration
         public int SyslogPort => GetValueInt("SyslogPort", 514, persist: false);
 
         public string SyslogLevel => GetValue("SyslogLevel", LogLevel, persist: false).ToLowerInvariant();
-
 
         public int GetValueInt(string key, int defaultValue, bool persist = true)
         {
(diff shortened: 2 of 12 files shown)


src/Sonarr.Api.V3/Config/HostConfigModule.cs
```diff
@@ -38,6 +38,7 @@ public HostConfigModule(IConfigFileProvider configFileProvider, IConfigService c
             SharedValidator.RuleFor(c => c.Port).ValidPort();
 
             SharedValidator.RuleFor(c => c.UrlBase).ValidUrlBase();
+            SharedValidator.RuleFor(c => c.InstanceName).ContainsSonarr().When(c => c.InstanceName.IsNotNullOrWhiteSpace());
```
markus101, line 41: This shouldn't be conditional on it not being null, otherwise the value can be set to nothing and the UI now shows nothing or `Page Title - `.

Removing correctly validates that the name isn't null/empty and contains Sonarr, which is the expected outcome.

markus101 (changes requested):
Just the one issue with validation that needs to be resolved.

Taloth:
Question regarding syslog. Why not leave it to journalctl to forward to syslog?

RobinDadswell:
> Question regarding syslog. Why not leave it to journalctl to forward to syslog?

That's fine if being run on Linux, what about if its being run on Windows and wants syslogging. Or if it's on a seedbox without access to configure journalctl or rsyslog or similar and you want centralised logging back to your own syslog server etc.

RobinDadswell:
This also allows for syslog to be trace but files to be info as an example

src/Sonarr.Api.V3/Config/HostConfigModule.cs
RobinDadswell: Sorted, thanks for picking that one up!

Taloth:
> > Isn't it better to specify something like InstanceNameSuffix? So instead of requiring the user to specify 'Sonarr', you just append it. like "{title} - Sonarr {InstanceName}"
> 
> I would suggest that is not as encompassing; prime example is people who would like "{title} - 4k Sonarr" and "{title} - Low Res Sonarr". Contained within the string makes more sense as it is more flexible

Flexibility isn't always good, after all, Sonarr is the brand name.
I'm not opposed to the current impl, but I think it's unnecessarily complex. On top of that users are gonna be confronted with a warning about it having to contain 'Sonarr'.
Let's see what markus thinks.

markus101:
When we talked about being able to change the name before Qstick and I weren’t sure whether it should be a suffix or prefix made the most sense. From a branding perspective I think `Sonarr X` makes the most sense, but I wasn’t exactly sure how to communicate that clearly and haven’t thought about it much since.

A few thoughts:


- Either enforcing it at the beginning or applying it ourselves seems reasonable
- Applying it ourselves would also avoid the existing issue where currently if it’s set in the config file (bypassing the validation) it can be set to whatever
- We could go with a hybrid approach where it either starts with Sonarr or we append it
- Since it’s the app name I like the idea of the user including it, otherwise there is a gap where they don’t want to include Sonarr in the app name because we’re including it, but they want to include it because that’s how it shows up in syslogs
- From a branding POV the regex should be case sensitive


TLDR, I think I’m okay with the user setting it somewhere in the app name (this allows a bit of flexibility for the user to choose `4K Sonarr` vs `Sonarr 4K`, but maybe we change it to must be at the beginning or end specifically and require it to be title case.

I’d also take it a step further to inject it at the start if someone attempts to bypass the validation.

frontend/src/Settings/General/HostSettings.js
```diff
@@ -71,6 +72,22 @@ function HostSettings(props) {
...
+        advancedSettings={advancedSettings}
+        isAdvanced={true}
+      >
+        <FormLabel>Instance Name</FormLabel>
+
+        <FormInputGroup
+          type={inputTypes.TEXT}
+          name="instanceName"
+          helpText="Instance name in tab and for Syslog app name"
+          helpTextWarning="Requires restart to take effect"
```
markus101, line 85: Make sure this prompts the user to restart, currently it doesn’t, but reload applies it immediately, which feels a bit weird.

RobinDadswell:
I would agree with case sensitive; I will get that in now. I can also see the logic of having it either start with or end with it as opposed to being in the middle so happy to put that in. 

I would agree that I think the user should definitely have to put the name in themselves as it affects both the frontend and the syslog functionality. 

@Qstick do you have any issues with the change to the branding here outline by Markus as I will filter them downstream as well :D

Qstick:
No issues from me, we knew about the gap in being able to edit direct in config and needing to patch that up. I'm fine with the other stuff, my biggest thing is just keeping the App name requirements. However we feel is the best approach for that, I'm down with.

src/NzbDrone.Core/Configuration/ConfigFileProvider.cs
```diff
@@ -202,14 +206,33 @@ public string UrlBase
...
+            {
+                var instanceName = GetValue("InstanceName", BuildInfo.AppName);
+
+                if (instanceName.StartsWith(BuildInfo.AppName) || instanceName.EndsWith(BuildInfo.AppName) )
+                {
+                    return instanceName;
+                }
+                return BuildInfo.AppName;
+            }
+        }
```
RobinDadswell, line 221: @markus101 this is to stop it being worked around from not having Sonarr in the InstanceName

PR #5027, opened 2022-05-16
Added an option for minimum release score.

bejhan:
#### Database Migration
YES

#### Description
Add an option to only grab releases above a minimum score.

Radarr has a similar feature and I find it very useful as it facilitates enforcing of a whitelist and/or blacklist.
I realize that "must contain" and "must not contains" already serve as a whitelist and blacklist but this approach allows leveraging the "preferred" list rather than having to duplicate everything. Plus the "preferred" list allows for much finer control and more complex rules.

#### Todos
- [ ] Tests
- [ ] Wiki Updates

*


markus101:
As discussed on Discord, we're moving to Custom Formats in v4 and not going to land this in v3 before that releases.

PR #5029, opened 2022-05-18
Added language "Brazilian_Portuguese" to Language Profiles

r4f4w:
#### Database Migration
NO

#### Description
Added language "Brazilian_Portuguese" to Language Profiles.

#### Todos
- [ X ] Tests
- [ X ] Wiki Updates (Not needed)

*

### DIFF
--- a/src/NzbDrone.Core.Test/Languages/LanguageFixture.cs
+++ b/src/NzbDrone.Core.Test/Languages/LanguageFixture.cs
@@ -35,7 +35,7 @@ namespace NzbDrone.Core.Test.Languages
                 new object[] {20, Language.Greek},
                 new object[] {21, Language.Korean},
                 new object[] {22, Language.Hungarian},
-                new object[] {23, Language.Brazilian_Portuguese}
+                new object[] {23, Language.PortugueseBR}
             };
 
         public static object[] ToIntCases =
@@ -62,7 +62,7 @@ namespace NzbDrone.Core.Test.Languages
                 new object[] {Language.Greek, 20},
                 new object[] {Language.Korean, 21},
                 new object[] {Language.Hungarian, 22},
-                new object[] {Language.Brazilian_Portuguese, 23}
+                new object[] {Language.PortugueseBR, 23}
             };
 
         [Test, TestCaseSource("FromIntCases")]
--- a/src/NzbDrone.Core.Test/Languages/LanguageFixture.cs
+++ b/src/NzbDrone.Core.Test/Languages/LanguageFixture.cs
@@ -30,11 +30,12 @@ namespace NzbDrone.Core.Test.Languages
                 new object[] {15, Language.Norwegian},
                 new object[] {16, Language.Finnish},
                 new object[] {17, Language.Turkish},
-                new object[] {18, Language.Portuguese},
+                new object[] {18, Language.Portuguese},        
                 new object[] {19, Language.Flemish},
                 new object[] {20, Language.Greek},
                 new object[] {21, Language.Korean},
-                new object[] {22, Language.Hungarian}
+                new object[] {22, Language.Hungarian},
+                new object[] {23, Language.Brazilian_Portuguese}
             };
 
         public static object[] ToIntCases =
@@ -60,7 +61,8 @@ namespace NzbDrone.Core.Test.Languages
                 new object[] {Language.Flemish, 19},
                 new object[] {Language.Greek, 20},
                 new object[] {Language.Korean, 21},
-                new object[] {Language.Hungarian, 22}
+                new object[] {Language.Hungarian, 22},
+                new object[] {Language.Brazilian_Portuguese, 23}
             };
 
         [Test, TestCaseSource("FromIntCases")]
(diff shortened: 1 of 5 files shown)


src/NzbDrone.Core/Parser/IsoLanguages.cs
```diff
@@ -27,6 +27,7 @@ public static class IsoLanguages
                                                                new IsoLanguage("fi", "fin", Language.Finnish),
                                                                new IsoLanguage("tr", "tur", Language.Turkish),
                                                                new IsoLanguage("pt", "por", Language.Portuguese),
+                                                               new IsoLanguage("br", "bra", Language.Brazilian_Portuguese),
```
Taloth, line 30: This list is ISO 639-2 language codes, not country codes. 'bra' is the language code of Braj Bhāshā, a Western Hindi language.

bakerboy448:
Id suggest being consistent with how radarr has it 

https://github.com/Radarr/Radarr/commit/3ab9af06636f7d3986c77359231643ffcdda11ab

src/NzbDrone.Core/Parser/IsoLanguages.cs
r4f4w: From what I've searched, there isn't a ISO 639-2 for Brazilian Portuguese:
image

So we just use the same code as "Portuguese"?

r4f4w:
> Id suggest being consistent with how radarr has it
> 
> [Radarr/Radarr@3ab9af0](https://github.com/Radarr/Radarr/commit/3ab9af06636f7d3986c77359231643ffcdda11ab)

Yeah I saw that Radarr already had a Id for Brazilian portuguese. I'm going to change to be equal in both!

src/NzbDrone.Core/Parser/IsoLanguages.cs
r4f4w: And also, in this snippet of the code, is specifying that the ISO 639-1 code will only have two digits, 
```
if (isoCode.Length == 2)
            {
                //Lookup ISO639-1 code
                return All.SingleOrDefault(l => l.TwoLetterCode == isoCode);
            }
```

But from what I've searched, Brazilian Portuguese ISO 639-1 code is "pt-BR", as "br" is the code for Breton. 

So what is the approach here?

r4f4w:
> ID 30 which is what Radarr used for Brazilian Portuguese is already in use by Ukrainian in Sonarr.

I don't think that's an issue since Ukrainian in Radarr is 32 and in Sonarr is 30.

Sure, it would be better if they all shared the same ID's, but we're talking about changing all the other ID's in both projects. Maybe something to do in another Pull Request.

bakerboy448:
Language ID #s are unique to each project and need not be consistent

src/NzbDrone.Core/Parser/IsoLanguages.cs
bakerboy448: Radarr had to make various other changes as noted in that commit to support dual language codes

r4f4w: > Radarr had to make various other changes as noted in that commit to support dual language codes

How bad would it be if we just used the same ISO code for both Portuguese and Brazilian Portuguese?

It wasn't my intention to change the entire parse ISO logic of the project.

src/NzbDrone.Core.Test/Languages/LanguageFixture.cs
```diff
@@ -30,11 +30,12 @@ public class LanguageFixture : CoreTest
                 new object[] {15, Language.Norwegian},
                 new object[] {16, Language.Finnish},
                 new object[] {17, Language.Turkish},
-                new object[] {18, Language.Portuguese},
+                new object[] {18, Language.Portuguese},
```
markus101, line 33: Extra spaces added here

src/NzbDrone.Core/Parser/IsoLanguages.cs
markus101: > pt is a [language code](https://en.wikipedia.org/wiki/Language_code) for Portuguese, defined by [ISO standards](https://en.wikipedia.org/wiki/ISO_standard) (see [ISO 639-1](https://en.wikipedia.org/wiki/ISO_639-1) and [ISO 3166-1 alpha-2](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2)). There is no ISO code for spoken or written Brazilian Portuguese.

> pt-BR is a [language code](https://en.wikipedia.org/wiki/Language_code) for the Brazilian Portuguese, defined by [Internet standards](https://en.wikipedia.org/wiki/Internet_standard) (see [IETF language tag](https://en.wikipedia.org/wiki/IETF_language_tag)).

Source: https://en.wikipedia.org/wiki/Brazilian_Portuguese

Looks like there aren't separate language codes for Brazilian Portuguese.

> How bad would it be if we just used the same ISO 639-3 code for both Portuguese and Brazilian Portuguese? We can use "br" and "por".

You can't just use whatever you want there, `br` is used for Breton, even if it's not in Sonarr you can't just usurp it to make it work.

To do this right, you're going to need to do what Radarr did, there isn't a shortcut here that will work.

markus101:
Portuguese (Brazil) landed in https://github.com/Sonarr/Sonarr/pull/5311

PR #5047, opened 2022-06-06
Fixed: Validate if equals or child for startup folder

Qstick:
#### Database Migration
NO

#### Description
Currently startup folder validation on checks if folder settings is child, this change also checks to make sure it's not equal. This change is related to root folder and recycle bin settings.


bakerboy448:
Downstream identifying this bug fixed: https://github.com/Radarr/Radarr/issues/7329

PR #5077, opened 2022-07-10
Fixed: Improve moving file to location where another one exists

beschoenen:
#### Database Migration
NO

#### Description
This is a simpler mitigation for [this issue](https://github.com/Sonarr/Sonarr/pull/4160).

#### Todos
- [ ] Tests
- [x] Wiki Updates

### DIFF
--- a/src/NzbDrone.Common/Disk/FileAlreadyExistsException.cs
+++ b/src/NzbDrone.Common/Disk/FileAlreadyExistsException.cs
@@ -4,11 +4,11 @@ namespace NzbDrone.Common.Disk
 {
     public class FileAlreadyExistsException : Exception
     {
-        public string FilePath { get; set; }
+        public string Filename { get; set; }
 
-        public FileAlreadyExistsException(string message, string filepath) : base(message)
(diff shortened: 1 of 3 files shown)


src/NzbDrone.Common/Disk/FileAlreadyExistsException.cs
```diff
@@ -0,0 +1,14 @@
+using System;
+
+namespace NzbDrone.Common.Disk
+{
+    public class FileAlreadyExistsException : Exception
+    {
+        public string FilePath { get; set; }
```
markus101, line 7: I think it makes sense to match the naming from `SameFilenameException`, `Filename` instead of `FilePath`

src/NzbDrone.Common/Disk/FileAlreadyExistsException.cs
```diff
@@ -0,0 +1,14 @@
...
+
+        public FileAlreadyExistsException(string message, string filepath) : base(message)
```
markus101, line 9: Same here, `filename` instead of `filepath`

markus101 (changes requested):

PR #5092, opened 2022-07-26
Send all environment variable with Test

lucianor:
Update to send all environment variables with their name into the custom script test. Helpful for debugging and creation of scripts.

#### Database Migration
NO

#### Description
Send all environment variables when clicking on the test button. This will help a lot with scripts creation and debugging.

#### Todos
Not applicable

Not applicable


bakerboy448:
These are already listed in the docs
https://wiki.servarr.com/sonarr/custom-scripts

This PR does not add anything meaningful and due to the fake values means users that rely on the test payload would need to refactor their scripts.

Not all triggers have the same payloads as well.

this would also have to be updated in addition  to the docs when changes are made....

lots of work for no gain and no benefit.

lucianor:
I beg to differ, but not my call.

PR #5131, opened 2022-08-23
Fixed: Repack Preference Ignored

bakerboy448:
#### Database Migration
NO

#### Description
fixes repack spec not using config repack/proper preference

#### Todos
- [x] Tests
- Wiki Updates

*

### DIFF
--- a/src/NzbDrone.Core.Test/DecisionEngineTests/RepackSpecificationFixture.cs
+++ b/src/NzbDrone.Core.Test/DecisionEngineTests/RepackSpecificationFixture.cs
@@ -1,9 +1,11 @@
+using System;
 using System.Collections.Generic;
 using System.Linq;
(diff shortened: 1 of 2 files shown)


Qstick (changes requested):
Please add a test case for this, similar to what we have in the ProperSpecFixture. 

<https://github.com/Sonarr/Sonarr/blob/develop/src/NzbDrone.Core.Test/DecisionEngineTests/RepackSpecificationFixture.cs>

bakerboy448:
> Please add a test case for this, similar to what we have in the ProperSpecFixture.
> 
> File is: https://github.com/Sonarr/Sonarr/blob/develop/src/NzbDrone.Core.Test/DecisionEngineTests/RepackSpecificationFixture.cs

tests added
also fixed missing repack ignoring do not upgrade as well

Issue #861, opened 2015-10-18
Release wrongly classified as foreign due to ep title


thezoggy:
release wrongly classified as foreign due to ep title:

![sonarr147](http://zoggy.net/sonarr147.png)

need to update regex/eliminate ep title from that decision.. (use the tvdb data you have. cross reference release regex data)

CatmanIX:
That's actually kinda funny.

CatmanIX:
Also does this when multiple languages are listed, even when english is one of them.

thezoggy:
if any of the blacklisted words are found.. then yes. just because another is in there doesnt exclude it. sonarr knows the ep info.. thus it knows what the ep title is.. so easily should be able to eliminate false positives in that part of the naming.

thezoggy:
just got hit again with this:
`Anthony.Bourdain.Parts.Unknown.S07E03.The.Greek.Islands.HDTV.x264-MiNDTHEGAP`
result is: 'English is wanted, but found Greek'

l3uddz:
Encountered this issue with last nights episode of Below Deck Mediterranean. It was rejected for not being an english release. 

Release name:
Below.Deck.Mediterranean.S01E13.That.Was.Very.Greek.of.Us.HDTV.x264-CRiMSON

krin-san:
I have slightly different issue with Sonarr ignoring RSS-Feed from russian tracker http://www.lostfilm.tv because of incorrect language mapping.

From debug logs below I see release was rejected by 2 reasons: language and quality. I can deal with quality but I can't do anything with language mapping.

RSS feed (part):
```xml
<rss version="0.91">
<channel>
<title>LostFilm.TV</title>
<description>Свежачок от LostFilm.TV</description>
<link>https://www.lostfilm.tv/</link>
<lastBuildDate>Tue, 03 Oct 2017 15:39:29 +0000</lastBuildDate>
<language>ru</language>
<item>
	<title>Бойтесь ходячих мертвецов (Fear the Walking Dead). Эта земля — твоя земля (S03E13) [1080p]</title>
	<category>[1080p]</category>
	<pubDate>Tue, 03 Oct 2017 15:37:46 +0000</pubDate>
	<link>http://tracktor.in/rssdownloader.php?id=24772</link>
</item>
...
</channel>
</rss>
```
Debug log:
```
17-10-3 19:09:01.0|Debug|Parser|Parsing string '   (Fear the Walking Dead).      (S03E13) [1080p]'
[...]

PR #5132, opened 2022-08-23
Improve handling of multiple languages in post title

Alxandr:
#### Database Migration
NO

#### Description
This improves handling of posts/titles with multiple languages in the title. For instance, I had the following example:

```
[Erai-raws] Overlord IV - 08 [1080p][Multiple Subtitle] [ENG][POR-BR][SPA-LA][RUS]
```

Sonarr concluded this was a Russian release (only) - and as shuch did not choose to download it. This PR fixes that without breaking any of the other language-parsing tests (though I had to change one of the tests because it was order dependent).

#### Todos
- [x] Tests
- [ ] Wiki Updates

I'm not sure if this warrants a wiki update?

I didn't find a github issue for this exact problem. I can make one if wanted - but I wanted to check whether this was something I could easily fix myself first, and it didn't seem like a big problem to resolve.

### DIFF
--- a/src/NzbDrone.Core/Parser/LanguageParser.cs
+++ b/src/NzbDrone.Core/Parser/LanguageParser.cs
@@ -19,7 +19,7 @@ namespace NzbDrone.Core.Parser
                 new RegexReplace(@".*?[_. ](S\d{2}(?:E\d{2,4})*[_. ].*)", "$1", RegexOptions.Compiled | RegexOptions.IgnoreCase)
             };
 
-        private static readonly Regex LanguageRegex = new Regex(@"(?:\W|_)(?<italian>\b(?:ita|italian)\b)|(?<german>german\b|videomann|ger[. ]dub)|(?<flemish>flemish)|(?<greek>greek)|(?<french>(?:\W|_)(?:FR|VF|VF2|VFF|VFQ|TRUEFRENCH)(?:\W|_))|(?<russian>\brus\b)|(?<hungarian>\b(?:HUNDUB|HUN)\b)|(?<hebrew>\bHebDub\b)|(?<polish>\b(?:PL\W?DUB|DUB\W?PL|LEK\W?PL|PL\W?LEK)\b)|(?<chinese>\[(?:CH[ST]|BIG5|GB)\]|简|繁|字幕)|(?<bulgarian>\bbgaudio\b)|(?<spanish>\b(?:español|castellano)\b)|(?<ukrainian>\b(?:ukr)\b)",
+        private static readonly Regex LanguageRegex = new Regex(@"(?:\W|_)(?<english>\b(?:eng|english)\b)|(?<italian>\b(?:ita|italian)\b)|(?<german>german\b|videomann|ger[. ]dub)|(?<flemish>flemish)|(?<greek>greek)|(?<french>(?:\W|_)(?:FR|VF|VF2|VFF|VFQ|TRUEFRENCH)(?:\W|_))|(?<russian>\brus\b)|(?<hungarian>\b(?:HUNDUB|HUN)\b)|(?<hebrew>\bHebDub\b)|(?<polish>\b(?:PL\W?DUB|DUB\W?PL|LEK\W?PL|PL\W?LEK)\b)|(?<chinese>\[(?:CH[ST]|BIG5|GB)\]|简|繁|字幕)|(?<bulgarian>\bbgaudio\b)|(?<spanish>\b(?:español|castellano|spa-la)\b)|(?<portoguese>\b(?:portoguese|por-br)\b)|(?<ukrainian>\b(?:ukr)\b)",
                                                                 RegexOptions.IgnoreCase | RegexOptions.Compiled);
 
         private static readonly Regex CaseSensitiveLanguageRegex = new Regex(@"(?:(?i)(?<!SUB[\W|_|^]))(?:(?<lithuanian>\bLT\b)|(?<czech>\bCZ\b)|(?<polish>\bPL\b)|(?<bulgarian>\bBG\b))(?:(?i)(?![\W|_|^]SUB))",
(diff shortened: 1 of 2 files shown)


src/NzbDrone.Core/Parser/LanguageParser.cs
```diff
@@ -235,99 +235,111 @@ private static List<Language> RegexLanguage(string title)
...
+                {
+                    languages.Add(Language.Ukrainian);
+                }
+
+                if (match.Groups["spanish"].Success)
+                {
+                    languages.Add(Language.Spanish);
+                }
+
+                if (match.Groups["portoguese"].Success)
```
bakerboy448, line 339: portuguese

bakerboy448 (changes requested):
how's this display in the UI when multiple languages are detected?

src/NzbDrone.Core/Parser/LanguageParser.cs
Alxandr: Good catch - I'll fix it tomorrow morning.

Alxandr:
> how's this display in the UI when multiple languages are detected?

I have no idea. I didn't touch the GUI. I just noticed that the method for parsing languages returned a list, so I assumed the rest of the program handled it as a list. I'm running sonnar in docker on a NAS, so I can't just fire it up on the dev-computer. I can try spinning it up and just add the same rss and series and see what happens.

manubell:
I don't think this is correct.
All those languages listed are _subtitles_ not the actual spoken language. 

This group is also the only one doing this subtitle listing nonsense. They got criticized heavily and decided to keep going anyway.
One of the criticisms being the confusion over language and subtitle that we have here.

SPA-LA in the example is not even valid by ISO standards. There's going to be a lot more of these 'made up' language codes that the regex wouldn't catch.

Alxandr:
I *think* it's actually technically valid - but I don't think it means what they think it means. Doing `new Intl.Locale('SPA-LA')` in my browser gives me "Spanish as spoken in Laos" :P. That being said, your worry about the regexes not matching everything is definitely correct - and I see a couple of possible options there. In this PR I tried to change as little as possible just to get the specific cases I have working, but a more robust thing would probably be to match any 2/3 letter word with some sort of non-space delimeter on both sides, optionally followed by `-whatever`.

With regards to all those languages being subtitles, that is indeed correct - however, that is also the part I care about when talking about "language" in series. All the series I use Sonarr to monitor are always in the original spoken language, and I just want the releases with the English subs. Whether or not they include Russian subs I don't care about.

Sonarr also has no problem importing these files once I download them manually. What I do currently is just paste the magnet link (manually) into my torrent client, and assign the tv-sonarr category, and then Sonarr happily imports the show as "English" once it's done (even though it's Japanese).

It's definitely a bit of a tricky situation - but I do believe that this PR would improve the situation somewhat at least - and then whether or not the regex should be expanded/rewritten to handle more general locale patterns can be looked more into later.

bakerboy448:
Language of an episode in Sonarr is audio based not subtitle based.

Same reason VOSTFR is English not French

Alxandr:
It's still equally wrong though given that just about everything I monitor is japanese in spoken language yet Sonarr thinks it's English.

bakerboy448:
> It's still equally wrong though given that just about everything I monitor is japanese in spoken language yet Sonarr thinks it's English.

V4 uses audio language from the file, so that should be a non-issue.
https://github.com/Sonarr/Sonarr/blob/develop/src/NzbDrone.Core/MediaFiles/EpisodeImport/Aggregation/Aggregators/Augmenters/Language/AugmentLanguageFromMediaInfo.cs

as far as this PR, given it is attempting to parse subtitle languages to be treated as the file audio language....that makes no sense and IMO this shouldn't be added. Especially if it's to solve a single group actively refusing to follow any standards.

Alxandr:
This runs before any file is downloaded however - so there is no file to analyze the audio from. It's running as part of the RSS parsing to determine whether or not the file should be downloaded.

bakerboy448:
It does not only run for grabs. It'd be running on the file name as well.

And again, subtitle language(s) is not what sonarr should be treating as a language for the release.   If this PR were to be merged then VOSTFR should be treated as French lang.  since that's been rejected multiple times, I wouldn't see this being merged.

src/NzbDrone.Core/Parser/LanguageParser.cs
Alxandr: This should be fixed now.

markus101:
In this case all the "languages" are subtitle languages? If that's the case we don't want to treat subtitle languages as the language, which is a bigger problem than this PR is addressing and not sure there is really a solution to it at the moment.

At the moment how does this appear in the UI?

Alxandr:
The languages in the title is subtitle languages, yes - but they are also why sonarr is rejecting the release. I've not run the UI at all with this patch, as I started working on a RSS proxy-rewriter instead as I got the feeling that this is currently working as intended, and sonarr simply will not suport this pattern.

acg5159:
Hope this helps,
See related Issue: https://github.com/Sonarr/Sonarr/issues/861#issuecomment-1278327503

This patch Seems to Fix a Few issues, maybe they can be split up: 
- This code only chooses one language when parsing the File Name when it should handle multiple as that seems to be why it returns a List Item.
- The code does not include "English" is regex, so it is never chosen, if another language is parsed. 
  - Missing many language codes from regex. Maybe should use a DB built from ISO standard: https://www.loc.gov/standards/iso639-2/php/code_list.php

Possible Inclusions: 
- We should Be able to supply a Default Language per Series (e.g. Anime = Japanese; American TV = English), so if no code found Default to X (I think now this is a system level or just always English).
- A way to disable language parser of Filename for some cases (maybe from some torrent Sites like nyaa)

bakerboy448:
> * Missing many language codes from regex. Maybe should use a DB built from ISO standard:

Groups do not use ISO language codes for indicating the language of the release nor is that standard.

language codes and the language parsing is based on how releases are named.

markus101:
@Qstick landed multi-language support a few months ago in v4, if anything else is needed after that please open a separate PR.

PR #5135, opened 2022-08-26
Fixes Systemd Restart=on-failure endless restarts

Antorell:
#### Database Migration
NO

#### Description
Enforce a limit to how many times the service can restart in case of failure. 
 
When Sonarr is restarted from the WebUI as Sonar's UI uses an internal killall->start mechanic instead of using Systemd to restart Sonarr. If the Sonarr's Systemd Service is started manually afterward the service will fail since there's already an instance of Sonarr running on the desired **ip:port**, the sonarr.service will fail and trigger the service "Restart=on-failure" command. And since there's no limit set in the sonarr.service Systemd Service to limit how many times the server should restart, Systemd will try to restart the service indefinitely. 

I had two instances where the Sonarr's systemd service hit Systemd internal limit (5 restarts in 10 seconds), however, for all my other tries Systemd just keep going on trying to restart the service.

It's easy to reproduce the issue, start the sonarr Systemd Service, restart Sonarr from the webUI, restart sonarr's systemd service manually "systemctl start sonarr" and watch Systemd tirelessly trying to restart the service using "systemctl status sonarr" or "journalctl -xe -u sonarr". Otherwise I can forward journalctl logs privately. 

The same issue occurs with the Sonarr.service Systemd service that comes with the APT source install .deb file.  

StartLimitIntervalSec=interval
StartLimitBurst=burst
https://www.freedesktop.org/software/systemd/man/systemd.unit.html#StartLimitIntervalSec=interval

#### Todos
- [ ] Tests
- [ ] Wiki Updates

*

### DIFF
--- a/distribution/debian/sonarr.service
+++ b/distribution/debian/sonarr.service
@@ -3,6 +3,8 @@
 # Or use systemd built-in override functionality using 'systemctl edit sonarr'
 [Unit]
 Description=Sonarr Daemon
+StartLimitIntervalSec=60
+StartLimitBurst=3
 After=network.target
 
 [Service]


Taloth:
Rather the webui restart feature must be fixed to send the appropriate signals/exit code.

Antorell:
Good to know, it kind of silly to use systemd then use another mechanism to restart Sonarr on linux as it messes up Systemd.

Taloth:
Originally we didn't even provide a .service unit file, so it was a bit of a mess. and on ubuntu we used upstart. It was a bit... messy with the built-in updater.
Also why it must be =on-failure specifically instead of 'always', so 'shutdown' can return a zero exitcode. But the webui logic still needs to be adjusted to recognize it's running in systemd, and if so,  do a exitcode=0 for shutdown and exitcode=1 for restart. or something similar.

Antorell:
I deleted my previous post. I don't think I understood correctly. The ui send the correct exit signal as it doesn't trigger Systemd restart on failure, the UI needs to use systemd. Any other mechanism that doesn't rely on systemd will mess up systemd. 

This only happens when the Systemd service is down and the UI sends the correct/clean termination signal which doesn't trigger systemd's own restart mechanism. 

I had rewrite my systemd like this to work around the UI current restart mechanism

```
[Service]
User=Bob
Group=bob
UMask=0002
KillMode=process
Type=simple
ExecStartPre=-bash -tc 'kill -s 15 $(pidof -s mono)'
ExecStart=/usr/bin/mono --debug /usr/lib/sonarr/bin/Sonarr.exe -nobrowser -data=/var/lib/sonarr
RestartForceExitStatus=0
TimeoutStopSec=20

Restart=on-failure
```

Antorell:
To fix it either use a dirty kill signal or use Systemd own restart mechanism on Linux distro that rely on system. An option in General to enable it maybe since there are other distros that don't use Systemd.

bakerboy448:
> Rather the webui restart feature must be fixed to send the appropriate signals/exit code.

Fixed downstream and maybe already in v4 already?

can't find the commit unfortunately

Antorell:
I'm not sure we're talking about the same restart loop. The restart loop for which I pushed the pull request happens after the UI killed (gracefully) and restarted its own instance of Sonarr. It doesn't happen when you restart from the UI, it only happens if somebody or something manually start Sonarr systemd service later because it can't find the service to the configured port there's already an instance(the one that was launched by the UI) of Sonarr running on that port. 

If you only use Systemd to start Sonarr at boot and use your own mechanism to restart the process in the WebUI, use an unregistered service with the oneshot Type  without the restart on failure and a [systemd .timer ](https://www.freedesktop.org/software/systemd/man/systemd.timer) with OnBootSec= to start Sonarr on boot, or just remove the whole Systemd service from the install and simply use Cron instead. It's only going to cause you more headaches down the line if you still use the Simple Type= with a registered (systemctl enable service.service) service in v4.



> > Rather the webui restart feature must be fixed to send the appropriate signals/exit code.
> 
> Fixed downstream and maybe already in v4 already?
> 
> can't find the commit unfortunately
> 
> edit:
> 
> the endless restart is fixed in v4 (exit 0), but the systemd service is still killed.
> 
> ```
> Version
> 4.0.0.104
> ```
> 
>     * restart via UI
> 
>     * `on-failure` on systemd
> 
>     * No Restart Loop; however, the service remains killed.
> 
> 
> confirmed downstream on Radarr 4.2.3.6522 the same outcome. the systemd exits sucessfully and the service is killed & restarted externally

Taloth:
We're talking about the same thing, The UI  'Restart' option, should NOT restart Sonarr. it should exit with non-zero exit code, which would lead systemd to restart sonarr properly. Any subsequence internal or external restarts would therefore work properly.
The UI 'Shutdown' option should exit with zero exit code, which would lead to systemd NOT restarting sonarr.

Antorell:
Ok cool. Thanks for the clarification.

thomas-massey:
I sadly too - about a week ago got this issue from the apt repo and it is still happening now

Issue #3985, opened 2020-09-29
Enhance Discord Notifications


bakerboy448:
Title....

Implement support for Discordnotifer.com which provides enhanced discord notifications.
Feature exists in radarr https://github.com/Radarr/Radarr/pull/4611

PR #3819  would implement this.

markus101:
We're aware of the PR, don't need a GHI after the fact.

bakerboy448:
But the feature does not exist in Sonarr, correct?
PR's are not necessarily feature requests, they are code changes, correct?
Thus would a feature request not be appropriate to add a feature?

Just some thoughts, not expecting nor do I need a reply.  

I'm operating off the assumption that the PR is not merged because it is not a feature that you believe you or Taloth will use or enjoy and thus it is not a feature that will be added for the community to use. I'd certainly hope that is not the case, but 🤷

markus101:
Correct it does not.
They are not feature requests, but a feature request doesn't need to exist for a PR to be merged to Sonarr.
If the GHI added value, sure, but adding a GHI to say you're interested in an existing PR doesn't really add value, it's just noise.

If we saw no value in the change it'd be closed, at this point we just haven't reviewed it.

markus101:
Updated the title and re-opening because this isn't as cut and dry as landing that PR. Like Radarr, we can implement most if not all of the functionality that Discord Notifier offers, without using a third party site same as Radarr has done here: https://github.com/Radarr/Radarr/pull/5043

I'd much rather have native options for this, because it removes confusion for users and avoids reliance on a third party service. I'm not immediately closing that PR, but unless there is a compelling reason to go that route over doing it directly in Sonarr we'll do that.

bakerboy448:
While I don't disagree that purely native is ideal.  I'm not sure the features in said third party service can fully be implemented natively in Sonarr given Discord's current accessibility...or lack thereof....many of the prime features require a discord bot at this time. I'm assuming neither Team Radarr or Sonarr want to go down the route of creating and maintaining a discord bot, would that be accurate? (inherently that would then rely on a third party service, a servarr and/or skyhook bot...)

While the end goal is great, it is likely a pretty significant effort to implement and obviously there are various other issues to address in the interim.   Would it not make sense to roll this into two phases?  Phase 1, implement the PR to provide immediate proven functionality.  Phase 2, work on implementing that same functionality natively. 

[...]

PR #3819, opened 2020-06-16
Connect > DiscordNotifier option

austinwbest:
- Added Discord Notifier support to connect options

#### Description
Adds an integrated notification for discordnotfier.com

PR #5168, opened 2022-09-28
New: Add Notifiarr connection

bakerboy448:
#### Database Migration
NO

#### Description
- Implements native Notifiarr support to be used in lieu of existing webhook functionality
- Over a dozen other apps provide native Notifiarr support and do not require a webhook
  - Autobrr
  - Lidarr
  - Plex Meta Manager
  - Polinatarr
  - Prowlarr
  - Qbit Manage 
  - MDBList
  - Radarr
  - Readarr
  - Reciperr
  - Unpackerr
  - Whispar
  - CrossSeed
- Previous concerns regarding confusion with Discord are now moot
- Allows customization of fields specific to notifiarr without modifying webhooks
- Allows easier and simpler configuration by users

#### Todos

- Tests
- [ ] Wiki Updates

* Related to #3985
* Related to #3819

### DIFF
--- a/src/NzbDrone.Core/Notifications/Notifiarr/NotifiarrProxy.cs
+++ b/src/NzbDrone.Core/Notifications/Notifiarr/NotifiarrProxy.cs
@@ -1,0 +1,112 @@
+using System;
+using System.Collections.Specialized;
+using FluentValidation.Results;
+using NLog;
+using NzbDrone.Common.Extensions;
+using NzbDrone.Common.Http;
+using NzbDrone.Core.Configuration;
+
+namespace NzbDrone.Core.Notifications.Notifiarr
+{
+    public interface INotifiarrProxy
+    {
+        void SendNotification(StringDictionary message, NotifiarrSettings settings);
+        ValidationFailure Test(NotifiarrSettings settings);
+    }
+
+    public class NotifiarrProxy : INotifiarrProxy
+    {
+        private const string URL = "https://notifiarr.com";
+        private readonly IHttpClient _httpClient;
+        private readonly IConfigFileProvider _configFileProvider;
+        private readonly Logger _logger;
+
+        public NotifiarrProxy(IHttpClient httpClient, IConfigFileProvider configFileProvider, Logger logger)
+        {
+            _httpClient = httpClient;
+            _configFileProvider = configFileProvider;
+            _logger = logger;
+        }
+
+        public void SendNotification(StringDictionary message, NotifiarrSettings settings)
+        {
+            try
+            {
+                ProcessNotification(message, settings);
+            }
+            catch (NotifiarrException ex)
+            {
+                throw ex;
+            }
+        }
+
+        public ValidationFailure Test(NotifiarrSettings settings)
+        {
+            try
+            {
+                var variables = new StringDictionary();
+                variables.Add("Sonarr_EventType", "Test");
+
+                SendNotification(variables, settings);
+                return null;
+            }
+            catch (NotifiarrException ex)
+            {
+                return new ValidationFailure("ApiKey", ex.Message);
+            }
+            catch (Exception ex)
+            {
+                _logger.Error(ex, ex.Message);
+                return new ValidationFailure("", "Unable to send test notification. Check the log for more details.");
+            }
+        }
+
+        private void ProcessNotification(StringDictionary message, NotifiarrSettings settings)
+        {
+            try
+            {
+                var instanceName = _configFileProvider.InstanceName;
+                var requestBuilder = new HttpRequestBuilder(URL + "/api/v1/notification/sonarr").Post();
+                requestBuilder.AddFormParameter("instanceName", instanceName).Build();
+                requestBuilder.SetHeader("X-API-Key", settings.ApiKey);
+
+                foreach (string key in message.Keys)
+                {
+                    requestBuilder.AddFormParameter(key, message[key]);
+                }
+
+                var request = requestBuilder.Build();
+
+                _httpClient.Post(request);
+            }
+            catch (HttpException ex)
+            {
+                var responseCode = ex.Response.StatusCode;
+                switch ((int)responseCode)
+                {
+                    case 401:
+                        _logger.Error("Unauthorized", "HTTP 401 - API key is invalid");
+                        throw new NotifiarrException("API key is invalid");
+                    case 400:
+                        _logger.Error("Invalid Request", "HTTP 400 - Unable to send notification. Ensure Sonarr Integration is enabled & assigned a channel on Notifiarr");
+                        throw new NotifiarrException("Unable to send notification. Ensure Sonarr Integration is enabled & assigned a channel on Notifiarr");
+                    case 502:
+                    case 503:
+                    case 504:
+                        _logger.Error("Service Unavailable", "Unable to send notification. Service Unavailable");
+                        throw new NotifiarrException("Unable to send notification. Service Unavailable", ex);
+                    case 520:
+                    case 521:
+                    case 522:
+                    case 523:
+                    case 524:
+                        _logger.Error(ex, "Cloudflare Related HTTP Error - Unable to send notification");
+                        throw new NotifiarrException("Cloudflare Related HTTP Error - Unable to send notification", ex);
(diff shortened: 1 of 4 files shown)


markus101:
> This seems like it can already be accomplished with the existing webhooks functionality given that notifier seems to simply takes data from a custom script-like API.

> all that's needed is for someone to create the webhook receiver to handle the middleware.

If this was good enough for ntfy why does Notifiarr need a custom implementation?

> Allows customization of fields specific to notifiarr without modifying webhooks

Such as? Seems more like we’d do the work for webhooks, custom scripts and Notifarr.

> Allows easier and simpler configuration by users

Would this be true if we had presets like we do for newznab indexers?

bakerboy448:
> If this was good enough for ntfy why does Notifiarr need a custom implementation?

Since ntfy landed; seems the powers that be believed that concept was not good enough.

Also from previous discussions

> Webhooks sent from Sonarr aren’t a solution to everything, but often that’s because those implementations have an existing established API.

Notifiarr has an existing established api that every other automation related app uses.  Sonarr is the only app where the handling had to be changed to use sonarr's webhooks.

> Such as? Seems more like we’d do the work for webhooks, custom scripts and Notifarr.

Nothing immediate other than out of the gate the custom script and proposed notifiarr logic is significantly easier and simpler for contributors to understand rather than the object based webhook logic.

By similar logic, why not have all the notifications have better support for additional information?  Custom script and this Notifiarr implementation by and large provide the most information.  Perhaps every other notification (proper notification not triggers i.e. not the plex connect etc) should have the missing dozens of data points added to each of them?   It was previously noted as `trivial` to add information to connects.

> Would this be true if we had presets like we do for newznab indexers?

Yes the user would still need to modify the preset url to append their apikey to it, so unfortunately little to no benefit would be gained from this concept.    Taking this further perhaps adding an apikey field could work, but different webhook providers may need that as a header, appended as a parameter with a field prefix, or just appended to the url path.  The possibilities and complexity grow significantly which ultimately means more confusion for the user.

Infinity8806:
Just commenting to say as a ARR app user please implement this.

src/NzbDrone.Core/Notifications/Notifiarr/NotifiarrProxy.cs
```diff
@@ -0,0 +1,112 @@
...
+                    case 504:
+                        _logger.Error("Service Unavailable", "Unable to send notification. Service Unavailable");
+                        throw new NotifiarrException("Unable to send notification. Service Unavailable", ex);
+                    case 520:
+                    case 521:
+                    case 522:
+                    case 523:
+                    case 524:
+                        _logger.Error(ex, "Cloudflare Related HTTP Error - Unable to send notification");
+                        throw new NotifiarrException("Cloudflare Related HTTP Error  - Unable to send notification", ex);
```
markus101, line 104: Double space.

src/NzbDrone.Core/Notifications/Notifiarr/NotifiarrProxy.cs
```diff
@@ -0,0 +1,112 @@
...
+                    default:
+                        _logger.Error(ex, "Unknown HTTP  Error - Unable to send notification");
```
markus101, line 106: Double space.

src/NzbDrone.Core/Notifications/Notifiarr/NotifiarrProxy.cs
```diff
@@ -0,0 +1,112 @@
...
+            _logger = logger;
+        }
+
+        public void SendNotification(StringDictionary message, NotifiarrSettings settings)
+        {
+            try
+            {
+                ProcessNotification(message, settings);
+            }
+            catch (NotifiarrException ex)
```
markus101, line 37: This will result in every `HttpException` that is thrown by `ProcessNotification` getting logged twice. Does this need to capture the exception and throw it again with just the message or can it skip that?

src/NzbDrone.Core/Notifications/Notifiarr/NotifiarrProxy.cs
```diff
@@ -0,0 +1,112 @@
...
+            {
+                _logger.Error(ex, ex.Message);
+                throw new NotifiarrException(ex.Message);
```
markus101, line 40: If we want to throw the exception, just `throw` would accomplish that without including this in the stack trace, though this might also negatively affect the exception the caller receives as well.

src/NzbDrone.Core/Notifications/Notifiarr/NotifiarrSettings.cs
```diff
@@ -0,0 +1,28 @@
...
+            RuleFor(c => c.APIKey).NotEmpty();
+        }
+    }
+
+    public class NotifiarrSettings : IProviderConfig
+    {
+        private static readonly NotifiarrSettingsValidator Validator = new NotifiarrSettingsValidator();
+
+        [FieldDefinition(0, Label = "API Key", Privacy = PrivacyLevel.ApiKey, HelpText = "Your API key from your profile", HelpLink = "https://notifiarr.com")]
+        public string APIKey { get; set; }
```
markus101, line 21: `ApiKey` to maintain consistency with other naming

markus101 (changes requested):

markus101:
>> If this was good enough for ntfy why does Notifiarr need a custom implementation?

> Since ntfy landed; seems the powers that be believed that concept was not good enough.

That wasn't our "concept" that was something you came up with. You were happy to tell another user that their preferred notification method doesn't make sense, but now you're asking for us to consider your preferred method.

Appreciate the information on why webhooks won't work, it's unfortunate it can't be used as it seems like a better implementation than converting everything to a string before sending it.

src/NzbDrone.Core/Notifications/Notifiarr/NotifiarrProxy.cs
bakerboy448: The double logging is actually something that's been annoying, but didn't know the route to correct. Thanks

I'll test, but this should be able to just be passthrough then a simple throw

bakerboy448:
much cleaner logging now without the double ex.  TY

```
2022-10-28 17:30:10.2|Trace|HttpClient|Req: [POST] https://notifiarr.com/api/v1/notification/sonarr/(removed): instanceName=Sonarr&sonarr_eventtype=Test
2022-10-28 17:30:10.2|Trace|ConfigService|Using default config value for 'proxyenabled' defaultValue:'False'
2022-10-28 17:30:10.4|Trace|HttpClient|Res: [POST] https://notifiarr.com/api/v1/notification/sonarr/(removed): 400.BadRequest (371 bytes) (167 ms)
2022-10-28 17:30:10.4|Warn|HttpClient|HTTP Error - Res: [POST] https://notifiarr.com/api/v1/notification/sonarr/(removed): 400.BadRequest (371 bytes)
2022-10-28 17:30:10.6|Error|NotifiarrProxy|Invalid Request
2022-10-28 17:30:19.8|Trace|Scheduler|Pending Tasks: 0
2022-10-28 17:30:19.9|Error|Microsoft.AspNetCore.Diagnostics.ExceptionHandlerMiddleware|An unhandled exception has occurred while executing the request.

[v10.0.0.31477] FluentValidation.ValidationException: Validation failed: 
 -- APIKey: Unable to send notification. Ensure Sonarr Integration is enabled & assigned a channel on Notifiarr
[...]

2022-10-28 17:30:20.0|Trace|SonarrErrorPipeline|Handling Exception
2022-10-28 17:30:20.0|Warn|SonarrErrorPipeline|Invalid request Validation failed: 
 -- APIKey: Unable to send notification. Ensure Sonarr Integration is enabled & assigned a channel on Notifiarr
```

src/NzbDrone.Core/Notifications/Notifiarr/NotifiarrSettings.cs
Thirrian: I think Marcus wanted the _variable_ named ApiKey, not change the label ;)

markus101 (approved):
Looking much better, in the future avoid squashing changes into a single commit so the diff can be reviewed.

PR #5173, opened 2022-10-03
New: Store and use original Series language

Qstick:
#### Database Migration
YES (176 - add original language to Series)

#### Description
Store Original Language in DB for Series. 
Use Original Language for parsing fallback.
Add Original Language to Index Table as Column option
Add Original Language to Index as Sort and Filter option
Allow Original Language usage in Custom Formats

#### Todos
- [ ] Tests

### DIFF
--- a/src/NzbDrone.Core/Languages/Language.cs
+++ b/src/NzbDrone.Core/Languages/Language.cs
@@ -102,6 +102,7 @@ namespace NzbDrone.Core.Languages
         public static Language Malayalam => new Language(29, "Malayalam");
         public static Language Ukrainian => new Language(30, "Ukrainian");
         public static Language Slovak => new Language(31, "Slovak");
+        public static Language Original => new Language(-2, "Original");
 
         public static List<Language> All
         {
@@ -140,7 +141,8 @@ namespace NzbDrone.Core.Languages
                     Bulgarian,
                     Malayalam,
                     Ukrainian,
-                    Slovak
+                    Slovak,
+                    Original
                 };
             }
         }
--- a/src/NzbDrone.Core/MediaFiles/EpisodeImport/Manual/ManualImportService.cs
+++ b/src/NzbDrone.Core/MediaFiles/EpisodeImport/Manual/ManualImportService.cs
@@ -487,7 +487,7 @@ namespace NzbDrone.Core.MediaFiles.EpisodeImport.Manual
                 localEpisode.Quality = file.Quality;
                 localEpisode.Languages = file.Languages;
 
-                //TODO: Cleanup non-tracked downloads
+                // TODO: Cleanup non-tracked downloads
 
                 var importDecision = new ImportDecision(localEpisode);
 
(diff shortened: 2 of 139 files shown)


src/NzbDrone.Core/Languages/Language.cs
```diff
@@ -102,6 +102,7 @@ public override bool Equals(object obj)
         public static Language Malayalam => new Language(29, "Malayalam");
         public static Language Ukrainian => new Language(30, "Ukrainian");
         public static Language Slovak => new Language(31, "Slovak");
+        public static Language Original => new Language(-2, "Original");
```
markus101, line 105: Why `-2` instead of `-1`, what is `-1`?

src/NzbDrone.Core/Parser/ParsingService.cs
```diff
@@ -182,6 +184,29 @@ private RemoteEpisode Map(ParsedEpisodeInfo parsedEpisodeInfo, int tvdbId, int t
                 {
                     remoteEpisode.Episodes = GetEpisodes(parsedEpisodeInfo, series, remoteEpisode.MappedSeasonNumber, sceneSource, searchCriteria);
                 }
+
+                parsedEpisodeInfo.ExtraInfo["OriginalLanguage"] = series.OriginalLanguage;
+            }
+
+            //Use series language as fallback if we could't parse a language (more accurate than just using English)
```
markus101, line 191: Nit: Space before comment (not sure if we can style cop that?)

src/NzbDrone.Core/MediaFiles/EpisodeImport/Manual/ManualImportService.cs
```diff
@@ -139,6 +139,14 @@ public ManualImportItem ReprocessItem(string path, string downloadId, int series
             var rootFolder = Path.GetDirectoryName(path);
             var series = _seriesService.GetSeries(seriesId);
 
+            var languageParse = LanguageParser.ParseLanguages(path);
+
+            if (languageParse.Count <= 1 && languageParse.First() == Language.Unknown && series != null)
+            {
+                languageParse = new List<Language> { series.OriginalLanguage };
+                _logger.Debug("Language couldn't be parsed from release, fallback to series original language: {0}", series.OriginalLanguage.Name);
```
markus101, line 147: `fallback` -> `falling back` or maybe just `using`?

markus101 (approved):
Looking really good, just a few minor comments.

src/NzbDrone.Core/Languages/Language.cs
Qstick: In Radarr `-1` is `Any` which becomes obsolete when we drop it off of the quality profile to be consistent with Sonarr. This was to make CFs between the two consistent, but can use `-1` here if we want.

markus101: Seems reasonable to keep it the same, I was just wondering why it was `-2`.... `Any` is a separate language though... WTF.

Dnkhatri:
I think you will need a language map for some languages. 

[Chinese - Taiwan (臺灣國語) is being mapped to english instead of Chinese

@Qstick

DanAykroyd256:
@Dnkhatri or for those that have access to v4 (Alpha), is this change adding an option to grab the “Original" language somewhere; similar as it works in Radarr?

image

PR #5187, opened 2022-10-13
Add indexer name to the download report log

Dtaggart:
#### Database Migration
NO

#### Description
This was to add the indexer name that was responsible for download. We logged the client, this is another piece of that information.

#### Todos
- [x] Tests
- [x] Wiki Updates

I didn't see the need for a change to add logging.

None, this was mainly for my own personal needs that I think people would benefit from having when trying to debug a configuration issue.

### DIFF
--- a/src/NzbDrone.Core/Download/DownloadService.cs
+++ b/src/NzbDrone.Core/Download/DownloadService.cs
@@ -22,6 +22,7 @@ namespace NzbDrone.Core.Download
     {
         private readonly IProvideDownloadClient _downloadClientProvider;
         private readonly IDownloadClientStatusService _downloadClientStatusService;
+        private readonly IIndexerFactory _indexerFactory;
         private readonly IIndexerStatusService _indexerStatusService;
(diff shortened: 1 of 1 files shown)


src/NzbDrone.Core/Download/DownloadService.cs
```diff
@@ -109,7 +112,26 @@ public void DownloadReport(RemoteEpisode remoteEpisode)
                 episodeGrabbedEvent.DownloadId = downloadClientId;
             }
 
-            _logger.ProgressInfo("Report sent to {0}. {1}", downloadClient.Definition.Name, downloadTitle);
+            var indexerName = "Unknown";
```
markus101, line 115: This should already be available as `remoteEpisode.Release.Indexer`, is there a reason that can't be used?

Dtaggart: No, I didn't realize that was there, I was taking how the name was retrieved by the Id in other places I overlooked the release properties. I change it to pull from there.

PR #5200, opened 2022-10-17
Added Discord webhook support for manual interactions.

StickieBE:
#### Database Migration
YES - 177

#### Description
Adds [Discord] webhook support for manual interactions. (e.g.: an episode is downloaded and needs manual import).

**➡️ Updated frontend to include "On Manual Interaction".**

afbeelding

**➡️ Added Database migration file `#176`.**

**➡️ Added `ImportPendingNotified` TrackedDownloadState.**

**➡️ Added customizable `ManualInteractionFields` Discord settings, similar to the already present `GrabFields` & `ImportFields` (visible when using advanced mode).**

afbeelding

#### Todos
- [ ] Tests
- [ ] Wiki Updates [Corresponding wiki page here](https://wiki.servarr.com/sonarr/settings#connections)

* None that I can think of.

### DIFF
--- a/src/NzbDrone.Core/Datastore/Migration/176_add_on_manual_interaction_to_notifications.cs
+++ b/src/NzbDrone.Core/Datastore/Migration/176_add_on_manual_interaction_to_notifications.cs
@@ -1,0 +1,14 @@
+using FluentMigrator;
+using NzbDrone.Core.Datastore.Migration.Framework;
+
+namespace NzbDrone.Core.Datastore.Migration
+{
+    [Migration(176)]
+    public class add_on_manual_interaction_to_notifications : NzbDroneMigrationBase
+    {
+        protected override void MainDbUpgrade()
+        {
+            Alter.Table("Notifications").AddColumn("OnManualInteraction").AsBoolean().WithDefaultValue(0);
+        }
+    }
+}
--- a/src/NzbDrone.Core/Download/CompletedDownloadService.cs
+++ b/src/NzbDrone.Core/Download/CompletedDownloadService.cs
@@ -57,7 +57,8 @@ namespace NzbDrone.Core.Download
 
         public void Check(TrackedDownload trackedDownload)
         {
-            if (trackedDownload.DownloadItem.Status != DownloadItemStatus.Completed)
+            if (trackedDownload.DownloadItem.Status != DownloadItemStatus.Completed
+                || trackedDownload.State == TrackedDownloadState.ImportPendingNotified)
             {
                 return;
             }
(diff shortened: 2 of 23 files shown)


src/NzbDrone.Core/Datastore/Migration/176_add_on_manual_interaction_to_notifications.cs
```diff
@@ -0,0 +1,14 @@
+using FluentMigrator;
+using NzbDrone.Core.Datastore.Migration.Framework;
+
+namespace NzbDrone.Core.Datastore.Migration
+{
+    [Migration(176)]
```
markus101, line 6: This will need to be changed to 177, an existing PR is already using 176.

src/NzbDrone.Core/Download/CompletedDownloadService.cs
```diff
@@ -103,6 +104,15 @@ public void Check(TrackedDownload trackedDownload)
                 if (seriesMatchType == SeriesMatchType.Id)
                 {
                     trackedDownload.Warn("Found matching series via grab history, but release was matched to series by ID. Automatic import is not possible.");
+                    trackedDownload.State = TrackedDownloadState.ImportPendingNotified;
```
markus101, line 107: I don’t think this should be a separate state, if it needs to be tracked here then a separate property would be more appropriate.

markus101 (changes requested):
A couple comments to start, but the big thing will be adding support for all notifications, not just Discord.

StickieBE:
> A couple comments to start, but the big thing will be adding support for all notifications, not just Discord.

Exactly, figured I'd get the ball rolling. I primarily use and could test Discord.

StickieBE:
Will look into adding the other platforms soon-ish. I want this enough to justify the work. 😎

markus101:
https://github.com/Sonarr/Sonarr/commit/5ec282750bdea16f770282b80fdaec56a65749e2 implemented everything and fixed up some issues I came across.

austinwbest:
> [5ec2827](https://github.com/Sonarr/Sonarr/commit/5ec282750bdea16f770282b80fdaec56a65749e2) implemented everything and fixed up some issues I came across.


Reuse of event types without any field to distinguish is not very ideal to services that use the events as switches or to users who will have no clue what the difference is. With download being used twice it has the `isUpgrade` field as a flag at least but unless i missed it the webhook base is literally just reusing the grab event type and no extra fields added?

Why not just give it a unique event type instead of a different event type for custom scripts than webhooks and reusing without being able to tell them apart?

Hopefully i just missed it as this seems quite problematic otherwise. If i did miss it, please point it out as there needs to be some way to implement this logically.

austinwbest:
> That was purely a copy pasta mistake.

All good, it happens.

> That's the intention, not sure you you'd assume any differently.

Was not an assumption, it is literally the code that was pushed and seen by multiple devs and still merged. So all i did was asked a question about it for clarification. It was clarified in your previous statement of being just a c/p mistake.

PR #5214, opened 2022-11-01
New: Base API info endpoint

Qstick:
#### Database Migration
NO

#### Description
Adds authorized generic non-versioned endpoint for third party tools to grab latest API root and see API version status of all versions. 

**Endpoint**
```
192.168.1.1:8989/api
```

**Response**
```
{
  "latest": "/api/v3",
  "v2": "removed",
  "v3": "current"
}
```

#### Todos
- [ ] Wiki Updates

### DIFF
--- a/src/Sonarr.Http/ApiInfoController.cs
+++ b/src/Sonarr.Http/ApiInfoController.cs
@@ -1,0 +1,23 @@
+using System.Collections.Generic;
+using Microsoft.AspNetCore.Mvc;
+
+namespace NzbDrone.Http
+{
+    public class ApiInfoController : Controller
+    {
+        public ApiInfoController()
+        {
+        }
+
+        [HttpGet("/api")]
+        [Produces("application/json")]
+        public object GetApiInfo()
+        {
+            return new ApiInfoResource
+            {
+                Current = "v3",
+                Deprecated = new List<string>()
+            };
+        }
+    }
+}
--- a/src/Sonarr.Http/ApiInfoResource.cs
+++ b/src/Sonarr.Http/ApiInfoResource.cs
@@ -1,0 +1,14 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace NzbDrone.Http
+{
+    public class ApiInfoResource
+    {
+        public string Current { get; set; }
+        public List<string> Deprecated { get; set; }
+    }
+}


src/Sonarr.Http/ApiInfoController.cs
```diff
@@ -0,0 +1,34 @@
...
+            _configFileProvider = configFileProvider;
+        }
+
+        [HttpGet("/api")]
+        [Produces("application/json")]
+        public object GetApiInfo()
+        {
+            return new
+            {
+                Latest = $"{_configFileProvider.UrlBase}/api/v3",
```
Qstick, line 21: `v3` is hardcoded here similar to initializeJSController.cs, IntegrationTestBase.cs and Startup.cs. Probably a good idea that we consolidate these into BuildInfo or somewhere at some point in a latest API prop.

rcdailey:
You are amazing Q. I'll review tomorrow!

src/Sonarr.Http/ApiInfoController.cs
```diff
@@ -0,0 +1,34 @@
...
+                v2 = ApiStatus.Removed,
+                v3 = ApiStatus.Current
+            };
+        }
+
+        private enum ApiStatus
+        {
+            Current,
+            Depricated,
```
rcdailey, line 30: Spelling: `Deprecated`

src/Sonarr.Http/ApiInfoController.cs
```diff
@@ -0,0 +1,34 @@
...
+        }
+
+        [HttpGet("/api")]
+        [Produces("application/json")]
+        public object GetApiInfo()
+        {
+            return new
+            {
+                Latest = $"{_configFileProvider.UrlBase}/api/v3",
+                v2 = ApiStatus.Removed,
```
rcdailey, line 22: To reduce schema changes, would it be better to do this instead?

```json
"removed": ["v2"],
"deprecated": [],
"current": "v2"
```

And technically, `removed` isn't even needed I guess. If it's not deprecated and not current, it's obviously removed.

Thoughts? I don't have a strong opinion one way or the other. Just food for thought.

markus101: Personally I like this format better than what I suggested, both from the schema perspective and the fact that you can exclude the removed ones which makes it even cleaner.

Qstick: As long as we are fine eliminating the return of the full api root (which includes any urlbase that user has setup) and just returning the version we can switch to this.

markus101: I think that's fine, they managed to hit this endpoint which also includes the URL base.

rcdailey (approved):

src/Sonarr.Http/ApiInfoController.cs
rcdailey: > As long as we are fine eliminating the return of the full api root (which includes any urlbase that user has setup) and just returning the version we can switch to this.

I'm personally OK with that because at least in Recyclarr's use case, I ask the user to input their base URL. Most software I've seen that interacts with Sonarr also asks for this information from the user.

I think as long as the general structure of the URL never changes, that `v3` is sufficient to fill in the gap:

```
{baseurl}/api/{version}
```

- `{baseurl}`: Requested from the user (client responsibility)
- `{version}`: Nondeterministic; provided by API

Thanks again Q!

markus101 (approved):
Looks good!

PR #5235, opened 2022-11-25
New: Rework List sync interval logic

### LINKED ISSUE #5011: Adjustable list sync task schedule interval

Seems there's been a ([poor](https://github.com/Sonarr/Sonarr/issues/4325)) attempt to request this feature in the past. 

My issue is with the import list feature. As of now I don't know of another way to have separate Sonarr instances and keeping these in sync for keeping different libraries.
My current connection allows me to use my Jellyfin instance to watch media on the go and locally. However my speeds and hardware aren't good for transcoding 4K content. For this reason I keep a separate 4K library on a separate instance of Sonarr for local use.

When adding Media to sonarr on the go I use the nzb360 app. The way I set it up a quality profile in the main sonarr instance which the 4K instance imports. This way I can have both a 1080p version and 4K version available.

In order to solve this i'd need an option for multiple quality options within a single instance, the ability to sync multiple instances, or my suggestion, the ability to change the task interval in order to keep the instance more synchronized as there is a big delay in between the time needed to gather the 4k version for it.

A simple drop-down menu in the task menu or an option within the Import list setting would help tons.

Any other alternatives besides my current set-up seem cumbersome and unfriendly to use.
I've tried renaming the series 

### Anything else?

Used the following guide to attempt synced sonarr instances: [TraSH-Guides](https://trash-guides.info/Sonarr/Tips/Sync-2-radarr-sonarr/)
Having the same thing with Radarr

Qstick:
#### Database Migration
YES `(178 - Add LastSync Column to List Status Table)`

#### Description
This makes a few changes to move list sync interval from global to by list implementation, allowing ListTypes which are user controlled (Sonarr Instances) to be minimal, while lists like Trakt and Plex remain more conservative. 

Currently Min Intervals by type are setup like so:
```
Sonarr - 5 mins
Plex - 6 hours (we can adjust this based on teams thoughts)
Trakt - 12 hours
IMDb - 12 hours
Custom - 6 hours
```

I choose not to make list interval user adjustable, but that's an easy add if we think that's useful in addition (this comes down to if we think users want something greater than the min intervals)

#### Todos
- [ ] Tests
- [ ] Wiki Updates

* Fixes #5011

### DIFF
--- a/frontend/src/Settings/ImportLists/ImportLists/EditImportListModalContent.css
+++ b/frontend/src/Settings/ImportLists/ImportLists/EditImportListModalContent.css
@@ -7,3 +7,9 @@
 .labelIcon {
   margin-left: 8px;
 }
+
+.message {
+  composes: alert from '~Components/Alert.css';
+
+  margin-bottom: 30px;
+}
--- a/frontend/src/Settings/ImportLists/ImportLists/EditImportListModalContent.js
+++ b/frontend/src/Settings/ImportLists/ImportLists/EditImportListModalContent.js
@@ -2,6 +2,7 @@ import PropTypes from 'prop-types';
 import React from 'react';
 import SeriesMonitoringOptionsPopoverContent from 'AddSeries/SeriesMonitoringOptionsPopoverContent';
 import SeriesTypePopoverContent from 'AddSeries/SeriesTypePopoverContent';
+import Alert from 'Components/Alert';
 import Form from 'Components/Form/Form';
 import FormGroup from 'Components/Form/FormGroup';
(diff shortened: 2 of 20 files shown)


markus101:
Migration 177 is already to be used by https://github.com/Sonarr/Sonarr/pull/5200

Qstick:
> Migration 177 is already to be used by #5200

Fixed

markus101 (commented):
I think the defaults seem alright, but maybe we should cut trakt's sync to 12 hours. I think we'd also want to show these values in the UI otherwise people might be confused and think they're running every 5 minutes.

markus101 (approved):
Looking great

PR #5236, opened 2022-11-25
Imdb lists

### LINKED ISSUE #276: Import new Series from IMDB Watch List

Qstick:
#### Database Migration
NO

#### Description
This PR adds IMDB LS Lists as an option for import list. As well this significantly speeds up ImportListSync due to eliminating 2 DB calls per result and instead making the 2 per batch of results. 

#### Todos
- [x] Tests
- [ ] Wiki Updates

This needs to go in before #5235 or be reworked after that's merged

* Fixes #276

### DIFF
--- a/src/NzbDrone.Core.Test/ImportListTests/ImportListSyncServiceFixture.cs
+++ b/src/NzbDrone.Core.Test/ImportListTests/ImportListSyncServiceFixture.cs
@@ -25,17 +25,21 @@ namespace NzbDrone.Core.Test.ImportListTests
 
             _importListReports = new List<ImportListItemInfo> { importListItem1 };
 
-            Mocker.GetMock<IFetchAndParseImportList>()
-                  .Setup(v => v.Fetch())
-                  .Returns(_importListReports);
+            Mocker.GetMock<ISeriesService>()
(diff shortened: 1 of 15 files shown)


markus101 (approved):
Just some minor changes then good to land.

mkanet:
@markus101 Sorry to intrude.  Which version of Sonarr will we see IMDB list support?  Thank you so much.

JudgeDreddKLC:
Hi, when should we start seeing this in production?

markus101:
When v4 beta is released, whenever that is.

JudgeDreddKLC:
> When v4 beta is released, whenever that is.

Fair enough, big thanks to the team for all the efforts

PR #5237, opened 2022-11-25
New: Auto tagging of series

### LINKED ISSUE #3870: Automatically Add Tags Based on Series Attributes
**Describe the problem**
Right now there is no way to automatically set tags based on the series type or any other attribute, without a way to apply tags automatically users need to manually add tags for each series they add, which may differ for anime vs standard or adults vs kids shows.

A clear and concise description of any solutions or features you've considered.

A custom filter-like implementation where users can create rules that result in tags being added.

With series that have the type anime, add the `anime` tag, for ones in the `/tv-kids` root folder add the `kids` tag, etc.

To start these could be single rules resulting in one or more tags being added, in the future there could be more complex rules, such as `series type && root folder`.

Tag settings seems like a reasonable place for this to exist.

markus101:
#### Database Migration
YES - 179

#### Description
Adds support for rules that will automatically tag series with the specified tags and the option to remove tags if the rules no longer match.

It's based on the same idea of Custom Formats to check the rules and the similar UI changes (based on react hooks).

#### Todos
- [X] Tests
- [ ] Wiki Updates

* Closes #3870

### DIFF
--- a/frontend/src/Components/Form/ProviderFieldFormGroup.js
+++ b/frontend/src/Components/Form/ProviderFieldFormGroup.js
@@ -35,6 +35,8 @@ function getType({ type, selectOptionsProviderAction }) {
       return inputTypes.TEXT;
     case 'oAuth':
       return inputTypes.OAUTH;
+    case 'rootFolder':
+      return inputTypes.ROOT_FOLDER_SELECT;
     default:
       return inputTypes.TEXT;
   }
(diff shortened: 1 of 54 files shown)


Qstick (changes requested):
Some initial UI things, testing this out today a bit

Qstick (approved):
This is working well in testing all current specifications and settings (removal). All scenarios functioned as expected. 

Last thing I can see is we need to add this table to CleanupUnusedTags so that tags used by these don't disappear in housekeeping.

PR #5255, opened 2022-11-29
Fixed: Use route Id for PUT requests if not passed in body

### LINKED ISSUE #5254: Monitor/Unmonitor episode from episode details fails.

### Current Behavior

When you click on a specific episode, e.g. from the series page, activity queue, etc., and click the flag in the upper-left of the popup, Sonarr fails to toggle the monitored/unmonitored status of that episode.

### Expected Behavior

Clicking the flag toggles the monitored/unmonitored status.

### Steps To Reproduce

1. Open the series page for one of your monitored series
2. Click the episode of one episode
3. Click the flag in the upper-left corner of the pop-up

### Environment

```markdown
- OS: Windows 11
- Sonarr: 4.0.0.189
- Docker Install:No
- Using Reverse Proxy: No
- Browser: Chrome 107.0.5304.108
```

Develop

### Trace Logs?

https://gist.github.com/azkar0/f17e8d5a08238dbed25b4d7792b6cc40

### Anything else?

Worked as expected in v3. Monitoring/unmonitoring individual episodes from the series page (not clicking a specific episode) works as expected.

Qstick:
#### Database Migration
NO

#### Description
When accepting a PUT API request, if no Id is passed in the body then the request fails due to resource Id validation. This PR uses the Id from the route in those cases to supplement. The current code requires that an Id always be set in the body resource even though it's set in the route (slightly different behavior than v3). 

Example is episode PUT for monitor toggle with the following payload:

`http://localhost:8989\api\v3\episode\1234`

```
{
    monitored: true
}
```

* Fixes #5254

### DIFF
--- a/src/Sonarr.Http/REST/RestController.cs
+++ b/src/Sonarr.Http/REST/RestController.cs
@@ -70,6 +70,12 @@ namespace Sonarr.Http.REST
 
                 foreach (var resource in resourceArgs)
                 {
+                    // Map route Id to body resource if not set in request
+                    if (Request.Method == "PUT" && resource.Id == 0 && context.RouteData.Values.TryGetValue("id", out var routeId))
+                    {
+                        resource.Id = Convert.ToInt32(routeId);
+                    }
+
                     ValidateResource(resource, skipValidate, skipShared);
                 }
             }
--- a/src/Sonarr.Api.V3/Episodes/EpisodeController.cs
+++ b/src/Sonarr.Api.V3/Episodes/EpisodeController.cs
@@ -8,6 +8,7 @@ using NzbDrone.SignalR;
 using Sonarr.Http;
 using Sonarr.Http.Extensions;
 using Sonarr.Http.REST;
+using Sonarr.Http.REST.Attributes;
 
 namespace Sonarr.Api.V3.Episodes
 {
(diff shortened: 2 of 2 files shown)


Qstick:
Other option here is we leave as is and require Id in route and body. If we want to stay with this route, I can modify PR to have the UI pass a correct request.

src/Sonarr.Http/REST/RestController.cs
```diff
@@ -70,6 +70,12 @@ public override void OnActionExecuting(ActionExecutingContext context)
 
                 foreach (var resource in resourceArgs)
                 {
+                    // Map route Id to body resource if not set in request
+                    if (Request.Method == "PUT" && resource.Id == 0 && context.RouteData.Values.TryGetValue("id", out var routeId))
+                    {
+                        resource.Id = int.Parse(routeId.ToString());
```
markus101, line 76:
```suggestion
                        resource.Id = Conver.ToInt32(routeId);
```

This avoids the warning with converting to string if `routeId` is null, but otherwise functions the same.

markus101 (approved):
Nice fix here, just a small suggestion then good to land.

src/Sonarr.Http/REST/RestController.cs
Qstick: Looks good. This one actually effects downstream as well, so good catch on the issue OP

PR #5258, opened 2022-11-30
Fixed: Add back LanguageProfiles endpoints and deprecate

Qstick:
#### Database Migration
NO

#### Description
Adds back shell endpoints for Language Profiles and Language Profile Schemas.
Adds Deprecated Attribute which results in response marked with deprecation header as well as Logs a Warn to Sonarr. 

#### Todos
- [ ] Tests


markus101 (approved):
I think converting `DEPRECATED_ATTRIBUTE` to a single type would make it a little cleaner and I can't imagine we'll have multiple deprecated attributes. Good to ship after that though.

PR #5278, opened 2022-12-08
New: Speed up mass deletes from Series Editor

Qstick:
#### Database Migration
NO

#### Description
This change significantly speeds up mass delete of series by ensuring most DB calls are only called once per mass delete instead of once per series per mass delete (including a GetAllSeries call).

#### Todos
- [ ] Tests


markus101 (approved):
Some of these are really WTF as to why we did it that way to begin with, good fixes, approving to unblock after you add the tests and CI is happy.

PR #5289, opened 2022-12-10
New: Reduce size of Custom Formats in responses for most endpoints

markus101:
#### Database Migration
NO

#### Description
As discussed on Discord this reduces the size of responses that include Custom Formats outside of `CustomFormatController` responses that need the full information.


Qstick (approved):
Looks good, tested a few cases in UI (creation, editing, etc) all seems good.

PR #5292, opened 2022-12-12
Fixed: Store Custom Format score during import

### LINKED ISSUE #5291: CF score of 0 for episodes from season pack on history

### Current Behavior

season pack has score of 5, on ep import that episode has score of 0. even though it has the same CF matches as the season pack.
i went to grab another version of the episode and it shows up with correct score of 5 for having the same matches

### Expected Behavior

show correct CF score

### Steps To Reproduce

download season pack, compare season pack CF vs season pack episode's CF 

### Environment

```markdown
- OS: windows 10
- Sonarr: 4.0.0.277
- Docker Install: no
- Using Reverse Proxy: no
- Browser: firefox 107.1
```

Develop

### Trace Logs?

sonarr

### Anything else?

_No response_

markus101:
#### Database Migration
NO

#### Description
This now stores CF score when importing episodes to align with grabs, but other events would still be lacking this score so it feels a bit janky, but not sure we want to store it for others or at all. Given CF is recalculated for everything else should the score be as well? Should we have some sort of caching for CF to allow quicker look ups when given the same information over and over?

#### TODOs

- [ ] Add to notifications/connections

* Closes #5291

### DIFF
--- a/frontend/src/Activity/History/HistoryRow.js
+++ b/frontend/src/Activity/History/HistoryRow.js
@@ -63,6 +63,7 @@ class HistoryRow extends Component {
       languageCutoffNotMet,
       quality,
       customFormats,
+      customFormatScore,
       qualityCutoffNotMet,
       eventType,
       sourceTitle,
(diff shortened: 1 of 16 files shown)


Qstick:
Yea, we should calculate this on the fly shouldn't we? I'd say it doesn't hurt to store the score at the time of the event as a reference point as to why Sonarr did what it did, but any decisions made on new releases would be due to comparisons with current scores of the event. 

Caching would be good if we could do it reliably.

bpoxy:
The season history view still doesn't reflect the custom score.

Season: 
season

Episode:
episode

PR #5293, opened 2022-12-12
Fixup ProgressBar Text Colors

thezak48:
#### Database Migration
NO

#### Description
Fixed the progress bar text colors so that they are white regardless of the background, for some reason there was a weird bug causing different parts of the episode progress text to be different colors as the bar moved along as shown in the image

image

#### Todos
- N/A

*

### DIFF
--- a/src/NzbDrone.Mono/Disk/ProcMountProvider.cs
+++ b/src/NzbDrone.Mono/Disk/ProcMountProvider.cs
@@ -27,6 +27,8 @@ namespace NzbDrone.Mono.Disk
 
         private static Dictionary<string, bool> _fileSystems;
 
+        private bool _hasLoggedProcMountFailure = false;
+
         public ProcMountProvider(Logger logger)
         {
             _logger = logger;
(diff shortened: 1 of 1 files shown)


frontend/src/Components/ProgressBar.css
```diff
@@ -23,6 +23,10 @@
   color: var(--white);
 }
 
+.backTextContainer {
+  color: var(--white);
```
markus101, line 27: It's different colours so it shows up on the background for light theme properly as well as the various bar colours. This change completely breaks the light theme.

image

markus101 (changes requested):

frontend/src/Components/ProgressBar.css
Qstick: instead of setting this to a color like `--white` maybe helpful to move to a variable like `--progressBarBackgroundColor` or set of variables for progress bar text color. This will allow it to be different per theme.

PR #5305, opened 2022-12-16
New: Add bypass if above Custom Format Score to Delay Profile

### LINKED ISSUE #4711: Add Bypass if Custom Format Above X for DelayProfiles

Many people use release profiles to grab their preferred groups, etc.  Delay profiles allow for delaying releases to wait for the best one and can be overrode for quality only.

Some series have many releases in the highest quality, but at less than preferred word scores.  So currently users have two options: delay releases for an extended period of time and miss out on ratios and quick episode adds to their media server OR bypass if the highest quality is reached and have multiple/several upgrades occur wasting indexer grabs, download/upload usage, and storage space (first grab file would need to seed until X ratio on private trackers) as the initial highest quality was low quality based on release profiles.

Currently delay profiles only allow bypassing based on quality.  This should be enhanced to add a bypass if the preferred word score is > X.

Thoughts to consider:
- AND or OR relationship between Bypass for Quality & Bypass for Score
- Edge cases?

Have perfectly suitable downloads sit held in delay and manually grab

### Anything else?

N/A

markus101:
#### Database Migration
YES - 182

#### Description
Adds two more options to Delay Profiles which will bypass the delay profile if a minimum CF score is reached.

A few things:
- This operates independently from the quality bypass
- The minimum value is based on a second property in the delay profile, alternatively it could use the minimum score or cutoff from the profile instead, which would make it less related to the profile's scoring
- If we want to switch this to requiring both quality and score if configured it'd be possible with some changes (and potentially could be another value for the delay profile)

#### Todos
- [x] Tests

* Closes #4711

### DIFF
--- a/src/NzbDrone.Core/Datastore/Migration/182_add_bypass_to_delay_profile.cs
+++ b/src/NzbDrone.Core/Datastore/Migration/182_add_bypass_to_delay_profile.cs
@@ -8,7 +8,7 @@ namespace NzbDrone.Core.Datastore.Migration
     {
         protected override void MainDbUpgrade()
         {
-            Alter.Table("DelayProfiles").AddColumn("BypassIfAboveCustomFormatScore").AsBoolean().Nullable();
+            Alter.Table("DelayProfiles").AddColumn("BypassIfAboveCustomFormatScore").AsBoolean().WithDefaultValue(false);
             Alter.Table("DelayProfiles").AddColumn("MinimumCustomFormatScore").AsInt32().Nullable();
         }
     }
--- a/src/NzbDrone.Core/Datastore/Migration/182_add_bypass_to_delay_profile.cs
+++ b/src/NzbDrone.Core/Datastore/Migration/182_add_bypass_to_delay_profile.cs
@@ -1,0 +1,15 @@
+using FluentMigrator;
+using NzbDrone.Core.Datastore.Migration.Framework;
+
+namespace NzbDrone.Core.Datastore.Migration
+{
+    [Migration(182)]
+    public class add_custom_format_score_bypass_to_delay_profile : NzbDroneMigrationBase
+    {
+        protected override void MainDbUpgrade()
+        {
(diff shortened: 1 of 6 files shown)


src/NzbDrone.Core/Datastore/Migration/182_add_bypass_to_delay_profile.cs
```diff
@@ -0,0 +1,15 @@
...
+using NzbDrone.Core.Datastore.Migration.Framework;
+
+namespace NzbDrone.Core.Datastore.Migration
+{
+    [Migration(182)]
+    public class add_custom_format_score_bypass_to_delay_profile : NzbDroneMigrationBase
+    {
+        protected override void MainDbUpgrade()
+        {
+            Alter.Table("DelayProfiles").AddColumn("BypassIfAboveCustomFormatScore").AsBoolean().Nullable();
```
Qstick, line 11: Should we make this boolean non-nullable with default false, similar to the bypass quality setting that was added

```suggestion
            Alter.Table("DelayProfiles").AddColumn("BypassIfAboveCustomFormatScore").AsBoolean().WithDefaultValue(false);
```

Qstick (commented):
Just the one comment on the migration. 

Regarding the comments, I can see the case for using the quality profile min or cutoff score instead of a separate score on the delay profile, given CFs could be scored differently on different quality profiles and scores are really quality profile based. There is also some flexibility to have it separate like this as well though.

markus101:
Yeah, I'm leaning more towards using a value in the profile for it, like we do with the quality bypass, which I think without adding anything extra would make the most sense to be the cutoff. @bakerboy448 thoughts?

bakerboy448:
Quality Profile CF Cutoff Score would be moot since users that use CFs to upgrade releases would set a nice high school like 10,000.

Min Score could maybe be an option, but that seems rather redundant.

User entered score is probably the least worst option and then they can use tags as needed.

PR #5307, opened 2022-12-16
New: Seed Goals are not advanced

bakerboy448:
#### Database Migration
NO

#### Description

#### Todos
- Tests
- [ ] Wiki Updates

*


markus101:
We discussed this previously and opted to leave seed time advanced, since we're referring to seed ratio directly we made it a normal setting and left time advanced so people don't mess it up. Open to discussing it further though because on one hand it's a bit advanced, but on the other it's different from the similar setting.

bakerboy448:
Considering trackers are typically min time or ratio based.   It probably makes sense to make it not advanced...with some tweaks.  
if not otherwise existing:
- seed time / ratio goals should be validated to be > 0
- Ideally the caps response from Prowlarr/Jackett can be updated to include the min seed time / min ratio which sonarr can use to warn and require force saving accordingly....I believe I saw some (likely unusused) logic that supports this

markus101:
We discussed this and will be leaving seed time as advanced, at least for the time being.

PR #5310, opened 2022-12-18
Fixed: Don't enforce minimum on single list fetch

Qstick:
#### Database Migration
NO

#### Description
Don't check minimum when single list is fetched (provider updates and tests)


markus101 (approved):
Looks good to me and the failing test looks to be XEM related, but re-running.

PR #5315, opened 2022-12-21
New: Don't block imports when release was matched by ID if they were grabbed by Interactive Search

### LINKED ISSUE #5043: Store Grab Type (RSS, Search, Interactive Search) and Automatically Import Interactive Search with Mismatched title

#4698
this issue had mentioned:
when manual grab the release rejected by sonarr with: Unknown Series
sonaar will import into serial 01 automatically and cover

#4826
this issue had mentioned:
when the release title don't match indexer's TVDB Id
sonaar import automatically 
> pay attention: the release didn't rejected by sonarr

and i found the solution:
[Fixed: Don't automatically import if release title doesn't match series title](https://github.com/stevietv/Sonarr/commit/4cefc1ed19db72296affdfc72cd1bfd0b234e244)
and i saw: 
```
if (series == null)
```
it is related to #4698 : Unknown Series
but commit message show that solved #4826

it confuse me: 
#4698 user think the release is right but rejected by sonarr
#4826 user think the release is wrong but accepted by sonarr
they are different

but it did solve #4698 
i found if exist S01EXX, sonarr will not import and cover and wait user deal
but still import into series 01 automatically if there don't exist S01EXX
image

by the way to #4826
i think it should not blame to sonarr, it should blame to why indexer provide wrong TVDB id.
if sonarr can't believe the indexer TVDB id, why should believe the title?
it could be optimized,  but can't be solve fully
and it is right to download and import the release accepted by sonarr
otherwise what should be download and import?

import automatically with episode and series which is searching when manual grab Unknown Series

> when user manual grab
> i think they know which episode and series they want
> and they think it is the right release but rejected by sonarr so they manual grab
> and the episode and series must be what they are searching

> it would be wonderfull if sonarr import automatically with episode and series which is searching
> instead of editting manually

if there exist S0EXX
i have to wait download finished and manual edit the episode and series.

if no
i have to wait download finished and go to folder, move it to the right path and rename.

### Anything else?

no, thanks

markus101:
#### Database Migration
NO

#### Description
Stores the grab type and uses it when checking whether a release should be blocked from automatic importing because it was matched by ID.

* Closes #5043


src/NzbDrone.Core/DecisionEngine/DownloadDecisionMaker.cs
```diff
@@ -156,6 +156,26 @@ private IEnumerable<DownloadDecision> GetDecisions(List<ReleaseInfo> reports, Se
 
                 if (decision != null)
                 {
+                    var source = pushedRelease ? ReleaseSourceType.ReleasePush : ReleaseSourceType.Rss;
+
+                    if (searchCriteria != null)
+                    {
+                        if (searchCriteria.InteractiveSearch)
```
bakerboy448, line 163: Would a case be cleaner than the nested ifs?

markus101: In this case it's not possible because these are different properties on `searchCriteria`, not a single property that can be used in a switch statement.

PR #5317, opened 2022-12-22
New: Use file's format title for quality if parsed

### LINKED ISSUE #3930: Try To Detect Source Quality from Metadata
**Describe the problem**
In cases where media source quality is not explicitly specified by the title or filename, Sonarr currently seems to detect resolution on import and assign the lowest source quality (HDTV/SDTV).  So an unmarked 720p episode would be imported as HDTV-720p

I've noticed that sometimes there are indications of the source quality in the metadata of the file.  Can Sonarr use the same logic it would use to detect source quality from a filename or title, but also search the metadata?  

An example snippet is included below where it has assigned HDTV-720p, but from the metadata it could detect that it is WEB-DL.  The tag is found in the Movie Name metadata tag most often in my experience

**Additional context**

> $ mediainfo 3x08-The\ Hero\ \[HDTV-720p\]\[x265\].mkv 
> General
> Complete name                            : 3x08-The Hero [HDTV-720p][x265].mkv
> Format                                   : Matroska
> ...
> Overall bit rate                         : 4 351 kb/s
> Movie name                               : Nathan for You S03E08 The Hero 720p WEB-DL DD+2 0

markus101:
Closes #3930

#### Database Migration
NO

#### Description
I checked some files for this value and only found one with a title, but it was just the name of the episode, but grabbing the title to use during import doesn't require much additional work. I opted to not store it since it's just extra data that doesn't do much after the import, but could change that if you think it's worth storing

#### Todos
- [x] Tests

* Closes #3930


src/NzbDrone.Core/MediaFiles/MediaInfo/MediaInfoModel.cs
```diff
@@ -57,5 +58,8 @@ public class MediaInfoModel : IEmbeddedDocument
         public List<string> Subtitles { get; set; }
 
         public string ScanType { get; set; }
+
+        [JsonIgnore]
+        public string Title { get; set; }
```
markus101, line 63: This is a bit generic, but not sure it's wroth going into much detail since it's not stored.

src/NzbDrone.Core/MediaFiles/MediaInfo/MediaInfoModel.cs
```diff
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using FFMpegCore;
+using Newtonsoft.Json;
```
Qstick, line 4: Should use from System.Text.Json instead of Newtonsoft.Json since that's what the DB uses now and we should migrate that way. I think it's the same attribute.

Qstick (changes requested):
Are we able to add this to the video info reader fixture as well just to make sure that part works, or does our included sample not have a title set? Looks good otherwise

Qstick:
Also, looks like a test failure on NzbDrone.Core.Test.MediaFiles.EpisodeImport.Aggregation.Aggregators.Augmenters.Quality.AugmentQualityFromMediaInfoFixture.should_have_unknown_source_if_no_source_extracted_from_title

 `Expected object to be Web, but found Television.`

Qstick (approved):

PR #5319, opened 2022-12-23
Fixed: Progress bar text colour in Dark theme

thezak48:
#### Database Migration
NO

#### Description
Buggered up my local git so continuation of this [PR 5293](https://github.com/Sonarr/Sonarr/pull/5293)

Fixed the progress bar text colors so that they are white regardless of the background, for some reason there was a weird bug causing different parts of the episode progress text to be different colors as the bar moved along as shown in the image
image

Moved the text to be a variable so can be changed depending on theme
image

#### Todos
- 

*

### DIFF
--- a/frontend/src/Styles/Themes/dark.js
+++ b/frontend/src/Styles/Themes/dark.js
@@ -235,6 +235,8 @@ module.exports = {
   //
   // misc
 
+  progressBarFrontTextColor: white,
+  progressBarBackTextColor: white,
   progressBarBackgroundColor: '#727070',
   logEventsBackgroundColor: '#2a2a2a'
 };
--- a/frontend/src/Components/ProgressBar.css
+++ b/frontend/src/Components/ProgressBar.css
@@ -20,7 +20,11 @@
 
 .frontTextContainer {
   z-index: 1;
-  color: var(--white);
+  color: var(--progressBarFrontTextColor);
+}
+
+.backTextContainer {
+  color: var(--progressBarBackTextColor);
 }
 
 .backTextContainer,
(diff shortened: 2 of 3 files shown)


frontend/src/Styles/Themes/dark.js
```diff
@@ -235,6 +235,7 @@ module.exports = {
   //
   // misc
 
+  progressBarTextColor: white,
```
markus101, line 238: Use hex codes instead of color names.

markus101 (changes requested):
> Fixed the progress bar text colors so that they are white regardless of the background, for some reason there was a weird bug causing different parts of the episode progress text to be different colors as the bar moved along as shown in the image

Different text colours is intentional, it probably makes sense for the dark theme to use the same colour for both on and off of the progress bar colour, but the light theme should continue to use different colours based on the state.

frontend/src/Styles/Themes/dark.js
thezak48: That is calling a previously defined hex code, if you want me to make it all hex I can but kinda sounds like extra work for no change

thezak48:
> > Fixed the progress bar text colors so that they are white regardless of the background, for some reason there was a weird bug causing different parts of the episode progress text to be different colors as the bar moved along as shown in the image
> 
> Different text colours is intentional, it probably makes sense for the dark theme to use the same colour for both on and off of the progress bar colour, but the light theme should continue to use different colours based on the state.

Made two vars for front and background text
image

markus101 (approved):

PR #5332, opened 2022-12-30
Fixed: Custom Formats upgrading even when profile doesn't allow upgrades

### LINKED ISSUE #5330: Disabling `Upgrades Allowed` does not prevent custom format score upgrades.

### Current Behavior

Sonarr is grabbing a release with a higher custom format score than the existing file even if `Upgrades Allowed` is unchecked, if the existing file's custom format score is lower than `Upgrade Until Custom Format Score`.

### Expected Behavior

`Upgrades Allowed` would apply to both quality and custom format score.

### Steps To Reproduce

1. Configure a custom format "Higher" that matches on release title for term "NTb".
2. Configure a custom format "Lower" that matches on release title for term "CAKES".
3. Configure a quality profile "Test":
    i. Score "Higher" as 1000 and "Lower" as 500.
    ii. Set `Upgrade Until Custom Format Score` to 1000.
    iii. Uncheck has `Upgrades Allowed`.
5. Add the show "Doom Patrol" and assign it "Test".
6. Do a manual search of S04E05 and trigger download of the CAKES release.
7. Once that's done, trigger an automatic search.
8. Notice the NTb release is grabbed.

### Environment

```markdown
- OS: Windows 10 x64
- Sonarr: 4.0.0.319
- Docker Install: No
- Using Reverse Proxy: No
- Browser: N/A
```

Develop

### Trace Logs?

https://0bin.net/paste/DsXRtXEL#E3dlScJCmuolo4EOyRvY1uHDLn5BXoCsCKcEhJdK7eY

### Anything else?

Setting `Upgrade Until Custom Format Score` sufficiently low can be used to prevent this but unchecking `Upgrades Allowed` hides this field, making it extra unintuitive. 

In my reproduction steps, I set `Upgrade Until Custom Format Score` to 500, for demonstration purposes. I realize this is a strange configuration, that a real user would likely never set up. In my actual case, it is set to 0 but the existing file's custom format score is negative so the cutoff is unmet. I realize this sort of scoring is also unusual but I still think that `Upgrades Allowed` not applying to custom format scores while also hiding `Upgrade Until Custom Format Score` is a confusing experience, hence the bug report.

markus101:
#### Database Migration
NO

#### Description
Fixes custom format upgrades when the quality profile doesn't allow upgrading.

#### Todos
- [X] Tests

* Closes #5330

### DIFF
--- a/src/NzbDrone.Core/DecisionEngine/Specifications/UpgradableSpecification.cs
+++ b/src/NzbDrone.Core/DecisionEngine/Specifications/UpgradableSpecification.cs
@@ -14,7 +14,7 @@ namespace NzbDrone.Core.DecisionEngine.Specifications
         bool QualityCutoffNotMet(QualityProfile profile, QualityModel currentQuality, QualityModel newQuality = null);
         bool CutoffNotMet(QualityProfile profile, QualityModel currentQuality, List<CustomFormat> currentCustomFormats, QualityModel newQuality = null);
         bool IsRevisionUpgrade(QualityModel currentQuality, QualityModel newQuality);
-        bool IsUpgradeAllowed(QualityProfile qualityProfile, QualityModel currentQuality, QualityModel newQuality);
+        bool IsUpgradeAllowed(QualityProfile qualityProfile, QualityModel currentQuality, List<CustomFormat> currentCustomFormats, QualityModel newQuality, List<CustomFormat> newCustomFormats);
     }
 
     public class UpgradableSpecification : IUpgradableSpecification
@@ -26,13 +26,6 @@ namespace NzbDrone.Core.DecisionEngine.Specifications
         {
             _configService = configService;
             _logger = logger;
-        }
-
-        private bool IsPreferredWordUpgradable(int currentScore, int newScore)
-        {
-            _logger.Debug("Comparing preferred word score. Current: {0} New: {1}", currentScore, newScore);
-
-            return newScore > currentScore;
         }
 
         public bool IsUpgradable(QualityProfile qualityProfile, QualityModel currentQuality, List<CustomFormat> currentCustomFormats, QualityModel newQuality, List<CustomFormat> newCustomFormats)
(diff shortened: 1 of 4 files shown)


src/NzbDrone.Core/DecisionEngine/Specifications/UpgradableSpecification.cs
```diff
@@ -142,17 +136,20 @@ public bool IsRevisionUpgrade(QualityModel currentQuality, QualityModel newQuali
...
-            if (isQualityUpgrade && qualityProfile.UpgradeAllowed)
+            if ((isQualityUpgrade || isCustomFormatUpgrade) && qualityProfile.UpgradeAllowed)
             {
-                _logger.Debug("At least one profile allows upgrading");
+                _logger.Debug("Quality profile allows upgrading");
                 return true;
             }
 
-            if (isQualityUpgrade && !qualityProfile.UpgradeAllowed)
+            if ((isQualityUpgrade || isCustomFormatUpgrade) && !qualityProfile.UpgradeAllowed)
```
markus101, line 152: Left this as a separate check so lower/same quality/CF score aren't rejected incorrectly as they'll be handled with another spec.

src/NzbDrone.Core/DecisionEngine/Specifications/UpgradableSpecification.cs
```diff
@@ -142,17 +136,20 @@ public bool IsRevisionUpgrade(QualityModel currentQuality, QualityModel newQuali
             return false;
         }
 
-        public bool IsUpgradeAllowed(QualityProfile qualityProfile, QualityModel currentQuality, QualityModel newQuality)
+        public bool IsUpgradeAllowed(QualityProfile qualityProfile, QualityModel currentQuality, List<CustomFormat> currentCustomFormats, QualityModel newQuality, List<CustomFormat> newCustomFormats)
         {
             var isQualityUpgrade = new QualityModelComparer(qualityProfile).Compare(newQuality, currentQuality) > 0;
+            var isCustomFormatUpgrade = qualityProfile.CalculateCustomFormatScore(newCustomFormats) > qualityProfile.CalculateCustomFormatScore(currentCustomFormats);
+            var t1 = qualityProfile.CalculateCustomFormatScore(newCustomFormats);
```
Qstick, line 143: Assume these were dev variables? Can be removed?

Qstick (approved):
Looks good, just the cleanup of the two unused variables.

PR #5335, opened 2023-01-02
Fixed: Kodi Metadata Subtitle Language

stevietv:
#### Database Migration
NO

#### Description
currently for episodes with multiple subs, only one subtitle element is created as follows:
```
     <subtitle>
        <language>engengengaracatczedangergergergrespaspabaqfinfilfreglghebhrvhuninditajpnkormaynobdutpolporporrumrusswethaturukrviechichi</language>
      </subtitle>
```

as per https://kodi.wiki/view/NFO_files/Movies multiple subtitle elements are allowed, where https://github.com/xbmc/xbmc/blob/master/xbmc/video/VideoInfoTag.cpp iterates through them.
This PR creates a seperate subtitle element for each langugae


Qstick:
Link for episodes for reference <https://kodi.wiki/view/NFO_files/Episodes> (format is same as movies link provided)

markus101 (approved):

PR #5362, opened 2023-01-10
Analysis tweaks

Qstick:
#### Database Migration
NO

#### Description
Ensure all code analyzers run so that we can control suppression with editor config. 

Also fixes most of the performance related code quality analyzer failures. Let me know if you don't agree with any of these. 

#### Todos
- [ ] Test


src/NzbDrone.Common/OAuth/WebParameterCollection.cs
```diff
@@ -16,7 +16,7 @@ public class WebParameterCollection : IList<WebParameter>
             {
                 var parameters = this.Where(p => p.Name.Equals(name));
 
-                if (parameters.Count() == 0)
+                if (!parameters.Any())
```
markus101, line 19: Could use `.Empty()` extension

src/NzbDrone.Common/Http/HttpUri.cs
```diff
@@ -170,7 +170,7 @@ private static string CombineRelativePath(string basePath, string relativePath)
 
             if (baseSlashIndex >= 0)
             {
-                return basePath.Substring(0, baseSlashIndex) + "/" + relativePath;
+                return string.Concat(basePath.AsSpan(0, baseSlashIndex), "/", relativePath);
```
markus101, line 173: String interpolation instead?

src/NzbDrone.Common/Extensions/StringExtensions.cs
```diff
@@ -131,7 +131,7 @@ public static bool ContainsIgnoreCase(this string text, string contains)
 
         public static string WrapInQuotes(this string text)
         {
-            if (!text.Contains(" "))
+            if (!text.Contains(' '))
```
markus101, line 134: Single quotes feels weird here, even though it's just a char, is there an advantage to it?

Qstick: From Microsoft rule docs:

When searching for a single character, using string.Contains(char) offers better performance than string.Contains(string).

src/NzbDrone.Common/OAuth/WebParameterCollection.cs
Qstick: Yep, I keep forgetting that one 😂

Qstick:
Tweaked based on comments, build on TC isn't respecting rules it seems. Something is funny there

Qstick:
This should be fixed now, just failing the unit tests that are bad on dev

markus101:
Doh, fixed the tests.

markus101 (approved):
Looks good.

PR #5367, opened 2023-01-12
Fixed: Changed extensions regex to js/jsx for babel-loader

mynameisbogdan:
NO

#### Description
Minor change to fix use of `.j` files in babel-loader.


markus101:
I already have a fix for this in another branch, not going to land this in the interim.

PR #5368, opened 2023-01-13
New: Improve Manual Import logging when not parsing files

bakerboy448:
#### Database Migration
NO

#### Description

Logs at warn that parse is being skipped; previously this was not clearer to the user

#### Todos
- Tests
- Wiki Updates

* https://forums.sonarr.tv/t/sonarr-v4-interactive-manual-import-not-detecting-series/31336

### DIFF
--- a/src/NzbDrone.Core/MediaFiles/EpisodeImport/Manual/ManualImportService.cs
+++ b/src/NzbDrone.Core/MediaFiles/EpisodeImport/Manual/ManualImportService.cs
@@ -246,7 +246,7 @@ namespace NzbDrone.Core.MediaFiles.EpisodeImport.Manual
 
                 if (files.Count() > 100)
                 {
-                    _logger.Warn("Found more than 100 files for multiple series. Skipping parsing.");
+                    _logger.Warn("Found more than 100 files for multiple series or no series. Skipping parsing.");
                     return ProcessDownloadDirectory(rootFolder, files);
                 }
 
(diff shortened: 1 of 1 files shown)


src/NzbDrone.Core/MediaFiles/EpisodeImport/Manual/ManualImportService.cs
```diff
@@ -246,6 +246,7 @@ private List<ManualImportItem> ProcessFolder(string rootFolder, string baseFolde
 
                 if (files.Count() > 100)
                 {
+                    _logger.Warn("Found more than 100 files for multiple series. Skipping parsing.");
```
markus101, line 249: It's possible they aren't for multiple series, they're just in a folder that doesn't have a series name so Sonarr doesn't bother parsing them all.

markus101 (changes requested):
This only improves things if people go an look at the logs,

bakerboy448:
It's a quick fix that at least communicates it to the users _somewhere_

Best solution would be to have it show in the UI, but not sure on the most optimal way to do that.

src/NzbDrone.Core/MediaFiles/EpisodeImport/Manual/ManualImportService.cs
bakerboy448, line 249:
```suggestion
                    _logger.Warn("Found more than 100 files for multiple series or no series. Skipping parsing.");
```

src/NzbDrone.Core/MediaFiles/EpisodeImport/Manual/ManualImportService.cs
```diff
@@ -246,6 +246,7 @@ private List<ManualImportItem> ProcessFolder(string rootFolder, string baseFolde
 
                 if (files.Count() > 100)
                 {
+                    _logger.Warn("Found more than 100 files for multiple series or no series. Skipping parsing.");
```
markus101, line 249:
```suggestion
                    _logger.Warn("Unable to determine series from folder name and found more than 100 files. Skipping parsing");
```

Which should avoid any confusion that it's the file names that matter in this case.

markus101 (changes requested):

PR #5370, opened 2023-01-13
Remove incorrect exclamation mark

ColinHebert:
There is no reason for `Series Title\'s` to gain an exclamation mark.

#### Database Migration
YES

#### Description
Fix incorrect/misleading embedded documentation

#### Todos
- [ ] Tests
- [ ] Wiki Updates

* N/A


markus101:
It's not incorrect, `CleanTitle` doesn't actually remove exclamation marks, it's doesn't Scenify the title (a title that all release groups would use). The `$#*! My Dad Says` test case for `CleanTitleFixture` shows this. This example has an exclamation point to show that in the UI.

ColinHebert:
@markus101 none of the other examples have an exclamation mark, so either the other ones filter exclamation marks or this one example is misleading in comparison to the others

PR #5374, opened 2023-01-15
Bump ImageSharp to 2.1.3

Qstick:
#### Database Migration
NO

#### Description
TVDB Image conversions when uploading PNGs seem to be just changing the file extension, thus many images are PNG saved as JPG. There is an old bug in ImageSharp when handling these. 

Pulls in this https://github.com/SixLabors/ImageSharp/issues/1704

#### Todos
- [ ] Tests


markus101 (approved):
Thanks!

PR #5389, opened 2023-01-21
New: Added new series title rename tokens without year

### LINKED ISSUE #5369: Add `Series TitleWithoutYear` Renaming token

Currently when working with serie titles, some pre-embed the year, while others do not. This inconsistency in behaviour makes it difficult to customise the naming pattern.

For example the series Kingdom has the following title: "Kingdom (2019)", the different options for the title are:
`{Series Title}` -> Kingdom (2019)
`{Series CleanTitle}` -> Kingdom 2019
`{Series TitleYear}` -> Kingdom (2019)
`{Series CleanTitleYear}` -> Kingdom 2019

If I want to have a `CleanTitle`, while retaining parenthesis around the year "Kingdom (2019)", my options are to do `{Series CleanTitle} ({Series Year})` which would give me "Kingdom 2019 (2019)", or `{Series CleanTitleYear}` which would give me "Kingdom 2019".

No option really fits the needs.

Introduce `VanillaTitle` (and it's `CleanVanillaTitle` equivalent) which is the vanilla name of the series (ie. without the year) allowing to customise how the title and year are displayed.

An alternative which would break backward compatibility is having `Title` and `CleanTitle` being the vanilla title.

### Anything else?

The introduction of "VanillaTitle" would remove the need for `TitleYear`, `TitleTheYear`, and any *Year equivalent, as the year can be managed entirely by the user however they want.

That said, the way `TitleYear` works currently assumes that [any 4 digits number specified in parenthesis in the title is a year](https://github.com/Sonarr/Sonarr/blob/develop/src/NzbDrone.Core/Organizer/FileNameBuilder.cs#L364-L368), and that it should be used as "the Year" for the series, taking precedence over `series.Year`.
Unfortunately the `{Series Year}` field only uses the `series.Year` entry and does not attempt to extract the year from the title.

An argument could be made for the sake of consistency for entries such as `Year`, `TitleYear`, and `CleanTitleYear` to either:
- Always have the year being `series.Year`
- Always have the year being the one specified in the non-vanilla series title, then fallback on `series.Year`

stevietv:
#### Database Migration
NO

#### Description
adds the naming tokens: 

* `{Series TitleWithoutYear}`
* `{Series TitleTheWithoutYear}`
* `{Series CleanTitleWithoutYear}`

* fixes #5369

### DIFF
--- a/src/NzbDrone.Core/Organizer/FileNameBuilder.cs
+++ b/src/NzbDrone.Core/Organizer/FileNameBuilder.cs
@@ -372,12 +372,6 @@ namespace NzbDrone.Core.Organizer
 
         public static string TitleWithoutYear(string title)
         {
-            // Regex match incase the year in the title doesn't match the year, for whatever reason.
-            // if (YearRegex.IsMatch(title))
-            // {
-            //    return title;
-            // }
-
             title = YearRegex.Replace(title, "");
 
             return title;
--- a/src/NzbDrone.Core/Organizer/FileNameBuilder.cs
+++ b/src/NzbDrone.Core/Organizer/FileNameBuilder.cs
@@ -66,7 +66,7 @@ namespace NzbDrone.Core.Organizer
 
         public static readonly Regex AirDateRegex = new Regex(@"\{Air(\s|\W|_)Date\}", RegexOptions.Compiled | RegexOptions.IgnoreCase);
 
-        public static readonly Regex SeriesTitleRegex = new Regex(@"(?<token>\{(?:Series)(?<separator>[- ._])(Clean)?Title(The)?(Year)?\})",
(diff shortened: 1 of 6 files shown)


src/NzbDrone.Core/Organizer/FileNameBuilder.cs
```diff
@@ -370,6 +370,19 @@ public static string TitleYear(string title, int year)
             return $"{title} ({year})";
         }
 
+        public static string TitleWithoutYear(string title)
+        {
+            // Regex match incase the year in the title doesn't match the year, for whatever reason.
```
markus101, line 375: Copy pasta?

markus101 (approved):
Looks like some commented code to remove then good to land.

src/NzbDrone.Core/Organizer/FileNameBuilder.cs
stevietv: fixed!

PR #5414, opened 2023-02-05
New: Make Release Group Outline Not Show as Required

thezak48:
#### Database Migration
NO

#### Description
When doing a manual import the Release Group setting shows as if it is a required setting, this changes that to not look like that

image

*

### DIFF
--- a/frontend/src/InteractiveImport/Interactive/InteractiveImportRow.js
+++ b/frontend/src/InteractiveImport/Interactive/InteractiveImportRow.js
@@ -319,7 +319,9 @@ class InteractiveImportRow extends Component {
         >
           {
             showReleaseGroupPlaceholder ?
-              <InteractiveImportRowCellPlaceholder /> :
+              <InteractiveImportRowCellPlaceholder
+                className={styles.optional}
+              /> :
               releaseGroup
           }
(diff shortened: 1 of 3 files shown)


thezak48:
image

frontend/src/InteractiveImport/Interactive/InteractiveImportRow.js
```diff
@@ -319,7 +319,9 @@ class InteractiveImportRow extends Component {
         >
           {
             showReleaseGroupPlaceholder ?
-              <InteractiveImportRowCellPlaceholder /> :
+              <InteractiveImportRowCellPlaceholder
+                className={styles.optional}
```
markus101, line 323: Instead of passing in the `classname` and duplicating the CSS from `InteractiveImportRowCellPlaceholder.css` to `Interactive/InteractiveImportRow.css`, pass in a property of `isOptional`, which will default to true. If true it should either use a different class for the component or combine classes with `classNames()` (we do this in other components already).

markus101 (changes requested):

markus101:
Fixed up and landed in 89e363fd145ae6c531e6b3fa4fde258af05afca0

PR #5421, opened 2023-02-08
Fixed: Show calculated (rather than stored) custom format score on season and series history, to match episode history

bpoxy:
#### Database Migration
NO

#### Description
@markus101 updated episode history to show freshly calculated custom format score (rather than the one stored at time of import) in #5292. However, series and season history still show the stored value. This PR updates series and season to display the freshly calculated score.

#### Todos
- [ ] Tests
- [ ] Wiki Updates

*

### DIFF
--- a/frontend/src/Series/History/SeriesHistoryRow.js
+++ b/frontend/src/Series/History/SeriesHistoryRow.js
@@ -191,8 +191,8 @@ SeriesHistoryRow.propTypes = {
   fullSeries: PropTypes.bool.isRequired,
   series: PropTypes.object.isRequired,
   episode: PropTypes.object.isRequired,
-  onMarkAsFailedPress: PropTypes.func.isRequired,
-  customFormatScore: PropTypes.number.isRequired
+  customFormatScore: PropTypes.number.isRequired,
+  onMarkAsFailedPress: PropTypes.func.isRequired
 };
 
 export default SeriesHistoryRow;
--- a/frontend/src/Series/History/SeriesHistoryRow.js
+++ b/frontend/src/Series/History/SeriesHistoryRow.js
@@ -191,7 +191,8 @@ SeriesHistoryRow.propTypes = {
   fullSeries: PropTypes.bool.isRequired,
   series: PropTypes.object.isRequired,
   episode: PropTypes.object.isRequired,
-  onMarkAsFailedPress: PropTypes.func.isRequired
+  onMarkAsFailedPress: PropTypes.func.isRequired,
+  customFormatScore: PropTypes.number.isRequired
 };
 
 export default SeriesHistoryRow;
--- a/frontend/src/Series/History/SeriesHistoryRow.js
+++ b/frontend/src/Series/History/SeriesHistoryRow.js
@@ -75,7 +75,8 @@ class SeriesHistoryRow extends Component {
       data,
       fullSeries,
       series,
-      episode
+      episode,
+      customFormatScore
     } = this.props;
 
     const {
@@ -145,7 +146,7 @@ class SeriesHistoryRow extends Component {
         <TableRowCell className={styles.customFormatScore}>
           <Tooltip
             anchor={
-              formatPreferredWordScore(data.customFormatScore, customFormats.length)
+              formatPreferredWordScore(customFormatScore, customFormats.length)
             }
             tooltip={<EpisodeFormats formats={customFormats} />}
             position={tooltipPositions.BOTTOM}


thezoggy:
if i have x CF setup and it grabbed something. a week later..  lets say made various changes to cf. wouldnt you not want to see what the score was when it grabbed? not what it would be if it was grabbed today.  as history should show the history context..

bpoxy:
> shouldnt the history show the CF that it was back then? as the current calculated value could be different if CF have been modified. which would get confusing if someone was looking at why something happen while comparing to current values

Personally, I like to see score based on my current rules so that I can determine whether it will be replaced by another release.
I can see the value in seeing the historical score too though, maybe we could show both on the UI? The values are both already accessible.

Currently, the episode and series/season views are different though. I think they should be consistent.

bakerboy448:
> Perhaps it makes sense to show the current score on the "Details" tab and the historical score on the "History" tab.

this makes the most sense. History is a snapshot of the point in time.  Details is live

bpoxy:
> > Perhaps it makes sense to show the current score on the "Details" tab and the historical score on the "History" tab.
> 
> this makes the most sense. History is a snapshot of the point in time. Details is live

There's a few issues with this approach though:
1. All pre-v4 history will have no custom format score to display.
2. The custom formats at the time of grab/import are not stored so a historical score would be shown next to current custom formats.

thezoggy:
> > > Perhaps it makes sense to show the current score on the "Details" tab and the historical score on the "History" tab.
> > 
> > 
> > this makes the most sense. History is a snapshot of the point in time. Details is live
> 
> There's a few issues with this approach though:
> 
>     1. All pre-v4 history will have no custom format score to display.
> 
>     2. The custom formats at the time of grab/import are not stored so a historical score would be shown next to current custom formats.

1. not sure how that is an issue, CF didnt exist until v4.

bpoxy:
> 1. not sure how that is an issue, CF didnt exist until v4.

That's a fair point. Personally, I do find it useful to see the scores of the pre-v4 grabs/imports though.
Curious on @markus101's thoughts on this, as he made the initial change.

markus101:
My understanding was Radarr showed the latest calculated score in the UI in all places, so my goal was to make that consistent. If that's the case I think keeping it consistent makes the most sense, then show the historic score (from grab/import time) in the details modal/popover. If Radarr doesn't show the current value then maybe we inverse that and show the historic value by default and the calculated value in details.

bpoxy:
> My understanding was Radarr showed the latest calculated score in the UI in all places, so my goal was to make that consistent. If that's the case I think keeping it consistent makes the most sense, then show the historic score (from grab/import time) in the details modal/popover. If Radarr doesn't show the current value then maybe we inverse that and show the historic value by default and the calculated value in details.

Radarr doesn't seem to show the calculated score anywhere.
The score at time of grab is shown in history along with the current custom formats.
The current custom formats are also shown on the "Files" tab.

I don't like this because (A) the calculated score isn't shown and (B) it seems inconsistent to show the current custom formats with a score that was potentially based on different custom formats.

I propose showing the historical score and historical custom formats in the history views (this would require storing custom format names in history) and showing the calculated score and current custom formats on the "Details" tab.

Thoughts?

markus101:
For now let's make it consistent with Radarr, which is what you did originally. We'll discuss if we want to change things later on.

You will need to fix this PR though:

> C:\BuildAgent\work\853acb111afc90d8\frontend\src\Series\History\SeriesHistoryRow.js
[21:56:07 ](https://build.sonarr.tv/buildConfiguration/Sonarr_V4PullRequests_1BuildFrontend/186772?buildTab=log&focusLine=152&logView=flowAware&linesState=152)      79:7  error  'customFormatScore' is missing in props validation  react/prop-types

bpoxy:
> For now let's make it consistent with Radarr, which is what you did originally. We'll discuss if we want to change things later on.
> 
> You will need to fix this PR though:
> 
> > C:\BuildAgent\work\853acb111afc90d8\frontend\src\Series\History\SeriesHistoryRow.js
> > [21:56:07 ](https://build.sonarr.tv/buildConfiguration/Sonarr_V4PullRequests_1BuildFrontend/186772?buildTab=log&focusLine=152&logView=flowAware&linesState=152)      79:7  error  'customFormatScore' is missing in props validation  react/prop-types

I actually did the opposite of make it consistent with Radarr since Radarr shows only historical score while my change shows calculated score. In order to be consistent with Radarr, we'd have to reject my changes and revert a change you made in https://github.com/Sonarr/Sonarr/pull/5292.

markus101:
Radarr changed it recently as well. https://github.com/Radarr/Radarr/commit/cbcf3d1058f47a87c43658084eafac13e7e0889f

bpoxy:
> Radarr changed it recently as well. [Radarr/Radarr@cbcf3d1](https://github.com/Radarr/Radarr/commit/cbcf3d1058f47a87c43658084eafac13e7e0889f)

Sorry, forgot that I'm not on Radarr's develop branch. :man_facepalming: :sweat_smile:

frontend/src/Series/History/SeriesHistoryRow.js
```diff
@@ -190,7 +191,8 @@ SeriesHistoryRow.propTypes = {
   fullSeries: PropTypes.bool.isRequired,
   series: PropTypes.object.isRequired,
   episode: PropTypes.object.isRequired,
-  onMarkAsFailedPress: PropTypes.func.isRequired
+  onMarkAsFailedPress: PropTypes.func.isRequired,
+  customFormatScore: PropTypes.number.isRequired
```
markus101, line 195: Put this above `onMarkAsFailedPress`

markus101 (changes requested):

PR #5433, opened 2023-02-16
Fixed: Improve Validation Failure Messaging

bakerboy448:
#### Database Migration
NO

#### Description

give users better instructions for where to look in the logs as many users only look at the validationfailure line saying to check the logs; ignoring the errors/details above it

#### Todos
- Tests
- Wiki Updates

* N/A


markus101:
This seems unnecessary and needlessly verbose, I don't think telling people to look above a specific error type will actually convince them to read the logs and even if they do, a lot of the time the error is complex enough that they're going to be asking for help (which is why we don't communicate the exact problem like we do in other cases).

bakerboy448:
typically they tend to just look (or more importantly) share the validation failure in lieu of some context above that, so the goal is to at least give a slight nudge to users who read the error message

fine to remove the exception messaging - figured it wouldn't hurt to give users the details explicitly to help reduce back and forth with gathering logs / helps with the issue of users only sharing the validation error and gives a hint as to the issue

src/NzbDrone.Core/ImportLists/HttpImportListBase.cs
```diff
@@ -239,7 +239,7 @@ protected virtual ValidationFailure TestConnection()
             {
                 _logger.Warn(ex, "Unable to connect to import list");
 
-                return new ValidationFailure(string.Empty, "Unable to connect to import list, check the log for more details");
+                return new ValidationFailure(string.Empty, "Unable to connect to import list, check the log above the ValidationFailure for more details");
```
markus101, line 242: Maybe something like `check the log surrounding this error for details`, that along with the actual error message will give them a better starting point.


```suggestion
                return new ValidationFailure(string.Empty, $"Unable to connect to import list: {ex.Message}. Check the log surrounding this error for details");
```

markus101 (changes requested):

PR #5446, opened 2023-02-22
New: Specials monitoring options

stevietv:
#### Database Migration
NO

#### Description
Given the removal of the season pass it can be arduous to mass monitor seasons using the season pass mouseover. The functionality is mostly covered by the option to mass edit the monitoring state, however if one wants to mass monitor specials there is no quick way anymore.

We never had it before but theoretically you could scroll down the season pass page and check every specials season.

I took a quick pass at this to see if there is interest.

#### Todos
- [X] Tests

*


stevietv:
Some options I had considered:

* make it "toggle specials monitoring" instead of only turning on monitoring.
* right now specials would be unmonitored if a different monitoring option is chosen. I'm ok with that, or we could rework the other monitoring options to only act on season > 0.

If this isn't something you're into I'm fine, was more for a bit of self-education also!

frontend/src/Utilities/Series/monitorOptions.js
```diff
@@ -6,6 +6,7 @@ const monitorOptions = [
   { key: 'pilot', value: 'Pilot Episode' },
   { key: 'firstSeason', value: 'Only First Season' },
   { key: 'latestSeason', value: 'Only Latest Season' },
+  { key: 'specials', value: 'Special Episodes'},
```
markus101, line 9: `A space is required before '}'  object-curly-spacing`

frontend/src/AddSeries/SeriesMonitoringOptionsPopoverContent.js
```diff
@@ -35,6 +35,11 @@ function SeriesMonitoringOptionsPopoverContent() {
         data="Monitor all episodes of the latest season and future seasons"
       />
 
+      <DescriptionListItem
+        title="Specials"
```
markus101, line 39: I think `Monitor Specials` with a description that indicates it will only affect special episodes would be better.

markus101 (changes requested):
I think it's useful, but needs some tweaks, thanks for diving in!

This seems like the most reasonable way to add the ability to monitor specials en masse without having + specials for every other option.

I think "specials" on it's own is too generic and probably needs a way to `Unmonitor Specials` without affecting other episodes.

frontend/src/Utilities/Series/monitorOptions.js
stevietv: fixed with yarn lint-fix
Special Episodes renamed to Monitor Specials

frontend/src/AddSeries/SeriesMonitoringOptionsPopoverContent.js
stevietv: This is reworded to Monitor specials with an explanation

stevietv:
> probably needs a way to Unmonitor Specials without affecting other episodes.

@markus101 - this is something I haven't tackled yet as wanted to check the direction you would prefer. I see two ways:

1. add another option to the dialog `Unmonitor Specials`
2.  rename this new option from `Monitor Specials` to `Toggle Specials Monitoring` and adjust the functionality so it toggles it on or off depending on current status.

I feel option 1 could be a cleaner implemenation as it'd not set some to true and some to false in one run (depending on current status)

src/NzbDrone.Core/Tv/MonitoringOptions.cs
```diff
@@ -19,6 +19,7 @@ public enum MonitorTypes
         FirstSeason,
         LatestSeason,
         Pilot,
+        Specials,
```
markus101, line 22: `MonitorSpecials` here as well

frontend/src/Utilities/Series/monitorOptions.js
```diff
@@ -6,6 +6,7 @@ const monitorOptions = [
   { key: 'pilot', value: 'Pilot Episode' },
   { key: 'firstSeason', value: 'Only First Season' },
   { key: 'latestSeason', value: 'Only Latest Season' },
+  { key: 'specials', value: 'Monitor Specials' },
```
markus101, line 9: This will need to be updated as well.

markus101 (changes requested):
I agree, option 1 is the better choice, that way like you said they'll be consistent instead of random.

stevietv:
Thanks Marcus - option 1 is added

frontend/src/Series/Index/Select/SeasonPass/ChangeMonitoringModalContent.tsx
```diff
@@ -48,7 +51,16 @@ function ChangeMonitoringModalContent(
       <ModalBody>
         <Form {...otherProps}>
           <FormGroup>
-            <FormLabel>{translate('Monitoring')}</FormLabel>
+            <FormLabel>
+              {translate('Monitoring')}
+
+              <Popover
+                anchor={<Icon className={styles.labelIcon} name={icons.INFO} />}
+                title="Monitoring Options"
```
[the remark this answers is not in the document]
stevietv: done

frontend/src/Series/MonitoringOptions/MonitoringOptionsModalContent.js
```diff
@@ -1,16 +1,20 @@
 import PropTypes from 'prop-types';
 import React, { Component } from 'react';
+import styles from 'AddSeries/AddNewSeries/AddNewSeriesModalContent.css';
```
markus101, line 3: Don't import CSS for another component, you'll need to create a new CSS file for this component.

stevietv: done! thanks for the guidance

frontend/src/Series/MonitoringOptions/MonitoringOptionsModalContent.css
```diff
@@ -0,0 +1,69 @@
...
+
+.info {
+  flex-grow: 1;
+}
+
+.overview {
+  margin-bottom: 30px;
+}
+
+.labelIcon {
```
markus101, line 24: This is the only class being used, everything else can be nuked, otherwise looks good.

markus101 (changes requested):
Just a little clean up of styles for the new CSS copied over

frontend/src/Series/MonitoringOptions/MonitoringOptionsModalContent.css
stevietv: done

markus101 (approved):
Thanks for the fixups!

PR #5469, opened 2023-03-03
Fixed: Mapping of parsed titles when one doesn't have an alias

msergein:
#### Database Migration
NO

#### Description
mapping all titles

#### Example
release: `Одни из нас (The Last of Us) S1E1-7 (2023)`
it is parsed as two title:
- `Одни из нас`
- `The Last of Us`

##### before fix
error: `Release Rejected: Unknown Series`
##### after fix: 
mapped

### DIFF
--- a/src/NzbDrone.Core/Parser/ParsingService.cs
+++ b/src/NzbDrone.Core/Parser/ParsingService.cs
@@ -93,13 +93,13 @@ namespace NzbDrone.Core.Parser
                 if (!tvdbId.HasValue)
                 {
                     _logger.Trace("Title {0} not matching any series.", title);
-                    return null;
+                    continue;
                 }
 
                 if (foundTvdbId.HasValue && tvdbId != foundTvdbId)
                 {
                     _logger.Trace("Title {0} both matches tvdbid {1} and {2}, no series selected.", parsedEpisodeInfo.SeriesTitle, foundTvdbId, tvdbId);
-                    return null;
+                    continue;
                 }
 
                 if (foundSeries == null)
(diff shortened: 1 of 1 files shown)


src/NzbDrone.Core/Parser/ParsingService.cs
```diff
@@ -93,13 +93,13 @@ private Series GetSeriesByAllTitles(ParsedEpisodeInfo parsedEpisodeInfo)
...
                     _logger.Trace("Title {0} not matching any series.", title);
-                    return null;
+                    continue;
                 }
 
                 if (foundTvdbId.HasValue && tvdbId != foundTvdbId)
                 {
                     _logger.Trace("Title {0} both matches tvdbid {1} and {2}, no series selected.", parsedEpisodeInfo.SeriesTitle, foundTvdbId, tvdbId);
-                    return null;
+                    continue;
```
markus101, line 102: Comment above the loop: 
```
Match each title individually, they must all resolve to the same tvdbid
```

Continuing here means this will no longer be enforced, which we don't want.

markus101 (changes requested):

src/NzbDrone.Core/Parser/ParsingService.cs
msergein: understood, thanks, returned as it was, but the first `continue` can be keep? If it doesn't find `tvdbId` for title - just skip it? This will help for localized titles

markus101: Yeah, I think that's reasonable, if it's unknown then it's ignored instead of blocking.

PR #5472, opened 2023-03-05
Fixed: Parsing of multi season packs with only space between the season numbers

stevietv:
#### Database Migration
NO

#### Description
Adds matching of a multi-season pack if there is only a space between the two season numbers.
Also strips the text "Complete Series" from a potential title match in multi-season packs/

#### Todos
- [X] Tests

*

### DIFF
--- a/src/NzbDrone.Core.Test/ParserTests/SeasonParserFixture.cs
+++ b/src/NzbDrone.Core.Test/ParserTests/SeasonParserFixture.cs
@@ -87,6 +87,7 @@ namespace NzbDrone.Core.Test.ParserTests
         [TestCase("Series Title S01 - S07 BluRay 1080p x264 REPACK -SacReD", "Series Title", 1)]
         [TestCase("Series Title Season 01-07 BluRay 1080p x264 REPACK -SacReD", "Series Title", 1)]
         [TestCase("Series Title Season 01 - Season 07 BluRay 1080p x264 REPACK -SacReD", "Series Title", 1)]
+        [TestCase("Series Title Complete Series S01 S04 (1080p BluRay x265 HEVC 10bit AAC 5.1 Vyndros)", "Series Title", 1)]
         public void should_parse_multi_season_release(string postTitle, string title, int firstSeason)
         {
             var result = Parser.Parser.ParseTitle(postTitle);
--- a/src/NzbDrone.Core.Test/ParserTests/SeasonParserFixture.cs
+++ b/src/NzbDrone.Core.Test/ParserTests/SeasonParserFixture.cs
@@ -88,6 +88,8 @@ namespace NzbDrone.Core.Test.ParserTests
         [TestCase("Series Title Season 01-07 BluRay 1080p x264 REPACK -SacReD", "Series Title", 1)]
         [TestCase("Series Title Season 01 - Season 07 BluRay 1080p x264 REPACK -SacReD", "Series Title", 1)]
         [TestCase("Series Title Complete Series S01 S04 (1080p BluRay x265 HEVC 10bit AAC 5.1 Vyndros)", "Series Title", 1)]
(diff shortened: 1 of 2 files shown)


src/NzbDrone.Core.Test/ParserTests/SeasonParserFixture.cs
```diff
@@ -87,6 +87,7 @@ public void should_parse_partial_season_release(string postTitle, string title,
         [TestCase("Series Title S01 - S07 BluRay 1080p x264 REPACK -SacReD", "Series Title", 1)]
         [TestCase("Series Title Season 01-07 BluRay 1080p x264 REPACK -SacReD", "Series Title", 1)]
         [TestCase("Series Title Season 01 - Season 07 BluRay 1080p x264 REPACK -SacReD", "Series Title", 1)]
+        [TestCase("Series Title Complete Series S01 S04 (1080p BluRay x265 HEVC 10bit AAC 5.1 Vyndros)", "Series Title", 1)]
```
markus101, line 90: Please add a case for just a space between the seasons.

src/NzbDrone.Core/Parser/Parser.cs
```diff
@@ -184,7 +184,7 @@ public static class Parser
                           RegexOptions.IgnoreCase | RegexOptions.Compiled),
 
                 // Multi-season pack
-                new Regex(@"^(?<title>.+?)[-_. ]+(?:S|(?:Season|Saison|Series|Stagione)[_. ])(?<season>(?<!\d+)(?:\d{1,2})(?!\d+))(?:-|[-_. ]{3})(?:S|(?:Season|Saison|Series|Stagione)[_. ])?(?<season>(?<!\d+)(?:\d{1,2})(?!\d+))",
+                new Regex(@"^(?<title>.+?)(Complete Series)?[-_. ]+(?:S|(?:Season|Saison|Series|Stagione)[_. ])(?<season>(?<!\d+)(?:\d{1,2})(?!\d+))(?:-|[-_. ]{3}|\s)(?:S|(?:Season|Saison|Series|Stagione)[_. ])?(?<season>(?<!\d+)(?:\d{1,2})(?!\d+))",
```
markus101, line 187: 1. I think it might be better to change the separator from `(?:-|[-_. ]{3}|\s)` to `(?:[-_. ]{1}|[-_. ]{3}|)` that way other space-like separators in release names are included.

markus101 (changes requested):

src/NzbDrone.Core.Test/ParserTests/SeasonParserFixture.cs
stevietv: added two more cases

src/NzbDrone.Core/Parser/Parser.cs
stevietv: refined the regex, I left out the final `|`  from your suggestion - not sure if you put that intentionally, but I was concerned that allowing no space would perhaps mess up those anime shows that are in 4 digit abs eps already.

markus101:
Thanks Stevie!

src/NzbDrone.Core.Test/ParserTests/SeasonParserFixture.cs
CLHatch: The test case with "S01 S04", is it supposed to recognize it as seasons 1 through 4, or as seasons 1 and 4?  Because I would think it would be 1 and 4.  I wouldn't think that specific one would show up, more likely they would be consecutive seasons, of course.

markus101: Sonarr expects either consecutive seasons or a range of inclusive seasons

PR #5493, opened 2023-03-17
New: Add indexer option for Discord on grab notifications

lodu:
#### Database Migration
NO

#### Description
When the OnGrab Discord notification is triggered it can now display the Indexer's name from the grab.

Minimal example of discord notification

#### Todos
- [-] Tests: [no tests](https://github.com/lodu/Sonarr/tree/discord-notification-grab-indexer/src/NzbDrone.Core.Test/NotificationTests)
- [-] Wiki Updates: [no wiki entry for fields](https://wiki.servarr.com/sonarr/supported#notifications)

### DIFF
--- a/src/NzbDrone.Core/Notifications/Discord/DiscordFieldType.cs
+++ b/src/NzbDrone.Core/Notifications/Discord/DiscordFieldType.cs
@@ -11,7 +11,8 @@ namespace NzbDrone.Core.Notifications.Discord
         Links,
         Release,
         Poster,
-        Fanart
+        Fanart,
+        Indexer
     }
 
     public enum DiscordImportFieldType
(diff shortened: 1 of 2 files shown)


src/NzbDrone.Core/Notifications/Discord/DiscordFieldType.cs
```diff
@@ -11,7 +11,8 @@ public enum DiscordGrabFieldType
         Links,
         Release,
         Poster,
-        Fanart
+        Fanart,
+        Indexer
```
lodu, line 15: Not sure if there is a specific order these should be in

markus101: Nope, at the end keeps them in the same order.

markus101:
Thanks!

PR #5504, opened 2023-03-24
Revert argument exception swallowing for Plex library update

markus101:
#### Database Migration
NO

#### Description
A better handling of path comparisons from different OSes which will fix Plex library path comparisons.


markus101:
Squashed and landed.

PR #5524, opened 2023-03-31
Fixed: Matching of custom formats during episode file import

bpoxy:
#### Database Migration
NO

#### Description
Fixed matching of Release Title custom formats during import by populating empty `CustomFormatInput` properties from `LocalEpisode` that are relied on when `SceneName` is not parsed.

#### Todos
- [ ] Tests
- [ ] Wiki Updates

* A release may be grabbed by Sonarr as a custom format upgrade then blocked with warnings indicating that the file(s) are not custom format upgrades for existing episode file(s).


markus101:
Thanks for the fix.

PR #5525, opened 2023-04-01
Migrate to fluent validation 9 and update SignalR

goremykin:
#### Database Migration
NO

#### Description
Migration to FlientValidation 9. See my comments below

#### Todos
- [x] Tests
- [ ] Wiki Updates

### DIFF
--- a/src/NzbDrone.Core.Test/NotificationTests/EmailTests/EmailSettingsValidatorFixture.cs
+++ b/src/NzbDrone.Core.Test/NotificationTests/EmailTests/EmailSettingsValidatorFixture.cs
@@ -62,7 +62,6 @@ namespace NzbDrone.Core.Test.NotificationTests.EmailTests
         }
 
         [TestCase("sonarr")]
-        [TestCase("sonarr@sonarr")]
         [TestCase("email.me")]
         [Ignore("Allowed coz some email servers allow arbitrary source, we probably need to support 'Name <email>' syntax")]
         public void should_not_be_valid_if_from_is_invalid(string email)
@@ -73,7 +72,6 @@ namespace NzbDrone.Core.Test.NotificationTests.EmailTests
         }
 
         [TestCase("sonarr")]
-        [TestCase("sonarr@sonarr")]
         [TestCase("email.me")]
         public void should_not_be_valid_if_to_is_invalid(string email)
         {
@@ -83,7 +81,6 @@ namespace NzbDrone.Core.Test.NotificationTests.EmailTests
         }
 
         [TestCase("sonarr")]
-        [TestCase("sonarr@sonarr")]
         [TestCase("email.me")]
         public void should_not_be_valid_if_cc_is_invalid(string email)
         {
@@ -93,7 +90,6 @@ namespace NzbDrone.Core.Test.NotificationTests.EmailTests
         }
 
         [TestCase("sonarr")]
-        [TestCase("sonarr@sonarr")]
         [TestCase("email.me")]
         public void should_not_be_valid_if_bcc_is_invalid(string email)
         {
--- a/src/NzbDrone.Core/Download/Clients/rTorrent/RTorrentDirectoryValidator.cs
+++ b/src/NzbDrone.Core/Download/Clients/rTorrent/RTorrentDirectoryValidator.cs
@@ -17,8 +17,10 @@ namespace NzbDrone.Core.Download.Clients.rTorrent
                                           PathExistsValidator pathExistsValidator,
                                           MappedNetworkDriveValidator mappedNetworkDriveValidator)
         {
+#pragma warning disable CS0618
             RuleFor(c => c.TvDirectory).Cascade(CascadeMode.StopOnFirstFailure)
-                                       .IsValidPath()
+#pragma warning restore CS0618
+                .IsValidPath()
                                        .SetValidator(rootFolderValidator)
                                        .SetValidator(mappedNetworkDriveValidator)
                                        .SetValidator(pathExistsValidator)
--- a/src/NzbDrone.Core/Download/Clients/rTorrent/RTorrentDirectoryValidator.cs
+++ b/src/NzbDrone.Core/Download/Clients/rTorrent/RTorrentDirectoryValidator.cs
@@ -17,9 +17,7 @@ namespace NzbDrone.Core.Download.Clients.rTorrent
                                           PathExistsValidator pathExistsValidator,
                                           MappedNetworkDriveValidator mappedNetworkDriveValidator)
         {
-#pragma warning disable CS0618
-            RuleFor(c => c.TvDirectory).Cascade(CascadeMode.StopOnFirstFailure)
-#pragma warning restore CS0618
+            RuleFor(c => c.TvDirectory).Cascade(CascadeMode.Stop)
                 .IsValidPath()
                                        .SetValidator(rootFolderValidator)
                                        .SetValidator(mappedNetworkDriveValidator)
--- a/src/NzbDrone.Core/ImportLists/Exclusions/ImportListExclusionExistsValidator.cs
+++ b/src/NzbDrone.Core/ImportLists/Exclusions/ImportListExclusionExistsValidator.cs
@@ -7,10 +7,11 @@ namespace NzbDrone.Core.ImportLists.Exclusions
         private readonly IImportListExclusionService _importListExclusionService;
 
         public ImportListExclusionExistsValidator(IImportListExclusionService importListExclusionService)
-            : base("This exclusion has already been added.")
         {
             _importListExclusionService = importListExclusionService;
         }
+
+        protected override string GetDefaultMessageTemplate() => "This exclusion has already been added.";
 
         protected override bool IsValid(PropertyValidatorContext context)
         {
--- a/src/NzbDrone.Core/Organizer/FileNameValidation.cs
+++ b/src/NzbDrone.Core/Organizer/FileNameValidation.cs
@@ -1,4 +1,3 @@
-using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using System.Text.RegularExpressions;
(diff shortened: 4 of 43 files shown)


src/NzbDrone.Core.Test/NotificationTests/EmailTests/EmailSettingsValidatorFixture.cs
```diff
@@ -62,7 +62,6 @@ public void should_not_be_valid_if_from_is_empty()
         }
 
         [TestCase("sonarr")]
-        [TestCase("sonarr@sonarr")]
```
goremykin, line 65: FluentValidation 9 switched to .NET Core’s `EmailAddressAttribute`:
https://docs.fluentvalidation.net/en/latest/upgrading-to-9.html

`"sonarr@sonarr"` is actually a valid email address:
https://en.wikipedia.org/wiki/Email_address#Examples

src/NzbDrone.Core/Download/Clients/rTorrent/RTorrentDirectoryValidator.cs
```diff
@@ -17,8 +17,10 @@ public class RTorrentDirectoryValidator : AbstractValidator<RTorrentSettings>, I
                                           PathExistsValidator pathExistsValidator,
                                           MappedNetworkDriveValidator mappedNetworkDriveValidator)
         {
+#pragma warning disable CS0618
```
goremykin, line 20: `CascadeMode.StopOnFirstFailure` is deprecated, but still exists in FluentValidation 11:
https://docs.fluentvalidation.net/en/latest/cascade.html#stop-vs-stoponfirstfailure

I tried `CascadeMode.Stop` - it works, our tests don't crash, but I decided not to change the behavior and suppress the warning. We can simulate `StopOnFirstFailure` behavior in FluentValidation 11 using `AbstractValidator.RuleLevelCascadeMode` and `AbstractValidator.ClassLevelCascadeMode`

src/NzbDrone.Core/ImportLists/Exclusions/ImportListExclusionExistsValidator.cs
```diff
@@ -7,11 +7,12 @@ public class ImportListExclusionExistsValidator : PropertyValidator
         private readonly IImportListExclusionService _importListExclusionService;
 
         public ImportListExclusionExistsValidator(IImportListExclusionService importListExclusionService)
-            : base("This exclusion has already been added.")
```
goremykin, line 10: Is deprecated, `GetDefaultMessageTemplate` should be used instead

src/NzbDrone.Core/Organizer/FileNameValidation.cs
```diff
@@ -66,96 +65,70 @@ public static class FileNameValidation
...
-        }
+        protected override string GetDefaultMessageTemplate() => "Must contain season and episode numbers OR Original Title";
 
         protected override bool IsValid(PropertyValidatorContext context)
         {
-            var value = context.PropertyValue as string;
-
-            if (!FileNameBuilder.SeasonEpisodePatternRegex.IsMatch(value) &&
-                !FileNameValidation.OriginalTokenRegex.IsMatch(value))
+            if (context.PropertyValue is not string value)
```
goremykin, line 72: Here ArgumentNullException was possible and I decided to check the type.
But I am not sure about this change. In some places null is a valid value, in some cases not.
So what do u think - should I return true or false in this case?

src/NzbDrone.Core/Organizer/FileNameValidation.cs
```diff
@@ -66,96 +65,70 @@ public static class FileNameValidation
...
+        protected override string GetDefaultMessageTemplate() => "Must contain Air Date OR Season and Episode OR Original Title";
 
         protected override bool IsValid(PropertyValidatorContext context)
         {
-            var value = context.PropertyValue as string;
-
-            if (!FileNameBuilder.SeasonEpisodePatternRegex.IsMatch(value) &&
-                !FileNameBuilder.AirDateRegex.IsMatch(value) &&
-                !FileNameValidation.OriginalTokenRegex.IsMatch(value))
+            if (context.PropertyValue is not string value)
```
goremykin, line 88: the same here

src/NzbDrone.Core/Organizer/FileNameValidation.cs
```diff
@@ -66,96 +65,70 @@ public static class FileNameValidation
...
+            "Must contain Absolute Episode number OR Season and Episode OR Original Title";
 
         protected override bool IsValid(PropertyValidatorContext context)
         {
-            var value = context.PropertyValue as string;
-
-            if (!FileNameBuilder.SeasonEpisodePatternRegex.IsMatch(value) &&
-                !FileNameBuilder.AbsoluteEpisodePatternRegex.IsMatch(value) &&
-                !FileNameValidation.OriginalTokenRegex.IsMatch(value))
+            if (context.PropertyValue is not string value)
```
goremykin, line 106: and here

src/NzbDrone.Core/Organizer/FileNameValidation.cs
```diff
@@ -66,96 +65,70 @@ public static class FileNameValidation
...
+            return FileNameBuilder.SeasonEpisodePatternRegex.IsMatch(value) ||
+                   FileNameBuilder.AbsoluteEpisodePatternRegex.IsMatch(value) ||
+                   FileNameValidation.OriginalTokenRegex.IsMatch(value);
         }
     }
 
     public class IllegalCharactersValidator : PropertyValidator
     {
-        private readonly char[] _invalidPathChars = Path.GetInvalidPathChars();
+        private static readonly char[] InvalidPathChars = Path.GetInvalidPathChars();
```
goremykin, line 119: No need to allocate a new array each time

src/NzbDrone.Integration.Test/Sonarr.Integration.Test.csproj
```diff
@@ -4,7 +4,7 @@
     <OutputType>Library</OutputType>
   </PropertyGroup>
   <ItemGroup>
-    <PackageReference Include="Microsoft.AspNetCore.SignalR.Client" Version="6.0.5" />
+    <PackageReference Include="Microsoft.AspNetCore.SignalR.Client" Version="6.0.15" />
```
goremykin, line 7: I think it's a safe change. All our AspNetCore packages are 6.0.15

src/Sonarr.Api.V3/Profiles/Quality/QualityCutoffValidator.cs
```diff
@@ -15,30 +15,17 @@ public static class QualityCutoffValidator
...
-            if (cutoffItem == null)
-            {
-                return false;
-            }
-
-            if (!cutoffItem.Allowed)
-            {
-                return false;
-            }
+            var cutoffItem = items?.SingleOrDefault(i => (i.Quality == null && i.Id == cutoff) || i.Quality?.Id == cutoff);
```
goremykin, line 26: The same behavior except the situation when items is null.
Previously, this would have been an null ref exception, but now we just return false.
So the same question here, should we return true instead?

src/Sonarr.Api.V3/Profiles/Quality/QualityItemsValidator.cs
```diff
@@ -23,139 +23,104 @@ public static class QualityItemsValidator
...
-        {
-        }
+        protected override string GetDefaultMessageTemplate() => "Individual qualities should not be named";
 
         protected override bool IsValid(PropertyValidatorContext context)
         {
-            var items = context.PropertyValue as IList<QualityProfileQualityItemResource>;
-
-            if (items.Any(i => i.Name.IsNotNullOrWhiteSpace() && i.Quality != null))
+            if (context.PropertyValue is not IList<QualityProfileQualityItemResource> items)
```
goremykin, line 56: The same here

src/Sonarr.Api.V3/Profiles/Quality/QualityItemsValidator.cs
```diff
@@ -23,139 +23,104 @@ public static class QualityItemsValidator
...
-        {
-        }
+        protected override string GetDefaultMessageTemplate() => "Groups must have a name";
 
         protected override bool IsValid(PropertyValidatorContext context)
         {
-            var items = context.PropertyValue as IList<QualityProfileQualityItemResource>;
-
-            if (items.Any(i => i.Quality == null && i.Name.IsNullOrWhiteSpace()))
+            if (context.PropertyValue is not IList<QualityProfileQualityItemResource> items)
```
goremykin, line 71: and here

src/Sonarr.Api.V3/Profiles/Quality/QualityItemsValidator.cs
```diff
@@ -23,139 +23,104 @@ public static class QualityItemsValidator
...
-        {
-        }
+        protected override string GetDefaultMessageTemplate() => "Groups must have an ID";
 
         protected override bool IsValid(PropertyValidatorContext context)
         {
-            var items = context.PropertyValue as IList<QualityProfileQualityItemResource>;
-
-            if (items.Any(i => i.Quality == null && i.Id == 0))
+            if (context.PropertyValue is not IList<QualityProfileQualityItemResource> items)
```
goremykin, line 86: and here

src/Sonarr.Api.V3/Profiles/Quality/QualityItemsValidator.cs
```diff
@@ -23,139 +23,104 @@ public static class QualityItemsValidator
...
-        {
-        }
+        protected override string GetDefaultMessageTemplate() => "Groups must have a unique ID";
 
         protected override bool IsValid(PropertyValidatorContext context)
         {
-            var items = context.PropertyValue as IList<QualityProfileQualityItemResource>;
-
-            if (items.Where(i => i.Id > 0).Select(i => i.Id).GroupBy(i => i).Any(g => g.Count() > 1))
+            if (context.PropertyValue is not IList<QualityProfileQualityItemResource> items)
```
goremykin, line 101: and here

src/Sonarr.Api.V3/Profiles/Quality/QualityItemsValidator.cs
```diff
@@ -23,139 +23,104 @@ public static class QualityItemsValidator
...
             {
                 return false;
             }
 
-            return true;
+            var ids = items.Where(i => i.Id > 0).Select(i => i.Id);
```
goremykin, line 106: The same behavior, I just tried to make it more readable

src/Sonarr.Api.V3/Profiles/Quality/QualityItemsValidator.cs
```diff
@@ -23,139 +23,104 @@ public static class QualityItemsValidator
...
-        public UniqueQualityIdValidator()
-            : base("Qualities can only be used once")
-        {
-        }
+        protected override string GetDefaultMessageTemplate() => "Qualities can only be used once";
 
         protected override bool IsValid(PropertyValidatorContext context)
         {
-            var items = context.PropertyValue as IList<QualityProfileQualityItemResource>;
+            if (context.PropertyValue is not IList<QualityProfileQualityItemResource> items)
```
goremykin, line 119: the same here

src/Sonarr.Api.V3/Series/SeriesFolderAsRootFolderValidator.cs
```diff
@@ -11,19 +11,20 @@ public class SeriesFolderAsRootFolderValidator : PropertyValidator
...
+
         protected override bool IsValid(PropertyValidatorContext context)
         {
             if (context.PropertyValue == null)
             {
                 return true;
             }
 
-            var seriesResource = context.Instance as SeriesResource;
+            var seriesResource = context.InstanceToValidate as SeriesResource;
```
goremykin, line 27: From the documentation:
`Instance` was removed from `PropertyValidatorContext` - use `InstanceToValidate` instead

src/Sonarr.Api.V3/Series/SeriesFolderAsRootFolderValidator.cs
```diff
@@ -37,7 +38,7 @@ protected override bool IsValid(PropertyValidatorContext context)
                 return true;
             }
 
-            var rootFolder = new DirectoryInfo(rootFolderPath).Name;
+            var rootFolder = new DirectoryInfo(rootFolderPath!).Name;
```
goremykin, line 41: suppress the warning, it cannot be null

src/Sonarr.Http/REST/ResourceValidator.cs
```diff
@@ -15,7 +14,7 @@ public class ResourceValidator<TResource> : AbstractValidator<TResource>
         {
             var rule = new PropertyRule(fieldListAccessor.GetMember(), c => GetValue(c, fieldListAccessor.Compile(), fieldName), null, () => CascadeMode.Continue, typeof(TProperty), typeof(TResource));
             rule.PropertyName = fieldName;
-            rule.DisplayName = new StaticStringSource(fieldName);
```
goremykin, line 18: It was removed, `SetDisplayName` should be used

src/NzbDrone.Core/Organizer/FileNameValidation.cs
markus101: I think it makes sense to be false if null in this case.

src/Sonarr.Api.V3/Profiles/Quality/QualityCutoffValidator.cs
markus101: I think false is fine if the expected list is null.

src/NzbDrone.Integration.Test/Sonarr.Integration.Test.csproj
Qstick: Make sure to bump the frontend package to match, just to maintain consistency

https://github.com/Sonarr/Sonarr/blob/6ea3d8c127eafbcf9d1b6e9338b737e91e256875/package.json#L33

goremykin: @Qstick, Nice point!

markus101:
With FluentValidation 11 out is the upgrade to 9 an interim step to going to 10 then 11? Just thinking about the best way to handle this as I think it might make the most sense to do it in chunks, but land it all at once.

goremykin:
@markus101, Initially I wanted to migrate to 11, but then i realized that migration to 10 requires a lot of effort.
I decided to migrate to version 9 because I do not know when I will have so much time

src/NzbDrone.Core/Download/Clients/rTorrent/RTorrentDirectoryValidator.cs
markus101: If `Stop` is working then I think it makes sense to use that instead of using the deprecated `StopOnFirstFailure` and having to disable rules on it. No point making this a future us problem if we can solve it now.

markus101 (changes requested):
Talked with @Qstick snd we think it’d be good to move to `Stop` and avoid the deprecated warning avoidance.

Also looks like there is another signalR bump that we might as well include.

markus101:
Thanks for the fixes, I've squashed and merged into 2 commits.

PR #5526, opened 2023-04-01
Use MinBy and MaxBy instead of OrderBy + First

goremykin:
#### Database Migration
NO

#### Description
This PR replaces `OrderBy().First()` by less resource-intensive `MinBy` and `MaxBy`.

Very simple benchmark:
```
public class Bench
{
    private IEnumerable<int> numbers = Enumerable.Range(0, 100).OrderBy(_ => Random.Shared.Next()).ToArray();

    [Benchmark]
    public void Order()
    {
        var max = numbers.OrderBy(_ => _).First();
    }
    
    [Benchmark]
    public void Min()
    {
        var max = numbers.MinBy(_ => _);
    }
}
```

```
| Method |     Mean |   Error |  StdDev |   Gen0 | Allocated |
|------- |---------:|--------:|--------:|-------:|----------:|
|  Order | 807.8 ns | 1.68 ns | 1.49 ns | 0.0010 |     128 B |
|    Min | 512.0 ns | 9.53 ns | 7.96 ns |      - |      32 B |
```

#### Todos
- [ ] Tests
- [ ] Wiki Updates

### DIFF
--- a/src/NzbDrone.Core/Download/Pending/PendingReleaseService.cs
+++ b/src/NzbDrone.Core/Download/Pending/PendingReleaseService.cs
@@ -10,7 +10,6 @@ using NzbDrone.Core.DecisionEngine;
 using NzbDrone.Core.Download.Aggregation;
 using NzbDrone.Core.Indexers;
 using NzbDrone.Core.Jobs;
-using NzbDrone.Core.Languages;
 using NzbDrone.Core.Messaging.Events;
 using NzbDrone.Core.Parser;
 using NzbDrone.Core.Parser.Model;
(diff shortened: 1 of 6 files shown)


src/NzbDrone.Core/Download/Pending/PendingReleaseService.cs
```diff
@@ -97,7 +96,7 @@ public void AddMany(List<Tuple<DownloadDecision, PendingReleaseReason>> decision
 
                     var episodeIds = decision.RemoteEpisode.Episodes.Select(e => e.Id);
 
-                    var existingReports = episodeIds.SelectMany(v => alreadyPendingByEpisode[v] ?? Enumerable.Empty<PendingRelease>())
```
goremykin, line 100: `alreadyPendingByEpisode[v]` cannot be null as `ILookup` never return null. It returns an empty enumerable instead

PR #5538, opened 2023-04-05
Import via Script

### LINKED ISSUE #791: Ability to execute a script before import (Pre-import script)
Would be great if sonarr was able to use Syncthing's API to determine when a synced folder has finished so it can scan for files to import at that time. It seems both Syncthing's and BTSync's api could be used to easily determine that info. Since Syncthing is open source it would probably be better to go that route over BTSync. Could there just be an option to have sonarr make an api call to syncthing before running downloaded episodes scan to see if files are still syncing or not? [Here](http://docs.syncthing.net/dev/events.html) is Syncthing's API.

This would be great for those systems that don't support an SSHFS mount.

JeWe37:
#### Database Migration
NO

#### Description
This pull request selecting a script that will be run instead of Sonarr's standard import system for both regular episode files and extras. This is useful if the user wants to for instance transcode or remux in the process of copying from the downloads to the library folder. Using the returncode 2, control of copying/hardlinking/moving can be returned to Sonarr.

The environment variables passed to the script are largely inspired by the `OnDownload` connection script type. Currently I am still leaving out whatever information couldn't as easily be obtained at this point in the code, however this might be subject to change.

This is a draft, as such there remain a few open questions:
- Which environment variables might and might not make sense to still retain for the scripts? Ultimately all information can be obtained via either running ffprobe in the script or using the API, as such one is flexible one way or another. Environment variables however do provide a convenient means of getting at least basic information.
- This question already came up in the issue: If one uses the script to do a remux whilst copying it is possible that a rescan and rename is required afterwards in order to get everything into sync. Should I consider that out of scope(seeing as the script could always just call these via the API if it needs to) or should support be directly integrated into Sonarr itself? Say via a special return code, similar to the return code for telling Sonarr to handle the file that has already been implemented.
- How exactly should I, if at all, go about fixturing and testing this code? In so far as I can see the extras service didn't have any tests previously either and the episode mover only had very few. What is the policy exactly in that respect?

I'll also need to clean up the commit. I'll also say that I come from the C++ world and have written hardly any C#, as such any notes on style would of course be welcome.

So far I've only tested this very briefly, and while it worked there, some more thorough testing will be required.

#### Todos
- [ ] Tests
- [ ] Wiki Updates

* Fixes #791


JeWe37:
> This question already came up in the issue: If one uses the script to do a remux whilst copying it is possible that a rescan and rename is required afterwards in order to get everything into sync. Should I consider that out of scope(seeing as the script could always just call these via the API if it needs to) or should support be directly integrated into Sonarr itself? Say via a special return code, similar to the return code for telling Sonarr to handle the file that has already been implemented.

I have been doing some investigating here, this could be solved by simply calling `UpdateMediaInfoService` and then using a (perhaps slightly adapted version of) `RenameEpisodeFileService.RenameFiles`. Probably worthwhile to add this as this will be more efficient than a script triggering a rescan/rename.

Currently can't test, I'll get to checking which of the environment variables matter once I can.

.editorconfig
```diff
@@ -51,6 +51,7 @@ dotnet_diagnostic.SA1101.severity = none
 dotnet_diagnostic.SA1116.severity = none
 dotnet_diagnostic.SA1118.severity = none
 dotnet_diagnostic.SA1122.severity = none
+dotnet_diagnostic.SA1200.severity = none
```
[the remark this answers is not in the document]
JeWe37: Yes, I'm aware, for some reason Sonarr wasn't building for me with this on(on Linux), I'll remove this for the final version.

src/NzbDrone.Core/MediaFiles/ScriptImportDecider.cs
```diff
@@ -0,0 +1,117 @@
...
+            // }
+
+            _logger.Debug("Executing external script: {0}", _configService.ScriptImportPath);
+
+            var processOutput = _processProvider.StartAndCapture(_configService.ScriptImportPath, $"\"{sourcePath}\" \"{destinationFilePath}\"", environmentVariables);
+
+            _logger.Debug("Executed external script: {0} - Status: {1}", _configService.ScriptImportPath, processOutput.ExitCode);
+            _logger.Debug("Script Output: \r\n{0}", string.Join("\r\n", processOutput.Lines));
+
+            if (processOutput.ExitCode == 1)
```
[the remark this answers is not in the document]
JeWe37: That already happens actually, the boolean return value indicates whether we are deferring to Sonarr

frontend/src/Settings/MediaManagement/MediaManagement.js
```diff
@@ -200,6 +200,40 @@ class MediaManagement extends Component {
...
+                          type={inputTypes.CHECK}
+                          name="scriptImport"
+                          helpText="Copy files for importing using a script, say for transcoding"
+                          onChange={onInputChange}
+                          {...settings.scriptImport}
+                        />
+                      </FormGroup>
+
+                      {
+                        settings.scriptImport.value &&
```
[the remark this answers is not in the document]
JeWe37: Copy pasted that from just below in that file, I would say either keep it consistent or also change that one?

src/NzbDrone.Core/MediaFiles/ScriptImportDecider.cs
```diff
@@ -0,0 +1,117 @@
...
+            environmentVariables.Add("Sonarr_EpisodeFile_EpisodeOverviews", string.Join("|", episodeFile.Episodes.Value.Select(e => e.Overview)));
+            environmentVariables.Add("Sonarr_EpisodeFile_Quality", episodeFile.Quality.Quality.Name);
+            environmentVariables.Add("Sonarr_EpisodeFile_QualityVersion", episodeFile.Quality.Revision.Version.ToString());
+            environmentVariables.Add("Sonarr_EpisodeFile_ReleaseGroup", episodeFile.ReleaseGroup ?? string.Empty);
+            environmentVariables.Add("Sonarr_EpisodeFile_SceneName", episodeFile.SceneName ?? string.Empty);
+
+            // environmentVariables.Add("Sonarr_Download_Client", message.DownloadClientInfo?.Name ?? string.Empty);
+            // environmentVariables.Add("Sonarr_Download_Client_Type", message.DownloadClientInfo?.Type ?? string.Empty);
+            // environmentVariables.Add("Sonarr_Download_Id", message.DownloadId ?? string.Empty);
+            environmentVariables.Add("Sonarr_EpisodeFile_MediaInfo_AudioChannels", MediaInfoFormatter.FormatAudioChannels(episodeFile.MediaInfo).ToString());
```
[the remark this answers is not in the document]
JeWe37: Yes I definitely should check this in more detail still.

src/NzbDrone.Core/MediaFiles/ScriptImportDecider.cs
```diff
@@ -0,0 +1,117 @@
...
+                                   IConfigFileProvider configFileProvider,
+                                   Logger logger)
+        {
+            _processProvider = processProvider;
+            _configService = configService;
+            _configFileProvider = configFileProvider;
+            _logger = logger;
+        }
+
+        public bool TryImport(EpisodeFile episodeFile, Series series, string sourcePath, string destinationFilePath, TransferMode mode)
```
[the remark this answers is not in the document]
JeWe37: Hmmm the only major difference is potentially in the environment vars you might want. I'm thinking at least some shared functionality makes sense.

frontend/src/Settings/MediaManagement/MediaManagement.js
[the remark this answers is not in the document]
markus101: Preferably replace them all, but at least one new code.

src/NzbDrone.Core/MediaFiles/ScriptImportDecider.cs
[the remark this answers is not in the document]
markus101: My comment about other script options is irrelevant, it's beyond the scope of this currently.

frontend/src/Settings/MediaManagement/MediaManagement.js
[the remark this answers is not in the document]
JeWe37: Changed all of them within the file, left them as they were in other files.

src/NzbDrone.Core/MediaFiles/ScriptImportDecider.cs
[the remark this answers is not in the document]
JeWe37: I went over all environment variables and all of them with the exception of those still commented out in the current version made sense. Getting all data required to actually add them was somewhat cumbersome, if you have an idea for how could be made easier do let me know.

src/NzbDrone.Core/MediaFiles/ScriptImportDecider.cs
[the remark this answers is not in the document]
JeWe37: Alright, it now works as follows:
0: Successful import. We're done.
2: Request the file to be renamed after import.
3: Defer all work to Sonarr.
Anything else: fail import

I intentionally skipped 1 as it is a common error code when a command just fails. We don't want to be mistaking failures for renames or deferring too much.

JeWe37:
Alright so, the latest commit broadly implements everything that I wanted to add. It is however still horribly messy code and in no state to get merged, cleanup will be the next step. For that reason I also didn't fix up the existing tests yet, as I am not totally sure which interfaces might change again in that process. Any ideas for how to do that would be much appreciated.

One particularly nasty bit is currently that the request for renaming has to be passed far enough up the stack for the file to first be added to the `MediaFileService`. I intentionally chose to avoid just publishing an event as those trigger in this case potentially unwanted side effects. I might investigate if I can't adjust the renaming events to handle this better.

src/NzbDrone.Core/MediaFiles/EpisodeFileMovingService.cs
```diff
@@ -51,12 +53,13 @@ public class EpisodeFileMovingService : IMoveEpisodeFiles
...
             _diskProvider = diskProvider;
             _mediaFileAttributeService = mediaFileAttributeService;
+            _scriptImportDecider = scriptImportDecider;
             _eventAggregator = eventAggregator;
             _configService = configService;
             _logger = logger;
         }
 
-        public EpisodeFile MoveEpisodeFile(EpisodeFile episodeFile, Series series)
+        public (EpisodeFile, bool) MoveEpisodeFile(EpisodeFile episodeFile, Series series)
```
markus101, line 62: This case is only used for existing episode files, already in the series folder, I don't think we want to handle these renames with an "import" script.

src/NzbDrone.Core/MediaFiles/EpisodeFileMovingService.cs
```diff
@@ -68,18 +71,18 @@ public EpisodeFile MoveEpisodeFile(EpisodeFile episodeFile, Series series)
             return TransferFile(episodeFile, series, episodes, filePath, TransferMode.Move);
         }
 
-        public EpisodeFile MoveEpisodeFile(EpisodeFile episodeFile, LocalEpisode localEpisode)
+        public (EpisodeFile, bool) MoveEpisodeFile(EpisodeFile episodeFile, LocalEpisode localEpisode, ScriptImportDecisionInfo scriptImportDecisionInfo)
```
markus101, line 74: A lot of this information is already part of `LocalEpisode`, if there is anything missing could it be added instead of passing along values for a specific import method that may not be used?

src/NzbDrone.Core/MediaFiles/EpisodeFileMovingService.cs
```diff
@@ -68,18 +71,18 @@ public EpisodeFile MoveEpisodeFile(EpisodeFile episodeFile, Series series)
...
             EnsureEpisodeFolder(episodeFile, localEpisode, filePath);
 
             _logger.Debug("Moving episode file: {0} to {1}", episodeFile.Path, filePath);
 
-            return TransferFile(episodeFile, localEpisode.Series, localEpisode.Episodes, filePath, TransferMode.Move);
+            return TransferFile(episodeFile, localEpisode.Series, localEpisode.Episodes, filePath, TransferMode.Move, scriptImportDecisionInfo);
         }
 
-        public EpisodeFile CopyEpisodeFile(EpisodeFile episodeFile, LocalEpisode localEpisode)
+        public (EpisodeFile, bool) CopyEpisodeFile(EpisodeFile episodeFile, LocalEpisode localEpisode, ScriptImportDecisionInfo scriptImportDecisionInfo)
```
markus101, line 85: Not really a fan of using a tuple here, though I'm not sure it's necessary, episodes that haven't imported yet don't aren't committed to the DB so the path of the episode file that was imported could be changed.

src/NzbDrone.Core/MediaFiles/EpisodeFileMovingService.cs
```diff
@@ -113,7 +117,33 @@ private EpisodeFile TransferFile(EpisodeFile episodeFile, Series series, List<Ep
                 throw new SameFilenameException("File not moved, source and destination are the same", episodeFilePath);
             }
 
-            _diskTransferService.TransferFile(episodeFilePath, destinationFilePath, mode);
+            bool needsRename = false;
+            bool transfer = true;
+
+            if (scriptImportDecisionInfo is not null)
```
markus101, line 123: I think this logic might be better suited for `UpgradeMediaFileService`, it'll avoid having to pass information back to it and it already knows whether it's going to be a move or copy.

src/NzbDrone.Core/MediaFiles/EpisodeImport/ImportApprovedEpisodes.cs
```diff
@@ -130,9 +149,19 @@ public List<ImportResult> Import(List<ImportDecision> decisions, bool newDownloa
                     episodeFile = _mediaFileService.Add(episodeFile);
                     importResults.Add(new ImportResult(importDecision));
 
+                    if (rename)
```
markus101, line 152: I'm not sure I understand why this is necessary, the EpisodeFile isn't committed to the DB until line 149, any changes to episode file would then be committed, instead this waits for that to happen then immediately updates it.

src/NzbDrone.Core/MediaFiles/ScriptImportDecider.cs
```diff
@@ -0,0 +1,143 @@
...
+        {
+            _processProvider = processProvider;
+            _configService = configService;
+            _configFileProvider = configFileProvider;
+            _logger = logger;
+        }
+
+        public ScriptImportDecision TryImport(string sourcePath, string destinationFilePath, ScriptImportDecisionInfo scriptImportDecisionInfo)
+        {
+            EpisodeFile episodeFile = scriptImportDecisionInfo.episodeFile;
```
markus101, line 43: use `var` instead of specifying the type.

markus101 (changes requested):
Looks like where things are being handled is leading to more passing of information than is needed and some unnecessary calls/updates.

The extra files are going to be a bit interesting since a script to rename them is probably unnecessary a lot of the time.

src/NzbDrone.Core/MediaFiles/EpisodeFileMovingService.cs
JeWe37: I've been wondering about this too. Yes a lot of it is already in LocalEpisode, but basically everything added in this commit needed to come from elsewhere. I need to try the other way and see if it's alright, but I don't love having functions with 5+ parameters only used like once inside them.

src/NzbDrone.Core/MediaFiles/EpisodeFileMovingService.cs
JeWe37: I noticed the same thing. I've been wondering if really the `EpisodeFileMover` should be broken up into two classes, the `EpisodeFikeMover` itself and perhaps an `EpisodeFileRenamer`, probably still sharing an abstract base with `TransferFile`. At one point I was running into recursive dependencies die zu that but that rectified itself in the process of fixing other issues. I might still break it up entirely if it becomes necessary.

JeWe37:
> The extra files are going to be a bit interesting since a script to rename them is probably unnecessary a lot of the time.

I don't really see a usecase for it either really. It might be best to remove the functionality. What would be nice is being able to add extras from the script, e.g. extracted subtitles. The script would need some way to tell Sonarr about those though. Is there anything available in
Sonarr for(non network) IPC? If not, named pipes I would imagine would generally be the easiest to handle from a script. I don't really want to impose requirements on stdout/stderr.

src/NzbDrone.Core/MediaFiles/EpisodeImport/ImportApprovedEpisodes.cs
markus101: Why would a file that's just been imported need to immediately be renamed?

If the file name has changed that should be updated before committing, if it needs to be renamed after importing can we not calculate that before committing?

JeWe37: Yes, that's broadly what I was thinking of wrt. "lower level methods"-using the basic "update media info, rename file" verbs that I am using currently relies on the file already being committed. But I'm guessing one can pull apart their functionality to actually do this to uncommitted files.

markus101:
The only interface is via the API, no IPC.

I think the easiest would just be name them like the video file and Sonarr can import them, you could import with script and tell Sonarr to rescan, but then things are going to get more complicated.

JeWe37:
> I think the easiest would just be name them like the video file and Sonarr can import them, you could import with script and tell Sonarr to rescan, but then things are going to get more complicated.

Well so far I have been trying to avoid triggering full series rescan/rename for the import of potentially only a single file. Honestly using the return codes we have so far is already somewhat of a hack though, really some better way to communicate back to Sonarr might be nice. But honestly yes, this isn't a priority for now and could be dealt with in a later PR.

JeWe37:
> I think the easiest would just be name them like the video file and Sonarr can import them, you could import with script and tell Sonarr to rescan, but then things are going to get more complicated.

Is that actually possible at the moment? It seems to me like the functionality for adding extras on rescan is currently not implemented, see [here](https://github.com/Sonarr/Sonarr/blob/6413d212e587461803b03df512e9471af08d0b5b/src/NzbDrone.Core/Extras/Others/OtherExtraService.cs#L39-L57) and [here](https://github.com/Sonarr/Sonarr/blob/6413d212e587461803b03df512e9471af08d0b5b/src/NzbDrone.Core/Extras/Subtitles/SubtitleService.cs#L46-L64)

JeWe37:
Alright, these four commits start to do some of the required cleanup. In order:
- The moving code has been simplified by using lower level methods that do not require the episode to have been committed yet
- Extras can be rejected and won't be imported by returning return code 2, renaming wouldn't have made sense for them anyways.
- When importing subtitle files using a script that script is given some additional information about the subtitle files language from Sonarr
- To clean up the structure somewhat, EpisodeFileMovingService was broken into a part for renaming and one for moving, basically importing, them. This is optional and really just clean up, if you do not like it that can be removed.

In general I have removed any null-annotations as they do not seem to be used anywhere else throughout the codebase. This is now in a decently clean state. Collecting the information for the environment variables is still perhaps not terribly pretty, but I frankly cannot see how this could be improved.

I'll wait for your review before I get to actually fixing up the existing tests, some of which were broken by all this, and commit history.

src/NzbDrone.Core/MediaFiles/EpisodeFileMovingService.cs
```diff
@@ -1,85 +1,85 @@
...
-            _logger.Debug("Renaming episode file: {0} to {1}", episodeFile, filePath);
-
-            return TransferFile(episodeFile, series, episodes, filePath, TransferMode.Move);
+            switch (scriptImportDecision)
+            {
+                case ScriptImportDecision.DeferMove:
+                    _diskTransferService.TransferFile(episodeFilePath, destinationFilePath, mode);
+                    break;
+                case ScriptImportDecision.RenameRequested:
+                    _updateMediaInfo.Update(episodeFile, series, false);
```
markus101, line 62: Why would the script import the file and then immediately need it to be renamed and moved? If it needs to be moved why can't the script do that and communicate the new file name or instead of moving it, leave it and let Sonarr handle it?

src/NzbDrone.Core/MediaFiles/EpisodeFileMovingService.cs
```diff
@@ -1,85 +1,85 @@
...
-using NzbDrone.Core.Tv;
 
 namespace NzbDrone.Core.MediaFiles
 {
     public interface IMoveEpisodeFiles
     {
-        EpisodeFile MoveEpisodeFile(EpisodeFile episodeFile, Series series);
-        EpisodeFile MoveEpisodeFile(EpisodeFile episodeFile, LocalEpisode localEpisode);
-        EpisodeFile CopyEpisodeFile(EpisodeFile episodeFile, LocalEpisode localEpisode);
+        EpisodeFile MoveEpisodeFile(EpisodeFile episodeFile, LocalEpisode localEpisode, ScriptImportDecisionInfo scriptImportDecisionInfo);
```
markus101, line 14: Most of the information in `ScriptImportDecisionInfo` is already part of `localEpisode`, including `LocalEpisode` itself, there is no need to pass it in to a function as a separate parameter, if more information is necessary then it should be added to `LocalEpisode`.

src/NzbDrone.Core/MediaFiles/EpisodeImport/ImportApprovedEpisodes.cs
```diff
@@ -132,7 +141,8 @@ public List<ImportResult> Import(List<ImportDecision> decisions, bool newDownloa
 
                     if (newDownload)
                     {
-                        _extraService.ImportEpisode(localEpisode, episodeFile, copyOnly);
+                        scriptImportDecisionInfo.isEpisodeFile = false;
```
markus101, line 144: Extra service should already know that it's not importing an episode file, this shouldn't be mutated and re-used.

src/NzbDrone.Core/MediaFiles/MediaInfo/UpdateMediaInfoService.cs
```diff
@@ -55,18 +55,18 @@ public void Handle(SeriesScannedEvent message)
...
                 _logger.Debug("MediaInfo is disabled");
                 return false;
             }
 
-            return UpdateMediaInfo(episodeFile, series);
+            return UpdateMediaInfo(episodeFile, series, update);
         }
 
-        private bool UpdateMediaInfo(EpisodeFile episodeFile, Series series)
+        private bool UpdateMediaInfo(EpisodeFile episodeFile, Series series, bool update = true)
```
markus101, line 69: If the mediainfo for a file needs to be reprocessed before the file in in the DB that shouldn't be done with this, which is meant to update media info for files already stored in the DB.

src/NzbDrone.Core/MediaFiles/ScriptImportDecisionInfo.cs
```diff
@@ -0,0 +1,35 @@
...
+using NzbDrone.Core.Download;
+using NzbDrone.Core.Extras.Subtitles;
+using NzbDrone.Core.Parser.Model;
+
+namespace NzbDrone.Core.MediaFiles
+{
+    public class ScriptImportDecisionInfo
+    {
+        // needed
+        public bool isEpisodeFile { get; set; }
```
markus101, line 12: This file should be going away, but properties should be CamelCase like they are in other instances.

src/NzbDrone.Core/MediaFiles/ScriptImportDecider.cs
```diff
@@ -0,0 +1,144 @@
...
+            environmentVariables.Add("Sonarr_EpisodeFile_CustomFormatScore", localEpisode.CustomFormatScore.ToString());
+
+            if (oldFiles.Any())
+            {
+                environmentVariables.Add("Sonarr_DeletedRelativePaths", string.Join("|", oldFiles.Select(e => e.RelativePath)));
+                environmentVariables.Add("Sonarr_DeletedPaths", string.Join("|", oldFiles.Select(e => Path.Combine(series.Path, e.RelativePath))));
+                environmentVariables.Add("Sonarr_DeletedDateAdded", string.Join("|", oldFiles.Select(e => e.DateAdded)));
+            }
+
+            if (subtitleFile is not null)
```
markus101, line 109: This should have it's own function, a common function to gather up the common variables makes sense, but additional variables can be handled separately

src/NzbDrone.Core/Extras/Subtitles/SubtitleService.cs
```diff
@@ -207,16 +208,21 @@ public override IEnumerable<ExtraFile> ImportFiles(LocalEpisode localEpisode, Ep
                     var suffix = GetSuffix(language, copy, file.LanguageTags, groupCount > 1);
                     try
                     {
-                        var subtitleFile = ImportFile(localEpisode.Series, episodeFile, path, isReadOnly, extension, suffix);
-                        subtitleFile.Language = language;
-                        subtitleFile.LanguageTags = file.LanguageTags;
+                        scriptImportDecisionInfo.subtitleFile = file;
+                        var subtitleFile = ImportFile(localEpisode.Series, episodeFile, scriptImportDecisionInfo, path, isReadOnly, extension, suffix);
+                        scriptImportDecisionInfo.subtitleFile = null;
+                        if (subtitleFile is not null)
```
markus101, line 214: Nit: newline before `if`

src/NzbDrone.Core/Extras/Others/OtherExtraService.cs
```diff
@@ -137,10 +138,13 @@ public override IEnumerable<ExtraFile> ImportFiles(LocalEpisode localEpisode, Ep
             {
                 try
                 {
-                    var extraFile = ImportFile(localEpisode.Series, episodeFile, file, isReadOnly, Path.GetExtension(file), null);
-                    _mediaFileAttributeService.SetFilePermissions(file);
-                    _otherExtraFileService.Upsert(extraFile);
-                    importedFiles.Add(extraFile);
+                    var extraFile = ImportFile(localEpisode.Series, episodeFile, scriptImportDecisionInfo, file, isReadOnly, Path.GetExtension(file), null);
+                    if (extraFile is not null)
```
markus101, line 142: Nit: newline before `if`

markus101 (changes requested):
There are some pieces of this that don't need to be done or at least should be done in a more streamlined way.

I don't think it makes sense to include the refactoring of  `EpisodeFileMovingService`, there is enough to process in this PR without diving into that since it's not necessary.

The extra file handling probably needs to be rethought some, it feels more like "make it work" than well designed.

src/NzbDrone.Core/MediaFiles/EpisodeFileMovingService.cs
JeWe37: Well the issue is that Sonarr's naming is dynamic and can be pretty involved, so if you still want the ability to use Sonarr's rename, you'd have to replicate that perfectly in the script, which would be rather impractical.

src/NzbDrone.Core/MediaFiles/EpisodeFileMovingService.cs
JeWe37: I have been meaning to ask, what is the semantic difference between an `EpisodeFile` and a `LocalEpisode`? I'm asking to see if information such as say the `DownloadClientItem` even belongs into either of those.

src/NzbDrone.Core/MediaFiles/EpisodeImport/ImportApprovedEpisodes.cs
JeWe37: I suppose if I am already supposed to break up `ScriptImportDecisionInfo` I can also turn this into a parameter to the `TryImport` method instead.

src/NzbDrone.Core/MediaFiles/MediaInfo/UpdateMediaInfoService.cs
JeWe37: I guess I can also just break it into two methods, one which updates the `MediaInfo` and one which does that plus writes to the DB.

JeWe37:
> The extra file handling probably needs to be rethought some, it feels more like "make it work" than well designed.

Well, you seem somewhat skeptical of handling the extra files with a script at all. I don't entirely disagree, it isn't crucial. The only reason I found for potentially wanting them is for rejecting files as I noted previously, but optimally that should really be handled via some sort of custom format system in Sonarr itself, i.e. subtitles only get imported if they match a custom format or something like that. But that would be an entirely separate discussion once again and probably is out of scope here.

Would just removing that functionality entirely work for you?

src/NzbDrone.Core/MediaFiles/EpisodeFileMovingService.cs
JeWe37: Going by what I am seeing so far are the `downloadId`(or I could instead replace the `DownloadClientItemClientInfo` with just the `DownloadClientItem` that contains both it and the `downloadId`), which makes a fair amount of sense, and the `OldFiles` which I am not totally sure how much sense it really does make. I would also definitely have to mutate it part way through to add those.

markus101: `LocalEpisode` is the file at time of import, used to decide whether the file should be imported and hold information about it until it's imported, after which its unused. `EpisodeFile` is what gets stored in the DB and kept going forward.

src/NzbDrone.Core/MediaFiles/MediaInfo/UpdateMediaInfoService.cs
markus101: Just getting the media info for the updated file can be retrieved directly from `_videoFileInfoReader.GetMediaInfo` and then set on the episode file.

markus101:
Removing support for extra/subtitle files and revisiting would be good with me, in order to facilitate it using the same script we could add a "type" like we do for Custom , without having the logic for it present.

src/NzbDrone.Core/MediaFiles/EpisodeFileMovingService.cs
markus101: Ahh gotcha, in that case I wonder if it's more practical to have Sonarr process the file again the next pass (but that could be problematic with multiple files on disk) or just recalculate the filename (after updating mediainfo). Right now forcing a move would be problematic for seeding torrents, even if the file actually hadn't been altered.

JeWe37: I don't think it's an issue actually. The move is from the imported file, that's why I'm setting the Path to null there, then it gets recalculated to the path inside the library.

JeWe37:
Alright with the latest commit the refactoring has been revered and the information has been put into LocalEpisode. The extra support has been removed. It is actually quite a bit cleaner this way.

src/NzbDrone.Core/MediaFiles/ScriptImportDecider.cs
```diff
@@ -0,0 +1,128 @@
...
+                                   IConfigFileProvider configFileProvider,
+                                   Logger logger)
+        {
+            _processProvider = processProvider;
+            _configService = configService;
+            _configFileProvider = configFileProvider;
+            _logger = logger;
+        }
+
+        private void oldFilesInfo(StringDictionary environmentVariables, List<EpisodeFile> oldFiles, Series series)
```
markus101, line 39: No need for this to be a separate function

src/NzbDrone.Core/MediaFiles/ScriptImportDecider.cs
```diff
@@ -0,0 +1,128 @@
...
+            _logger.Debug("Executing external script: {0}", _configService.ScriptImportPath);
+
+            var processOutput = _processProvider.StartAndCapture(_configService.ScriptImportPath, $"\"{sourcePath}\" \"{destinationFilePath}\"", environmentVariables);
+
+            _logger.Debug("Executed external script: {0} - Status: {1}", _configService.ScriptImportPath, processOutput.ExitCode);
+            _logger.Debug("Script Output: \r\n{0}", string.Join("\r\n", processOutput.Lines));
+
+            switch (processOutput.ExitCode)
+            {
+            case 0: // Copy complete
```
markus101, line 117: Indentation for the cases is incorrect

src/NzbDrone.Core/MediaFiles/EpisodeFileMovingService.cs
```diff
@@ -113,10 +126,35 @@ private EpisodeFile TransferFile(EpisodeFile episodeFile, Series series, List<Ep
                 throw new SameFilenameException("File not moved, source and destination are the same", episodeFilePath);
             }
 
-            _diskTransferService.TransferFile(episodeFilePath, destinationFilePath, mode);
+            bool transfer = true;
```
markus101, line 129: `bool` -> `var`

src/NzbDrone.Core/MediaFiles/EpisodeFileMovingService.cs
```diff
@@ -59,6 +67,11 @@ public class EpisodeFileMovingService : IMoveEpisodeFiles
         public EpisodeFile MoveEpisodeFile(EpisodeFile episodeFile, Series series)
         {
             var episodes = _episodeService.GetEpisodesByFileId(episodeFile.Id);
+            return MoveEpisodeFile(episodeFile, series, episodes);
+        }
+
+        public EpisodeFile MoveEpisodeFile(EpisodeFile episodeFile, Series series, List<Episode> episodes)
```
markus101, line 73: Does this need to be public?

src/NzbDrone.Core/MediaFiles/EpisodeFileMovingService.cs
```diff
@@ -113,10 +126,35 @@ private EpisodeFile TransferFile(EpisodeFile episodeFile, Series series, List<Ep
...
+                var scriptImportDecision = _scriptImportDecider.TryImport(episodeFilePath, destinationFilePath, localEpisode, mode);
+
+                switch (scriptImportDecision)
+                {
+                    case ScriptImportDecision.DeferMove:
+                        break;
+                    case ScriptImportDecision.RenameRequested:
+                        _updateMediaInfo.UpdateData(episodeFile, series);
+                        episodeFile.Path = null;
+                        MoveEpisodeFile(episodeFile, series, episodeFile.Episodes);
```
markus101, line 144: Instead of passing all this logic back out of `ScriptImportDecider.Import`, let that handle getting the media info for the new file as well as getting the new file name and basically treats it as a `DeferMove` result.

`_buildFileNames.BuildFilePath(episodes, series, episodeFile, Path.GetExtension(episodeFile.RelativePath));` to calculate the new name and `_videoFileInfoReader.GetMediaInfo(path)` for the media info

markus101 (changes requested):
Looking better, but still some cruft to clean up

src/NzbDrone.Core/MediaFiles/EpisodeFileMovingService.cs
JeWe37: I don't quite follow. Deferring will trigger a call to the disk service to move, copy or hardlink the file. We don't know which of these it'll do. Here we want the move to _always_ be a move. And the paths for how it's moved around also are calculated before this? 

The current process for RenameRequested:
1. Script moves file to the path specified by Sonarr, essentially as though it was a regular move/copy/hardlink. 
2. Return code of the script first triggers Sonarr to recheck the MediaInfo, then the file will be renamed(i.e. moved) _within_ the library. The downloads folder doesn't get touched in this step at all.

A defer move doesn't move within the library, it still moves/copies/hardlinks from the downloads folder. I don't see how these are compatible.

markus101: Ahh, it's already "imported" in this case, but this is still leaking more logic out than necessary, part of that is due to how `EpisodeFileMovingService` is structured (to actually get the new filename and move it), which might change if refactored. In the mean time it can be left as-is, but updating the mediainfo can be done within the script decider and this can just be responsible for taking care of the moving.

JeWe37: Yeah, that's what I tried to do with the refactor before, but alright, I'll fix this and let's hope once this is done at least the actual code will be fine and I can move on to checking everything still tests ok and history cleanup.

JeWe37:
Well since the tests actually seem to be passing with the now simplified code, I decided to also do a preliminary rebase. Probably makes the most sense to just make all this functionality a single commit for now, see [here](https://github.com/JeWe37/Sonarr/commit/876279a086fe8b8e2dbb2ee1529e2482231a5221).

src/NzbDrone.Core/MediaFiles/EpisodeFileMovingService.cs
markus101: It doesn't look like this is called externally and doesn't need to be public.

src/NzbDrone.Core/MediaFiles/ScriptImportDecider.cs
```diff
@@ -0,0 +1,127 @@
...
+            if (oldFiles.Any())
+            {
+                environmentVariables.Add("Sonarr_DeletedRelativePaths", string.Join("|", oldFiles.Select(e => e.RelativePath)));
+                environmentVariables.Add("Sonarr_DeletedPaths", string.Join("|", oldFiles.Select(e => Path.Combine(series.Path, e.RelativePath))));
+                environmentVariables.Add("Sonarr_DeletedDateAdded", string.Join("|", oldFiles.Select(e => e.DateAdded)));
+            }
+
+            _logger.Debug("Executing external script: {0}", _configService.ScriptImportPath);
+
+            var processOutput = _processProvider.StartAndCapture(_configService.ScriptImportPath, $"\"{sourcePath}\" \"{destinationFilePath}\"", environmentVariables);
```
markus101, line 108: Any particular reason to send these as arguments instead of environment variables? It's a bit of a deviation from custom scripts and I think a lot of people will be looking for environment variables for these. Maybe worth setting an environment variable in addition to the arguments?

markus101 (commented):
Thanks for all the fixups so far, have a few more comments/questions, but it's getting closer. @Qstick it'd be great if you could also take a look and give your thoughts on everything from a downstream perspective.

src/NzbDrone.Core/MediaFiles/ScriptImportDecider.cs
JeWe37: Well the idea was to make something with the syntax of say `cp A B` work. But sure I can offer both.

frontend/src/Settings/MediaManagement/MediaManagement.js
```diff
@@ -200,6 +200,40 @@ class MediaManagement extends Component {
...
+                        advancedSettings={advancedSettings}
+                        isAdvanced={true}
+                        size={sizes.MEDIUM}
+                      >
+                        <FormLabel>Import Using Script</FormLabel>
+
+                        <FormInputGroup
+                          type={inputTypes.CHECK}
+                          name="useScriptImport"
+                          helpText="Copy files for importing using a script, say for transcoding"
```
Qstick, line 213: Should we be a bit more formal here `Copy files for importing using a script (ex. for transcoding)`

frontend/src/Settings/MediaManagement/MediaManagement.js
```diff
@@ -200,6 +200,40 @@ class MediaManagement extends Component {
...
+                          <FormGroup
+                            advancedSettings={advancedSettings}
+                            isAdvanced={true}
+                          >
+                            <FormLabel>Import Script Path</FormLabel>
+
+                            <FormInputGroup
+                              type={inputTypes.TEXT}
+                              name="scriptImportPath"
+                              helpText="The path to the script to use for importing"
```
Qstick, line 230: Can we use file browser here instead of just text input?

JeWe37:
Alright, made all the required changes - I hope this will be everything. Do you have a policy for the commit history I should follow?

markus101:
Nothing formal, but in this case we'll squash it all down when we merge it.

markus101 (approved):

PR #5546, opened 2023-04-11
New: Add version and timestamp to backup archive

mynameisbogdan:
(cherry picked from commit 43fa4cb5b9dca1aed25acb169c63ddd9a8e2712b)

#### Database Migration
NO

#### Description
Currently `CreateVersionInfo` doesn't seem to do anything, so I'm thinking of adding version and timestamp to the zip file.

### DIFF
--- a/src/NzbDrone.Core/Backup/BackupService.cs
+++ b/src/NzbDrone.Core/Backup/BackupService.cs
@@ -69,7 +69,8 @@ namespace NzbDrone.Core.Backup
             _diskProvider.EnsureFolder(_backupTempFolder);
             _diskProvider.EnsureFolder(GetBackupFolder(backupType));
 
-            var backupFilename = string.Format("sonarr_backup_v{0}_{1:yyyy.MM.dd_HH.mm.ss}.zip", BuildInfo.Version, DateTime.Now);
+            var dateNow = DateTime.Now;
+            var backupFilename = $"sonarr_backup_v{BuildInfo.Version}_{dateNow:yyyy.MM.dd_HH.mm.ss}.zip";
             var backupPath = Path.Combine(GetBackupFolder(backupType), backupFilename);
 
             Cleanup();
@@ -81,7 +82,7 @@ namespace NzbDrone.Core.Backup
 
             BackupConfigFile();
             BackupDatabase();
-            CreateVersionInfo();
+            CreateVersionInfo(dateNow);
 
             _logger.ProgressDebug("Creating backup zip");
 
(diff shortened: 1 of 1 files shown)


src/NzbDrone.Core/Backup/BackupService.cs
```diff
@@ -205,9 +205,10 @@ private void BackupConfigFile()
 
         private void CreateVersionInfo()
         {
-            var builder = new StringBuilder();
+            var tempFile = Path.Combine(_backupTempFolder, "VERSION");
```
markus101, line 208: If we're going to include non-version info then I don't think naming it `VERSION` makes sense.

src/NzbDrone.Core/Backup/BackupService.cs
```diff
@@ -205,9 +205,10 @@ private void BackupConfigFile()
...
+            var contents = $"v{BuildInfo.Version}-{DateTime.Now:yyyy-MM-dd HH:mm:ss}";
```
markus101, line 209: Is including a different time stamp than the file name necessary?

src/NzbDrone.Core/Backup/BackupService.cs
markus101: Maybe `INFO` would be better?

src/NzbDrone.Core/Backup/BackupService.cs
markus101: I guess my concerns were two fold.

1. The time stamp here differs from the file name of the backup
2. Was it necessary at all.

Sounds like 2 is a yes, but a consistent time stamp should be used for both the filename and the info in the file.

thezoggy: shouldnt the start time of the backup be used for the filename+referred to the file instead of it being generated on each event which then means different times can be seen..

src/NzbDrone.Core/Backup/BackupService.cs
```diff
@@ -203,11 +205,12 @@ private void BackupConfigFile()
             _diskTransferService.TransferFile(configFile, tempConfigFile, TransferMode.Copy);
         }
 
-        private void CreateVersionInfo()
+        private void CreateVersionInfo(DateTime dateNow)
         {
-            var builder = new StringBuilder();
+            var tempFile = Path.Combine(_backupTempFolder, "INFO");
+            var contents = $"v{BuildInfo.Version}\n{dateNow:yyyy-MM-dd HH:mm:ss}";
```
markus101, line 211: Use a string builder to build the contents instead of building a string with new lines manually.

src/NzbDrone.Core/Backup/BackupService.cs
```diff
@@ -69,7 +69,9 @@ public void Backup(BackupType backupType)
             _diskProvider.EnsureFolder(_backupTempFolder);
             _diskProvider.EnsureFolder(GetBackupFolder(backupType));
 
-            var backupFilename = string.Format("sonarr_backup_v{0}_{1:yyyy.MM.dd_HH.mm.ss}.zip", BuildInfo.Version, DateTime.Now);
+            var dateNow = DateTime.Now;
+
```
markus101, line 73: Nit: unnecessary new line

markus101 (changes requested):

markus101 (changes requested):

markus101:
Thanks for the fixes!

PR #5557, opened 2023-04-15
Rename to FolderChmodValidator to match class name

mynameisbogdan:
#### Database Migration
NO


markus101:
FYI, small changes like this don't need a public release note, which is what `Fixed:` or `New:` will result in.

PR #5561, opened 2023-04-16
Fixed: Return properly formatted episode file for deleted files Webhook

mpeleshenko:
#### Database Migration
NO

#### Description
Return WebhookEpisodeFile for onEpisodeFileDelete events to match the type used in onDownload events.


markus101:
Thanks!

PR #5566, opened 2023-04-18
Fixed: Improved Blocklist Comparison logic

bakerboy448:
#### Database Migration
NO

#### Description
- New: Always compare Tracker for Torrent Blocklisting
- Fixed: Improved Blocklist Comparison logic

- ToDo: Add Blocklist tests
How / One mocked up would be a good start.

#### Todos
- [ ] Tests
- Wiki Updates

* https://github.com/Radarr/Radarr/issues/8086

### DIFF
--- a/src/NzbDrone.Core.Test/Blocklisting/BlocklistServiceFixture.cs
+++ b/src/NzbDrone.Core.Test/Blocklisting/BlocklistServiceFixture.cs
@@ -12,47 +12,145 @@ namespace NzbDrone.Core.Test.Blocklisting
     [TestFixture]
     public class BlocklistServiceFixture : CoreTest<BlocklistService>
     {
-        private DownloadFailedEvent _event;
+        private DownloadFailedEvent _nzbEvent;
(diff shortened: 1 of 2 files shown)


markus101 (commented):
Blocklist intentionally ignores the indexer, this allows the file to be blocked once and not just grabbed again from another indexer.

Sounds like the issue is when blocking the wrong item from history is used, so instead of fixing the comparison when grabbing the correct item needs to be blocked.

Is this a fix or a feature you've included change logs for both.

bakerboy448:
poked around the logic and can't see where it'd be sending the wrong item in history.

Any guidance for writing 1 of the tests for this to make tracking that down easier?

markus101:
I did more digging and pushed up a fix in cf48bf304122bacc597de1d4d4429065d2358fa8 the issue is the item that was marked as failed will only be used in history if it's the first returned history item. It will now use that item first and use any others to mark other episodes in the same release as failed.

PR #5569, opened 2023-04-20
Add Apprise notifications

mynameisbogdan:
#### Database Migration
NO

#### Description
[Apprise](https://github.com/caronc/apprise) is a self-hosted solution that supports a lot of [notifications providers](https://github.com/caronc/apprise#supported-notifications). It can be used as a [docker image from LS](https://github.com/linuxserver/docker-apprise-api#usage).

### DIFF
--- a/frontend/src/Components/Form/ProviderFieldFormGroup.js
+++ b/frontend/src/Components/Form/ProviderFieldFormGroup.js
@@ -65,6 +65,7 @@ function ProviderFieldFormGroup(props) {
     label,
     helpText,
     helpLink,
+    placeholder,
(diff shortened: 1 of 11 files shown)


markus101 (commented):
Untested, but looks fine, just some naming cleanup suggestions.

markus101:
Can you remove the `New:` from the placeholders commit so it won't show up in the change log please?

markus101:
The Apprise commit was correct, the other one needs `New:` dropped instead, we don't need a change log message for the placeholders, but do for the actual addition of Apprise.

PR #5575, opened 2023-04-22
Ensure CustomFormat is taken from disk filename

merqwury:
#### Database Migration
NO

#### Description
As it stands, sonarr only relies on the scene filename, which may be incorrect (i.e. mismatch between release title and release filename). This can be worked around, by having a custom renaming scheme. However, sonarr doesn't look at the filename on disk as a priority, and thus the CF score is always based on the original release filename. This PR aims to always use what's on disk to determine the imported CF score.

#### Todos
- [ ] Tests
- [ ] Wiki Updates

*

### DIFF
--- a/src/NzbDrone.Core/CustomFormats/CustomFormatCalculationService.cs
+++ b/src/NzbDrone.Core/CustomFormats/CustomFormatCalculationService.cs
@@ -107,7 +107,7 @@ namespace NzbDrone.Core.CustomFormats
             var episodeInfo = new ParsedEpisodeInfo
             {
                 SeriesTitle = localEpisode.Series.Title,
-                ReleaseTitle = localEpisode.SceneName.IsNotNullOrWhiteSpace() ? localEpisode.SceneName : Path.GetFileName(localEpisode.Path),
+                ReleaseTitle = Path.GetFileName(localEpisode.Path),
                 Quality = localEpisode.Quality,
                 Languages = localEpisode.Languages,
                 ReleaseGroup = localEpisode.ReleaseGroup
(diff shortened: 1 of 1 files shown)


bakerboy448 (commented):
Logic starting on line 160 is what needs updating, no?

bakerboy448:
Current logic understanding:
- if scene name exists use scene name for custom formats
- If original file path exists use that for custom formats
- then if neither exists use current on disk name


In effect - no matter what you rename files to - it does not matter for CF purposes as the original filename will always be used.

if you manage episode and reimport the renamed file then the post-named file and any probed media info included in naming / original title if used for naming will be used for CF calc

markus101:
My concern with this change is now you must include custom format with a score greater than 0 in your file name, including media info if that's relevant, regardless of whether you want them included or not.

Instead of that we should use all the available information for a file, including the source file name and current file name, then dedupe the results so things aren't counted multiple times.

bakerboy448:
> including the source file name and current file name, then dedupe the results so things aren't counted multiple times

how would it handle where the scenename / on disk name have mutually exclusive matches?

markus101:
In that case they'd both be counted.

The biggest issue with everything is there are up to 4 different things involved, you have the name the indexer says it is, which may or may not match the release name from the release group, then you have the file name, which might match the release name or might be something else made up by the release group or not match because of season packs and finally you have what the user decided to call the file, which can be some concoction of all the previous things + media info.

markus101:
I don't think it's significantly larger in scope, but probably worth a GHI to discus the pros and cons, opened https://github.com/Sonarr/Sonarr/issues/5598

PR #5576, opened 2023-04-24
Build download requests from indexer implementation

mynameisbogdan:
#### Database Migration
NO

#### Description
This allows building the HTTP request from the indexer implementation, eg. when custom headers like authorization are needed.

Example for RED in Lidarr, since this is how I discovered the inability for this:
```csharp
        public override HttpRequest GetDownloadRequest(Uri link)
        {
            var request = new HttpRequestBuilder(link.AbsoluteUri)
                .SetHeader("Authorization", Settings.ApiKey)
                .Build();

            return request;
        }
```

Also fixing the redirect without host issue in `DownloadFromWebUrl`.

### DIFF
--- a/src/NzbDrone.Core/Download/TorrentClientBase.cs
+++ b/src/NzbDrone.Core/Download/TorrentClientBase.cs
@@ -41,7 +41,7 @@ namespace NzbDrone.Core.Download
         protected abstract string AddFromMagnetLink(RemoteEpisode remoteEpisode, string hash, string magnetLink);
         protected abstract string AddFromTorrentFile(RemoteEpisode remoteEpisode, string hash, string filename, byte[] fileContent);
 
-        public override string Download(RemoteEpisode remoteEpisode)
+        public override string Download(RemoteEpisode remoteEpisode, IIndexer indexer)
         {
             var torrentInfo = remoteEpisode.Release as TorrentInfo;
 
@@ -68,7 +68,7 @@ namespace NzbDrone.Core.Download
                 {
                     try
                     {
-                        return DownloadFromWebUrl(remoteEpisode, torrentUrl);
+                        return DownloadFromWebUrl(remoteEpisode, indexer, torrentUrl);
                     }
                     catch (Exception ex)
                     {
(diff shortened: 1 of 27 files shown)


src/NzbDrone.Core/Indexers/HttpIndexerBase.cs
```diff
@@ -109,6 +108,11 @@ public override IList<ReleaseInfo> Fetch(SpecialEpisodeSearchCriteria searchCrit
             return FetchReleases(g => g.GetSearchRequests(searchCriteria));
         }
 
+        public override HttpRequest GetDownloadRequest(Uri link)
```
markus101, line 111: Instead of a URI can this accept a string? Looks like the two callers of this accept a string and then have to convert to a URI, when that could be done here instead.

src/NzbDrone.Core/Download/TorrentClientBase.cs
```diff
@@ -149,7 +149,9 @@ private string DownloadFromWebUrl(RemoteEpisode remoteEpisode, string torrentUrl
                             return DownloadFromMagnetUrl(remoteEpisode, locationHeader);
                         }
 
-                        return DownloadFromWebUrl(remoteEpisode, locationHeader);
+                        request.Url += new HttpUri(locationHeader);
```
markus101, line 152: Is this going to work correctly with both relative and absolute location headers?

src/NzbDrone.Core/Indexers/IIndexer.cs
```diff
@@ -1,4 +1,6 @@
+using System;
```
markus101, line 1: This name space doesn't appear to be used.

src/NzbDrone.Core/Indexers/HttpIndexerBase.cs
markus101: It doesn't look that applies here, plus as-is it's a string converted to a URI then back to a string when `link.AbsoluteUri` is used in `HttpIndexerBase`. If it is invalid then that should be dealt with by `HttpUri`

markus101 (changes requested):
Also looks like a bunch of tests are unhappy at the moment.

Qstick:
He closed all his PRs it seems

PR #5581, opened 2023-04-27
New: Updated button and calendar outline colors for dark theme

Stanzilla:
#### Database Migration
NO

#### Description
This upgrades the border color for buttons and the calendar grid from very bright to a pleasant darker color.

##### Before

image

##### After

image

I assume that these were never supposed to be that bright considering dark mode is still work in progress(?).

There is a consideration to use the same color as the calendar headers for the calendar borders (#2a2a2a) so the connections look a bit more natural but that is also very close to the background color, contrast wise, and almost makes them disappear. Please let me know if you'd prefer that.

### DIFF
--- a/frontend/src/Styles/Themes/dark.js
+++ b/frontend/src/Styles/Themes/dark.js
@@ -94,9 +94,9 @@ module.exports = {
 
   defaultButtonTextColor: '#eee',
   defaultBackgroundColor: '#333',
-  defaultBorderColor: '#eaeaea',
+  defaultBorderColor: '#393f45',
   defaultHoverBackgroundColor: '#444',
-  defaultHoverBorderColor: '#d6d6d6',
+  defaultHoverBorderColor: '#5a6265',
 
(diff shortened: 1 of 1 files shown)


markus101:
This needs to be rebased.

Stanzilla:
> This needs to be rebased.

Done!

frontend/src/Styles/Themes/dark.js
```diff
@@ -209,7 +209,7 @@ module.exports = {
 
   calendarTodayBackgroundColor: '#3e3e3e',
   calendarBackgroundColor: '#2a2a2a',
-  calendarBorderColor: '#cecece',
+  calendarBorderColor: ''#393f45',
```
markus101, line 212: Need to remove the extra quote here

Stanzilla: Oh wow, nice rebase fail here, thank you!

PR #5587, opened 2023-04-27
Fixed some aria violations

Stanzilla:
#### Database Migration
NO

#### Description

Fixes a few aria violations that I found when running lighthouse, the remaining one is in 3rd party code at https://github.com/moroshko/react-autosuggest/pull/830

### DIFF
--- a/frontend/src/Components/DescriptionList/DescriptionListItem.js
+++ b/frontend/src/Components/DescriptionList/DescriptionListItem.js
@@ -18,7 +18,7 @@ class DescriptionListItem extends Component {
     } = this.props;
 
     return (
-      <span className={className}>
+      <div className={className}>
         <DescriptionListItemTitle
           className={titleClassName}
         >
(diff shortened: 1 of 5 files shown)


frontend/src/Components/Link/IconButton.js
```diff
@@ -23,6 +23,7 @@ function IconButton(props) {
         className,
         isDisabled && styles.isDisabled
       )}
+      name={name}
```
markus101, line 26: `name` is an object, I don't think we want to pass that as the name attribute on the button

frontend/src/Components/DescriptionList/DescriptionListItem.js
```diff
@@ -18,7 +18,7 @@ class DescriptionListItem extends Component {
     } = this.props;
 
     return (
-      <span className={className}>
+      <div className={className}>
```
markus101, line 21: aria related or something else?

Stanzilla: HTML spec actually, span are not valid inside of there

frontend/src/Components/Link/IconButton.js
Stanzilla: Fixed

PR #5590, opened 2023-04-28
Cleanup used frontend tools

Stanzilla:
#### Database Migration
NO

#### Description
To make it easier for new contributors (including me), I took a look at which tools that have config files for vs. are actually used.

Please correct me if I'm wrong but it looks like we are using the following actively:

- Eslint
- Prettier
- ESFormatter (run manually?)
- Stylelint

I also moved the .vscode folder to the `frontend` folder since that made more sense to me, if someone wants to work on the frontend only, it's more likely they open up this folder in VSC instead of `src`. Also added a extension suggestion file for Eslint, Prettier and Stylelint since those are the ones that have good support in VSC and are actively used in the project.

It would probably make sense to then mention these tools in the [Contribution guide](https://github.com/Sonarr/Sonarr/blob/develop/CONTRIBUTING.md) as well?

#### TODO

- [x] Is EsFormatter actually used/run regularly? 
- [ ] Is tern still actually used/run regularly? 
- [ ] Update documentation
- [ ] More testing
- [ ] 73369:220 Warning: Failed prop type: The prop `lazy` is marked as required in `SeriesPoster`, but its value is `undefined`.


markus101:
> ESFormatter (run manually?)

I don't know that we're actually using this at all, maybe previously, but prettier will handle formatting in TS files, with the goal to eventually convert everything, but that's not a short term goal.

Stanzilla:
Is there a way to see the TeamCity build log without an account?

Qstick:
Click "Login as Guest"

Stanzilla:
> Click "Login as Guest"

Thanks, it's `Sonarr\frontend\src\Components\Form\TagInputInput.js
  41:9  error  Unknown property 'component' found  react/no-unknown-property`

I'll take a look tomorrow.

Qstick:
Yea, likely from the lint package bumps triggering a new rule

Stanzilla:
> Yea, likely from the lint package bumps triggering a new rule

Looks more like something was fixed so this is now detected? It's a custom property and I think in those you always have to use the `data-*` prefix. I added it to the config to be ignored now.

Qstick:
> Looks more like something was fixed so this is now detected? It's a custom property and I think in those you always have to use the `data-*` prefix? I added it to the config to be ignored now.

TagInputInput.js hasn't been touched in 8 months. It's most certainly something that's been changed in eslint which is now picking up the issue in TagInputInput. This happens quite frequently. Instead of ignoring the rule, let's either fix it or leave the version where it is until we fix it. Ignoring rules almost always means we'll never go back and remember to fix it. 

```diff
diff --git a/frontend/src/Components/Form/TagInputInput.js b/frontend/src/Components/Form/TagInputInput.js
index d88c0ce8a..86628b134 100644
--- a/frontend/src/Components/Form/TagInputInput.js
+++ b/frontend/src/Components/Form/TagInputInput.js
@@ -38,7 +38,6 @@ class TagInputInput extends Component {
       <div
         ref={forwardedRef}
         className={className}
-        component="div"
```

Stanzilla:
> > Looks more like something was fixed so this is now detected? It's a custom property and I think in those you always have to use the `data-*` prefix? I added it to the config to be ignored now.
> 
> TagInputInput.js hasn't been touched in 8 months. It's most certainly something that's been changed in eslint which is now picking up the issue in TagInputInput. This happens quite frequently. Instead of ignoring the rule, let's either fix it or leave the version where it is until we fix it. Ignoring rules almost always means we'll never go back and remember to fix it.
> 
> ```diff
> diff --git a/frontend/src/Components/Form/TagInputInput.js b/frontend/src/Components/Form/TagInputInput.js
> index d88c0ce8a..86628b134 100644
> --- a/frontend/src/Components/Form/TagInputInput.js
> +++ b/frontend/src/Components/Form/TagInputInput.js
> @@ -38,7 +38,6 @@ class TagInputInput extends Component {
>        <div
>          ref={forwardedRef}
>          className={className}
> -        component="div"
> ```

The same code is used in several more files though

Qstick:
It's valid in those files, thus why it doesn't throw a lint error for them (only the TagInputInput.js instance). There is no `component` property for a `div` block, but there is for other components like `Link` which are the 9 other uses of this.

Stanzilla:
> It's valid in those files, thus why it doesn't throw a lint error for them (only the TagInputInput.js instance). There is no `component` property for a `div` block, but there is for other components like `Link` which are the 9 other uses of this.
> 
> Eslint is doing what is should here and calling us out for using a property incorrectly in a `<div>`.

Got it, thanks. I should have left it at figuring it out the next day but I'm allergic to broken CI :D

package.json
```diff
@@ -86,71 +85,71 @@
...
     "ts-loader": "9.4.2",
-    "typescript-plugin-css-modules": "4.1.1",
+    "typescript-plugin-css-modules": "5.0.1",
     "url-loader": "4.1.1",
     "webpack": "5.81.0",
     "webpack-cli": "5.0.2",
     "webpack-livereload-plugin": "3.0.2",
     "worker-loader": "3.0.8"
   },
+  "resolutions": {
```
markus101, line 150: Why is this now necessary? Also seems like yet another thing that needs to be updated.

frontend/.stylelintrc
```diff
@@ -15,96 +15,46 @@
...
       {
         "ignoreAtRules": [
           "/^add\\-mixin$/",
           "/^define\\-mixin$/"
-        ] 
+        ]
       }
     ],
     "at-rule-no-vendor-prefix": true,
-    "at-rule-semicolon-newline-after": "always",
```
markus101, line 31: Why are all these rules being removed?

markus101 (commented):
Haven't tested yet, but the major version bumps to packages is concerning given some previous issues.

frontend/src/Components/Form/TagInputInput.css
```diff
@@ -1,12 +1,9 @@
...
-  right: -1px;
-  bottom: -1px;
-  left: -1px;
   display: flex;
   align-items: start;
   flex-wrap: wrap;
   padding: 1px 16px;
   min-height: 33px;
   cursor: default;
+  inset: -1px;
```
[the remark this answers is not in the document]
Stanzilla: Yup!

package.json
Stanzilla: It's temporary, can be removed as soon as stylelint is done with their ESM transition.

frontend/.stylelintrc
Stanzilla: they are deprecated, Stylelint decided to go back to being just about correctness rules and leave stylistic choices up to tools like Prettier, see https://github.com/stylelint/stylelint/blob/main/docs/migration-guide/to-15.md

frontend/src/Components/Form/TagInputInput.css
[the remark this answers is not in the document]
markus101: `inset` will need to be added to `stylelintrc`

markus101:
Makes sense to land some of this as separate commits, but right now they are too granular.

Removing the unused configs/tools can be done in a single commit.
The commits fixing the various issues found after upgrading should be part of the upgrade commit so the commits are all functional on their own.

markus101:
Thanks!

Stanzilla:
Oh I had a few open todos but I guess that means you want me to address them in a follow-up :P

markus101:
Doh, I guess I missed them on mobile, a follow up is fine though.

PR #5591, opened 2023-04-29
Log contents for invalid torrent files

mynameisbogdan:
#### Database 
NO

#### Description
Useful for debugging download issues.

### DIFF
--- a/src/NzbDrone.Core/MediaFiles/TorrentInfo/TorrentFileInfoReader.cs
+++ b/src/NzbDrone.Core/MediaFiles/TorrentInfo/TorrentFileInfoReader.cs
@@ -1,4 +1,6 @@
-﻿using MonoTorrent;
+﻿using System.Text;
+using MonoTorrent;
+using NLog;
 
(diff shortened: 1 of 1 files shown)


src/NzbDrone.Core/MediaFiles/TorrentInfo/TorrentFileInfoReader.cs
```diff
@@ -9,9 +11,24 @@ public interface ITorrentFileInfoReader
...
         public string GetHashFromTorrentFile(byte[] fileContents)
         {
-            return Torrent.Load(fileContents).InfoHash.ToHex();
+            try
+            {
+                return Torrent.Load(fileContents).InfoHash.ToHex();
+            }
+            catch
+            {
+                _logger.Trace("Invalid torrent file contents: {0}", Encoding.ASCII.GetString(fileContents));
```
markus101, line 29: I assume the value here is we'll be able to log something like HTML content when we expected a torrent?

Similar to logging (IIRC) indexer responses, should we limit this to a maximum size instead of spewing out everything?

PR #5600, opened 2023-04-30
Fix parsing empty string as boolean

mynameisbogdan:
#### Database Migration
NO

#### Description
```
System.FormatException: String '' was not recognized as a valid Boolean.
   at System.Boolean.Parse(ReadOnlySpan`1 value)
   at System.Boolean.Parse(String value)
```

### DIFF
--- a/src/Sonarr.Http/Extensions/RequestExtensions.cs
+++ b/src/Sonarr.Http/Extensions/RequestExtensions.cs
@@ -55,9 +55,9 @@ namespace Sonarr.Http.Extensions
         {
             var parameterValue = request.Query[parameter];
 
-            if (parameterValue.Any())
+            if (parameterValue.Any() && bool.TryParse(parameterValue.ToString(), out var value))
             {
-                return bool.Parse(parameterValue.ToString());
+                return value;
             }
 
             return defaultValue;


markus101:
An empty string isn’t a valid boolean value, so it seems reasonable to throw an error if that is encountered because the request is invalid.

Instead of the 500 error that I assume is being thrown this should throw a 400 error instead.

markus101:
Is this coming from the UI or something else?

If the value is indeterminate then sending nothing would fall through to the default value, as opposed to sending anything and having that fall through to the default value instead of an error.

This feels more of a API client issue (either send a valid Boolean value or nothing) as opposed to an API issue beyond properly reporting that the client made a bad request.

Qstick:
This was caused by a bad UI variable in Radarr, which just got fixed in https://github.com/Radarr/Radarr/commit/bf3e23cc3bef351315e3239d7e4b7e143c059a6d (It's different in Radarr to Sonarr)

While it does make some sense to make the API more robust on this front, this was not the root cause of the issue that we saw pop up today. 

Really in this csae we should use native ASPNet style query parameters for DeleteSeries instead of calling to `GetBooleanQueryParameter`. This would also fix the issue of these not showing up in API Docs

Qstick:
Below is Radarr, but something like this 

```diff
diff --git a/src/Radarr.Api.V3/Movies/MovieController.cs b/src/Radarr.Api.V3/Movies/MovieController.cs
index 538311826e..aba70c22b9 100644
--- a/src/Radarr.Api.V3/Movies/MovieController.cs
+++ b/src/Radarr.Api.V3/Movies/MovieController.cs
@@ -245,12 +245,9 @@ public ActionResult<MovieResource> UpdateMovie(MovieResource moviesResource)
         }

         [RestDeleteById]
-        public void DeleteMovie(int id)
+        public void DeleteMovie(int id, bool? addImportExclusion, bool? deleteFiles)
         {
-            var addExclusion = Request.GetBooleanQueryParameter("addImportExclusion");
-            var deleteFiles = Request.GetBooleanQueryParameter("deleteFiles");
-
-            _moviesService.DeleteMovie(id, deleteFiles, addExclusion);
+            _moviesService.DeleteMovie(id, deleteFiles.GetValueOrDefault(), addImportExclusion.GetValueOrDefault());
         }

         private void MapCoversToLocal(MovieResource movie)
```

PR #5604, opened 2023-04-30
Check response status code in classes that extend RssParser

mynameisbogdan:
#### Database Migration
NO

### DIFF
--- a/src/NzbDrone.Core/Indexers/EzrssTorrentRssParser.cs
+++ b/src/NzbDrone.Core/Indexers/EzrssTorrentRssParser.cs
@@ -1,6 +1,4 @@
 using System.Linq;
-using System.Xml.Linq;
-using NzbDrone.Common.Extensions;
 using NzbDrone.Core.Indexers.Exceptions;
 
 namespace NzbDrone.Core.Indexers
@@ -20,6 +18,8 @@ namespace NzbDrone.Core.Indexers
 
         protected override bool PreProcess(IndexerResponse indexerResponse)
         {
+            base.PreProcess(indexerResponse);
+
             var document = LoadXmlDocument(indexerResponse);
             var items = GetItems(document).ToList();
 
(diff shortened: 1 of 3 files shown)


Qstick:
And info on what we are attempting to solve here? Torznab spec indicates we should always get a 200 back, and the exception, 429s are handled in HttpClient.

Qstick:
Ok, so issues in upstream app/site that cause 500 errors to be returned. We should probably add to NewsnabRssParser as well in that case, as a 500 from Prowlarr or Hydra could also be seen there.

Qstick:
> I'm thinking if maybe `indexerResponse.HttpResponse.HasHttpError` would be a better condition in case someone likes to use >= 200 <300 codes.

Yea, agree

src/NzbDrone.Core/Indexers/EzrssTorrentRssParser.cs
```diff
@@ -20,6 +18,11 @@ public EzrssTorrentRssParser()
 
         protected override bool PreProcess(IndexerResponse indexerResponse)
         {
+            if (indexerResponse.HttpResponse.HasHttpError)
```
markus101, line 21: I think we always want to call this, same as we did before to catch HTML content responses.

markus101 (changes requested):
> Any info on what we are attempting to solve here?

This is why the template asks for this information, please don't nuke it.

src/NzbDrone.Core/Indexers/EzrssTorrentRssParser.cs
markus101: I understand, but EZRSS parser always checked this, unless there is a report of an issue then it should continue to do so.

markus101: You don't want to revert it, just don't require an HTTP error to check it here.

markus101: Only for EZRSS, it was the only one previously using `base.PreProcess`

markus101 (approved):

PR #5607, opened 2023-05-01
Fix downloading releases without an indexer

### LINKED ISSUE #5606: Grabbing pushed release fails due to 0 indexer id

### Current Behavior

Pushed releases cannot be grabbed

### Expected Behavior

Pushed releases should be grabbed 

### Steps To Reproduce

1. Have pending release via push api
2. Release grab attempt
3. Fails due to no indexer id

### Environment

```markdown
- OS: agnostic
- Sonarr: 4.0.0.470
- Docker Install: agnostic
- Using Reverse Proxy: agnostic
- Browser: agnostic
```

Develop

### Trace Logs?

https://gist.github.com/bakerboy448/fe0a2e91bf3a195ed6efea06df975ccb

### Anything else?

Regression  #5576  https://github.com/Sonarr/Sonarr/commit/a0b08f6c6f106d92cdb12fbb959dd2605c22fe6a

mynameisbogdan:
#### Database Migration
NO

#### Description
Changed the download request not to assume the indexer is always defined.

* Fixes #5606

### DIFF
--- a/src/NzbDrone.Core/Download/DownloadService.cs
+++ b/src/NzbDrone.Core/Download/DownloadService.cs
@@ -71,13 +71,19 @@ namespace NzbDrone.Core.Download
 
             // Get the seed configuration for this release.
             remoteEpisode.SeedConfiguration = _seedConfigProvider.GetSeedConfiguration(remoteEpisode);
-            var indexer = _indexerFactory.GetInstance(_indexerFactory.Get(remoteEpisode.Release.IndexerId));
 
             // Limit grabs to 2 per second.
             if (remoteEpisode.Release.DownloadUrl.IsNotNullOrWhiteSpace() && !remoteEpisode.Release.DownloadUrl.StartsWith("magnet:"))
             {
                 var url = new HttpUri(remoteEpisode.Release.DownloadUrl);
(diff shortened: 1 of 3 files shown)


src/NzbDrone.Core/Download/DownloadService.cs
```diff
@@ -71,7 +72,15 @@ private void DownloadReport(RemoteEpisode remoteEpisode, IDownloadClient downloa
 
             // Get the seed configuration for this release.
             remoteEpisode.SeedConfiguration = _seedConfigProvider.GetSeedConfiguration(remoteEpisode);
-            var indexer = _indexerFactory.GetInstance(_indexerFactory.Get(remoteEpisode.Release.IndexerId));
+
+            IIndexer indexer = null;
```
markus101, line 76: If the `IndexerId` is zero, don't look it up, if the indexer is any other value it should be looked up and fail if the ID is invalid. Right now an invalid ID greater than 0 will not fail.

Qstick:
This looks solid now. An invalid Id greater than 0 will get changed to 0 in the ReleasePushController, so there is also a failsafe there. Good here.

Qstick (approved):

PR #5612, opened 2023-05-02
Fix validation for boolean query parameters

mynameisbogdan:
#### Database Migration
NO

#### Description
Going from https://github.com/Sonarr/Sonarr/pull/5600#issuecomment-1529155922, switched some params to native ASP.NET style to fix validation and to show up in swagger. 

I didn't use nullable bool since it skips the validation.

Before:
before

After:
after

### DIFF
--- a/src/Sonarr.Api.V3/Series/SeriesController.cs
+++ b/src/Sonarr.Api.V3/Series/SeriesController.cs
@@ -138,9 +138,8 @@ namespace Sonarr.Api.V3.Series
 
         [RestPutById]
         [Consumes("application/json")]
-        public ActionResult<SeriesResource> UpdateSeries(SeriesResource seriesResource)
+        public ActionResult<SeriesResource> UpdateSeries(SeriesResource seriesResource, bool moveFiles = false)
         {
-            var moveFiles = Request.GetBooleanQueryParameter("moveFiles");
             var series = _seriesService.GetSeries(seriesResource.Id);
 
             if (moveFiles)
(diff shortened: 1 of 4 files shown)


src/Sonarr.Api.V3/Series/SeriesController.cs
```diff
@@ -167,11 +166,8 @@ public ActionResult<SeriesResource> UpdateSeries(SeriesResource seriesResource)
         }
 
         [RestDeleteById]
-        public void DeleteSeries(int id)
+        public void DeleteSeries(int id, bool deleteFiles = false, bool addImportListExclusion = false)
```
markus101, line 169: Does this need `[FromQuery]` to ensure it comes from the query string?

src/Sonarr.Api.V3/Series/SeriesController.cs
```diff
@@ -138,9 +138,8 @@ public ActionResult<SeriesResource> AddSeries(SeriesResource seriesResource)
 
         [RestPutById]
         [Consumes("application/json")]
-        public ActionResult<SeriesResource> UpdateSeries(SeriesResource seriesResource)
+        public ActionResult<SeriesResource> UpdateSeries(SeriesResource seriesResource, bool moveFiles = false)
```
markus101, line 141: Does this need `[FromQuery]` to ensure it comes from the query string?

markus101 (commented):
What else needs to be done in order to convert this from a draft PR?

src/Sonarr.Api.V3/Series/SeriesController.cs
markus101: Yeah, it definitely works without, seems like it can be used for more complex model binding https://learn.microsoft.com/en-us/aspnet/core/mvc/models/model-binding?view=aspnetcore-7.0

Qstick:
Should we include the new openapi.json as well with the PR (it's not auto PRd separate in Sonarr)

markus101:
Thanks for the fixes and the new action, I've added the token secret.

PR #5613, opened 2023-05-02
New: On Health Restored notification

### LINKED ISSUE #4262: Notifications "on health restored"
**Describe the problem**
There is an option to have notifications sent "On Health Issue", and also "Include Health Warnings"
I sometimes receive a notification telling me an indexer is unavailable.
I connect to Sonarr, click the "test all" button, and the error goes away. It was a very short-live problem, like some kind on network glitch or something.

It would great to avoid those kind of alarms
I see other tickets, like #3813, basically asking for a failure to have to happen twice in a row before raising an alarm.
That idea, that I find very good, has been dismissed it seems, unfortunately

As an alternative then, would it be possible to have the option to have notifications sent "on health restored", or something like that?
In my case I'd receive the failure notification, and some minutes later would receive the "back to normal" message, telling me I don't have to bother connecting to Sonarr to check what's going on.

Thanks

CheAle14:
#### Database Migration
YES (190)

#### Description
Adds a notification when a previously failing check is now passing. Includes the message of the previously failing check to aid identifying what's now fixed, in case there are multiple health issues. 

* Closes #4262

### DIFF
--- a/src/NzbDrone.Core/HealthCheck/HealthCheckRestoredEvent.cs
+++ b/src/NzbDrone.Core/HealthCheck/HealthCheckRestoredEvent.cs
@@ -1,0 +1,16 @@
+using NzbDrone.Common.Messaging;
+
+namespace NzbDrone.Core.HealthCheck
+{
+    public class HealthCheckRestoredEvent : IEvent
+    {
+        public HealthCheck PreviousCheck { get; private set; }
+        public bool IsInStartupGracePeriod { get; private set; }
+
+        public HealthCheckRestoredEvent(HealthCheck previousCheck, bool isInStartupGracePeriod)
+        {
+            PreviousCheck = previousCheck;
+            IsInStartupGracePeriod = isInStartupGracePeriod;
+        }
+    }
+}
--- a/src/NzbDrone.Core/HealthCheck/HealthCheckService.cs
+++ b/src/NzbDrone.Core/HealthCheck/HealthCheckService.cs
@@ -83,6 +83,13 @@ namespace NzbDrone.Core.HealthCheck
             {
                 if (result.Type == HealthCheckResult.Ok)
                 {
+                    var previous = _healthCheckResults.Find(result.Source.Name);
+
+                    if (previous != null)
+                    {
+                        _eventAggregator.PublishEvent(new HealthCheckRestoredEvent(previous, !_hasRunHealthChecksAfterGracePeriod));
+                    }
+
                     _healthCheckResults.Remove(result.Source.Name);
                 }
                 else
--- a/src/NzbDrone.Core/Notifications/Discord/Discord.cs
+++ b/src/NzbDrone.Core/Notifications/Discord/Discord.cs
@@ -317,6 +317,29 @@ namespace NzbDrone.Core.Notifications.Discord
             _proxy.SendPayload(payload, Settings);
         }
 
+        public override void OnHealthRestored(HealthCheck.HealthCheck previousCheck)
+        {
+            var attachments = new List<Embed>
+                              {
+                                  new Embed
+                                  {
+                                      Author = new DiscordAuthor
+                                      {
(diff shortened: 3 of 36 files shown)


src/NzbDrone.Core/HealthCheck/HealthCheckService.cs
```diff
@@ -83,6 +83,12 @@ private void PerformHealthCheck(IProvideHealthCheck[] healthChecks)
             {
                 if (result.Type == HealthCheckResult.Ok)
                 {
+                    var previous = _healthCheckResults.Find(result.Source.Name);
+                    if (previous != null)
```
markus101, line 87: New line before `if` block

src/NzbDrone.Core/Notifications/Discord/Discord.cs
```diff
@@ -317,6 +317,29 @@ public override void OnHealthIssue(HealthCheck.HealthCheck healthCheck)
...
+            var attachments = new List<Embed>
+                              {
+                                  new Embed
+                                  {
+                                      Author = new DiscordAuthor
+                                      {
+                                          Name = Settings.Author.IsNullOrWhiteSpace() ? Environment.MachineName : Settings.Author,
+                                          IconUrl = "https://raw.githubusercontent.com/Sonarr/Sonarr/develop/Logo/256.png"
+                                      },
+                                      Title = "Successful " + previousCheck.Source.Name,
```
markus101, line 331: `Health Issue Resolved: previousCheck.Source.Name` to match other cases? `Successful` doesn't really cover it I feel.

src/NzbDrone.Core/HealthCheck/HealthCheckRestoredEvent.cs
```diff
@@ -0,0 +1,16 @@
+using NzbDrone.Common.Messaging;
+
+namespace NzbDrone.Core.HealthCheck
+{
+    public class HealthCheckRestoredEvent : IEvent
+    {
+        public HealthCheck PreviousCheck { get; private set; }
+        public bool IsInStartupGraceperiod { get; private set; }
```
markus101, line 8: `IsInStartupGracePeriod` eventually we should fix the health issue event as well.

src/NzbDrone.Core/Notifications/Slack/Slack.cs
```diff
@@ -134,6 +134,23 @@ public override void OnHealthIssue(HealthCheck.HealthCheck healthCheck)
             _proxy.SendPayload(payload, Settings);
         }
 
+        public override void OnHealthRestored(HealthCheck.HealthCheck previousCheck)
+        {
+            var attachments = new List<Attachment>
+                              {
+                                  new Attachment
+                                  {
+                                      Title = "Successful " + previousCheck.Source.Name,
```
markus101, line 143: Same thing here, `Successful` is too generic.

markus101 (changes requested):
Looking pretty good, just a few smaller comments.

src/NzbDrone.Core/HealthCheck/HealthCheckRestoredEvent.cs
CheAle14: Done, should I fix the other one whilst we're here or leave it?

src/NzbDrone.Core/Notifications/Slack/Slack.cs
CheAle14: I've removed the prefix entirely since Slack uses the message in the CreatePayload beneath to provide context on what the notification is for

src/NzbDrone.Core/HealthCheck/HealthCheckRestoredEvent.cs
markus101: You can leave it, I've already fixed it locally.

markus101 (approved):

PR #5618, opened 2023-05-03
Add trace log checkbox to bug report

bakerboy448:
Git


markus101:
We don’t need another field to guide people to add trace logs, we already have a field for that.

PR #5620, opened 2023-05-05
Fix usage of special characters in apikey

mynameisbogdan:
#### Database Migration
NO

#### Description
Allow usage of `#` in the API key.

Radarr/Radarr#8441

### DIFF
--- a/src/NzbDrone.Core/HealthCheck/Checks/ApiKeyValidationCheck.cs
+++ b/src/NzbDrone.Core/HealthCheck/Checks/ApiKeyValidationCheck.cs
@@ -1,0 +1,34 @@
+﻿using System.Text.RegularExpressions;
+using NLog;
+using NzbDrone.Core.Configuration;
+using NzbDrone.Core.Configuration.Events;
+using NzbDrone.Core.Lifecycle;
+
+namespace NzbDrone.Core.HealthCheck.Checks
+{
+    [CheckOn(typeof(ApplicationStartedEvent))]
+    [CheckOn(typeof(ConfigSavedEvent))]
+    public class ApiKeyValidationCheck : HealthCheckBase
+    {
+        private readonly IConfigFileProvider _configFileProvider;
+        private readonly Logger _logger;
+
+        public ApiKeyValidationCheck(IConfigFileProvider configFileProvider, Logger logger)
+        {
+            _configFileProvider = configFileProvider;
+            _logger = logger;
+        }
+
+        public override HealthCheck Check()
+        {
+            if (_configFileProvider.ApiKey.Length < 20)
+            {
+                _logger.Warn("Please update your API key to be at least 20 characters long. You can do this via settings or the config file");
+
+                return new HealthCheck(GetType(), HealthCheckResult.Warning, "Please update your API key to be at least 20 characters long. You can do this via settings or the config file", "#invalid-api-key");
+            }
+
+            return new HealthCheck(GetType());
+        }
+    }
+}
--- a/frontend/src/Calendar/iCal/CalendarLinkModalContent.js
+++ b/frontend/src/Calendar/iCal/CalendarLinkModalContent.js
@@ -40,7 +40,7 @@ function getUrls(state) {
     icalUrl += `tags=${tags.toString()}&`;
   }
 
-  icalUrl += `apikey=${window.Sonarr.apiKey}`;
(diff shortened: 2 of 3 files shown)


markus101:
Instead of trying to use an invalid API key that has been set in the config file should we throw an error on startup instead?

markus101:
Sure, but an API key set to a single character would also be valid, but something we should not allow. Since we don't really have a way to make sure users aren't overwriting the API key we should instead ensure it meets our expectations.

markus101:
Could do it as part of `ConfigFileProvider.ApiKey` if it's invalid just regenerate it. Checking it in application started event and logging an error/regenerating it there would work too.

Qstick:
I like the two prong approach. Fix the UI for do signalR doesn't blow up for url components controlled by users, and rework the key if it's no good. 

This looking good to me.

src/NzbDrone.Core/Configuration/ConfigFileProvider.cs
```diff
@@ -391,10 +392,25 @@ private string GenerateApiKey()
             return Guid.NewGuid().ToString().Replace("-", "");
         }
 
+        private void ValidateApiKey()
+        {
+            var apiKey = ApiKey;
+
+            if (new Regex("^[a-zA-Z0-9]{20,}$").IsMatch(apiKey))
```
Qstick, line 399: Do we want to enforce 20? This may kill alot of instances for people that are using custom keys (we know they were many based on some previous issues). 

Could we do something like at least 8, then health warning from 8-20? @markus101 thoughts here?

markus101: For Sonarr at least I think it’s fine since it’s going to hit in a major release, but is a somewhat breaking change so I can see showing a health check warning instead.

Qstick: True, I forget we are not released on 4 yet. We'll see if we here anything, in which case we can advise they can still set it during deployment, just needs to be 20+

Qstick: 30 might be a bit high for some for a health nag? I'm fine with it, just anticipating the issues.

Also Im not against the regenerate for very poor keys, just need to know that we might get some backlash from people setting via auto deployment.

Qstick: Yea, the check for now seems sane

src/NzbDrone.Core/HealthCheck/Checks/ApiKeyValidationCheck.cs
```diff
@@ -0,0 +1,36 @@
...
+using NzbDrone.Core.Configuration.Events;
+using NzbDrone.Core.Lifecycle;
+
+namespace NzbDrone.Core.HealthCheck.Checks
+{
+    [CheckOn(typeof(ApplicationStartedEvent))]
+    [CheckOn(typeof(ConfigSavedEvent))]
+    public class ApiKeyValidationCheck : HealthCheckBase
+    {
+        private static readonly Regex ValidateApiKeyRegex = new ("^[a-zA-Z0-9]{20,}$", RegexOptions.Compiled);
```
markus101, line 13: Since we’re just going to throw an health check back telling them to choose a more secure API key, should we instead just check for a minimum length and avoid a specific format? We’ll now handle anything they throw at us, but we should expect a minimum length.

This will also avoid us calling out a more complex 20 character token as insecure.

src/NzbDrone.Core/HealthCheck/Checks/ApiKeyValidationCheck.cs
```diff
@@ -0,0 +1,36 @@
...
+        {
+            _configFileProvider = configFileProvider;
+            _logger = logger;
+        }
+
+        public override HealthCheck Check()
+        {
+            if (!ValidateApiKeyRegex.IsMatch(_configFileProvider.ApiKey))
+            {
+                _logger.Warn("Please consider using a valid or more secure API Key in your configuration");
```
markus101, line 28: If we just focus on length we could change this to something like:

`Please update your API key to be at least 20 characters long. You can do this via settings or the config file`

markus101 (commented):
I like the move to health check, but worry that we’ll call out a more secure API key as less secure, so probably worth focusing on length instead of format.

PR #5622, opened 2023-05-05
Fixed: Make tld generic Website cleanse

bakerboy448:
#### Database Migration
NO

#### Description
Title

#### Todos
- [ ] Tests
- [] Wiki Updates

*

### DIFF
--- a/src/NzbDrone.Core/Parser/Parser.cs
+++ b/src/NzbDrone.Core/Parser/Parser.cs
@@ -462,11 +462,12 @@ namespace NzbDrone.Core.Parser
                                                                 string.Empty,
                                                                 RegexOptions.IgnoreCase | RegexOptions.Compiled);
 
-        private static readonly RegexReplace WebsitePrefixRegex = new RegexReplace(@"^\[\s*[-a-z]+(\.[a-z]+)+\s*\][- ]*|^www\.[a-z]+\.(?:com|net|org)[ -]*",
+        private static readonly Regex WebsiteRegex = new Regex(@"\[\s*[-a-z]+(\.[a-z]+)+\s*\][- ]*|^www\.[a-z]+(\.(?![0-9]*$)[a-z0-9]{2,24})[ -]*", RegexOptions.Compiled);
+        private static readonly RegexReplace WebsitePrefixRegex = new RegexReplace(@"^" + WebsiteRegex,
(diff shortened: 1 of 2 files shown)


src/NzbDrone.Core/Parser/Parser.cs
```diff
@@ -462,7 +462,7 @@ public static class Parser
                                                                 string.Empty,
                                                                 RegexOptions.IgnoreCase | RegexOptions.Compiled);
 
-        private static readonly RegexReplace WebsitePrefixRegex = new RegexReplace(@"^\[\s*[-a-z]+(\.[a-z]+)+\s*\][- ]*|^www\.[a-z]+\.(?:com|net|org)[ -]*",
+        private static readonly RegexReplace WebsitePrefixRegex = new RegexReplace(@"^\[\s*[-a-z]+(\.[a-z]+)+\s*\][- ]*|^www\.[a-z]+(\.[a-z]+)[ -]*",
```
markus101, line 465: Are there actually single character TLDs?

markus101 (changes requested):
> A few sentences describing the overall goals of the pull request's commits.

Mmmhmm.

This needs tests.

src/NzbDrone.Core/Parser/Parser.cs
Qstick: RFC-952 allows it, but there currently doesn't seem to be any and there is some debate in recent RFC proposals.. Also, they can be alpha numeric but can't be all numeric, and must be under 24 chars. 

```
A "name" (Net, Host, Gateway, or Domain name) is a text string up
to 24 characters drawn from the alphabet (A-Z), digits (0-9), minus
sign (-), and period (.).
```

```
There is an additional rule that essentially requires
     that top-level domain names not be all-numeric.
```

REF current list: https://data.iana.org/TLD/tlds-alpha-by-domain.txt

Maybe `^\[\s*[-a-z]+(\.[a-z]+)+\s*\][- ]*|^www\.[a-z]+(\.(?![0-9]*$)[a-z0-9]{2,24})[ -]*`

markus101:
@bakerboy448 are you going to address the regex?

markus101:
This regex had several failing tests still, I've redone in e273f16c3905e0c2451f43cf98b9b7ad1cbdc777

PR #5627, opened 2023-05-07
Catch and log errors when loading invalid config file

mynameisbogdan:
#### Database Migration
NO

#### Description
Cherry-picks from Prowlarr:
- Catch InvalidDataException during initial config to prevent boot loop
- Provide a nicer error message if the config file is invalid, and logging the exception so the user can see the stack trace to know exactly what was the issue with loading the XML file.


Qstick:
Markus, this clears quite a bit of sentry nonsense from certain config boot loops.

markus101 (approved):

PR #5594, opened 2023-04-29
Fixed: Various HealthCheck Improvements

bakerboy448:
#### Database Migration
NO

#### Description
- Align Radarr and Sonarr HealthChecks
- Backport various Radarr changes
- Add ProviderAddedEvent
- Cleanup usings
- Add HealthCheck Wiki Fragments
- Add filterBlockedClients to Download Clients

based partially on Radarr cae4faae61fe47ee6dacf334d8c4e4d483f5e5c8

#### Todos
- [] Tests
- [] Wiki Updates

#### Issues Fixed or Closed by this PR

*

PR #5628, opened 2023-05-08
Remove unused imports and constructors

mynameisbogdan:
#### Database Migration
NO

#### Description
Going from https://github.com/Sonarr/Sonarr/pull/5594#pullrequestreview-1415844773: remove used `using`, remove BOM and use LF line-endings.


markus101:
I'm fine with removing the unused usings, but changing the line endings is unnecessary churn on files.

markus101:
For BOM, are they causing issues?
For unused using statements is there a rule we can enable to prevent more from being added?

PR #5629, opened 2023-05-08
Fix migrations running in PostgreSQL

mynameisbogdan:
#### Database Migration
NO

#### Description
```
ALTER TABLE "public"."Notifications" ADD "OnHealthRestored" boolean NOT NULL DEFAULT 0;
The error was 42804: column "OnHealthRestored" is of type boolean but default expression is of type integer

  ----> Npgsql.PostgresException : 42804: column "OnHealthRestored" is of type boolean but default expression is of type integer
Data:
  Severity: ERROR
  InvariantSeverity: ERROR
  SqlState: 42804
  MessageText: column "OnHealthRestored" is of type boolean but default expression is of type integer
  Hint: You will need to rewrite or cast the expression.
```

### DIFF
--- a/src/NzbDrone.Core/Datastore/Migration/135_health_issue_notification.cs
+++ b/src/NzbDrone.Core/Datastore/Migration/135_health_issue_notification.cs
@@ -8,8 +8,8 @@ namespace NzbDrone.Core.Datastore.Migration
     {
         protected override void MainDbUpgrade()
         {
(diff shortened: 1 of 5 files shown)


markus101:
Postgres is not supported until https://github.com/Sonarr/Sonarr/pull/5559 lands. Most of these are already addressed here.

markus101:
Looks like 190 needs to land there, but that change should be made into that PR's branch.

PR #5636, opened 2023-05-11
New: Added UI for parsing release names

### LINKED ISSUE #5263: Add UI for Parse Endpoint

Sometimes I need to check whether something parses or not and for that, I use `/api/v3/parse?title=filename`. It works but isn't the most convenient experience. Existing options like Notifiarr's parse isn't ideal since it can't match the title with the correct show if it's a XEM alias. Plus a built-in solution would be preferable.

A front-end web page for the parse api in the web-ui making it easier to check parsing from Sonarr itself. 

using the parse api `/api/v3/parse?title=filename` or notifiarr's parser

### Anything else?

For reference, medusa has this:
image

markus101:
#### Database Migration
NO

#### Description
Adds a parse testing link under system. Not sure that System is the best spot for it, so open to suggestions there.

* Closes #5263


Qstick:
Parse text input box seems to prevent spaces during typing. However you can go back and add them after (or paste in fine). 

Also are Proper and Real supposed to be a int? If so, we may want to tell users what those mean. 
image

bakerboy448:
System (thinking generic system tool) or Activity (thinking what does this release do and parsing is an action) would make the most sense for location

Qstick:
IDK, it does feel weird in System where it is... though I'm having difficulty in thinking of a better spot

markus101:
> Also are Proper and Real supposed to be a int? If so, we may want to tell users what those mean.

Yeah they are, you can have some really fun things like proper.repack.real.proper, but probably sufficient to just show it as a boolean since that's what it is most of the time.

markus101:
Made some changes and also added a modal to Custom Formats settings and Series list for testing parsing. I don't think it makes sense to have it as a modal and under system as a separate page, but both are there until we decide where we want it to exist.

I'm leaning towards the modal, it's more flexible, the content doesn't take up the full screen width anyways and I like being able to link to it from multiple places.

markus101:
Modal (also available on Series list)
image

Dedicated page
image

frontend/src/System/Parse/ParseModalContent.tsx
```diff
@@ -0,0 +1,124 @@
...
+        dispatch(clear());
+      };
+    },
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    []
+  );
+
+  return (
+    <ModalContent onModalClose={onModalClose}>
+      <ModalHeader>Tags</ModalHeader>
```
Qstick, line 61: `Tags` to `Release Parser` or `Test Parsing` or something.

Qstick (commented):
I like the modal

PR #5640, opened 2023-05-12
New: Translations support for Health Checks

mynameisbogdan:
(cherry picked from commit bfc036178487fe0b692f306a53f2a334cdf7f9d5)

#### Database Migration
NO

#### Description
Start using the localization service a little, first with the heath checks.


markus101:
Looks like some of the backporting I did stomped on things this is changing and this will need a rebase.

Thanks for pulling this one upstream, it looked pretty gnarly when I looked at it.

src/NzbDrone.Core/Localization/Core/en.json
```diff
@@ -1,6 +1,60 @@
 {
+  "ApiKeyValidationHealthCheckMessage": "Please update your API key to be at least {0} characters long. You can do this via settings or the config file",
```
markus101, line 2: Similar to the UI the keys should be the English translation

markus101 (changes requested):
Changes look good, but the keys for the translations should be the English translation.

markus101:
> I've seen the discussion from 2 years ago and still no improvements were made regarding multi-languages were made.

Translation support was only added in v4, which is still in beta.

> You really should reconsider this approach since it actually doesn't improve DX at all.

I disagree, I've worked on several projects using English translations as the key, I find it a lot easier to know the text without having to look it up in a giant dictionary of keys and values, similarly, unless translating from another language you can use the text in logs/messages to locate the code without having to look up the key from the value in a table. That said, DX is subjective and "feely".

> For example if there's a typo, you'll make all other languages invalid.

That does happen, but it hasn't been too problematic in my experience. This problem also extends to keys, if poor keys are used then fixing them will lead to loss of translations, though less of a concern if they are well constructed.

> The whole point of this PR was to make it easier to share code between arrs, but might as well close it.

This is probably the biggest reason to stick with the status quo as much as I'd like to go another route, unnecessary conflicts are just that, unnecessary.

markus101:
Haven't had time to come back to review with fresh eyes, but will when I can.

markus101 (approved):
Looks good, just needs a rebase.

PR #5642, opened 2023-05-13
Update UI dependencies

mynameisbogdan:
#### Database Migration
NO

#### Description
I wanted to fix formatBytes.js for use with TS, but I ended up updating the minor versions of outdated front-end libraries as well.


frontend/src/Utilities/Number/formatBytes.js
```diff
@@ -10,7 +10,7 @@ function formatBytes(input) {
   return filesize(size, {
     base: 2,
     round: 1
-  });
+  }).toString();
```
markus101, line 13: Wouldn’t this already be a string?

markus101: Given that it could be any[] or a specific object shape, calling toString on it could lead to something unexpected.

Converting to TS then checking the type before toString would ensure we got an expected result. This return type has me a little unsure how to know what we’re getting back.

PR #5651, opened 2023-05-16
 New: Add path mapping for partial library updates in Plex Server Notifications

mynameisbogdan:
(cherry picked from commit 24206ad0a3095c4bf5c860af516d8543bd6525d6)

#### Database Migration
NO

#### Description
Nice addition for partial scan when paths don't match.

### DIFF
--- a/src/NzbDrone.Core/Notifications/Plex/Server/PlexServerSettings.cs
+++ b/src/NzbDrone.Core/Notifications/Plex/Server/PlexServerSettings.cs
@@ -1,4 +1,5 @@
 using FluentValidation;
+using NzbDrone.Common.Extensions;
 using NzbDrone.Core.Annotations;
 using NzbDrone.Core.ThingiProvider;
 using NzbDrone.Core.Validation;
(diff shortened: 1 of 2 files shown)


Qstick:
Might want to change the name here, path scan is already supported on Sonarr, this just adds mapping capability

src/NzbDrone.Core/Notifications/Plex/Server/PlexServerSettings.cs
```diff
@@ -43,6 +46,12 @@ public PlexServerSettings()
         [FieldDefinition(5, Label = "Update Library", Type = FieldType.Checkbox)]
         public bool UpdateLibrary { get; set; }
 
+        [FieldDefinition(6, Label = "Map Paths From", Type = FieldType.Textbox, Advanced = true, HelpText = "Sonarr Path, Used to modify series paths when Plex sees library path location differently from Sonarr")]
```
markus101, line 49: `Sonarr path, used to modify series paths when Plex sees library path location differently from Sonarr`

or

`Sonarr path. Used to modify series paths when Plex sees library path location differently from Sonarr`

would be better. Similar issue with the `MapTo` path.

PR #5656, opened 2023-05-18
Add `forceSave` to the OpenAPI docs

mynameisbogdan:
#### Database Migration
NO

#### Description
I didn't add this in #5612 since I'm wary about this, but since it's something supported I'll leave it up to you to decide.


markus101:
Seems reasonable to add since we support it.

PR #5668, opened 2023-05-20
New: Improve validation messages

mynameisbogdan:
#### Database Migration
NO

#### Description
This is redundant from the start when using the UI, but logs with `Path: Folder is not writable by user abc` or `RootFolderPath: Invalid Path` aren't helpful at all for support or the users.

Should I continue with the rest of them?


markus101:
If we're not logging the paths at all in the logs then we should update them, if they have to scroll up one line to see the path used then I don't see a ton of value since we shouldn't just be looking at an error.

markus101:
Yeah, we definitely shouldn't log in the validator. Adding more information to the validation message as you've done is a good idea.

markus101 (approved):

PR #5670, opened 2023-05-21
Fix tests

mynameisbogdan:
#### Database Migration
NO

#### Description
TorrentDownloadStationFixture: Conditions changed in 3cd33d3f44097b4cb4fb291bca70a0aa53c4b844
IndexerServiceFixture: while okay on CI, now seems to fail locally on Mac.


markus101:
Odd that mac started failing locally, was there an update that changed something?

PR #5674, opened 2023-05-21
Add import list clean library option

### LINKED ISSUE #5201: Delete series and files via list sync

Using the List import feature for a separate 4K library, it looks like Radarr allows a List Update Interval and Clean Library Level option, Sonarr does not.  In Radarr I can manage the 4K library by utilizing a sync'd profile and the file will delete from my datastore automatically, without having to log in and manage the 4K instance manually.

Files attached for Radarr and Sonarr List import option screenshots, for comparison.
Radarr
Sonarr

Sonarr to have the same or similar capability that Radarr does, so that the 4K library can be managed 100% through my existing, standard Sonarr instance.

Manually log into the Sonarr4K library to delete old series occasionally.

### Anything else?

Found this issue when comparing the Radarr and Sonarr sync guides found on TRaSH guides:

https://trash-guides.info/Sonarr/Tips/Sync-2-radarr-sonarr/

CheAle14:
#### Database Migration
YES (193)

#### Description
Backports Radarr's import list clean library level option, with the frontend component converted to typescript. Stores each list's import list items in the database. 
I've kept the `createSettingSectionSelector` typing in a separate commit as its not directly related; I can drop it if its not wanted.  

* Closes #5201


src/NzbDrone.Core/Configuration/IConfigService.cs
```diff
@@ -49,6 +49,9 @@ public interface IConfigService
         int MaximumSize { get; set; }
         int MinimumAge { get; set; }
 
+        int ImportListSyncInterval { get; set; }
```
markus101, line 52: This setting isn't valuable to Sonarr. The scheduled task runs every 5 minutes and each list has it's own pre-configured interval, which for Sonarr syncing is the same 5 minute interval, for Plex Watchlist, it's 6 hours.

markus101 (changes requested):
The customizable interval doesn't apply to Sonarr, I believe the same change will eventually make it's way to Radarr.

Qstick:
Correct, Radarr will function the same soon enough

src/NzbDrone.Core/Configuration/IConfigService.cs
CheAle14: I somehow managed to read the task's interval as 5 hours so thought the setting made sense. I've removed it now

Qstick:
The clean library option is a bit dangerous, I almost wish we never had it in Radarr. There are a few things we need to make sure work correctly if we are bringing this in Sonarr. One of the main ones is ensuring cleaning doesn't happen on a partial or errored list pull.

CheAle14:
There are tests which should check that cleaning doesn't occur when any list returns with a failure, but I've just realised a problem whilst thinking about the partial pulls. Cleaning only occurs when all of the lists have been fetched (at the same time),  but since Sonarr essentially fetches each list on its own interval (using the last sync time + min refresh interval) they _probably_ wouldn't line up, so at least one list would be missing, and cleaning would never occur with multiple lists.     

I think this would require keeping track of what lists each series belongs to (like Radarr does). Cleaning could then be done when all lists have been 'recently' fetched (e.g. within half their refresh interval?) on series that don't belong to any of them. I'll mark this as a draft until I've done that.

CheAle14:
Import lists' items are now stored to the database, and clean library occurs on its own scheduled task that starts off disabled, whenever an item is removed from a list that task gets set to a 5 minute interval so it can repeatedly check whether all lists are recent. Once they are, the library is cleaned (comparing against the stored list items) and the task is disabled again. If a list fails then its last sync time is set to null, which ensures that it is never seen as 'recent' until it succeeds again, so the library won't be cleaned until that happens.

markus101 (changes requested):
Haven't reviewed fully, but some of the front end changes are failing lint checks.

https://build.sonarr.tv/buildConfiguration/Sonarr_V4PullRequests_1BuildFrontend/196028?buildTab=log&focusLine=0&logView=flowAware&linesState=86.113

markus101:
@Qstick any suggestions for handling this better in Sonarr? I see the advantage of this functionality, but partially fetched lists or a list being temporarily offline has the potential to really muck things up.

If we tracked the list an item came from going forward, would that be more useful so we could decide on a list by list basis?

CheAle14:
I've made it keep track of the number of items on the last fetch, and if there's a big difference (25%) it blocks any cleaning until the user approves (or until the number returns to normal, if it were a temporary issue/outage):
image
image

src/NzbDrone.Core/ImportLists/FetchAndParseImportListService.cs
```diff
@@ -86,16 +99,37 @@ public List<ImportListItemInfo> Fetch()
...
-        public List<ImportListItemInfo> FetchSingleList(ImportListDefinition definition)
+        private void updateListStatus(int listId, List<Parser.Model.ImportListItemInfo> listReports, ImportListFetchResult result)
+        {
+            var previousCount = _importListStatusService.GetImportListStatus(listId).LastCount ?? 0;
+            var currentCount = listReports.Count;
+
+            if (previousCount > 10)
+            {
+                var percentageDiff = Math.Abs((currentCount - previousCount) / (double)(previousCount + currentCount));
+                if (percentageDiff > 0.25)
```
markus101, line 117: New line before block

src/NzbDrone.Core/ImportLists/FetchAndParseImportListService.cs
```diff
@@ -114,16 +148,20 @@ public List<ImportListItemInfo> FetchSingleList(ImportListDefinition definition)
...
+                    var importListReports = fetchResult.Series;
 
                     lock (result)
                     {
                         _logger.Debug("Found {0} reports from {1} ({2})", importListReports.Count, importList.Name, importListLocal.Definition.Name);
 
-                        result.AddRange(importListReports);
+                        result.Series.AddRange(importListReports);
+                        result.AnyFailure |= fetchResult.AnyFailure;
+                        if (!fetchResult.AnyFailure)
```
markus101, line 160: New line before block

src/NzbDrone.Core/ImportLists/ImportListItems/ImportListItemRepository.cs
```diff
@@ -0,0 +1,42 @@
...
+
+        public List<ImportListItemInfo> GetAllForLists(List<int> listIds)
+        {
+            return Query(x => listIds.Contains(x.ImportListId));
+        }
+
+        public bool Exists(int tvdbId, string imdbId)
+        {
+            List<ImportListItemInfo> items;
+            if (string.IsNullOrWhiteSpace(imdbId))
```
markus101, line 30: New line before block

markus101 (changes requested):
Discussing if whether this approach makes sense with @Qstick, though at first glance the UI will need some work.

frontend/src/Store/Actions/Settings/importListOptions.js
```diff
@@ -0,0 +1,64 @@
...
+  return {
+    section,
+    ...payload
+  };
+});
+
+//
+// Details
+
+export default {
```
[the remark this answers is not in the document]
CheAle14: All of the other `Store/Actions/Settings/*.js` things appear to use a big `default` block, as it seems they're all just imported together in `settingsActions.js` rather than separately. I can still make them separate if you wish (just checking before I make it inconsistent)

frontend/src/Settings/ImportLists/ImportLists/ImportList.js
```diff
@@ -86,11 +90,20 @@ class ImportList extends Component {
           </Label>
         </div>
 
+        {
+          blockingCount >= 0 ? <div className={styles.enabled}>
```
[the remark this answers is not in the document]
CheAle14: As above, that was poorly worded; it would be 0 if the list previously returned some items and recently returned none at all.

markus101:
I've discussed with QStick and we have concerns that this functionality will be too buried to be useful and people will have to figure out things are not working then stumble onto settings and see it was blocked.

Based on this and conversations with the various support people I think it makes sense that we don't add the block/approval process at this time and instead of the options for `Remove and Keep` and `Remove and Delete` there is a `Keep and Tag` option to add a tag (and a corresponding option for specifying the tag).

If someone really wants to remove series automatically they could set the tag and have something hit the API and remove it, turning it from a simple setting to a more complex setup that requires the forethought to understand what it's going to do.

src/NzbDrone.Core/ImportLists/ImportListSyncService.cs
```diff
@@ -1,43 +1,91 @@
...
+
+                var lastTime = _importListStatusService.GetLastSyncListInfo(list.Definition.Id);
+                if (!lastTime.HasValue)
+                {
+                    // list has never been synced, or failed the last time it did.
+                    return false;
+                }
+
+                var minimum = Math.Max((list.MinRefreshInterval / 2).TotalMinutes, 5);
+                if ((DateTime.UtcNow - lastTime.Value).TotalMinutes > minimum)
```
markus101, line 80: Same thing here

src/Sonarr.Api.V3/Config/ImportListConfigController.cs
```diff
@@ -0,0 +1,26 @@
...
+
+    public class ImportListConfigController : ConfigController<ImportListConfigResource>
+    {
+        public ImportListConfigController(IConfigService configService)
+            : base(configService)
+        {
+            SharedValidator.RuleFor(x => x.ListSyncTag)
+               .ValidId()
+               .WithMessage("Tag must be specified")
+               .When(x => x.ListSyncLevel == NzbDrone.Core.ImportLists.ListSyncLevelType.KeepAndTag);
```
markus101, line 18: Add a using statement for `NzbDrone.Core.ImportLists` and change to `ListSyncLevelType.KeepAndTag`

markus101 (changes requested):

src/NzbDrone.Core/ImportLists/FetchAndParseImportListService.cs
```diff
@@ -64,16 +66,28 @@ public List<ImportListItemInfo> Fetch()
...
+                             var importListReports = fetchResult.Series;
 
                              lock (result)
                              {
                                  _logger.Debug("Found {0} reports from {1} ({2})", importListReports.Count, importList.Name, importListLocal.Definition.Name);
 
-                                 result.AddRange(importListReports);
+                                 if (fetchResult.AnyFailure)
+                                 {
+                                     _importListStatusService.ClearListSyncStatus(importList.Definition.Id);
```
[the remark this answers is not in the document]
CheAle14: Yes, I've changed it to use the `DisabledTill` property from the provider base now to tell when a failure has occurred.

CheAle14:
I've gotten rid of the extra task, and to avoid attempting to clean after every single sync I've added a boolean to the `ImportListStatus` to track whether any items on that list have been removed since the last time cleaning occured, so it should only clean when there might actually be something to remove.

markus101 (approved):
Thanks for addressing everything.

PR #5680, opened 2023-05-24
Enforce rule IDE0005 on build

mynameisbogdan:
#### Database Migration
NO

#### Description
Since c6b543e0729bc20f15e37d074fbf31d8c76c187a apparently a few unused import already got in, thus let's add a Github Action to prevent this.

I noticed the build.sh is little different from downstream and if I'm assuming correctly that in TC it's used with `build.sh --all` and not step by step like `build.sh --lint`, then I think I can add `LintBackend` to run on `$LINT_BACKEND == "YES"`.

### DIFF
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -30,6 +30,13 @@
     <!-- A test project gets the test sdk packages automatically added -->
     <TestProject>false</TestProject>
     <TestProject Condition="$(MSBuildProjectName.EndsWith('.Test'))">true</TestProject>
+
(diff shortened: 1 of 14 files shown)


markus101:
Backend: `./build.sh --backend --enable-extra-platforms --packages`
Frontend: `./build.sh --lint --frontend`

Ideally this would get caught during compilation and the build would fail regardless of a PR being opened and failing via GH action.

markus101 (approved):

PR #5685, opened 2023-05-24
Fixed: those stupid lines

bakerboy448:
just some trapped hamster's typos


Qstick:
Does this really need to be a changlog commit? ie. `Fixed:`

bakerboy448:
Feels like it was in line with the humor of the messages

PR #5686, opened 2023-05-25
 Fix sorting search releases by language

mynameisbogdan:
#### Database Migration
NO

#### Description
This is to fix sorting by empty languages in https://github.com/Sonarr/Sonarr/blob/f05405fe1ce4c78a8c75e27920c863c5b83686bd/frontend/src/Store/Actions/releaseActions.js#L47 which does the same as https://github.com/Sonarr/Sonarr/blob/f05405fe1ce4c78a8c75e27920c863c5b83686bd/src/NzbDrone.Core/Parser/ParsingService.cs#L194 for a successful parsing. 

Before
before
After
after

### DIFF
--- a/frontend/src/Store/Actions/releaseActions.js
+++ b/frontend/src/Store/Actions/releaseActions.js
@@ -44,7 +44,7 @@ export const defaultState = {
         return 10000;
       }
 
-      return item.languages[0].id;
+      return item.languages[0]?.id ?? 0;
     },
 
     rejections: function(item, direction) {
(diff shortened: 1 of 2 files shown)


frontend/src/Store/Actions/releaseActions.js
```diff
@@ -44,7 +44,7 @@ export const defaultState = {
         return 10000;
       }
 
-      return item.languages[0].id;
+      return item.languages[0]?.id ?? 0;
```
Qstick, line 47: Is this solved by the fixed to DownloadDecisionMaker, or are their other cases of no language?

PR #5691, opened 2023-05-26
Fixed: Enforce validation warnings

mynameisbogdan:
#### Database Migration
NO

#### Description
~~Validation with `AsWarning()` doesn't work nice with arrs, and this might a regression from b95c4d37d7e1c2d699dbfd995b496a2fffde6705 allowing use of <= 0 values.~~

~~Since I fixed this in Prowlarr a while ago a few users [complained](https://github.com/Prowlarr/Prowlarr/issues/1692) that Prowlarr doesn't allow `0` and Sonarr/Radarr does.~~

~~Should we allow >= 0 or stick to `>0`?~~

Fix showing warnings in provider validation.

### DIFF
--- a/frontend/src/Store/Actions/Creators/createSaveProviderHandler.js
+++ b/frontend/src/Store/Actions/Creators/createSaveProviderHandler.js
@@ -32,9 +32,9 @@ function createSaveProviderHandler(section, url, options = {}) {
     const params = { ...queryParams };
 
     // If the user is re-saving the same provider without changes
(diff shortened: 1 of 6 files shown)


markus101:
> Validation with AsWarning() doesn't work nice with arrs

How so?
It's intention is to show the warning, but still allow the user to save.

markus101:
That is an issue, but we should fix the problem instead of removing it because it's not working.

markus101 (approved):
Looks good, thanks for fixing!

PR #5694, opened 2023-05-27
Faster tag view in UI for large libraries

Qstick:
#### Database Migration
NO

#### Description
Avoid `AllSeries` call on Tag UI page, particularly detrimental on large libraries. 

Pulled from Radarr


markus101 (approved):
Looks good.

PR #5696, opened 2023-05-28
Fix priority input for provider manage edit modal

Qstick:
#### Database Migration
NO

#### Description
Avoid Manage Edit modal from crashing on click in and out of priority number input

#### Todos
- [ ] Tests
- [ ] Wiki Updates

### DIFF
--- a/frontend/src/Components/Form/NumberInput.js
+++ b/frontend/src/Components/Form/NumberInput.js
@@ -10,7 +10,7 @@ function parseValue(props, value) {
   } = props;
 
   if (value == null || value === '') {
-    return min;
+    return null;
   }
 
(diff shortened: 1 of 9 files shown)


frontend/src/Components/Form/NumberInput.js
```diff
@@ -10,7 +10,7 @@ function parseValue(props, value) {
   } = props;
 
   if (value == null || value === '') {
-    return min;
+    return null;
```
Qstick, line 13: This ensures that a '' or null value in the input doesn't get changed to min value on a click in and out.

markus101 (approved):

PR #5705, opened 2023-05-31
Add style for default kind in ProgressBar

mynameisbogdan:
(cherry picked from commit 4ab1cb393ac682cb06625f42ffc6131e9bd72500)

#### Database Migration
NO

#### Description
This fixes in advance any future usage of the default kind in ProgressBar, since default + 100% progressbar would result in something like this:

firefox_2023-05-30_19-03-56

### DIFF
--- a/frontend/src/Components/ProgressBar.css
+++ b/frontend/src/Components/ProgressBar.css
@@ -45,6 +45,10 @@
   cursor: default;
 }
 
+.default {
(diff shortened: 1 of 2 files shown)


markus101:
This isn’t a progress bar kind we want to support, instead of allowing it to be used, even by mistake we should instead prevent it from being used because it’s not something intended to be used.

PR #5706, opened 2023-05-31
New: Download Client Tags

### LINKED ISSUE #402: Tags for download clients
This would use the tagged download client if it matches a series (same as notifications).

We'd need a health check in the event there is not a download client untagged (default download client) or at least to validate that every series has a corresponding download client via a tag.

Qstick:
#### Database Migration
YES (191)

#### Description
Adds tags to download clients to allow sending series to a certain client. May want to gut check the logic to ensure it's functioning as we believe it should. 

This would currently filter all clients down to those with no tag or a matching tag when a release is sent from a series. The round robin logic then kicks in with the clients are are left after the tag filter. (similar to how we search indexers with matching tags)

#### Todos
- [x] Add some tests

* Fixes #402


src/NzbDrone.Core/Extras/Metadata/Consumers/Xbmc/XbmcMetadata.cs
```diff
@@ -24,19 +24,19 @@ public class XbmcMetadata : MetadataBase<XbmcMetadataSettings>
     {
         private readonly Logger _logger;
         private readonly IMapCoversToLocal _mediaCoverService;
-        private readonly ITagService _tagService;
+        private readonly ITagRepository _tagRepo;
```
Qstick, line 27: Had to switch this direct to repo to avoid a recursive dependency loop (outlined below)

image

markus101 (approved):
Changes look good, but it'd be good to move towards translated strings with these new additions.

frontend/src/Settings/DownloadClients/DownloadClients/Manage/Tags/TagsModalContent.tsx
```diff
@@ -0,0 +1,180 @@
...
+import DownloadClient from 'typings/DownloadClient';
+import styles from './TagsModalContent.css';
+
+interface TagsModalContentProps {
+  ids: number[];
+  onApplyTagsPress: (tags: number[], applyTags: string) => void;
+  onModalClose: () => void;
+}
+
+function TagsModalContent(props: TagsModalContentProps) {
```
markus101, line 28: Also probably worth DRYing this out at some point since it's mostly the same thing as applying tags in other places.

PR #5711, opened 2023-06-03
Fixed: Treat redirects as errors in Sonarr Import List

mynameisbogdan:
#### Database Migration
NO

#### Description
Fix for silent errors with null results when redirects happen due to possible wrong URL base, making `GetSeries` successful in tests but not in real use.

### DIFF
--- a/src/NzbDrone.Core/ImportLists/Sonarr/SonarrV3Proxy.cs
+++ b/src/NzbDrone.Core/ImportLists/Sonarr/SonarrV3Proxy.cs
@@ -69,6 +69,12 @@ namespace NzbDrone.Core.ImportLists.Sonarr
                     return new ValidationFailure("ApiKey", "API Key is invalid");
                 }
 
+                if (ex.Response.HasHttpRedirect)
+                {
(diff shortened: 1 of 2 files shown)


src/NzbDrone.Core/ImportLists/Sonarr/SonarrV3Proxy.cs
```diff
@@ -90,11 +96,18 @@ private List<TResource> Execute<TResource>(string resource, SonarrSettings setti
...
-            var request = new HttpRequestBuilder(baseUrl).Resource(resource).Accept(HttpAccept.Json)
-                .SetHeader("X-Api-Key", settings.ApiKey).Build();
+            var request = new HttpRequestBuilder(baseUrl).Resource(resource)
+                .Accept(HttpAccept.Json)
+                .SetHeader("X-Api-Key", settings.ApiKey)
+                .Build();
 
             var response = _httpClient.Get(request);
 
+            if ((int)response.StatusCode > 299)
```
Qstick, line 106: Can we just do `response.HasHttpRedirect` here? HttpClient already throws for 400 and up given SuppressHttpError is false, right?

PR #5712, opened 2023-06-03
Add HelpTextWarning support in FieldDefinition

mynameisbogdan:
#### Database Migration
NO

#### Description
Sometimes it's useful to use `helpTextWarning` in the UI.


markus101:
Is there a concrete example where this would be used?

PR #5716, opened 2023-06-04
New: More info for script import and custom script

JeWe37:
#### Database Migration
NO

#### Description
In the process of writing the script to actually use #5538 I found a few more environment variables that would be useful for that and also added those in the corresponding places in the custom scripts. Leaving this as a draft for now until I have finished said script and verified I didn't miss any others that might be useful.

#### Todos
---

---

### DIFF
--- a/src/NzbDrone.Core/MediaFiles/ScriptImportDecider.cs
+++ b/src/NzbDrone.Core/MediaFiles/ScriptImportDecider.cs
@@ -7,7 +7,9 @@ using NzbDrone.Common.Extensions;
 using NzbDrone.Common.Processes;
 using NzbDrone.Core.Configuration;
 using NzbDrone.Core.MediaFiles.MediaInfo;
+using NzbDrone.Core.Parser;
 using NzbDrone.Core.Parser.Model;
+using NzbDrone.Core.Tags;
 
 namespace NzbDrone.Core.MediaFiles
 {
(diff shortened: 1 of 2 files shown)


JeWe37:
Alright, seems that this is all that will be needed-there's always more info that could be added, but I figure till the day that someone needs it, it's best to keep it to a minimum.

src/NzbDrone.Core/MediaFiles/ScriptImportDecider.cs
```diff
@@ -22,18 +24,21 @@ public class ImportScriptService : IImportScript
...
                                    IConfigService configService,
                                    IConfigFileProvider configFileProvider,
+                                   ITagRepository tagRepository,
                                    Logger logger)
         {
             _processProvider = processProvider;
             _videoFileInfoReader = videoFileInfoReader;
             _configService = configService;
             _configFileProvider = configFileProvider;
+            _tagRepository = tagRepository;
```
markus101, line 41: This should use the `TagService` instead of `TagRepository` (same for other uses).

markus101 (changes requested):

src/NzbDrone.Core/MediaFiles/ScriptImportDecider.cs
JeWe37: That's what I tried first, but that actually gives a dependency loops because the TagService was involved in sending notifications, but obviously notifications depend on the various custom scripts, so those can't depend on the TagService. Using the repository directly breaks that dependency cycle.

markus101 (changes requested):
I thought I submitted an updated review, but I guess I never clicked it.

JeWe37:
Alright fixed that up

markus101 (approved):

markus101:
Thanks for the fixes

PR #5733, opened 2023-06-12
New: Smarter media info audio stream selection

Qstick:
#### Database Migration
NO

#### Description
Be smarter about which stream we pick

#### Todos
- [ ] Tests
- [ ] Wiki Updates

### DIFF
--- a/src/NzbDrone.Core/MediaFiles/MediaInfo/VideoFileInfoReader.cs
+++ b/src/NzbDrone.Core/MediaFiles/MediaInfo/VideoFileInfoReader.cs
@@ -69,11 +69,13 @@ namespace NzbDrone.Core.MediaFiles.MediaInfo
 
                 var analysis = FFProbe.AnalyseStreamJson(ffprobeOutput);
                 var primaryVideoStream = GetPrimaryVideoStream(analysis);
+                var primaryAudioStream = GetPrimaryAudioStream(analysis);
 
-                if (analysis.PrimaryAudioStream?.ChannelLayout.IsNullOrWhiteSpace() ?? true)
+                if (primaryAudioStream?.ChannelLayout.IsNullOrWhiteSpace() ?? true)
                 {
                     ffprobeOutput = FFProbe.GetStreamJson(filename, ffOptions: new FFOptions { ExtraArguments = "-probesize 150000000 -analyzeduration 150000000" });
                     analysis = FFProbe.AnalyseStreamJson(ffprobeOutput);
+                    primaryAudioStream = GetPrimaryAudioStream(analysis);
                 }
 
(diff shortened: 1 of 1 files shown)


src/NzbDrone.Core/MediaFiles/MediaInfo/VideoFileInfoReader.cs
```diff
@@ -174,6 +176,20 @@ private VideoStream GetPrimaryVideoStream(IMediaAnalysis mediaAnalysis)
             return mediaAnalysis.VideoStreams.FirstOrDefault(s => !codecFilter.Contains(s.CodecName)) ?? mediaAnalysis.PrimaryVideoStream;
         }
 
+        private AudioStream GetPrimaryAudioStream(IMediaAnalysis mediaAnalysis)
```
bakerboy448, line 179: `GetPrimaryOrBestAudioStream` no?

src/NzbDrone.Core/MediaFiles/MediaInfo/VideoFileInfoReader.cs
```diff
@@ -174,6 +176,20 @@ private VideoStream GetPrimaryVideoStream(IMediaAnalysis mediaAnalysis)
...
+        {
+            var audioStreams = mediaAnalysis.AudioStreams;
+
+            if (audioStreams.Count <= 1 || audioStreams.Any(s => s.BitRate == 0))
```
markus101, line 183: Should the `Any` be `All` since the sorting should take care of picking the best stream.

src/NzbDrone.Core/MediaFiles/MediaInfo/VideoFileInfoReader.cs
markus101: This is similar to getting the video stream, which isn't always the primary stream, I think the naming is fine to maintain that consistency.

markus101 (approved):
Just the one question, otherwise looks good.

JeWe37:
I hope you don't mind if I chip in, but I've been working on a custom import script relying on the current behavior, and given some of my experience with that, I'm generally not sure that this is the best idea. For one, bitrate doesn't necessarily mean much as different codecs and different encoders can result in significantly different bitrates. There is also the matter of different codecs potentially having features(e.g., Atmos), that should make them "primary" but which this doesn't cover.

I think if anything what may be worth considering is the disposition(default audio tracks) and potentially selecting the track by its language tag. Failing that, it might be best to just go for the first track as is currently the case.

Qstick:
I tend to agree that this needs a lot more logic/testing than I currently have the time it patience too put into it right now.

PR #5736, opened 2023-06-13
Show loading errors as Alerts

mynameisbogdan:
before
after

### DIFF
--- a/frontend/src/Activity/Blocklist/Blocklist.js
+++ b/frontend/src/Activity/Blocklist/Blocklist.js
@@ -1,5 +1,6 @@
 import PropTypes from 'prop-types';
 import React, { Component } from 'react';
+import Alert from 'Components/Alert';
 import LoadingIndicator from 'Components/Loading/LoadingIndicator';
 import ConfirmModal from 'Components/Modal/ConfirmModal';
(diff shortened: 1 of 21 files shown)


doomwithdon:
Build broken due to this commit with the following error. Import statements missing in some files.

Sonarr/frontend/src/RootFolder/RootFolders.js
  46:8   error  'Alert' is not defined  react/jsx-no-undef
  46:20  error  'kinds' is not defined  no-undef

doomwithdon:
> > Build broken due to this commit with the following error. Import statements missing in some files.
> 
> You're right, but I think I'm going insane since I'm 100% sure the UI build/lint was okay.
> 
> I'm not going insane, both `yarn build --env=development` and `yarn build --env=production` are building fine.

Still failing on my end with that error. Am i missing something. The build was fine before this commit.

doomwithdon:
> > Still failing on my end with that error. Am i missing something. The build was fine before this commit.
> 
> image
> 
> 🤷‍♂️

build successful after adding missing imports as in this commit
https://github.com/Sonarr/Sonarr/commit/5601a940164db41257e1ab9008db58d79a2e49f3

PR #5750, opened 2023-06-20
Create overload for `ToJson()` with Formatting param

mynameisbogdan:
#### Database Migration
NO

#### Description
- Added overload for use with `x.ToJson(Formatting.None)` to prevent spamming the logs across multiple lines.
- As an example added ContentSummary for HDBits for easier debugging.


markus101:
Looks like this needs to be rebased.

PR #5757, opened 2023-06-23
No thumb tags with local images in .nfo

JeWe37:
#### Database Migration
NO

#### Description
According to the [Jellyfin docs](https://jellyfin.org/docs/general/server/metadata/nfo/#image-paths-and-urls-in-nfo-files), .nfo file thumb image tags are prioritized over local images. However Sonarr always adds the thumb image tags, meaning Jellyfin attempts to grab the images from the internet rather than using those that are available locally. This PR simply removes the tags if episode images are enabled. 

#### Todos
- [ ] Tests
- [ ] Wiki Updates

### DIFF
--- a/src/NzbDrone.Core/Extras/Metadata/Consumers/Xbmc/XbmcMetadataSettings.cs
+++ b/src/NzbDrone.Core/Extras/Metadata/Consumers/Xbmc/XbmcMetadataSettings.cs
@@ -36,13 +36,16 @@ namespace NzbDrone.Core.Extras.Metadata.Consumers.Xbmc
         [FieldDefinition(3, Label = "Episode Metadata", Type = FieldType.Checkbox, Section = MetadataSectionType.Metadata, HelpText = "<filename>.nfo")]
         public bool EpisodeMetadata { get; set; }
 
-        [FieldDefinition(4, Label = "Series Images", Type = FieldType.Checkbox, Section = MetadataSectionType.Image, HelpText = "fanart.jpg, poster.jpg, banner.jpg")]
+        [FieldDefinition(4, Label = "Episode Metadata Image Thumbs", Type = FieldType.Checkbox, Section = MetadataSectionType.Image, HelpText = "Include image thumb tags in <filename>.nfo (Requires 'Episode Metadata')", Advanced = true)]
+        public bool EpisodeImageThumb { get; set; }
+
+        [FieldDefinition(5, Label = "Series Images", Type = FieldType.Checkbox, Section = MetadataSectionType.Image, HelpText = "fanart.jpg, poster.jpg, banner.jpg")]
         public bool SeriesImages { get; set; }
 
-        [FieldDefinition(5, Label = "Season Images", Type = FieldType.Checkbox, Section = MetadataSectionType.Image, HelpText = "season##-poster.jpg, season##-banner.jpg, season-specials-poster.jpg, season-specials-banner.jpg")]
+        [FieldDefinition(6, Label = "Season Images", Type = FieldType.Checkbox, Section = MetadataSectionType.Image, HelpText = "season##-poster.jpg, season##-banner.jpg, season-specials-poster.jpg, season-specials-banner.jpg")]
         public bool SeasonImages { get; set; }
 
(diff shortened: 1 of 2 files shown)


markus101:
This particular metadata is first designed for Kodi's consumption, so how is that affected? Seems a bit odd that Jellyfin would half use local metadata, half use online metadata.

JeWe37:
Took a bit of digging, but this behavior aligns with what Kodi does as well, it states as much in the [docs](https://kodi.wiki/view/NFO_files/Movies#nfo_Tags)(where it explains the thumb tags) and I've also confirmed this is how it works [in the code](https://github.com/xbmc/xbmc/blob/6e768251d7b9042a03598ee199660b37f298b050/xbmc/video/VideoThumbLoader.cpp#L412-L432).

markus101:
Thanks for digging in. I'm trying to think if this change could negatively affect Kodi, which I think would only happen if Sonarr failed to download the image for some reason, but Kodi was able to. Making this optional (similar to `Series Metadata URL`) would ensure we don't change the behaviour at all for Kodi and still be able to handle Jellyfin's alternate reality, which seems like a small trade off to avoid having to do it later.

src/NzbDrone.Core/Extras/Metadata/Consumers/Xbmc/XbmcMetadataSettings.cs
```diff
@@ -45,6 +45,9 @@ public XbmcMetadataSettings()
         [FieldDefinition(6, Label = "Episode Images", Type = FieldType.Checkbox, Section = MetadataSectionType.Image, HelpText = "<filename>-thumb.jpg")]
         public bool EpisodeImages { get; set; }
 
+        [FieldDefinition(7, Label = "Disable Image Thumb Tags", Type = FieldType.Checkbox, Section = MetadataSectionType.Image, HelpText = "Do not include image thumb tags in <filename>.nfo (Requires 'Episode Images')", Advanced = true)]
+        public bool NoImageThumbTags { get; set; }
```
markus101, line 49: A few things:

1. In keeping with the other options, this should be renamed to `EpisodeImageThumb` with the Label of `Episode Metadata Image Thumbs` and the default value set to true. This will be more inline with `SeriesMetadataEpisodeGuide` and `SeriesMetadataUrl`
2. Should to be moved to `4` (under `Episode Metadata`).
3. The reference to `Episode Images` should be changed to `Episode Metadata`

markus101 (changes requested):
Thanks for making it a setting, just a few changes around that

src/NzbDrone.Core/Extras/Metadata/Consumers/Xbmc/XbmcMetadataSettings.cs
JeWe37: Fixed

src/NzbDrone.Core/Extras/Metadata/Consumers/Xbmc/XbmcMetadataSettings.cs
```diff
@@ -36,13 +36,16 @@ public XbmcMetadataSettings()
         [FieldDefinition(3, Label = "Episode Metadata", Type = FieldType.Checkbox, Section = MetadataSectionType.Metadata, HelpText = "<filename>.nfo")]
         public bool EpisodeMetadata { get; set; }
 
-        [FieldDefinition(4, Label = "Series Images", Type = FieldType.Checkbox, Section = MetadataSectionType.Image, HelpText = "fanart.jpg, poster.jpg, banner.jpg")]
+        [FieldDefinition(4, Label = "Episode Metadata Image Thumbs", Type = FieldType.Checkbox, Section = MetadataSectionType.Image, HelpText = "Do not include image thumb tags in <filename>.nfo (Requires 'Episode Metadata')", Advanced = true)]
```
markus101, line 39: `Do not include` -> `Include`

markus101 (approved):
Just one more fix for the help text then it's good.

src/NzbDrone.Core/Extras/Metadata/Consumers/Xbmc/XbmcMetadataSettings.cs
JeWe37: I keep missing the small things. Fixed!

Issue #5635, opened 2023-05-10
Simkl Anime List integration


funforgiven:
### Is there an existing issue for this?

- [X] I have searched the existing open and closed issues

### Is your feature request related to a problem? Please describe

I believe that adding support for Simkl was a great idea. However, Simkl separates Anime and TV shows so it would be beneficial to be able to import Anime list from Simkl as well, as Sonarr already supports anime.

### Describe the solution you'd like

One potential solution could be to provide an option within the Import Lists feature, specifically in the Simkl User Watchlist section, to allow users to selectively import their desired media list, whether it be for TV Shows or Anime. Given the apparent similarity between the APIs of TV Shows and Anime, this implementation should prove to be relatively straightforward.

https://simkl.com/
https://simkl.docs.apiary.io/#

### Describe alternatives you've considered

N/A

#### Issue and pull request for Simkl List integration for TV Shows:
#5306
#5313

thezoggy:
sonarr uses tvdb, which is why simkl list support for 'tv' was added (as they use tvdb). for sonarr to add 'anime' list would require sonarr to add anidb support which it does not do.

funforgiven:
To my understanding, they also offer tvdb support for anime.

I have observed [an example in their API documentation](https://simkl.docs.apiary.io/#introduction/about-simkl-api/standard-media-objects), where they provided a tvdb id for an anime.

This is the example in their documentation:

```
{
    "title": "Attack on Titan",
    "year": 2013,
    "ids": {
        "simkl": 39687,
        "mal": 16498,
        "tvdb": 267440,
        "imdb": "tt2560140",
        "anidb": 9541
    },
    "seasons":[ 
        {
            "number": 2
        }
    ]
}
```

Are there some animes that do not provide tvdb ids?

ShubhamPatilsd:
I would like to reserve this issue for our students over at [CodeDay Labs](https://labs.codeday.org/) during our summer session.

A bit about CodeDay Labs:

> CodeDay Labs provides “micro-internships” for college students. This means working on real issues on real-world projects. The best part about this is that we’re helping the open source community at the same time by providing opportunities for these students to work on open-source projects.

jtnguyen45:
I am working on this from CodeDay Labs.

PR #5766, opened 2023-06-29
Simkl Anime List integration

jtnguyen45:
#### Database Migration
NO

#### Description
This is our progress with issue 5635 so far. We added a button when going to "Settings→Import Lists→Add List→Simkl List" that allows the user to add an anime list where the button is called "Simkl Anime Watchlist". We also renamed the other button to "Simkl Shows Watchlist". 

When creating the new list for Simkl Anime Watchlist, we are able to create it in the same manner as Simkl Shows Watchlist. On the initial import, the code was able to import most of the anime shows into the Series tab with no issue. For a couple anime shows, it would import the wrong show. Some anime shows don't have a IMDB id when fetched from the Simkl API (it also does not list an IMDB link on Simkl's website). The issue is happening in ImportListServices.cs on line 99 where it is trying to search for the show on the sonarr skyhook website, and it chooses the first entry by default, which may lead to the wrong output.

Example: adding the anime show "Kono Subarashii Sekai ni Bakuen o!" returns the wrong show because when this line, 
`var mappedSeries = _seriesSearchService.SearchForNewSeries(report.Title).FirstOrDefault()`, 
is executed, it chooses the first entry from the http request from the skyhook sonarr website, which is incorrect. 

This is the [link](https://skyhook.sonarr.tv/v1/tvdb/search/en/?term=kono%20subarashii%20sekai%20ni%20bakuen%20o%21), that was used when trying to fetch "Kono Subarashii Sekai ni Bakuen o!". 

* https://github.com/Sonarr/Sonarr/issues/5635

### DIFF
--- a/src/NzbDrone.Core/ImportLists/Simkl/SimklAPI.cs
+++ b/src/NzbDrone.Core/ImportLists/Simkl/SimklAPI.cs
@@ -11,6 +11,7 @@ namespace NzbDrone.Core.ImportLists.Simkl
         public string Imdb { get; set; }
         public string Tmdb { get; set; }
         public string Tvdb { get; set; }
+        public string Mal { get; set; }
     }
 
     public class SimklSeriesPropsResource
@@ -23,11 +24,15 @@ namespace NzbDrone.Core.ImportLists.Simkl
     public class SimklSeriesResource
     {
         public SimklSeriesPropsResource Show { get; set; }
+
+        [JsonProperty("anime_type")]
+        public SimklAnimeType AnimeType { get; set; }
     }
 
     public class SimklResponse
     {
         public List<SimklSeriesResource> Shows { get; set; }
+        public List<SimklSeriesResource> Anime { get; set; }
     }
 
     public class RefreshRequestResponse
@@ -65,5 +70,11 @@ namespace NzbDrone.Core.ImportLists.Simkl
     public class SimklTvSyncActivityResource
     {
         public DateTime All { get; set; }
+    }
+
+    public enum SimklAnimeType
+    {
+        Tv,
+        Movie
     }
 }
--- a/src/NzbDrone.Core/ImportLists/Simkl/SimklParser.cs
+++ b/src/NzbDrone.Core/ImportLists/Simkl/SimklParser.cs
@@ -1,6 +1,8 @@
 using System.Collections.Generic;
 using System.Net;
+using NLog;
 using NzbDrone.Common.Extensions;
+using NzbDrone.Common.Instrumentation;
 using NzbDrone.Common.Serializer;
 using NzbDrone.Core.ImportLists.Exceptions;
 using NzbDrone.Core.Parser.Model;
@@ -10,6 +12,11 @@ namespace NzbDrone.Core.ImportLists.Simkl
     public class SimklParser : IParseImportListResponse
     {
         private ImportListResponse _importResponse;
+        private static readonly Logger Logger = NzbDroneLogger.GetLogger(typeof(SimklParser));
+
+        public SimklParser()
+        {
+        }
 
         public virtual IList<ImportListItemInfo> ParseResponse(ImportListResponse importResponse)
         {
@@ -22,7 +29,7 @@ namespace NzbDrone.Core.ImportLists.Simkl
                 return series;
             }
 
-            var jsonResponse = STJson.Deserialize<SimklResponse>(_importResponse.Content);
+            var jsonResponse = Json.Deserialize<SimklResponse>(_importResponse.Content);
 
             // no shows were return
             if (jsonResponse == null)
@@ -30,14 +37,40 @@ namespace NzbDrone.Core.ImportLists.Simkl
                 return series;
             }
 
-            foreach (var show in jsonResponse.Shows)
+            if (jsonResponse.Anime != null)
             {
-                series.AddIfNotNull(new ImportListItemInfo()
+                foreach (var show in jsonResponse.Anime)
                 {
-                    Title = show.Show.Title,
-                    TvdbId = int.TryParse(show.Show.Ids.Tvdb, out var tvdbId) ? tvdbId : 0,
-                    ImdbId = show.Show.Ids.Imdb
-                });
+                    var tentativeTvdbId = int.TryParse(show.Show.Ids.Tvdb, out var tvdbId) ? tvdbId : 0;
+
+                    if (tentativeTvdbId > 0 && show.AnimeType == SimklAnimeType.Tv)
+                    {
+                        series.AddIfNotNull(new ImportListItemInfo()
+                        {
+                            Title = show.Show.Title,
+                            ImdbId = show.Show.Ids.Imdb,
+                            TvdbId = tvdbId,
+                            MalId = int.TryParse(show.Show.Ids.Mal, out var malId) ? malId : 0
+                        });
+                    }
+                    else
+                    {
+                        Logger.Warn("Skipping info grabbing for '{0}' because it is a movie or it is not the first season of the show", show.Show.Title);
+                    }
+                }
+            }
+
+            if (jsonResponse.Shows != null)
+            {
+                foreach (var show in jsonResponse.Shows)
+                {
+                    series.AddIfNotNull(new ImportListItemInfo()
+                    {
+                        Title = show.Show.Title,
+                        TvdbId = int.TryParse(show.Show.Ids.Tvdb, out var tvdbId) ? tvdbId : 0,
+                        ImdbId = show.Show.Ids.Imdb
+                    });
+                }
             }
 
             return series;
--- a/src/NzbDrone.Core/ImportLists/Simkl/User/SimklUserRequestGenerator.cs
+++ b/src/NzbDrone.Core/ImportLists/Simkl/User/SimklUserRequestGenerator.cs
@@ -21,7 +21,7 @@ namespace NzbDrone.Core.ImportLists.Simkl.User
 
         private IEnumerable<ImportListRequest> GetSeriesRequest()
         {
-            var link = $"{Settings.BaseUrl.Trim()}/sync/all-items/shows/{((SimklUserListType)Settings.ListType).ToString().ToLowerInvariant()}";
+            var link = $"{Settings.BaseUrl.Trim()}/sync/all-items/{((SimklUserShowType)Settings.ShowType).ToString().ToLowerInvariant()}/{((SimklUserListType)Settings.ListType).ToString().ToLowerInvariant()}";
 
             var request = new ImportListRequest(link, HttpAccept.Json);
 
--- a/src/NzbDrone.Core/ImportLists/Simkl/User/SimklUserSettings.cs
+++ b/src/NzbDrone.Core/ImportLists/Simkl/User/SimklUserSettings.cs
@@ -19,9 +19,13 @@ namespace NzbDrone.Core.ImportLists.Simkl.User
         public SimklUserSettings()
         {
             ListType = (int)SimklUserListType.Watching;
+            ShowType = (int)SimklUserShowType.Shows;
         }
 
         [FieldDefinition(1, Label = "List Type", Type = FieldType.Select, SelectOptions = typeof(SimklUserListType), HelpText = "Type of list you're seeking to import from")]
         public int ListType { get; set; }
+
+        [FieldDefinition(1, Label = "Show Type", Type = FieldType.Select, SelectOptions = typeof(SimklUserShowType), HelpText = "Type of show you're seeking to import from")]
+        public int ShowType { get; set; }
     }
 }
--- a/src/NzbDrone.Core/Parser/Model/ImportListItemInfo.cs
+++ b/src/NzbDrone.Core/Parser/Model/ImportListItemInfo.cs
@@ -11,6 +11,7 @@ namespace NzbDrone.Core.Parser.Model
         public int TvdbId { get; set; }
         public int TmdbId { get; set; }
         public string ImdbId { get; set; }
+        public int MalId { get; set; }
         public DateTime ReleaseDate { get; set; }
 
         public override string ToString()
--- a/src/NzbDrone.Core/ImportLists/Simkl/User/SimklUserShowType.cs
+++ b/src/NzbDrone.Core/ImportLists/Simkl/User/SimklUserShowType.cs
@@ -1,0 +1,8 @@
+namespace NzbDrone.Core.ImportLists.Simkl.User
+{
+    public enum SimklUserShowType
+    {
+        Shows = 0,
+        Anime = 1
+    }
+}


src/NzbDrone.Core/ImportLists/Simkl/User/SimklUserImport.cs
```diff
@@ -17,7 +17,7 @@ public class SimklUserImport : SimklImportBase<SimklUserSettings>
         {
         }
 
-        public override string Name => "Simkl User Watchlist";
```
markus101, line 20: This isn't needed if this is DRYed up.

markus101 (changes requested):
I think this could be DRYed out by adding a new enum setting to `SimklUserSettings` (with a default value to continue fetching shows). This would reuse the whole fetching/processing that already exists for Simkl user lists and in the request generator it would decide whether it'd use `shows` or `anime` int he route.

With this all the files in `src/NzbDrone.Core/ImportLists/Simkl/Anime/` would be removed, a new setting would need to be added and the request in `SimklUserRequestGenerator` would need to chose the right path based on the new setting.

markus101:
The issue with Skyhook not matching the correct title is going to be a problem, but unless an ID is known or TheTVDB has additional information Skyhook can use I don't think there is anything else we can really do.

funforgiven:
The issue with the Skyhook is that TVDB uses English titles for animes but Simkl main titles are usually in romaji. Fortunately, Simkl also provides English titles (at least it does for the animes I have checked).

Using the English title provided by Simkl just under the original name, your example actually [provides the correct entry as first result.](https://skyhook.sonarr.tv/v1/tvdb/search/en/?term=Konosuba:%20An%20Explosion%20on%20this%20Wonderful%20World!)

jtnguyen45:
> I think this could be DRYed out by adding a new enum setting to `SimklUserSettings` (with a default value to continue fetching shows). This would reuse the whole fetching/processing that already exists for Simkl user lists and in the request generator it would decide whether it'd use `shows` or `anime` int he route.
> 
> With this all the files in `src/NzbDrone.Core/ImportLists/Simkl/Anime/` would be removed, a new setting would need to be added and the request in `SimklUserRequestGenerator` would need to chose the right path based on the new setting.

Just implemented the changes so we could DRY it out. It looks way cleaner, thank you for your suggestion.

src/NzbDrone.Core/ImportLists/Simkl/User/SimklUserImport.cs
jtnguyen45: If this line of code is removed, how is the button implemented? Currently we have two buttons: "Simkl User Watchlist" and "Simkl Anime WatchList". Not sure where to go from there.

markus101: The renaming isn't necessary as setting the list type will be handled by a setting in `SimklUserSettings`

src/NzbDrone.Core/ImportLists/Simkl/User/SimklUserRequestGenerator.cs
```diff
@@ -21,7 +23,19 @@ public virtual ImportListPageableRequestChain GetListItems()
 
         private IEnumerable<ImportListRequest> GetSeriesRequest()
         {
-            var link = $"{Settings.BaseUrl.Trim()}/sync/all-items/shows/{((SimklUserListType)Settings.ListType).ToString().ToLowerInvariant()}";
+            var link = "";
+
+            switch (ShowType)
```
markus101, line 28: Instead of being set on the generator this should come from settings, similar to `Settings.ListType`

src/NzbDrone.Core/ImportLists/Simkl/SimklAPI.cs
```diff
@@ -27,7 +28,8 @@ public class SimklSeriesResource
 
     public class SimklResponse
     {
-        public List<SimklSeriesResource> Shows { get; set; }
+        public List<SimklSeriesResource> Shows { get; set; }  // get { return shows } set {this.shows = shows}
```
markus101, line 31: Looks like `// get { return shows } set {this.shows = shows}` is unneeded,

src/NzbDrone.Core/ImportLists/Simkl/User/SimklUserSettings.cs
```diff
@@ -24,4 +24,10 @@ public SimklUserSettings()
         [FieldDefinition(1, Label = "List Type", Type = FieldType.Select, SelectOptions = typeof(SimklUserListType), HelpText = "Type of list you're seeking to import from")]
         public int ListType { get; set; }
     }
+
+    public enum SimklShowType
```
markus101, line 28: A new `FieldDefinition` needs to be added for this in `SimklUserSettings`

markus101 (commented):
Thanks for the clean up. I've left some additional comments  on how to utilize the setting.

markus101:
> While working on this issue for a couple of hours, we have tested the animeplanet ID trick, and it seems to have worked. However, some animes that have an IMDB still get screwed up. For example, when trying to import [Why Raeliana Ended Up at the Duke's Mansion](https://www.imdb.com/title/tt27484853/), skyhook returns [this](https://skyhook.sonarr.tv/v1/tvdb/search/en/?term=imdb%3Att27484853), which is completely incorrect. I don't know why this is happening, but it is.

This looks to be because there isn't a series on TheTVDB with that IMDB ID, but instead of returning an empty array of results it searches by title. I'll look into that issue on the Skyhook side.

Unfortunately I don't think MAL or Anidb ID's are available on TheTVDB.

funforgiven:
> Unfortunately I don't think MAL or AniDB ID's are available on TheTVDB.

They are not available on TheTVDB but there is a [frequently updated mapping information between AniDB, TheTVDB, TMDB and IMDB](https://github.com/Anime-Lists/anime-lists) which could be used for AniDB ids. Notably, significant projects depend on it such as HAMA agent and Overseer/JellySeer.

funforgiven:
> Yes, Simkl does have an English title, but the API request only gives the title in romaji. Fortunately, the animeplanet ID that simkl gives back is in english.

I would suggest using crunchyroll id, because animeplanet is not really consistent with their naming and would also cause problems with sequels. 

Some examples:

isekai-shoukan-wa-nidome-desu (animeplanet) 
summoned-to-another-world-for-a-second-time (crunchyroll)

kaminaki-sekai-no-kamisama-katsudou (animeplanet)
kamikatsu-working-for-god-in-a-godless-world (crunchyroll)

hells-paradise-jigokuraku (animeplanet)
hells-paradise (crunchyroll)

While I realize that these examples already have IMDB ids, I just wanted to point out the inconsistent naming of animeplanet.

I also want to point out that Simkl separates seasons to different entries like many anime sites. However, TheTVDB, IMDB and TMDB combines all seasons in a single entry. Therefore, animes that does not have IMDB or TMDB ids are mostly sequels so they are added with first season already.

Only entry I could find that does not have an IMDB id is Kono Subarashii Sekai ni Bakuen o! which could be solved by using crunchyroll id.

Using animeplanet would create problems with sequels. I can give an example:

Shingeki no Kyojin Season 3 entry in Simkl does not have an IMDB or TMDB id because it is a sequel. If you use animeplanet id, which is attack-on-titan-3rd-season-part-ii, you would get the wrong entry from Skyhook which is ["Attack on Titan: Junior High"](https://skyhook.sonarr.tv/v1/tvdb/search/en/?term=attack-on-titan-3rd-season-part-ii)

funforgiven:
> Sadly, not all animes have a Crunchyroll ID.

Unfortunately, yes. However, as I presented in my example, there are problems when using animeplanet ids with sequels.

 I think the best solution would be to check for IMDB id first as you already do. If it does not exist, it is almost always a sequel. However, there may be some weird exceptions like Kono Subarashii Sekai ni Bakuen o!. That is when crunchyroll id comes in because it does not exist on sequels so we can be sure that it is not a sequel and still be able to import it. There may be still missed entries but I don't think we should try to solve them using animeplanet ids because they create more problems than they solve. I believe Simkl will eventually add IMDB ids if it not a sequel.

funforgiven:
> The only one big solution that will solve all of this is if Simkl added a TVDB ID in their API requests for anime.

It is weird that they don't have TheTVDB ids because they have it in their examples. Instead, they have TVDB slug, which does not work with Skyhook as far as I know. Anyway, that also would not solve the current problem because they would only add TVDB ids to first seasons of animes just like IMDB ids. However, it is actually enough for Sonarr since it would import the whole series with all its seasons from just the first season Simkl entry.

funforgiven:
Hello @iceypotato, Simkl developers are willing to help with the implementation. They have added TVDB ids for missing non sequel animes and can also provide tvdb ids on their /sync/ end point for your and Sonarr's client ids. If you [join the Simkl Discord](https://discord.gg/u89XfYn), and write to the dev-discussion channel, we can communicate further.

jtnguyen45:
We were able to fix the issue with some of the anime imports not loading correctly. We discussed the issue with the Simkl developers and they fixed the TVDB issue on their end. With how it is set up on Simkl, users are only able to import the anime show if they choose the first season of the show (which is fine as doing so will load the subsequent seasons with it). The reason is Simkl's API doesn't have a TVDB id if you pick the second season of that anime. We have made it so Sonarr will not try to import the anime in that case, so that wrong shows are not imported instead, and also will not try to import anime movies.

Thanks again @funforgiven for helping us on the TVDB issue and connecting us with the Simkl developers!

src/NzbDrone.Core/ImportLists/Simkl/User/SimklUserRequestGenerator.cs
```diff
@@ -21,7 +23,7 @@ public virtual ImportListPageableRequestChain GetListItems()
 
         private IEnumerable<ImportListRequest> GetSeriesRequest()
         {
-            var link = $"{Settings.BaseUrl.Trim()}/sync/all-items/shows/{((SimklUserListType)Settings.ListType).ToString().ToLowerInvariant()}";
+            var link = $"{Settings.BaseUrl.Trim()}/sync/all-items/{ShowType}/{((SimklUserListType)Settings.ListType).ToString().ToLowerInvariant()}";
```
markus101, line 26: When coming from settings this will need to convert the enum to a string, but a ternary can handle that while there are only two values.

Something like:
`var showType = Settings.ShowType == (int)SimklShowType.Anime ? 'anime' : 'shows'`

src/NzbDrone.Core/ImportLists/Simkl/User/SimklUserSettings.cs
```diff
@@ -19,9 +19,13 @@ public class SimklUserSettings : SimklSettingsBase<SimklUserSettings>
...
         {
             ListType = (int)SimklUserListType.Watching;
+
+            // ShowType = "";
         }
 
         [FieldDefinition(1, Label = "List Type", Type = FieldType.Select, SelectOptions = typeof(SimklUserListType), HelpText = "Type of list you're seeking to import from")]
         public int ListType { get; set; }
+
+        // public string ShowType { get; set; }
```
markus101, line 29: This is heading in the correct direction, but needs to be an enum and needs the `FieldDefinition` treatment

src/NzbDrone.Core/Parser/Model/ImportListItemInfo.cs
```diff
@@ -7,10 +7,13 @@ public class ImportListItemInfo
         public int ImportListId { get; set; }
         public string ImportList { get; set; }
         public string Title { get; set; }
+
```
markus101, line 10: No need for a new line

src/NzbDrone.Core/ImportLists/Simkl/SimklParser.cs
```diff
@@ -30,14 +30,40 @@ public virtual IList<ImportListItemInfo> ParseResponse(ImportListResponse import
...
+                        {
+                            Title = show.Show.Title,
+                            ImdbId = show.Show.Ids.Imdb,
+                            TvdbId = tentativeTVDBid,
+                            MalId = int.TryParse(show.Show.Ids.Mal, out var malId) ? malId : 0,
+                        });
+                    }
+                    else
+                    {
+                        System.Console.WriteLine("Skipping import for " + show.Show.Title + " because it is not the first season of the show or is a movie.");
```
markus101, line 51: This will need to use the logger, you can do construction injection like `SimklUserImport` does to inject the Logger: `Logger logger`

src/NzbDrone.Core/ImportLists/Simkl/SimklParser.cs
```diff
@@ -30,14 +30,40 @@ public virtual IList<ImportListItemInfo> ParseResponse(ImportListResponse import
...
+            if (jsonResponse.Anime != null)
             {
-                series.AddIfNotNull(new ImportListItemInfo()
+                foreach (var show in jsonResponse.Anime)
                 {
-                    Title = show.Show.Title,
-                    TvdbId = int.TryParse(show.Show.Ids.Tvdb, out var tvdbId) ? tvdbId : 0,
-                    ImdbId = show.Show.Ids.Imdb
-                });
+                    var tentativeTVDBid = int.TryParse(show.Show.Ids.Tvdb, out var tvdbId) ? tvdbId : 0;
```
markus101, line 37: `tentativeTvdbId` is more consistent with other usages of `TvdbId` in variables

src/NzbDrone.Core/ImportLists/Simkl/SimklParser.cs
```diff
@@ -30,14 +30,40 @@ public virtual IList<ImportListItemInfo> ParseResponse(ImportListResponse import
...
+
+                    if (tentativeTVDBid != 0 && string.Equals("tv", show.Anime_type))
```
markus101, line 39: `>= 0` just incase a negative number is returned one day by mistake.

markus101: Also as mentioned elsewhere can we avoid the magic string comparison?

src/NzbDrone.Core/ImportLists/Simkl/SimklParser.cs
markus101: If the anime is not the first season of a series in TheTVDB is it possible to add that anime if the user doesn't have the first season already?

src/NzbDrone.Core/ImportLists/Simkl/User/SimklUserImport.cs
```diff
@@ -17,14 +23,40 @@ public class SimklUserImport : SimklImportBase<SimklUserSettings>
...
+            return new SimklUserRequestGenerator()
+            {
+                Settings = Settings,
+                ClientId = ClientId,
+                ShowType = SimklShowTypes.Shows
+            };
+        }
+    }
+
+    public class SimklAnimeImport : SimklImportBase<SimklUserSettings>
```
markus101, line 39: A separate class for anime isn't needed, as mentioned this can be done via a setting on `SimklUserSettings` instead.

markus101 (changes requested):
I've left some suggestions on improving names and removing the need to have a separate import list type.

src/NzbDrone.Core/ImportLists/Simkl/SimklParser.cs
funforgiven: > If the anime is not the first season of a series in TheTVDB is it possible to add that anime if the user doesn't have the first season already?

I think it is not possible currently. Simkl API does not provide ids of prequels. However, they are planning on adding TVDB ids for the sequels in the future.

src/NzbDrone.Core/ImportLists/Simkl/User/SimklUserImport.cs
markus101: A dropdown for selecting anime or shows instead of the separate button. You should be able to use `ListType` on `SimklUserSettings` as a guide on how it should be implemented.

markus101: In the case of other lists the lists are quite different, in this case it's just a different type of series, not too different from a different list type as it's all coming from the same source.

markus101 (changes requested):
Thanks for moving to the setting! I hope my comment on the deserialization helps and I left a few more small comments.

src/NzbDrone.Core/ImportLists/Simkl/SimklAPI.cs
```diff
@@ -66,4 +73,10 @@ public class SimklTvSyncActivityResource
     {
         public DateTime All { get; set; }
     }
+
+    public class AnimeType
```
markus101, line 77: I wonder is this is could be conflicting with the `AnimeType` on `SimklSeriesResource`, could try renaming one of them. This should also be converted to an enum.

markus101: Okay, figured it out, the wrong JSON deserialization is being used in `SimklParser`,

```
var jsonResponse = STJson.Deserialize<SimklResponse>(_importResponse.Content);
```

should be 

```
var jsonResponse = Json.Deserialize<SimklResponse>(_importResponse.Content);
```

src/NzbDrone.Core/ImportLists/Simkl/SimklAPI.cs
```diff
@@ -23,11 +24,17 @@ public class SimklSeriesPropsResource
     public class SimklSeriesResource
     {
         public SimklSeriesPropsResource Show { get; set; }
+
+        [JsonProperty(PropertyName = "anime_type")]
```
markus101, line 28: For consistency with the other uses in this file you can change to `[JsonProperty("anime_type")]`

markus101:
This PR will need to be rebased on `develop` currently it's merging develop into this PR's branch. `git rebase origin/develop` should do it.

src/NzbDrone.Core/ImportLists/Simkl/SimklAPI.cs
jtnguyen45: This did the trick and AnimeType works just fine now. Thanks for looking into it for us!

src/NzbDrone.Core/ImportLists/Simkl/SimklAPI.cs
```diff
@@ -23,11 +24,13 @@ public class SimklSeriesPropsResource
     public class SimklSeriesResource
     {
         public SimklSeriesPropsResource Show { get; set; }
+        public string Anime_type { get; set; }
```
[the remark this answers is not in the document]
jtnguyen45: Was able to fix using your suggestion in another comment.

jtnguyen45:
Implemented the changes above and our pull request is ready for review. For now, until Simkl API implements TVDB ids in their subsequent anime seasons (any season after the first one), this feature is only able to import anime shows if the user chooses the first season. Once Simkl API implements this, this feature will need to be looked at again in the future.

markus101 (changes requested):
Just a few minor things and this needs a rebase on `develop` to remove the merge commit.

src/NzbDrone.Core/ImportLists/Simkl/SimklParser.cs
```diff
@@ -22,22 +29,48 @@ public virtual IList<ImportListItemInfo> ParseResponse(ImportListResponse import
...
+                        {
+                            Title = show.Show.Title,
+                            ImdbId = show.Show.Ids.Imdb,
+                            TvdbId = tvdbId,
+                            MalId = int.TryParse(show.Show.Ids.Mal, out var malId) ? malId : 0
+                        });
+                    }
+                    else
+                    {
+                        Logger.Warn("Skipping info getting on \"" + show.Show.Title + "\" because it is a movie or it is not the first season of the show (Simkl does not currently support TVDB IDs for >= 2nd seasons).");
```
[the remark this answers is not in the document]
Thirrian: "Skip getting info on..." ?

src/NzbDrone.Core/ImportLists/Simkl/SimklAPI.cs
```diff
@@ -23,11 +24,15 @@ public class SimklSeriesPropsResource
     public class SimklSeriesResource
     {
         public SimklSeriesPropsResource Show { get; set; }
+
+        [JsonProperty("anime_type")]
+        public string AnimeType { get; set; }
```
markus101, line 29: This should be `SimklAnimeType` instead of `string`

src/NzbDrone.Core/ImportLists/Simkl/SimklParser.cs
```diff
@@ -22,22 +29,48 @@ public virtual IList<ImportListItemInfo> ParseResponse(ImportListResponse import
...
             }
 
-            foreach (var show in jsonResponse.Shows)
+            if (jsonResponse.Anime != null)
+            {
+                foreach (var show in jsonResponse.Anime)
+                {
+                    var tentativeTvdbId = int.TryParse(show.Show.Ids.Tvdb, out var tvdbId) ? tvdbId : 0;
+
+                    if (tentativeTvdbId > 0 && show.AnimeType.Equals(SimklAnimeType.Tv.ToString().ToLowerInvariant()))
```
markus101, line 46: Along with changing the prop to the correct enum:

```suggestion
                    if (tentativeTvdbId > 0 && show.AnimeType == SimklAnimeType.Tv)
```

markus101 (changes requested):
A couple things to clean up the enum usage and then this needs to be squashed down and rebased on `develop` to get rid of all the merge commits

markus101 (approved):
Looks great, thanks for iterating on it!

Issue #5769, opened 2023-07-01
New: Add CF and Score option for Discord on grab notifications


jack-mil:
### Is there an existing issue for this?

- [X] I have searched the existing open and closed issues

### Is your feature request related to a problem? Please describe

New in v4, support custom formats and custom format score fields into Discord webhook notification template.
This would require implementing code similar to Radarr.
https://github.com/Radarr/Radarr/blob/690bab3264f897e101a79bf881f8fd8ee3c746cf/src/NzbDrone.Core/Notifications/Discord/Discord.cs#L102C1-L109C31

### Describe the solution you'd like

Similar to how Radarr has implemented this feature.
![image](https://github.com/Sonarr/Sonarr/assets/62065280/538ffa0e-4133-4607-9c71-5bab94c68cd1)

### Describe alternatives you've considered

Could be accomplished similarily to #5493 

### Anything else?

I will attempt to make implement this change. The code is easy, but setting up the build environment will be more complex.
Is there a supported method for building and testing in Docker?
I use Linux and do not use Windows/Visual studio code,

readstead:
Please also add the release group to the notifications (or an option to). I need to know what all the upgrades are about within the same quality profile.

jack-mil:
There is currently the "Release" field that can be included in Discord, which is the full release title. 

I agree that the parsed release group would be a good addition.

I haven't used any of the other notification types, so I don't know what is available there. But it would be best if they all matched for sure.

PR #5782, opened 2023-07-06
New: Add CustomFormat field types to Discord OnGrab Notification

jack-mil:
Similar to Radarr, adds the ability to include parsed custom formats and CF score in the Discord webhook notification.
Are there even more options that could be included here?
Database Migration

NO
Description

See https://github.com/Sonarr/Sonarr/issues/5769.
Similar in nature to PR https://github.com/Sonarr/Sonarr/pull/5493
Todos

Tests (How?)

    Wiki Updates (None, except all the possible field types listed somewhere would be nice. Might open a PR for the wiki at some point)

Issues Fixed or Closed by this PR

Closes https://github.com/Sonarr/Sonarr/issues/5769

### DIFF
--- a/src/NzbDrone.Core/Notifications/Discord/DiscordSettings.cs
+++ b/src/NzbDrone.Core/Notifications/Discord/DiscordSettings.cs
@@ -19,7 +19,7 @@ namespace NzbDrone.Core.Notifications.Discord
         public DiscordSettings()
         {
             // Set Default Fields
-            GrabFields = new[] { 0, 1, 2, 3, 5, 6, 7, 8, 9 };
+            GrabFields = new[] { 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12 };
             ImportFields = new[] { 0, 1, 2, 3, 4, 6, 7, 8, 9, 10, 11, 12 };
             ManualInteractionFields = new[] { 0, 1, 2, 3, 5, 6, 7, 8, 9 };
         }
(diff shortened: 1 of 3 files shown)


src/NzbDrone.Core/Notifications/Discord/DiscordSettings.cs
```diff
@@ -19,7 +19,7 @@ public class DiscordSettings : IProviderConfig
         public DiscordSettings()
         {
             // Set Default Fields
-            GrabFields = new[] { 0, 1, 2, 3, 5, 6, 7, 8, 9 };
```
jack-mil, line 22: I am not a C# expert, but it seems odd to store the default field types as an int[] and constantly cast to the `DiscordGrabFieldType` enum. Is this done for some reason I don't understand with the frontend?

augustuen:
> Is there a specific reason that the defaults are hardcoded instead of automatically including everything? I could tryto clean up the code a little

Since the fields are an advanced setting it is reasonable to expect that a good portion of users will not know to change them. Therefore it is a good idea to only include the most likely fields they will want. 

And while we are on it, is there a reason you didnt include `indexer` in the GrabField defaults?

src/NzbDrone.Core/Notifications/Discord/DiscordSettings.cs
markus101: I believe this is mostly a limitation of our dynamic handling of settings when serialized for the API, not really a burden to cast as necessary since it makes the conversion via the API painless.

jack-mil:
I have mentioned the potential additions on the Wiki, https://github.com/Servarr/Wiki/commit/f1e5f80aee9fb2239406ea222bcb6f399d44e70e

markus101 (approved):

PR #5785, opened 2023-07-07
Translate Sidebar Strings

stevietv:
#### Database Migration
NO

#### Description
My first attempt at bringing in some more translation strings to Sonarr. Started with just the sidebar.
Let me know if this is the right approach or if I should be doing something else and I'm happy to work on more.

#### Todos
- [ ] Tests
- [ ] Wiki Updates

*


frontend/src/Components/Page/Sidebar/PageSidebar.js
```diff
@@ -25,11 +26,11 @@ const links = [
     alias: '/series',
     children: [
       {
-        title: 'Add New',
+        title: 'AddNew',
```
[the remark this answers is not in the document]
stevietv: That was my first approach. I tried it like this and it didn't work: 

for example, using this:
image

and reverting back to `title={child.title}` on line 499 results in no translation:

image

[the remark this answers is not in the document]
stevietv: thanks @mynameisbogdan - wouldn't have got that without you!

frontend/src/Utilities/String/translate.js
```diff
@@ -1,18 +1,22 @@
 import createAjaxRequest from 'Utilities/createAjaxRequest';
 
 function getTranslations() {
-  return createAjaxRequest({
-    global: false,
+  let localization = null;
+  const ajaxOptions = {
+    async: false,
```
markus101, line 6: `async: false` should be avoided, if we need to wait for the translations to load then we need to wait on a promise during initialization instead of forcing the browser to make the request synchonously.

markus101 (changes requested):
The translations look good, but the changes to the initialization of the strings needs to avoid the synchronous call.

frontend/src/Utilities/String/translate.js
Qstick: The promise will be needed. As doing this purely async with large number of translations can result in weird half translated scenarios in the UI (experienced on Radarr)

stevietv: ok, then I'll leave it as I have submitted?

markus101: I'm going to change it a different way, I have a local change for it started so you can revert.

stevietv: done

src/NzbDrone.Core/Localization/Core/en.json
```diff
@@ -141,17 +164,24 @@
...
+  "Tasks": "Tasks",
+  "UI": "UI",
   "UI Language": "UI Language",
   "Unmonitored": "Unmonitored",
+  "Updates":  "Updates",
   "UpdateAvailableHealthCheckMessage": "New update is available",
   "UpdateStartupNotWritableHealthCheckMessage": "Cannot install update because startup folder '{0}' is not writable by the user '{1}'.",
   "UpdateStartupTranslocationHealthCheckMessage": "Cannot install update because startup folder '{0}' is in an App Translocation folder.",
   "UpdateUINotWritableHealthCheckMessage": "Cannot install update because UI folder '{0}' is not writable by the user '{1}'.",
+  "Wanted": "Wanted"
```
markus101, line 185: I think the missing comma at the end is causing the test failures.

markus101 (approved):
Just the one JSON issue then looks good, thanks!

src/NzbDrone.Core/Localization/Core/en.json
stevietv: > I think the missing comma at the end is causing the test failures.

oops :D

stevietv:
@markus101 commit 3fa571c3 doesn't fix the translations not being loaded

image

markus101:
The translation issue should be fixed, but needs a rebase before landing.

stevietv:
> The translation issue should be fixed, but needs a rebase before landing.

done :)

PR #5786, opened 2023-07-07
Fixed: Minor improvements and translations for managing bulk indexers, lists and clients

mynameisbogdan:
#### Database Migration
NO


src/Sonarr.Api.V3/ProviderBulkResource.cs
```diff
@@ -9,6 +8,18 @@ public class ProviderBulkResource<T>
...
         public List<int> Tags { get; set; }
         public ApplyTags ApplyTags { get; set; }
+
+        public ProviderBulkResource()
+        {
+            Ids = new List<int>();
+        }
+    }
+
+    public enum ApplyTags
```
markus101, line 18: I think moving this is the right idea, but it shouldn't be part of this resource, it can have it's own file instead.

markus101 (changes requested):
Please separate the "minor fixes" into their own commit and elaborate on the purpose instead of combining everything into one.

markus101:
Landed these after some adjustments locally.

PR #5792, opened 2023-07-11
Fixed: (BTN) Prevent NullRef when Result.Torrents is null

mynameisbogdan:
#### Database Migration
NO

#### Description
Fix for a NullRef described in https://www.reddit.com/r/sonarr/comments/14v1neb/bug_report_nullreferenceexception_in/

### DIFF
--- a/src/NzbDrone.Core/Indexers/BroadcastheNet/BroadcastheNetParser.cs
+++ b/src/NzbDrone.Core/Indexers/BroadcastheNet/BroadcastheNetParser.cs
@@ -50,7 +50,7 @@ namespace NzbDrone.Core.Indexers.BroadcastheNet
                 throw new IndexerException(indexerResponse, "Indexer API call returned an error [{0}]", jsonResponse.Error);
             }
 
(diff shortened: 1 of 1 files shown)


src/NzbDrone.Core/Indexers/BroadcastheNet/BroadcastheNetParser.cs
```diff
@@ -50,7 +50,7 @@ public IList<ReleaseInfo> ParseResponse(IndexerResponse indexerResponse)
                 throw new IndexerException(indexerResponse, "Indexer API call returned an error [{0}]", jsonResponse.Error);
             }
 
-            if (jsonResponse.Result.Results == 0)
+            if (jsonResponse.Result.Results == 0 || jsonResponse.Result?.Torrents?.Values == null)
```
markus101, line 53: `jsonResponse.Result` will always have a value if it hits this condition since `jsonResponse.Result.Results` was called first

markus101 (changes requested):

markus101 (approved):

PR #5799, opened 2023-07-12
New: Show CF Score in episode row

jack-mil:
#### Database Migration
NO

#### Description
This commit adds the Custom Format column in the series episode row view. This is to be consistent with fields available in 
History, Queue, Search, etc.. This is something I wanted for my personal use at least

image

Disclaimer: I am not a React Developer and am hacking this together from existing code. If there is a better way to integrate all table views with consistent column view options, I am completely open to redesigning.

#### Todos
- [ ] Tests (None?)
- [x] Wiki Updates (N/A)

*

### DIFF
--- a/frontend/src/Series/Details/EpisodeRow.js
+++ b/frontend/src/Series/Details/EpisodeRow.js
@@ -4,6 +4,7 @@ import MonitorToggleButton from 'Components/MonitorToggleButton';
 import RelativeDateCellConnector from 'Components/Table/Cells/RelativeDateCellConnector';
 import TableRowCell from 'Components/Table/Cells/TableRowCell';
 import TableRow from 'Components/Table/TableRow';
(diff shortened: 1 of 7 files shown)


frontend/src/Series/Details/EpisodeRow.js
```diff
@@ -193,6 +195,17 @@ class EpisodeRow extends Component {
               );
             }
 
+            if (name === 'customFormatScore') {
```
markus101, line 198: Thoughts on having a tooltip to show the formats as well? That way you could have 1 column with both the score and formats (on mouse over) without having to have formats as it's own separate column

jack-mil: Thats a great idea! Thanks for the snippet below

jack-mil:
Here is the tooltip
image

markus101 (approved):

PR #5804, opened 2023-07-14
Fixed: Improve quality and episode info output in parse result

mynameisbogdan:
#### Database Migration
NO

#### Description
!image


markus101 (changes requested):
Instead of just flexing things into place it'd make more sense to keep things grouped together logically.

For example with Episode Info:

```
Season Number
Episode Number(s)
Absolute Episode Number(s)
Daily
Air Date
```

Would be on the left and the remaining on the right, when two columns are not possible then then vertically it'd be left then right stacked up.

markus101:
Cleaned up a couple things locally and landed.

PR #5806, opened 2023-07-14
New: Show tooltips with custom formats in History and Queue

mynameisbogdan:
#### Database Migration
NO

#### Description
For consistency.


jack-mil:
great call, this can cut down on row length while still providing useful info

markus101 (approved):

PR #5807, opened 2023-07-15
Naming modal update examples

stevietv:
#### Database Migration
NO

#### Description
Brought naming modal examples in line with how Clean works in filename tokens

*

### DIFF
--- a/frontend/src/Settings/MediaManagement/Naming/NamingModal.js
+++ b/frontend/src/Settings/MediaManagement/Naming/NamingModal.js
@@ -30,28 +30,28 @@ const caseOptions = [
 const fileNameTokens = [
   {
     token: '{Series Title} - S{season:00}E{episode:00} - {Episode Title} {Quality Full}',
-    example: 'Series Title (2010) - S01E01 - Episode Title HDTV-720p Proper'
+    example: 'The Series Title\'s! (2010) - S01E01 - Episode Title HDTV-720p Proper'
   },
   {
     token: '{Series Title} - {season:0}x{episode:00} - {Episode Title} {Quality Full}',
-    example: 'Series Title (2010) - 1x01 - Episode Title HDTV-720p Proper'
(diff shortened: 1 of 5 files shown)


stevietv:
One thing I haven't fixed is that using {Series CleanTitle} for example does not correctly clean the example filename shown under the box

markus101:
> One thing I haven't fixed is that using {Series CleanTitle} for example does not correctly clean the example filename shown under the box

That's because `{Series CleanTitle}` doesn't actually remove the `'`, you can verify this by adding a test to `CleanTitleFixture`.

stevietv:
> That's because `{Series CleanTitle}` doesn't actually remove the `'`, you can verify this by adding a test to `CleanTitleFixture`.`

[current tests](https://github.com/Sonarr/Sonarr/blob/dee8820b1f31e9180c55c6d29b950ff6cfe0205f/src/NzbDrone.Core.Test/OrganizerTests/FileNameBuilderTests/CleanTitleFixture.cs#L72) demonstrate that the `'` is actually removed.

This however doesn't get reflected in the example under the naming box when CleanTitle is used and not sure how to fix it as the other types of naming tokens do get applied: 

image

markus101:
But not `'s`

image

stevietv:
> But not `'s`
> 
> image

ahh thats the difference! I missed that nuance in the regex for following `s` and `m`.
I'll tweak the modal again to be in line with that :)

stevietv:
I see, so we keep the `'s` or `'m` only when its on the last word, otherwise its cleaned up.

I've added some test cases to demonstrate the functionality of CleanTitle, or specifically the lack of consistency of it. if there is a Year at the end it will remove it, if not the it leaves it in. Further if we use a token that adds the year, then the apostrophe will also get removed

I see what you mean now when you said CleanTitle was a bit of a mess!

PR #5808, opened 2023-07-15
Use specific How To Apply Tags per setting section

mynameisbogdan:
#### Database Migration
NO

#### Description
Not all sections are indexers, so let's split them per use case. Also sorting the translation keys alphabetically.


markus101 (approved):
Good to land once rebased

markus101:
Already approved... don't need to review again...

PR #5810, opened 2023-07-16
Translate System Menus

stevietv:
#### Database Migration
NO

#### Description
add translations to all strings within the System menu/submenus


frontend/src/System/Logs/Files/LogFiles.js
```diff
@@ -76,13 +77,13 @@ class LogFiles extends Component {
         <PageContentBody>
           <Alert>
             <div>
-              Log files are located in: {location}
+              {translate('LogFilesLocation')}: {location}
```
markus101, line 80: `location` should be part of the translation so languages that prefer a different syntax can translate it properly.

frontend/src/System/Status/About/About.js
```diff
@@ -30,54 +31,54 @@ class About extends Component {
...
             data={version}
           />
 
           {
             packageVersion &&
               <DescriptionListItem
-                title="Package Version"
-                data={(packageAuthor ? <span> {packageVersion} {' by '} <InlineMarkdown data={packageAuthor} /> </span> : packageVersion)}
+                title={translate('PackageVersion')}
+                data={(packageAuthor ? <span> {packageVersion} {translate('by')} <InlineMarkdown data={packageAuthor} /> </span> : packageVersion)}
```
markus101, line 45: Similar thing here, though this is more complicated with the HTML.

frontend/src/System/Status/About/About.js
```diff
@@ -30,54 +31,54 @@ class About extends Component {
...
               />
           }
 
           {
             isNetCore &&
               <DescriptionListItem
-                title=".Net Version"
+                title={translate('NetCore')}
```
markus101, line 52: Core isn't part of the official name any more.

markus101 (changes requested):

frontend/src/System/Status/About/About.js
stevietv: i used NetCore to be in line with radarr, but adjusted to DotNetVersion instead.

frontend/src/System/Status/About/About.js
stevietv: So for this I went with translating the whole thing and if a packageAuthor exists wrapping it all in an `InlineMarkdown`.

frontend/src/System/Status/MoreInfo/MoreInfo.js
```diff
@@ -12,52 +13,70 @@ class MoreInfo extends Component {
...
           <DescriptionListItemDescription>
             <Link to="https://discord.sonarr.tv/">discord.sonarr.tv</Link>
           </DescriptionListItemDescription>
 
-          <DescriptionListItemTitle>IRC</DescriptionListItemTitle>
+          <DescriptionListItemTitle>
+            {translate('IRC')}
+          </DescriptionListItemTitle>
           <DescriptionListItemDescription>
             <Link to="irc://irc.libera.chat/#sonarr">#sonarr on Libera</Link>
```
markus101, line 57: I think we could translate the link text as well

markus101 (approved):

frontend/src/System/Status/MoreInfo/MoreInfo.js
stevietv: added in #5823

PR #5772, opened 2023-07-02
New: Async and HTTP/2 in HttpClient

mynameisbogdan:
#### Database Migration
NO

#### Description
- Releases/Indexer search changed to use the async requests
- Download/Grab releases changed to use the async requests
- Add async methods to RateLimitService
- Changed ProcessDownloadDecisions to be async
- Changed the default Request HTTP version to 2.0

PR #5811, opened 2023-07-16
New: Async HttpClient

Qstick:
#### Database Migration
NO

#### Description
Creates Async methods for HttpClient as a starting point for carrying async operations up the stack.

#### Todos
- [ ] Tests


Qstick:
Seems #5772 already adds async as well... closing this

Issue #5616, opened 2023-05-02
Anime Standard Format Search  (allow search by seasons)


pixelroll:
### Is there an existing issue for this?

- [X] I have searched the existing open and closed issues

### Is your feature request related to a problem? Please describe

This option allows you to search with the standard type but only on episodes and this is all the problem since this type also allows you to find season packs. 

### Describe the solution you'd like

The solution would be that when the standard search is launched, sonarr also looks at the season packs and not just the episodes.

### Describe alternatives you've considered

I tested with the series type and it works very well with some anime, I have the series pack but I miss the search with the anime type...

### Anything else?
2023-05-02 21:18:42.7|Trace|HttpClient|Req: [GET] http://ip:9696/8/api?t=tvsearch&cat=5070,102179&extended=1&apikey=(removed)&offset=0&limit=100&q=Hajime%20no%20Ippo&season=1&ep=69

if we raise the ep parameter we have the season pack
_No response_

fryfrog:
I'm not sure how hard it would be to actually achieve, but it seems logical to do a standard *season* search when the per indexer setting "Anime Standard Format Search" is used. It would be more efficient to do *one* standard search and find X episodes and season packs vs. doing X searches for SxxEyy style results and *not* getting season packs.

ShubhamPatilsd:
Will be reserving this for students at [CodeDay Labs](https://labs.codeday.org/) to be working on in our summer cohort.

markus101:
For anime typical season packs don't always exist, so individual episodes are searched, this returns a lot of results, but is expected given the lack of consistent season packs and unknown batch sizes.

The desired change here is to keep the existing behaviour for season searches of anime and when `AnimeStandardFormatSearch` is enabled also search for season packs.

ttran913:
Ok, I think I have a better picture of this issue now. So I just want to confirm its _current_ behavior for searches.

- When setting up one of a desired anime series, there's a checkbox for `Anime Standard Format Search`. 
- Then I only used interactive searches.
- I've found that there's no difference currently whether the `Anime Standard Format Search` option is checked or not. 
- I get the same results. When searching for a season of an anime, season packs are found, in both on or off for `Anime Standard Format Search`, which makes sense, I'm purposely looking for the whole season. 
- While when searching for individual episodes, only single episodes are found, in, again, in both on or off for `Anime Standard Format Search`, no season packs are listed.

My search setup for current understanding of the app's behavior:

[...]

PR #5814, opened 2023-07-17
AnimeStandardFormatSearch option also looks for season packs

ttran913:
#### Database Migration
NO

#### Description
Implemented the new search criteria object, `AnimeSeasonSearchCriteria`, and applied it where it might be needed throughout the codebase. Generally paired with wherever `AnimeEpisodeSearchCriteria` is referenced. Added some minor tests that are similar to the already existing tests. Now, when the `AnimeStandardFormatSearch` option is enabled in the indexer settings, season packs will also be found in addition to its original function.

#### Todos
- [x] Tests
- [ ] Wiki Updates

* https://github.com/Sonarr/Sonarr/issues/5616

### DIFF
--- a/src/NzbDrone.Core/Indexers/BroadcastheNet/BroadcastheNetRequestGenerator.cs
+++ b/src/NzbDrone.Core/Indexers/BroadcastheNet/BroadcastheNetRequestGenerator.cs
@@ -42,8 +42,8 @@ namespace NzbDrone.Core.Indexers.BroadcastheNet
         public virtual IndexerPageableRequestChain GetSearchRequests(SingleEpisodeSearchCriteria searchCriteria)
         {
             var pageableRequests = new IndexerPageableRequestChain();
-
             var parameters = new BroadcastheNetTorrentQuery();
+
             if (AddSeriesSearchParameters(parameters, searchCriteria))
             {
                 foreach (var episode in searchCriteria.Episodes)
@@ -63,8 +63,8 @@ namespace NzbDrone.Core.Indexers.BroadcastheNet
         public virtual IndexerPageableRequestChain GetSearchRequests(SeasonSearchCriteria searchCriteria)
         {
             var pageableRequests = new IndexerPageableRequestChain();
-
             var parameters = new BroadcastheNetTorrentQuery();
+
             if (AddSeriesSearchParameters(parameters, searchCriteria))
             {
                 foreach (var seasonNumber in searchCriteria.Episodes.Select(v => v.SeasonNumber).Distinct())
@@ -89,8 +89,8 @@ namespace NzbDrone.Core.Indexers.BroadcastheNet
         public virtual IndexerPageableRequestChain GetSearchRequests(DailyEpisodeSearchCriteria searchCriteria)
         {
             var pageableRequests = new IndexerPageableRequestChain();
-
             var parameters = new BroadcastheNetTorrentQuery();
+
             if (AddSeriesSearchParameters(parameters, searchCriteria))
             {
                 parameters.Category = "Episode";
@@ -117,8 +117,8 @@ namespace NzbDrone.Core.Indexers.BroadcastheNet
         public virtual IndexerPageableRequestChain GetSearchRequests(DailySeasonSearchCriteria searchCriteria)
         {
             var pageableRequests = new IndexerPageableRequestChain();
-
             var parameters = new BroadcastheNetTorrentQuery();
+
             if (AddSeriesSearchParameters(parameters, searchCriteria))
             {
                 parameters.Category = "Episode";
@@ -145,8 +145,8 @@ namespace NzbDrone.Core.Indexers.BroadcastheNet
         public virtual IndexerPageableRequestChain GetSearchRequests(AnimeEpisodeSearchCriteria searchCriteria)
         {
             var pageableRequests = new IndexerPageableRequestChain();
-
             var parameters = new BroadcastheNetTorrentQuery();
+
             if (AddSeriesSearchParameters(parameters, searchCriteria))
             {
                 foreach (var episode in searchCriteria.Episodes)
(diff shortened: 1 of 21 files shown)


src/NzbDrone.Core/Indexers/Newznab/NewznabRequestGenerator.cs
```diff
@@ -404,17 +404,45 @@ public virtual IndexerPageableRequestChain GetSearchRequests(AnimeEpisodeSearchC
                         "search",
                         $"&q={NewsnabifyTitle(queryTitle)}+{searchCriteria.AbsoluteEpisodeNumber:00}"));
 
-                    if (Settings.AnimeStandardFormatSearch && searchCriteria.SeasonNumber > 0 && searchCriteria.EpisodeNumber > 0)
+                    if (Settings.AnimeStandardFormatSearch && searchCriteria.SeasonNumber > 0)
```
markus101, line 407: A separate search is required, right now this will get called for every episode in the season.

src/NzbDrone.Core/Indexers/Newznab/NewznabRequestGenerator.cs
```diff
@@ -404,17 +404,45 @@ public virtual IndexerPageableRequestChain GetSearchRequests(AnimeEpisodeSearchC
...
-                            searchCriteria,
-                            $"&season={NewznabifySeasonNumber(searchCriteria.SeasonNumber)}&ep={searchCriteria.EpisodeNumber}");
-
-                        pageableRequests.Add(GetPagedRequests(MaxPages,
-                            Settings.AnimeCategories,
-                            "tvsearch",
-                            $"&q={NewsnabifyTitle(queryTitle)}&season={NewznabifySeasonNumber(searchCriteria.SeasonNumber)}&ep={searchCriteria.EpisodeNumber}"));
+                        if (searchCriteria.IsSeasonSearch)
+                        {
+                            if (searchCriteria.SeasonNumber < 10)
```
markus101, line 411: With the `season` query param Sonarr doesn't need to pad the season number, the indexer should handle it correctly regardless and in the case of Prowlarr and Jackett they know best how to query the different indexers. This is the same as `ep`, which isn't padded.

markus101 (changes requested):

markus101:
In order to get the latest changes you need to rebase on `develop`, instead of merging develop into your branch. Depending how you have your git setup, it'll be something like `git rebase upstream/develop`.

ttran913:
We think we made some successful modifications. We hope we followed your suggestions sufficiently?

src/NzbDrone.Core/Indexers/Newznab/NewznabRequestGenerator.cs
ttran913: I made it so that it does another search separately in `ReleaseSearchService` for the season alone.

src/NzbDrone.Core/Indexers/Newznab/NewznabRequestGenerator.cs
ttran913: We removed on what we think the unnecessary code was since the indexers should already know best.

ttran913:
Okay, we made so that it searches for the season pack only on the search of the first episode when conducting a whole season search. It will, of course, _not_ look for season packs when doing the individual search on the first episode.

markus101 (changes requested):
Instead of only handling this when searching for the first episode a separate search should be triggered to the indexer.

This will require a new search criteria, `AnimeSeasonSearchCriteria` and handling in each indexer.

ttran913:
We implemented the new search criteria object, `AnimeSeasonSearchCriteria`, and implemented it where it might be needed throughout the codebase. We also added some minor tests that are similar to the already existing tests.

src/NzbDrone.Core/Indexers/Newznab/NewznabRequestGenerator.cs
```diff
@@ -422,6 +422,33 @@ public virtual IndexerPageableRequestChain GetSearchRequests(AnimeEpisodeSearchC
...
+            var pageableRequests = new IndexerPageableRequestChain();
+
+            if (SupportsSeasonSearch)
+            {
+                var queryTitles = TextSearchEngine == "raw" ? searchCriteria.SceneTitles : searchCriteria.CleanSceneTitles;
+                foreach (var queryTitle in queryTitles)
+                {
+                    if (Settings.AnimeStandardFormatSearch && searchCriteria.SeasonNumber > 0)
+                    {
+                        AddTvIdPageableRequests(pageableRequests,
```
markus101, line 436: This needs to be moved outside of the loop, right now it's triggered for each query title instead of just once. Looks like the original search for anime episodes does this as well, I'll fix those up.

markus101 (changes requested):

markus101:
Tested well except for the duplicate ID searches, just some things to address.

src/NzbDrone.Core/Indexers/BroadcastheNet/BroadcastheNetRequestGenerator.cs
```diff
@@ -173,6 +173,32 @@ public virtual IndexerPageableRequestChain GetSearchRequests(AnimeEpisodeSearchC
             return pageableRequests;
         }
 
+        public virtual IndexerPageableRequestChain GetSearchRequests(AnimeSeasonSearchCriteria searchCriteria)
+        {
+            var pageableRequests = new IndexerPageableRequestChain();
+
+            var parameters = new BroadcastheNetTorrentQuery();
```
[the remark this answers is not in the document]
ttran913: All our formatting here and below was due to the fact that we were just trying to match style of the rest of the file.

ttran913:
> Tested well except for the duplicate ID searches, just some things to address.

@markus101 How did you test for that? Was it in the testing files?

markus101 (approved):
Thanks for the fixes and cleanup, working as expected now.

PR #5824, opened 2023-07-20
Add support for deprecated values in field select options

mynameisbogdan:
#### Database Migration
NO

#### Description
Adding support for deprecated enums in SelectOptions with the attribute Obsolete by not displaying them in values anymore while maintaining BC.


markus101:
What's the use case?

Issue #5828, opened 2023-07-22
Jellyfin notification API changes


JeWe37:
### Is there an existing issue for this?

- [X] I have searched the existing open and closed issues

### Current Behavior

With jellyfin/jellyfin#9147 having been merged notifications will not be sent to Jellyfin properly anymore.

### Expected Behavior

This should likely be updated to use the Events API.

### Steps To Reproduce

1. Run the jellyfin/jellyfin:unstable docker container
2. Add a connection to it from Sonarr
3. Add something that would cause a notification
4. See that no notifications are sent

### Environment

```markdown
- OS: Irrelevant
- Sonarr: develop
- Docker Install: Irrelevant
- Using Reverse Proxy: Irrelevant
- Browser: Irrelevant
```

### What branch are you running?

Develop

### Trace Logs?

```
[v10.0.0.6501] NzbDrone.Common.Http.HttpException: HTTP request failed: [404:NotFound] [POST] at [https://jellyfin.jendrik.malined.com:443/Notifications/Admin]
   at NzbDrone.Common.Http.HttpClient.Execute(HttpRequest request) in ./Sonarr.Common/Http/HttpClient.cs:line 108
   at NzbDrone.Common.Http.HttpClient.Post(HttpRequest request) in ./Sonarr.Common/Http/HttpClient.cs:line 318
   at NzbDrone.Core.Notifications.Emby.MediaBrowserProxy.ProcessRequest(HttpRequest request, MediaBrowserSettings settings) in ./Sonarr.Core/Notifications/MediaBrowser/MediaBrowserProxy.cs:line 97
   at NzbDrone.Core.Notifications.Emby.MediaBrowserProxy.Notify(MediaBrowserSettings settings, String title, String message) in ./Sonarr.Core/Notifications/MediaBrowser/MediaBrowserProxy.cs:line 32
   at NzbDrone.Core.Notifications.Emby.MediaBrowserService.Notify(MediaBrowserSettings settings, String title, String message) in ./Sonarr.Core/Notifications/MediaBrowser/MediaBrowserService.cs:line 30
   at NzbDrone.Core.Notifications.Emby.MediaBrowser.OnEpisodeFileDelete(EpisodeDeleteMessage deleteMessage) in ./Sonarr.Core/Notifications/MediaBrowser/MediaBrowser.cs:line 54
[...]

PR #5829, opened 2023-07-22
New: Allow updating Jellyfin/Emby library by name

### LINKED ISSUE #5826: Jellyfin/Emby Update by Name

Currently, Jellyfin and Emby are notified of changes to files [via the path of the series](https://github.com/Sonarr/Sonarr/blob/dee8820b1f31e9180c55c6d29b950ff6cfe0205f/src/NzbDrone.Core/Notifications/MediaBrowser/MediaBrowserProxy.cs#L34C25-L34C25), which doesn't work when, for instance, Jellyfin and Sonarr are in containers with different mounts.

Especially when Metadata is on, Jellyfin's title and year should line up with Sonarr's, allowing us to obtain the path from Jellyfin's API and use that rather than Sonarr's. This should be an option as for cases where Metadata is off, there may be edge cases where this mapping won't be definitive.

One alternative would be to integrate some form of remote path mapping into the connection, however that will be difficult with the current menu system and presents problems if in addition to using a different base directory, the names of your files do not actually line up between Sonarr and Jellyfin(either via some form of filesystem that causes that or maintaining a hardlink set). I happen to use such a setup in order to get some Jellyfin features that are very particular about the naming and incompatible with what I need for my Sonarr library.

### Anything else?

There seem to be [some problems](https://github.com/jellyfin/jellyfin/pull/7965) with some parts of the Jellyfin API that might be useful here when using an API token as Sonarr would, but those are fixed in development versions so can be used now.

JeWe37:
#### Database Migration
NO

#### Description

As described in the issue. The API situation is a little complicated right now. For this to work properly, you have to use 10.9.0 of Jellyfin which is still in development. However, in that version the notification APIs have also changed(see #5828 ), the one for updating the library has not though. So if you are running 10.9.0, Sonarr's notifications won't work, but updating the library with the names and with paths will. On 10.8.x, notifications will work and updating the library by path will work, but not by name.

One note: I removed the `/mediaprovider` prefix to all the notification paths. In so far as I can tell that is simply not part of the API. I am not entirely sure how this ever worked.

I'm not addressing the notification API situation here, because the Webhooks plugin that is supposed to replace the prior notification API seems to not have been updated yet.

It probably makes sense to just merge this for now as it is an entirely optional feature, and add the changes as required for #5828 once Jellyfin reaches a new version officially.

* Fixes #5826


markus101:
Instead of name can it use one of the IDs Sonarr already has?

JeWe37:
Not for constraining the query. For the FindAll it is possible though, but then the question becomes how to prioritize depending on which IDs are available(tvdb, IMDb, TVMaze etc.). Do you have any preference?

markus101:
TheTVDB should always be preferred as we know that's as accurate as it can be unless it was removed as a duplicate, from there IMDB is the most common.

JeWe37:
Alright. I ended up implementing a ranking system. First compares by tvdbid, then imdbid, then tvmazeid, then tvrageid and finally if all else fails by name. The best scoring shows, if any of these matched at all, are then _all_ updated. This might be useful people have multiple libraries referencing the same files.

JeWe37:
I also added path mappings analogously to Plex in v4 as suggested by @bakerboy448 [here](https://github.com/Sonarr/Sonarr/issues/5826#issuecomment-1648892842). Both options have their use cases I think.

src/NzbDrone.Core/Notifications/MediaBrowser/MediaBrowserSettings.cs
```diff
@@ -43,6 +45,15 @@ public MediaBrowserSettings()
         [FieldDefinition(5, Label = "Update Library", HelpText = "Update Library on Import, Rename, or Delete?", Type = FieldType.Checkbox)]
         public bool UpdateLibrary { get; set; }
 
+        [FieldDefinition(5, Label = "Update Library By Name", HelpText = "Update Library by name rather than path(Requires 'UpdateLibrary')", Type = FieldType.Checkbox, Advanced = true)]
+        public bool UpdateLibraryByName { get; set; }
+
+        [FieldDefinition(5, Label = "Map Paths From", HelpText = "Sonarr path, used to modify series paths when Emby/Jellyfin sees library path location differently from Sonarr(Requires 'UpdateLibrary')", Type = FieldType.Textbox, Advanced = true)]
```
markus101, line 51: They all can't be the same index.

src/NzbDrone.Core/Notifications/MediaBrowser/MediaBrowserSettings.cs
```diff
@@ -43,6 +45,15 @@ public MediaBrowserSettings()
         [FieldDefinition(5, Label = "Update Library", HelpText = "Update Library on Import, Rename, or Delete?", Type = FieldType.Checkbox)]
         public bool UpdateLibrary { get; set; }
 
+        [FieldDefinition(5, Label = "Update Library By Name", HelpText = "Update Library by name rather than path(Requires 'UpdateLibrary')", Type = FieldType.Checkbox, Advanced = true)]
```
markus101, line 48: It's not just name anymore, but we don't provide this option elsewhere we should just assume that it's wanted and update an de-duped paths found by ID or name as well as anything that is remapped.

markus101 (changes requested):

src/NzbDrone.Core/Notifications/MediaBrowser/MediaBrowserProxy.cs
```diff
@@ -31,6 +35,71 @@ public void Notify(MediaBrowserSettings settings, string title, string message)
...
+                .Resource(path)
+                .AddQueryParam("recursive", "true")
+                .AddQueryParam("includeItemTypes", "Series")
+                .AddQueryParam("fields", "Path,ProviderIds")
+                .AddQueryParam("years", series.Year)
+                .Build();
+
+            try
+            {
+                var paths = ProcessGetRequest<MediaBrowserItems>(request, settings).Items.GroupBy(item =>
```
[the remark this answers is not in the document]
JeWe37: Essentially match if any of the IDs are the same, then fall back on names if that fails?

src/NzbDrone.Core/Notifications/MediaBrowser/MediaBrowserSettings.cs
JeWe37: If I understand correctly, you would want to automatically just update every possible option and let Jellyfin figure out whether there's actually anything to do? That seems a bit wasteful perhaps. I don't think this is obscure enough as that having it would be confusing.

src/NzbDrone.Core/Notifications/MediaBrowser/MediaBrowserProxy.cs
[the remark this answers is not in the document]
markus101: Yeah, exactly.

src/NzbDrone.Core/Notifications/MediaBrowser/MediaBrowserSettings.cs
markus101: That's what we do with Kodi and previously with Plex (before a lot of people changed away from TheTVDB metadata in Plex), we smartly find the matching series and update that instead of updating nothing because the paths don't match before falling back to everything.

The vast majority of people are going to have a single library where that series exists, Sonarr will find the matching series and update the path that Jellyfin reports. The only thing wasted would be an API request to try and find the matching series.

If the user has set up a mapping of paths and Sonarr finds a path from JF and they're the same then it's still only one path being updated, if they happen to be different then there would be multiple.

JeWe37:
Alright, I hope that that's what you had in mind. Kept the ordering, but now only distinguishing between found by id, found by name and not found at all.

markus101 (approved):
Thanks for the fixes.

JeWe37:
Actually hold up before you merge this, I believe as is this might break updating in 10.8.10 and below. I'll have to test that still. But should amount to adding a try catch block at most to avoid that.

JeWe37:
Alright, got that fixed. Just amounted to not throwing an exception and just defaulting to returning an no paths.

PR #5834, opened 2023-07-22
New: Add ValueHash for hidden API keys via the API

mynameisbogdan:
#### Database Migration
NO

#### Description
Looks like 570be882154e73f8ad1de5b16b957bcb964697fd is breaking the applications sync from Prowlarr making it see the remote indexer always changed due to mismatch between Prowlarr's apikey and `********` thus triggering a new indexer sync every 6 hours or when sync apps is manually triggered. So I'm adding the value hash to have something to compare.

### DIFF
--- a/src/Sonarr.Http/ClientSchema/Field.cs
+++ b/src/Sonarr.Http/ClientSchema/Field.cs
@@ -13,6 +13,7 @@ namespace Sonarr.Http.ClientSchema
         public string HelpTextWarning { get; set; }
         public string HelpLink { get; set; }
         public object Value { get; set; }
+        public string ValueHash { get; set; }
         public string Type { get; set; }
         public bool Advanced { get; set; }
         public List<SelectOption> SelectOptions { get; set; }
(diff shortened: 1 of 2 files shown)


markus101:
This needs to be done differently, returning a hashed password like this is a bad idea.

markus101:
There isn't anything you can do to compare those fields directly otherwise you expose what you're trying to hide.

markus101 (approved):

Taloth:
Fyi, there are two scenarios where this is relevant:
- Prowlarr api key changes, in which case Prowlarr could just force a sync, without having to check individually if it's different.
- User manually changed Sonarr and fucked something up, a force sync or deleting the indexer would fix that.

Neither scenario _really_ needs this hash.

The reason I bring this up is that either Prowlarr needs to be aware that valueHash is sha256 of first 6 chars, which is a very bad interface contract to make regarding hashes (The hash should be opaque). Or Prowlarr needs to store the hash after a sync has been performed to compare it, which you don't want to do coz it adds db stuff in Prowlarr.
So yeah, simply having a force sync button in prowlarr and otherwise ignore the apikey would be better, imo.

I wouldn't reject this PR based on that, but it's a point of concern.

Issue #5839, opened 2023-07-24
Bug with Importing Anime Watchlist from Simkl


jtnguyen45:
### Is there an existing issue for this?

- [X] I have searched the existing open and closed issues

### Current Behavior

When importing a watchlist from Simkl, users are unable to import their lists if they have an anime type that is categorized as "ova", "ona", "music video", or "special". Currently, we only account of anime types that were "tv" or "movie". We need to add the enum values of SimklAnimeType to account for the different anime types.

This is the error message the user gets when they are trying to import their anime list with a different anime type:
```
Unable to connect to import list: Error converting value "ona" to type 'NzbDrone.Core.ImportLists.Simkl.SimklAnimeType'. Path 'anime[2].anime_type', line 1, position 2094
```

### Expected Behavior

The expected behavior was to be able to import an anime watchlist, as long as the anime was the first season of a series and not of movie type.

### Steps To Reproduce

1. Go to "Settings"
2. Go to "Import Lists"
3. Click the large "+" button
4. Click "Simkl User Watchlist" button
5. Go to the "Show Type" drop down and select "Anime"
6. Click "Authenticate with Simkl" button
7. Click the "Save" button

At this point, if you have any anime type of ova, ona, music video, or special in your anime watchlist, it will not import your watchlist and give an error message.

### Environment

```markdown
- OS: Windows 10 (OS Build 19045.3086)
- Sonarr: 10.0.0.19541
- Docker Install: No
- Using Reverse Proxy: No
```

### What branch are you running?

Develop

### Trace Logs?

No

### Anything else?

Bug introduced from new feature [Simkl Anime List Integration](https://github.com/Sonarr/Sonarr/pull/5766)

jtnguyen45:
@iceypotato and I are working on this. We just need to add to the enum SimklAnimeType in the SimklAPI.cs and change the logic in the SimklParser.cs. We will skip over importing the music videos, since Lidarr handles music.

Not all OVAs, ONAs, and specials have a TVDB id, so we will change the logic to not import any anime that do not have a TVDB id.

PR #5840, opened 2023-07-24
Importing additional Anime Types from Simkl

jtnguyen45:
#### Database Migration
NO

#### Description
Users were getting error messages when trying to import their anime watchlist if they had the any OVAs, ONAs, Special, or Music Video anime types included in their watchlist. Added enum entries in simkl API, as well as tweaked the import warning message to fix this issue.

Fixes https://github.com/Sonarr/Sonarr/issues/5839

### DIFF
--- a/src/NzbDrone.Core/ImportLists/Simkl/SimklAPI.cs
+++ b/src/NzbDrone.Core/ImportLists/Simkl/SimklAPI.cs
@@ -73,8 +73,10 @@ namespace NzbDrone.Core.ImportLists.Simkl
         public DateTime All { get; set; }
     }
 
+    [JsonConverter(typeof(TolerantEnumConverter))]
     public enum SimklAnimeType
     {
+        Unknown,
         Tv,
         Movie,
         Ova,
--- a/src/NzbDrone.Core/ImportLists/Simkl/SimklAPI.cs
+++ b/src/NzbDrone.Core/ImportLists/Simkl/SimklAPI.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using System.Runtime.Serialization;
 using Newtonsoft.Json;
 
 namespace NzbDrone.Core.ImportLists.Simkl
@@ -75,6 +76,12 @@ namespace NzbDrone.Core.ImportLists.Simkl
     public enum SimklAnimeType
     {
         Tv,
-        Movie
+        Movie,
+        Ova,
+        Ona,
+        Special,
+
+        [EnumMember(Value = "music video")]
+        Music
     }
 }
--- a/src/NzbDrone.Core/ImportLists/Simkl/SimklParser.cs
+++ b/src/NzbDrone.Core/ImportLists/Simkl/SimklParser.cs
@@ -43,7 +43,7 @@ namespace NzbDrone.Core.ImportLists.Simkl
                 {
                     var tentativeTvdbId = int.TryParse(show.Show.Ids.Tvdb, out var tvdbId) ? tvdbId : 0;
 
-                    if (tentativeTvdbId > 0 && (show.AnimeType == SimklAnimeType.Tv || show.AnimeType == SimklAnimeType.Ona || show.AnimeType == SimklAnimeType.Ova || show.AnimeType == SimklAnimeType.Special))
+                    if (tentativeTvdbId > 0 && (show.AnimeType is SimklAnimeType.Tv or SimklAnimeType.Ona or SimklAnimeType.Ova or SimklAnimeType.Special))
                     {
                         series.AddIfNotNull(new ImportListItemInfo()
                         {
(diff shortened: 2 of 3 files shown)


src/NzbDrone.Core/ImportLists/Simkl/SimklAPI.cs
```diff
@@ -75,6 +76,12 @@ public class SimklTvSyncActivityResource
     public enum SimklAnimeType
     {
         Tv,
-        Movie
```
markus101, line 78: A value of `Unknown` above `Tv` would be best to capture all the types Sonarr doesn't know about, which will be used if it doesn't match a known type.

src/NzbDrone.Core/ImportLists/Simkl/SimklParser.cs
```diff
@@ -43,7 +43,7 @@ public virtual IList<ImportListItemInfo> ParseResponse(ImportListResponse import
                 {
                     var tentativeTvdbId = int.TryParse(show.Show.Ids.Tvdb, out var tvdbId) ? tvdbId : 0;
 
-                    if (tentativeTvdbId > 0 && show.AnimeType == SimklAnimeType.Tv)
+                    if (tentativeTvdbId > 0 && show.AnimeType != SimklAnimeType.Movie && show.AnimeType != SimklAnimeType.Music)
```
markus101, line 46: Leaving this to compare to Tv makes it easier to read and doesn't need to be updated every time a new value is added.

src/NzbDrone.Core/ImportLists/Simkl/SimklParser.cs
```diff
@@ -55,7 +55,7 @@ public virtual IList<ImportListItemInfo> ParseResponse(ImportListResponse import
                     }
                     else
                     {
-                        Logger.Warn("Skipping info grabbing for '{0}' because it is a movie or it is not the first season of the show", show.Show.Title);
+                        Logger.Warn($"Skipping info grabbing for '{show.Show.Title}' because it is missing a TVDB ID, or it is movie or music video");
```
markus101, line 58: Similar here, but instead of movie it should say something like "unsupported content type"

markus101 (changes requested):

src/NzbDrone.Core/ImportLists/Simkl/SimklParser.cs
jtnguyen45: We were accounting for the different anime types ONAs, OVAs, and Specials if they had a TVDB id. Should we not include them? Cyberpunk: Edgerunners is listed as an ONA, for example.

markus101: In that case opting into the ones we want is probably safer than excluding the ones we don't

jtnguyen45: That makes sense. I will do that.

jtnguyen45:
We added a [class](https://gist.github.com/gubenkoved/999eb73e227b7063a67a50401578c3a7) that helps handle AnimeTypes that are not defined in the SimklAnimeType enum. This should fix the import list bug.

src/NzbDrone.Core/ImportLists/Simkl/SimklParser.cs
```diff
@@ -55,7 +55,7 @@ public virtual IList<ImportListItemInfo> ParseResponse(ImportListResponse import
                     }
                     else
                     {
-                        Logger.Warn("Skipping info grabbing for '{0}' because it is a movie or it is not the first season of the show", show.Show.Title);
+                        Logger.Warn($"Skipping info grabbing for '{0}' because it is an unsupported content type.", show.Show.Title);
```
markus101, line 58: The `$` is unneeded.

src/NzbDrone.Core/ImportLists/Simkl/SimklAPI.cs
markus101: My mistake, I forgot that parsing that enum wouldn't work with the unknown values out of the box, but what you found will work.

src/NzbDrone.Core/ImportLists/Simkl/SimklParser.cs
```diff
@@ -43,7 +43,7 @@ public virtual IList<ImportListItemInfo> ParseResponse(ImportListResponse import
                 {
                     var tentativeTvdbId = int.TryParse(show.Show.Ids.Tvdb, out var tvdbId) ? tvdbId : 0;
 
-                    if (tentativeTvdbId > 0 && show.AnimeType == SimklAnimeType.Tv)
+                    if (tentativeTvdbId > 0 && (show.AnimeType == SimklAnimeType.Tv || show.AnimeType == SimklAnimeType.Ona || show.AnimeType == SimklAnimeType.Ova || show.AnimeType == SimklAnimeType.Special))
```
markus101, line 46: With C# 9 you can now do the following which makes it a little easier to read and add more options if the need arises.

```suggestion
                    if (tentativeTvdbId > 0 && (show.AnimeType is SimklAnimeType.Tv or SimklAnimeType.Ona or SimklAnimeType.Ova or SimklAnimeType.Special))
```

markus101 (changes requested):
Nice solution for the enum problem, just a couple small suggestions.

markus101 (approved):

PR #5845, opened 2023-07-26
Fixed: Add dedupe releases rule based on indexer priority

mynameisbogdan:
#### Database Migration
NO

#### Description
This comes handy when you're using two instances of the same indexer (eg. one freeleech and the other normal), and would rather show the FL releases first based on the indexer priority.

### DIFF
--- a/src/NzbDrone.Core/IndexerSearch/ReleaseSearchService.cs
+++ b/src/NzbDrone.Core/IndexerSearch/ReleaseSearchService.cs
@@ -548,9 +548,10 @@ namespace NzbDrone.Core.IndexerSearch
 
         private List<DownloadDecision> DeDupeDecisions(List<DownloadDecision> decisions)
         {
-            // De-dupe reports by guid so duplicate results aren't returned. Pick the one with the least rejections.
-
-            return decisions.GroupBy(d => d.RemoteEpisode.Release.Guid).Select(d => d.OrderBy(v => v.Rejections.Count()).First()).ToList();
+            // De-dupe reports by guid so duplicate results aren't returned. Pick the one with the least rejections and higher indexer priority.
+            return decisions.GroupBy(d => d.RemoteEpisode.Release.Guid)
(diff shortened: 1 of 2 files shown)


src/NzbDrone.Core/IndexerSearch/ReleaseSearchService.cs
```diff
@@ -548,9 +548,10 @@ private List<DownloadDecision> Dispatch(Func<IIndexer, IEnumerable<ReleaseInfo>>
 
         private List<DownloadDecision> DeDupeDecisions(List<DownloadDecision> decisions)
         {
-            // De-dupe reports by guid so duplicate results aren't returned. Pick the one with the least rejections.
-
-            return decisions.GroupBy(d => d.RemoteEpisode.Release.Guid).Select(d => d.OrderBy(v => v.Rejections.Count()).First()).ToList();
+            // De-dupe reports by guid so duplicate results aren't returned. Pick the one with the least rejections and higher indexer priority.
+            return decisions.GroupBy(d => d.RemoteEpisode.Release.Guid)
+                .Select(d => d.OrderBy(v => v.Rejections.Count()).ThenBy(v => v.RemoteEpisode?.Release?.IndexerPriority ?? 100).First())
```
markus101, line 553: The default priority is 25: https://github.com/Sonarr/Sonarr/blob/1a4403e0ab9ab824c97acf97174f894b3770f2ef/src/NzbDrone.Core/Indexers/IndexerDefinition.cs#L14 though worth extracting to a constant so we're not using a magic number

markus101 (changes requested):

src/NzbDrone.Core/IndexerSearch/ReleaseSearchService.cs
markus101: Should be pretty rare that we have duplicates with the same GUID and no indexer, since that'd mostly come from pushed releases I think.

markus101 (approved):

PR #5846, opened 2023-07-26
Added table identifier to OrderBy to avoid column ambiguity on joins.

kharenis:
#### Database Migration
NO

#### Description
Resolve query ambiguity issue for Episode Missing Search and Episode Cutoff unmet. 

* https://github.com/Sonarr/Sonarr/pull/5559#issuecomment-1636852942


Qstick:
Views like queue use order by keys from different tables, thus using `_table` (the current repo table name) will result in errors in other places.

Qstick:
Integrated

Issue #5148, opened 2022-09-09
My Anime List integration


snoopy82481:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your feature request related to a problem? Please describe

The issue is not related to a problem.  My Anime List has an API for connecting to the lists available there.

### Describe the solution you'd like

The ability to add My Anime List as an option in `Settings > Import List`.

### Describe alternatives you've considered

Continue with manually adding Anime series to the application.

### Anything else?

[MyAnimeList authorization Documentation](https://myanimelist.net/apiconfig/references/authorization)
[MyAnimeList API (beta ver.) (2) Documentation](https://myanimelist.net/apiconfig/references/api/v2)

thezoggy:
sonarr uses tvdb, i dont see tvdb info being returned on myanimelist? so doesnt look like it could be used

markus101:
In addition to lacking TheTVDB information it appears there would be multiple entries for different "seasons" (as they are listed on TheTVDB) of the same series so even if the title matched for the original series it's likely going to result in bad matching.

I don't see this being something we'd implement or support given how poorly title based match can be.

GNS7:
Support for Anilist would be good too, but I don't know if it is possible.

snoopy82481:
> In addition to lacking TheTVDB information it appears there would be multiple entries for different "seasons" (as they are listed on TheTVDB) of the same series so even if the title matched for the original series it's likely going to result in bad matching.
> 
> I don't see this being something we'd implement or support given how poorly title based match can be.

I think if there is XEM linking it could work.  Since there is already logic for searching Anime against XEM to TVDB id.  Other option would be to add an additional metadata source to AniDB, then again using XEM to aggregate the data to TVDB id. It could allow for XEM to be updated programmatically instead of manually. 

I think the benefit of the additional metadata source would be better grabs on anime.  The downside, another metadata source to monitor for API changes and updates.

thezoggy:
> > In addition to lacking TheTVDB information it appears there would be multiple entries for different "seasons" (as they are listed on TheTVDB) of the same series so even if the title matched for the original series it's likely going to result in bad matching.
> > I don't see this being something we'd implement or support given how poorly title based match can be.
> 
[...]

PR #5850, opened 2023-07-28
AniList import list module

Xabis:
#### Database Migration
NO

#### Description
Adds the ability to import lists from the anilist.co web service.

This is related to issue #5148.

#### Mappings:
As Sonarr exclusively uses TVDB for its content, a mapping between services is required.
	
I have chosen the mapping data hosted on this github repository:
https://github.com/Fribb/anime-lists/

When there is no mapping between services available, the item is skipped, and a warning is emitted to the log.
		
It is unknown what the long-term support for this resource is, but seems to be actively maintained at least in the interim.

#### Authentication
Anilist uses oAuth, so authentication has been plumbed in with that in mind.
	
However, I do not own the auth.servarr.com service, so the client id and token urls that are baked in this PR will not work out of the box. If this PR is accepted, it is expected that an official developer will make an anilist.co account, generate a client, update the servarr auth server, then update the code here on my behalf.
	
I have successfully tested the authentication flow using a local Servarr.Auth instance.

#### Paging
Anilist has a unique requirement, where queried lists have a maximum number of returned items of fifty, and may not be increased.
	
Instead, consumers of the API are expected to paginate the results, using the returned page information in each query response.
	
This means the pagination must be reactionary, rather than known upfront. As such, the fetching logic was overridden to accommodate this flow. It was unclear if the original pagination supported this type of flow, so let me know if this needs another look.

#### Future
Anilist tracks watching progress and returns that data during import. A typical setup will automatically sync progress as episodes are watched, using tools such as taiga, jellyfin, kody, etc.
	
A way to incorporate this information back into the sonarr pipeline would be great. It may be possible to map an episode season to an absolute episode number, using another 3rd party mapping service. This information could then be used to automatically turn monitoring off for those ranges.

However, setting the monitor type to "future episodes' may be enough for most users.

Additionally, allowing the importer to automatically add tags would be a boon. In this case, the importer could automatically tag a series with the user's current watching status such as: watching, planning, on hold, etc. The generic tag input setting would not be sufficient in this case without resorting to managing multiple importers.


src/NzbDrone.Core/ImportLists/AniList/AniListImportBase.cs
```diff
@@ -0,0 +1,205 @@
...
+        public override TimeSpan MinRefreshInterval => TimeSpan.FromHours(12);
+
+        private const string _cacheKey = "animemappings";
+        private readonly TimeSpan _cacheInterval = TimeSpan.FromHours(20);
+        private readonly ICached<Dictionary<int, MediaMapping>> _cache;
+
+        public const string OAuthUrl = "https://anilist.co/api/v2/oauth/authorize";
+        public const string RedirectUri = "http://localhost:5000/anilist/auth";
+        public const string RenewUri = "http://localhost:5000/anilist/renew";
+        public const string ClientId = "13737";
```
markus101, line 31:
```suggestion
        public const string RedirectUri = "https://auth.servarr.com/v1/anilist_sonarr/auth";
        public const string RenewUri = "https://auth.servarr.com/v1/anilist_sonarr/renew";
        public const string ClientId = "13780";
```

src/NzbDrone.Core/ImportLists/AniList/List/AniListSettings.cs
```diff
@@ -0,0 +1,72 @@
...
+            ImportReleasing = true;
+            ImportFinished = true;
+        }
+
+        protected override AbstractValidator<AniListSettings> Validator => new AniListSettingsValidator();
+
+        [FieldDefinition(1, Label = "Username", HelpText = "Username for the List to import from")]
+        public string Username { get; set; }
+
+        [FieldDefinition(0, Label = "Import Watching", Type = FieldType.Checkbox, Section = sectionImport, Unit = unitListStatus)]
```
markus101, line 39: I don't think using unit for this makes sense, it's not really a unit and as-is it looks awkward at best in the UI. `HelpText` with more information would be more valuable.

src/NzbDrone.Core/ImportLists/AniList/List/AniListParser.cs
```diff
@@ -0,0 +1,146 @@
...
+
+namespace NzbDrone.Core.ImportLists.AniList.List
+{
+    public class AniListParser : IParseImportListResponse
+    {
+        private readonly Dictionary<int, MediaMapping> _mappings;
+        private readonly Logger _logger;
+        private readonly AniListSettings _settings;
+
+        public AniListParser(Logger logger, AniListSettings settings, Dictionary<int, MediaMapping> mappings)
```
markus101, line 18: Nit: Logger last

src/NzbDrone.Core/ImportLists/AniList/List/AniListImport.cs
```diff
@@ -0,0 +1,157 @@
...
+                ClientId = ClientId
+            };
+        }
+
+        public override AniListParser GetParser()
+        {
+            return new AniListParser(_logger, Settings, Mappings);
+        }
+
+        /// <summary>
```
markus101, line 47: This comment makes more sense by the do/while loop

src/NzbDrone.Core/ImportLists/AniList/List/AniListParser.cs
```diff
@@ -0,0 +1,146 @@
...
+
+            var jsonResponse = STJson.Deserialize<MediaPageResponse>(importListResponse.Content);
+
+            if (jsonResponse?.Data?.Page?.MediaList == null)
+            {
+                pageInfo = null;
+                return result;
+            }
+
+            // Filter out unwanted series status types
```
markus101, line 48:
```suggestion
            // Anilist currently does not support filtering this at the query level, they will get filtered out here.
```

src/NzbDrone.Core/ImportLists/AniList/List/AniListParser.cs
```diff
@@ -0,0 +1,146 @@
...
+                {
+                    _logger.Warn("'{1}' (id:{0}) could not be imported, because there is no mapping available.", media.Id, media.Title.UserPreferred ?? media.Title.UserRomaji);
+                }
+            }
+
+            pageInfo = jsonResponse.Data.Page.PageInfo;
+            return result;
+        }
+
+        /// <summary>
```
markus101, line 95: Not necessary

markus101 (changes requested):
Looking pretty good, just some cleanup and updating of the redirect URL to start.

src/NzbDrone.Core/ImportLists/AniList/AniListImportBase.cs
```diff
@@ -0,0 +1,210 @@
...
+                { IsWarning = true };
+            }
+
+            return base.TestConnection();
+        }
+
+        protected void CheckToken()
+        {
+            Settings.Validate().Filter("AccessToken", "RefreshToken").ThrowOnError();
+            _logger.Trace($"Access token expires at {Settings.Expires}");
```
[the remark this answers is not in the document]
Xabis: This pattern was copied from the Trakt import base class, so you may want to have a look for consistency.

src/NzbDrone.Core/ImportLists/AniList/AniListSettingsBase.cs
```diff
@@ -0,0 +1,65 @@
...
+    public class AniListSettingsBase<TSettings> : IImportListSettings
+        where TSettings : AniListSettingsBase<TSettings>
+    {
+        protected virtual AbstractValidator<TSettings> Validator => new AniListSettingsBaseValidator<TSettings>();
+
+        public AniListSettingsBase()
+        {
+            BaseUrl = "https://graphql.anilist.co";
+            SignIn = "startOAuth";
+            MapSourceUrl = "https://raw.githubusercontent.com/Fribb/anime-lists/master/anime-list-full.json";
```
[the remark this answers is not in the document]
Xabis: There seems to be a few of these lists available, but in slightly different formats. I wasn't sure what the best approach here, but as it stands currently a setting does seem silly. Originally, i also had the source and destination keys as settings as well, that way it could be changed on the fly, but not every source is a flat list, so I ended up backing it off.

As you say, a service would be better in order to separate this specific data parsing from the importer itself, especially with multiple imports now in play.

I am assuming that will be a later refactor, though, and if so, the setting here can be removed at that time?

src/NzbDrone.Core/ImportLists/AniList/List/AniListImport.cs
```diff
@@ -0,0 +1,152 @@
...
+                }
+                else
+                {
+                    _logger.Error(ex, "CAPTCHA token required for {0}, check import list settings.", this);
+                }
+            }
+            catch (ImportListException ex)
+            {
+                _importListStatusService.RecordFailure(Definition.Id);
+                _logger.Warn(ex, "{0}", url);
```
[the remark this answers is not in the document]
Xabis: This is copied from the base class ( HttpImportListBase ) as-is, because the FetchItems method could not be used.

Please confirm. And if confirmed, please suggest a message. Whatever is decided here should probably be made to the base class for consistency.

src/NzbDrone.Core/ImportLists/AniList/List/AniListSettings.cs
Xabis: The idea was to separate the status filters that apply only to the users watching status, versus the media airing status.

When they are clumped together with no separation on the UI, it gets confusing (ex: "Import Completed vs "Import Finished')

I was hoping the "Section' would group them visually, but it doesn't seem to do anything at all? I will look into the suggested property. I will remove the unit; I am not even sure what it does, other than add a label to the UI.

src/NzbDrone.Core/ImportLists/AniList/List/AniListParser.cs
Xabis: The reason logger was first here, was because it is a dependency from the container, whereas the setting and mapping objects are not. Typically dependency injection parameters are ordered first, even though it is manually passed in here.

Please confirm.

src/NzbDrone.Core/ImportLists/AniList/List/AniListSettings.cs
Xabis: Help was a better choice. I have updated the labels with prefixes to help differentiate between the two filter sets.

src/NzbDrone.Core/ImportLists/AniList/AniListSettingsBase.cs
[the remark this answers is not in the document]
markus101: I need to look at what it's going to need on the Skyhook side, but regardless I think the setting should be removed.

src/NzbDrone.Core/ImportLists/AniList/List/AniListParser.cs
markus101: Generally logger is one of the last parameters passed in. Though since the parser is not being constructed through DI we shouldn't be passing in the logger from `AniListImport`, instead we should instantiate the logger in the class, like `SimklParser` does.

src/NzbDrone.Core/ImportLists/AniList/AniListSettingsBase.cs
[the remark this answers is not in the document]
Xabis: I removed the setting as requested.

I would like to point out, however, that BaseUrl is also a setting that doesn't really make sense in the same context. However, it is an existing pattern.

src/NzbDrone.Core/ImportLists/AniList/List/AniListParser.cs
Xabis: Refactored as recommended.

Xabis:
@markus101 are there any further issues or concerns blocking this PR that I may address?

src/NzbDrone.Core/ImportLists/AniList/List/AniListParser.cs
```diff
@@ -0,0 +1,143 @@
...
+            {
+                var media = item.Media;
+
+                if (_mappings.TryGetValue(media.Id, out var mapping))
+                {
+                    // Base required data
+                    var entry = new ImportListItemInfo()
+                    {
+                        TvdbId = mapping.Tvdb.Value,
+                        Title = media.Title.UserPreferred ?? media.Title.UserRomaji ?? default
```
markus101, line 64: `?? default` isn't required since it's null for a string

```suggestion
                        Title = media.Title.UserPreferred ?? media.Title.UserRomaji
```

src/NzbDrone.Core/ImportLists/AniList/List/AniListParser.cs
```diff
@@ -0,0 +1,143 @@
...
+                .Where(x => ValidateMediaStatus(x.Media));
+
+            foreach (var item in filtered)
+            {
+                var media = item.Media;
+
+                if (_mappings.TryGetValue(media.Id, out var mapping))
+                {
+                    // Base required data
+                    var entry = new ImportListItemInfo()
```
markus101, line 61:
```suggestion
                    var entry = new ImportListItemInfo
```

src/NzbDrone.Core/ImportLists/AniList/List/AniListParser.cs
```diff
@@ -0,0 +1,143 @@
...
+                    }
+
+                    // Optional Year/ReleaseDate data
+                    if (media.StartDate?.Year != null)
+                    {
+                        entry.Year = (int)media.StartDate.Year;
+                        entry.ReleaseDate = (System.DateTime)media.StartDate.Convert();
+                    }
+
+                    result.AddIfNotNull(entry);
```
markus101, line 85: `Add` is sufficient since we know it's not null because it was instantiated within the block

markus101 (changes requested):
I wanted to make sure that the Skyhook anime list was complete before reviewing again so that we can avoid each instance needing to pull that list (as well as avoiding that need for MAL to do the same).

For lookups you can use anilist:<id> as the search term to skyhook, similar to tvdb:<id> and skyhook will return the TVDB series if a mapping is known. For example: https://skyhook.sonarr.tv/v1/tvdb/search/en?term=anilist:185 which will return Initial D.

`ISearchForNewSeries` and `SkyHookProxy` will need a `SearchForNewSeriesByAniListId` method similar to `SearchForNewSeriesByImdbId` in order to do the search.

Xabis:
> For lookups you can use anilist: as the search term to skyhook

Updated to use the suggested service call.

However, there are noticeable performance impacts importing a large list, due to the nature of involving an external service for each item. Additionally, due to how the test connection logic functions, these calls also occur while saving, which causes the save to take quite a while.

Suggestions/Guidance welcome.

src/NzbDrone.Core/ImportLists/AniList/List/AniListParser.cs
```diff
@@ -0,0 +1,128 @@
...
+            }
+
+            // Anilist currently does not support filtering this at the query level, they will get filtered out here.
+            var filtered = jsonResponse.Data.Page.MediaList
+                .Where(x => ValidateMediaStatus(x.Media));
+
+            foreach (var item in filtered)
+            {
+                var media = item.Media;
+                var mappedSeries = _seriesSearchService.SearchForNewSeriesByAniListId(media.Id.ToString())
```
markus101, line 57: Instead of mapping during the processing results should be returned with their AniListId and when all the results are processed in `ImportListSyncService` Sonarr will perform the search. This will mean Skyhook doesn't get called twice as well, since it would just get the series data when it's looking to add the series instead of once here and once when adding the series.

How it's done for IMDB: https://github.com/Sonarr/Sonarr/blob/b1d12b8ee9b828c780e88ef029f2bb8e0934f053/src/NzbDrone.Core/ImportLists/ImportListSyncService.cs#L84

markus101 (changes requested):
Ahh, that would definitely be slow to map them like that and also means the whole series is fetched, mostly thrown out then fetched again when it's being added.

src/NzbDrone.Core/ImportLists/AniList/List/AniListParser.cs
Xabis: This approach is much better, thanks.

src/NzbDrone.Core/MetadataSource/ISearchForNewSeries.cs
```diff
@@ -7,5 +7,6 @@ public interface ISearchForNewSeries
     {
         List<Series> SearchForNewSeries(string title);
         List<Series> SearchForNewSeriesByImdbId(string imdbId);
+        List<Series> SearchForNewSeriesByAniListId(string aniListId);
```
markus101, line 10:
```suggestion
        List<Series> SearchForNewSeriesByAniListId(int aniListId);
```

src/NzbDrone.Core/ImportLists/AniList/List/AniListImport.cs
```diff
@@ -0,0 +1,152 @@
...
+            catch (RequestLimitReachedException)
+            {
+                _importListStatusService.RecordFailure(Definition.Id, TimeSpan.FromHours(1));
+                _logger.Warn("API Request Limit reached for {0}", this);
+            }
+            catch (CloudFlareCaptchaException ex)
+            {
+                _importListStatusService.RecordFailure(Definition.Id);
+                ex.WithData("FeedUrl", url);
+                if (ex.IsExpired)
```
markus101, line 128: New line before `if`

markus101 (changes requested):
Working great, just a little nit and making `aniListId` an int for the search.

src/NzbDrone.Core/ImportLists/AniList/List/AniListImport.cs
Xabis: This was copied exactly as-is from HttpImportListBase; you may want to apply these formatting standards to the base code.

markus101 (approved):

markus101:
Thanks for working on this!

reitenth:
it seems as though it is re adding shows already on my list.

Xabis:
@reitenth the underlying import service is responsible for de-duplication, not this module. If you believe there is an issue please open a new issue.

pervertkirito:
About pagination, can't you just use normal quary request like you view your anime list profile?
[here example](https://anilist.co/graphiql?query=query%20(%24userId%3A%20Int%2C%20%24userName%3A%20String%2C%20%24type%3A%20MediaType)%20%7B%0A%20%20MediaListCollection(userId%3A%20%24userId%2C%20userName%3A%20%24userName%2C%20type%3A%20%24type)%20%7B%0A%20%20%20%20lists%20%7B%0A%20%20%20%20%20%20name%0A%20%20%20%20%20%20isCustomList%0A%20%20%20%20%20%20isCompletedList%3A%20isSplitCompletedList%0A%20%20%20%20%20%20entries%20%7B%0A%20%20%20%20%20%20%20%20...mediaListEntry%0A%20%20%20%20%20%20%7D%0A%20%20%20%20%7D%0A%20%20%7D%0A%7D%0A%0Afragment%20mediaListEntry%20on%20MediaList%20%7B%0A%20%20id%0A%20%20mediaId%0A%20%20status%0A%20%20score%0A%20%20progress%0A%20%20progressVolumes%0A%20%20repeat%0A%20%20priority%0A%20%20private%0A%20%20hiddenFromStatusLists%0A%20%20customLists%0A%20%20advancedScores%0A%20%20notes%0A%20%20updatedAt%0A%20%20startedAt%20%7B%0A%20%20%20%20year%0A%20%20%20%20month%0A%20%20%20%20day%0A%20%20%7D%0A%20%20completedAt%20%7B%0A%20%20%20%20year%0A%20%20%20%20month%0A%20%20%20%20day%0A%20%20%7D%0A%20%20media%20%7B%0A%20%20%20%20id%0A%20%20%20%20title%20%7B%0A%20%20%20%20%20%20userPreferred%0A%20%20%20%20%20%20romaji%0A%20%20%20%20%20%20english%0A%20%20%20%20%20%20native%0A%20%20%20%20%7D%0A%20%20%20%20coverImage%20%7B%0A%20%20%20%20%20%20extraLarge%0A%20%20%20%20%20%20large%0A%20%20%20%20%7D%0A%20%20%20%20type%0A%20%20%20%20format%0A%20%20%20%20status(version%3A%202)%0A%20%20%20%20episodes%0A%20%20%20%20volumes%0A%20%20%20%20chapters%0A%20%20%20%20averageScore%0A%20%20%20%20popularity%0A%20%20%20%20isAdult%0A%20%20%20%20countryOfOrigin%0A%20%20%20%20genres%0A%20%20%20%20bannerImage%0A%20%20%20%20startDate%20%7B%0A%20%20%20%20%20%20year%0A%20%20%20%20%20%20month%0A%20%20%20%20%20%20day%0A%20%20%20%20%7D%0A%20%20%7D%0A%7D%0A&variables=%7B%0A%20%20%22type%22%3A%20%22ANIME%22%2C%0A%20%20%22userId%22%3A%2010%0A%7D)

Xabis:
> pagination

Collections are grouped, have their own record limits, and may also return _duplicative_ entries if the user has created any custom lists.

Below is an example, pulled from a random profile. In this example, the custom list "Priority numero uno (Planning)" contains duplicated entries that are repeated in the "Planning" list. There is no argument to suppress custom lists in the result that I could tell.
```
{
  "data": {
    "MediaListCollection": {
      "lists": [
        {
          "name": "Watching",
          "isCustomList": false,
          "isSplitCompletedList": false,
          "entries": [...]
        },
        {
          "name": "Planning",
          "isCustomList": false,
          "isSplitCompletedList": false,
          "entries": [...]
        },
        {
          "name": "Priority numero uno (Planning)",
          "isCustomList": true,
          "isSplitCompletedList": false,
          "entries": [...]
        }
      ]
    }
  }
}
```

The MediaListCollection query has the following arguments:
```
chunk: Int
Which chunk of list entries to load

perChunk: Int
The amount of entries per chunk, max 500
```

This means that the maximum total entries that may be returned with this query is 500.

Even if this seems overly generous compared to the Page query, invariably there will be someone who exceeds this limit. So, some form of "pagination" would need to be done regardless.

As such, using the simpler Page query to return a flat de-duplicated list was considered the better approach, at the expense of additional requests.

PR #5852, opened 2023-07-29
Support for MyAnimeList list import

iceypotato:
#### Database Migration
NO

#### Description
Added support for MAL importing, fixes issue 5148

* #5148 

#### Users working on this issue
@jtnguyen45
@iceypotato

#### Background Info
With the help of https://github.com/Fribb/anime-lists, We were able to get MAL list imports working; however, there are still a couple of issues.

We based our refresh token code on Simkl's, and it does not seem to refresh properly. In the "Settings" for the particular ListImport entry, the access and refresh tokens do not get changed and we are still trying to figure out how to fix that. This also raises the question if the Simkl refresh token function works properly; though, I have not tested it.

We also got the "Auth with MAL" button to work, but we do not know if we are following conventions, as we do not fully understand how to use the `RequestAction` function. Our current workaround was to use a separate function to make an HTTP post request to MAL's auth server. MAL requires that the user create an "App" in their MAL account, in which they get the generated Client ID and Client Secret. Then, in order to get the access token, they have to do a http get request to https://myanimelist.net/v1/oauth2/authorize, then use that information to do a post request to https://myanimelist.net/v1/oauth2/token, which the response is in JSON format.


src/NzbDrone.Core/ImportLists/MyAnimeList/MalImport.cs
```diff
@@ -0,0 +1,227 @@
...
+using NzbDrone.Core.Parser;
+using NzbDrone.Core.Parser.Model;
+
+namespace NzbDrone.Core.ImportLists.MyAnimeList
+{
+    public class MalImport : HttpImportListBase<MalListSettings>
+    {
+        public const string OAuthUrl = "https://myanimelist.net/v1/oauth2/authorize";
+        public const string OAuthTokenUrl = "https://myanimelist.net/v1/oauth2/token";
+        public const string RedirectUri = "http://localhost:8989/oauth.html";
```
markus101, line 19:
```suggestion
        public const string RedirectUri = "https://auth.servarr.com/v1/mal_sonarr/auth";
        public const string RenewUri = "https://auth.servarr.com/v1/mal_sonarr/renew";
        public const string ClientId = "402de1a90f8eb545f625739fa69ddb98";
```

src/NzbDrone.Core/ImportLists/MyAnimeList/MalImport.cs
```diff
@@ -0,0 +1,227 @@
...
+        {
+            if (action == "startOAuth")
+            {
+                // The workaround
+                // Have mal redirect, then make copy and paste the returned stuff into a text box for sonarr to use.
+                // Create those boxes in MalListSettings
+                _Codechallenge = GenCodeChallenge();
+                var request = new HttpRequestBuilder(OAuthUrl)
+                    .AddQueryParam("response_type", "code")
+                    .AddQueryParam("client_id", Settings.ClientId)
```
markus101, line 71: This shouldn't come from settings.

src/NzbDrone.Core/ImportLists/MyAnimeList/MalListSettings.cs
```diff
@@ -0,0 +1,64 @@
...
+        public MalListSettings()
+        {
+            BaseUrl = "https://api.myanimelist.net/v2";
+            ListStatus = MalAnimeStatus.Watching;
+        }
+
+        [FieldDefinition(0, Label = "List Status", Type = FieldType.Select, SelectOptions = typeof(MalAnimeStatus), HelpText = "Type of list status you're seeking to import from")]
+        public MalAnimeStatus ListStatus { get; set; }
+
+        [FieldDefinition(0, Label = "Client ID", Type = FieldType.Textbox)]
```
markus101, line 41: This something handled server side

src/NzbDrone.Core/ImportLists/MyAnimeList/MalListSettings.cs
```diff
@@ -0,0 +1,64 @@
...
+        public string ClientId { get; set; }
+
+        [FieldDefinition(0, Label = "Client Secret", Type = FieldType.Textbox)]
+        public string ClientSecret { get; set; }
```
markus101, line 45: This something handled server side

src/NzbDrone.Core/ImportLists/MyAnimeList/MalImport.cs
```diff
@@ -0,0 +1,227 @@
...
+            }
+            else if (action == "getOAuthToken")
+            {
+                var jsonResult = Json.Deserialize<MalAuthToken>(GetAuthToken(query["code"]));
+                return new
+                {
+                    accessToken = jsonResult.AccessToken,
+                    refreshToken = jsonResult.RefreshToken,
+                    expires = DateTime.UtcNow.AddSeconds(int.Parse(jsonResult.ExpiresIn))
+                    //expires = DateTime.UtcNow.AddSeconds(5)
```
markus101, line 90: Extraneous

markus101 (changes requested):
The problem with the OAuth is it requires a server side implementation, instead of trying to do it another way. With the information I provided in comments you should be able to follow the Simkl implementation.

User's do not need to create their own app on MAL when the redirect through Servarr is created correctly.

Also lots of code issues being caught by CI.

src/NzbDrone.Core/ImportLists/MyAnimeList/MalImport.cs
markus101: I setup an App for Sonarr on MAL and had it setup on the Servarr auth side of things.

jtnguyen45:
We are trying to implement a redirect similar to Simkl but we are having issues because MAL uses a [code challenge](https://myanimelist.net/blog.php?eid=835707#:~:text=Step%201%20%2D%20Generate%20a%20new%20Code%20Verifier%20and%20Code%20Challenge) for their OAuth. We are not sure how to adjust the code using Simkl's format and accounting for the code challenge, as we still get `Received oauth token was invalid.` error.

markus101:
Yeah, that's going to require some additional work to support, not yet sure how best to achieve that.

jtnguyen45:
Yeah it seems to be a bit tricky. Hopefully we find a way. As a heads up, @iceypotato and I are wrapping up our internship this Friday (8/18). We will not be as attentive in working on this issue but we hope to still keep contributing and working on it when are able to.

markus101:
Thanks for the initial work on this, finally had a chance to dig into the MAL OAuth and got that working, I've opened a new PR with it all implemented. https://github.com/Sonarr/Sonarr/pull/6654

PR #5859, opened 2023-07-30
Filter user issues from Sentry

Qstick:
#### Database Migration
NO

#### Description
Some more sentry filtering to clean things up


markus101 (approved):
Not sure if there are more you want to add, but these 3 look good.

PR #5866, opened 2023-08-01
New: Add root folder select to EditSeriesModal

### LINKED ISSUE #5544: A way to change root path inside a show

Every time i want to move a show i'm fixing, to another root path, or if i want to apply renaming to the folder name, i have to go back in main dashboard, search the show, use the multiselect to select it, then press the EDIT in the bottom and select the new or same root path, then apply

using the mass edit, but finding the show between thousands without using search (can't be used because it will open the show page, not filter the show list in dashboard), it's a pain

under the folder path inside a show edit window, i would like to have the same listbox as in the mass edit, to change the root path of that single show

A button to rename the show folder, not the episodes. Will let rename the folder as per config, but no help in changing root folder if needed.

### Anything else?

_No response_

ricci2511:
#### Database Migration
NO

#### Description
Simply added the root folder select to the edit series modal. Changing the root folder also changes the series path accordingly.

#### Todos
- [ ] Tests
- [ ] Wiki Updates

#### Screenshot (if UI related)
root-folder1
root-folder2

* Closes #5544

### DIFF
--- a/frontend/src/Series/Edit/EditSeriesModalConnector.js
+++ b/frontend/src/Series/Edit/EditSeriesModalConnector.js
@@ -2,13 +2,21 @@ import PropTypes from 'prop-types';
 import React, { Component } from 'react';
 import { connect } from 'react-redux';
 import { clearPendingChanges } from 'Store/Actions/baseActions';
+import { fetchRootFolders } from 'Store/Actions/rootFolderActions';
 import EditSeriesModal from './EditSeriesModal';
 
 const mapDispatchToProps = {
-  clearPendingChanges
+  clearPendingChanges,
+  fetchRootFolders
 };
 
 class EditSeriesModalConnector extends Component {
+
+  //
+  // Lifecycle
+  componentDidMount() {
+    this.props.fetchRootFolders();
+  }
 
   //
   // Listeners
@@ -34,7 +42,8 @@ class EditSeriesModalConnector extends Component {
 EditSeriesModalConnector.propTypes = {
   ...EditSeriesModal.propTypes,
   onModalClose: PropTypes.func.isRequired,
-  clearPendingChanges: PropTypes.func.isRequired
+  clearPendingChanges: PropTypes.func.isRequired,
+  fetchRootFolders: PropTypes.func.isRequired
 };
 
 export default connect(undefined, mapDispatchToProps)(EditSeriesModalConnector);
--- a/frontend/src/Series/Edit/EditSeriesModalContent.js
+++ b/frontend/src/Series/Edit/EditSeriesModalContent.js
@@ -77,6 +77,7 @@ class EditSeriesModalContent extends Component {
       qualityProfileId,
       seriesType,
       path,
+      rootFolderPath,
       tags
     } = item;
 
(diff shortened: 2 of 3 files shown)


frontend/src/Series/Edit/EditSeriesModalConnector.js
```diff
@@ -2,14 +2,22 @@ import PropTypes from 'prop-types';
...
-  clearPendingChanges
+  clearPendingChanges,
+  fetchRootFolders
 };
 
 class EditSeriesModalConnector extends Component {
 
+  //
+  // Lifecycle
+  componentDidMount() {
```
markus101, line 17: New line after comment

frontend/src/Series/Edit/EditSeriesModalContent.js
```diff
@@ -148,6 +149,23 @@ class EditSeriesModalContent extends Component {
...
+              <FormLabel>{translate('Root Folder')}</FormLabel>
+
+              <FormInputGroup
+                type={inputTypes.ROOT_FOLDER_SELECT}
+                name="rootFolderPath"
+                {...rootFolderPath}
+                includeNoChange={true}
+                includeNoChangeDisabled={false}
+                selectedValueOptions={{ includeFreeSpace: false }}
+                helpText={translate(
```
markus101, line 162: This translation needs to be extracted to a key

frontend/src/Series/Edit/EditSeriesModalContentConnector.js
```diff
@@ -80,7 +84,20 @@ class EditSeriesModalContentConnector extends Component {
...
+      isWindows,
+      title,
+      item: { path }
+    } = this.props;
+
+    dispatchSetSeriesValue({ name, value });
+
+    // Also update the path if the root folder path changes
+    if (name === 'rootFolderPath' && value !== 'noChange' && path.value.indexOf(value) === -1) {
+      const newPath = combinePath(isWindows, value, [title]);
```
markus101, line 98: Using the series title isn't sufficient, this path needs to be generated on the backend to take the format into account.

markus101 (changes requested):
The fact that you can change both the root folder and the path separately at the same time is going to cause some confusion and likely some issues, especially with how the new path is being updated.

Instead of a drop down a modal that either communicates with the backend and gets the new path given the chosen root folder and series or uses a pre-calculated value, like how it's done in `SeriesLookupController`, but the latter option could lead to a stale value.

ricci2511:
Yeah, your concerns make sense. Just one thing, so if the root folder dropdown is ditched in favor of the modal approach how would the expected "user flow" be? Can you maybe provide a really high level example of the SeriesEditModal UI/UX? Just to make sure I understood correctly.

markus101:
I think something like `Update Monitoring` when using series selection.

image

When the modal is opened the path for the series folder could be constructed, (precomputing could be useful for other things, but it's not a free operation and often not needed).

Once the root folder is chosen we could show the user the new path that will be used like we do in add series:
image

markus101:
Closing, but we can reopen once updated.

PR #5872, opened 2023-08-03
Remove Test Parsing button from Series Index

mynameisbogdan:
#### Database Migration
NO

#### Description
Feels very out of place in Series Index.


stevietv:
I don't mind it there, nice and accessible. Otherwise you then only have it in custom formats.

PR #5873, opened 2023-08-03
Update API docs page to clarify v3/v4 independence

stevietv:
#### Database Migration
NO

#### Description
Expand the description on the swagger UI page to state that the API docs are for API v3 for the v4 frontend

### DIFF
--- a/src/NzbDrone.Host/Startup.cs
+++ b/src/NzbDrone.Host/Startup.cs
@@ -99,7 +99,7 @@ namespace NzbDrone.Host
                 {
                     Version = "3.0.0",
                     Title = "Sonarr",
-                    Description = "Sonarr API docs",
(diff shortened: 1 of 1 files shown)


markus101:
I don't think this makes sense to add, the API version is the same, the fact that v4 has better handling for an enum value for the history event type doesn't invalidate the docs.

stevietv:
> I don't think this makes sense to add, the API version is the same, the fact that v4 has better handling for an enum value for the history event type doesn't invalidate the docs.

What about then a simple statement - 'the v3 API docs apply to both v3 and v4 versions of sonarr. Some functionality may only be available in v4 of Sonarr app'

markus101:
Yeah, I think that a lot better

stevietv:
Done :)

markus101 (approved):

PR #5887, opened 2023-08-07
Show warning when using the docker update mechanism

mynameisbogdan:
#### Database Migration
NO

#### Description
image

Also
- Prevent NullRef in IsPathValid for null paths
- Replace legacy master branch with main

### DIFF
--- a/frontend/src/Settings/General/UpdateSettings.js
+++ b/frontend/src/Settings/General/UpdateSettings.js
@@ -9,7 +9,7 @@ import titleCase from 'Utilities/String/titleCase';
 import translate from 'Utilities/String/translate';
 
 const branchValues = [
-  'master',
+  'main',
   'develop'
 ];
 
@@ -83,6 +83,7 @@ function UpdateSettings(props) {
                 type={inputTypes.CHECK}
                 name="updateAutomatically"
                 helpText={translate('UpdateAutomaticallyHelpText')}
+                helpTextWarning={updateMechanism.value === 'docker' ? translate('AutomaticUpdatesDisabledDocker', { appName: 'Sonarr' }) : undefined}
                 onChange={onInputChange}
                 {...updateAutomatically}
               />
--- a/src/NzbDrone.Core/Localization/Core/en.json
+++ b/src/NzbDrone.Core/Localization/Core/en.json
@@ -95,6 +95,7 @@
   "Automatic": "Automatic",
   "AutomaticAdd": "Automatic Add",
   "AutomaticSearch": "Automatic Search",
+  "AutomaticUpdatesDisabledDocker": "Automatic updates are not directly supported when using the Docker update mechanism. You will need to update the container image outside of {appName} or use a script",
   "Backup": "Backup",
   "BackupFolderHelpText": "Relative paths will be under Sonarr's AppData directory",
   "BackupIntervalHelpText": "Interval between automatic backups",
--- a/frontend/src/System/Updates/Updates.js
+++ b/frontend/src/System/Updates/Updates.js
@@ -111,7 +111,6 @@ class Updates extends Component {
                   name={icons.CHECK_CIRCLE}
                   size={30}
                 />
-
                 <div className={styles.message}>
                   {translate('OnLatestVersion')}
                 </div>
@@ -149,8 +148,8 @@ class Updates extends Component {
                           </div>
 
                           {
-                            update.branch === 'master' ?
-                              null:
+                            update.branch === 'main' ?
+                              null :
                               <Label
                                 className={styles.label}
                               >
--- a/src/NzbDrone.Common/Extensions/PathExtensions.cs
+++ b/src/NzbDrone.Common/Extensions/PathExtensions.cs
@@ -149,7 +149,7 @@ namespace NzbDrone.Common.Extensions
 
         public static bool IsPathValid(this string path, PathValidationType validationType)
         {
-            if (path.ContainsInvalidPathChars() || string.IsNullOrWhiteSpace(path))
+            if (string.IsNullOrWhiteSpace(path) || path.ContainsInvalidPathChars())
             {
                 return false;
             }
--- a/src/Sonarr.Api.V3/Config/HostConfigController.cs
+++ b/src/Sonarr.Api.V3/Config/HostConfigController.cs
@@ -58,7 +58,7 @@ namespace Sonarr.Api.V3.Config
                 .Must((resource, path) => IsValidSslCertificate(resource)).WithMessage("Invalid SSL certificate file or password")
                 .When(c => c.EnableSsl);
 
-            SharedValidator.RuleFor(c => c.Branch).NotEmpty().WithMessage("Branch name is required, 'master' is the default");
+            SharedValidator.RuleFor(c => c.Branch).NotEmpty().WithMessage("Branch name is required, 'main' is the default");
             SharedValidator.RuleFor(c => c.UpdateScriptPath).IsValidPath().When(c => c.UpdateMechanism == UpdateMechanism.Script);
 
             SharedValidator.RuleFor(c => c.BackupFolder).IsValidPath().When(c => Path.IsPathRooted(c.BackupFolder));


stevietv:
does this actually work in lidarr? 

image

markus101:
Do docker images have the ability to tell Sonarr which branch to display updates for? If not then that should still be configurable.

thezoggy:
just fyi in sab we actually had to change from using this docker detection because it doesnt work for everything anymore

now we just look to see if `.dockerenv` exists instead. which works for older+newer docker (Ubuntu 18.04 and 22.04).

bakerboy448:
agree with taloth; the detect docker fix is good.  but docker / other third party maintainers maintainers should be using package_info appropriately to disable it and tell users how to update (with that said - some additional documentation on package_info would likely be useful)

bakerboy448:
Run a script to run compose update or any various scripts that would update the container(s)

Taloth:
> > But... disabling of the auto updater goes via package_info. It doesn't need to be hardcoded and users should be able to select Script option if they want to.
> 
> Also it seems like if it's docker it doesn't allow any type of updates, so this PR only updates the UI to reflect the current state.
> 
> https://github.com/Sonarr/Sonarr/blob/a645681cc8a3683876270af1d1a7b052a86e87e1/src/NzbDrone.Core/Update/InstallUpdateService.cs#L240-L244

That should not have been added imo. coz a container dev can specify the desired behavior in package_info. There's no reason to block it just because it's docker. And thus I'd rather revert that, even though it was added a year ago, possible as pull from Radarr. I'll talk to markus about it.

I agree package_info should be documented better, but the two major image maintainers know and use it.

markus101:
I agree with Taloth here, seems like we shouldn't have added the disabling of updating for docker completely in favour of relying on `package_info` instead.

frontend/src/Settings/General/UpdateSettings.js
```diff
@@ -83,8 +83,10 @@ function UpdateSettings(props) {
                 type={inputTypes.CHECK}
                 name="updateAutomatically"
                 helpText={translate('UpdateAutomaticallyHelpText')}
+                helpTextWarning={updateMechanism.value === 'docker' ? translate('AutomaticUpdatesAreDisabledWhenUsingTheDockerUpdateMechanism') : undefined}
```
markus101, line 86: This translation key is too verbose, something like `AutomaticUpdatedDisabledDocker` would be more appropriate, I've already discussed downstream syncing up with Q.

src/NzbDrone.Core/Localization/Core/en.json
```diff
@@ -83,6 +83,7 @@
   "Automatic": "Automatic",
   "AutomaticAdd": "Automatic Add",
   "AutomaticSearch": "Automatic Search",
+  "AutomaticUpdatesAreDisabledWhenUsingTheDockerUpdateMechanism": "Automatic updates are disabled when using the Docker update mechanism. Update the container image instead. Contact your container maintainer for additional details.",
```
markus101, line 86: > Contact your container maintainer for additional details.

What's the value in this? Seems like unnecessary noise.

markus101 (changes requested):

frontend/src/Settings/General/UpdateSettings.js
markus101: We don't want to use the English Translation as a key, so we shouldn't convert the English translation to a key.

src/NzbDrone.Core/Localization/Core/en.json
bakerboy448: userproof else they'd be coming to the sonarr discord?

markus101: Potentially, but now we're sending them to the maintainer's discord when it's not really something they need to deal with either. We already have docker disabled for updates and don't have many people asking how to update it.

```suggestion
  "AutomaticUpdatesAreDisabledWhenUsingTheDockerUpdateMechanism": "Automatic updates are disabled when using the Docker update mechanism. You will need to update the container image outside of Sonarr.",
```

frontend/src/Settings/General/UpdateSettings.js
```diff
@@ -83,8 +83,10 @@ function UpdateSettings(props) {
...
                 onChange={onInputChange}
                 {...updateAutomatically}
+                isDisabled={updateMechanism.value === 'docker'}
```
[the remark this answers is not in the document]
markus101: I think the latter

src/NzbDrone.Core/Localization/Core/en.json
```diff
@@ -83,6 +83,7 @@
   "Automatic": "Automatic",
   "AutomaticAdd": "Automatic Add",
   "AutomaticSearch": "Automatic Search",
+  "AutomaticUpdatesDisabledDocker": "Automatic updates are disabled when using the Docker update mechanism. You will need to update the container image outside of Sonarr.",
```
markus101, line 86: In an effort to make this re-usable:
```suggestion
  "AutomaticUpdatesDisabledDocker": "Automatic updates are disabled when using the Docker update mechanism. You will need to update the container image outside of {appName}.",
```

Then pass `{ appName: 'Sonarr' }` into `translate`

markus101 (changes requested):

markus101 (approved):

frontend/src/Settings/General/UpdateSettings.js
```diff
@@ -83,8 +85,10 @@ function UpdateSettings(props) {
                 type={inputTypes.CHECK}
                 name="updateAutomatically"
                 helpText={translate('UpdateAutomaticallyHelpText')}
+                helpTextWarning={updateMechanism.value === 'docker' ? translate('AutomaticUpdatesDisabledDocker', { appName: 'Sonarr' }) : undefined}
                 onChange={onInputChange}
                 {...updateAutomatically}
+                isDisabled={!canUpdateAutomatically}
```
markus101, line 91: Actually, as mentioned by @bakerboy448 script updating is technically possible, so this is wholly accurate and would disable that ability.

src/NzbDrone.Core/Localization/Core/en.json
```diff
@@ -95,6 +95,7 @@
   "Automatic": "Automatic",
   "AutomaticAdd": "Automatic Add",
   "AutomaticSearch": "Automatic Search",
+  "AutomaticUpdatesDisabledDocker": "Automatic updates are disabled when using the Docker update mechanism. You will need to update the container image outside of {appName}.",
```
markus101, line 98:
```suggestion
  "AutomaticUpdatesDisabledDocker": "Automatic updates are not directly supported when using the Docker update mechanism. You will need to update the container image outside of {appName} or use a script",
```

markus101 (changes requested):
On further review I think disabling automatic updates completely accurate as it could be possible with a script and this would gatekeep that ability.

frontend/src/Settings/General/UpdateSettings.js
markus101: If the update mechanism is `docker`, the user either needs to update via docker (outside of Sonarr) or as baker mentioned before, they could use a script to update docker through compose or some other tooling.

Disabling this means they would be unable to use a script to do the updating automatically.

src/NzbDrone.Core/Update/ConfigureUpdateMechanism.cs
```diff
@@ -65,6 +65,12 @@ private void ChangeUpdateMechanism(UpdateMechanism updateMechanism)
             {
                 [nameof(_configFileProvider.UpdateMechanism)] = updateMechanism
             };
+
+            if (updateMechanism >= UpdateMechanism.External)
+            {
+                config.Add(nameof(_configFileProvider.UpdateAutomatically), false);
+            }
```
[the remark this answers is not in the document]
markus101: Yeah I think so, it'll keep turning it off if they've enabled it, but if we can do it when the value hasn't yet been written to the file then that would be fine.

frontend/src/Settings/General/UpdateSettings.js
stevietv: > How can one update with docker compose from inside of a running container?

for example, on identifying an update you could call a script that triggers another container (watchtower) to update this one.

markus101 (approved):

PR #5895, opened 2023-08-08
Fixed: Detect Docker when using control group v2

mynameisbogdan:
#### Database Migration
NO

#### Description
As per title.


markus101 (changes requested):
Should include the sentry change as well.

markus101 (approved):

PR #5904, opened 2023-08-10
Translate App Frontend (updated and connection lost modals)

stevietv:
#### Database Migration
NO

#### Description
add translations where required to the app pages.

*


frontend/src/App/AppUpdatedModalContent.css
```diff
@@ -1,6 +1,7 @@
-.version {
+code {
```
markus101, line 1: Does this scope `code`  styling to this component only? We might need to tweak the Markdown parser to allow for passing in a class for the code blocks.

Styling specific elements like this is less efficient and could result in it affecting other usages.

frontend/src/App/AppUpdatedModalContent.css
```diff
@@ -1,6 +1,7 @@
...
   margin: 0 3px;
   font-weight: bold;
+  font-family: "Roboto", "open sans", "Helvetica Neue", "Helvetica", "Arial", sans-serif;
```
markus101, line 4: Should be able to use the `defaultFontFamily` CSS variable

src/NzbDrone.Core/Localization/Core/en.json
```diff
@@ -160,6 +162,9 @@
   "Connect": "Connect",
   "ConnectSettings": "Connect Settings",
   "ConnectSettingsSummary": "Notifications, connections to media servers/players, and custom scripts",
+  "ConnectionLost": "Connection Lost",
+  "ConnectionLostReconnect": "{appName}  will try to connect automatically, or you can click reload below.",
```
markus101, line 166:
```suggestion
  "ConnectionLostReconnect": "{appName} will try to connect automatically, or you can click reload below.",
```

markus101 (changes requested):

frontend/src/App/AppUpdatedModalContent.css
stevietv: yeah it applies to only this block. I can work on the InlineMarkdown component then :)

stevietv: ok, tweaked InlineMarkdown to allow for a different class for the <code> part of the data passed to it.

frontend/src/Components/Markdown/InlineMarkdown.js
```diff
@@ -47,7 +48,11 @@ class InlineMarkdown extends Component {
           markdownBlocks.push(data.substr(endIndex, match.index - endIndex));
         }
 
-        markdownBlocks.push(<code key={`code-${match.index}`}>{match[0].substring(1, match[0].length - 1)}</code>);
+        if (blockClassName) {
+          markdownBlocks.push(<code key={`code-${match.index}`} className={blockClassName}>{match[0].substring(1, match[0].length - 1)}</code>);
+        } else {
+          markdownBlocks.push(<code key={`code-${match.index}`}>{match[0].substring(1, match[0].length - 1)}</code>);
+        }
```
markus101, line 55:
```suggestion
        markdownBlocks.push(<code key={`code-${match.index}`} className={blockClassName ?? ''}>{match[0].substring(1, match[0].length - 1)}</code>);
```

I think we could do that, but if that's not working I think this would be a little DRYer.

```
        const codeBlockKey = `code-${match.index}`;
        const codeBlockContent = match[0].substring(1, match[0].length - 1);
        
        if (blockClassName) {
          markdownBlocks.push(<code key={codeBlockKey} className={blockClassName}>{codeBlockContent}</code>);
        } else {
          markdownBlocks.push(<code key={codeBlockKey}>{codeBlockContent}</code>);
        }
```

markus101 (changes requested):
Just one suggestion then looks good!

frontend/src/Components/Markdown/InlineMarkdown.js
stevietv: the first suggestion results in <code class> being created, but null instead of '' works

markus101 (approved):
Thanks for the fixes

PR #5907, opened 2023-08-11
Fixed: Use preset or implementation name as default name when adding providers

mynameisbogdan:
#### Database Migration
NO

#### Description
Use the preset name or implementation name as default name, and fix issues with `undefined` where the implementation name should be.


frontend/src/Components/Form/RootFolderSelectInput.js
```diff
@@ -74,6 +74,7 @@ class RootFolderSelectInput extends Component {
     return (
       <div>
         <EnhancedSelectInput
+          value=""
```
markus101, line 77: This should be set either by the caller or a default prop value.

frontend/src/Settings/ImportLists/ImportLists/EditImportListModalContent.js
```diff
@@ -58,7 +59,7 @@ function EditImportListModalContent(props) {
   return (
     <ModalContent onModalClose={onModalClose}>
       <ModalHeader>
-        {id ? translate('EditImportList') : translate('AddImportList')}
+        {`${id ? translate('EditImportList') : translate('AddImportList')} - ${implementationName}`}
```
markus101, line 62: For metadata we use `Edit {metadataType} Metadata`, should that be changed to a similar format?

I think regardless the implementation name should be part of the translation so the full context of how it's used is presented in the translation in the event that affects the translation in other language.

markus101 (changes requested):

frontend/src/Settings/ImportLists/ImportLists/EditImportListModalContent.js
markus101: It gives additional information on how the translation is being used, including the secondary title being appended, which could affect how it's supposed to be translated. In a RTL language it's also necessary because the display would be wrong, whether that's more wrong than mixing a proper English name in RTL, I'm not sure.

markus101: Yeah, having the implementation name at the end like that still, just wanted to make sure it was part of the translation.

markus101 (approved):
Thanks for the fixes, I'll land after CI stops being dumb.

frontend/src/Settings/CustomFormats/CustomFormats/Specifications/EditSpecificationModalContent.js
```diff
@@ -41,7 +41,7 @@ function EditSpecificationModalContent(props) {
   return (
     <ModalContent onModalClose={onCancelPress}>
       <ModalHeader>
-        {`${id ? translate('EditCondition') : translate('AddCondition')} - ${implementationName}`}
+        {id ? translate('EditConditionImplementation', { implementation: implementationName }) : translate('AddConditionImplementation', { implementation: implementationName })}
```
markus101, line 44: I think changing these all to `implementationName` would be good, it'll shorten down every call to translate and is a bit more explicit as to what "implementation" is.

```suggestion
        {id ? translate('EditConditionImplementation', { implementationName }) : translate('AddConditionImplementation', { implementationName })}
```

src/NzbDrone.Core/Localization/Core/en.json
```diff
@@ -9,18 +9,21 @@
   "AddAutoTagError": "Unable to add a new auto tag, please try again.",
   "AddCondition": "Add Condition",
   "AddConditionError": "Unable to add a new condition, please try again.",
-  "AddConnection": "Add Connection",
+  "AddConditionImplementation": "Add Condition - {implementation}",
```
markus101, line 12: This should be connection, not condition.

src/NzbDrone.Core/Localization/Core/en.json
```diff
@@ -34,6 +37,7 @@
   "AddNewSeriesSearchForMissingEpisodes": "Start search for missing episodes",
   "AddNotification": "Add Notification",
   "AddNotificationError": "Unable to add a new notification, please try again.",
+  "AddNotificationImplementation": "Add Notification - {implementation}",
```
markus101, line 40: `Connection` instead of `Notification`

src/NzbDrone.Core/Localization/Core/en.json
```diff
@@ -290,16 +294,17 @@
...
+  "EditDownloadClientImplementation": "Edit Download Client - {implementation}",
   "EditGroups": "Edit Groups",
-  "EditImportList": "Edit Import List",
   "EditImportListExclusion": "Edit Import List Exclusion",
-  "EditIndexer": "Edit Indexer",
+  "EditImportListImplementation": "Edit Import List - {implementation}",
+  "EditIndexerImplementation": "Edit Indexer - {implementation}",
   "EditListExclusion": "Edit List Exclusion",
   "EditMetadata": "Edit {metadataType} Metadata",
+  "EditNotificationImplementation": "Edit Notification - {implementation}",
```
markus101, line 307: `Connection` instead of `Notification`

markus101 (changes requested):
Thanks for updating the rest, just a little final fixes

src/NzbDrone.Core/Localization/Core/en.json
markus101: Yeah, but I think `AddNotificationError` should be renamed changed to reference connections since that is the external facing name and `notifications` for the file/folder structure is from when it was really just notifications.

markus101 (approved):
Thanks!

PR #5909, opened 2023-08-11
Fixed: Use `#` for special chars in TitleFirstCharacter

mynameisbogdan:
#### Database Migration
NO

#### Description
https://github.com/Radarr/Radarr/issues/5642

Open to any other wildcard character.

### DIFF
--- a/src/NzbDrone.Core/Organizer/FileNameBuilder.cs
+++ b/src/NzbDrone.Core/Organizer/FileNameBuilder.cs
@@ -452,7 +452,7 @@ namespace NzbDrone.Core.Organizer
             tokenHandlers["{Series TitleWithoutYear}"] = m => TitleWithoutYear(series.Title);
             tokenHandlers["{Series TitleTheYear}"] = m => TitleYear(TitleThe(series.Title), series.Year);
             tokenHandlers["{Series TitleTheWithoutYear}"] = m => TitleWithoutYear(TitleThe(series.Title));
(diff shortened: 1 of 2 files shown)


src/NzbDrone.Core/Organizer/FileNameBuilder.cs
```diff
@@ -1150,6 +1150,13 @@ private static string CleanFileName(string name, NamingConfig namingConfig)
 
             return result.TrimStart(' ', '.').TrimEnd(' ');
         }
+
+        private static string GetFirstCharacter(string title)
+        {
+            var firstCharacter = title[..1].FirstCharToUpper();
+
+            return char.IsLetterOrDigit(firstCharacter, 0) ? firstCharacter : "#";
```
markus101, line 1158: `#` feels too similar to digits for me and I'd expect titles that start with `0-9` to be in that folder, not sure which character would be better though.

markus101 (changes requested):
I disagree that this is a fix, but more of a feature that would need to be considered, we don't sort `0-9` into a `#` subfolder or `A-Z` into a subfolder for all letters.

For sorting there are a number of characters that are typically ignored: https://www.niso.org/sites/default/files/2017-08/tr03.pdf which is effectively what's being done with using the first letter as a subfolder.

That Radarr GHI also asked about titles that start with numeric numbers to be grouped with special characters.

PR #5911, opened 2023-08-11
revert translations on defaultProps for QueueStatus.js

stevietv:
#### Database Migration
NO

#### Description
revert translations on defaultProps for QueueStatus.js
This wasn't required

*

### DIFF
--- a/frontend/src/Activity/Queue/QueueStatus.js
+++ b/frontend/src/Activity/Queue/QueueStatus.js
@@ -154,8 +154,8 @@ QueueStatus.propTypes = {
 };
 
 QueueStatus.defaultProps = {
-  trackedDownloadStatus: translate('Ok'),
-  trackedDownloadState: translate('Downloading'),
+  trackedDownloadStatus: 'Ok',
+  trackedDownloadState: 'Downloading',
   canFlip: false
 };
 
--- a/frontend/src/Activity/Queue/QueueStatus.js
+++ b/frontend/src/Activity/Queue/QueueStatus.js
@@ -154,8 +154,8 @@ QueueStatus.propTypes = {
 };
 
 QueueStatus.defaultProps = {
-  trackedDownloadStatus: 'Ok',
-  trackedDownloadState: 'Downloading',
+  trackedDownloadStatus: 'ok',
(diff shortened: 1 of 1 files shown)


frontend/src/Activity/Queue/QueueStatus.js
```diff
@@ -154,8 +154,8 @@ QueueStatus.propTypes = {
 };
 
 QueueStatus.defaultProps = {
-  trackedDownloadStatus: translate('Ok'),
-  trackedDownloadState: translate('Downloading'),
+  trackedDownloadStatus: 'Ok',
```
markus101, line 157: These still should be translated.

This might be a problem for swapping languages while the app is running, but off hand I'm not sure how default props are treated and whether they get evaluated every time they're called on just the first time.

markus101 (changes requested):
This revert isn't necessary, but we may have to change how `defaultProps` are handled for hotswapping translations,. but we can figure that out later.

frontend/src/Activity/Queue/QueueStatus.js
stevietv: OK, i wasn't sure if this one was needed, since its doing a string comparison on lines 51/52 based on what I assume is a query from of the status DL client.
it also seems like trackedDownloadState isn't used here at all?

frontend/src/Activity/Queue/QueueStatus.js
```diff
@@ -154,8 +154,8 @@ QueueStatus.propTypes = {
...
+  trackedDownloadState: 'Downloading',
```
markus101, line 158: Looks like these should be

```suggestion
  trackedDownloadStatus: 'ok',
  trackedDownloadState: 'downloading',
```

frontend/src/Activity/Queue/QueueStatus.js
markus101: Ahh, they're not strings that should be translated, but values from the API.

markus101 (changes requested):

markus101 (approved):

PR #5916, opened 2023-08-13
Add translations to frontend/InteractiveImport

stevietv:
#### Database Migration
NO

#### Description
adds translations to Interactive Import.

**special note**: I had to allow `token` to be `undefined` in `translate.js`, otherwise `SelectEpisodeModalContent.tsx` would give a type error as modalTitle is optional.

*


frontend/src/Utilities/String/translate.ts
```diff
@@ -25,7 +25,7 @@ export async function fetchTranslations(): Promise<boolean> {
 
 export default function translate(
   key: string,
-  tokens?: Record<string, string | number | boolean>
+  tokens?: Record<string, string | number | boolean | undefined>
```
markus101, line 28: I think overall it'd be better to have the caller set a default value if it's undefined, so it can properly deal with the differences, otherwise the translation is going to look off.

markus101 (changes requested):

frontend/src/InteractiveImport/Episode/SelectEpisodeModalContent.tsx
```diff
@@ -197,21 +197,23 @@ function SelectEpisodeModalContent(props: SelectEpisodeModalContentProps) {
...
-        : `${selectedCount} selected file`;
+        ? translate('CountSelectedFiles', { selectedCount })
+        : translate('CountSelectedFile', { selectedCount });
   }
 
   return (
     <ModalContent onModalClose={onModalClose}>
-      <ModalHeader>{modalTitle} - Select Episode(s)</ModalHeader>
+      <ModalHeader>
+        {translate('SelectEpisodesModalTitle', { modalTitle })}
```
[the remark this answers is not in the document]
stevietv: done, made this prop no longer optional and reverted translate changes

frontend/src/Utilities/String/translate.ts
stevietv: reverted change

markus101 (approved):
Thanks!

PR #5930, opened 2023-08-16
No error uneccessary custom import rename

JeWe37:
#### Database Migration
NO

#### Description
Minor oversight for custom import scripts. When requesting a rename that is not required, an error is thrown because the file would be moved to the location it is already in. In that case we can simply catch the error and ignore it.

*

### DIFF
--- a/src/NzbDrone.Core/MediaFiles/EpisodeFileMovingService.cs
+++ b/src/NzbDrone.Core/MediaFiles/EpisodeFileMovingService.cs
@@ -121,16 +121,10 @@ namespace NzbDrone.Core.MediaFiles
                 throw new SameFilenameException("File not moved, source and destination are the same", episodeFilePath);
             }
 
-            var transfer = true;
-
             episodeFile.RelativePath = series.Path.GetRelativePath(destinationFilePath);
 
-            if (localEpisode is not null)
+            if (localEpisode is not null && _scriptImportDecider.TryImport(episodeFilePath, destinationFilePath, localEpisode, episodeFile, mode) is var scriptImportDecision && scriptImportDecision != ScriptImportDecision.DeferMove)
             {
-                var scriptImportDecision = _scriptImportDecider.TryImport(episodeFilePath, destinationFilePath, localEpisode, episodeFile, mode);
-
-                transfer = scriptImportDecision == ScriptImportDecision.DeferMove;
-
                 if (scriptImportDecision == ScriptImportDecision.RenameRequested)
                 {
                     try
@@ -143,8 +137,7 @@ namespace NzbDrone.Core.MediaFiles
                     }
                 }
             }
-
-            if (transfer)
+            else
             {
                 _diskTransferService.TransferFile(episodeFilePath, destinationFilePath, mode);
             }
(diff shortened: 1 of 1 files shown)


src/NzbDrone.Core/MediaFiles/EpisodeFileMovingService.cs
```diff
@@ -134,7 +134,15 @@ private EpisodeFile TransferFile(EpisodeFile episodeFile, Series series, List<Ep
                     case ScriptImportDecision.DeferMove:
                         break;
                     case ScriptImportDecision.RenameRequested:
-                        MoveEpisodeFile(episodeFile, series, episodeFile.Episodes);
+                        try
```
markus101, line 137: This switch case should be converted to an if/else block instead of moving more logic inside of it.

markus101 (changes requested):
What's the case where this occurs? Seems like it should only happen when it's an existing file being renamed, which I'd expect to be caught by `RenameEpisodeFileService`.

src/NzbDrone.Core/MediaFiles/EpisodeFileMovingService.cs
JeWe37: Done

JeWe37:
Not quite, this line exists for the following purpose: Consider the following situation: Renaming is turned on, but due to a remux done by a custom import script the name of the file changes, say because the audio was reordered, changing the primary audio. Then the name needs to be changed again, after the actual copy from the import location to the library has been completed(by the custom import script). This means the move here is happening _within_ the library. 

Now consider the case where a rename was requested, but it isn't actually needed. Say the script thought the name might change, but it actually didn't. In that case, we should just not move the file silently. But currently an error is being raised, failing the import from Sonarr's perspective, because when we request a rename we always try to move the file and the method doesn't account for that. So we just catch the error here to ignore it, as no action was necessary. 

This is the best way to do this rather than checking beforehand because the actual filename to move to is computed within the method that moves the file in this case.

src/NzbDrone.Core/MediaFiles/EpisodeFileMovingService.cs
```diff
@@ -129,17 +129,18 @@ private EpisodeFile TransferFile(EpisodeFile episodeFile, Series series, List<Ep
             {
                 var scriptImportDecision = _scriptImportDecider.TryImport(episodeFilePath, destinationFilePath, localEpisode, episodeFile, mode);
 
-                switch (scriptImportDecision)
+                transfer &= scriptImportDecision == ScriptImportDecision.DeferMove;
```
markus101, line 132: Since `transfer` is only used with `DeferMove` I think this could be simplified to:

```
if (scriptImportDecision == ScriptImportDecision.DeferMove)
{
    _diskTransferService.TransferFile(episodeFilePath, destinationFilePath, mode);
}
else if (scriptImportDecision == ScriptImportDecision.RenameRequested)
{
    try
    {
        MoveEpisodeFile(episodeFile, series, episodeFile.Episodes);
    }
    catch (SameFilenameException)
    {
        _logger.Debug("No rename was required. File already exists at destination.");
    }
}
```

markus101 (changes requested):
Thanks for the explanation, that makes sense.

src/NzbDrone.Core/MediaFiles/EpisodeFileMovingService.cs
JeWe37: Well not quite it's also used when local episode is null. I didn't want to repeat the transfer call. That's why it's structured like this.

JeWe37: Actually thought up a solution that should be ok with pattern matching. This should be ok now.

markus101: :( I was completely ignoring that case, but this change looks reasonable.

markus101 (approved):

PR #5934, opened 2023-08-18
Translate Frontend Organize + Rename modal

stevietv:
#### Database Migration
NO

#### Description
Translate Frontend Organize + Rename modal

*


frontend/src/Organize/OrganizePreviewModalContent.js
```diff
@@ -93,7 +96,7 @@ class OrganizePreviewModalContent extends Component {
     return (
       <ModalContent onModalClose={onModalClose}>
         <ModalHeader>
-          Organize & Rename {seasonNumber != null && <SeasonNumber seasonNumber={seasonNumber} />}
+          {translate('OrganizeModalHeader', { season: seasonNumber != null && ReactDOMServer.renderToString(<SeasonNumber seasonNumber={seasonNumber} />) || '' })}
```
markus101, line 99: I don't think we should be using `ReactDOMServer` on the client side, but given how simple the `SeasonNumber` component is we can just replace it with a function. Looks like we use it in a few spots, but use it the same way in all.

markus101 (changes requested):

frontend/src/Organize/OrganizePreviewModalContent.js
stevietv: replaced the SeasonNumber component with a seasonFormat function. wasn't too sure on the name of it!

frontend/src/Season/seasonFormat.ts
```diff
@@ -0,0 +1,14 @@
+import translate from 'Utilities/String/translate';
+
+export default function seasonFormat(seasonNumber: number, separator?: string) {
+  separator ??= '- ';
+  if (seasonNumber === 0) {
+    return `${separator}${translate('Specials')}`;
+  }
```
markus101, line 7: Should be able to set the default value like this:

```suggestion
export default function seasonFormat(seasonNumber: number, separator?: string = '- ') {

  if (seasonNumber === 0) {
    return `${separator}${translate('Specials')}`;
  }
```

Though given the super limited use, should the separator just be dropped and handled in the translation? That would be a bit more contextual and more along the lines of what we do for modal titles for providers.

markus101: Maybe `formatSeason` or `getSeasonName`?

stevietv: I agree, getting rid of the separator is better, otherwise might screw up translations with different orders.

stevietv: only issue I face now is the translation for this modal header:
`  "OrganizeModalHeader": "Organize & Rename {season}",`
`  {translate('OrganizeModalHeader', { season: seasonNumber != null && formatSeason(seasonNumber) || '' })}`

if I add the ` - ` in the translation itself then it shows always, even when there is no seasonNumber

frontend/src/Organize/OrganizePreviewModalContent.js
```diff
@@ -93,7 +95,10 @@ class OrganizePreviewModalContent extends Component {
     return (
       <ModalContent onModalClose={onModalClose}>
         <ModalHeader>
-          Organize & Rename {seasonNumber != null && <SeasonNumber seasonNumber={seasonNumber} />}
+          { seasonNumber ?
+            translate('OrganizeModalHeaderSeason', { season: formatSeason(seasonNumber) }) :
+            translate('OrganizeModalHeader')
+          }
```
markus101, line 101: I think this'll need to be:

```suggestion
          { seasonNumber == null ?
            translate('OrganizeModalHeader') :
            translate('OrganizeModalHeaderSeason', { season: formatSeason(seasonNumber) })
          }
```

otherwise specials are also skipped, but I like the approach of splitting them to different translations.

markus101 (changes requested):
Just one thing to make sure specials don't regress, but looks solid!

markus101 (approved):

PR #5935, opened 2023-08-18
Translate Frontend Parse modal

stevietv:
#### Database Migration
NO

#### Description
Translate Frontend Parse modal further than it already was

*


frontend/src/Parse/ParseResult.tsx
```diff
@@ -228,7 +228,13 @@ function ParseResult(props: ParseResultProps) {
 
         <ParseResultItem
           title={translate('CustomFormats')}
-          data={<EpisodeFormats formats={customFormats} />}
+          data={
+            customFormats && customFormats.length > 0 ? (
```
markus101, line 232:
```suggestion
            !!customFormats?.length ? (
```

or keep the `> 0` 🤷

markus101 (changes requested):

frontend/src/Parse/ParseResult.tsx
stevietv: lint doesn't like this:
```
  232:13  warning  Unexpected negated condition  no-negated-condition
  232:13  error    Redundant double negation     no-extra-boolean-cast
```

markus101: Ahh right, `customFormats?.length` should be sufficient

stevietv: yep

markus101 (approved):

PR #5945, opened 2023-08-20
Improve various messaging for UX

bakerboy448:
#### Database Migration
NO

#### Todos
- Tests
- [x] Wiki Updates

*


src/NzbDrone.Core/Localization/Core/en.json
```diff
@@ -322,13 +322,13 @@
   "Download": "Download",
   "DownloadClient": "Download Client",
   "DownloadClientCheckNoneAvailableHealthCheckMessage": "No download client is available",
-  "DownloadClientCheckUnableToCommunicateWithHealthCheckMessage": "Unable to communicate with {0}.",
+  "DownloadClientCheckUnableToCommunicateWithHealthCheckMessage": "Unable to communicate with {0}. Check your logs for details.",
   "DownloadClientOptionsLoadError": "Unable to load download client options",
-  "DownloadClientRootFolderHealthCheckMessage": "Download client {0} places downloads in the root folder {1}. You should not download to a root folder.",
+  "DownloadClientRootFolderHealthCheckMessage": "Download client {0} downloads completed downloads in to the root folder {1}. You should not download to the app's Library ('root') folder.",
```
markus101, line 327: Too many downloads, `places` is more correct.

`You should not download to the {2}'s root folder.` - where `{2}` is Sonarr/Radarr/Lidarr/etc

src/NzbDrone.Core/Localization/Core/en.json
```diff
@@ -630,7 +630,7 @@
   "LibraryImport": "Library Import",
   "LibraryImportHeader": "Import series you already have",
   "LibraryImportTips": "Some tips to ensure the import goes smoothly:",
-  "LibraryImportTipsDontUseDownloadsFolder": "Do not use for importing downloads from your download client, this is only for existing organized libraries, not unsorted files.",
+  "LibraryImportTipsDontUseDownloadsFolder": "Do not use for importing downloads from your download client. This is only for importing existing organized libraries and not unsorted files.",
```
markus101, line 633: The `,` is correct, it shouldn't be replaced with `and`

src/NzbDrone.Core/Localization/Core/en.json
```diff
@@ -369,7 +369,7 @@
   "EditSeries": "Edit Series",
   "Enable": "Enable",
   "EnableAutomaticAdd": "Enable Automatic Add",
-  "EnableAutomaticAddHelpText": "Add series from this list to Sonarr when syncs are performed via the UI or by Sonarr",
+  "EnableAutomaticAddHelpText": "Add series from this list to Sonarr when syncs are performed via the UI or as a scheduled app task",
```
markus101, line 372: `or automatically via scheduled tasks`

markus101 (changes requested):
Do we really want to add `Check your logs for details.` to every health check message?

bakerboy448:
> Do we really want to add `Check your logs for details.` to every health check message?

given the quantity of users that see the error and are like "what is this" spoonfeeding seems prudent

markus101:
If they need to be spoon fed checking logs wouldn't likely be sufficient and they don't need to be fed the same spoon in the same part of the app multiple times. 

Instead of telling the user `n` times to check logs we can put it once in the UI with all the health checks.

bakerboy448:
Replaced with https://github.com/Sonarr/Sonarr/issues/5958

PR #5948, opened 2023-08-21
Allow remuxed files in script import

JeWe37:
#### Database Migration
NO

#### Description
If a custom import script wants to remux a file, say from mp4 to mkv, it will have to change the extension, but therefore won't be able to import to the destination path Sonarr instructs it to. For the case of a rename having been requested, if the file is not imported with its original extension, all other possible extensions will now be attempted before actually failing the import.

### DIFF
--- a/src/NzbDrone.Core/MediaFiles/ScriptImportDecider.cs
+++ b/src/NzbDrone.Core/MediaFiles/ScriptImportDecider.cs
@@ -10,6 +10,7 @@ using NzbDrone.Core.MediaFiles.MediaInfo;
 using NzbDrone.Core.Parser;
 using NzbDrone.Core.Parser.Model;
 using NzbDrone.Core.Tags;
+using NzbDrone.Core.Tv;
 
 namespace NzbDrone.Core.MediaFiles
 {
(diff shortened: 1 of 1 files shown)


JeWe37:
The alternative to something like this is figuring out how to properly let the called process communicate with Sonarr. Then it could inform Sonarr where the file was actually imported to and we could also properly register say subtitle files generated as part of the remux. I'm still unsure how best to realize such communication though.

JeWe37:
I have an initial draft(very much untested and WIP) for this in the JeWe37/Sonarr comms branch. Will need a bit of work still, but I would appreciate some feedback on whether the basic concept is sound.

markus101 (changes requested):
Sonarr is capturing the standard/error output of the script, instead of trying to guess at the new extension and everything that could go wrong with that, the script can tell Sonarr where the file is. SABnzbd does this by expecting the output in a specific order, but instead of that Sonarr should use a specific pattern of the line written to the standard output.

JeWe37:
If you prefer that, what about 8216b56? I chose to not parse stdout, as it's often rather polluted in scripts. Instead I just used a temporary file.

JeWe37:
Closed in favor of #5956. Having flashed it out a little more I am convinced that is the better path forward.

PR #5952, opened 2023-08-22
Fix translations and default values in Interactive Search

mynameisbogdan:
#### Database Migration
NO

#### Description
- Mostly to re-add the default values for InteractiveSearchRow removed in 07f0fbf9a51d54e44681fd0f74df4e048bff561a.


frontend/src/InteractiveSearch/InteractiveSearch.js
```diff
@@ -65,6 +65,7 @@ const columns = [
   },
   {
     name: 'customFormatScore',
+    columnLabel: () => translate('CustomFormatScore'),
```
markus101, line 68: Do we need `columnLabel` here? Icon already has a title.

markus101: You already changed it, but those are required because the columns are optional, in this case it's not.

markus101 (approved):

PR #5956, opened 2023-08-24
New: communication back from custom import script

JeWe37:
#### Database Migration
NO

#### Description
There are a few cases that were not handled for custom import scripts yet:

* Custom import scripts that generate additional extra files to be imported.
* Custom import scripts that remux the original media file to a different container, such that its extension must change
* Wanting to import extra files via a custom import script(as deferred to a later date in the original #5538)

All of these use cases are addressed as follows: A custom import script can now communicate back to Sonarr by means of a temporary file, the path to which is passed as an environment variable. This avoids parsing a potentially polluted stdout, however I am open to changes in this respect. This communication is used in order to inform Sonarr of: The potentially changed name of the media file, and which extras have been imported, or equivalently generated, by the custom import script.

Still a WIP as I haven't really tested this yet and am just leaving it for initial feedback.

#### Todos
- [ ] Tests
- [ ] Wiki Updates

*

### DIFF
--- a/src/NzbDrone.Core/Extras/ExistingExtraFileService.cs
+++ b/src/NzbDrone.Core/Extras/ExistingExtraFileService.cs
@@ -10,7 +10,7 @@ namespace NzbDrone.Core.Extras
 {
     public interface IExistingExtraFiles
     {
-        List<string> ImportFileList(Series series, List<string> possibleExtraFiles);
+        List<string> ImportExtraFiles(Series series, List<string> possibleExtraFiles);
     }
 
     public class ExistingExtraFileService : IExistingExtraFiles, IHandle<SeriesScannedEvent>
@@ -25,7 +25,7 @@ namespace NzbDrone.Core.Extras
             _logger = logger;
         }
 
-        public List<string> ImportFileList(Series series, List<string> possibleExtraFiles)
+        public List<string> ImportExtraFiles(Series series, List<string> possibleExtraFiles)
         {
             _logger.Debug("Looking for existing extra files in {0}", series.Path);
 
@@ -47,7 +47,7 @@ namespace NzbDrone.Core.Extras
 
             var possibleExtraFiles = message.PossibleExtraFiles;
 
-            var importedFiles = ImportFileList(series, possibleExtraFiles);
+            var importedFiles = ImportExtraFiles(series, possibleExtraFiles);
 
             _logger.Info("Found {0} possible extra files, imported {1} files.", possibleExtraFiles.Count, importedFiles.Count);
         }
--- a/src/NzbDrone.Core/Extras/ExistingExtraFileService.cs
+++ b/src/NzbDrone.Core/Extras/ExistingExtraFileService.cs
@@ -44,9 +44,7 @@ namespace NzbDrone.Core.Extras
         public void Handle(SeriesScannedEvent message)
         {
             var series = message.Series;
-
             var possibleExtraFiles = message.PossibleExtraFiles;
-
             var importedFiles = ImportExtraFiles(series, possibleExtraFiles);
 
             _logger.Info("Found {0} possible extra files, imported {1} files.", possibleExtraFiles.Count, importedFiles.Count);
--- a/src/NzbDrone.Core/Extras/ExistingExtraFileService.cs
+++ b/src/NzbDrone.Core/Extras/ExistingExtraFileService.cs
@@ -2,33 +2,25 @@ using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using NLog;
-using NzbDrone.Common.Disk;
-using NzbDrone.Core.MediaFiles;
 using NzbDrone.Core.MediaFiles.Events;
 using NzbDrone.Core.Messaging.Events;
 using NzbDrone.Core.Tv;
 
 namespace NzbDrone.Core.Extras
 {
-    public interface IExistingExtraFiles : IHandle<SeriesScannedEvent>
+    public interface IExistingExtraFiles
     {
         List<string> ImportFileList(Series series, List<string> possibleExtraFiles);
     }
 
-    public class ExistingExtraFileService : IExistingExtraFiles
+    public class ExistingExtraFileService : IExistingExtraFiles, IHandle<SeriesScannedEvent>
     {
-        private readonly IDiskProvider _diskProvider;
-        private readonly IDiskScanService _diskScanService;
         private readonly List<IImportExistingExtraFiles> _existingExtraFileImporters;
         private readonly Logger _logger;
 
-        public ExistingExtraFileService(IDiskProvider diskProvider,
-                                        IDiskScanService diskScanService,
-                                        IEnumerable<IImportExistingExtraFiles> existingExtraFileImporters,
+        public ExistingExtraFileService(IEnumerable<IImportExistingExtraFiles> existingExtraFileImporters,
                                         Logger logger)
         {
-            _diskProvider = diskProvider;
-            _diskScanService = diskScanService;
             _existingExtraFileImporters = existingExtraFileImporters.OrderBy(e => e.Order).ToList();
             _logger = logger;
         }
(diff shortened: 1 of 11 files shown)


markus101 (changes requested):
Left some inline comments.

JeWe37:
Ultimately I am not horribly invested in the file idea. I'll switch to just looking at stdout(and fix the other things).

JeWe37:
Switched to using stdout and addressed the other comments. In testing it turned out there was a dependency loop, broke that by adding some more information to the series scanned event which I think makes sense architecturally.

src/NzbDrone.Core/MediaFiles/ScriptImportDecider.cs
```diff
@@ -25,23 +28,75 @@ public class ImportScriptService : IImportScript
...
+                        throw new ScriptImportException("Script output contains invalid extra file: {0}", lastAdded);
+                    }
+                    else if (!_diskProvider.FileExists(lastAdded))
+                    {
+                        throw new ScriptImportException("Script output contains non-existent extra file: {0}", lastAdded);
+                    }
+                }
+            }
+
+            return new ScriptImportInfo(possibleExtraFiles, mediaFile ?? defaultMediaFile);
```
markus101, line 97: It shouldn't be the responsibility of the output parser to return a default.

src/NzbDrone.Core/MediaFiles/ScriptImportDecider.cs
```diff
@@ -25,23 +28,75 @@ public class ImportScriptService : IImportScript
...
             _tagRepository = tagRepository;
+            _diskProvider = diskProvider;
             _logger = logger;
         }
 
+        private static readonly Regex MediaFileLine = new Regex(@"^\[SonarrMediaFile\]\s*(?<mediaFileName>.*)$", RegexOptions.Compiled);
+
+        private static readonly Regex ExtraFileLine = new Regex(@"^\[SonarrExtraFile\]\s*(?<extraFileName>.*)$", RegexOptions.Compiled);
+
+        private ScriptImportInfo ProcessStdout(List<ProcessOutputLine> processOutputLines, string defaultMediaFile)
```
markus101, line 55: `ProcessStdout` -> `ProcessOutput`

src/NzbDrone.Core/MediaFiles/ScriptImportDecider.cs
```diff
@@ -118,12 +173,27 @@ public ScriptImportDecision TryImport(string sourcePath, string destinationFileP
...
+
+            episodeFile.RelativePath = series.Path.GetRelativePath(mediaFile);
+            episodeFile.Path = mediaFile;
+
+            if ((processOutput.ExitCode & 0x4) == 0x4)
+            {
+                localEpisode.ShouldImportExtras = true;
+            }
+
+            switch (processOutput.ExitCode & 0x3)
```
markus101, line 189: Should be able to achieve all this without mixing in the bit-wise operator along with an if and a switch which will make it easier to read and maintain.

src/NzbDrone.Core/MediaFiles/ScriptImportDecider.cs
```diff
@@ -25,23 +28,75 @@ public class ImportScriptService : IImportScript
...
             _configService = configService;
             _configFileProvider = configFileProvider;
             _tagRepository = tagRepository;
+            _diskProvider = diskProvider;
             _logger = logger;
         }
 
+        private static readonly Regex MediaFileLine = new Regex(@"^\[SonarrMediaFile\]\s*(?<mediaFileName>.*)$", RegexOptions.Compiled);
+
+        private static readonly Regex ExtraFileLine = new Regex(@"^\[SonarrExtraFile\]\s*(?<extraFileName>.*)$", RegexOptions.Compiled);
```
markus101, line 53: I think something like:

```suggestion
        private static readonly Regex OutputRegex = new Regex(@"^\[(?:(?<mediaFile>MediaFile)|(?<extraFile>ExtraFile))\]\s(?<fileName>.+)$", RegexOptions.Compiled);
```

for both these regexes to be a little more strict on the format and ensuring the script provided information instead of a null/empty value.

Changed to `\[MediaFile\]` to make it universal with downstream apps and merged into a single regex.

src/NzbDrone.Core/MediaFiles/ScriptImportDecider.cs
```diff
@@ -25,23 +28,75 @@ public class ImportScriptService : IImportScript
...
+
+        private ScriptImportInfo ProcessStdout(List<ProcessOutputLine> processOutputLines, string defaultMediaFile)
+        {
+            var possibleExtraFiles = new List<string>();
+            string mediaFile = null;
+
+            foreach (var line in processOutputLines)
+            {
+                if (MediaFileLine.Match(line.Content) is var match && match.Success)
```
markus101, line 62: With the suggestion to use one regex this will need to be reworked a bit, but should be pretty similar.

markus101 (changes requested):
Left some ideas to simplify it a little bit, but looking good.

src/NzbDrone.Core/MediaFiles/ScriptImportDecider.cs
JeWe37: Personally I don't really like my new alternative any better...

markus101: Did you forget to push these changes up?

src/NzbDrone.Core/MediaFiles/ScriptImportDecider.cs
JeWe37: I did indeed...... Fixed.

src/NzbDrone.Core/MediaFiles/ScriptImportDecider.cs
JeWe37: I did indeed...... Fixed.

markus101 (changes requested):
Getting away from the exit code driving behaviour would be good, since this is still in a beta release a breaking change there would be acceptable.

src/NzbDrone.Core/MediaFiles/ScriptImportDecider.cs
```diff
@@ -25,23 +27,71 @@ public class ImportScriptService : IImportScript
...
+                }
+                else if (match.Groups["extraFile"].Success)
+                {
+                    possibleExtraFiles.Add(match.Groups["fileName"].Value);
+
+                    var lastAdded = possibleExtraFiles.Last();
+
+                    if (!_diskProvider.FileExists(lastAdded))
+                    {
+                        throw new ScriptImportException("Script output contains non-existent possible extra file: {0}", lastAdded);
```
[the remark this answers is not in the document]
JeWe37: Switched to emitting a warning. Import state should still be fine I think, I don't think anything was added to the database by this point? But changed anyway.

src/NzbDrone.Core/MediaFiles/EpisodeImport/ImportApprovedEpisodes.cs
```diff
@@ -130,7 +133,20 @@ public List<ImportResult> Import(List<ImportDecision> decisions, bool newDownloa
 
                     if (newDownload)
                     {
-                        _extraService.ImportEpisode(localEpisode, episodeFile, copyOnly);
+                        if (localEpisode.ScriptImported)
+                        {
+                            _existingExtraFiles.ImportFileList(localEpisode.Series, localEpisode.PossibleExtraFiles);
```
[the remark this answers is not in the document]
JeWe37: This is actually intended, "ShouldImportExtras" only is supposed to instruct sonarr to import extras. Even if we prevent Sonarr from importing in the script, I think it still makes sense to be able to import extras via the script ourselves. This is mostly just for the case where the script doesn't handle extras, but we want to still import them via Sonarr's stock logic.

src/NzbDrone.Core/Extras/ExistingExtraFileService.cs
```diff
@@ -50,6 +38,17 @@ public void Handle(SeriesScannedEvent message)
...
+            return importedFiles;
+        }
+
+        public void Handle(SeriesScannedEvent message)
+        {
+            var series = message.Series;
+
+            var possibleExtraFiles = message.PossibleExtraFiles;
+
+            var importedFiles = ImportExtraFiles(series, possibleExtraFiles);
```
markus101, line 50:
```suggestion
            var series = message.Series;
            var possibleExtraFiles = message.PossibleExtraFiles;
            var importedFiles = ImportExtraFiles(series, possibleExtraFiles);
```

markus101 (approved):

Issue #5955, opened 2023-08-23
Support preserving title tag when renaming subtitle files


JeWe37:
### Is there an existing issue for this?

- [X] I have searched the existing open and closed issues

### Is your feature request related to a problem? Please describe

Jellyfin allows for the [specification of the titles](https://jellyfin.org/docs/general/server/media/external-files/) of subtitle files
> Any arbitrary text not parsable to a language or flag will be combined and used as the title of the stream (if there is not a stream title already embedded in the file metadata).

These titles, if present on existing imported subtitle files, will not be properly preserved when a Sonarr rename operation is executed. 

### Describe the solution you'd like

Sonarr should parse these titles and include them when renaming of the episode is requested.

### Describe alternatives you've considered

Can't think of any.

### Anything else?

This is especially nice for extracting subtitle files from media files, as the subtitle streams are often tagged with identifying information as a title, such as whether they are commentary tracks or the like.

bakerboy448:
Example names? standard language tags are already supported 

xyz.en.srt
xyz.en.forced.srt
xyz.en.sdh.srt
xyz.en.hi.srt

https://github.com/Sonarr/Sonarr/commit/b20e247feb24084348261ef2fad61a05f1fd69e8
https://github.com/Sonarr/Sonarr/commit/3141bf7f9d8afcce676c3fbfccbde66e5b317b2b

JeWe37:
EPISODENAME.Full Subtitles.eng.default.srt
EPISODENAME.eng.Signs & Songs.forced.ass

JeWe37:
I don't follow, I'm not sure we mean the same thing. What I mean is that with current Sonarr the following will happen: 
```
EPISODENAME.Full Subtitles.eng.default.srt > EPISODENAME.eng.default.srt
EPISODENAME.eng.Signs & Songs.forced.ass > EPISODENAME.eng.forced.ass
```
In my testing sure the matcher itself was sufficiently robust to handle this without issue. The issue is that the title tags are lost.

JeWe37:
Maybe it'll be clearer when I explain what actually needs to change in the code:
The LanguageTags parser needs to be changed to take anything that isn't a language tag but still isn't just the part of the name of the corresponding EpisodeFile and put it into the SubtitleFile as a title. From there that title needs to be used in rename operations, if it is present, and added to the list of tags.

markus101:
[...]

PR #5964, opened 2023-08-26
New: Parsing subtitle titles

JeWe37:
#### Database Migration
YES (198)

#### Description
Essentially the idea here is as follows: Try to parse the format that would be expected if there was a title. If the title tag that is parsed is first, there is the risk that for files named like
```
Name. Rest of name.eng.default.srt
```
We would incorrectly parse the title to be `Rest of name`. To fix this, this PR checks if `Rest of name` occurs within the name of the corresponding episode file, on the assumption that the title of the episode should mostly only differ by the tags. Some of the rationale for this was explained in the comments on #5955.

Note that with the way this works, the tags that make subtitles unique will now be parsed as titles. I would argue this is actually desirable behavior but I still need to fix the generation of the unique names for this I believe.

That aside this PR also relaxes the order in which tags have to appear in the name of the file. Now the language code can occur anywhere, it doesn't have to be first.

Very much WIP and untested, mostly a sketch.

#### Todos
- [ ] Tests

* #5955


src/NzbDrone.Core/Parser/LanguageParser.cs
```diff
@@ -25,7 +26,11 @@ public static class LanguageParser
         private static readonly Regex CaseSensitiveLanguageRegex = new Regex(@"(?:(?i)(?<!SUB[\W|_|^]))(?:(?<lithuanian>\bLT\b)|(?<czech>\bCZ\b)|(?<polish>\bPL\b)|(?<bulgarian>\bBG\b)|(?<slovak>\bSK\b))(?:(?i)(?![\W|_|^]SUB))",
                                                                 RegexOptions.Compiled);
 
-        private static readonly Regex SubtitleLanguageRegex = new Regex(".+?[-_. ](?<iso_code>[a-z]{2,3})([-_. ](?<tags>full|forced|foreign|default|cc|psdh|sdh))*$", RegexOptions.Compiled | RegexOptions.IgnoreCase);
+        private static readonly Regex SubtitleLanguageRegex = new Regex(".+?([-_. ](?<tags>full|forced|foreign|default|cc|psdh|sdh))*[-_. ](?<iso_code>[a-z]{2,3})([-_. ](?<tags>full|forced|foreign|default|cc|psdh|sdh))*$", RegexOptions.Compiled | RegexOptions.IgnoreCase);
+
+        private static readonly Regex SubtitleLanguageTitleRegex = new Regex(".+?(\\.((?<tags1>full|forced|foreign|default|cc|psdh|sdh)|(?<iso_code>[a-z]{2,3})))*\\.(?<title>[^.]*)(\\.((?<tags2>full|forced|foreign|default|cc|psdh|sdh)|(?<iso_code>[a-z]{2,3})))*$", RegexOptions.Compiled | RegexOptions.IgnoreCase);
```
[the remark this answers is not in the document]
markus101: I don't think we can assume title will never have a period

src/NzbDrone.Core/Parser/LanguageParser.cs
```diff
@@ -226,6 +231,56 @@ public static Language ParseSubtitleLanguage(string fileName)
...
+                .Select(tag => tag.Value.ToLower());
+            var title = matchTitle.Groups["title"].Captures.Cast<Capture>().First().ToString();
+
+            if (matchTitle.Groups["tags1"].Captures.Empty())
+            {
+                var episodeFile = episode.EpisodeFile.Value;
+                var episodeFileTitle = Path.GetFileNameWithoutExtension(episodeFile.RelativePath);
+                var originalEpisodeFileTitle = Path.GetFileNameWithoutExtension(episodeFile.OriginalFilePath) ?? string.Empty;
+
+                if (episodeFileTitle.Contains(title, StringComparison.OrdinalIgnoreCase) || originalEpisodeFileTitle.Contains(title, StringComparison.OrdinalIgnoreCase))
```
markus101, line 257: This logic shouldn't live in the parser, like language parsing this should probably live in a augmenter.

src/NzbDrone.Core/Parser/LanguageParser.cs
```diff
@@ -226,6 +231,56 @@ public static Language ParseSubtitleLanguage(string fileName)
...
         }
 
+        public static (List<string> languageTags, string title) ParseLanguageTagsAndTitle(string fileName, Episode episode)
+        {
+            var simpleFilename = Path.GetFileNameWithoutExtension(fileName);
+            var matchTitle = SubtitleLanguageTitleRegex.Match(simpleFilename);
+
+            if (matchTitle.Groups["iso_code"].Captures.Count != 1)
+            {
+                throw new ArgumentException("Subtitle file title probably parsed incorrectly, not using.");
```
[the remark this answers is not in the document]
JeWe37: Suspected that would be preferred, will change that once I get there.

src/NzbDrone.Core/Parser/LanguageParser.cs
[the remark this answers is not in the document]
JeWe37: Well the reason that is necessary is to check if the title is first. If it isn't first, parsing is a little easier because the other tags are sufficiently specific as that they should mostly discard everything that comes at the end of an episode title and isn't a tag/subtitle title. The question of whether there is a dot only comes into play if the title is first(or non-existent). To guard against incorrectly parsing a title when there is none, the heuristic that if the episode file path contains the string that was parsed as a subtitle file title, it will not be used. See some of the testcases for examples of that.

src/NzbDrone.Core/Parser/LanguageParser.cs
```diff
@@ -226,6 +231,56 @@ public static Language ParseSubtitleLanguage(string fileName)
...
+                if (episodeFileTitle.Contains(title, StringComparison.OrdinalIgnoreCase) || originalEpisodeFileTitle.Contains(title, StringComparison.OrdinalIgnoreCase))
+                {
+                    throw new ArgumentException("Subtitle file title probably parsed incorrectly, not using.");
+                }
+            }
+
+            return (languageTags.ToList(), title);
+        }
+
+        public static int CopyFromTitle(string title)
```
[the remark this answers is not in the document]
JeWe37: The thing is that Sonarr previously(has been changed) will add something like `.3` to mark subs that would have the same names. The issue is that that format would be parsed as a title, so we can't use that anymore really. Handling for that remains very much work in progress(as does everything in this PR).

src/NzbDrone.Core/Parser/LanguageParser.cs
JeWe37: In so far as I understand the augmenters exist to add the rest of the available info to a partial local episode where possible. What exactly should I be augmenting here? Should I create a new set of augmenters that add info such as title, copies etc.? Those would be entirely separate from the existing local episode augmenters tho.

JeWe37:
Alright, made some more changes, migration is tested pretty thoroughly now, also tested it in the actual renaming use case etc. Assuming that the core expectations(regarding how dots may be placed and such) are fine, this is functional. Feel free to review in detail.

JeWe37:
I'm not totally sure why CI isn't happy with this, the tests do work locally. Is there anything specific to the tests for migrations that's breaking this?

src/NzbDrone.Core/Parser/LanguageParser.cs
markus101: They do several things, but I'm specifically referring to this: https://github.com/Sonarr/Sonarr/blob/develop/src/NzbDrone.Core/MediaFiles/EpisodeImport/Aggregation/Aggregators/Augmenters/Language/AugmentLanguageFromFileName.cs#L22-L27

The parser should only parse out information, it shouldn't try to make sense of what it parsed out and strip information. Instead that can be stripped out later.

src/NzbDrone.Core/Parser/LanguageParser.cs
```diff
@@ -226,6 +231,63 @@ public static Language ParseSubtitleLanguage(string fileName)
             return Language.Unknown;
         }
 
+        public static (List<string> languageTags, string title, Language language) ParseLanguageTagsAndTitle(string fileName, Episode episode)
```
markus101, line 234: Maybe `ParseSubtitleLanguageInformation`? and return an object with that information.

src/NzbDrone.Core/Parser/LanguageParser.cs
```diff
@@ -226,6 +231,63 @@ public static Language ParseSubtitleLanguage(string fileName)
...
+        {
+            var simpleFilename = Path.GetFileNameWithoutExtension(fileName);
+            var matchTitle = SubtitleLanguageTitleRegex.Match(simpleFilename);
+
+            if (matchTitle.Groups["iso_code"].Captures.Count is var languageCodeNumber && languageCodeNumber != 1)
+            {
+                throw new LanguageParsingException(string.Format("Subtitle file title probably parsed incorrectly, {0} language codes found, not using.", languageCodeNumber));
```
markus101, line 241: I don't think we should throw if the parsing fails.

src/NzbDrone.Core/Parser/LanguageParser.cs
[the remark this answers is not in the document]
markus101: What about something like `series.title.s01e01.my.name.is.default.eng.forced.srt`? If the episode file is `My Name is Default` this will still parse it incorrectly while trying to guard against it and it should still be cleaned up via an augmenter.

src/NzbDrone.Core/Parser/LanguageParser.cs
JeWe37: The example you gave in the other comment is why I'm actually checking both the original and current file name for matches. I guess I could look at the entire history of renames of the file(somehow, need to see how that might be possible)? Or at least at the previous name.

JeWe37: What I still find strange is how augmenters typically operate on local episodes-this would essentially mean having the list of the subtitle files as part of the local episode and then just import that I suppose?

JeWe37:
> What about something like `series.title.s01e01.my.name.is.default.eng.forced.srt`? If the episode file is `My Name is Default` this will still parse it incorrectly while trying to guard against it and it should still be cleaned up via an augmenter.

Changed the first few things, using the augmenters I'll do once I know what exactly to do about this case

src/NzbDrone.Core/Parser/LanguageParser.cs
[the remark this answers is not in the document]
JeWe37: This is why I'm actually checking both the original and current file name for matches. I guess I could look at the entire history of renames of the file(somehow, need to see how that might be possible)? Or at least at the previous name.

markus101:
Similar to what we're doing with language augmentation (where languages that are part of the episode title are removed) we'd want to remove `default` from the tags when processing this and recognizing that the title is part of the name, see `AugmentLanguageFromFileName`.

I ended up stealing migration 197 in a recent commit, if you could update to 198 please, sorry for the extra work there.

JeWe37:
So, I believe that should complete the move to using aggregators. Took me a bit, I have been somewhat busy. Minor note: There is now an ordering to the aggregators as the episode needs to be resolved first before we can figure out the title and all that.

I'd love to get this and #5956 to done soon. I'm guessing a round more of reviews might be needed here, but over there I'm hoping most everything is in order.

JeWe37:
Seems that CI still doesn't like my migration tests despite them working locally, any idea what may be causing that? I'm not familiar with your infrastructure.

markus101:
The migration is failing against Postgres, nothing obvious strikes me as being wrong at first glance.

src/NzbDrone.Core/Datastore/Migration/198_parse_titles_from_existing_subtitle_files.cs
```diff
@@ -0,0 +1,68 @@
...
+        }
+
+        private void UpdateTitles(IDbConnection conn, IDbTransaction tran)
+        {
+            var updates = new List<object>();
+
+            using (var cmd = conn.CreateCommand())
+            {
+                cmd.Transaction = tran;
+                cmd.CommandText = "SELECT \"Id\", \"RelativePath\", \"EpisodeFileId\", \"Language\", \"LanguageTags\" FROM \"SubtitleFiles\"";
```
markus101, line 30: Looks like `Language` and `LanguageTags` can be removed.

src/NzbDrone.Core/Parser/Model/SubtitleTitleInfo.cs
```diff
@@ -0,0 +1,34 @@
...
+
+namespace NzbDrone.Core.Parser.Model
+{
+    public class SubtitleTitleInfo
+    {
+        private static readonly Regex SubtitleTitleRegex = new Regex("((?<title>.+) - )?(?<copy>\\d+)", RegexOptions.Compiled);
+        public List<string> LanguageTags { get; set; }
+        public Language Language { get; set; }
+        public string RawTitle { get; set; }
+        private Lazy<(string Title, int Copy)> TitleAndCopy => new Lazy<(string title, int copy)>(() =>
```
markus101, line 14: The model should be dumb and just a method to store the data parsed elsewhere and passed in. Can this logic instead live in `ParseSubtitleLanguageInformation`?

src/NzbDrone.Core/Datastore/Migration/198_parse_titles_from_existing_subtitle_files.cs
```diff
@@ -0,0 +1,68 @@
...
+
+                using var reader = cmd.ExecuteReader();
+                while (reader.Read())
+                {
+                    var id = reader.GetInt32(0);
+                    var relativePath = reader.GetString(1);
+                    var episodeFileId = reader.GetInt32(2);
+
+                    var episodeFile = conn.QuerySingle<EpisodeFile>("SELECT * FROM \"EpisodeFiles\" WHERE \"Id\" = @Id", new { Id = episodeFileId });
```
markus101, line 39: We should avoid a separate query for this, maybe via a join getting only the pieces of information that this needs.

src/NzbDrone.Core/Datastore/Migration/198_parse_titles_from_existing_subtitle_files.cs
```diff
@@ -0,0 +1,68 @@
...
+
+                    var subtitleTitleInfo = AggregateSubtitleInfo.CleanSubtitleTitleInfo(episodeFile, relativePath);
```
markus101, line 41: The migration should be a snapshot in time and instead of calling this function we should extract the necessary pieces into a function in this migration, that way if the structure of episode file changes or we refactor `AggregateSubtitleInfo.CleanSubtitleTitleInfo` we don't break this migration.

src/NzbDrone.Core/MediaFiles/EpisodeImport/Aggregation/Aggregators/AggregateSubtitleInfo.cs
```diff
@@ -0,0 +1,59 @@
...
+using NzbDrone.Core.Download;
+using NzbDrone.Core.Extras.Subtitles;
+using NzbDrone.Core.Parser;
+using NzbDrone.Core.Parser.Model;
+
+namespace NzbDrone.Core.MediaFiles.EpisodeImport.Aggregation.Aggregators
+{
+    public class AggregateSubtitleInfo : IAggregateLocalEpisode
+    {
+        private static readonly Logger Logger = NzbDroneLogger.GetLogger(typeof(AggregateSubtitleInfo));
```
markus101, line 15: With the other suggestions, this should be injected.

src/NzbDrone.Core/MediaFiles/EpisodeImport/Aggregation/Aggregators/AggregateSubtitleInfo.cs
```diff
@@ -0,0 +1,59 @@
...
+            }
+
+            var firstEpisode = localEpisode.Episodes.First();
+            var episodeFile = firstEpisode.EpisodeFile.Value;
+            localEpisode.SubtitleInfo = CleanSubtitleTitleInfo(episodeFile, path);
+
+            return localEpisode;
+        }
+
+        public static SubtitleTitleInfo CleanSubtitleTitleInfo(EpisodeFile episodeFile, string path)
```
markus101, line 34: No need for this to be static

src/NzbDrone.Core/MediaFiles/EpisodeImport/Aggregation/Aggregators/AggregateSubtitleInfo.cs
```diff
@@ -0,0 +1,59 @@
...
+    public class AggregateSubtitleInfo : IAggregateLocalEpisode
+    {
+        private static readonly Logger Logger = NzbDroneLogger.GetLogger(typeof(AggregateSubtitleInfo));
+        public int Order => 6;
+
+        public LocalEpisode Aggregate(LocalEpisode localEpisode, DownloadClientItem downloadClientItem)
+        {
+            var path = localEpisode.Path;
+            var isSubtitleFile = SubtitleFileExtensions.Extensions.Contains(Path.GetExtension(path));
+            if (!isSubtitleFile)
```
markus101, line 22: New line before block
```suggestion

            if (!isSubtitleFile)
```

src/NzbDrone.Core/Datastore/Migration/198_parse_titles_from_existing_subtitle_files.cs
```diff
@@ -0,0 +1,68 @@
...
+                        Id = id,
+                        Title = subtitleTitleInfo.Title,
+                        Language = subtitleTitleInfo.Language,
+                        LanguageTags = subtitleTitleInfo.LanguageTags,
+                        Copy = subtitleTitleInfo.Copy
+                    });
+                }
+            }
+
+            var serializerSettings = new JsonSerializerOptions
```
markus101, line 54: Copy/pasta from another migration?

markus101 (changes requested):
Not sure if making the migration simpler, will help with the postgres issue directly, but it should help overall.

src/NzbDrone.Core/Datastore/Migration/198_parse_titles_from_existing_subtitle_files.cs
JeWe37: Not quite, there may be some parsing differences where tags/languages that previously weren't detected because of a title that messed with the old regex.

src/NzbDrone.Core/Datastore/Migration/198_parse_titles_from_existing_subtitle_files.cs
JeWe37: Yea I was wondering about this. It's taken from migration 195 – not sure why it's there either to be honest. If you say I can just get rid of it and it is actually just as unused as it looks that's what I'll do.

src/NzbDrone.Core/Datastore/Migration/198_parse_titles_from_existing_subtitle_files.cs
JeWe37: Well I changed that, however it does still depend on the `LanguageParser`. If I wanted to avoid that I would have to copy in several methods from that as well, which admittedly seems like a horrific amount of code duplication.

JeWe37:
Looks as though this did indeed resolve the Postgres issues.

src/NzbDrone.Core/Datastore/Migration/198_parse_titles_from_existing_subtitle_files.cs
markus101: Yeah, we do use parsers elsewhere, but their inputs are usually pretty basic, so should be safe, the extra copied code here does clean up the need for a full episode file at least.

JeWe37:
I was hoping that I could get this in before v4, but seeing as that was released now I am now wondering what the policy for new features like this one will be: Will these now go in a point release of v4? 

I think this should be rebased correctly now(still waiting on CI right now). I'm guessing with the v4 release there are currently other priorities but I would appreciate if this could be merged some time soon/if you could tell me what further adjustments you would like prior to that.

JeWe37:
Github CI is throwing `Resource not accessible by integration`, which to me sounds like perhaps there are still some issues with #6333 having just been freshly merged? The old CI seems to be happy with the PR.

markus101:
I typed a reply, but I guess I never sent it. This should be good to rebase to make github actions happy and we'll give it another review.

src/NzbDrone.Core/Extras/Subtitles/SubtitleFile.cs
```diff
@@ -13,15 +13,19 @@ public SubtitleFile()
...
         public List<string> LanguageTags { get; set; }
 
+        public string Title { get; set; }
+
         private string LanguageTagsAsString => string.Join(".", LanguageTags);
 
         public override string ToString()
         {
-            return $"[{Id}] {RelativePath} ({Language}{(LanguageTags.Count > 0 ? "." : "")}{LanguageTagsAsString}{Extension})";
+            return $"[{Id}] {RelativePath} ({Language}{(Title is not null ? "." : "")}{Title ?? ""}{(LanguageTags.Count > 0 ? "." : "")}{LanguageTagsAsString}{Extension})";
```
markus101, line 28: I think doing this with a string builder will be more readable in the long run, especially with the chained conditionals.

src/NzbDrone.Core/MediaFiles/EpisodeImport/Aggregation/Aggregators/AggregateEpisodes.cs
```diff
@@ -10,6 +10,8 @@ namespace NzbDrone.Core.MediaFiles.EpisodeImport.Aggregation.Aggregators
 {
     public class AggregateEpisodes : IAggregateLocalEpisode
     {
+        public int Order => 1;
```
markus101, line 13: Since most of these don't have a strict order I think it'd make sense to group everything that doesn't have a requirement at `1` and then things that do at `2`. That way multiple files don't need to be changed to insert something later.

markus101 (changes requested):
Thanks for the rebasing and fixes so far, just a couple minor things then good to land I think

markus101 (approved):

PR #5967, opened 2023-08-27
Fix: Fallback to English translations if invalid UI language in config

stevietv:
#### Database Migration
 NO

#### Description
We saw on Discord that it was possible for someone to have -2 as an original language, which resulted in translations failing.
This falls back to English if an invalid Language is returned during evaluating what language the UI should be in.

*

### DIFF
--- a/frontend/src/Settings/UI/UISettings.js
+++ b/frontend/src/Settings/UI/UISettings.js
@@ -220,6 +220,13 @@ class UISettings extends Component {
                       helpTextWarning={translate('BrowserReloadRequired')}
                       onChange={onInputChange}
                       {...settings.uiLanguage}
+                      errors={
+                        languages.some((language) => language.key === settings.uiLanguage.value) ?
+                          settings.uiLanguage.errors :
+                          [
+                            ...settings.uiLanguage.errors,
+                            { message: translate('InvalidUILanguage') }
+                          ]}
                     />
                   </FormGroup>
                 </FieldSet>
--- a/src/Sonarr.Api.V3/Config/UiConfigController.cs
+++ b/src/Sonarr.Api.V3/Config/UiConfigController.cs
@@ -1,7 +1,9 @@
 using System.Linq;
 using System.Reflection;
+using FluentValidation;
 using Microsoft.AspNetCore.Mvc;
 using NzbDrone.Core.Configuration;
+using NzbDrone.Core.Languages;
 using Sonarr.Http;
 using Sonarr.Http.REST.Attributes;
 
@@ -16,6 +18,17 @@ namespace Sonarr.Api.V3.Config
             : base(configService)
         {
             _configFileProvider = configFileProvider;
+            SharedValidator.RuleFor(c => c.UILanguage).Custom((value, context) =>
+            {
+                if (!Language.All.Any(o => o.Id == value))
+                {
+                    context.AddFailure("Invalid UI Language value");
+                }
+            });
+
+            SharedValidator.RuleFor(c => c.UILanguage)
+                           .GreaterThanOrEqualTo(1)
+                           .WithMessage("The UI Language value cannot be less than 1");
         }
 
         [RestPutById]
--- a/src/NzbDrone.Core/Localization/Core/en.json
+++ b/src/NzbDrone.Core/Localization/Core/en.json
@@ -612,6 +612,7 @@
   "InteractiveSearchResultsFailedErrorMessage": "Search failed because its {message}. Try refreshing the series info and verify the necessary information is present before searching again.",
   "Interval": "Interval",
   "InvalidFormat": "Invalid Format",
+  "InvalidUILanguage": "Your UI is set to an invalid language, correct it and save your settings",
   "KeyboardShortcuts": "Keyboard Shortcuts",
   "KeyboardShortcutsCloseModal": "Close Current Modal",
   "KeyboardShortcutsConfirmModal": "Accept Confirmation Modal",
--- a/src/NzbDrone.Core/Localization/LocalizationService.cs
+++ b/src/NzbDrone.Core/Localization/LocalizationService.cs
@@ -87,7 +87,7 @@ namespace NzbDrone.Core.Localization
 
         public string GetLanguageIdentifier()
         {
-            var isoLanguage = IsoLanguages.Get((Language)_configService.UILanguage);
+            var isoLanguage = IsoLanguages.Get((Language)_configService.UILanguage) ?? IsoLanguages.Get(Language.English);
             var language = isoLanguage.TwoLetterCode;
 
             if (isoLanguage.CountryCode.IsNotNullOrWhiteSpace())


markus101 (changes requested):
I think we should also add some validation to `UiConfigController` to tell users to fix it on save. `UI Language` on `settings/ui` will likely default to Arabic (first option), which we might want to do something about.

stevietv:
> I think we should also add some validation to `UiConfigController` to tell users to fix it on save. `UI Language` on `settings/ui` will likely default to Arabic (first option), which we might want to do something about.

I added a warning when the value in the settings is not a valid language.
for the validation I stuck with making sure its greater than or equal to 1. Not sure how to tackle here checking its an actual UI language.

frontend/src/Settings/UI/UISettings.js
```diff
@@ -210,6 +210,13 @@ class UISettings extends Component {
                 </FieldSet>
 
                 <FieldSet legend={translate('Language')}>
+                  {
```
markus101, line 213: I was thinking something along the lines of a fake language entry (disabled) that is `Choose Language` or similar, something with the form validation could work too, but I don't think this is needed.

frontend/src/Settings/UI/UISettings.js
```diff
@@ -210,6 +210,13 @@ class UISettings extends Component {
...
+                    !(languages.some((language) => language.key === settings.uiLanguage.value)) &&
```
markus101, line 214: Use a ternary instead of `&&`

src/Sonarr.Api.V3/Config/UiConfigController.cs
```diff
@@ -16,6 +17,8 @@ public UiConfigController(IConfigFileProvider configFileProvider, IConfigService
             : base(configService)
         {
             _configFileProvider = configFileProvider;
+            SharedValidator.RuleFor(c => c.UILanguage)
```
markus101, line 20: Could do something like what we do in `LanguageSpecification` to make it a little more strict, though I'd also add in a check to make sure the value is greater than or equal to 1 with a custom message.

markus101 (changes requested):

frontend/src/Settings/UI/UISettings.js
stevietv: So I got the other two changes pushed, but this one I can't get. I was able to add an additional Language to the list 'Choose Language' and set it as disabled, but I can't find a way to set the value of the dropdown to it if the language is disabled.

currently this is what it looks like when an invalid code is selected: 
image

markus101: Ahh, I was expecting it to select the first value, but if it's just showing the invalid valid I don't think we need to do anything and instead we could just use the the error handling for the form input and do something like:

```
<FormGroup>
  <FormLabel>{translate('UiLanguage')}</FormLabel>
  <FormInputGroup
    type={inputTypes.SELECT}
    name="uiLanguage"
    values={languages}
    helpText={translate('UiLanguageHelpText')}
    helpTextWarning={translate('BrowserReloadRequired')}
    onChange={onInputChange}
    {...settings.uiLanguage}
    errors={[
      { message: 'Invalid language selected' }
    ]}
  />
</FormGroup>
```

But we should use `settings.errors` and only show that error if it's an invalid value

stevietv: I've got it like this for now.
did you want to leave the alert above, or just keep the error message below.

```
  <FormGroup>
	<FormLabel>{translate('UiLanguage')}</FormLabel>
	<FormInputGroup
	  type={inputTypes.SELECT}
	  name="uiLanguage"
	  values={languages}
	  helpText={translate('UiLanguageHelpText')}
	  helpTextWarning={translate('BrowserReloadRequired')}
	  onChange={onInputChange}
	  {...settings.uiLanguage}
	  errors={
              languages.some((language) => language.key === settings.uiLanguage.value) ?
		[] :
		[
		  { message: 'Invalid Language Selected' }
		]}
	/>
  </FormGroup>
 ```

Comment (markus101) [reply frontend/src/Settings/UI/UISettings.js:213]: I don't think we also need the alert above, but also with the server side validation maybe we don't even need this, since they can't save changes with it in the error state.

You'll also need to do `...settings.errors` to not hide errors coming back from the server.

Comment (stevietv) [reply frontend/src/Settings/UI/UISettings.js]: I took the alert out and left only the error message.
I think the error message is nice, in-case somehow the language did get set incorrectly, like we saw in the original case from discord.

File: frontend/src/Settings/UI/UISettings.js [hunk 4]
```diff
@@ -220,6 +220,13 @@ class UISettings extends Component {
                       helpTextWarning={translate('BrowserReloadRequired')}
                       onChange={onInputChange}
                       {...settings.uiLanguage}
+                      {...settings.errors}
+                      errors={
+                        languages.some((language) => language.key === settings.uiLanguage.value) ?
+                          [] :
+                          [
+                            { message: translate('InvalidUILanguage') }
+                          ]}
```
markus101, line 229: It wasn't super clear what I meant with my previous comment, but it'll need to be like this to spread the server errors correctly.

```suggestion
                      errors={
                        languages.some((language) => language.key === settings.uiLanguage.value) ?
                          settings.error :
                          [
                            ...settings.error,
                            { message: translate('InvalidUILanguage') }
                          ]}
```

markus101 (changes requested):

frontend/src/Settings/UI/UISettings.js
stevietv: this works when the language is valid, but throws an error when the language in the db is invalid:
```
UISettings.js:227 Uncaught TypeError: settings.error is not iterable
    at UISettings.render (UISettings.js:227:41)```

doing this works but not sure if thats right?

```
                      errors={
                        languages.some((language) => language.key === settings.uiLanguage.value) ?
                          settings.error :
                          [
                            { ...settings.error },
                            { message: translate('InvalidUILanguage') }
                          ]}
                          ```

markus101: Oops, should be `errors` not `error`

stevietv: same issue though
```
UISettings.js:227 Uncaught TypeError: settings.errors is not iterable
    at UISettings.render (UISettings.js:227:41)
    ```

markus101: Double oops, it's `settings.uiLanguage.errors`

stevietv: > Double oops, it's `settings.uiLanguage.errors`... not sure why I made up `settings.errors`

thanks, that was it! still learning my around js, react, and C#, and trying to find ways to use JetBrains Rider to investigate variables. Lot of learning :)

markus101 (approved):

PR #5968, opened 2023-08-27
Return original token when missing from translation

mynameisbogdan:
#### Database Migration
NO

#### Description
before
before

after
after


stevietv:
Good idea.
How did you even get into the select episode modal with no episodes selected?!

markus101 (approved):

PR #5973, opened 2023-08-28
Improve messaging in Authentication Required modal

mynameisbogdan:
#### Database Migration
NO


frontend/src/FirstRun/AuthenticationRequiredModalContent.js
```diff
@@ -95,27 +97,38 @@ function AuthenticationRequiredModalContent(props) {
...
-
-                <FormInputGroup
-                  type={inputTypes.TEXT}
-                  name="username"
-                  onChange={onInputChange}
-                  {...username}
-                />
-              </FormGroup>
+              {
+                authenticationEnabled ?
```
markus101, line 101: I don't think there is a lot of value in hiding this until the user makes a selection since they're going to have to fill it out regardless, whereas it might be hidden in settings because of how `External` works.

markus101 (changes requested):
Just my comment inline

frontend/src/FirstRun/AuthenticationRequiredModalContent.js
```diff
@@ -63,20 +63,22 @@ function AuthenticationRequiredModalContent(props) {
...
+                <FormLabel>{translate('AuthenticationMethod')}</FormLabel>
 
                 <FormInputGroup
                   type={inputTypes.SELECT}
                   name="authenticationMethod"
                   values={authenticationMethodOptions}
-                  helpText={translate('AuthenticationMethodHelpText')}
+                  helpText={translate('AuthenticationMethodHelpText', { appName: 'Sonarr' })}
+                  helpTextWarning={authenticationMethod.value === 'none' ? translate('AuthenticationMethodHelpTextWarning') : undefined}
+                  helpLink="https://wiki.servarr.com/sonarr/faq-v4#forced-authentication"
```
bakerboy448, line 81: this can be 
```suggestion
                  helpLink="https://wiki.servarr.com/sonarr/faq#forced-authentication"
```

v3 faq has an entry at the bottom with the link to the v4 faq; this means once v4 is stable we wont need to update this

markus101: 👍 Seems reasonable to future proof it

markus101 (approved):

PR #5989, opened 2023-08-31
Fixed: Qbit only map remote path map for completed items

bakerboy448:
#### Database Migration
NO

#### Description
reduce log spam

#### Todos
- Tests
- Wiki Updates

*


markus101 (changes requested):
All this logic should move to `GetImportItem` so it's only done after completion and just in time for Sonarr to import it.

Bye bye needlessly large image added (a few lines would have sufficed).

bakerboy448:
Other than the warning if savePath = ContentPath; it looks like this code is completely redundant for what's in `GetImportItem` ? Is that correct?

markus101:
It's a bit different because 2.6.1 and later has a better API, so checking the version would still make sense, but now it'd only be when getting the import item at the end.

bakerboy448:
should be better;  calling the version for each item seems redundant, but not sure where that can get called once and passed through.

src/NzbDrone.Core/Download/Clients/QBittorrent/QBittorrent.cs
```diff
@@ -328,9 +315,23 @@ public override void RemoveItem(DownloadClientItem item, bool deleteData)
 
         public override DownloadClientItem GetImportItem(DownloadClientItem item, DownloadClientItem previousImportAttempt)
         {
+            var properties = Proxy.GetTorrentProperties(item.DownloadId.ToLower(), Settings);
+            var savePath = new OsPath(properties.SavePath);
+
             // On API version >= 2.6.1 this is already set correctly
```
markus101, line 321: Comment is no longer relevant.

src/NzbDrone.Core/Download/Clients/QBittorrent/QBittorrent.cs
```diff
@@ -328,9 +315,23 @@ public override void RemoveItem(DownloadClientItem item, bool deleteData)
...
-            if (!item.OutputPath.IsEmpty)
+            var version = Proxy.GetApiVersion(Settings);
```
markus101, line 322: New line before if block and can group this with the vars above.

markus101 (changes requested):
A couple small things + a broken test needs to be fixed.

bakerboy448:
don't see why the test is failing.... :/

bakerboy448:
@markus101  what am I missing here for the test? 
Properties.ContentPath is empty
image

markus101:
What's the result of `_remotePathMappingService.RemapRemoteToLocal`? If that's not mocked then it's going to return a default value.

bakerboy448:
looks like the issue is there are no torrentproperties   so need to mock that?  even though the torrent itself is mocked?

don't really follow the test logic for what is / isn't needed here.

markus101:
The mocked response for `GetTorrentProperties` doesn't return anything for `ContentPath` so the comparison will always be false and the value for `properties.ContentPath` being null means and empty path gets returned.

I think this should be sufficient to resolve.

```
diff --git a/src/NzbDrone.Core.Test/Download/DownloadClientTests/QBittorrentTests/QBittorrentFixture.cs b/src/NzbDrone.Core.Test/Download/DownloadClientTests/QBittorrentTests/QBittorrentFixture.cs
index 3ae0edd84..f0dc2ee8e 100644
--- a/src/NzbDrone.Core.Test/Download/DownloadClientTests/QBittorrentTests/QBittorrentFixture.cs
+++ b/src/NzbDrone.Core.Test/Download/DownloadClientTests/QBittorrentTests/QBittorrentFixture.cs
@@ -137,7 +137,7 @@ protected virtual void GivenTorrents(List<QBittorrentTorrent> torrents)
             {
                 Mocker.GetMock<IQBittorrentProxy>()
                     .Setup(s => s.GetTorrentProperties(torrent.Hash.ToLower(), It.IsAny<QBittorrentSettings>()))
-                    .Returns(new QBittorrentTorrentProperties { SavePath = torrent.SavePath });
+                    .Returns(new QBittorrentTorrentProperties { ContentPath = torrent.ContentPath, SavePath = torrent.SavePath });
 
                 Mocker.GetMock<IQBittorrentProxy>()
                     .Setup(s => s.GetTorrentFiles(torrent.Hash.ToLower(), It.IsAny<QBittorrentSettings>()))
```

markus101 (approved):

markus101:
Turns out Qbit's torrent properties doesn't return `content_path` so I reverted it back closer to your original approach.

PR #5993, opened 2023-09-01
Translate Frontend Utilities

stevietv:
#### Database Migration
NO

#### Description
Translate Frontend Utilities

*


frontend/src/Utilities/Date/formatDateTime.js
```diff
@@ -10,15 +11,15 @@ function getRelativeDay(date, includeRelativeDate) {
   }
 
   if (isYesterday(date)) {
-    return 'Yesterday, ';
+    return `${translate('Yesterday')}, `;
```
markus101, line 14: This is going to be another place where we need to have more context to translate properly. In this case, I think we just want to return the relative day without the comma then translate it again in `formatDateTime`, which could look something like: 

```
const relativeDay = getRelativeDay(date, includeRelativeDay);
const formattedDate = moment(date).format(dateFormat);
const formattedTime = formatTime(date, timeFormat, { includeMinuteZero: true, includeSeconds });

if (relativeDay) {
  return translate('{relativeDay}, ${formattedDate} ${formattedTime}', { relativeDay, formattedDate, formattedTime });
}

return translate('${formattedDate} ${formattedTime}', { formattedDate, formattedTime });
```

This way different locales that prefer the relative day at the end or time before the date can be handled better.

frontend/src/Utilities/Number/formatAge.js
```diff
@@ -1,17 +1,19 @@
...
+
 function formatAge(age, ageHours, ageMinutes) {
   age = Math.round(age);
   ageHours = parseFloat(ageHours);
   ageMinutes = ageMinutes && parseFloat(ageMinutes);
 
   if (age < 2 && ageHours) {
     if (ageHours < 2 && !!ageMinutes) {
-      return `${ageMinutes.toFixed(0)} ${ageHours === 1 ? 'minute' : 'minutes'}`;
+      return `${ageMinutes.toFixed(0)} ${ageHours === 1 ? translate('Minute').toLowerCase() : translate('Minutes').toLowerCase()}`;
```
markus101, line 10: If we don't need `Minute`, `Hour`, `Day` and the plural values capitalized I think we're better off not calling the `toLowerCase` after translating

markus101 (changes requested):

frontend/src/Utilities/Number/formatAge.js
stevietv: we do have Day capitalized being used elsewhere, so I chose to just capitalize them all and lowercase them here.
Other wise we end up with a mix of `Day`  and `days`

stevietv: hmm, thinking about it more, what if a particular locale doesn't want lowercase for their indication. I will create new keys for this

markus101 (approved):

PR #6006, opened 2023-09-07
Fixed detection of newer versions of macOS

mynameisbogdan:
#### Database Migration
NO


thezoggy:
wonder if this detection works for betas as well.. since they do some funky stuff
https://stackoverflow.com/questions/40271041/detect-if-macos-version-is-a-beta-release

markus101 (approved):

Issue #6003, opened 2023-09-05
Add missing paging parameters in declared OpenAPI endpoints 


Ezwen:
### Is there an existing issue for this?

- [X] I have searched the existing open and closed issues

### Is your feature request related to a problem? Please describe

The following Sonarr API endpoints provide paginated responses:

- `/blocklist`
- `/wanted/cutoff`
- `/wanted/missing`
- `/history`
- `/log`
- `/queue`

In particular, the `page` parameter can be used to ask these endpoints a specific page.

For example, this query is valid and works well

```
$ curl   -H "X-Api-Key: XXXXX" -X GET "http://my.local.sonarr:8989/api/v3/queue?page=2"
```

However, in the [Sonarr OpenAPI definition file](https://github.com/Sonarr/Sonarr/blob/develop/src/Sonarr.Api.V3/openapi.json), the paging parameters that these endpoints accept are not declared nor documented.

A consequence it that tools that rely on this definition file, for instance the [sonarr-py library](https://github.com/devopsarr/sonarr-py/), do not provide any way to query specific pages in these endpoints.

### Describe the solution you'd like

I would expect the [Sonarr OpenAPI definition file](https://github.com/Sonarr/Sonarr/blob/develop/src/Sonarr.Api.V3/openapi.json) to define all parameters that all endpoints accept − and in particular the paging parameters mentioned above.

### Describe alternatives you've considered

I don't see any for now

### Anything else?

Issue opened on sonarr-py because their client code for the queue endpoint is not usable due to a lack of paging parameters: https://github.com/devopsarr/sonarr-py/issues/41

PR #6008, opened 2023-09-08
Paging params in API docs

markus101:
#### Database Migration
NO

#### Description
Adds paging params to API docs.

#### Todos
- [ ] Tests
- [ ] Wiki Updates

* #6003


src/NzbDrone.Host/Startup.cs
```diff
@@ -158,6 +158,8 @@ public void ConfigureServices(IServiceCollection services)
                 {
                     { apikeyQuery, Array.Empty<string>() }
                 });
+
+                c.DescribeAllParametersInCamelCase();
```
markus101, line 162: This gives camelCase names for properties on `PagingRequestResource` and other models

src/Sonarr.Api.V3/Blocklist/BlocklistController.cs
```diff
@@ -23,9 +23,9 @@ public class BlocklistController : Controller
 
         [HttpGet]
         [Produces("application/json")]
-        public PagingResource<BlocklistResource> GetBlocklist()
+        public PagingResource<BlocklistResource> GetBlocklist([FromQuery] PagingRequestResource paging)
         {
-            var pagingResource = Request.ReadPagingResourceFromRequest<BlocklistResource>();
+            var pagingResource = new PagingResource<BlocklistResource>(paging);
```
markus101, line 28: This drops support for `FilterKey` and `FilterValue` as of right now params for filtering, but we could probably add that behaviour back. This was originally added for backwards compatibility with v2 API.

src/Sonarr.Api.V3/openapi.json
```diff
@@ -381,6 +381,50 @@
...
+          },
+          {
+            "name": "sortKey",
+            "in": "query",
+            "schema": {
+              "type": "string"
+            }
+          },
+          {
+            "name": "sortDirection",
```
markus101, line 411: Getting a default for the sort direction requires inlining enum values, which doesn't seem great.

src/Sonarr.Api.V3/openapi.json
```diff
@@ -381,6 +381,50 @@
...
+            "name": "pageSize",
+            "in": "query",
+            "schema": {
+              "type": "integer",
+              "format": "int32",
+              "default": 10
+            }
+          },
+          {
+            "name": "sortKey",
```
markus101, line 404: This changes depending on the resource so no default value is set.

src/Sonarr.Api.V3/Blocklist/BlocklistController.cs
Qstick: I think this is likely fine with the v4 bump

Qstick (approved):

PR #6011, opened 2023-09-10
Use variable for App name in translations

Qstick:
#### Database Migration
NO

#### Description
Replace "Sonarr" with a variable for translations, this allows translation key sharing in Weblate with other apps without clobbering translations with the wrong app name.


src/NzbDrone.Core/Localization/Core/cs.json
```diff
@@ -5,21 +5,21 @@
     "AgeWhenGrabbed": "Stáří (kdy bylo získáno)",
     "Always": "Vždy",
     "AnalyseVideoFiles": "Analyzovat video soubory",
-    "AnalyseVideoFilesHelpText": "Extrahujte ze souborů informace o videu, jako je rozlišení, doba běhu a informace o kodeku. To vyžaduje, aby Sonarr četl části souboru, což může způsobit vysokou aktivitu disku nebo sítě během skenování.",
+    "AnalyseVideoFilesHelpText": "Extrahujte ze souborů informace o videu, jako je rozlišení, doba běhu a informace o kodeku. To vyžaduje, aby {appName} četl části souboru, což může způsobit vysokou aktivitu disku nebo sítě během skenování.",
     "ApplicationURL": "URL aplikace",
     "ApplicationUrlHelpText": "Externí adresa URL této aplikace včetně http(s)://, portu a základní adresy URL",
-    "AuthenticationMethodHelpText": "Vyžadovat uživatelské jméno a heslo pro přístup k Sonarru",
+    "AuthenticationMethodHelpText": "Vyžadovat uživatelské jméno a heslo pro přístup k {appName}u",
```
bakerboy448, line 11: /bSonarr/

Qstick: Nah, need to do all languages to fix things or they may never get changed.

frontend/src/Utilities/String/translate.ts
```diff
@@ -27,7 +27,9 @@ export default function translate(
   key: string,
   tokens?: Record<string, string | number | boolean>
 ) {
-  const translation = translations[key] || key;
+  let translation = translations[key] || key;
+
+  translation = translation.replace(/\{(appName)\}/gi, 'Sonarr');
```
markus101, line 32: I don't think we should replace tokens outside of the normal means, I'm not sure we need to pass in `appName` in every situation this is used, but should use tokens to replace it.

markus101 (changes requested):

frontend/src/Utilities/String/translate.ts
Qstick: Your're going to make me work for this one I see :D

markus101: Hey, I gave you a shortcut to just add it to tokens if it's not there 😄

markus101 (approved):

PR #6013, opened 2023-09-10
Preserve the protocol in Series Image

mynameisbogdan:
#### Database Migration
NO


markus101 (commented):
What else needs to be done in this PR?

markus101 (approved):

PR #6028, opened 2023-09-17
Fixed: Show correct error on unauthorized caps call

Qstick:
#### Database Migration
NO

#### Description
Currently CapabilitiesProvider doesn't throw on APIKey error, thus default categories are returned resulting in confusing messaging to the user. This change catches and throws the ApiKeyException that is generated instead of handling it with other exceptions

image

### DIFF
--- a/src/NzbDrone.Core/Indexers/Newznab/NewznabCapabilitiesProvider.cs
+++ b/src/NzbDrone.Core/Indexers/Newznab/NewznabCapabilitiesProvider.cs
@@ -7,6 +7,7 @@ using NzbDrone.Common.Cache;
 using NzbDrone.Common.Extensions;
 using NzbDrone.Common.Http;
 using NzbDrone.Common.Serializer;
+using NzbDrone.Core.Indexers.Exceptions;
 
 namespace NzbDrone.Core.Indexers.Newznab
 {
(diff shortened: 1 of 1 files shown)


src/NzbDrone.Core/Indexers/Newznab/NewznabCapabilitiesProvider.cs
```diff
@@ -73,6 +74,13 @@ private NewznabCapabilities FetchCapabilities(NewznabSettings indexerSettings)
                 _logger.Debug(ex, "Failed to parse newznab api capabilities for {0}", indexerSettings.BaseUrl);
                 throw;
             }
+            catch (ApiKeyException ex)
+            {
+                ex.WithData(response, 128 * 1024);
+                _logger.Trace("Unexpected Response content ({0} bytes): {1}", response.ResponseData.Length, response.Content);
+                _logger.Debug(ex, "Failed to parse newznab api capabilities for {0}, invalid key", indexerSettings.BaseUrl);
```
markus101, line 81:
```suggestion
                _logger.Debug(ex, "Failed to parse newznab api capabilities for {0}, invalid API key", indexerSettings.BaseUrl);
```

to more closely match the UI?

markus101 (approved):
Just the one suggestion.

Issue #1972, opened 2017-06-11
Put.io Download Client


markus101:
Cloud torrent provider, which will require transfer providers to be implemented: https://github.com/Sonarr/Sonarr/issues/282
API Docs: https://put.io/v2/docs/index.html

mepcotterell:
People have been asking about this for over a year. Has there been any progress?

markus101:
There are hundreds of open issues and this hasn't made it to the top of the priority list and as referenced here requires a transfer providers for importing which would is a lot of work in itself.

It's also been almost a year since this PR was opened and was abandoned: https://github.com/Sonarr/Sonarr/pull/1357

Ramblurr:
Would love to see this feature developed!

chierichetto:
+1

ZerithZA:
Good day, I was wondering if anyone has taken an interest in developing this. Would be nice to have integrated support which promotes sharing.

oimctony:
Hello ! lots of interest from me, but i have to admit i've very limited dev skills right now ! If anybody wants to work on it, i'll be glad to test.   ( FYI, the couchpotato implementation is quite well done, visible here : https://github.com/CouchPotato/CouchPotatoServer/tree/b538f9a08db9a0be1af2f298727834edeafa322e/couchpotato/core/downloaders/putio )

arthursoares:
Any updates on this?

oimctony:
Nope :(

dskvr:
I started on this an hour ago. However, be aware, I'm not proficient in C# and this is just going to be a little side project for me. No promises. I did find some reference material from a Windows Phone (RIP)  App that implemented Put.io v1 wrapper that may be able to help me out a little, but so far it's not looking great.

My plan is as follows:
1. Generate a C# SDK from the PUt.io 2.7.0 OAS3 Swagger File (https://app.swaggerhub.com/apis-docs/putio/putio/2.7.0)
2. Implement the SDK as a Download client
3. Learn about C# dependencies, workflows and tooling.

Wish me luck 🤞

oimctony:
Good luck man ! It would be incredible if you make it work :)


Le 21 mars 2020 à 06:43:05, Sandwich (notifications@github.com) a écrit:

I started on this an hour ago. However, be aware, I'm not proficient in C#
and this is just going to be a little side project for me. No promises. I
did find some reference material from a Windows Phone (RIP) App that
[...]

PR #6029, opened 2023-09-17
Put.io Download Client

michaelfeinbier:
#### Database Migration
NO

#### Description
This is a first draft to implement a Put.io Download client as described in https://github.com/Sonarr/Sonarr/issues/1972. I took the initial work of https://github.com/Sonarr/Sonarr/pull/1357 and fixed a lot of compile issue (as this one was pretty old). However its a bit far from done yet but I post it here to gather feedback and maybe clarify some questions or ideas as I am not very familiar with the .NET stack (but getting used to 😉 )

So Feedback welcome!

#### Todos
- [ ] Finish implementing the proxy
- [ ] Tests
- [ ] Wiki Updates

* https://github.com/Sonarr/Sonarr/issues/1972


michaelfeinbier:
Hey y'all
I'm making quite good progress on this :) However I need your input on one fundamental functionality. Put.io works in a way, that finished files end up in a remote bucket and they could be downloaded to a local dir via URL which we get from the Put.io API. So in order to make Sonarr importing the file to the library, we need to download these files.

In general, I do see 2 options here
### 1) Download "on-demand"
Basically, during iteration on all transfers in `IEnumerable<DownloadClientItem> GetItems()` we could check, if finished downloads on putio are already downloaded locally and are not marked as imported yet. If the file is finished on the remote and not yet downloaded we could initiate the download with
```c#
 await _httpClient.DownloadFileAsync(url, localPath);
```
However, obviously these downloads might take long and I don't think the `GetItems()` is really a good place to be misused as a download manager 😉 

### 2) Blackhole / `rclone` approach
The other option I do see here is to not care about downloading finished torrents from put.io at all but instead rely on 3rd party tools like `rclone` which can mount or sync the remote to a local folder. This is how usually all "workarounds" for put.io work and this was also the approach in the original PR #1357. Here we would just "fake" the `item.OutputPath` to the predicted path based on the settings of the client.

Overall I wanted to have a full-working client without the need of another tool but I am not sure if solution 1) is the best way to go here 🤔 
Do you guys maybe have other ideas how we could implement the downloading path? Or do we think solution 2) is good enough?

markus101:
Not relying on another service to handle the downloading would be nice, but that is effectively what we do for seedboxes already, so I don't think it's a huge deal to rely on that and consider doing something completely within Sonarr in the future. Option 1 in itself would be a pretty big undertaking to ensure it's functional and supports slower connections and/or interruptions.

sankalpsingha:
Hi all! Any update on this? Option 1 would be great tbh!

ZerithZA:
Personally for me using Rclone to mount the putio and have sonarr grab it from the putio folder or even have rclone download it and have sonarr look for when it's done fetching will work either way. The main problem I currently have is downloads that have stalled and having sonarr remove them and try a different one.

paulirish:
> In general, I do see 2 options here

@michaelfeinbier So glad you took this on!  

Perhaps obvious but a a challenge I see here is that with putio/seedbox-style-services, users may either want to **download** the files locally or B) ~**stream** them from an rclone mount. I prefer download because I have the storage, and trust in that more than a reliable connection. But at this point, I bet more putio users are used to the stream-style. 

My evolved putio+*arr setup is using https://gitlab.com/paulirish/krantor  configured as a torrent blackhole, then a `rclone move` cmd in cron. As a result, the current PR with [option 2 above](https://github.com/Sonarr/Sonarr/pull/6029#issuecomment-1725640651) roughly offers the same value as krantor, though the detailed download status is real nice!  

It seems like option 1 ~= running a download manager running within a recurring checkActiveItemStatus handler.. and that's less attractive then running an out-of-process long-running utility to handle downloads, resuming, etc etc.  (Is that right? Makes sense...)

Do any other seedboxy download clients try to offer options for users that either want to use the mounted location vs the post-download location?

markus101:

PR #6035, opened 2023-09-18
Fixed: Skip parsing releases without title

### LINKED ISSUE #6030: Interactive searches throwing error

### Current Behavior

When doing an interactive search for some episodes/seasons I'm met with the following error
image
In this case it was Game of Thrones Season 8.   I tried Season 7 and 6 and the result was the same.  

After doing some digging it was suggested to refresh the series and  remove-re-add.  Neither bore fruit.   I did find an issue report from a few years back saying it was addressed in a nightly build and nothing since. 

### Expected Behavior

Interactive searches should be presenting a list of results as opposed to "Search failed because its object reference not set to an instance of an object"

### Steps To Reproduce

1.  Run an interactive search for season 8 of game of thrones.  (have had it appear on other shows as well but wasnt keeping track of which until now)
2.  Receive an error as opposed to results. 

### Environment

```markdown
- OS: Windows 11
- Sonarr: 4.0.0.671
- Docker Install: No
- Using Reverse Proxy: No
- Browser: Brave 
- Database: Sqlite 3.36.0
```

Develop

### Trace Logs?

https://privatebin.net/?27c50d87a75cb1f9#AxbHZYbRTCNYnNgMtHKBLa8C5LcJkamtuCXzDAjbCYpy

### Anything else?

If you guys need more logs or anything just lmk.

mynameisbogdan:
#### Database Migration
NO

* Fixes #6030

### DIFF
--- a/src/NzbDrone.Common/Extensions/PathExtensions.cs
+++ b/src/NzbDrone.Common/Extensions/PathExtensions.cs
@@ -171,7 +171,7 @@ namespace NzbDrone.Common.Extensions
         {
             if (text.IsNullOrWhiteSpace())
             {
-                throw new ArgumentNullException("text");
+                throw new ArgumentNullException(nameof(text));
             }
 
             return text.IndexOfAny(Path.GetInvalidPathChars()) >= 0;
--- a/src/NzbDrone.Core/Indexers/HttpIndexerBase.cs
+++ b/src/NzbDrone.Core/Indexers/HttpIndexerBase.cs
@@ -307,9 +307,17 @@ namespace NzbDrone.Core.Indexers
 
         protected virtual bool IsValidRelease(ReleaseInfo release)
         {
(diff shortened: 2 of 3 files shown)


markus101 (changes requested):
I don't think we even want to get this far as we're never going to be able to do anything with it. If the release title is null/empty we should log a message and skip it earlier in the process.

markus101:
That makes sense, but I'm not sure how practical that is, in this case the UI showed an error which lead to the GHI. If we "fix the glitch" and return the result in this case it would have been buried at the bottom of the list or potentially not even shown if it was filtered to only show season packs.

Instead of relying on users can they be caught on the Jackett/Prowlarr side are report to sentry/other error handling to look into?

markus101 (approved):

markus101:
Thanks Bogdan!

PR #6045, opened 2023-09-21
Fix SabNZBd history retention to allow at least 14 days

### LINKED ISSUE #6044: SABnzbd health check is too eager

### Current Behavior

Sonarr complains that "Download client SABnzbd is set to remove completed downloads. This can result in downloads being removed from your client before Sonarr can import them.". According to the code [here](https://github.com/Sonarr/Sonarr/blame/c7824bb593291634bf14a5f7aa689666969b03bf/src/NzbDrone.Core/Download/Clients/Sabnzbd/Sabnzbd.cs#L279), the check fails if history retention is not disabled (set to 0).

My history_retention is set to 31 days which seems reasonable to me. Sonarr may not be the only client being served by SABnzbd and I think I need to configure it to not keep everything forever just so that Sonarr does not complain. 31 days should be enough time to import anything.

The change in question was added in #6020.

### Expected Behavior

Don't complain about reasonable SABnzbd configuration. Maybe the check should be against "-1" which I think means that everything is removed immediately (see https://github.com/sabnzbd/sabnzbd/blob/1e4ef9c38123abfe3701382f9d4646d288f6d52f/sabnzbd/database.py#L247).

### Steps To Reproduce

1. Use SABnzbd.
2. Set history retention to 31 days (Switches -> Post processing)
3. Check Sonarr status messages.

### Environment

```markdown
- OS: Ubuntu 20.04
- Sonarr: 4.0.0.682
- Docker Install: yes
- Using Reverse Proxy: yes
- Browser: 
- Database:
```

Develop

### Trace Logs?

n/a

### Anything else?

_No response_

stevietv:
#### Database Migration
 NO

#### Description
Fixes #6044 

allows for retention of at least 14 days also

*

### DIFF
--- a/src/NzbDrone.Core/Download/Clients/Sabnzbd/Sabnzbd.cs
+++ b/src/NzbDrone.Core/Download/Clients/Sabnzbd/Sabnzbd.cs
@@ -276,7 +276,7 @@ namespace NzbDrone.Core.Download.Clients.Sabnzbd
                 status.OutputRootFolders = new List<OsPath> { _remotePathMappingService.RemapRemoteToLocal(Settings.Host, category.FullPath) };
             }
 
-            if (config.Misc.history_retention.EndsWith("d"))
+            if (!string.IsNullOrEmpty(config.Misc.history_retention) && config.Misc.history_retention.EndsWith("d"))
             {
                 int.TryParse(config.Misc.history_retention.AsSpan(0, config.Misc.history_retention.Length - 1),
                     out var daysRetention);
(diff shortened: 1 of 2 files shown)


src/NzbDrone.Core/Download/Clients/Sabnzbd/Sabnzbd.cs
```diff
@@ -276,7 +276,16 @@ public override DownloadClientInfo GetStatus()
                 status.OutputRootFolders = new List<OsPath> { _remotePathMappingService.RemapRemoteToLocal(Settings.Host, category.FullPath) };
             }
 
-            status.RemovesCompletedDownloads = config.Misc.history_retention != "0";
+            if (!string.IsNullOrEmpty(config.Misc.history_retention) && config.Misc.history_retention.EndsWith("d"))
```
markus101, line 279:
```suggestion
            if (config.Misc.history_retention.IsNotNullOrWhiteSpace() && config.Misc.history_retention.EndsWith("d"))
```

markus101 (approved):
Just a suggestion to use the extension method, but otherwise looks good

src/NzbDrone.Core/Download/Clients/Sabnzbd/Sabnzbd.cs
stevietv: done 👍🏻

stevietv:
Also just went ahead and added some tests for Sab

PR #6059, opened 2023-09-30
Fix: return '#' as first character for non alphanumeric

### LINKED ISSUE #6055: Series TitleFirstCharacter Special Characters

Hi there, new Sonarr user here, please take this into consideration if I overlooked something obvious ;)

I have ran into a slight issue adding series which start with a special character ([example](https://thetvdb.com/series/the-80s-greatest-us))

Sonarr is configured to use the following folder format:
`{Series TitleFirstCharacter}/{Series CleanTitleYear}`

On manually importing the related files, Sonarr will try to create the _{Series TitleFirstCharacter}_ folder, and fails here because, well, it is a special character, and (I assume) internal code catches it...

Feature request:
Add another label: **{Series CleanTitleFirstCharacter}**
This label should use the first character from the {Series CleanTitle} label, effectively disallowing special character folders

I don't really see any alternative solutions, but I'm very open to suggestions!

### Anything else?

I searched the Issue tracker for similar requests, but did not see any with the terms I used.
The only comment related I found was [this one](https://github.com/Sonarr/Sonarr/issues/5410#issuecomment-1416878727), but I don't think it is relevant really?

Thank you for your time and this wonderful collection of software!

stevietv:
#### Database Migration
NO

#### Description
This returns '#' as the first character to use in the `{Series TitleFirstCharacter}` token when that character is not alphanumeric

#### Todos
- [X] Tests

* fixes #6055 6055

### DIFF
--- a/src/NzbDrone.Core/Organizer/FileNameBuilder.cs
+++ b/src/NzbDrone.Core/Organizer/FileNameBuilder.cs
@@ -381,9 +381,16 @@ namespace NzbDrone.Core.Organizer
         {
             if (char.IsLetterOrDigit(title[0]))
             {
-                return title.Substring(0, 1).FirstCharToUpper();
+                return title.Substring(0, 1).ToUpper();
             }
 
+            // try the second character if the first was non alphanumeric
+            if (char.IsLetterOrDigit(title[1]))
+            {
+                return title.Substring(1, 1).ToUpper();
+            }
+
+            // default to "_" if no alphanumeric character can be found in the first 2 positions
             return "_";
         }
 
--- a/src/NzbDrone.Core/Organizer/FileNameBuilder.cs
+++ b/src/NzbDrone.Core/Organizer/FileNameBuilder.cs
@@ -384,7 +384,7 @@ namespace NzbDrone.Core.Organizer
                 return title.Substring(0, 1).ToUpper().RemoveAccent();
             }
 
-            // try the second character if the first was non alphanumeric
+            // Try the second character if the first was non alphanumeric
             if (char.IsLetterOrDigit(title[1]))
             {
                 return title.Substring(1, 1).ToUpper().RemoveAccent();
--- a/src/NzbDrone.Core/Organizer/FileNameBuilder.cs
+++ b/src/NzbDrone.Core/Organizer/FileNameBuilder.cs
@@ -381,13 +381,13 @@ namespace NzbDrone.Core.Organizer
         {
             if (char.IsLetterOrDigit(title[0]))
             {
-                return title.Substring(0, 1).ToUpper();
+                return title.Substring(0, 1).ToUpper().RemoveAccent();
             }
 
             // try the second character if the first was non alphanumeric
             if (char.IsLetterOrDigit(title[1]))
             {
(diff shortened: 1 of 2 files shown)


stevietv:
> [#5909 (review)](https://github.com/Sonarr/Sonarr/pull/5909#pullrequestreview-1574370534)

I see that, though I'm not really sure what other char we would want to use. I would see these as options:

- `-`
- `!`
- `@`
- `+`

markus101:
Underscore or Dash maybe?

I believe we've followed https://www.niso.org/sites/default/files/2017-08/tr03.pdf for some sorting rules before, but it's not super clear to me at a glance what the expected sorting would be in this case.

stevietv:
I switched it to `_` - this also sits better with me.
Thinking about this some more though, should a show like `The '80s Greatest` really be stored under `_/The '80s Greatest` or does it really belong under the `8` folder.  It feels kind of contraintuitive to me to go search for that show under the _ instead of 8.

I.e. should we just be taking the first LetterOrDigit from the title and filing it under there?

markus101:
Yeah, I was thinking about what would a library do, but it's probably a little more complex for us as we don't maintain a separate manually set title for for sorting and it's best effort.

I'd be in favour of see what it'd looks like to use the first alphanumeric character, but that could be funky for some series and we may need to limit how far it looks for that. Maybe just second character to differentiate between special characters used as the title and an apostrophe as a way to shorten the word?

stevietv:
I've gone for now taking the first, or second char, and defaulting to _ if neither are good.
I'm wondering now if we should also remove diacritics with `RemoveAccent` also?

markus101:
The doc I linked previously said they should be removed for alphabetical sorting, which I think makes sense here as well, but not sure what that will look like with other character sets.

stevietv:
now removing diacritics. the test present in a different char set still passes.

src/NzbDrone.Core/Organizer/FileNameBuilder.cs
```diff
@@ -377,6 +377,23 @@ public static string TitleWithoutYear(string title)
...
         }
 
+        public static string TitleFirstCharacter(string title)
+        {
+            if (char.IsLetterOrDigit(title[0]))
+            {
+                return title.Substring(0, 1).ToUpper().RemoveAccent();
+            }
+
+            // try the second character if the first was non alphanumeric
```
markus101, line 387:
```suggestion
            // Try the second character if the first was non alphanumeric
```

src/NzbDrone.Core/Organizer/FileNameBuilder.cs
```diff
@@ -377,6 +377,23 @@ public static string TitleWithoutYear(string title)
...
+            if (char.IsLetterOrDigit(title[1]))
+            {
+                return title.Substring(1, 1).ToUpper().RemoveAccent();
+            }
+
+            // default to "_" if no alphanumeric character can be found in the first 2 positions
```
markus101, line 393:
```suggestion
            // Default to "_" if no alphanumeric character can be found in the first 2 positions
```

markus101 (approved):

PR #6064, opened 2023-10-02
New: Additional tooltips

stevietv:
#### Database Migration
NO

#### Description

chrome_2023-10-02_21-49-23

*

### DIFF
--- a/frontend/src/Helpers/Props/icons.js
+++ b/frontend/src/Helpers/Props/icons.js
@@ -162,7 +162,6 @@ export const KEYBOARD = farKeyboard;
 export const LOGOUT = fasSignOutAlt;
 export const MANAGE = fasListCheck;
 export const MEDIA_INFO = farFileInvoice;
-export const MENU = fasBars;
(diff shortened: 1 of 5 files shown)


frontend/src/Helpers/Props/icons.js
```diff
@@ -162,6 +162,7 @@ export const KEYBOARD = farKeyboard;
 export const LOGOUT = fasSignOutAlt;
 export const MANAGE = fasListCheck;
 export const MEDIA_INFO = farFileInvoice;
+export const MENU = fasBars;
```
markus101, line 165: This icon is already used for the collapsing sidebar on mobile, so I don't think we want to use it here.

markus101 (changes requested):

frontend/src/Helpers/Props/icons.js
stevietv: I've reverted the Icon change

markus101 (approved):

PR #6082, opened 2023-10-10
Use named tokens for backend translations

### LINKED ISSUE #6051: Backend translations don't handle tokens

### Current Behavior

Since commit ddabe66 there are a number of translations where `{appName}` has been added but is not handled.
This is because the translation takes place in the backend via [GetLocalizedString](https://github.com/Sonarr/Sonarr/blob/583eb52ddc01b608ab6cb17e863a8830c17b7b75/src/NzbDrone.Core/Localization/LocalizationService.cs#L61).

This method doesn't handle tokens. Currently only a few strings are used in backend translations, and only in healthcheck messages.

for example: `No indexers available with RSS sync enabled, {appName} will not grab new releases automatically`

### Expected Behavior

tokens are replaced in the translation string before being passed to the frontend

### Steps To Reproduce

_No response_

### Environment

```markdown
- OS: agnostic
- Sonarr: 4.0.0.682+
- Docker Install: agnostic
- Using Reverse Proxy: agnostic
- Browser: agnostic
- Database: agnostic
```

Develop

### Trace Logs?

N/A

### Anything else?

_No response_

markus101:
#### Database Migration
NO

#### Description
Migrates backend translations to named tokens instead of numbered tokens to match the front end and replace `{appName}` properly.

* Closes #6051


src/NzbDrone.Core.Test/Localization/LocalizationServiceFixture.cs
```diff
@@ -32,35 +32,28 @@ public void should_get_string_in_dictionary_if_lang_exists_and_string_exists()
         [Test]
         public void should_get_string_in_default_language_dictionary_if_no_lang_country_code_exists_and_string_exists()
```
stevietv, line 33: Is the test name still relevant?

stevietv (commented):
looks great!, one small question about a test only.

src/NzbDrone.Core.Test/Localization/LocalizationServiceFixture.cs
markus101: I don't know if it was ever relevant, but updated it it to the bare minimum, getting the translation in French.

Issue #6047, opened 2023-09-25
Trace Logs do not filter out private tracker info


Rubicj:
### Is there an existing issue for this?

- [x] I have searched the existing open and closed issues

### Current Behavior

Reopen of https://github.com/Sonarr/Sonarr/issues/4341

The return from QBittorrent includes secrets about trackers. These secrets seem to not be redacted properly. I've attached a manually redacted file where I've replaced my keys and the website names.

### Expected Behavior

```
"tracker":"https://t.foo.com/tracker.php/p6jqYMVbKn7FNoAz7NLdU6JRv548M30V/announce",
"magnet_uri":"magnet:?xt=urn:btih:4da7319561bf054dfd6135c284587b844ec16efc&dn=Nikki%20St.%20Crowe%20-%20Midnight%20Harbor%2003%20-%20Hot%20Vampire%20Next%20Door%20Season%20Three.epub&tr=https%3a%2f%2ft.foo.com%2ftracker.php%2fp6jqYMVbKn7FNoAz7NLdU6JRv548M30V%2fannounce"
```
should be replace with
```
"tracker":"[REMOVED]",
"magnet_uri":"[REMOVED]"
```

### Steps To Reproduce

1. Add QBittorrent as a download client
2. Turn on Trace logging
3. Observe that the full logs are printed without redaction

### Environment

```markdown
- OS: Unraid
- Sonarr: 4.0.0.682 and Readarr 0.3.5.2217
- Docker Install: Yes
- Using Reverse Proxy: No
- Browser: Chrome
- Database: Default
```

### What branch are you running?

Develop

### Trace Logs?

[fakelogs.txt](https://github.com/Sonarr/Sonarr/files/12719917/fakelogs.txt)
Note that I replaced the actual 32 digit passkey with one I made up, `p6jqYMVbKn7FNoAz7NLdU6JRv548M30V`

### Anything else?

The regex used is visible in https://github.com/Sonarr/Sonarr/commit/e67864fecb4a66792152ff8fb0889ee1a176919f , but it doesn't look like normal regex to me -- the bars and forward slashes are confusing me.

I do see that it's only matching lowercase letters, and my secret contains uppercase.  That might be it? In my opinion, it's actually preferrable to mask out the whole website name -- I'd actually expect it to redact the entirety of the magnet uri and the tracker.

Rubicj:
[...]

PR #6091, opened 2023-10-12
Fixed: Ignore case when cleansing announce URLs

mynameisbogdan:
#### Database Migration
NO

#### Description
Update the announce URL cleanse regex for trackers that use generated hashes with upper case chars in the passkeys.


markus101:
Closes https://github.com/Sonarr/Sonarr/issues/6047

markus101 (approved):

PR #6119, opened 2023-10-22
New: Option to control which new series episodes get monitored

### LINKED ISSUE #5083: Don't automatically monitor latest season/future seasons When Adding Series

Right now, there is no way to monitor select seasons/episodes without monitoring future episodes.
This leads to confusing UX when adding new series using Monitor -> Pilot Episode and similar options. Its not clear that these options lead to a temporary state that can change without user interaction. The existence of a distinct "Future Episodes" option further pushed me as a user towards wrong assumptions.

A user limits monitoring to the first episode/season because they don't want to commit more resources until they watched the show some time later. Fetching more than that doesn't make sense.

But even for users aware of this behavior, it requires manual intervention for automated list imports just to stick with the episodes that were asked for.

Make monitored/active and "monitor future episodes" distinct states of a series.

Maybe there could be a bogus item like an empty season representing future episodes. I could imagine this blending in well with the current UI.

Instead of monitoring pilots, first/last season etc. upon _add_, allow for a one-off search of these items without monitoring them.
That way the user could get exactly what he asked for without having to manually unmonitor the series afterwards.
This is an inferior solution for the lack of monitoring upgrades.

### Anything else?

.

Qstick:
#### Database Migration
YES (200)

#### Description
Adds setting to prevent future episodes from being monitored automatically. 

#### Todos
- [ ] Tests
- [ ] Wiki Updates

* Fixes #5083


frontend/src/AddSeries/SeriesMonitorNewItemsOptionsPopoverContent.js
```diff
@@ -0,0 +1,22 @@
...
+import DescriptionList from 'Components/DescriptionList/DescriptionList';
+import DescriptionListItem from 'Components/DescriptionList/DescriptionListItem';
+import translate from 'Utilities/String/translate';
+
+function SeriesMonitorNewItemsOptionsPopoverContent() {
+  return (
+    <DescriptionList>
+      <DescriptionListItem
+        title={translate('MonitorAllEpisodes')}
+        data="Monitor all new episodes"
```
markus101, line 11: I think this is a bit misleading because episodes should follow the season's monitored state and this should be more about monitoring new seasons.

Having this also control episodes is going to make the interaction with `MonitoringOptions` confusing as well.

frontend/src/Utilities/Series/monitorNewItemsOptions.js
```diff
@@ -0,0 +1,18 @@
+import translate from 'Utilities/String/translate';
+
+const monitorNewItemsOptions = [
+  {
+    key: 'all',
+    get value() {
+      return translate('MonitorAllScenes');
```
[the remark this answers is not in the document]
Qstick: Should be Monitor all Episodes there

frontend/src/AddSeries/SeriesMonitorNewItemsOptionsPopoverContent.js
Qstick: Yea, makes sense to change to seasons for Sonarr, all the others are just date based releases with no grouping

PR #6120, opened 2023-10-22
Set busy timeout for SQLite

mynameisbogdan:
#### Database Migration
NO

#### Description
Adding a busy timeout in order to avoid db locks. I used 100 for more than week and 0 db locks since. 

https://www.sqlite.org/c3ref/busy_timeout.html


Qstick:
Saw your Radarr PR to bump to 1000 now, what's the reason for going higher? Should we align here?

markus101:
Are we confirmed to be good with 100?

markus101 (approved):

PR #6138, opened 2023-10-29
Fixed: Blocklisting torrents from indexers that do not provide torrent hash

### LINKED ISSUE #6108: Blocklist does not block GrabIB: Sonarr grabbing same Download from different Indexers (Same GrabID)

### Current Behavior

I believe this is related to this bug, which was marked as "resolved" by the bot: https://github.com/Radarr/Radarr/issues/3863

It directly impacts my application that tries to detect stalled downloads and replace them with good ones, and we tried to get to the bottom of the problem here: https://github.com/ManiMatter/decluttarr/issues/8

What I try to accomplish:
- Download a TV Show (or Movie in Radarr, for that matter, same bug over there) and block a certain GrabID (for instance because it it is stalled), and expect that Sonarr then picks up a different torrent

What I do:
- I use the function "Blocklist Release - Starts a search for this episode again and prevents this release from being grabbed again" under "Activity/Queue"

What happens:
- Sonarr adds the release to the blocklist, but still grabs the same download again

### Expected Behavior

I expect that once a GrabID is marked as blocked, no torrent with the identical grabID should be picked.

**What we believe the root cause is:**
Blocking seems to take into consideration the friendly name of the torrent and the indexer.
That is a problem, since the very same download may be present on many different indexers, with slightly different names.

**How we believe this could be fixed:**
Blocking should merely take the GrabID into account. If a movie or tvshow has block listed a GrabID, no torrent, regardless on what name or from what indexer, should be grabbed

### Steps To Reproduce

We spent quite some time trying to figure out what's going on here:
https://github.com/ManiMatter/decluttarr/issues/8#issuecomment-1765332736

For testing, we have created is a python script that deletes all downloads in Sonarr, and adds them to the blocklist:
https://pastebin.com/raw/wweQWhYt

What we have observed that running the same script within 3 hours results in the same queueItemID and GrabID to be present again (eventhough they were first deleted and blocked):

> 2023-10-15 18:41:07,662 [VERBOSE]: queue id: 855858022, download id: 68F83966142792A0884C9B16813C0F903273C469, action: deleted, title:
> The.Breaker.Upperers.2018.1080p.Bluray.X264-EVO[EtHD]
> 2023-10-15 18:41:08,744 [VERBOSE]: **************************************************
> 2023-10-15 18:41:08,744 [VERBOSE]: END - REFRESHING QUEUE
> 2023-10-15 18:41:08,756 [VERBOSE]: End - Queue is empty
> 
> 2023-10-15 21:50:18,778 [VERBOSE]: queue id: 855858022, download id: 68F83966142792A0884C9B16813C0F903273C469, action: deleted, title: The.Breaker.Upperers.2018.1080p.Bluray.X264-EVO[EtHD]
> 2023-10-15 21:50:26,394 [VERBOSE]: **************************************************
> 2023-10-15 21:50:26,394 [VERBOSE]: END - REFRESHING QUEUE
> 2023-10-15 21:50:26,409 [VERBOSE]: End - Queue is empty

What you can see from the screenshots that [MikeFalcor](https://github.com/MikeFalcor) posted [here](https://github.com/ManiMatter/decluttarr/issues/8#issuecomment-1765332736) the same GrabID is picked up 50+ times, just to be blocked again.

### Environment

```markdown
- OS: Ubuntu
- Sonarr: Nightly
- Docker Install: Yes
- Using Reverse Proxy: Traefik
- Browser: Chrome
- Database: Nightly
```

Main

### Trace Logs?

I believe that is not applicable, since the execution of the calls to delete and block the downloads work correctly

> 2023-10-12T20:55:39.193501836Z [DEBUG]: http://172.20.0.101:7878 "DELETE /api/v3/queue/2072900635?removeFromClient=true&blocklist=True HTTP/1.1" 200 0

If needed, we can run trace logs, but we don't think they will add much benefit. The problem here seems to conceptual with the design of what is considered when determining if something has been blocked previously or not.

### Anything else?

_No response_

markus101:
#### Database Migration
NO

#### Description
A lot of low end indexers accessed through torznab don't provide hashes during parsing and we're forced to get it from the torrent/magnet link, which currently means it's not stored in history and not available for blocklisting.

This PR adds/fixes a few things:

- Blocklisting the last grabbed history item instead of the first (which would use the wrong title/indexer combination if the same release was re-grabbed from a different indexer)
- The hash from the torrent/magnet to the release so it's stored in history and used when blocklisting
- An option on indexers that is used after grabbing the torrent, but before sending it to the client to allow *arr to reject it

* Closes #6108

### DIFF
--- a/src/NzbDrone.Core/Indexers/BroadcastheNet/BroadcastheNetSettings.cs
+++ b/src/NzbDrone.Core/Indexers/BroadcastheNet/BroadcastheNetSettings.cs
@@ -37,6 +37,9 @@ namespace NzbDrone.Core.Indexers.BroadcastheNet
         [FieldDefinition(3)]
         public SeedCriteriaSettings SeedCriteria { get; set; } = new SeedCriteriaSettings();
 
+        [FieldDefinition(4, Type = FieldType.Checkbox, Label = "Reject Blocklisted Torrent Hashes While Grabbing", HelpText = "If a torrent is blocked by hash it may not properly be rejected during RSS/Search for some indexers, enabling this will allow it to be rejected after the torrent is grabbed, but before it is sent to the client.", Advanced = true)]
+        public bool RejectBlocklistedTorrentHashesWhileGrabbing { get; set; }
+
         public NzbDroneValidationResult Validate()
         {
             return new NzbDroneValidationResult(Validator.Validate(this));
--- a/src/NzbDrone.Core/Blocklisting/BlocklistService.cs
+++ b/src/NzbDrone.Core/Blocklisting/BlocklistService.cs
@@ -15,6 +15,7 @@ namespace NzbDrone.Core.Blocklisting
     public interface IBlocklistService
     {
         bool Blocklisted(int seriesId, ReleaseInfo release);
+        bool BlocklistedTorrentHash(int seriesId, string hash);
         PagingSpec<Blocklist> Paged(PagingSpec<Blocklist> pagingSpec);
         void Block(RemoteEpisode remoteEpisode, string message);
         void Delete(int id);
@@ -59,6 +60,12 @@ namespace NzbDrone.Core.Blocklisting
 
             return blocklistedByTitle.Where(b => b.Protocol == DownloadProtocol.Usenet)
                                      .Any(b => SameNzb(b, release));
+        }
+
+        public bool BlocklistedTorrentHash(int seriesId, string hash)
+        {
+            return _blocklistRepository.BlocklistedByTorrentInfoHash(seriesId, hash).Any(b =>
+                b.TorrentInfoHash.Equals(hash, StringComparison.InvariantCultureIgnoreCase));
         }
 
         public PagingSpec<Blocklist> Paged(PagingSpec<Blocklist> pagingSpec)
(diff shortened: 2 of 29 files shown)


src/NzbDrone.Core/Indexers/BroadcastheNet/BroadcastheNetSettings.cs
```diff
@@ -37,6 +37,9 @@ public BroadcastheNetSettings()
         [FieldDefinition(3)]
         public SeedCriteriaSettings SeedCriteria { get; set; } = new SeedCriteriaSettings();
 
+        [FieldDefinition(4, Type = FieldType.Checkbox, Label = "Reject Blocklisted Torrent Hashes While Grabbing", HelpText = "If a torrent is blocked by hash it may not properly be rejected during RSS/Search for some indexers, enabling this will allow it to be rejected after the torrent is grabbed, but before it is sent to the client.", Advanced = true)]
```
markus101, line 40: Both the `Label` and `HelpText` feel pretty wordy, but wanted to explicit

markus101:
If the hash is different then the rejection after grabbing won't apply. The first fix I mentioned applies for the "same" torrent grabbed from different indexers, but it needs to be blocked per indexer.

markus101:
Blocking based on GUID would only work if the GUID is the same though, which for different indexers or even the same torrent with a different name on the same tracker (assuming this happens) won't often be the case. I think we looked at GUID blocking before with usenet, but since it didn't add much over title blocking we never used it, I don't see much of an advantage with torrents either.

PR #6148, opened 2023-11-02
New: Parser for Turkish releases

Xetera:
#### Database Migration
NO

#### Description
Some Turkish release groups like TURG and turkseed like to localize episode names and add air dates to titles which sometimes confuses Sonarr, making detection flaky because it can't make up its mind between daily show vs absolute episodes.

The regex for this is specifically placed before this one:
```js
/^(?<title>.+?)?\W*(?<airmonth>[0-1][0-9])[-_. ]+(?<airday>[0-3][0-9])[-_. ]+(?<airyear>\d{4})(?!\d+)/
```
since that's what made parsing flaky in the first place.

I didn't include the air dates in the pattern because release titles are often very inconsistent. I don't think that's going to cause false positives though.

#### Todos
- [x] Tests
- [ ] Wiki Updates

*

### DIFF
--- a/src/NzbDrone.Core/Parser/Parser.cs
+++ b/src/NzbDrone.Core/Parser/Parser.cs
@@ -330,8 +330,8 @@ namespace NzbDrone.Core.Parser
                 new Regex(@"^(?<title>.+?)?\W*(?<airyear>\d{4})[-_. ]+(?<airmonth>[0-1][0-9])[-_. ]+(?<airday>[0-3][0-9])(?![-_. ]+[0-3][0-9])",
                     RegexOptions.IgnoreCase | RegexOptions.Compiled),
 
-                new Regex(@"^(?<title>.+?)\b(?<absoluteepisode>\d{1,4})(?:\.\s*|\s+)(?:BLM|B[oö]l[uü]m)", RegexOptions.IgnoreCase | RegexOptions.Compiled),
                 // Turkish tracker releases (01 BLM, 3. Blm, 04.Bolum, etc)
+                new Regex(@"^(?<title>.+?)[_. ](?<absoluteepisode>\d{1,4})(?:[_. ]+)(?:BLM|B[oö]l[uü]m)", RegexOptions.IgnoreCase | RegexOptions.Compiled),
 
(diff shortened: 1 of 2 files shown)


src/NzbDrone.Core/Parser/Parser.cs
```diff
@@ -330,6 +330,9 @@ public static class Parser
                 new Regex(@"^(?<title>.+?)?\W*(?<airyear>\d{4})[-_. ]+(?<airmonth>[0-1][0-9])[-_. ]+(?<airday>[0-3][0-9])(?![-_. ]+[0-3][0-9])",
                     RegexOptions.IgnoreCase | RegexOptions.Compiled),
 
+                // Turkish tracker releases
+                new Regex(@"^(?<title>.+?)\b(?<absoluteepisode>\d{1,4})(?:\.\s*|\s+)(?:BLM|B[oö]l[uü]m)", RegexOptions.IgnoreCase | RegexOptions.Compiled),
```
markus101, line 334:
```suggestion
                new Regex(@"^(?<title>.+?)[_. ](?<absoluteepisode>\d{1,4})(?:[_. ]+)(?:BLM|B[oö]l[uü]m)", RegexOptions.IgnoreCase | RegexOptions.Compiled),
```

src/NzbDrone.Core/Parser/Parser.cs
```diff
@@ -330,6 +330,9 @@ public static class Parser
                 new Regex(@"^(?<title>.+?)?\W*(?<airyear>\d{4})[-_. ]+(?<airmonth>[0-1][0-9])[-_. ]+(?<airday>[0-3][0-9])(?![-_. ]+[0-3][0-9])",
                     RegexOptions.IgnoreCase | RegexOptions.Compiled),
 
+                // Turkish tracker releases
```
markus101, line 333: Please update to include an example like we do above: `(01 BLM or 79.Bolum)`

markus101 (changes requested):
A small improvement on the regex to allow for `_` to be used as a separator which comes up if spaces are replaced with them.

markus101 (approved):

PR #6155, opened 2023-11-05
Fix translation for GrabReleaseMessageText

ZephireNZ:
#### Database Migration
NO

#### Description
Translation uses `appName` but this parameter was not being passed.

#### Todos
- [ ] Tests
- [ ] Wiki Updates

*


stevietv:
appName is translated without passing the token in: https://github.com/Sonarr/Sonarr/blob/165e3dbae8c9ba478d4aac4bd1b8514f54f0dfec/frontend/src/Utilities/String/translate.ts#L28

stevietv:
I see now that indeed this token doesn't work, and we need to change translate.ts
translate.ts currently only works when there are no tokens passed in.

PR #6156, opened 2023-11-05
Fix: translate for empty appName token

stevietv:
#### Database Migration
NO

#### Description
when there tokens are passed to translate function appName is not added to the list, thus is skipped in multiple translation strings.
translate() function now defaults to it (when no tokens are passed) and adds it when tokens are passed.

* #6155

### DIFF
--- a/frontend/src/Utilities/String/translate.ts
+++ b/frontend/src/Utilities/String/translate.ts
@@ -25,15 +25,13 @@ export async function fetchTranslations(): Promise<boolean> {
 
 export default function translate(
   key: string,
-  tokens: Record<string, string | number | boolean> = { appName: 'Sonarr' }
+  tokens: Record<string, string | number | boolean> = {}
 ) {
   const translation = translations[key] || key;
 
-  if (tokens) {
-    return translation.replace(/\{([a-z0-9]+?)\}/gi, (match, tokenMatch) =>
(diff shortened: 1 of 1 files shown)


ZephireNZ:
Instead of always overriding this field, would it be better to set it only when it's not already set?

stevietv:
> Instead of always overriding this field, would it be better to set it only when it's not already set?

We always want this field to be set to "Sonarr", there shouldn't be another case where the token doesn't equal Sonarr if passed in.

stevietv:
oh and @ZephireNZ thanks for spotting the error!!

frontend/src/Utilities/String/translate.ts
```diff
@@ -29,11 +29,9 @@ export default function translate(
 ) {
   const translation = translations[key] || key;
 
-  if (tokens) {
-    return translation.replace(/\{([a-z0-9]+?)\}/gi, (match, tokenMatch) =>
-      String(tokens[tokenMatch] ?? match)
-    );
-  }
+  tokens.appName = 'Sonarr';
```
markus101, line 32: Should update line 28 to only initialize the object.

markus101 (changes requested):

markus101 (approved):

PR #6216, opened 2023-11-26
New: Add qBittorrent option for Content Layout

### LINKED ISSUE #6213: Add content layout parameter in qbittorrent settings on torrent addition

Many of the private trackers I use don't use folders for single episode torrents. This results in all the files to be saved in the download folder. And then when sonarr imports those episodes if the folder has subtitles of other show they are imported as well. Qbittorrent has a setting where a folder can be added for a torrent without folders when the torrent is added. This can be done globally or per torrent. I prefer to keep this setting for only torrent sent by sonarr as they are single episode torrent that will be deleted anyway once the show is batched. But for other torrents can follow the original file structure. I need it this way as the private trackers I use also provide subtitles which I auto download in to the download folder automatically. 

The solution I would like is for sonarr to give the option in advance settings for give better control to the user. I asked about it in discord but one of the mods said this setting was not available through the web api but it seems to be availble looking at the latest api document 
https://qbittorrent-api.readthedocs.io/en/latest/apidoc/torrents.html

I have considered enabling set subfolder option globally in qbittorrent but I think that is not the optimal way to do it the optimal way should always be make it available to user through the settings

### Anything else?

_No response_

markus101:
#### Database Migration
NO

#### Description
Content Layout option for qBit used when adding torrents.

* Closes #6213

### DIFF
--- a/src/NzbDrone.Core/Download/Clients/QBittorrent/QBittorrentContentLayout.cs
+++ b/src/NzbDrone.Core/Download/Clients/QBittorrent/QBittorrentContentLayout.cs
@@ -1,0 +1,9 @@
+namespace NzbDrone.Core.Download.Clients.QBittorrent
+{
+    public enum QBittorrentContentLayout
+    {
+        Default = 0,
+        Original = 1,
+        Subfolder = 2
+    }
+}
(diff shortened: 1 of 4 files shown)


Dnkhatri:
what is the reason for this not being merged does it not work?. I closed the issue because I thought this was submitted but it never went into development.

markus101:
The issue will be closed automatically when the PR is merged. This feature was not needed for the v4 release and we didn't want to introduce more features until it was released.

PR #6218, opened 2023-11-28
New: Show Proper or Repack tag in interactive search

stevietv:
#### Database Migration
NO

#### Description
Show a P or R tag for proper or repack in the interactive search results, with possibility to expand to other places quality label is shown.

image

*


stevietv:
@bakerboy448 interesting one, We discussed possible colours in discord and settled on this one. the idea here was to trigger people to think why something is higher in the list even with a lower score.
Wondering what the others think here before I go change it

markus101 (approved):
Regarding the colour I think just having it is good enough for the time being, but we can revisit later.

Issue #1604, opened 2016-12-29
Investigate support for Asian series (E### numbering)


markus101:
We somewhat support this format via RSS by treating them as mini series which works if the episode is in season 1, but will fail otherwise. If there are absolute episode numbers for the series then setting the series as anime may also work for multi-season series, but searching will still fail for search since the format is fairly strict (and would fail immediately if no absolute episode numbers is available).

This may need to rely on the advanced parser or may not be feasible if it can't be done reliably.

Buster14:
AFAIK,  the only reliable tracker for asian tvshows for me is avistaz.to, and the format that always works in avistaz search is 
tvshow_title E##
e.g : 
Running Man E435
Infinity Challenge E421
The Legend of the Blue Sea E13

It works that way in jackett too. 
So they seem to disregard season numbering.
If so perhaps the solution is to have an option to not include season number?

thezoggy:
sounds like absolute numbering..

Buster14:
Absolute numbering sets the season number = 1, while not all asian series have single season.. some famous ongoing series have multiple season due to their long running time..
Running Man
http://thetvdb.com/?tab=series&id=284209
Infinity Challenge
http://thetvdb.com/?tab=seasonall&id=248870
Easiest way in my opinion is to have an option to remove season numbering...

Taloth:
So, which episode is Running Man E435?
http://thetvdb.com/?tab=episode&seriesid=284209&seasonid=592288&id=5267598&lid=7
or
http://thetvdb.com/?tab=episode&seriesid=284209&seasonid=592288&id=4944357&lid=7

Buster14:
Running Man uses aired year as season name now.
Running Man E435, isn't released yet, i'm just using it as an example :)
http://thetvdb.com/?tab=seasonall&id=248870
Running Man E1 would be Running Man S2010E01
But in avistaz.tv tracker it would be accepted as simply Running Man E001
Running Man E235 for example, is Running Man S2015E235 in thetvdb,
but in avistaz it's Running Man E235

Infinity Challenge uses normal season naming, 
http://thetvdb.com/?tab=episode&seriesid=284209&seasonid=592288&id=5267598&lid=7
thetvdb = Infinity Challenge S03e435
Avistaz = Infinity Challenge e435

So the consistent pattern is without season numbering, 
Setting the season value as 1 won't work on avistaz.

Taloth:
[...]

PR #6220, opened 2023-11-29
fix: searching with incorrect absolute episode number in multiple-season anime

DWJZ:
#### Database Migration
NO

#### Description
  When I searched for S02E18 of jujutsu-kaisen , most of the results are S01E18. Sonarr sent queries including "S02E18", "S2 18", "18", but not "42". I believe it is correct to use 42 instead of 18 as absolute episode number here. After switching the `SceneAbsoluteEpisodeNumber` and `AbsoluteEpisodeNumber`, now sonarr can search properly with correct responses.
  Let me know if there is anything needs to be changed.

#### Todos
- [ ] Tests
- [ ] Wiki Updates

may be related to #1604

### DIFF
--- a/src/NzbDrone.Core/IndexerSearch/ReleaseSearchService.cs
+++ b/src/NzbDrone.Core/IndexerSearch/ReleaseSearchService.cs
@@ -350,7 +350,7 @@ namespace NzbDrone.Core.IndexerSearch
 
             searchSpec.SeasonNumber = episode.SceneSeasonNumber ?? episode.SeasonNumber;
             searchSpec.EpisodeNumber = episode.SceneEpisodeNumber ?? episode.EpisodeNumber;
-            searchSpec.AbsoluteEpisodeNumber = episode.SceneAbsoluteEpisodeNumber ?? episode.AbsoluteEpisodeNumber ?? 0;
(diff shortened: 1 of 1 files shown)


markus101:
This will ignore the scene episode number from TheXEM for anything that has an absolute episode number from TheTVDB, which is incorrect. While this fixes this particular issue it'll break countless others.

Looks like this is an issue with the data coming from TheXEM, not a bug.

image

stevietv:
the issue is that subsplease were one of the only groups not to reset their absolute numbering for this show on season 2. all other groups do.
Automatic grabs from RSS for subsplease will work, interactive searches will not.

Please use one of the support channels: [forums](https://forums.sonarr.tv/), [subreddit](https://www.reddit.com/r/sonarr/), [discord ](https://discord.gg/Ex7FmFK), or [IRC](https://web.libera.chat/?channels=#sonarr) for support/questions.

DWJZ:
Thanks for the quick reply! I’m not searching in subsplease but mainly in Mikan. I understand this change will bring break changes. What if I add another query with episode number 42 and keep the 18 query, will this solution be acceptable?

markus101:
I can see some value in it, but also a lot of waste in extra queries that yield no results, but take time to come back + the API hits involved in that. Ultimately it comes down to whether searching 1.5-2x for every series makes sense to solve a few problematic series (the exceptions to the rule) and I'm not sure it does given the already slow querying for anime.

DWJZ:
ok. I will give it a try. Thanks for your patience.

PR #6253, opened 2023-12-15
Fixed: Bump media info revision for DV HDR10Plus

mynameisbogdan:
#### Database Migration
NO

#### Description
Needed to properly detect DV HDR10+ for already imported files.


salesawagner:
following

PR #6269, opened 2023-12-25
New: Use Torznab minimum seed and ratio when available

Faustvii:
Prefer indexer seed configuration if it is available otherwise fallback to Torznab when available

#### Database Migration
NO

#### Description
This change will make Sonarr use `minimumseedtime` and `minimumratio`  from torznab indexers that supply it as long a the indexers haven't been manually configured with a Ratio or Seedtime. This is especially useful for example AnimeBytes where the seed time is based on the size of the torrent and Prowlarr does this calculation but Sonarr isn't using it.

#### Todos
- [x] Tests
- [ ] Wiki Updates

* None to my knowledge, since I didn't make one just looked into implementing it.

### DIFF
--- a/src/NzbDrone.Core.Test/IndexerTests/SeedConfigProviderFixture.cs
+++ b/src/NzbDrone.Core.Test/IndexerTests/SeedConfigProviderFixture.cs
@@ -37,6 +37,7 @@ namespace NzbDrone.Core.Test.IndexerTests
         {
             var settings = new TorznabSettings();
             settings.SeedCriteria.SeasonPackSeedTime = 10;
+            settings.SeedCriteria.SeedRatio = 1.0;
 
             Mocker.GetMock<IIndexerFactory>()
                      .Setup(v => v.Get(It.IsAny<int>()))
@@ -47,10 +48,12 @@ namespace NzbDrone.Core.Test.IndexerTests
 
             var result = Subject.GetSeedConfiguration(new RemoteEpisode
             {
-                Release = new ReleaseInfo()
+                Release = new TorrentInfo()
                 {
                     DownloadProtocol = DownloadProtocol.Torrent,
-                    IndexerId = 1
+                    IndexerId = 1,
+                    MinimumRatio = 1.5,
+                    MinimumSeedTime = 120
                 },
                 ParsedEpisodeInfo = new ParsedEpisodeInfo
                 {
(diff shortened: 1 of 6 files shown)


Faustvii:
> Sadly I disagree with this change.
> 
>     * doesn't allow long term seeding
> 
>     * MST in some indexers where is based on a formula, like in case of AB, isn't something to rely on due to lack of accuracy.

I pretty agree and think you're right, even though you could still do long term by changing the indexer's ratio/seed time. My thoughts were that if Sonarr started to listen to this (Though it should be maybe behind a toggle on indexers) then the indexers in Prowlarr could always be updated if the MST formula wasn't working as expected. But I very much get where you are coming from, just thought it was a shame that we had this data available but wasn't using it for anything, especially when it is being maintained in Prowlarr for instance.

bakerboy448:
Could this just be a UI warning if and maybe show as suggested in the UI? Warn if
- the indexer has configured seed values less than the MST 
OR
- The indexer has no configured seed values AND the client configured values are less than MST

Taloth:
Some scenarios (the below applies to minimumratio as well):
a. User has long seed time manually configured because they want to be sure it seeds for the required time.
b. Indexer provides accurate minimumseedtime and therefore can seed for less than the 'safe default'.
c. Indexer provides accurate minimumseedtime and therefore can force seed longer than the 'default' since the indexer requires it.

Option b is 'nice for the user', option c is 'nice for the indexer'. Imo c should always be forced unless sonarr doesn't control seed time. Indexer that don't seed criteria (public trackers) then minimumseedtime shouldn't be provided simply because '1:1 seed ratio is good'.

The question is whether b is desirable, and I think that comes down to preference.

If someone has no seed time set on indexers, then the client settings win and minimumseedtime should be ignored. Otherwise the configured settings are considered 'fallback' or 'minimum'.

So imo there should be a setting (in the per-indexer seed settings):
- 'Minimum' Use minimumseedtime if it's longer than the configured values. (setting should default to this)
- 'Fallback' Use minimumseedtime if it's available, treating the configured values as default.
- Optional: Do not use minimumseedtime at all. (i'm no fan of this, but it's trivial to add as option since there's already a setting)

The default should make sure it behaves in a 'safe' fashion for everyone. And ppl can elect to use minimumseedtime/ratio to reduce their seed time (due to 0.5 seed ratio criteria specials on their indexer) when they want to.

I don't believe it should be opt-in completely. But by default it shouldn't seed less than the user originally configured. This effectively makes it opt-in since nobody seeds for less than the required values for private indexers.

markus101:
Closing since development on this has stalled, happy to reopen once changes are made.

Issue #6270, opened 2023-12-25
Queue should have the option to show Grab Date


Rubicj:
### Is there an existing issue for this?

- [X] I have searched the existing open and closed issues

### Is your feature request related to a problem? Please describe

When looking at a large queue of downloading items, I would like to be able to see the oldest/newest items in the queue.

### Describe the solution you'd like

Add "Grab Date" as an option in the "Table Columns" options of the Queue:
![image](https://github.com/Sonarr/Sonarr/assets/14086872/86208e4e-2ec6-4943-a87c-382dda5b244a)

### Describe alternatives you've considered

I can use the download client's organization to modify downloads there, but then Sonarr gets confused about downloads dissapearing, and sometimes attempts to regrab the same download later.

### Anything else?

I think the columns in https://github.com/Sonarr/Sonarr/blob/e291834393ce29a1c5c631875e47c62362031e1d/frontend/src/Components/Table/TableOptions/TableOptionsModal.js

are autopopulated from 
https://github.com/Sonarr/Sonarr/blob/e291834393ce29a1c5c631875e47c62362031e1d/src/Sonarr.Api.V3/Queue/QueueResource.cs#L18

which only has access to the fields in the Queue defined in 
https://github.com/Sonarr/Sonarr/blob/e291834393ce29a1c5c631875e47c62362031e1d/src/NzbDrone.Core/Queue/Queue.cs#L13

which doesn't include add date. I _think_ a  NzbDrone.Core.Queue.Queue is a Database object, but I can't find where Queue objects are created using GitHub's online search.

I know that the Grab Date is knowable, because it shows up in individual items History, which is presumeably written at the same time that the Queue object is created. 

I'm not sure how much this project tries to avoid database migrations, but the most direct solution seems to be adding "Grab Date" to the Queue object, and populating that information when the Queue is added to the DB.

Rubicj:
If a database migration is to be avoided, it's probably possible to populate this field from the History table, rather than inserting a new column in the Queue table -- it'll just be a bit messy to pull from multiple tables in the QueueResourceMapper.

markus101:
There is no queue table, queue is built from things Sonarr gets from your download client and anything in pending releases.

`Queue` and `QueueResource` would need a new property, `QueueService` would need to populate it, probably the same way `Indexer` is done from the `TrackedDownload`.

[...]

PR #6271, opened 2023-12-26
Added the option to see the time a Queued item was sent to the download

Rubicj:
#### Database Migration
NO

#### Description
Show the time each of the Queued items was grabbed. Done in the same way as the Indexer, using the last Grab for that downloaded item.

#### Todos
- [x] Tests
- [x] Wiki Updates (N/A)

* https://github.com/Sonarr/Sonarr/issues/6270


src/NzbDrone.Core/Download/Pending/PendingReleaseService.cs
```diff
@@ -368,6 +368,7 @@ private Queue.Queue GetQueueItem(PendingRelease pendingRelease, Lazy<DateTime> n
                 RemoteEpisode = pendingRelease.RemoteEpisode,
                 Timeleft = timeleft,
                 EstimatedCompletionTime = ect,
+                AddedOn = pendingRelease.Added, // Check that this is right
```
Rubicj, line 371: I'm 60% sure this isn't right -- is a pending release one done in the calendar, which shouldn't have an AddedOn time yet?

frontend/src/Activity/Queue/QueueRow.js
```diff
@@ -100,6 +100,7 @@ class QueueRow extends Component {
       outputPath,
       downloadClient,
       estimatedCompletionTime,
+      addedOn,
```
Rubicj, line 103: Should this be named `addedOn` or `added`?

src/NzbDrone.Core/Download/TrackedDownloads/TrackedDownloadService.cs
```diff
@@ -135,6 +135,7 @@ public TrackedDownload TrackDownload(DownloadClientDefinition downloadClient, Do
                     var grabbedEvent = historyItems.FirstOrDefault(v => v.EventType == EpisodeHistoryEventType.Grabbed);
 
                     trackedDownload.Indexer = grabbedEvent?.Data["indexer"];
```
Rubicj, line 137: This line is suspicious to me. We use `?.` to access `Data` on the possibly-null `grabbedEvent`, but then directly index into that data with `["indexer"]` -- if `grabbedEvent` is null, won't we try to do `null["indexer"]` and fail anyway?

Note that the line I added below doesn't use `?.` -- if I did, I'd have to mark `AddedOn` as nullable.

src/NzbDrone.Core/Localization/Core/en.json
```diff
@@ -1573,6 +1573,7 @@
   "Time": "Time",
   "TimeFormat": "Time Format",
   "TimeLeft": "Time Left",
+  "AddedOn": "Added On",
```
Rubicj, line 1576: What should I do for other languages? Plug it though Google Translate and say good enough?

src/NzbDrone.Core/Queue/DatetimeComparer.cs
```diff
@@ -3,7 +3,7 @@
 
 namespace NzbDrone.Core.Queue
 {
-    public class EstimatedCompletionTimeComparer : IComparer<DateTime?>
+    public class DatetimeComparer : IComparer<DateTime?>
```
Rubicj, line 6: I renamed this file -- is this going to cause issues pulling this commit into Radarr/Lidarr/etc?

src/NzbDrone.Core/Localization/Core/en.json
stevietv: Don't do anything with other languages. The Key will be added to weblate and translators can add their translations.
please do sort your en.json though.

Rubicj: Updated using VS's built in "sort keys" option -- there were a lot of changes . . .

stevietv: I think VS sorts badly compared to JetBrains Rider, as in it sorts capitals oddly. As you're only adding one Key it'd be easier just to add it in the right line :)

Rubicj:
Do multiple commits automatically get squashed, or should I squash them at the end?

src/NzbDrone.Core/Localization/Core/en.json
Rubicj: image
That explains it -- yeah,VS was doing case insensitive.

Working on fixing tests too, will push up soon.

stevietv:
> Do multiple commits automatically get squashed, or should I squash them at the end?

you can squash them at the end once the PR is reviewed.

src/NzbDrone.Core/Download/TrackedDownloads/TrackedDownloadService.cs
Rubicj: It seems like I don't correctly understand `?.` --
```
a = grabbedEvent?.Data;
b = a["indexer"]
```
Throws a null pointer exception, but `b = grabbedEvent?.Data["indexer"]` doesn't.
Additionally, the `.Indexer` field of the TrackedDownload isn't nullable, yet null is clearly being written in.

However, whatever's making this line work doesn't apply to `AddedOn` below -- the compiler correctly recognizes that `grabbedEvent?.AddedOn` may be null. Solution is either to mark it as nullable in TrackedDownload, or hide the write inside a `.HasValue` call. Chose to go with the first here.

Rubicj:
Anything else that I should change here? I still have 2 questions -- `added` vs `addedOn`, and whether I should add it in `PendingReleaseService.cs`, and my comment on `?.` probably should get a second set of eyes.

frontend/src/Activity/Queue/QueueRow.js
markus101: I think `added`/`Added` is more consistent with existing fields.

frontend/src/Store/Actions/queueActions.js
```diff
@@ -158,6 +158,12 @@ export const defaultState = {
         isSortable: true,
         isVisible: true
       },
+      {
+        name: 'addedOn',
+        label: () => translate('AddedOn'),
+        isSortable: true,
+        isVisible: true
```
markus101, line 165: Should be `false` so it's not shown by default for existing users.

src/NzbDrone.Core/Download/Pending/PendingReleaseService.cs
markus101: Knowing when it's added to pending would also be useful, so I think this is right, just needs the comment removed.

src/NzbDrone.Core/Download/TrackedDownloads/TrackedDownload.cs
```diff
@@ -15,6 +15,7 @@ public class TrackedDownload
         public TrackedDownloadStatusMessage[] StatusMessages { get; private set; }
         public DownloadProtocol Protocol { get; set; }
         public string Indexer { get; set; }
+        public DateTime? AddedOn { get; set; }
```
markus101, line 18: `Added` to match pending release.

stevietv:
@Rubicj - focus from the Dev team this week has been launching v4 to stable, so theres been a pause on reviewing and merging active PRs. I'm sure Markus will take a look at it after the holidays

frontend/src/typings/Queue.ts
```diff
@@ -28,6 +28,7 @@ interface Queue extends ModelBase {
   sizeleft: number;
   timeleft: string;
   estimatedCompletionTime: string;
+  addedOn: string;
```
markus101, line 31:
```suggestion
  addedOn?: string;
```

src/NzbDrone.Core/Download/TrackedDownloads/TrackedDownloadService.cs
markus101: `?.` only does something when the left side of the expression is not null, it could also be written a longer like this like this: 

```
trackedDownload.AddedOn = grabbedEvent == null ? null : grabbedEvent.Date;
```

I agree that making it nullable is the correct solution since we wouldn't want to show a default date for items that Sonarr didn't grab.

src/NzbDrone.Core/Queue/DatetimeComparer.cs
markus101: It shouldn't be a problem at all.

src/Sonarr.Api.V3/Queue/QueueResource.cs
```diff
@@ -71,6 +72,7 @@ public static QueueResource ToResource(this NzbDrone.Core.Queue.Queue model, boo
                 Sizeleft = model.Sizeleft,
                 Timeleft = model.Timeleft,
                 EstimatedCompletionTime = model.EstimatedCompletionTime,
+                AddedOn = model.AddedOn,
```
markus101, line 75:
```suggestion
                Added = model.Added,
```

src/Sonarr.Api.V3/Queue/QueueResource.cs
```diff
@@ -29,6 +29,7 @@ public class QueueResource : RestResource
         public decimal Sizeleft { get; set; }
         public TimeSpan? Timeleft { get; set; }
         public DateTime? EstimatedCompletionTime { get; set; }
+        public DateTime? AddedOn { get; set; }
```
markus101, line 32:
```suggestion
        public DateTime? Added { get; set; }
```

markus101 (changes requested):
Looks good, mostly just `AddedOn` to `Added`

src/NzbDrone.Core/Localization/Core/en.json
Rubicj: Note that I left this as "AddedOn" rather than "Added" -- that seems right for the human-consumption string.

Rubicj:
Done, no rush though. This is my first time contributing to any open-source project, I want to get it right.

Rubicj:
Anything else I should do here? The PR still shows Changes Requested, but all changes have been implemented.

frontend/src/Store/Actions/queueActions.js
```diff
@@ -158,6 +158,12 @@ export const defaultState = {
         isSortable: true,
         isVisible: true
       },
+      {
+        name: 'added',
+        label: () => translate('AddedOn'),
```
markus101, line 163: I think this should use the already existing `Added` translation token, which will drop the `on`.

src/NzbDrone.Core/Localization/Core/en.json
```diff
@@ -53,6 +53,7 @@
   "AddToDownloadQueue": "Add to download queue",
   "Added": "Added",
   "AddedDate": "Added: {date}",
+  "AddedOn": "Added On",
```
markus101, line 56: Can remove when changing the token.

markus101 (changes requested):

markus101:
Thanks for the fixes, just one more minor thing (and another related to it).

markus101 (approved):

markus101 (changes requested):
Everything looks good, but this will need to be rebased on `develop` to pickup the latest action changes so the build will pass.

markus101 (approved):

markus101:
Thanks for adding this!

PR #6273, opened 2023-12-28
Fixed: Monitor last season only if there's at least one future monitored episode

mynameisbogdan:
#### Database Migration
NO

#### Description
Fix for last season monitored with no future episodes when applying monitor future episodes.


markus101 (changes requested):

markus101 (changes requested):
Just need to update the comment:

```
// - Set to monitor all episodes 
// - Set to monitor future episodes and series is continuing or not yet aired
```

markus101 (approved):

PR #6276, opened 2023-12-29
add install.sh script and remove old debian install methods

stevietv:
#### Database Migration
NO

#### Description
adjust the install method for v4 debian based installs


distribution/debian/sonarr.service
```diff
@@ -11,7 +11,7 @@ Group=sonarr
 UMask=002
 
 Type=simple
-ExecStart=/usr/lib/sonarr/bin/Sonarr -nobrowser -data=/var/lib/sonarr
+ExecStart=/opt/sonarr/Sonarr -nobrowser -data=/var/lib/sonarr
```
markus101, line 14: I think this is actually `/opt/Sonarr`, but I think this file can be nuked since it's built into the installation script

markus101 (changes requested):

distribution/debian/sonarr.service
stevietv: I kept this file as its linked to from the wiki and the announcements. an example of how people might want to change their service file too if they run v3 already and dont want to use the install script

PR #6282, opened 2023-12-31
New: Custom formats in episode history

mynameisbogdan:
#### Database Migration
NO

#### Description
image

### DIFF
--- a/frontend/src/Episode/History/EpisodeHistoryRow.js
+++ b/frontend/src/Episode/History/EpisodeHistoryRow.js
@@ -9,7 +9,6 @@ import RelativeDateCellConnector from 'Components/Table/Cells/RelativeDateCellCo
 import TableRowCell from 'Components/Table/Cells/TableRowCell';
 import TableRow from 'Components/Table/TableRow';
 import Popover from 'Components/Tooltip/Popover';
-import Tooltip from 'Components/Tooltip/Tooltip';
 import EpisodeFormats from 'Episode/EpisodeFormats';
 import EpisodeLanguages from 'Episode/EpisodeLanguages';
 import EpisodeQuality from 'Episode/EpisodeQuality';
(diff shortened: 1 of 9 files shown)


stevietv:
this will certainly help when providing support via screenshots to identify CF matches

frontend/src/Episode/History/EpisodeHistoryRow.js
```diff
@@ -102,14 +101,28 @@ class EpisodeHistoryRow extends Component {
...
         <RelativeDateCellConnector
           date={date}
         />
 
-        <TableRowCell className={styles.details}>
+        <TableRowCell className={styles.actions}>
           <Popover
             anchor={
-              <Icon
+              <IconButton
```
markus101, line 125: Why is it changing to a button?

frontend/src/Episode/History/EpisodeHistoryRow.js
```diff
@@ -102,14 +101,28 @@ class EpisodeHistoryRow extends Component {
           />
         </TableRowCell>
 
+        <TableRowCell>
+          <EpisodeFormats formats={customFormats} />
+        </TableRowCell>
+
+        <TableRowCell>
+          <Tooltip
```
markus101, line 109: Doesn't seem super useful to have a tooltip with the custom formats when they're always shown in the column to the left.

frontend/src/Episode/History/EpisodeHistoryRow.js
markus101: I understand that grouping these saves horizontal space, but information under `Actions` doesn't seem right and we should consider removing the column title.

frontend/src/Episode/History/EpisodeHistoryRow.js
stevietv: would make sense to me to remove the tooltip indeed!

markus101 (approved):

PR #6295, opened 2024-01-03
Correctly parse German DL and ML tags in releases

JBou:
#### Database Migration
NO

#### Description
This is an upstream port of https://github.com/Radarr/Radarr/pull/9512

Correctly parses German DL and ML tags

- The DL tag (Dual Language) is quite like the `Multi` tag, but for the German Scene, it specifies that the release contains the German and the original language.
- The ML tag (Multi Language) contains the following languages: `German`, `Original`, `English` or more ([examples on xrel.to](https://www.xrel.to/p2p/releases.html?search_results=v6anRHibr9167592559&query=german+ml+s01&maxed_out=1))

#### Todos
- [x] Tests
- [ ] Wiki Updates

*

### DIFF
--- a/src/NzbDrone.Core.Test/ParserTests/LanguageParserFixture.cs
+++ b/src/NzbDrone.Core.Test/ParserTests/LanguageParserFixture.cs
@@ -385,5 +385,37 @@ namespace NzbDrone.Core.Test.ParserTests
             var result = LanguageParser.ParseLanguages(postTitle);
             result.Should().BeEquivalentTo(new[] { Language.English, Language.Spanish, Language.Catalan });
         }
+
+        [TestCase("Series.Title.S01E01.German.DL.1080p.BluRay.x264-RlsGrp")]
+        [TestCase("Series.Title.S01E01.GERMAN.DL.1080P.WEB.H264-RlsGrp")]
+        [TestCase("Series.Title.2023.S01E01.German.DL.EAC3.1080p.DSNP.WEB.H264-RlsGrp")]
+        public void should_add_original_language_to_german_release_with_dl_tag(string postTitle)
+        {
+            var result = Parser.Parser.ParseTitle(postTitle);
+            result.Languages.Count.Should().Be(2);
+            result.Languages.Should().Contain(Language.German);
+            result.Languages.Should().Contain(Language.Original);
+        }
+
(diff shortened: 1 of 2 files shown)


stevietv:
Isn't this currently going to parse ALL german WEB-DL files as dual lanugage. Web-DL is a quality source, not a definition of the language. Surely not all German WEBDL releases are dual language?

markus101:
Unless the release name has something other than a quality tag I don't think treating all Web DL as being dual language is viable it's too generic to be something we'll use for that purpose.

JBou:
> Isn't this currently going to parse ALL german WEB-DL files as dual lanugage. Web-DL is a quality source, not a definition of the language. Surely not all German WEBDL releases are dual language?

No, this PR is **NOT** parsing all German WEB-DL files as dual language. That is not the intention. I already added a test specifically for this case that excludes the following strings from being parsed as dual language: `WEBDL`, `WEB.DL`, `WEB DL`, `WEB-DL`, `WEB_DL`.
https://github.com/Sonarr/Sonarr/pull/6295/files#diff-6e24515432aa1dd3b4667bb1012400fb5103f94b152a4578b8190dd2bb3accd3R400-R409
There is no intention to parse `WEB-DL` as dual language, this would be wrong, as you have written @markus101.
This PR should only parse a standalone `.DL.` tag as dual language, as you can see in the tests. `WEB-DL` is a quality tag with the same meaining as in the English scene, but `DL` is an **additional** separate tag in the German scene. The regex only contains `WEB-DL` to **exclude it** (to avoid false positives). Maybe there was a misunderstanding?
As you can see [at these releases](https://www.xrel.to/p2p/releases.html?search_results=a838WgdaLU167594989&page=7) or also [here](https://www.xrel.to/p2p/327797-Fargo-S05E07-Linda-GERMAN-5-1-DUBBED-DL-AC3-2160p-WEB-DL-SDR-HEVC-TvR/nfo.html) (`Fargo.S05E07.Linda.GERMAN.5.1.DUBBED.DL.AC3.2160p.WEB-DL.SDR.HEVC-TvR`), it can be a `WEB-DL` that additionally has dual language (`DL` tag).

stevietv:
I see, I'm sorry I missed the negative look behind in the Regex! Thanks for
taking the time to explain it to me!

markus101 (approved):

PR #6307, opened 2024-01-05
Exit Debian Install Script when running from install dir

stevietv:
Exit the install script with a warning when trying to run it from the planned install directory.


distribution/debian/install.sh
bakerboy448, line 1: On 126 the comment is now moot

stevietv: thanks, made a tweak also as it shouldn't be ran from installdir or bindir

markus101 (approved):

PR #6315, opened 2024-01-06
Fixed: Loading series details when Quality Profile is invalid

mynameisbogdan:
#### Database Migration
NO

#### Description
Don't break the whole page when series has an invalid quality profile.

ss


markus101:
The backend prevents deleting a quality profile if it's used by a series or an import list, so I'm curious how someone would even get into this scenario without editing the DB directly?

If they edit the DB manually and break things this much, they should get an error page instead of just a subtle indication.

markus101:
Still worth having something like this to handle that, but probably more in your face would be better.

PR #6321, opened 2024-01-08
Fixed: Latvian and Russian language parsing

AndrejsK:
#### Database Migration
NO

#### Description
Added proper support for Latvian with test cases I have encountered in the wild and fixed a case where Russian is not recognized (RU instead of RUS).

#### Todos
- [x] Tests
- N/A Wiki Updates


markus101 (approved):
Thanks!

PR #6324, opened 2024-01-09
Fixed: Remove dead Torznab presets

ilike2burnthing:
#### Database Migration
NO

#### Description
Remove dead Torznab presets - HD4Free.xyz and Nyaa Pantsu.

Again, tidy/change whatever.

#### Todos
- [x] Tests
- [ ] Wiki Updates

*

### DIFF
--- a/src/NzbDrone.Core/Indexers/Torznab/Torznab.cs
+++ b/src/NzbDrone.Core/Indexers/Torznab/Torznab.cs
@@ -49,9 +49,7 @@ namespace NzbDrone.Core.Indexers.Torznab
         {
             get
             {
-                yield return GetDefinition("HD4Free.xyz", GetSettings("http://hd4free.xyz"));
                 yield return GetDefinition("AnimeTosho Torrents", GetSettings("https://feed.animetosho.org", apiPath: @"/nabapi", categories: Array.Empty<int>(), animeCategories: new[] { 5070 }));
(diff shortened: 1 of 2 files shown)


stevietv:
Thanks for this. This one fails to build right now until the `using System.Linq;` is removed from `TorznabSettings.cs` - its no longer used anymore.

stevietv (changes requested):
update using directives
Thanks for this. This one fails to build right now until the `using System.Linq;` is removed from `TorznabSettings.cs` - its no longer used anymore.

stevietv (approved):
I was just about to comment my review when I saw a new commit:

> The intention of validation of an having an API was because we knew that these specific indexer templates required an API key. By removing the && check against the whitelist we are now requiring every Torznab implemented indexer to have an API key, which may not be correct.
> I see two options here:
> 
> 1. Keep the ApiKeyWhiteList string array, only empty, and keep the validation against it, just incase a new template torznab is added in the future where we know an ApiKey is required. This may be an inefficient check, though minimal.
> 2. remove the validation rule ` RuleFor(c => c.ApiKey).NotEmpty().When(ShouldHaveApiKey);` and the function `ShouldHaveApiKey`

Seems you also caught that and went for option 1 :D

ilike2burnthing:
I swear that check was fine a minute ago :/

`using System;` missing?

stevietv:
> `using System;` missing?

correct, your using now Array.Empty :)

Build checks take a little while to run.

markus101 (approved):

PR #6335, opened 2024-01-11
Sort Custom Filters

### LINKED ISSUE #6334: Ability to sort/rearrange Custom Filters

When adding a custom filter, the new filter is always placed at the bottom of the list of available Custom Filters.

Adding the ability to rearrage/sort the Custom Filters list so that similar filters can be grouped together and so newly added filters don't appear at bottom of list by default.

If changing the order of the Custom Filters isn't feasible, sorting them alphabetically by default could be helpful as the user could group like filters by name.

### Anything else?

_No response_

stevietv:
#### Database Migration
NO

#### Description
Sorts custom filters alphabetically

#### Todos
- [ ] Tests
- [ ] Wiki Updates

* closes #6334

### DIFF
--- a/frontend/src/Components/Menu/FilterMenuContent.js
+++ b/frontend/src/Components/Menu/FilterMenuContent.js
@@ -40,18 +40,25 @@ class FilterMenuContent extends Component {
         }
 
         {
-          customFilters.map((filter) => {
-            return (
-              <FilterMenuItem
-                key={filter.id}
-                filterKey={filter.id}
-                selectedFilterKey={selectedFilterKey}
-                onPress={onFilterSelect}
-              >
-                {filter.label}
-              </FilterMenuItem>
-            );
-          })
+          (customFilters.length > 0) &&
+            <MenuItemSeparator />
+        }
+
+        {
+          customFilters
+            .sort((a, b) => a.label.localeCompare(b.label))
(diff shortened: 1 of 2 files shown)


markus101 (changes requested):
We also need to sort them in filter menu

stevietv:
> We also need to sort them in filter menu

done! I missed that as every time I checked the list they were already sorted due to the state being sorted (coming from the modal)

I also added a separator here to split standard from custom filters. Helps a little to distinguish why there are two lists.

frontend/src/Components/Menu/FilterMenuContent.js
```diff
@@ -40,18 +40,25 @@ class FilterMenuContent extends Component {
...
-                filterKey={filter.id}
-                selectedFilterKey={selectedFilterKey}
-                onPress={onFilterSelect}
-              >
-                {filter.label}
-              </FilterMenuItem>
-            );
-          })
+          (customFilters.length > 0) &&
+            <MenuItemSeparator />
```
markus101, line 44:
```suggestion
          customFilters.length > 0 ?
            <MenuItemSeparator /> :
            null
```

markus101 (approved):
Just the suggestion to make it a ternary, but will commit that change and land once CI is happy.

PR #6336, opened 2024-01-11
Avoid blocklisted by title query for releases with InfoHash

mynameisbogdan:
#### Database Migration
NO

#### Description
This change simply avoids running `BlocklistedByTitle` (which is a `LIKE` query with wildcards) if the release has an infohash and the result isn't used anyways.

### DIFF
--- a/src/NzbDrone.Core/Blocklisting/BlocklistService.cs
+++ b/src/NzbDrone.Core/Blocklisting/BlocklistService.cs
@@ -36,30 +36,28 @@ namespace NzbDrone.Core.Blocklisting
 
         public bool Blocklisted(int seriesId, ReleaseInfo release)
         {
-            var blocklistedByTitle = _blocklistRepository.BlocklistedByTitle(seriesId, release.Title);
-
             if (release.DownloadProtocol == DownloadProtocol.Torrent)
             {
-                var torrentInfo = release as TorrentInfo;
-
-                if (torrentInfo == null)
+                if (release is not TorrentInfo torrentInfo)
                 {
                     return false;
                 }
 
-                if (torrentInfo.InfoHash.IsNullOrWhiteSpace())
+                if (torrentInfo.InfoHash.IsNotNullOrWhiteSpace())
                 {
-                    return blocklistedByTitle.Where(b => b.Protocol == DownloadProtocol.Torrent)
-                                             .Any(b => SameTorrent(b, torrentInfo));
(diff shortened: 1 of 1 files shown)


src/NzbDrone.Core/Blocklisting/BlocklistService.cs
```diff
@@ -36,30 +36,34 @@ public BlocklistService(IBlocklistRepository blocklistRepository)
...
 
-                var blocklistedByTorrentInfohash = _blocklistRepository.BlocklistedByTorrentInfoHash(seriesId, torrentInfo.InfoHash);
+                var blocklistedByTitle = _blocklistRepository.BlocklistedByTitle(seriesId, release.Title);
 
-                return blocklistedByTorrentInfohash.Any(b => SameTorrent(b, torrentInfo));
+                return blocklistedByTitle
+                    .Where(b => b.Protocol == DownloadProtocol.Torrent)
+                    .Any(b => SameTorrent(b, torrentInfo));
             }
+            else
```
markus101, line 59: The `else` is unnecessary since all the paths in the `if` return.

markus101: No real need for the interim variable since it's only used the one time.

```
_blocklistRepository.BlocklistedByTitle(seriesId, release.Title).blocklistedByTitle
                    .Where(b => b.Protocol == DownloadProtocol.Torrent)
                    .Any(b => SameTorrent(b, torrentInfo));
```

src/NzbDrone.Core/Blocklisting/BlocklistService.cs
```diff
@@ -36,30 +36,34 @@ public BlocklistService(IBlocklistRepository blocklistRepository)
...
                 }
 
-                if (torrentInfo.InfoHash.IsNullOrWhiteSpace())
+                if (torrentInfo.InfoHash.IsNotNullOrWhiteSpace())
                 {
-                    return blocklistedByTitle.Where(b => b.Protocol == DownloadProtocol.Torrent)
-                                             .Any(b => SameTorrent(b, torrentInfo));
+                    var blocklistedByTorrentInfohash = _blocklistRepository.BlocklistedByTorrentInfoHash(seriesId, torrentInfo.InfoHash);
+
+                    return blocklistedByTorrentInfohash.Any(b => SameTorrent(b, torrentInfo));
```
markus101, line 50: Same thing as usenet here

markus101 (approved):

Issue #6356, opened 2024-01-14
Clarify download client priority


nc7s:
### Is there an existing issue for this?

- [X] I have searched the existing open and closed issues

### Is your feature request related to a problem? Please describe

The "client priority" setting in a download client's setting dialog does not clarify which direction the priority goes. As a comparison, indexers have this clarification.

### Describe the solution you'd like

Add description in `PriorityHelpText`, i18n. I can send in a PR, but needs confirmation of the direction (i.e. is 1 highest or lowest?).

### Describe alternatives you've considered

No?

markus101:
It's the same as indexers, 1 is the highest priority.

https://github.com/Sonarr/Sonarr/blob/7e011df2b2c66ede0a0fbb3507a6ab89a7a08673/src/NzbDrone.Core/Download/DownloadClientProvider.cs#L93-L96

stevietv:
this'd be a fine improvement.
It'd be best to rename the translation key to `DownloadClientPriorityHelpText` in [`frontend/src/Settings/DownloadClients/DownloadClients/EditDownloadClientModalContent.js`](https://github.com/Sonarr/Sonarr/blob/7e011df2b2c66ede0a0fbb3507a6ab89a7a08673/frontend/src/Settings/DownloadClients/DownloadClients/EditDownloadClientModalContent.js#L136) and the translation in `en.json` updated to be more inline with the IndexerPriorityHelpText

1 is highest priority. and 1 is the default priority

nc7s:
I see that translations are synced from Weblate. Should I update the key only and submit translation updates there? Or is it OK to go for a direct change in code?

markus101:
Only `en.json` should be updated, everything else comes from webplate.

PR #6357, opened 2024-01-14
Narrow download client PriorityHelpText key name and clarify direction

nc7s:
#### Description
Clarify priority direction and change the key name as @stevietv suggested in #6356. `fi`, `fr`, `pr_BR` and `zh_CN` too have this key, so I also updated them.

#### Database Migration
NO

* Closes https://github.com/Sonarr/Sonarr/issues/6356


stevietv (changes requested):
As the content of the translation key changed. I would say its better not edit the other languages, let weblate drop the current translation and a new one be created by translators.

nc7s:
@stevietv: updated to only change `en.json`.

markus101:
Thanks for fixing this!

PR #6388, opened 2024-01-20
Fixed: Avoid upgrades for custom formats cut-off already met

mynameisbogdan:
#### Description
A check if existing quality meets the custom formats cutoff, since it's possible that an intermediary quality to bypass the custom formats cutoff. 

Seems reasonable?


markus101 (approved):
I don't see anything wrong with this approach, any reason to delay before landing (since it's a draft)?

markus101:
Yeah, that code doesn't particularly care whether what's an upgrade just that it's an upgrade and the upgrade is allowed on the profile.

PR #6394, opened 2024-01-21
Display database migration version in Status

mynameisbogdan:
#### Description
Would be easier to know which is the current DB migration version.

`DatabaseMigration` vs `DbMigration`?


src/NzbDrone.Core/Localization/Core/en.json
```diff
@@ -285,6 +285,7 @@
   "DailyEpisodeTypeFormat": "Date ({format})",
   "Dash": "Dash",
   "Database": "Database",
+  "DatabaseMigration": "Database migration",
```
markus101, line 288: Since we use this as a title of sorts I think `Database Migration` makes more sense. `AppData Directory` and `Startup Directory` should probably also be updated.

PR #6397, opened 2024-01-21
New: Expand seasons with all episodes having missing air dates

mynameisbogdan:
#### Description
Expand by default seasons where the episodes, usually TBA, have the air date missing. 

Should make it easier to not get scared that you missed a whole season, when it's actually TBA. 😄 

Screenshot


markus101 (commented):
Does this need a separate condition to make sure we don't expand specials or is that already handled?

PR #6398, opened 2024-01-21
fix translation in CF Spec

stevietv:
#### Description
fix translation in CF Spec


markus101 (approved):

stevietv:
> I think we should also use the same translation key in the UI notice check, otherwise it might not show if the translations differ.
> 
> https://github.com/Sonarr/Sonarr/blob/0128dca5ac1e26e6d04b26b3f492aa1db8530c96/frontend/src/Settings/CustomFormats/CustomFormats/Specifications/EditSpecificationModalContent.js#L52

Oh absolutely!

PR #6444, opened 2024-01-27
Wrapped fields on series details page in div

jab416171:
#### Description
This allows you to triple click on the series details page to select the path for instance, similar to the details page in radarr.

#### Database Migration
NO

### DIFF
--- a/frontend/src/Series/Details/SeriesDetails.js
+++ b/frontend/src/Series/Details/SeriesDetails.js
@@ -428,14 +428,16 @@ class SeriesDetails extends Component {
                     className={styles.detailsLabel}
                     size={sizes.LARGE}
                   >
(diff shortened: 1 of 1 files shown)


frontend/src/Series/Details/SeriesDetails.js
```diff
@@ -444,16 +446,18 @@ class SeriesDetails extends Component {
...
-                          name={icons.DRIVE}
-                          size={17}
-                        />
 
-                        <span className={styles.sizeOnDisk}>
-                          {
-                            formatBytes(sizeOnDisk || 0)
-                          }
-                        </span>
+                        <div>
```
markus101, line 450: Do these all need to be wrapped now?

PR #6446, opened 2024-01-27
fix/correct translation keys

stevietv:
#### Description
fix missing translation key in TransmissionSettings.cs
and fix one in Delay Profiles Modal


stevietv:
> I would also remove this empty line too:
> 
> https://github.com/Sonarr/Sonarr/blob/70807a9dcf0fbbd39768779ec8b416c78e6804f8/src/NzbDrone.Core/Download/Clients/Transmission/TransmissionSettings.cs#L51-L53

gone, as if it never existed

PR #6450, opened 2024-01-29
Add additional On Manual Interaction Required notifications

### LINKED ISSUE #6448: "On Manual Interaction Required" notification not working

### Current Behavior

I have set up a connection to send a notification when a manual interaction is required. My understanding is that a waiting import is such an event, but sonarr does not send a notification.

### Expected Behavior

Sonarr should issue a notification when an import is waiting to be imported and a connection is set up with "On Manual Interaction Required".

### Steps To Reproduce

Sonarr has a download that it can't import automatically.

In my case, the name of the downloaded file is different from the grabbed title and the import is no longer a custom format upgrade.

### Environment

```markdown
- OS: Debian 6.9.12-2
- Sonarr: 4.0.1.929-ls222 by linuxserver.io
- Docker Install: Yes
- Using Reverse Proxy: No
- Database: Sqlite 3.44.2
```

Main

### Trace Logs?

https://privatebin.net/?0ce5c7dfb3962134#Fzmmcp8tKNEhWPKvssrPQWncq4dG7oEnvmSwXNRqsgSo

### Anything else?

Let me know if the event is missing in the logs. After enabling trace logs, I changed the category from the download client and back to trigger the import again.

[sonarr.trace.txt](https://github.com/Sonarr/Sonarr/files/14077362/sonarr.trace.txt)

connect
Screenshot_20240128-132444_ntfy
act2
act1

stevietv:
#### Description
Trigger notification on other manual interaction types.

* Closes #6448

### DIFF
--- a/src/NzbDrone.Core/Download/CompletedDownloadService.cs
+++ b/src/NzbDrone.Core/Download/CompletedDownloadService.cs
@@ -73,6 +73,11 @@ namespace NzbDrone.Core.Download
             var grabbedHistories = _historyService.FindByDownloadId(trackedDownload.DownloadItem.DownloadId).Where(h => h.EventType == EpisodeHistoryEventType.Grabbed).ToList();
             var historyItem = grabbedHistories.MaxBy(h => h.Date);
 
+            if (grabbedHistories.Count == 0)
+            {
+                grabbedHistories = new List<EpisodeHistory> { new EpisodeHistory() };
+            }
+
             if (historyItem == null && trackedDownload.DownloadItem.Category.IsNullOrWhiteSpace())
             {
                 trackedDownload.Warn("Download wasn't grabbed by Sonarr and not in a category, Skipping.");
--- a/src/NzbDrone.Core/Download/CompletedDownloadService.cs
+++ b/src/NzbDrone.Core/Download/CompletedDownloadService.cs
@@ -260,14 +260,9 @@ namespace NzbDrone.Core.Download
             {
                 var grabbedHistories = _historyService.FindByDownloadId(trackedDownload.DownloadItem.DownloadId).Where(h => h.EventType == EpisodeHistoryEventType.Grabbed).ToList();
 
-                if (grabbedHistories.Count == 0)
-                {
-                    grabbedHistories = new List<EpisodeHistory> { new EpisodeHistory() };
-                }
-
                 trackedDownload.HasNotifiedManualInteractionRequired = true;
 
-                var releaseInfo = new GrabbedReleaseInfo(grabbedHistories);
+                var releaseInfo = grabbedHistories.Count > 0 ? new GrabbedReleaseInfo(grabbedHistories) : null;
                 var manualInteractionEvent = new ManualInteractionRequiredEvent(trackedDownload, releaseInfo);
 
                 _eventAggregator.PublishEvent(manualInteractionEvent);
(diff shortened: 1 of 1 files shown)


src/NzbDrone.Core/Download/CompletedDownloadService.cs
```diff
@@ -96,6 +96,17 @@ public void Check(TrackedDownload trackedDownload)
                 if (series == null)
                 {
                     trackedDownload.Warn("Series title mismatch; automatic import is not possible. Check the download troubleshooting entry on the wiki for common causes.");
+
+                    if (!trackedDownload.HasNotifiedManualInteractionRequired)
```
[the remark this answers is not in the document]
stevietv: you mean in this case don't notify if it was an interactive search, with the expectation that user is already present and ready to import?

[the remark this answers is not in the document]
stevietv: I'm not sure I agree here. I this case it would stop all warnings when the title was a mismatch. I didn't change the logic here of the warning creation, only added the notifcation

src/NzbDrone.Core/Download/CompletedDownloadService.cs
```diff
@@ -133,9 +149,27 @@ public void Import(TrackedDownload trackedDownload)
...
+            if (grabbedHistories.Count == 0)
+            {
+                grabbedHistories = new List<EpisodeHistory> { new EpisodeHistory() };
+            }
+
             if (trackedDownload.RemoteEpisode == null)
             {
                 trackedDownload.Warn("Unable to parse download, automatic import is not possible.");
+
+                if (!trackedDownload.HasNotifiedManualInteractionRequired)
```
markus101, line 163: I think this can all be pulled out to a separate method that can get the history (if not provided), check whether it's been notified before and send the notification. This way we don't need to look up the history if it's never going to be used and don't have to duplicate this same block of code whether we are checking whether to send a notification.

markus101 (changes requested):

src/NzbDrone.Core/Download/CompletedDownloadService.cs
stevietv: split out the notification decision into a separate method

src/NzbDrone.Core/Download/CompletedDownloadService.cs
```diff
@@ -258,6 +254,26 @@ public bool VerifyImport(TrackedDownload trackedDownload, List<ImportResult> imp
...
+                var grabbedHistories = _historyService.FindByDownloadId(trackedDownload.DownloadItem.DownloadId).Where(h => h.EventType == EpisodeHistoryEventType.Grabbed).ToList();
+
+                if (grabbedHistories.Count == 0)
+                {
+                    grabbedHistories = new List<EpisodeHistory> { new EpisodeHistory() };
+                }
+
+                trackedDownload.HasNotifiedManualInteractionRequired = true;
+
+                var releaseInfo = new GrabbedReleaseInfo(grabbedHistories);
```
markus101, line 270: Since I didn't like `grabbedHistories = new List<EpisodeHistory> { new EpisodeHistory() };` I did some digging and everything that handles `releaseInfo` later will ignore it when it's null so we can simply this to something like:

```suggestion
                trackedDownload.HasNotifiedManualInteractionRequired = true;

                var releaseInfo = grabbedHistories.Count > 0 ? new GrabbedReleaseInfo(grabbedHistories) : null;
```

markus101 (approved):

PR #6473, opened 2024-02-04
Improve add/loading error notices

mynameisbogdan:
#### Description
- Reverting b76bf373717edff8e475fde31fbaec86c65903fe to use `name` for consistency across the board.
- Changed translation token for loading error in ImportListOptions for consistency
- Replaced lodash's find with native in a few places
- Improve add/loading errors notices by showing them as alerts
- Made Manage Import Lists modal wider since paths can be long
- Use the sorted selector in ImportListsConnector
- Removed some `c={true}`
- Fixed 2 translation tokens not related to Jackett.
- Fix translation token for DL client directory help text
- Fix typo for Downloaded Episodes Scan command name


frontend/src/Store/Actions/episodeActions.js
```diff
@@ -106,8 +106,7 @@ export const defaultState = {
     {
       name: 'size',
       label: () => translate('Size'),
-      isVisible: false,
-      isSortable: true
```
markus101, line 110: It should be sortable, the fact that it's not working correctly is the bug.

markus101 (changes requested):

frontend/src/Store/Actions/episodeActions.js
markus101: Neither are a number of these changes...

markus101 (approved):

PR #6505, opened 2024-02-12
Fixed: Correctly persist calendar custom filter selection

Qstick:
#### Description
Correctly persist calendar custom filter selection


frontend/src/Store/Actions/calendarActions.js
```diff
@@ -97,7 +97,7 @@ export const persistState = [
   'calendar.view',
   'calendar.selectedFilterKey',
   'calendar.options',
-  'seriesIndex.customFilters'
+  'calendar.customFilters'
```
markus101, line 100: 🙈

markus101 (approved):
Thanks for fixing the bad copy/pasta

PR #6522, opened 2024-02-18
Search for new series rather than going to first suggestion

robjackstewart:
#### Description
Most search engines do not select the first suggestion upon pressing enter, they perform a search for your term. Sonarr will currently do this when there are no existing series that match the search term, but it will go to the first matching existing series if one exists.

Update the Sonarr search to look for a new series upon pressing Tab or Enter instead of selecting the first existing series if there is one. The behaviour for no matching existing series remains the same. It will navigate to the highlighted selection if selected.

Behavioural change rather than visual change.

#### Database Migration
NO

N/A

#### Other info
[Corresponding PR for Radarr](https://github.com/Radarr/Radarr/pull/9786)

### DIFF
--- a/frontend/src/Components/Page/Header/SeriesSearchInput.js
+++ b/frontend/src/Components/Page/Header/SeriesSearchInput.js
@@ -161,13 +161,12 @@ class SeriesSearchInput extends Component {
       return;
     }
 
-    // If an suggestion is not selected go to the first series,
-    // otherwise go to the selected series.
-
-    if (highlightedSuggestionIndex == null) {
-      this.goToSeries(suggestions[0]);
-    } else {
+    // If a suggestion is highlighted, go to that movie
+    // otherwise search for a new movie
+    if (highlightedSuggestionIndex) {
       this.goToSeries(suggestions[highlightedSuggestionIndex]);
+    } else {
+      this.props.onGoToAddNewSeries(value);
     }
 
     this._autosuggest.input.blur();
--- a/frontend/src/Components/Page/Header/SeriesSearchInput.js
+++ b/frontend/src/Components/Page/Header/SeriesSearchInput.js
@@ -161,12 +161,12 @@ class SeriesSearchInput extends Component {
       return;
     }
 
-    // If a suggestion is highlighted, go to that series
-    // otherwise search for a new series
-    if (highlightedSuggestionIndex) {
-      this.goToSeries(suggestions[highlightedSuggestionIndex]);
-    } else {
+    // If no suggestion is highlighted, search for a new series
+    // otherwise go to that suggestion
+    if (highlightedSuggestionIndex == null) {
       this.props.onGoToAddNewSeries(value);
+    } else {
+      this.goToSeries(suggestions[highlightedSuggestionIndex]);
     }
 
     this._autosuggest.input.blur();
--- a/frontend/src/Components/Page/Header/SeriesSearchInput.js
+++ b/frontend/src/Components/Page/Header/SeriesSearchInput.js
@@ -161,8 +161,8 @@ class SeriesSearchInput extends Component {
       return;
     }
 
-    // If a suggestion is highlighted, go to that movie
-    // otherwise search for a new movie
+    // If a suggestion is highlighted, go to that series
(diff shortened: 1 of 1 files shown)


PearsonFlyer:
Your comments in the file reference "movie" twice. This is Sonarr.

robjackstewart:
> Your comments in the file reference "movie" twice. This is Sonarr.

Sorry about that! Fixed.

frontend/src/Components/Page/Header/SeriesSearchInput.js
```diff
@@ -161,13 +161,12 @@ class SeriesSearchInput extends Component {
...
 
-    // If an suggestion is not selected go to the first series,
-    // otherwise go to the selected series.
-
-    if (highlightedSuggestionIndex == null) {
-      this.goToSeries(suggestions[0]);
-    } else {
+    // If a suggestion is highlighted, go to that series
+    // otherwise search for a new series
+    if (highlightedSuggestionIndex) {
```
markus101, line 166: `null`, `undefined` and `0` will all fall through this check, which means the first item will never be selectable.

Something like this would be more appropriate.

```
    if (highlightedSuggestionIndex == null) {
      this.props.onGoToAddNewSeries(value);
    } else {
      this.goToSeries(suggestions[highlightedSuggestionIndex]);
    }
```

markus101 (changes requested):

frontend/src/Components/Page/Header/SeriesSearchInput.js
robjackstewart: Fixed 😄

markus101:
I've discussed this a bit on Discord with @stevietv and @mynameisbogdan and the consensus is changing the default behaviour that has existed for years doesn't make sense. It will lead to extra searches against Skyhook and disrupt the known experience.

One suggestion was to make the search option always available and extend the height a bit to see more results, which would make it easier to see that option without actually triggering an unexpected search.

robjackstewart-microsoft:
> I've discussed this a bit on Discord with @stevietv and @mynameisbogdan and the consensus is changing the default behaviour that has existed for years doesn't make sense. It will lead to extra searches against Skyhook and disrupt the known experience.
> 
> One suggestion was to make the search option always available and extend the height a bit to see more results, which would make it easier to see that option without actually triggering an unexpected search.

Hey, 

Thats a shame. It does catch myself and others out given its difference from other search engines. I understand it changes the existing behaviour but it means new users need to learn how the search works contrary to other search features they use every day, so I think the current behaviour deducts from the UX.

But hey, its a one line change that didn't take long and its not the end of the world and you reviewed it pretty quickly 😄.

Thanks for chatting about it and reviewing the PR! 😸   

Regarding the proposed UX improvement, just so I understand correctly. Are you suggesting that the `Add new movie` search option should always show in the dropdown box with the suggestions being separately scrollable? 😸

robjackstewart:
> I've discussed this a bit on Discord with @stevietv and @mynameisbogdan and the consensus is changing the default behaviour that has existed for years doesn't make sense. It will lead to extra searches against Skyhook and disrupt the known experience.
> 
> One suggestion was to make the search option always available and extend the height a bit to see more results, which would make it easier to see that option without actually triggering an unexpected search.

Hey, 

Thats a shame. It does catch myself and others out given its difference from other search engines. I understand it changes the existing behaviour but it means new users need to learn how the search works contrary to other search features they use every day, so I think the current behaviour deducts from the UX.

But hey, its a one line change that didn't take long and its not the end of the world and you reviewed it pretty quickly 😄.

Thanks for chatting about it and reviewing the PR! 😸   

Regarding the proposed UX improvement, just so I understand correctly. Are you suggesting that the `Add new movie` search option should always show in the dropdown box with the suggestions being separately scrollable? 😸

markus101:
Yeah, effectively pinned to the bottom, like you'd see if there were only a couple of results:
image

I'm not sure how feasible it is to achieve given how the results are constructed especially if we want to keep the arrow navigation the same as it is now.

robjackstewart:
Change is not desired. Will review suggested alternative.

PR #6531, opened 2024-02-21
Fixed: Parsing of subtitle languages separated by dash

### LINKED ISSUE #6494: Sonarr erases subtitles extensions when renaming episodes

### Current Behavior

When renaming episodes, Sonarr also renames associated external subtitles. 
If subtitles extension is not just country code, Sonarr erases it. 
Examples:
```
01.ru-st_whisper-dfnet.srt -> Wellington Paranormal - S04E01 - Bird Woman Bluray-1080p.1.srt
01.ru-st_whisper-vad.srt -> Wellington Paranormal - S04E01 - Bird Woman Bluray-1080p.2.srt
01.ru-st_whisper-vad-dfnet.srt -> Wellington Paranormal - S04E01 - Bird Woman Bluray-1080p.3.srt
```
What's the reasong behind this? Sonarr didn't create these files, why is it deleting important information?
I believe this shouldn't happen, or at least there must be an option in the settings.

Other software (KODI, jellyfin) can properly detect country code and additional information for such subtitles.

### Expected Behavior

Only file names of the associated subtitles should be changed, extensions must be preserved.

### Steps To Reproduce

1) Add any TV show, add episodes manually
2) Add subtitle with extension `en-extra-info.srt`
3) Ask Sonarr to rename episodes
4) See that new subtitles extension is `1.srt`

### Environment

```markdown
- OS: Linux
- Sonarr: 3.0.10.1567
- Docker Install: yes, from linuxserver.io
- Using Reverse Proxy: no
- Browser: Chrome Version 121.0.6167.160
- Database:
```

Main

markus101:
#### Description

Allow for additional separators other than period between ISO language and additional subtitle information.

* Closes #6494

### DIFF
--- a/src/NzbDrone.Core.Test/ParserTests/LanguageParserFixture.cs
+++ b/src/NzbDrone.Core.Test/ParserTests/LanguageParserFixture.cs
@@ -442,6 +442,9 @@ namespace NzbDrone.Core.Test.ParserTests
         [TestCase("Name (2020) - S01E20 - [AAC 2.0].fra.forced.testtitle.ass", new[] { "forced" }, "testtitle", "French")]
         [TestCase("Name (2020) - S01E20 - [AAC 2.0].forced.fra.testtitle.ass", new[] { "forced" }, "testtitle", "French")]
         [TestCase("Name (2020) - S01E20 - [AAC 2.0].ru-something-else.srt", new string[0], "something-else", "Russian")]
+        [TestCase("Name (2020) - S01E20 - [AAC 2.0].Full Subtitles.eng.ass", new string[0], "Full Subtitles", "English")]
+        [TestCase("Name (2020) - S01E20 - [AAC 2.0].mytitle - 1.en.ass", new string[0], "mytitle", "English")]
+        [TestCase("Name (2020) - S01E20 - [AAC 2.0].mytitle.en.ass", new string[0], "mytitle", "English")]
         public void should_parse_title_and_tags(string postTitle, string[] expectedTags, string expectedTitle, string expectedLanguage)
         {
             var subtitleTitleInfo = LanguageParser.ParseSubtitleLanguageInformation(postTitle);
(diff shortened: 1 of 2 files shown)


JeWe37:
I believe this may cause problems. Subtitle titles with spaces are fairly common, and I believe e.g.
```
Name (2020) - S01E20 - [AAC 2.0].Full Subtitles.eng.ass
```
will break this. I would argue that in order to accept a title, it should be in the narrowly defined format from before, and rely on the solution proposed in [my previous comment](https://github.com/Sonarr/Sonarr/issues/6494#issuecomment-1947769387) to ignore the title.

JeWe37:
I think you're actually right about the EDIT, this came to mind later and I forgot to account for the fact that this still requires a known tag or any alpha string of less 2 or 3 characters to be present in the title preceding the number.

The "parse as full and clean up the subtitle tag" admittedly doesn't sound great to me. Which makes me ask, does anything actually support the full flag? 
- Plex doesn't based on its docs https://support.plex.tv/articles/200471133-adding-local-subtitles-to-your-media/
- Jellyfin/Emby definitely doesn't https://github.com/jellyfin/jellyfin/blob/ef96b5fa20f6f09e2671cf20efd075f134274314/Emby.Naming/Common/NamingOptions.cs#L291-L308
- Kodi doesn't seem to support anything except forced https://kodi.wiki/view/Subtitles

I think just removing this flag should actually cover most of the breakage based on a brief check in my own library. All the subs I saw where you had another tag that would get parsed incorrectly always just had that tag as the title. I.e. something like `Forced` as the title rather than `Forced Subtitles` so that reinterpretation may actually be valid.

markus101:
`full` does seems a little weird, not even sure what it'd mean other than the full episode, but then Sonarr wouldn't really care because a partial file isn't really supported. If the file has that it'd be considered part of the title and extracted, but that's reasonable.

I've added a commit that drops full and adds some test cases around the things mentioned here and of course the original issue is addressed.

PR #6533, opened 2024-02-21
Import list exclusion pagination

### LINKED ISSUE #6079: Pagination for Import List Exclusions

When having a long exclusion list for import lists, the Lists page in settings can become excessively laggy to load.

Use pagination for import list exclusions.

Using uBlock to filter out the import list exclusions.

### Anything else?

_No response_

CheAle14:
#### Description

Adds pagination and sorting to the import list exclusions table.  

I've also converted the import list exclusion section and its modal to typescript. This uses `pendingChanges` on the `state.settings.importListExclusions`, which appears to be present in all `AppSectionItemState<T>`s, so I added it to that interface to avoid having to cast it in the modal file.  

I've also fixed the `createSettingsSectionSelector` as I had assumed it was used by both sections that extended  `AppSectionItemState` and `AppSectionState`, but it is only used by the item ones, and relies on `item` and `pendingChanges` being present in it, so wouldn't work otherwise. As such, I've made it only allow sections which extend `AppSectionItemState`, and the above addition of `pendingChanges` means I could get rid of the cast to `AppStateWithPending`.

And finally, I noticed whilst converting the modal to typescript that I kept writing something like:

```ts
const [isOpen, setModalOpen] = useState(false);

const onModalClose = useCallback(() => {
  setModalOpen(true)
}, [setModalOpen]);

const onEditButtonEtcPress = useCallback(() => {
  setModalOpen(false)
}, [setModalOpen]);

```

So I've also added a `useModalOpenState` hook, which can be used like:

```ts
const [isOpen, onModalClose, onEditButtonEtcPress] = useModalOpenState(false);
```

I've kept those changes in separate commits so I can get rid of them or send them in a different PR if you'd like.

image

image

* Closes #6079


src/Sonarr.Api.V3/ImportLists/ImportListExclusionController.cs
```diff
@@ -29,9 +29,12 @@ protected override ImportListExclusionResource GetResourceById(int id)
 
         [HttpGet]
         [Produces("application/json")]
-        public List<ImportListExclusionResource> GetImportListExclusions()
+        public PagingResource<ImportListExclusionResource> GetImportListExclusions([FromQuery]PagingRequestResource paging)
```
markus101, line 32: Since this changes the structure of the return it's a breaking change. Instead of breaking the API this should be added as a separate route and the old route can be deprecated.

`/paged` on the existing route should be sufficient.

markus101 (changes requested):
Looks good, just need to address the breaking API change.

markus101 (approved):

CheAle14:
Sorry must've missed them, I've opened #6593 to fix those.

Issue #5598, opened 2023-04-30
Improve CF Comparison Between Release and File


markus101:
### Is there an existing issue for this?

- [X] I have searched the existing open and closed issues

### Is your feature request related to a problem? Please describe

Sometimes indexers create their own special release names that don't match the actual release name leading to download loops due to CF upgrades.

### Describe the solution you'd like

1. Use the current file name, which would require all CFs to be part of the file name format
2. Use current file name + import file name / scene name, this would need to dedupe all the available formats

### Describe alternatives you've considered

 See above

thezoggy:
If the release is grabbed from indexer (lets use usenet as example). I'd prefer that release name  be used as source of truth. As the filename may be obfuscated / might have mistake which indexer fixed up or packing mistake. Id rather use indexer and if wrong user can blacklist and move on. 

Now if the release from indexer is season pack, then set flag to fallback to using filename? 

Then if release is being imported manually, no grab. Then only thing you really can use is the foldername/filename?

bakerboy448:
Zoggy so you're saying even once on disk ignore the existing file name and only use the release title?  That's a problem and means the ffprobed media info will always be ignored.

It's a non-starter IMO.

post import:
- Current On-Disk Name
- SceneTitle OR original file name
- Q: How to handle when SceneTitle matched a CF that Current Name explicitly cannot be matched to
- Several downstream GHI in radarr: to review

Pre import:
- season packs: use scenetitle (file name)
- Single files: use scenetitle (file name fallback to folder fallback to release name

Grab:
- release name

yammes08:
Is some of this dependent on users choosing a naming strategy that captures various items of metadata? E.g., if I decided to name my files `Series Name (2023) - s01e01 - Episode Name`, then CFs that would match against ffprobed metadata wouldn't ever be aggregated from the on-disk name. Also not sure if that's a problem, as in this scenario you wouldn't be getting conflicts, even if the matched CFs are technically incorrect based on the actual file properties.

You could use a (temporary?) intermediate file that contains all the file metadata for matching/aggregating?

Regardless of the above, aggregating from each step of the process sounds like a good idea so long it's possible to a) discard CFs that end up not matching and b) correctly pick the right CF when conflicting options present. Would a user selectable advanced option be a step too far? 

(caveat, I probably don't understand this topic as well as others so feel free to disregard!)

thezoggy:
[...]

Issue #6550, opened 2024-02-27
Grab single episode releases for multi-episode files


rg9400:
### Is there an existing issue for this?

- [X] I have searched the existing open and closed issues

### Is your feature request related to a problem? Please describe

This is reopening the discussion for https://github.com/Sonarr/Sonarr/issues/4864 and https://github.com/Sonarr/Sonarr/issues/734. Since most networks have shifted to streaming services now, we are seeing a lot of older shows get re-released on new platforms. This often means that those streamers will split apart double-header episodes, but I've noticed Sonarr will completely miss those upgrades since it may have already imported a multi-episode file instead. There is no way for the user to know that a potential upgrade was completely missed, and thus a multi-episode file can actually act as a very hard block to Sonarr functionality.

### Describe the solution you'd like

I think Sonarr should grab individual episodes as upgrades for sure. When importing, it can require manual intervention from the user end to determine if they want to replace the multi-episode file, but this way, the release is not completely skipped. Ideally, Sonarr would also have some intelligence to determine if both (or however many) episodes mapped to the existing file have releases in the queue/pending manual import, then it should override that block and allow the upgrades through automatically. 

### Describe alternatives you've considered

The alternatives are to either not import multi-episode files, or to pay very close attention to each and every upgrade. That way, if you see a season being upgraded, you can verify if every episode was upgraded in that season, and if not, you can use the interactive search to upgrade the multi-episode file manually. However, this requires you to be aware that a file was missed. Since some of these shows have 20+ episodes, it is very easy to miss a particular episode, so you may never even know that an upgrade could be grabbed.

markus101:
Dupe of #734

No short term plans to put effort into this.

markus101:
Thinking about this further, this has the same underlying requirements of https://github.com/Sonarr/Sonarr/issues/6378 and as long as it doesn't exclusively apply to season packs it'd work.

PR #6567, opened 2024-02-29
New: Release Type (Single/Multi episode and Season Pack) for Custom Formats

### LINKED ISSUE #3562: Preferred Words overrule season pack preference
**Describe the bug**

A bonus to a specific release group prevents Sonarr from grabbing season packs without such bonus since it will first grab the single episode (on btn) and then reject the pack.

A fix is a bit of a challenge since Preferred Words overrules Protocol too, so we can't just move episode count up.

markus101:
#### Description
Adds a new CF type of `Release Type` which can be used to prefer or avoid releases based on whether they are single episode, multi-episode or season packs. The value of the release type is stored in history, blocklist and with the final episode file to prevent download loops that previously existed with release title regex matching alone.

In a roundabout way it closes #3562 by embracing the lose of season/multi-episode preference that currently exists, but also adds some changes towards #5598.

image

#### Database Migration
YES - 203

* Closes #3562

### DIFF
--- a/src/NzbDrone.Core/Datastore/Migration/203_release_type.cs
+++ b/src/NzbDrone.Core/Datastore/Migration/203_release_type.cs
@@ -1,0 +1,58 @@
+using System.Collections.Generic;
+using System.Data;
+using System.IO;
+using Dapper;
+using FluentMigrator;
+using NzbDrone.Common.Extensions;
+using NzbDrone.Core.Datastore.Migration.Framework;
+using NzbDrone.Core.Parser.Model;
+
+namespace NzbDrone.Core.Datastore.Migration
+{
+    [Migration(203)]
+    public class release_type : NzbDroneMigrationBase
+    {
+        protected override void MainDbUpgrade()
+        {
+            Alter.Table("Blocklist").AddColumn("ReleaseType").AsInt32().WithDefaultValue(0);
+            Alter.Table("EpisodeFiles").AddColumn("ReleaseType").AsInt32().WithDefaultValue(0);
+
+            Execute.WithConnection(UpdateEpisodeFiles);
+        }
+
+        private void UpdateEpisodeFiles(IDbConnection conn, IDbTransaction tran)
+        {
+            var updates = new List<object>();
+
+            using (var cmd = conn.CreateCommand())
+            {
+                cmd.Transaction = tran;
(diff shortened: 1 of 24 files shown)


src/NzbDrone.Core/Datastore/Migration/203_release_type.cs
```diff
@@ -0,0 +1,58 @@
...
+{
+    [Migration(203)]
+    public class release_type : NzbDroneMigrationBase
+    {
+        protected override void MainDbUpgrade()
+        {
+            Alter.Table("Blocklist").AddColumn("ReleaseType").AsInt32().WithDefaultValue(0);
+            Alter.Table("EpisodeFiles").AddColumn("ReleaseType").AsInt32().WithDefaultValue(0);
+
+            Execute.WithConnection(UpdateEpisodeFiles);
```
markus101, line 20: This only updates episode files to keep it lighter than trying to update existing history and blocklist, though it would be possible.

owine:
Would this allow for #6550 to be handled via CF scoring (negative score MultiEpisode, for example)? I believe there is separate handling that blocks those upgrades but it would be nice to set a preference for SingleEpisode via CF.

markus101:
No, you'd still run into things not being an upgrade for all episodes and getting rejected.

src/NzbDrone.Core/Datastore/Migration/203_release_type.cs
markus101: Parsing should be pretty quick, but let me try and benchmark it a bit to see if that'll trigger often.

markus101: I pulled in a list of 12060 file paths and ran them through the parser, 12045 of them parsed successfully, with 15 failing for whatever reason. Total time taken was 923ms. There will be some additional overhead reading/writing from/to the DB, but I think we still have a lot of overhead here.

markus101: What's at it in Prowlarr now? Not opposed to increasing the limit, but also, of course it's FreeBSD.

markus101: I've bumped it to 5 minutes, let's see how that goes.

markus101:
Ahh, I forgot to rename when I changed directions a bit. I'll fix in a separate PR with a migration to fix as well.

PR #6569, opened 2024-02-29
Improve AddSeries Messaging and Logging

bakerboy448:
#### Description
Improve UX

```none
2024-02-29 10:09:49.0|Error|AddSeriesService|TVDB ID 0 was not found, it may have been removed from TheTVDB.  Path:
```

no way to debug from the error without finding the addseries line

```none
2024-02-29 10:09:45.0|Info|AddSeriesService|Adding Series [0][Masters of the Air] Root Folder Path: [/data/media/media/tv]
```

### DIFF
--- a/src/NzbDrone.Core/Tv/AddSeriesService.cs
+++ b/src/NzbDrone.Core/Tv/AddSeriesService.cs
@@ -84,7 +84,7 @@ namespace NzbDrone.Core.Tv
 
                     if (seriesToAdd.Any(f => f.TvdbId == series.TvdbId))
                     {
-                        _logger.Debug("TVDB ID {0} was not added due to validation failure: Series {1} already exists on list", s.TvdbId, s);
+                        _logger.Trace("TVDB ID {0} was already added from another import list, not adding series {1} again", s.TvdbId, s);
                         continue;
                     }
 
--- a/src/NzbDrone.Core/Tv/AddSeriesService.cs
+++ b/src/NzbDrone.Core/Tv/AddSeriesService.cs
@@ -104,7 +104,7 @@ namespace NzbDrone.Core.Tv
                         throw;
                     }
 
-                    _logger.Debug("Seires {0} with TVDB ID {1} was not added due to validation failures. {2}", s, s.TvdbId, ex.Message);
+                    _logger.Debug("Series {0} with TVDB ID {1} was not added due to validation failures. {2}", s, s.TvdbId, ex.Message);
                 }
             }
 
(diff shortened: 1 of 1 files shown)


src/NzbDrone.Core/Tv/AddSeriesService.cs
```diff
@@ -121,7 +121,7 @@ private Series AddSkyhookData(Series newSeries)
             }
             catch (SeriesNotFoundException)
             {
-                _logger.Error("TVDB ID {0} was not found, it may have been removed from TheTVDB.  Path: {1}", newSeries.TvdbId, newSeries.Path);
+                _logger.Error("Series {0} with TVDB ID {1} was not found, it may have been removed from TheTVDB.  Path: {2}", newSeries, newSeries.TvdbId, newSeries.Path);
```
markus101, line 124:
```suggestion
                _logger.Error("Series {0} with TVDB ID {1} was not found, it may have been removed from TheTVDB. Path: {2}", newSeries, newSeries.TvdbId, newSeries.Path);
```

src/NzbDrone.Core/Tv/AddSeriesService.cs
```diff
@@ -78,13 +78,13 @@ public List<Series> AddSeries(List<Series> newSeries, bool ignoreErrors = false)
...
                     {
-                        _logger.Debug("TVDB ID {0} was not added due to validation failure: Series already exists in database", s.TvdbId);
+                        _logger.Debug("TVDB ID {0} was not added due to validation failure: Series {1} already exists in database", s.TvdbId, s);
                         continue;
                     }
 
                     if (seriesToAdd.Any(f => f.TvdbId == series.TvdbId))
                     {
-                        _logger.Debug("TVDB ID {0} was not added due to validation failure: Series already exists on list", s.TvdbId);
+                        _logger.Debug("TVDB ID {0} was not added due to validation failure: Series {1} already exists on list", s.TvdbId, s);
```
markus101, line 87: I'm not sure how much value this message actually adds. It's not being added because it's on another list that was processed first, but the series will still be added, but at worst it has different monitoring.

bakerboy448: Should we demote to trace?

markus101: Yeah and tweaking the messaging.

```suggestion
                        _logger.Trace("TVDB ID {0} was already added from another import list, not adding series {1} again", s.TvdbId, s);
```

markus101 (approved):

PR #656, opened 2015-07-02
Reimplement ToTv changes

Terebi42:
Mostly the same as the PR you guys made, a few tweaks.

PR #6573, opened 2024-03-01
`{Release Hash}` Renaming Token

### LINKED ISSUE #6570: `{Release Hash}` Renaming Token

This an Anime problem, and really only if you want to retain the CRC hash on the files.
e.g. 
```
[FroZen-EviL]_Miyuki_-_06_[DVD][2EDD54DC].mkv
[AEOG][1080p] A Couple of Cuckoos - 02 [BD HEVC 10bit Dual Audio AC3][6FB18E24].mkv
```

At the moment, there's no way to retain the hashes. They are currently parsed via **Parser**, but nothing is retained.

I'd like to add the `{Release Hash}` token (maybe just for Anime to avoid confusion?) to the **FileNameBuilder**.

I don't want to retain anything in the database from the initial parse, but have it be treated similar to `{Original FileName}` where it's reparsed out to make it available via the token at the time of rename. 

If the file is ever renamed without a CRC (via Sonarr or externally), it's gone and would have to be manually re-added to the filename. Sonarr would just have an empty token. Sonarr would never attempt to hash the file itself.

I used to use Preferred Words to keep the CRC, but it was decided in #5274 not to keep the functionality with the move to CustomFormats.

Even if #6431 is completed, CustomFormats uses the name of the CustomFormat as it's token value and not the matched value.

### Anything else?

I'm just checking to see if there's any push back against the idea. I'll submit the PR. 
I just thought of it when looking into what the effort would be for #6431

Alanoll:
#### Description
Implementing the `{Release Hash}` for use in renaming.

The parsed Release Hash is stored in the DB table `EpisodeFile`.

#### Database Migration
YES - 204

* Closes #6570


src/NzbDrone.Core/Datastore/Migration/203_add_release_hash.cs
```diff
@@ -0,0 +1,14 @@
+using FluentMigrator;
+using NzbDrone.Core.Datastore.Migration.Framework;
+
+namespace NzbDrone.Core.Datastore.Migration
+{
+    [Migration(203)]
```
Alanoll, line 6: I wasn't sure how DB migrations are kept in sync across forks, so I just incremented to the next ID.

src/NzbDrone.Core/Parser/Parser.cs
Alanoll, line 1: I tweaked the RegEx's to open them up alittle more to get any trailing Hashes in the strings, and allow for parenthesis as well. Half of them tried to do end string, but also had a "stop at first period" assuming the period would be the period of the file extension which isn't always the case.

All the unit tests passed, and I added more for parenthesized hashes.

src/NzbDrone.Core/Datastore/Migration/203_add_release_hash.cs
markus101: 203 is already used in https://github.com/Sonarr/Sonarr/pull/6567 will need to be bumped to 204

src/NzbDrone.Core.Test/OrganizerTests/FileNameBuilderTests/FileNameBuilderFixture.cs
```diff
@@ -991,6 +991,67 @@ public void should_use_updated_media_info_if_token_configured_and_revision_is_ol
...
+        [TestCase("[Doremi].Show.Title.5.Go.Go!.31[1280x720].[C65D4B1F]", "C65D4B1F")]
+        [TestCase("[Doremi].Show.Title.5.Go.Go!.31.[1280x720].mkv", "")]
+        [TestCase("[K-F] Series Title 214", "")]
+        [TestCase("[K-F] Series Title S10E14 214", "")]
+        [TestCase("[K-F] Series Title 10x14 214", "")]
+        [TestCase("[K-F] Series Title 214 10x14", "")]
+        [TestCase("Series Title - 031 - The Resolution to Kill [Lunar].avi", "")]
+        [TestCase("[ACX]Series Title 01 Episode Name [Kosaka] [9C57891E].mkv", "9C57891E")]
+        [TestCase("[S-T-D] Series Title! - 06 (1280x720 10bit AAC) [59B3F2EA].mkv", "59B3F2EA")]
+        public void should_replace_release_space_hash_with_scenename_hash(string sceneName, string hash)
```
markus101, line 1020:
```suggestion
        public void should_replace_release_hash_with_scenename_hash(string sceneName, string hash)
```

src/NzbDrone.Core.Test/OrganizerTests/FileNameBuilderTests/FileNameBuilderFixture.cs
```diff
@@ -991,6 +991,67 @@ public void should_use_updated_media_info_if_token_configured_and_revision_is_ol
...
+        [TestCase("[Doremi].Show.Title.5.Go.Go!.31[1280x720].[C65D4B1F].mkv", "C65D4B1F")]
+        [TestCase("[Doremi].Show.Title.5.Go.Go!.31.[1280x720].mkv", "")]
+        [TestCase("[K-F] Series Title 214.mkv", "")]
+        [TestCase("[K-F] Series Title S10E14 214.mkv", "")]
+        [TestCase("[K-F] Series Title 10x14 214.mkv", "")]
+        [TestCase("[K-F] Series Title 214 10x14.mkv", "")]
+        [TestCase("Series Title - 031 - The Resolution to Kill [Lunar].avi", "")]
+        [TestCase("[ACX]Series Title 01 Episode Name [Kosaka] [9C57891E].mkv", "9C57891E")]
+        [TestCase("[S-T-D] Series Title! - 06 (1280x720 10bit AAC) [59B3F2EA].mkv", "59B3F2EA")]
+        public void should_replace_release_space_hash_with_filename_hash(string fileName, string hash)
```
markus101, line 1045:
```suggestion
        public void should_replace_release_hash_with_filename_hash(string fileName, string hash)
```

src/NzbDrone.Core/Organizer/FileNameBuilder.cs
```diff
@@ -1101,6 +1102,27 @@ private string GetOriginalFileName(EpisodeFile episodeFile, bool useCurrentFilen
             return Path.GetFileNameWithoutExtension(episodeFile.RelativePath);
         }
 
+        private string GetReleaseHash(EpisodeFile episodeFile)
```
markus101, line 1105: This logic can all be done in the migration, then it doesn't need to be done at the time of naming. You could use what this PR does as an example. https://github.com/Sonarr/Sonarr/pull/656

markus101 (changes requested):

src/NzbDrone.Core/Organizer/FileNameBuilder.cs
Alanoll: The db migration now performs the parsing. 
https://github.com/Sonarr/Sonarr/pull/6567/files#r1507044123

The timeout y'all discussed in the other PR may be an issue here as well, since it's also reparsing all titles.

Alanoll:
- Updated the DB migration to perform the parsing, and incremented it's ID to 204.
- remove the Unit Tests that were testing the title parsing within the FileNameBuilder
- added a Unit Test to ensure null/empty Release Hashes return empty string correctly

Alanoll:
Realized this morning that I forgot to add the new token to the File Naming Tokens dialog. Added it to the dialog and shown only for Anime patterns.

src/NzbDrone.Core/Datastore/Migration/204_add_release_hash.cs
```diff
@@ -0,0 +1,75 @@
...
+                cmd.Transaction = tran;
+                cmd.CommandText = "SELECT \"Id\", \"SceneName\", \"RelativePath\", \"OriginalFilePath\" FROM \"EpisodeFiles\" WHERE \"ReleaseHash\" IS NULL";
+
+                using var reader = cmd.ExecuteReader();
+                while (reader.Read())
+                {
+                    var id = reader.GetInt32(0);
+                    var sceneName = reader.IsDBNull(1) ? null : reader.GetString(1);
+                    var relativePath = reader.IsDBNull(2) ? null : reader.GetString(2);
+                    var originalFilePath = reader.IsDBNull(3) ? null : reader.GetString(3);
```
markus101, line 36: These can be simplified to `var x= reader[n] as string;` which correctly handles null values.

src/NzbDrone.Core/Datastore/Migration/204_add_release_hash.cs
```diff
@@ -0,0 +1,75 @@
...
+
+                    string title = null;
+                    if (sceneName.IsNotNullOrWhiteSpace())
+                    {
+                        title = sceneName;
+                    }
+                    else if (relativePath.IsNotNullOrWhiteSpace())
```
markus101, line 43: This will never be null. so the `originalFilePath` branch will never be hit.

src/NzbDrone.Core/Datastore/Migration/204_add_release_hash.cs
```diff
@@ -0,0 +1,75 @@
...
+                    {
+                        title = Path.GetFileNameWithoutExtension(relativePath);
+                    }
+                    else if (originalFilePath.IsNotNullOrWhiteSpace())
+                    {
+                        title = Path.GetFileNameWithoutExtension(originalFilePath);
+                    }
+
+                    if (!title.IsNullOrWhiteSpace())
```
markus101, line 52: Since the relative path probably doesn't have the hash (given it was only available via a Preferred Word match), I think this should be two tiered.

1. Check the scene name or original file name, giving preference to the scene name as above if either is available
2. Check the existing file name, which will cover cases where the original information is unavailable, but the file name does have it.

Something like: 

```
ParsedEpisodeInfo parsedEpisodeInfo = null;

var originalTitle = sceneName ?? originalFilePath.IsNotNullOrWhiteSpace() ? Path.GetFileNameWithoutExtension(originalFilePath) : null;

if (originalTitle.IsNotNullOrWhiteSpace())
{
  parsedEpisodeInfo = Parser.Parser.ParseTitle(originalTitle);
}

if ( parsedEpisodeInfo?.ReleaseHash.IsNullOrWhiteSpace())
{
  parseEpisodeInfo = Parser.Parser.ParseTitle(Path.GetFileNameWithoutExtension(relativePath));
}

if (parsedEpisodeInfo?.ReleaseHash.IsNotNullOrWhiteSpace())
{
  updates.Add(new
	{
	    Id = id,
	    ReleaseHash = parsedEpisodeInfo.ReleaseHash
	});
}
```

This way it's best effort to get the value and at most will parse 2x per file.

markus101 (changes requested):

src/NzbDrone.Core/Datastore/Migration/204_add_release_hash.cs
```diff
@@ -0,0 +1,75 @@
...
+        }
+
+        private void UpdateEpisodeFiles(IDbConnection conn, IDbTransaction tran)
+        {
+            var updates = new List<object>();
+
+            using (var cmd = conn.CreateCommand())
+            {
+                cmd.Transaction = tran;
+                cmd.CommandText = "SELECT \"Id\", \"SceneName\", \"RelativePath\", \"OriginalFilePath\" FROM \"EpisodeFiles\" WHERE \"ReleaseHash\" IS NULL";
```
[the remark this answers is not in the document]
Alanoll: Removed the condition

src/NzbDrone.Core/Datastore/Migration/204_add_release_hash.cs
Alanoll: Good tip. I'm so ingrained in the older pattern, never thought of using that.

src/NzbDrone.Core/Datastore/Migration/204_add_release_hash.cs
Alanoll: Obsolete

src/NzbDrone.Core/Datastore/Migration/204_add_release_hash.cs
Alanoll: Implemented the suggested changes, but was explicit instead of the null coalescing operator in case `sceneName` somehow is an empty string since FileNameBuilder performs the same check

src/NzbDrone.Core/Datastore/Migration/204_add_release_hash.cs
```diff
@@ -0,0 +1,71 @@
...
+                while (reader.Read())
+                {
+                    var id = reader.GetInt32(0);
+                    var sceneName = reader[1] as string;
+                    var relativePath = reader[2] as string;
+                    var originalFilePath = reader[3] as string;
+
+                    ParsedEpisodeInfo parsedEpisodeInfo = null;
+
+                    var originalTitle = sceneName.IsNotNullOrWhiteSpace() ? sceneName : (originalFilePath.IsNotNullOrWhiteSpace() ? Path.GetFileNameWithoutExtension(originalFilePath) : null);
```
markus101, line 41: Since `sceneName` is null when it's not set this can be shortened to:

```suggestion
                    var originalTitle = sceneName ?? originalFilePath.IsNotNullOrWhiteSpace() ? Path.GetFileNameWithoutExtension(originalFilePath) : null;
```

markus101 (changes requested):

src/NzbDrone.Core/Datastore/Migration/204_add_release_hash.cs
Alanoll: Can SceneName ever be the empty string in the database?
I went the IsNullOrWhiteSpace route since that's what FileNameBuilder did.

markus101: Should only be only be null or a valid value unless someone manually altered their DB to set it to empty string.

Yeah, we use `Is(Not)NullOrWhiteSpace` across the app to avoid having to check bot, but in the migration just the null check is fine.

markus101: Thinking about it a bit more, not much harm in the empty string check, so this would be what I'd do to avoid the double ternary.

```suggestion
                    var originalTitle = sceneName;
                    
                     if (originalTitle.IsNullOrWhiteSpace() && originalFilePath.IsNotNullOrWhiteSpace())
                     {
                         originalTitle = Path.GetFileNameWithoutExtension(originalFilePath);
                     }
```

src/NzbDrone.Core/Parser/Parser.cs
```diff
@@ -492,7 +492,7 @@ public static class Parser
         private static readonly Regex FileExtensionRegex = new Regex(@"\.[a-z0-9]{2,4}$",
                                                                 RegexOptions.IgnoreCase | RegexOptions.Compiled);
 
-        private static readonly RegexReplace SimpleTitleRegex = new RegexReplace(@"(?:(480|540|576|720|1080|2160)[ip]|[xh][\W_]?26[45]|DD\W?5\W1|[<>?*]|848x480|1280x720|1920x1080|3840x2160|4096x2160|(8|10)b(it)?|10-bit)\s*?",
```
Alanoll, line 495: This was causing any hash value with either `8B` or `10B` to be removed when the title is simplified. For example `[4228bb8c]` would become `[422b8c]` and then failed the hash regex because it's not 8 characters.

The updated expression should still match `8b`, `8bit`, `10b`, `10bit`, or `10-bit` for removal, but `8b` and `10b` can't have an immediate word character after it. I didn't use `\w` because of the underscore and `_8b_` could be what people use in their files.

Alanoll: There's still cases where the hash won't match, but I think a decision would need to be made.
Hashes that end with `10b` or `8b` still might not match e.g. `[15AF298B]` or `[15AF210B]`

Alanoll: An alternative could be to check if the `<hash>` match has content, and if not, do an explicit RegEx against the raw release title trying to find the hash.

Pros:
- Don't have to keep tweaking existing expressions for niche cases of Anime naming
- Could find hashes in non-anime releases if they had them (I don't know of any off the top of my head)

Con:
- Increases processing time with an additional check. The check is only perform when one of the title matches hits, so it's not a 2x across the board. It would be a simple regex `(?<hash>[(\[][0-9a-f]{8}[)\]])$`, so should be relatively quick.
- Could find hashes in non-anime releases (leading to questions of "why does this one have a hash but that one doesn't?")

markus101: I think you should be able to do a lookbehind and only match if it's proceeded by a space, underscore, or period, or a negative look behind to make sure it's not proceeded by a valid character in the hash.

src/NzbDrone.Core/Parser/Parser.cs
```diff
@@ -492,7 +492,7 @@ public static class Parser
...
+        private static readonly RegexReplace SimpleTitleRegex = new RegexReplace(@"(?:(480|540|576|720|1080|2160)[ip]|[xh][\W_]?26[45]|DD\W?5\W1|[<>?*]|848x480|1280x720|1920x1080|3840x2160|4096x2160|(8|10)(b(?![a-z0-9])|bit)|10-bit)\s*?",
```
markus101, line 495:
```suggestion
        private static readonly RegexReplace SimpleTitleRegex = new RegexReplace(@"(?:(480|540|576|720|1080|2160)[ip]|[xh][\W_]?26[45]|DD\W?5\W1|[<>?*]|848x480|1280x720|1920x1080|3840x2160|4096x2160|(?<![]a-f0-9])(8|10)b(it)?|10-bit)\s*?",
```

This works with the example title provided.

Alanoll: I combined our two patterns. Your look behind, but with my look ahead.

Some mocked up example parses:
```
[DHD] Series Title! - 01 - Episode Title [1080p][10bit][15AF298B]
[DHD] Series Title! - 01 - Episode Title [1080p 10bit][15AF298B]
[DHD] Series Title! - 01 - Episode Title [1080p 10b][15AF298B]
[DHD] Series Title! - 01 - Episode Title [1080p][10b][15AF298B]
[DHD] Series Title! - 01 - Episode Title [1080p 10b][8B15AF29]
[DHD] Series Title! - 01 - Episode Title [1080p][10b][8B15AF29]
[DHD] Series Title! - 01 - Episode Title [1080p][8bit][15AF298B]
[DHD] Series Title! - 01 - Episode Title [1080p 8bit][15AF298B]
[DHD] Series Title! - 01 - Episode Title [1080p 8b][15AF298B]
[DHD] Series Title! - 01 - Episode Title [1080p][8b][15AF298B]
[DHD] Series Title! - 01 - Episode Title [1080p 8b][8B15AF29]
[DHD] Series Title! - 01 - Episode Title [1080p][8b][8B15AF29]
```

With just the suggested change:
image

With the one I committed:
image

src/NzbDrone.Core.Test/ParserTests/AnimeMetadataParserFixture.cs
```diff
@@ -22,12 +22,39 @@ public class AnimeMetadataParserFixture : CoreTest
...
+        [TestCase("[K-F] Series Title 214", "K-F", "")]
+        [TestCase("[K-F] Series Title S10E14 214", "K-F", "")]
+        [TestCase("[K-F] Series Title 10x14 214", "K-F", "")]
+        [TestCase("[K-F] Series Title 214 10x14", "K-F", "")]
+        [TestCase("Series Title - 031 - The Resolution to Kill [Lunar].avi", "Lunar", "")]
+        [TestCase("[ACX]Series Title 01 Episode Name [Kosaka] (9C57891E).mkv", "ACX", "9C57891E")]
+        [TestCase("[S-T-D] Series Title! - 06 (1280x720 10bit AAC) (59B3F2EA).mkv", "S-T-D", "59B3F2EA")]
+        [TestCase("[DHD] Series Title! - 08 (1280x720 10bit AAC) (8B8BF2EA).mkv", "DHD", "8B8BF2EA")]
+        [TestCase("[DHD] Series Title! - 10 (1280x720 10bit AAC) (10BBF2EA).mkv", "DHD", "10BBF2EA")]
+        public void should_parse_absolute_numbers_parenthesized_hash(string postTitle, string subGroup, string hash)
```
markus101, line 52: No need to create a duplicate function with many duplicate tests, add the relevant tests to `should_parse_absolute_numbers`.

markus101 (changes requested):

src/NzbDrone.Core/Parser/Parser.cs
markus101: Currently only tests for `8b` and `10b` are covered, would be good to add additional ones.

src/NzbDrone.Core/Organizer/FileNameBuilder.cs
```diff
@@ -623,6 +623,7 @@ private void AddEpisodeFileTokens(Dictionary<string, Func<TokenMatch, string>> t
             tokenHandlers["{Original Title}"] = m => GetOriginalTitle(episodeFile, useCurrentFilenameAsFallback);
             tokenHandlers["{Original Filename}"] = m => GetOriginalFileName(episodeFile, useCurrentFilenameAsFallback);
             tokenHandlers["{Release Group}"] = m => Truncate(episodeFile.ReleaseGroup, m.CustomFormat) ?? m.DefaultValue("Sonarr");
+            tokenHandlers["{Release Hash}"] = m => GetReleaseHash(episodeFile);
```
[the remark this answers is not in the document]
Alanoll: Done

src/NzbDrone.Core.Test/ParserTests/AnimeMetadataParserFixture.cs
Alanoll: I recombined them to `should_parse_absolute_numbers` but also renamed it to reflect what is being tested.

src/NzbDrone.Core/MediaFiles/EpisodeImport/ImportApprovedEpisodes.cs
```diff
@@ -95,6 +95,7 @@ public List<ImportResult> Import(List<ImportDecision> decisions, bool newDownloa
                     episodeFile.SeasonNumber = localEpisode.SeasonNumber;
                     episodeFile.Episodes = localEpisode.Episodes;
                     episodeFile.ReleaseGroup = localEpisode.ReleaseGroup;
+                    episodeFile.ReleaseHash = localEpisode.FileEpisodeInfo?.ReleaseHash;
```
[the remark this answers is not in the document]
Alanoll: Implemented.

src/NzbDrone.Core/Parser/Parser.cs
Alanoll: What additional ones where you thinking of?
I split off the explicit 8b and 10b within the hashes to another function, with begins with, ends with, and contains variants.

Alanoll:
Rebased onto develop to clear merge conflict from #6608

src/NzbDrone.Core/Parser/Parser.cs
markus101: Perfect, I think I just missed them as I reviewed.

markus101 (approved):

PR #6574, opened 2024-03-02
Update to .NET 8

Gauvino:
#### Description
Overall update to .NET 8, need further testing and building test to see if everything working well.

If upgrading is simple as that I can do it on every other arr*

#### Changes
 - Update project to .NET 8
 - Update .NETFramework to 4.8.1
 - Update NuGet packages
 - Update Dockerfile from focal to jammy

PR #6580, opened 2024-03-02
Bump to net8.0

Gauvino:
#### Description
Overall update to .NET 8, need further testing and building test to see if everything working well.

#### Changes
 - Update project to .NET 8
 - Update .NETFramework to 4.8.1
 - Update NuGet deps
 - Remove clock and change add to append
 - Remove serialization since it's obsolete
 - OS-release update to Ubuntu 20.04

#### Todos
- [X] Tests
- [X] [Wiki Updates](https://wiki.servarr.com) : https://github.com/Servarr/Wiki/pull/327


Gauvino:
Switching discussion from https://github.com/Sonarr/Sonarr/pull/6574 to here here are the question :

The question is does mono is still needed ?

markus101:
Why a new PR to fracture the discussion?

No, mono isn't needed.

Gauvino:
> Why a new PR to fracture the discussion?
> 
> No, mono isn't needed.

New PR for less commit, and will investigate to remove mono, and could i do the same PR for radarr and lidarr ?

thezak48:
is .NET 8 even ported to BSD yet?

Gauvino:
Does this respond to your question ?
https://github.com/Thefrank/dotnet-freebsd-crossbuild/releases
from https://wiki.freebsd.org/.NET#releases

markus101:
You don't need to open a new PR to have less commits, git's interactive rebase can be used to fixup/squash commits into a single commit, but in such a large PR that will actually make reviewing more difficult as it can't be done commit by commit. Don't close a PR to open another one for the same thing, there is no reason to and doing so doesn't instill confidence.

What Mono pieces do you think need to be removed? Sonarr no longer uses mono for the runtime on linux/mac, but the Sonarr.Mono library still relies on a couple mono libraries for Unix specific things, if there are alternatives then a separate PR to replace them would be useful, but shouldn't be done as part of the move to .net 8.

Gauvino:
> You don't need to open a new PR to have less commits, git's interactive rebase can be used to fixup/squash commits into a single commit, but in such a large PR that will actually make reviewing more difficult as it can't be done commit by commit. Don't close a PR to open another one for the same thing, there is no reason to and doing so doesn't instill confidence.
> 
I'm not really good with git so I didn't knew I could make such thing to have a clean PR
> What Mono pieces do you think need to be removed? Sonarr no longer uses mono for the runtime on linux/mac, but the Sonarr.Mono library still relies on a couple mono libraries for Unix specific things, if there are alternatives then a separate PR to replace them would be useful, but shouldn't be done as part of the move to .net 8.
I don't think I know this well the project to be able to make a clean removal of mono.

What would it need to have .NET 8 on every repo for servarr, and also maybe I could look to use new optimization from new NET

markus101:
> I don't think I know this well the project to be able to make a clean removal of mono.

That's the problem when you open another PR and ask a different question than the one asked in that PR, Baker specifically asked about mono in the docker build, which is not needed. Sonarr.Mono and the Mono.Native libraries used within are needed and if they are intended to be removed should be removed in a separate PR.

distribution/docker-build/Dockerfile
```diff
@@ -1,21 +1,22 @@
-FROM ubuntu:focal AS builder
+FROM ubuntu:jammy AS builder
```
markus101, line 1: Don't make these docker changes as part of the .net 8 migration.

docker/tests/mono/complete/Dockerfile
```diff
@@ -1,11 +1,13 @@
-FROM ubuntu:xenial
+FROM ubuntu:jammy
```
markus101, line 1: Same thing here.

docker/tests/mono/sonarr/Dockerfile
```diff
@@ -1,13 +1,15 @@
-FROM ubuntu:xenial
+FROM ubuntu:jammy
```
markus101, line 1: Same here.

docker/tests/run-all.sh
```diff
@@ -25,10 +25,10 @@ done
 MONO_VERSIONS=""
 
 # Future versions
-MONO_VERSIONS="$MONO_VERSIONS 6.10=preview-xenial"
+MONO_VERSIONS="$MONO_VERSIONS 6.12=preview-focal"
```
markus101, line 28: And here.

markus101 (changes requested):
Don't include the docker mono changes.

distribution/docker-build/Dockerfile
Gauvino: It is a part of the .NET 8 update since .NET 8 is not supported on other version of Ubuntu than >22.04
All of the 20.04 and older distribution or not supported by .NET 8 

https://learn.microsoft.com/en-us/dotnet/core/install/linux-ubuntu#supported-distributions

markus101: I understand, but these docker containers were set up for testing mono which isn't necessary and can be upgraded independently of the .net 8 upgrade.

Gauvino: So if i understand everything right, you want me make a PR just for the .NET 8 update on csproj files and other but don't touch the docker file or else related to mono ?

markus101: No, just remove these changes from this PR.

joshka:
It looks like the requested changes were made by the submitter, and at a superficial level without running it, this looks ok. @markus101 I think you need to approve running the workflow checks in order to get CI feedback on this.

joshka:
That most recent push introduced a huge amount of changes. What's going on there @Gauvino ?

Gauvino:
> That most recent push introduced a huge amount of changes. What's going on there @Gauvino ?

Yes I made a mistake, I will revert everything their is no hurry I think

markus101:
Converting this to a draft since there are a number of conflicts to address and won't be landed until we finalize plans for v5 due to backwards compatibility issues with older OSes.

Issue #6431, opened 2024-01-26
Add Custom Format "Rename To"


salomj:
### Is there an existing issue for this?

- [X] I have searched the existing open and closed issues

### Is your feature request related to a problem? Please describe

I would like to have more qualified Custom Format names, to make it easier to see what is going on, such as `Audio: AAC` instead of just `aac`, while still allowing it to be renamed when checking `Include Custom Format when Renaming`.

### Describe the solution you'd like

Another text field called something like `Rename To` when checking `Include Custom Format when Renaming`

### Describe alternatives you've considered

The two alternatives is to set the `name` to what you want it to be, which makes `Custom Formats` pretty messy, or to add a second format with the exact same conditions with renaming checked.

markus101:
We need to determine what the behaviour is if this is not set or whether it must always be set if renaming is enabled or  if it'll fall back to use name if unset (which will be a little more flexible with the current API). This field will need to also be unique like name is (and also not be the same as the name).

bakerboy448:
Xref https://github.com/Radarr/Radarr/issues/6801

Alanoll:
@salomj would the Radarr suggestion would for you? e.g. `{CustomFormat:FormatName}`

I implemented something similar in the v3 release (#4453), but when work started on v4 and the transition to CustomFormats it was dropped because it seemed like I may have been the only one using it :)

I can look into adding it back in.

salomj:
> @salomj would the Radarr suggestion would for you? e.g. `{CustomFormat:FormatName}`
> 
> I implemented something similar in the v3 release (#4453), but when work started on v4 and the transition to CustomFormats it was dropped because it seemed like I may have been the only one using it :)
> 
> I can look into adding it back in.

Can you specify each individual Custom Formats that way? Because I'll have more than one Custom Format being renamed.

The way they are right now, I have to split my Custom Formats into its matching profile and its renaming profile.

Alanoll:
In theory you would be able to do:
`{Episode Title}.s{Season:00}e{Episode:00}.{Custom Format:AudioAAC}.{Quality Title}.{Custom Format:SomeOtherFormat}-{Custom Formats}`

And if the `Custom Format` was matched, it would put that format's name `AudioAAC` or `SomeOtherFormat` where you told it to. `Custom Formats` would be all matched formats like it is now.

Alanoll:
I implemented the token in PR #6585. 

[...]

PR #4453, opened 2021-04-21
Add support for more precise PreferredWord placement in Rename patterns

Alanoll:
#### Database Migration
NO

#### Description
Episode Naming formats currently support the {Preferred Words} token, which combines all matched preferred words from Release Profiles and replaces the token with them.

This PR adds support to specific Preferred Words to be more precise in the placement of the naming pattern.

Example Pattern:

> {Series Title}.s{season:00}e{episode:00}.{absolute:000}.{Preferred Words EpisodeVersion}.{Episode Title}.{Quality Title}.{Preferred Words}.{Release Group}

Example Release Profile

#### Todos
- [x] Tests
- [ ] Wiki Updates

#### Issues Fixed or Closed by this PR

*

PR #6585, opened 2024-03-02
`{Custom Format:Format}` Rename Token

Alanoll:
#### Description
Implementing the `{Custom Format:Format}` rename token to allow position of Custom Formats when renaming.

This re-implements #4453 for Custom Formats, and would satisfy [Radarr/Radarr#6801](https://github.com/Radarr/Radarr/issues/6801) if it gets cherry-picked to over there.

It is a smaller change because it's really just allowing exact placement of the Format Name instead of the old Preferred Words implementation that replaced with the matched values.

#### Database Migration
NO

* #6431 - It doesn't close this issue, but is related.

### DIFF
--- a/src/NzbDrone.Core/Organizer/FileNameBuilder.cs
+++ b/src/NzbDrone.Core/Organizer/FileNameBuilder.cs
@@ -705,7 +705,7 @@ namespace NzbDrone.Core.Organizer
                     return string.Empty;
                 }
 
-                return string.Join(" ", customFormats.Where(x => x.IncludeCustomFormatWhenRenaming && x.Name == m.CustomFormat));
+                return customFormats.Where(x => x.IncludeCustomFormatWhenRenaming && x.Name == m.CustomFormat).FirstOrDefault()?.ToString() ?? string.Empty;
             };
         }
 
--- a/src/NzbDrone.Core/Organizer/FileNameBuilder.cs
+++ b/src/NzbDrone.Core/Organizer/FileNameBuilder.cs
@@ -698,6 +698,15 @@ namespace NzbDrone.Core.Organizer
             }
 
             tokenHandlers["{Custom Formats}"] = m => string.Join(" ", customFormats.Where(x => x.IncludeCustomFormatWhenRenaming));
+            tokenHandlers["{Custom Format}"] = m =>
+            {
+                if (m.CustomFormat.IsNullOrWhiteSpace())
+                {
+                    return string.Empty;
+                }
+
+                return string.Join(" ", customFormats.Where(x => x.IncludeCustomFormatWhenRenaming && x.Name == m.CustomFormat));
+            };
         }
 
         private void AddIdTokens(Dictionary<string, Func<TokenMatch, string>> tokenHandlers, Series series)
--- a/src/NzbDrone.Core/Organizer/FileNameBuilder.cs
+++ b/src/NzbDrone.Core/Organizer/FileNameBuilder.cs
@@ -47,7 +47,7 @@ namespace NzbDrone.Core.Organizer
         private readonly ICached<bool> _patternHasEpisodeIdentifierCache;
         private readonly Logger _logger;
 
-        private static readonly Regex TitleRegex = new Regex(@"(?<escaped>\{\{|\}\})|\{(?<prefix>[- ._\[(]*)(?<token>(?:[a-z0-9]+)(?:(?<separator>[- ._]+)(?:[a-z0-9]+))?)(?::(?:[ ]+)?(?<customFormat>[ a-z0-9+-]+(?<!-| )))?(?<suffix>[- ._)\]]*)\}",
+        private static readonly Regex TitleRegex = new Regex(@"(?<escaped>\{\{|\}\})|\{(?<prefix>[- ._\[(]*)(?<token>(?:[a-z0-9]+)(?:(?<separator>[- ._]+)(?:[a-z0-9]+))?)(?::(?<customFormat>[ a-z0-9+-]+(?<![- ])))?(?<suffix>[- ._)\]]*)\}",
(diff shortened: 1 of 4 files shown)


src/NzbDrone.Core/Organizer/FileNameBuilder.cs
```diff
@@ -698,6 +698,15 @@ private void AddCustomFormats(Dictionary<string, Func<TokenMatch, string>> token
...
 
             tokenHandlers["{Custom Formats}"] = m => string.Join(" ", customFormats.Where(x => x.IncludeCustomFormatWhenRenaming));
+            tokenHandlers["{Custom Format}"] = m =>
+            {
+                if (m.CustomFormat.IsNullOrWhiteSpace())
+                {
+                    return string.Empty;
+                }
+
+                return string.Join(" ", customFormats.Where(x => x.IncludeCustomFormatWhenRenaming && x.Name == m.CustomFormat));
```
Alanoll, line 708: If Custom Format names are unique, then this could be simplified to just
`customFormats.Where(x => x.IncludeCustomFormatWhenRenaming && x.Name == m.CustomFormat).FirstOrDefault()?.ToString() ?? string.Empty`

markus101: They are unique. https://github.com/Sonarr/Sonarr/blob/develop/src/Sonarr.Api.V3/CustomFormats/CustomFormatController.cs#L28-L29

src/NzbDrone.Core/Organizer/FileNameBuilder.cs
```diff
@@ -698,6 +698,15 @@ private void AddCustomFormats(Dictionary<string, Func<TokenMatch, string>> token
             }
 
             tokenHandlers["{Custom Formats}"] = m => string.Join(" ", customFormats.Where(x => x.IncludeCustomFormatWhenRenaming));
+            tokenHandlers["{Custom Format}"] = m =>
+            {
+                if (m.CustomFormat.IsNullOrWhiteSpace())
+                {
+                    return string.Empty;
+                }
```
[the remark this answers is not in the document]
Alanoll: The original {Custom Formats} is untouched to display all formats matches

src/NzbDrone.Core/Organizer/FileNameBuilder.cs
```diff
@@ -698,6 +698,15 @@ private void AddCustomFormats(Dictionary<string, Func<TokenMatch, string>> token
...
+
+                return customFormats.Where(x => x.IncludeCustomFormatWhenRenaming && x.Name == m.CustomFormat).FirstOrDefault()?.ToString() ?? string.Empty;
```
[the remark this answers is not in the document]
Alanoll: This PR only adds the ability to position Custom Format name matches. The ability to use a different word (e.g. WEB-DL instead of the format name "WEB-DL PVP") I held off on because of unanswered questions and needs a separate PR.

markus101 (changes requested):
Please add some tests to verify the behaviour (and prevent breaking changes in the future) in a new fixture in https://github.com/Sonarr/Sonarr/tree/develop/src/NzbDrone.Core.Test/OrganizerTests/FileNameBuilderTests

Alanoll:
I added Unit Tests for both `{Custom Formats}` and `{Custom Format:Format}` tokens.

Something I ran into was that Custom Formats allow for spaces in their Names. Multiple Trash Guide formats have them ([example](https://trash-guides.info/Sonarr/sonarr-collection-of-custom-formats/#remux-tier-01)). I tweaked the `TitleRegex` to support spaces within the format, but to ignore any leading/trailing spaces.

Did some quick tests in an expression tester:
image

All existing FileNameBuilder tests passed after the change as well.

src/NzbDrone.Core/Organizer/FileNameBuilder.cs
```diff
@@ -47,7 +47,7 @@ public class FileNameBuilder : IBuildFileNames
         private readonly ICached<bool> _patternHasEpisodeIdentifierCache;
         private readonly Logger _logger;
 
-        private static readonly Regex TitleRegex = new Regex(@"(?<escaped>\{\{|\}\})|\{(?<prefix>[- ._\[(]*)(?<token>(?:[a-z0-9]+)(?:(?<separator>[- ._]+)(?:[a-z0-9]+))?)(?::(?<customFormat>[a-z0-9+-]+(?<!-)))?(?<suffix>[- ._)\]]*)\}",
+        private static readonly Regex TitleRegex = new Regex(@"(?<escaped>\{\{|\}\})|\{(?<prefix>[- ._\[(]*)(?<token>(?:[a-z0-9]+)(?:(?<separator>[- ._]+)(?:[a-z0-9]+))?)(?::(?:[ ]+)?(?<customFormat>[ a-z0-9+-]+(?<!-| )))?(?<suffix>[- ._)\]]*)\}",
```
markus101, line 50: Suggestion is more inline with what we do elsewhere with dashes and spaces and removed what appears to be an unnecessary space match

```suggestion
        private static readonly Regex TitleRegex = new Regex(@"(?<escaped>\{\{|\}\})|\{(?<prefix>[- ._\[(]*)(?<token>(?:[a-z0-9]+)(?:(?<separator>[- ._]+)(?:[a-z0-9]+))?)(?::(?<customFormat>[ a-z0-9+-]+(?<![- ])))?(?<suffix>[- ._)\]]*)\}",
```

Alanoll: Applied your suggestion. The leading space match was to allow for `{Custom Format: FORMATNAME}` with a leading space, but that seems unnecessary unless someone requests it

markus101 (approved):

PR #6592, opened 2024-03-03
Fixed: URL Base setting for Kodi connections

mynameisbogdan:
#### Description
Update Address for Kodi after 9fd193d2a82d5c2cdc0f36c1f984e4b6b68aaa8d

### DIFF
--- a/src/NzbDrone.Core/Notifications/Xbmc/XbmcSettings.cs
+++ b/src/NzbDrone.Core/Notifications/Xbmc/XbmcSettings.cs
@@ -14,16 +14,18 @@ namespace NzbDrone.Core.Notifications.Xbmc
         {
             RuleFor(c => c.Host).ValidHost();
             RuleFor(c => c.DisplayTime).GreaterThanOrEqualTo(2);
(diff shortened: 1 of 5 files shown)


src/NzbDrone.Core/Notifications/Xbmc/XbmcSettings.cs
```diff
@@ -65,7 +66,7 @@ public XbmcSettings()
         public bool AlwaysUpdate { get; set; }
 
         [JsonIgnore]
-        public string Address => $"{Host.ToUrlHost()}:{Port}";
```
markus101, line 68: `Address` is only used for logging, https://github.com/Sonarr/Sonarr/blob/develop/src/NzbDrone.Core/Notifications/Xbmc/XbmcJsonApiProxy.cs#L78 is where we use it for the actual connection. We either need to make the default `/jsonrpc` or combine them there. I think making the default is the better route, in the future if they add a different API then we could just add it as a separate connection.

markus101 (changes requested):
I definitely didn't wire anything up for Kodi, thanks for the fixes

markus101 (approved):

PR #6593, opened 2024-03-03
Fix import list exclusion props

CheAle14:
#### Description

Fixes a problem with the import list exclusion props declaring `useCurrentPage` and `totalRecords` but not actually being passed them. This removes those props and makes it so `totalRecords` is selected from the app state, and `useCurrentPage` is calculated from the `useHistory()` hook.


markus101:
I could see that behaviour being annoying too and we don't do it on other tables.

PR #6599, opened 2024-03-05
Apple WebApp

Helvio88:
New iPhones need different meta tags to see an app as a web-app. This should address those missing tags on index.ejs, namely Status Bar.


Helvio88:
Both changes performed here.

markus101 (approved):

PR #6603, opened 2024-03-06
Porla Client

### LINKED ISSUE #5682: Add support for Porla download client

I wanted to replace my current download client with another one and investigated which download clients are available for Sonarr. Sonarr lists Hadouken as one of the download clients, however navigating to [its Github page](https://github.com/hadouken/hadouken) states that it is no longer being worked on and that you should use [Porla](https://github.com/porla/porla) instead.

I would like to see support for the Porla download client

I tried using qBittorrent but it seems to bug out after a while pausing all new downloads hence why I want to switch download clients.

### Anything else?

_No response_

Wesselinator:
#### Description

Attempts to add porla client for porla support.

No UI changes.

#### Database Migration
NO

* Closes #5682


Wesselinator:
Still working on this.
Still need to cleanup, and live test it for a bit, and docs.
At this point it it doesn't blow up, and you can adding works.

More then that even I am unsure.

Would like some guidance on the `GetStatus()` method.

src/NzbDrone.Common/Http/JsonRpcRequestBuilder.cs
Wesselinator, line 1: Needed to add the capability to send a object, instead of a array.
Problem is **keeping** backwards compatibility (Unsure what else relies on the assumption it'll do array)

Figured out something to keep the array format and send an object.

Not hard tested, but it deals with un equal arrays (it'll drop the last term)
Should also be ok if you send something that isn't a string as the key: `.ToString()` should deal with it.

Wesselinator: Unknown performance impact on the Tuple create, but I've never had an opertunity to use one!
And I think it is the mechanism that "Drops the last unequal term".

src/NzbDrone.Core/Download/Clients/Porla/Porla.cs
```diff
@@ -0,0 +1,239 @@
...
+            // var sessettings = _proxy.GetSessionSettings(Settings);
+
+            var status = new DownloadClientInfo
+            {
+                IsLocalhost = Settings.Host == "127.0.0.1" || Settings.Host == "localhost",
+                RemovesCompletedDownloads = false, // should be determined through Flags From  somewhere? I think session?
+            };
+
+            // the default directory value is stored in the presets values... i think?
+            // but will not always be active if not set
```
markus101, line 128: If not set what happens is Sonarr doesn't send a path to download to?

src/NzbDrone.Core/Download/Clients/Porla/Porla.cs
```diff
@@ -0,0 +1,239 @@
...
+
+        public override DownloadClientInfo GetStatus()
+        {
+            // var preset = _proxy.ListPresets(Settings);
+            // var sessettings = _proxy.GetSessionSettings(Settings);
+
+            var status = new DownloadClientInfo
+            {
+                IsLocalhost = Settings.Host == "127.0.0.1" || Settings.Host == "localhost",
+                RemovesCompletedDownloads = false, // should be determined through Flags From  somewhere? I think session?
```
markus101, line 124: This is based on settings configured in Porla. For example qbit, a user can configure qbit to either pause or remove completed downloads, if it's pause that will work with Sonarr, if it's remove it can be removed before Sonarr can import and we want to warn the user.

src/NzbDrone.Core/Download/Clients/Porla/Porla.cs
```diff
@@ -0,0 +1,239 @@
...
+            try
+            {
+                var sysVers = _proxy.GetSysVersion(Settings);
+                var version = new Version(sysVers.Porla.Version);
+
+                if (version < new Version("0.37.0"))
+                {
+                    // return new ValidationFailure(string.Empty, _localizationService.GetLocalizedString("DownloadClientValidationErrorVersion",
+                    //        new Dictionary<string, object> { { "clientName", Name }, { "requiredVersion", "0.37.0" }, { "reportedVersion", version } }));
+                    _logger.Warn($"Porla: Unsure if your version is forwards compatible: {version}");
```
markus101, line 186: Not sure why this is needed if you're also checking for < 1.0

src/NzbDrone.Core/Download/Clients/Porla/Porla.cs
```diff
@@ -0,0 +1,239 @@
...
+                }
+
+                if (version > new Version("0.37.0"))
+                {
+                    _logger.Warn($"Porla: Unsure if your version is backwards compatible: {version}");
```
markus101, line 191: If it's under active development then I don't think we should be super strict on the version.

src/NzbDrone.Core/Download/Clients/Porla/README.md
```diff
@@ -0,0 +1,63 @@
+# Porla Client
```
markus101, line 1: Sections of this would be better as comments in the code, other pieces could be dropped.

src/NzbDrone.Core/Download/Clients/Porla/README.md
```diff
@@ -0,0 +1,63 @@
...
+### Presets
+
+Presets are being set on the torrent adds
+
+So I needed to do something with the `save_path`.  
+I wanted this to work "by default", so if you have NO presets set, you **NEED** to set the `save_path` _BUT_ if you want to use the presets values you shoule NOT set the `save_path` in the RPC request. That is why you see that intersting branch on the `AddTorrent` functions.
+
+### Tags
+
+Would be great to use to filter BIG lists of torrents even more! But it doesn't look like Sonarr expects that. looks like it expects everything
```
markus101, line 47: Sonarr used tags/categories/labels in the download client to reduce the number of items it gets from the client (or if not filtered on the client, filtered while processing all results), they aren't used to filter by series or anything more specific.

markus101 (commented):
Seems unlikely that you'll need a migration, so `No` is the answer there,

src/NzbDrone.Core/Download/Clients/Porla/Porla.cs
Wesselinator: If we don't set a path, it doesn't want to accept it.
If the preset is setup on porla side, it's cool with it.

src/NzbDrone.Core/Download/Clients/Porla/Porla.cs
Wesselinator: Ok.
I'll dig through the docs again. I think you can set up that behavior. But IIRC, I think you can set that behavior on a per-trorrent bassis

just double checking: this code needs to be able to tell what is set up by checking in with porla correct?

src/NzbDrone.Core/Download/Clients/Porla/Porla.cs
Wesselinator: Checking forward and backward and the exact version we know is fine.
Maybe the future breaks, maybe we are incompatible with the past.
Wanted to get the blocks in so it's easy to change the numbers when that happens.

src/NzbDrone.Core/Download/Clients/Porla/Porla.cs
Wesselinator: I am currently not throwing any exception, regardless of the version porla tells us, but I am logging when the version doesn't match my expectation, so when something breaks I know by looking at the logs that the version is different.

I am running the live test on `latest` so I can catch new issues as and when the

src/NzbDrone.Core/Download/Clients/Porla/README.md
Wesselinator: But to get that list Sonarr asks for everything from the client.
It's not "On-Demand" correct?

So I we have say 1000 active torrents in the client, everything is requested, and only the ones that make sense to sonarr are loaded into the db, and THEN filtered because Sonarr can do the Series/Episode/Category mapping inside itself.

We have the capability to do optimized queries to porla: "Give us all the torrents with the tag `TVDBID=123456789` in the catagory `sonarr-tv` please.".

Of course in, practice that isn't too big of a concern really.
I still added the tags because I would very much like to filter the items outside of the Sonarr's context (dealing with said 1000 actives).

src/NzbDrone.Core/Download/Clients/Porla/README.md
Wesselinator: I'll figure out what to do with this at the end.

src/NzbDrone.Core/Download/Clients/Porla/Porla.cs
Wesselinator: I'll update this block when I get to it.

But yeah, if you have it set in the presets, it'll use that, else it expects you to set it in the request for the torrent.
If it is set in the presets, AND you set it in the torrent, it uses the request's value.

src/NzbDrone.Core/Download/Clients/Porla/Porla.cs
Wesselinator: Oh the `< 1.0` check!

That is just a general warning in the logs for others.
The checks above it are doing the actual comparability stuff

src/NzbDrone.Core/Download/Clients/Porla/Porla.cs
Wesselinator: Oh the < 1.0 check!

That is just a general warning in the logs for others.
The checks above it are doing the actual comparability stuff

Wesselinator:
Added tests. 
Struggling with actually running them, so some of them are not passing yet.

Wesselinator:
Finally figured out Tests!
Proper testing underway now

Wesselinator:
Ok need to do some cleanup but it is working now with no-exceptions as far I can tell

src/NzbDrone.Common/Http/JsonRpcRequestBuilder.cs
```diff
@@ -28,6 +31,16 @@ public JsonRpcRequestBuilder(string baseUrl, string method, IEnumerable<object>
             Method = HttpMethod.Post;
             JsonMethod = method;
             JsonParameters = parameters.ToList();
+            JsonParametersToObject = false;
+        }
+
+        public JsonRpcRequestBuilder(string baseUrl, string method, bool paramToObj, IEnumerable<object> parameters)
```
markus101, line 37: `paramToObj` -> `convertParamsToObject` or something less shorthand

src/NzbDrone.Core.Test/Download/DownloadClientTests/PorlaTests/PorlaFixture.cs
```diff
@@ -0,0 +1,595 @@
...
+            Mocker.GetMock<ITorrentFileInfoReader>()
+                  .Setup(s => s.GetHashFromTorrentFile(It.IsAny<byte[]>()))
+                  .Returns(_somehash);
+
+            Mocker.GetMock<IHttpClient>()
+                  .Setup(s => s.Get(It.IsAny<HttpRequest>()))
+                  .Returns<HttpRequest>(r => new HttpResponse(r, new HttpHeader(), Array.Empty<byte>()));
+        }
+
+        /// <summary> Setups the `Add*` to fail, by throwing Exceptions </summary>
```
markus101, line 225: No need to add summary comments to these methods

src/NzbDrone.Core/Download/Clients/LibTorrent/Models/LibTorrentInfoHash.cs
```diff
@@ -0,0 +1,68 @@
...
+using System.Collections.Generic;
+using Newtonsoft.Json;
+
+namespace NzbDrone.Core.Download.Clients.LibTorrent.Models
+{
+    // TODO: Figure out correct json serilization of `null` values, currently setting them to ""
+    public class LibTorrentInfoHash
+    {
+        #nullable enable
+        public string? Hash { get; set; }
```
markus101, line 11: Strings are nullable by default.

src/NzbDrone.Core.Test/Download/DownloadClientTests/PorlaTests/PorlaFixture.cs
```diff
@@ -0,0 +1,595 @@
...
+        {
+            Mocker.GetMock<IPorlaProxy>()
+                .Setup(s => s.AddMagnetTorrent(It.IsAny<PorlaSettings>(), It.IsAny<string>(), It.IsAny<IList<string>>()))
+                .Throws<InvalidOperationException>();
+
+            Mocker.GetMock<IPorlaProxy>()
+                .Setup(s => s.AddTorrentFile(It.IsAny<PorlaSettings>(), It.IsAny<byte[]>(), It.IsAny<IList<string>>()))
+                .Throws<InvalidOperationException>();
+
+            #nullable enable
```
markus101, line 236: These should not be necessary.

src/NzbDrone.Core.Test/Download/DownloadClientTests/PorlaTests/PorlaFixture.cs
```diff
@@ -0,0 +1,595 @@
...
+            Mocker.GetMock<IPorlaProxy>()
+                .Setup(s => s.AddMagnetTorrent(It.IsAny<PorlaSettings>(), It.IsAny<string>(), It.IsAny<IList<string>?>()))
+                .Throws<InvalidOperationException>();
+
+            Mocker.GetMock<IPorlaProxy>()
+                .Setup(s => s.AddTorrentFile(It.IsAny<PorlaSettings>(), It.IsAny<byte[]>(), It.IsAny<IList<string>?>()))
+                .Throws<InvalidOperationException>();
+            #nullable disable
```
markus101, line 244: Same for the disabling

src/NzbDrone.Core/Download/Clients/LibTorrent/Models/LibTorrentInfoHash.cs
```diff
@@ -0,0 +1,68 @@
...
+            Status = list[1];
+        }
+
+        public ICollection<string> ToList()
+        {
+            var hash = string.IsNullOrEmpty(Hash) ? "" : Hash;
+            var status = string.IsNullOrEmpty(Status) ? "" : Status;
+            string[] ret = { hash, status };
+            return ret;
+        }
```
markus101, line 37: This is easier to read and less lines.

```suggestion
            var hash = Hash ?? string.Empty;
            var status = Status ?? string.Empty;

            return new[] { hash, status };
        }
```

src/NzbDrone.Core/Download/Clients/Porla/Models/PorlaPresetsList.cs
```diff
@@ -0,0 +1,86 @@
...
+{
+    /// <summary> Porla Session Extention Methods </summary>
+    public static class PorlaPresetsExtentions
+    {
+        /// <summary> Gets the spesified preset values merged with the values from the default preset </summary>
+        public static PorlaPreset GetEffective(this ReadOnlyDictionary<string, PorlaPreset> presets, string preset)
+        {
+            if (presets == null)
+            {
+                // presets is null
```
markus101, line 15: Unnecessary comment.

src/NzbDrone.Core/Download/Clients/Porla/Models/PorlaTorrent.cs
```diff
@@ -0,0 +1,36 @@
...
+        [JsonConverter(typeof(LibTorrentInfoHashConverter))]
+        public LibTorrentInfoHash InfoHash { get; set; }
+
+        [JsonConstructor]
+        public PorlaTorrent(string hash, string status)
+        {
+            InfoHash = new LibTorrentInfoHash(hash, status);
+        }
+
+        public PorlaTorrent(LibTorrentInfoHash ltif)
```
markus101, line 19: Use words instead of acronyms

```suggestion
        public PorlaTorrent(LibTorrentInfoHash infoHash)
```

src/NzbDrone.Core/Download/Clients/Porla/Models/PorlaTorrent.cs
```diff
@@ -0,0 +1,36 @@
...
+        {
+            InfoHash = ltif;
+        }
+
+        public object[] AsParam()
+        {
+            string[] ret = { InfoHash.Hash, null };
+            return ret;
```
markus101, line 27:
```suggestion
            return new[] { InfoHash.Hash, null };
```

src/NzbDrone.Core/Download/Clients/Porla/Models/PorlaTorrent.cs
```diff
@@ -0,0 +1,36 @@
...
+        }
+
+        public object[] AsParams()
+        {
+            object[] ret = { "info_hash", AsParam() };
```
markus101, line 32: Same thing here

src/NzbDrone.Core/Download/Clients/Porla/Porla.cs
```diff
@@ -0,0 +1,281 @@
...
+        public override IEnumerable<DownloadClientItem> GetItems()
+        {
+            var plist = _proxy.ListTorrents(Settings);
+
+            var items = new List<DownloadClientItem>();
+
+            // should probs paginate instead of cheating
+            foreach (var torrent in plist)
+            {
+                // we don't need to check the category, the filter did that for us, but we are checking anyway becuase why not :)
```
markus101, line 49: If we don't need to because it's already filtered, then we don't need to filter again

src/NzbDrone.Core/Download/Clients/Porla/Porla.cs
```diff
@@ -0,0 +1,281 @@
...
+                    DownloadId = torrent.InfoHash.Hash,
+                    OutputPath = outputPath + torrent.Name,
+                    RemainingSize = torrent.Total,
+                    RemainingTime = torrent.ETA < 1 ? (TimeSpan?)null : TimeSpan.FromSeconds((double)torrent.ETA), // LibTorent uses `-1` to denote "infinite" time i.e. I am stuck, or I am done. FromSeconds will convert `-1` to 1, so we need to do this.
+                    Title = torrent.Name,
+                    TotalSize = torrent.Size,
+                    SeedRatio = torrent.Ratio
+                };
+
+                // deal with moving_storage=true ?
```
markus101, line 71: Probably treat it as downloading

src/NzbDrone.Core/Download/Clients/Porla/Porla.cs
```diff
@@ -0,0 +1,281 @@
...
+                    if (torrent.ETA < 1)
+                    {
+                        // Sonarr wants to see a TimeSpan.Zero when it is done ( -1 -> 0 )
+                        item.RemainingTime = TimeSpan.Zero;
+                    }
+                }
+                else if (torrent.Flags.Contains("paused"))
+                {
+                   item.Status = DownloadItemStatus.Paused;
+                } /*  I don't know what the torent looks like if it is "Queued"
```
markus101, line 90: Does Porla provide a list of flags we can use to determine that?

src/NzbDrone.Core/Download/Clients/Porla/Porla.cs
```diff
@@ -0,0 +1,281 @@
...
+            return items;
+        }
+
+        public override void RemoveItem(DownloadClientItem item, bool deleteData)
+        {
+            // Kinda sucks we don't have a `RemoveItems`, porla has a batch interface for removals
+            PorlaTorrent[] singleItem = { new PorlaTorrent(item.DownloadId, "") };
+            _proxy.RemoveTorrent(Settings, deleteData, singleItem);
+
+            // when do we set item.Removed ?
```
markus101, line 119: No need to, when it's dropped from the client it'll be removed from Sonarr's queue.

src/NzbDrone.Core/Download/Clients/Porla/Porla.cs
```diff
@@ -0,0 +1,281 @@
...
+
+            var status = new DownloadClientInfo
+            {
+                IsLocalhost = Settings.Host == "127.0.0.1" || Settings.Host == "localhost",
+                RemovesCompletedDownloads = false, // TODO: I don't think porla has config for this, it feels like it should
+            };
+
+            var savePath = ((presetEffectiveSettings?.SavePath.IsNullOrWhiteSpace() ?? true) ? Settings.TvDirectory : presetEffectiveSettings?.SavePath) ?? "";
+            var destDir = new OsPath(savePath);
+            if (!destDir.IsEmpty)
```
markus101, line 136: new line before blocks

```suggestion

            if (!destDir.IsEmpty)
```

src/NzbDrone.Core/Download/Clients/Porla/Porla.cs
```diff
@@ -0,0 +1,281 @@
...
+            return tags;
+        }
+
+        // NOTE: If the torrent already exists in the client, it'll fail with error code -3
+        // {
+        //   "code": -3,
+        //   "data": null,
+        //   "message": "Torrent already in session 'default'"
+        // }
+        // IDK If I should deal with that...
```
markus101, line 166: If it already exists then we either need to throw an error (why isn't Sonarr tracking it?) or treat it as successful, pros and cons to both.

src/NzbDrone.Core/Download/Clients/Porla/Porla.cs
```diff
@@ -0,0 +1,281 @@
...
+
+        protected override string AddFromMagnetLink(RemoteEpisode remoteEpisode, string hash, string magnetLink)
+        {
+            PorlaTorrent torrent;
+            if (Settings.SeriesTag)
```
markus101, line 171: New line here too

src/NzbDrone.Core/Download/Clients/Porla/Porla.cs
```diff
@@ -0,0 +1,281 @@
...
+                torrent = _proxy.AddMagnetTorrent(Settings, magnetLink);
+            }
+
+            return torrent.InfoHash.Hash ?? "";
+        }
+
+        protected override string AddFromTorrentFile(RemoteEpisode remoteEpisode, string hash, string filename, byte[] fileContent)
+        {
+            PorlaTorrent torrent;
+            if (Settings.SeriesTag)
```
markus101, line 187: And here

src/NzbDrone.Core/Download/Clients/Porla/Porla.cs
```diff
@@ -0,0 +1,281 @@
...
+                torrent = _proxy.AddTorrentFile(Settings, fileContent);
+            }
+
+            return torrent.InfoHash.Hash ?? "";
+        }
+
+        protected override void Test(List<ValidationFailure> failures)
+        {
+            failures.AddIfNotNull(TestVersion());
+            if (failures.HasErrors())
```
markus101, line 203: And here

src/NzbDrone.Core/Download/Clients/Porla/PorlaProxy.cs
```diff
@@ -0,0 +1,234 @@
...
+        }
+
+        private T ProcessRequest<T>(PorlaSettings settings, string method, params object?[] parameters)
+        {
+            var baseUrl = HttpRequestBuilder.BuildBaseUrl(settings.UseSsl, settings.Host, settings.Port, settings.UrlBase);
+
+            var jwt = settings.InfinteJWT ??= string.Empty;
+            var apiurl = settings.ApiUrl ??= string.Empty;
+
+            // this block will run a lot, don't want to be too noisy in the logs
```
markus101, line 80: Do we need it at all? An empty if and else won't do anything.

src/NzbDrone.Core/Download/Clients/Porla/PorlaProxy.cs
```diff
@@ -0,0 +1,234 @@
...
+
+        private void LogSupposedToBeNothing(string method, string something)
+        {
+            if (!string.IsNullOrEmpty(something))
+            {
+                _logger.Warn($"method: {method} was not expected to return: {something}");
+            }
+        }
+
+        // sys
```
markus101, line 146: These comments don't provide much, if anything

markus101 (changes requested):
Just a quick pass, but left some comments/questions.

Why all the `#nullable enable` lines?

Still some cleanup of code that is commented out and some low value comments around. Some methods have summary comments, but generally not something we include.

src/NzbDrone.Core/Download/Clients/LibTorrent/Models/LibTorrentInfoHash.cs
Wesselinator: Yeah I need to rethink this class.
I don't have anything to test against so I don 't know how the distinction between a proper c++ `null` vs a C# _string null_ will behave.

I'll keep it C# for now and then someone can figure it out later when it's relevant

src/NzbDrone.Core.Test/Download/DownloadClientTests/PorlaTests/PorlaFixture.cs
Wesselinator: You would think so!

I think there there is something wrong between the JSONRPC decoding json nulls to C# objects that I should check somewhere. 

I get extremely confusing errors if I don't include the nullable flag

src/NzbDrone.Core/Download/Clients/Porla/Porla.cs
Wesselinator: Yeah, at that point we are testing if porla works.
I'll remove it

src/NzbDrone.Core/Download/Clients/Porla/Porla.cs
Wesselinator: Using porla for a bit, I actually got some examples, I'll be able to fill these in.

No. Yes? Kinda.
Libtorrent states do not match 1:1 with Sonarr states.
It's generally more "programy" and less "usery" if that makes sense.
I needed to "interpret" the logical state of a torrent by using stuff like it's time active, if it downloaded anything, etc.

src/NzbDrone.Core/Download/Clients/Porla/Porla.cs
Wesselinator: How did that slip in!
C# won't build of you screw up the formatting!

src/NzbDrone.Core/Download/Clients/Porla/PorlaProxy.cs
Wesselinator: I guess the idea is for the future when they do implement JWT's we need to deal with it here.

What I actually want to do is just log what mode of JWT is currently being used.

Am i allowed to log messages in the ClientSettings.cs ?

vktr:
Hey @Wesselinator - awesome stuff! 😸 Do you need any help from our (Porla) end for this?

arty01238:
What's the status on this as I am looking to use this download client?

vktr:
I'm more than happy to help finish this PR - just let me know and I'll get started!

onedr0p:
@vktr it looks like @Wesselinator hasn't been active on Github for several months, I see no reason why you couldn't take this over 👍🏼

markus101:
Closing as this has been sitting stale for some time, but if someone wants to pick up and complete it please do and open up another PR.

PR #6606, opened 2024-03-07
Prevent NullRef in naming when truncating a null Release Group

mynameisbogdan:
#### Description
Fix for https://discord.com/channels/383686866005917708/383688189941907459/1215276558424743997 when using truncate with a null value.

### DIFF
--- a/src/NzbDrone.Core/Organizer/FileNameBuilder.cs
+++ b/src/NzbDrone.Core/Organizer/FileNameBuilder.cs
@@ -1168,6 +1168,11 @@ namespace NzbDrone.Core.Organizer
 
         private string Truncate(string input, string formatter)
         {
+            if (input.IsNullOrWhiteSpace())
+            {
+                return null;
+            }
+
             var maxLength = GetMaxLengthFromFormatter(formatter);
 
             if (maxLength == 0 || input.Length <= Math.Abs(maxLength))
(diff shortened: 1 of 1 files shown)


src/NzbDrone.Core/Organizer/FileNameBuilder.cs
```diff
@@ -1168,6 +1168,11 @@ private static string CleanFileName(string name, NamingConfig namingConfig)
 
         private string Truncate(string input, string formatter)
         {
+            if (input.IsNullOrWhiteSpace())
+            {
+                return null;
```
markus101, line 1173:
```suggestion
                return string.Empty;
```

src/NzbDrone.Core/Organizer/FileNameBuilder.cs
```diff
@@ -1168,6 +1168,11 @@ private static string CleanFileName(string name, NamingConfig namingConfig)
 
         private string Truncate(string input, string formatter)
         {
+            if (input.IsNullOrWhiteSpace())
+            {
+                return string.Empty;
```
markus101, line 1173: I realized for release groups we want this to be null so we fall back to Sonarr, but not sure we should return null for other cases and an empty string may be better.

This will handle the release group scenario without forcing everything to null.

```csharp
tokenHandlers["{Release Group}"] = m => episodeFile.ReleaseGroup.IsNullOrWhiteSpace() ? m.DefaultValue("Sonarr") : Truncate(episodeFile.ReleaseGroup, m.CustomFormat);
```

markus101:
Landed via #6608

Issue #5342, opened 2023-01-03
Add IMDB Link to Telegram Notifications


Zantag:
### Is there an existing issue for this?

- [X] I have searched the existing open and closed issues

### Is your feature request related to a problem? Please describe

I have telegram group with telegram bot connected to sonarr, radarr and bazarr. 
The notifications from radarr has imdb link... notifications from sonarr doesn't... in the example below you will see how bad look like the notification from sonarr without preview from the imdb link...
![image](https://user-images.githubusercontent.com/1080384/210436469-b49856a6-274e-42a8-b1bf-4c0ccb50b224.png)

### Describe the solution you'd like

I want just to add imdb link for the tv series in all notification connected to telegram bots... the same way used in radarr.

### Describe alternatives you've considered

Hope this will be not hard to be integrated in the notifications feature...

markus101:
I imagine Radarr always has IMDB IDs, but for Sonarr that's not always the case, it'd be more reliable to have TheTVDB links instead.

Zantag:
> I imagine Radarr always has IMDB IDs, but for Sonarr that's not always the case, it'd be more reliable to have TheTVDB links instead.
 Every TV series on my sonarr has IMDB link... The links from IMDB also has nice preview... So i prefer IMDB.
 
 
 
![image](https://user-images.githubusercontent.com/1080384/210616405-f1d42358-d65d-420f-88ad-452c64b07fa5.png)

markus101:
Making it optional and giving a choice should be easy enough whenever someone picks this up.

ricci2511:
I recently thought about this too.
I guess we can add the imdb url in the `GetMessage` method in `NotificationService.cs` just like Radarr does. But maybe you have something else in mind? @markus101

Adair-GA:
I can do this, it does't seem to be really hard, but I'm a bit confused as how would I make it an optional feture. It seems that it could either be an option for the telegram connection, or it could be included in the `GetMessage` method, but that would make it appear on all connections, and I don't see how could I make that optional.

markus101:
I don't think we want to add it to every message by adding to to `NotificationService`.

The simplest solution would be to add an option to `TelegramSettings`, then conditionally append it to the message before sending, which can be done with a function in `Telegram.cs` that takes the message input and outputs the updated message.

qTipTip:
[...]

PR #6613, opened 2024-03-09
[Feature] Add clickable metadata links when using the Telegram connector. (IMDb / Trakt / TVDb, TVMaze)

qTipTip:
#### Description

Adds a checkbox and a dropdown allowing you to select whether to add a clickable link to your Telegram notification, and which database you want a link to. 

*Questions*:

1. Which events should this be added to. Currently only implemented for series added and deleted.
2. Should it be added for episode-specific events, like episode grabbed?
3. Should we only use the dropdown and have a `None` or `Disabled`-option instead? Probably easier to reason about for the user. 

I'd appreciate some input on wording. The setting-labels are not very enlightening at the moment. 

Screenshot 2024-03-09 135542
Screenshot 2024-03-09 135512
Screenshot 2024-03-09 001358

* Closes https://github.com/Sonarr/Sonarr/issues/5342


src/NzbDrone.Api.Test/Sonarr.Api.Test.csproj
```diff
@@ -1,6 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <TargetFrameworks>net6.0</TargetFrameworks>
+    <Platforms>AnyCPU;x86</Platforms>
```
qTipTip, line 4: I'll drop the changes in the `.csproj`-files. Not sure how I managed to commit them.

src/NzbDrone.Core/Notifications/Telegram/Telegram.cs
```diff
@@ -1,16 +1,78 @@
...
     {
+        private const string ImdbUrlFormat = "https://www.imdb.com/title/{0}";
+        private const string TvdbUrlFormat = "http://www.thetvdb.com/?tab=series&id={0}";
+        private const string TraktUrlFormat = "http://trakt.tv/search/tvdb/{0}?id_type=show";
+        private const string TvMazeUrlFormat = "http://www.tvmaze.com/shows/{0}/_";
+
         private readonly ITelegramProxy _proxy;
+        private readonly Dictionary<MetadataLinkType, Func<string, string, string>> _formatLinkFromIdMethods;
+
+        private string FormatImdbLinkFromId(string message, string id)
```
qTipTip, line 19: Could this be encapsulated in a utility class somehow? Might be useful for other notication providers?

src/NzbDrone.Core/Notifications/Telegram/TelegramSettings.cs
```diff
@@ -32,6 +40,12 @@ public class TelegramSettings : IProviderConfig
         [FieldDefinition(3, Label = "NotificationsTelegramSettingsSendSilently", Type = FieldType.Checkbox, HelpText = "NotificationsTelegramSettingsSendSilentlyHelpText")]
         public bool SendSilently { get; set; }
 
+        [DefaultValue(true)]
+        [FieldDefinition(4, Label = "NotificationsTelegramSettingsSendMetadataLink", Type = FieldType.Checkbox, HelpText = "NotificationsTelegramSettingsSendMetadataLink")]
```
stevietv, line 44: For the Label and Helptext to correctly translate, please add the key and value to the `src/NzbDrone.Core/Localization/Core/en.json` file

stevietv (commented):
nice first contribution :)
I was wondering why the choice for 4 different methods for formatting the string? Couldnt it be done in a single method with a case switch if you pass in the MedatdataLinkType?

How have you handled cases where there isn't an imdb, trakt or tvmaze Id associated with the show, but user has selected on?

qTipTip:
> nice first contribution :) I was wondering why the choice for 4 different methods for formatting the string? Couldnt it be done in a single method with a case switch if you pass in the MedatdataLinkType?
> 
> How have you handled cases where there isn't an imdb, trakt or tvmaze Id associated with the show, but user has selected on?

Thanks for the input! 

I'll try to create a switch. 

And no, I haven't considered that fact! 
Would that materialize as a null-value for the ID? If so I can make adding the link a no-op in that case.

src/NzbDrone.Core/Notifications/Telegram/Telegram.cs
markus101: Potentially, but it depends a lot on the format something else might need, so for now I wouldn't worry about it.

This whole group of methods can be cleaned up and combined with `FormatMessageWithLink`

Something like:

```
private string FormatMessageWithLink(string message, Series series)
{
	var linkType = (MetadataLinkType)Settings.MetadataLinkType;

	switch (linkType)
    {
        case MetadataLinkType.IMDb:
            return $"[{message}](https://www.imdb.com/title/{series.ImdbId})";
        ...
        default:
            throw new ArgumentException($"Unsupported link type: {linkType}", nameof(linkType));
    }
}

Comment (markus101) [reply src/NzbDrone.Core/Notifications/Telegram/Telegram.cs:19]: Also move these private functions to the bottom of the class.

File: src/NzbDrone.Core/Notifications/Telegram/Telegram.cs [hunk 4]
```diff
@@ -1,16 +1,78 @@
...
+                { MetadataLinkType.IMDb, FormatImdbLinkFromId },
+                { MetadataLinkType.TVDb, FormatTvdbLinkFromId },
+                { MetadataLinkType.Trakt, FormatTraktLinkFromId },
+                { MetadataLinkType.TVMaze, FormatTVMazeLinkFromId }
+            };
+        }
+
+        private string FormatMessageWithLink(string message, Series series)
+        {
+            if (Settings.SendMetadataLink && _formatLinkFromIdMethods.TryGetValue(Settings.MetadataLinkType, out var formatMethod))
```
markus101, line 70: Settings should validate that the stored `MetadataLinkType` is valid.

src/NzbDrone.Core/Notifications/Telegram/TelegramSettings.cs
```diff
@@ -1,10 +1,18 @@
+using System.ComponentModel;
 using FluentValidation;
 using NzbDrone.Core.Annotations;
 using NzbDrone.Core.ThingiProvider;
 using NzbDrone.Core.Validation;
-
 namespace NzbDrone.Core.Notifications.Telegram
 {
+    public enum MetadataLinkType
```
markus101, line 8: Move below the class

src/NzbDrone.Core/Notifications/Telegram/TelegramSettings.cs
markus101: No need for this setting, it can all be done with the enum.

src/NzbDrone.Core/Notifications/Telegram/TelegramSettings.cs
```diff
@@ -1,10 +1,18 @@
...
+    {
+        IMDb,
+        TVDb,
+        TVMaze,
+        Trakt,
```
markus101, line 13: `Imdb` instead of a stylized format, use the `FieldOption` attribute to give it a formatted name, each value in the enum should have a integer assigned to it for future safetyy.

Take a look at `ReleaseType` to see how it's done there.

src/NzbDrone.Core/Notifications/Telegram/TelegramSettings.cs
markus101: Add `None =0` as the first enum value to allow the single enum to control whether it's sent or not.

src/NzbDrone.Core/Notifications/Telegram/TelegramSettings.cs
```diff
@@ -32,6 +40,12 @@ public class TelegramSettings : IProviderConfig
         [FieldDefinition(3, Label = "NotificationsTelegramSettingsSendSilently", Type = FieldType.Checkbox, HelpText = "NotificationsTelegramSettingsSendSilentlyHelpText")]
         public bool SendSilently { get; set; }
 
+        [DefaultValue(true)]
+        [FieldDefinition(4, Label = "NotificationsTelegramSettingsSendMetadataLink", Type = FieldType.Checkbox, HelpText = "NotificationsTelegramSettingsSendMetadataLink")]
+        public bool SendMetadataLink { get; set; }
+        [FieldDefinition(5, Label = "NotificationsTelegramSettingsMetadataLinkType", Type = FieldType.Select, SelectOptions = typeof(MetadataLinkType), HelpText = "NotificationsTelegramSettingsMetadataLinkType")]
```
markus101, line 46: `TelegramSettingsValidator` needs a validation rule to ensure the value of the enum is valid.

An example how that's done for `ReleaseType`:
```
RuleFor(c => c.Value).Custom((releaseType, context) =>
{
    if (!Enum.IsDefined(typeof(ReleaseType), releaseType))
    {
        context.AddFailure($"Invalid release type condition value: {releaseType}");
    }
});
```

markus101 (changes requested):
Thanks for contributing! Left quite a few comments which should help clean things up a lot and reduce the number of changes to make it easier to follow.

markus101:
> Would that materialize as a null-value for the ID? If so I can make adding the link a no-op in that case.

We'll have a TVDB ID 100% of the time, IMDB should be null (off chance of an empty string) and everything else will be 0 if not set.

src/NzbDrone.Core/Notifications/Telegram/TelegramProxy.cs
```diff
@@ -35,13 +35,13 @@ public TelegramProxy(IHttpClient httpClient, ILocalizationService localizationSe
...
             // Format text to add the title before and bold using markdown
-            var text = $"<b>{HttpUtility.HtmlEncode(title)}</b>\n{HttpUtility.HtmlEncode(message)}";
+            var text = $"*{HttpUtility.HtmlEncode(title)}*\n{HttpUtility.HtmlEncode(message)}";
 
             var requestBuilder = new HttpRequestBuilder(URL).Resource("bot{token}/sendmessage").Post();
 
             var request = requestBuilder.SetSegment("token", settings.BotToken)
                                         .AddFormParameter("chat_id", settings.ChatId)
-                                        .AddFormParameter("parse_mode", "HTML")
+                                        .AddFormParameter("parse_mode", "MarkdownV2")
```
qTipTip, line 44: I changed the parser to `MarkdownV2` to have the link render properly, as I think the HTML `hrefs` are being escaped in `HTML`-mode when just added to the message. However, this breaks the `SeriesDeleteMessage` as it contains a reserved character `-`. 

Not quite sure how to handle this.

src/NzbDrone.Core/Notifications/Telegram/TelegramSettings.cs
```diff
@@ -32,9 +41,25 @@ public class TelegramSettings : IProviderConfig
...
         public NzbDroneValidationResult Validate()
         {
             return new NzbDroneValidationResult(Validator.Validate(this));
         }
     }
+
+    public enum MetadataLinkType
+    {
+        None = 0,
+        [FieldOption(Label = "IMDb")]
```
qTipTip, line 57: Should I add `FieldOption` to `None` and `Trakt` as well, where the Enum key aligns with the Label?

qTipTip:
> Thanks for contributing! Left quite a few comments which should help clean things up a lot and reduce the number of changes to make it easier to follow.

Thanks for the thorough review @markus101 ! I've cleaned up stuff, and it's a lot more readable now, so thanks for that! 🔥 

How can I have `None` set as the default option in TelegramSettings.cs? Does that happen just by virtue of the `= 0` in the enum?

src/NzbDrone.Core/Notifications/Telegram/Telegram.cs
```diff
@@ -69,5 +72,37 @@ public override ValidationResult Test()
...
+        private string FormatMessageWithLink(string message, Series series)
+        {
+            var linkType =  Settings.MetadataLinkType;
+
+            if (linkType == MetadataLinkType.None)
+            {
+                return message;
+            }
+
+            if (linkType == MetadataLinkType.Imdb && !string.IsNullOrWhiteSpace(series.ImdbId))
```
markus101, line 85:
```suggestion
            if (linkType == MetadataLinkType.Imdb && series.ImdbId.IsNotNullOrWhiteSpace())
```

src/NzbDrone.Core/Notifications/Telegram/TelegramProxy.cs
```diff
@@ -35,13 +35,13 @@ public TelegramProxy(IHttpClient httpClient, ILocalizationService localizationSe
         public void SendNotification(string title, string message, TelegramSettings settings)
         {
             // Format text to add the title before and bold using markdown
-            var text = $"<b>{HttpUtility.HtmlEncode(title)}</b>\n{HttpUtility.HtmlEncode(message)}";
+            var text = $"*{HttpUtility.HtmlEncode(title)}*\n{HttpUtility.HtmlEncode(message)}";
```
markus101, line 38: If this isn't sending HTML, do we still need to encode it for HTML?

src/NzbDrone.Core/Notifications/Telegram/TelegramSettings.cs
```diff
@@ -32,9 +39,27 @@ public class TelegramSettings : IProviderConfig
         [FieldDefinition(3, Label = "NotificationsTelegramSettingsSendSilently", Type = FieldType.Checkbox, HelpText = "NotificationsTelegramSettingsSendSilentlyHelpText")]
         public bool SendSilently { get; set; }
 
+        public bool SendMetadataLink { get; set; }
```
markus101, line 42: No longer needed

src/NzbDrone.Core/Notifications/Telegram/TelegramSettings.cs
markus101: I think leaving them off as fine as long as the UI is rendering them correctly.

src/NzbDrone.Core/Notifications/Telegram/TelegramProxy.cs
markus101: Which part of the message? A dash could also be in the series. We could escape the `-`, possibly with just `\-`, but what about `*` and other special characters in markdown? `_` could cause issues, especially if there are two in the title.

src/NzbDrone.Core/Notifications/Telegram/Telegram.cs
```diff
@@ -33,12 +34,14 @@ public override void OnEpisodeFileDelete(EpisodeDeleteMessage deleteMessage)
 
         public override void OnSeriesAdd(SeriesAddMessage message)
         {
-            _proxy.SendNotification(SERIES_ADDED_TITLE, message.Message, Settings);
+            var text = FormatMessageWithLink(message.Message, message.Series);
+            _proxy.SendNotification(SERIES_ADDED_TITLE, text, Settings);
```
markus101, line 38:
```suggestion
            var text = FormatMessageWithLink(message.Message, message.Series);

            _proxy.SendNotification(SERIES_ADDED_TITLE, text, Settings);
```

src/NzbDrone.Core/Notifications/Telegram/Telegram.cs
```diff
@@ -33,12 +34,14 @@ public override void OnEpisodeFileDelete(EpisodeDeleteMessage deleteMessage)
...
         }
 
         public override void OnSeriesDelete(SeriesDeleteMessage deleteMessage)
         {
-            _proxy.SendNotification(SERIES_DELETED_TITLE, deleteMessage.Message, Settings);
+            var text = FormatMessageWithLink(deleteMessage.Message, deleteMessage.Series);
+            _proxy.SendNotification(SERIES_DELETED_TITLE, text, Settings);
```
markus101, line 44:
```suggestion
            var text = FormatMessageWithLink(deleteMessage.Message, deleteMessage.Series);

            _proxy.SendNotification(SERIES_DELETED_TITLE, text, Settings);
```

markus101 (changes requested):
Great work on the clean up, much easier to review now, left some more comments and questions.

src/NzbDrone.Core/Notifications/Telegram/TelegramProxy.cs
markus101: Still need to determine the impact of this

qTipTip: Yes, apologies for the lack of activity here. 

I'll try to mess around with reverting back to HTML. 
Been unable to find a clean solution. 

The problem I had initially with the HTML parser was that the markup would render as raw strings in the message, instead of formatting the message.

src/NzbDrone.Core/Notifications/Telegram/Telegram.cs
```diff
@@ -33,12 +34,16 @@ public override void OnEpisodeFileDelete(EpisodeDeleteMessage deleteMessage)
 
         public override void OnSeriesAdd(SeriesAddMessage message)
         {
-            _proxy.SendNotification(SERIES_ADDED_TITLE, message.Message, Settings);
+            var text = FormatMessageWithLink(message.Message, message.Series);
```
markus101, line 37: These links should also be included for `OnGrab`, `OnDownload` and `OnEpisodeFileDelete`

src/NzbDrone.Core/Notifications/Telegram/TelegramProxy.cs
markus101: This is because `HttpUtility.HtmlEncode` is being called on the body text. The message body with the link can't be passed into `SendNotification`, the link needs to be passed in separately and then built in the proxy, which is better because it hides that implementation detail from `Telegram.cs`

src/NzbDrone.Core/Notifications/Telegram/TelegramSettings.cs
markus101: This still needs to be removed.

src/NzbDrone.Core/Notifications/Telegram/TelegramSettings.cs
```diff
@@ -32,9 +39,27 @@ public class TelegramSettings : IProviderConfig
...
+        [FieldDefinition(4, Label = "NotificationsTelegramSettingsMetadataLinkType", Type = FieldType.Select, SelectOptions = typeof(MetadataLinkType), HelpText = "NotificationsTelegramSettingsMetadataLinkType")]
```
markus101, line 43:
```suggestion
        [FieldDefinition(4, Label = "NotificationsTelegramSettingsMetadataLink", Type = FieldType.Select, SelectOptions = typeof(MetadataLinkType), HelpText = "NotificationsTelegramSettingsMetadataLink")]
```

src/NzbDrone.Core/Notifications/Telegram/TelegramSettings.cs
```diff
@@ -32,9 +39,27 @@ public class TelegramSettings : IProviderConfig
...
+        public MetadataLinkType MetadataLinkType { get; set; }
```
markus101, line 44:
```suggestion
        public int MetadataLink { get; set; }
```

src/NzbDrone.Core/Notifications/Telegram/Telegram.cs
```diff
@@ -69,5 +74,37 @@ public override ValidationResult Test()
 
             return new ValidationResult(failures);
         }
+
+        private string FormatMessageWithLink(string message, Series series)
```
markus101, line 78: This should be renamed to `GetLink` and just return the link, which gets passed to `SendNotification` and the message is formatted there.

src/NzbDrone.Core/Notifications/Telegram/TelegramProxy.cs
markus101: For test notifications something like the following inside `Test`:

```
var link = settings.MetadataLink == (int)MetadataLinkType.None ? null : "https://sonarr.tv";

SendNotification(settings.IncludeAppNameInTitle ? brandedTitle : title, body, link, settings);
```

This would allow people to see that the link works.

markus101 (changes requested):
I've tested this locally and added some additional comments and suggestions on how to make this functional.

I do wonder if a slightly different approach would be better, where individual links are added for each selected Metadata type, instead of turning the body of the message into a single link.

Let me know what you think about that and whether you'll be able to pick up the rest of the fixes here.

qTipTip:
> I've tested this locally and added some additional comments and suggestions on how to make this functional.
> 
> I do wonder if a slightly different approach would be better, where individual links are added for each selected Metadata type, instead of turning the body of the message into a single link.
> 
> Let me know what you think about that and whether you'll be able to pick up the rest of the fixes here.

Hey! 

I'll have a look at this sometime this week!

markus101:
Superceded by https://github.com/Sonarr/Sonarr/pull/6613

PR #6614, opened 2024-03-10
Add notification for build success/failures

markus101:
#### Description
Notifications when develop/main builds succeed/fail.


.github/workflows/build.yml
```diff
@@ -225,3 +225,25 @@ jobs:
       branch: ${{ github.ref_name }}
       major_version: ${{ needs.backend.outputs.major_version }}
       version: ${{ needs.backend.outputs.version }}
+
+  notifiy:
```
qTipTip, line 229:
```suggestion
  notify:
```

PR #6618, opened 2024-03-10
Add include/exclude logic for `{Custom Formats}` renaming token

### LINKED ISSUE #6615: Add rename token: 'remaining custom formats'

The new "{Custom Format:Format}"  allows you to choose the location of a custom format within renaming. However, when used with "{Custom Formats}", it creates a duplicate. Without this new token, multiple "{Custom Format:Format}" entries would be necessary for other formats, such as the web source, which would make the renaming convention rather long.

I'd like a new "{Remaining Custom Formats}" that would use all remaining custom formats that are not already mentioned in a {Custom Formats:Format}.

For example, 

{Series TitleYear} - S{season:00}E{episode:00} - {Episode Title} {Custom Format:Repack} ({Quality Title} {Remaining Custom Formats} {Mediainfo VideoCodec} {MediaInfo VideoBitDepth}bit {MediaInfo VideoDynamicRangeType} {Mediainfo AudioCodec}{ Mediainfo AudioChannels} {Release Group})

Whereas {Remaining Custom Formats} would include values such as the source, e.g. AMZN.

Otherwise, we would have to do something similar to:

{Series TitleYear} - S{season:00}E{episode:00} - {Episode Title} {Custom Format:Repack} ({Quality Title} {Custom Format:AMZN}{Custom Format:HBO}{Custom Format:DSNP},etc... 

To use a "{Custom Format:Format}" for every format.

or

Allow {Custom Format:Format} to work on formats that have "Include Custom Format when Renaming" deselected so that we can specify formats that would not be included in "{Custom Formats}" to prevent duplicity.

### Anything else?

Nope

Alanoll:
#### Description
Adds include/exclude logic similar to the `{MediaInfo}` languages tokens.

The delimiter is `,` instead however because I thought `,` would be less used in Custom Format names than the `-`. e.g. `{Custom Formats:-Name1,Name2}` instead of `{Custom Formats:-WEB-DL}`

#### Database Migration
NO

* Closes #6615

### DIFF
--- a/src/NzbDrone.Core.Test/OrganizerTests/FileNameBuilderTests/CustomFormatsFixture.cs
+++ b/src/NzbDrone.Core.Test/OrganizerTests/FileNameBuilderTests/CustomFormatsFixture.cs
@@ -93,6 +93,30 @@ namespace NzbDrone.Core.Test.OrganizerTests.FileNameBuilderTests
                    .Should().Be(expected);
         }
 
+        [TestCase("{Custom Formats:-INTERNAL}", "AMZN NAME WITH SPACES")]
+        [TestCase("{Custom Formats:-NAME WITH SPACES}", "INTERNAL AMZN")]
+        [TestCase("{Custom Formats:-INTERNAL,NAME WITH SPACES}", "AMZN")]
+        [TestCase("{Custom Formats:INTERNAL}", "INTERNAL")]
+        [TestCase("{Custom Formats:NAME WITH SPACES}", "NAME WITH SPACES")]
+        [TestCase("{Custom Formats:INTERNAL,NAME WITH SPACES}", "INTERNAL NAME WITH SPACES")]
+        public void should_replace_custom_formats_with_filtered_names(string format, string expected)
+        {
+            _namingConfig.StandardEpisodeFormat = format;
+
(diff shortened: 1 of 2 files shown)


Alanoll:
With the addition of this logic, it almost obsoletes the recently added `{Custom Format:Name}` token, but I left it in because of Token ordering.

For Example:
 `{Custom Formats}` is ordered by `Name`, so if I wanted to include both `AMZN` and `iNTERNAL`, `{Custom Formats:AMZN,iNTERNAL}` would produce `AMZN iNTERNAL`. But if I wanted `iNTERNAL` first, I couldn't achieve that. `{Custom Format:iNTERNAL} {Custom Format:AMZN}` would however.

thezoggy:
shouldnt logic for the custom format be with the actual custom format? not in its naming token

Alanoll:
> shouldnt logic for the custom format be with the actual custom format? not in its naming token

This isn't logic for what is matched as a custom format, but whether that custom format after having matched is used when renaming.

`{Series Title} - s{season:00}e{episode:00} - {Custom Format:WEB-DL} - {MediaInfo VideoCodec} - {Custom Formats:-WEB-DL}`
The user wants `WEB-DL` (if matched) to be before video codec, and then all remaining matched Custom Formats at the end.

markus101 (approved):

PR #6626, opened 2024-03-14
Fixed: Release push with only Magnet URL

### LINKED ISSUE #6622: HTTP 400 on new Magnet releases from autobrr

### Current Behavior

I setup a Torznab RSS feed from Prowlarr for Nyaa on my autobrr, which i am using to monitor new releases and send them to Sonarr. I setup this feed to use Magnet links.

Whenever autobrr tries to push a nyaa release, the push gets rejected with a message saying `[[error: NotEmptyValidator] DownloadUrl: 'Download Url' must not be empty. - got value: ]`. I checked Sonarr and it is indeed returning a 400 status code for these releases. 
This issue only happens when using Magnet links. If I set up the RSS feed with Torrent links, it works well.

### Expected Behavior

The release should be submitted to Sonarr by autobrr, and possibly rejected, but not with the above error.

### Steps To Reproduce

1. Add Nyaa indexer to Prowlarr
2. Add Generic Torznab indexer to autobrr, pointing to the Nyaa indexer in Prowlarr (**make sure the Download Type is set to Magnet**)
3. Create filter for Nyaa indexer with an Action to send to a Sonarr instance
4. Wait for a new release to be added to Nyaa
5. Confirm that the release always gets rejected by Sonarr with the above error message

Screenshot 2024-03-12 at 10 54 41 AM
Screenshot 2024-03-12 at 10 55 13 AM

### Environment

```markdown
- OS: Ubuntu LTS 22.04.4
- Sonarr: 4.0.1.929
- Docker Install: Yes
- Using Reverse Proxy: Yes
- Database: Sqlite 3.44.2
```

Main

### Trace Logs?

(see timestamp 13:45:56)

Sonarr Logs: https://pastebin.com/cnVYgLDB
Prowlarr Logs: https://pastebin.com/LKVdTntE

### Anything else?

I went through autobrr's source code and noticed that when i am using torrents, it sends the link under the field `downloadUrl` and when i use magnets it sends the link under `magnetUrl` ([torznab.go](https://github.com/autobrr/autobrr/blob/develop/internal/feed/torznab.go#L129-L132) and [sonarr.go]( https://github.com/autobrr/autobrr/blob/develop/internal/action/sonarr.go#L58-L59)). I feel like Sonarr always expects a `downloadUrl` and that's why the error is thrown. Is there a reason why this field would always be expected?

markus101:
#### Description
Allows `MagnetUrl` without `DownloadUrl` and vice versa.

* Closes #6622

### DIFF
--- a/src/Sonarr.Api.V3/Indexers/ReleasePushController.cs
+++ b/src/Sonarr.Api.V3/Indexers/ReleasePushController.cs
@@ -41,7 +41,8 @@ namespace Sonarr.Api.V3.Indexers
             _logger = logger;
 
             PostValidator.RuleFor(s => s.Title).NotEmpty();
-            PostValidator.RuleFor(s => s.DownloadUrl).NotEmpty();
+            PostValidator.RuleFor(s => s.DownloadUrl).NotEmpty().When(s => s.MagnetUrl.IsNullOrWhiteSpace());
+            PostValidator.RuleFor(s => s.MagnetUrl).NotEmpty().When(s => s.DownloadUrl.IsNullOrWhiteSpace());
             PostValidator.RuleFor(s => s.Protocol).NotEmpty();
             PostValidator.RuleFor(s => s.PublishDate).NotEmpty();
         }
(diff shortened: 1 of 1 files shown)


zze0s:
Thanks for the quick fix @markus101!

PR #6637, opened 2024-03-20
Fix bug in which Minimum retention was not being considered for qBittorrent client test

carlos-sarmiento:
#### Description
The validation added on #6050 was not included in the code for the connection Test between Sonarr and the qBittorrent client. This means that even with a correctly configured client, sonarr would complain that torrents could be deleted before they were imported. This PR fixes this issue by including the same validation on the Test function

### DIFF
--- a/src/NzbDrone.Core/Download/Clients/QBittorrent/QBittorrent.cs
+++ b/src/NzbDrone.Core/Download/Clients/QBittorrent/QBittorrent.cs
@@ -388,19 +388,18 @@ namespace NzbDrone.Core.Download.Clients.QBittorrent
                 }
             }
 
-            var minimumRetention = GetMinimumRetention();
-
             return new DownloadClientInfo
             {
                 IsLocalhost = Settings.Host == "127.0.0.1" || Settings.Host == "localhost",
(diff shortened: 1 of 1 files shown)


src/NzbDrone.Core/Download/Clients/QBittorrent/QBittorrent.cs
```diff
@@ -388,7 +388,7 @@ public override DownloadClientInfo GetStatus()
                 }
             }
 
-            var minimumRetention = 60 * 24 * 14;
+            var minimumRetention = GetMinimumRetention();
```
markus101, line 391: Instead of extracting the minimum value, extract the logic for calculating `RemovesCompletedDownloads` that way it can be re-used in `TestConnection` without duplicating it.

markus101 (changes requested):

src/NzbDrone.Core/Download/Clients/QBittorrent/QBittorrent.cs
carlos-sarmiento: Fixed

carlos-sarmiento:
Updated and ready to review

markus101 (approved):
Looks good, as long as CI is happy.

PR #6640, opened 2024-03-21
Fix incorrect subtitle copy regex and extra file script import

JeWe37:
#### Description
Previously, subtitle titles such as `test 1` would incorrectly be parsed as having no title with index `1`, despite the correct format for a title with copy being `test - 1`. This is fixed.

Quite literally a single character change.

#### Database Migration
NO

*

### DIFF
--- a/src/NzbDrone.Core/Extras/ExistingExtraFileService.cs
+++ b/src/NzbDrone.Core/Extras/ExistingExtraFileService.cs
@@ -10,7 +10,7 @@ namespace NzbDrone.Core.Extras
 {
     public interface IExistingExtraFiles
     {
-        List<string> ImportExtraFiles(Series series, List<string> possibleExtraFiles, bool keepExistingEntries);
+        List<string> ImportExtraFiles(Series series, List<string> possibleExtraFiles, string fileNameBeforeRename);
     }
 
     public class ExistingExtraFileService : IExistingExtraFiles, IHandle<SeriesScannedEvent>
@@ -25,7 +25,7 @@ namespace NzbDrone.Core.Extras
             _logger = logger;
         }
 
-        public List<string> ImportExtraFiles(Series series, List<string> possibleExtraFiles, bool keepExistingEntries)
+        public List<string> ImportExtraFiles(Series series, List<string> possibleExtraFiles, string fileNameBeforeRename)
         {
             _logger.Debug("Looking for existing extra files in {0}", series.Path);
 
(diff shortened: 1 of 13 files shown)


JeWe37:
Noticed something else that needs fixing, draft for now.

markus101:
Please add a test case as well.

JeWe37:
Test added. 

The other addition does the following: Previously, the existing subtitle service always deleted all subtitle files that weren't explicitly listed as possible subtitle files. This makes sense in the case of a series refresh, but when importing via a script we use the existing subtitle service with only the possible files of a single episode. This means that all other subtitle files would be incorrectly removed from the DB, which meant they wouldn't be properly deleted alongside the episodes. The fix for this essentially simply disables removal of subtitle files for this case entirely and assumes none exist, as in the case of an upgrade, the subtitle files should already be removed by the deletion of the old episode file.

src/NzbDrone.Core/Extras/ExistingExtraFileService.cs
```diff
@@ -10,7 +10,7 @@ namespace NzbDrone.Core.Extras
 {
     public interface IExistingExtraFiles
     {
-        List<string> ImportExtraFiles(Series series, List<string> possibleExtraFiles);
+        List<string> ImportExtraFiles(Series series, List<string> possibleExtraFiles, bool import = true);
```
markus101, line 13: Instead of the very generic `import` indicating what is calling it, this chould be renamed to how it will be used, something like `cleanRemovedFiles` or a better name for whether all series files are being passed in, maybe something like `possibleExtraFilesForEntireSeries`.

 I also don't think it needs a default value since it's only used in a couple places.

markus101 (changes requested):
Mostly thoughts on the naming.

JeWe37:
Alright another fix(if you want me to split things out into separate PRs let me know):

Consider the following scenario: You are importing a file and based on the MediaInfo Sonarr parses from it, it is decided it will be

JeWe37:
Found another minor issue, this one I am less sure how to fix though: If a series gets rescanned while an import via a script is running, its extra files will get imported before the import finishes, thus won't be picked up by the import at the end of the script import and also won't be renamed correctly alongside the episode.

Does Sonarr have any system that I can use to lock individual series for a task yet? I wouldn't love to introduce a whole new locking system for such an issue, but I am not sure if there are any alternatives.

markus101 (approved):

JeWe37:
Ah. I hadn't tested this with normal renaming. This should have been obvious in retrospect. It's a one line fix, I can submit that as a PR tomorrow if you want me to.

JeWe37:
It is. The FileNameBeforeRename is only relevant in the scenario that this is used through the script import in the import of new episodes. Whenever that path is used(and only then), the localEpisode should be non null.

PR #6647, opened 2024-03-22
UI change to show a footnote to indicate episode title/series title/release group truncation is supported

fireph:
#### Description

I noticed when looking through the code that episode title/series title/release group truncation is supported but not documented in the UI, so thought it would be helpful to add a footnote in the file naming dialog that indicated the support and how to use it.

image

image

image

#### Database Migration

Not needed.


src/NzbDrone.Core/Localization/Core/en.json
```diff
@@ -635,6 +635,7 @@
   "EpisodeRequested": "Episode Requested",
   "EpisodeSearchResultsLoadError": "Unable to load results for this episode search. Try again later",
   "EpisodeTitle": "Episode Title",
+  "EpisodeTitleFootNote": "Episode Title/CleanTitle support a `:MAX_LENGTH` suffix allowing you to truncate the episode title. Ellipsis (`...`) will be added when truncating. The length of the truncated text (including ellipsis) will be equal to the specificied max length. For example `{Episode Title:30}`.",
```
markus101, line 638: The information feels a little repetitive here, I don't think the token name needs to be repeated and can probably be cut down to a single sentence with examples.

Truncating from the beginning is also supported with `:-30` and the actual truncation is done on the number of bytes (which is mostly an issue when truncating due to file length limits).

markus101 (changes requested):
The maximum length is not limited to episode titles, series title and release group also support truncation.

fireph:
> The maximum length is not limited to episode titles, series title and release group also support truncation.

Ah ok, thank you for the feedback. I'll reword to be more clear and less wordy. I'll also add footnotes for series title/release group

fireph:
I updated the language and tried to make it less wordy (although it is tricky to make it short and still indicate that positive and negative max lengths are supported).

src/NzbDrone.Core/Localization/Core/en.json
```diff
@@ -635,6 +635,7 @@
   "EpisodeRequested": "Episode Requested",
   "EpisodeSearchResultsLoadError": "Unable to load results for this episode search. Try again later",
   "EpisodeTitle": "Episode Title",
+  "EpisodeTitleFootNote": "Truncate episode titles with `:MAX_LENGTH` (bytes). Ellipsis (`...`) is added. Positive length truncates from the end. (e.g. `{Episode Title:30}`). Negative length truncates from the beginning. (e.g. `{Episode Title:-30}`).",
```
markus101, line 638: I agree it's hard to cut it down too much, but I think the examples and `:MAX_LENGTH` end up being redundant, but maybe something like this.

```
Optionally control truncation to a maximum number of bytes including ellipsis (`...`). Truncating from the end (e.g. `{Episode Title:30}`) or the beginning (e.g. `{Episode Title:-30}`) are both supported.
```

For episode title it will also be automatically truncated to fit maximum file system lengths, so for this also add: `Episode titles will be automatically truncated to file system limitations if necessary`.

markus101 (changes requested):
Thanks for adding the others, left some thoughts on the wording.

markus101 (approved):
Thanks for the updates, will land after CI passes.

frontend/src/Settings/MediaManagement/Naming/NamingModal.js
```diff
@@ -468,6 +475,11 @@ class NamingModal extends Component {
                         )
                       }
                     </div>
+
```
markus101, line 478:
```suggestion

```

markus101 (commented):
Just need to fix the lint error.

PR #6659, opened 2024-03-26
New: Increase rate limit for automatic searches

mynameisbogdan:
#### Description
Since automatic searches are made in the background and some trackers allow 150 requests/hour, with a rate limit of 30s it should not go above this limit.


markus101:
I don't think it makes sense to cater to the maximum requests per hour this way, 150/h or 150/15 minutes and then nothing for 45 minutes is still the same thing as long as the indexer is okay with the rate.

markus101:
If we want to impose a limit of 150 requests per hour then that shouldn't be enforced by rate limiting requests to be spread out over the full hour, because it shouldn't matter if they use 150 requests in 10 minutes or 60 minutes, but then we need to track all the requests per indexer.

PR #6671, opened 2024-04-01
Update en.json

TheFiZi:
Sonarr, Radarr and Prowlarr all say "New update is available" but I can't determine which one it is without checking each one since I use the same notification method for each of them. This should resolve that.

#### Description
I just want the update message to mention which service has the update available.


markus101:
This is intentional so the same translations can be shared across all projects, this Gould be handled at the notification level instead.

TheFiZi:
> This is intentional so the same translations can be shared across all projects, this Gould be handled at the notification level instead.

Understood. I'll go digging in there.

Thirrian:
Can't `{appName}` be used like in other translations? A few lines below:
`"UpdateMechanismHelpText": "Use {appName}'s built-in updater or a script",`

markus101:
Normally yes, but it doesn't make a lot of sense to add `{appName}` to every health check message so the app name is always included with those messages.

PR #6674, opened 2024-04-01
Specify widely supported display mode for PWA

cuki:
#### Description
With the changes from #6231 by changing `display` to `minimal-ui` it broke Progressive Web App behavior on iOS.
Instead of opening the *arr application as a standalone (PWA) instance it opens a new tab in Safari.

By looking at [browser compatibility](https://developer.mozilla.org/en-US/docs/Web/Manifest/display#browser_compatibility) of `display` in **manifest.json**, you can see that `minimal-ui` is not supported on iOS.
image

Reverting these changes made to **manifest.json** the *arr application will properly open a standalone (PWA) instance on iOS.

In this video you can see the difference between current (4.0.2.1183) and fixed behavior on iOS.

https://github.com/Sonarr/Sonarr/assets/2996147/273b7c4c-049a-496a-867e-23e0f80edbcd

#### Database Migration
NO

### DIFF
--- a/frontend/src/Content/Images/Icons/manifest.json
+++ b/frontend/src/Content/Images/Icons/manifest.json
@@ -15,5 +15,5 @@
     "start_url": "../../../../",
     "theme_color": "#3a3f51",
     "background_color": "#3a3f51",
-    "display": "minimal-ui"
+    "display": "standalone"
 }


cuki:
> I agree with reverting this back, since it fully broke iOS just to support something on Android. 😬

Not sure the changes were introduced to support something on Android, based on original branch name (`macos-pwa-fix`) in pull request #6231, my assumption is that the intent was to add (navigation) support for [Safari web apps on Mac](https://support.apple.com/en-us/104996), although this requires macOS Sonoma or later (and support is questionable looking at iOS Safari support).

markus101 (approved):
Thanks for the fix.

PR #6675, opened 2024-04-01
Add info box to custom formats modal

stevietv:
#### Description
Add a small info box to the custom formats modal, clarifying how CFs are applied when multiple specs are chosen.

image


stevietv:
> Wording needs work
> 
> * not all CF conditions are release title based
> * Release title or file name is inaccurate as at times neither are used as the scene title is actually used

tweaked to say  "release or file" instead of "release name or file name".
This way we imply that it applies to releases (during search/rss/import) or files (in library/during import)

markus101 (approved):

PR #6687, opened 2024-04-05
Fixed: Skip move when source and destination are the same

mynameisbogdan:
#### Description
As per title.

### DIFF
--- a/src/NzbDrone.Core/Tv/MoveSeriesService.cs
+++ b/src/NzbDrone.Core/Tv/MoveSeriesService.cs
@@ -1,6 +1,7 @@
 using System.IO;
 using NLog;
 using NzbDrone.Common.Disk;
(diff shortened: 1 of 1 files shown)


src/NzbDrone.Core/Tv/MoveSeriesService.cs
```diff
@@ -51,6 +52,12 @@ private void MoveSingleSeries(Series series, string sourcePath, string destinati
                 _logger.ProgressInfo("Moving {0} from '{1}' to '{2}'", series.Title, sourcePath, destinationPath);
             }
 
+            if (sourcePath.PathEquals(destinationPath))
+            {
+                _logger.ProgressInfo("{0} is already in the specified location '{1}'.", series, destinationPath);
```
bakerboy448, line 57: `series` or `series.Title` ?

PR #6688, opened 2024-04-06
Changed Telegram notifications to use branded strings

TheFiZi:
Changed Telegram notifications to use branded strings

#### Description
This change will brand notifications sent from Sonarr via Telegram helping clear up which service send the notification for those of us running multiple arr's

### DIFF
--- a/src/NzbDrone.Core/Notifications/Telegram/Telegram.cs
+++ b/src/NzbDrone.Core/Notifications/Telegram/Telegram.cs
@@ -18,47 +18,47 @@ namespace NzbDrone.Core.Notifications.Telegram
 
         public override void OnGrab(GrabMessage grabMessage)
         {
-            _proxy.SendNotification(EPISODE_GRABBED_TITLE, grabMessage.Message, Settings);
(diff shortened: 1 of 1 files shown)


markus101:
Please give your PRs a meaningful title, rather than just the name of the file you changed.

The problem with this is now everyone that has Telegram setup, regardless of using multiple services or using separate bots that already differentiate this for them would have the branded title added, which is not something we want to do. This needs to be done in a way that doesn't force everyone's notifications to change, via an option or taking it a step further to add additional customization that https://github.com/Sonarr/Sonarr/issues/328 covers.

TheFiZi:
Sorry about the title, will keep that in mind for next time.

Since adding to / altering the UI is beyond my skill set I'll just close this off. Radarr and Prowarr accepted the same changes so at least in my case I'll know the unbranded notification means Sonarr which solves things for me.

TheFiZi:
> @TheFiZi Actually it wasn't merged in Prowlarr yet since I was waiting for this PR to merged with a proper commit message.

Oh. Is that something I need to do?

markus101:
No UI changes are required, settings on the backend control what's visible on the front end. https://github.com/Sonarr/Sonarr/pull/6691

PR #6690, opened 2024-04-06
Fixed: Improve AniList testing with Media filters

markus101:
#### Description

Testing of AniList will return a warning if all results are filtered out by 'Media' filters, this will now query additional pages to find at least one valid result before returning that warning.


markus101:
Yeah, if we were implementing it for the first time again I'd probably exclude the media filtering to avoid this issue, but at least this improves the behaviour.

Haven't noticed random 500s myself, but I haven't really been testing it too much.

PR #6695, opened 2024-04-07
Add DevContainer and VSCode config

joshka:
#### Description

This makes it possible to develop Sonarr in a DevContainer with VSCode (which is useful for running this easily on a Mac).

On an M1/M2 Mac, I found that the default linux-x64 runtime identifier
is not valid for the aarch64 docker image that will be used by default.
So instead of hardcoding the runtime identifier, I've added a check to
set the architecture based on the current process architecture.

#### Testing

Tested working on an M2 Mac with Docker Desktop not using Rosetta.

1. Install docker desktop (or your preferred flavor of docker)
2. Open the repo in VSCode and accept the suggestion to reload into the DevContainer
3. Run the npm: install and npm: build tasks 
4. Run | Run without Debugging
5. Click button to load in browser

image

#### Background on the architecture issue

The existing code specifies `linux-x64` for the architecture, and it's not possible to pass `-p:RuntimeIdentifier` (note singular property name, not plural) to the `dotnet msbuild command`. When the incorrect runtime identifier is used, the software builds, but when run cannot find a file in `/lib64` (which doesn't exist on arm64 / aarch64 docker containers). Using an x64 image instead of an arm64 image is much slower for pretty much everything.

I'd guess there's possibly a more correct way to do this (perhaps using Platform / PlatformTarget), but I'm not familiar with exactly how these should be used (and how the customization in Sonarr effect these props).

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,19 @@
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the
+// README at: https://github.com/devcontainers/templates/tree/main/src/dotnet
+{
+  "name": "Sonarr",
+  "image": "mcr.microsoft.com/devcontainers/dotnet:1-6.0",
+  "features": {
+    "ghcr.io/devcontainers/features/node:1": {
+      "nodeGypDependencies": true,
+      "version": "16",
+      "nvmVersion": "latest"
+    }
+  },
+  "forwardPorts": [8989],
+  "customizations": {
+    "vscode": {
+      "extensions": ["esbenp.prettier-vscode"]
+    }
+  }
+}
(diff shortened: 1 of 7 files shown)


joshka:
> What do you think of adding as well `extensions.json` with recommendations for `ms-dotnettools.csharp` or `ms-dotnettools.csdevkit` or devcontainers makes this redundant?

It's not redundant for anyone that wants to dev on their host directly instead of in a container. I added the kit and the devcontainers plugin to recccomendations in extensions.json in https://github.com/Sonarr/Sonarr/pull/6695/commits/bee4f71f978c8e1cefb89ae35cdb6135f7cdd099. The dotnet devcontainer automatically adds `"ms-dotnettools.csharp"` inside the container (from https://github.com/devcontainers/images/blob/main/src/dotnet/.devcontainer/devcontainer.json).

markus101:
Thanks for contributing!

PR #6712, opened 2024-04-11
Improve release notes for main releases

markus101:
#### Description

For all main releases release notes will be generated off the previous main release, instead of off of develop, this will provide a much more accurate list of actual changes in these releases. For develop releases the releases will be built off the previous develop release (though the impact this is much lower).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -217,7 +217,7 @@ jobs:
 
   deploy:
     if: ${{ github.ref_name == 'develop' || github.ref_name == 'main' }}
-    needs: [backend, unit_test, unit_test_postgres, integration_test]
+    needs: [backend, frontend, unit_test, unit_test_postgres, integration_test]
     secrets: inherit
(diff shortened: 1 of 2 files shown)


joshka:
As an aside, have you considered calling develop -> main and main -> release?

This would have the benefit of just working with the idiomatic approaches that most tooling and projects expect (that main is the branch that gets new features). I understand this isn't all projects, and I'm curious what the rationale was for this aproach in the first place. From a few docs / PRs I've read this is something that you sometimes have to educate new contributors about, but perhaps it would be easier to just do the common thing instead?

markus101:
Having two branches, typically develop and main is Gitflow, though I've used staging and main on different projects at work.
https://nvie.com/posts/a-successful-git-branching-model/
https://www.atlassian.com/git/tutorials/comparing-workflows/gitflow-workflow

They could be named anything, but at this stage there isn't a lot of confusion with the naming and I don't think naming it main and release would eliminate any confusion around closed GHI = in release build.

PR #6718, opened 2024-04-12
Update name for RSS only indexers

mynameisbogdan:
#### Description
Some users are confusing these indexers being search capable so I think would be better to indicate that's RSS only in the naming.


bakerboy448:
If they don't read that it's RSS only and they don't read the check boxes and they don't read that search is not available.... are they really going to read the 3 characters of RSS in the name?

(still a good change, just unsure if it'll solve the issue of users blindly clicking and not reading anything at all)

markus101:
I agree with Baker here, if they can't read it in the places where it's already bc Allen out they aren't going to read it when they add the indexer.

I think we'd get more value out of a badge/label that is shown on the indexer card, but even then the value is pretty limited due to the previously mentioned reading issues.

PR #6721, opened 2024-04-13
Update action and add merge conflict labeler

Gauvino:
#### Description
Added a merge conflict labeler workflow for all the PR when a merge conflict is detected so people can resolve it when commit on develop branch is made

`merge-conflict` label need to be made before merging this PR

Update action version to remove the warning and adapt parameters on labeler

### DIFF
--- a/.github/workflows/conflict_labeler.yml
+++ b/.github/workflows/conflict_labeler.yml
@@ -21,3 +21,4 @@ jobs:
         with:
           dirtyLabel: 'merge-conflict'
           repoToken: "${{ secrets.GITHUB_TOKEN }}"
+          
--- a/.github/workflows/conflict_labeler.yml
+++ b/.github/workflows/conflict_labeler.yml
@@ -1,0 +1,23 @@
+name: Merge Conflict Labeler
+
+on:
+  push:
+    branches:
+      - develop
+  pull_request_target:
+  issue_comment:
+
+permissions: {}
+
+jobs:
+  label:
+    name: Labeling
+    runs-on: ubuntu-latest
+    if: ${{ github.repository == 'Sonarr/Sonarr' }}
+    steps:
+      - name: Apply label
+        uses: eps1lon/actions-label-merge-conflict@v3
+        if: ${{ github.event_name == 'push' || github.event_name == 'pull_request_target'}}
+        with:
+          dirtyLabel: 'merge-conflict'
+          repoToken: "${{ secrets.GITHUB_TOKEN }}"


.github/workflows/conflict_labeler.yml
```diff
@@ -0,0 +1,24 @@
...
+  pull_request_target:
+  issue_comment:
+
+permissions: {}
+
+jobs:
+  label:
+    name: Labeling
+    runs-on: ubuntu-latest
+    if: ${{ github.repository == 'Sonarr/Sonarr' }}
```
markus101, line 16: This will prevent it from being used down stream without changes, which makes it a little awkward. Is there a way to check if it's a fork and not run on forks?

.github/workflows/conflict_labeler.yml
```diff
@@ -0,0 +1,24 @@
...
+    steps:
+      - name: Apply label
+        uses: eps1lon/actions-label-merge-conflict@v3
+        if: ${{ github.event_name == 'push' || github.event_name == 'pull_request_target'}}
+        with:
+          dirtyLabel: 'merge conflict'
```
markus101, line 22: This would be more in line with our existing labels.

```suggestion
          dirtyLabel: 'merge-conflict'
```

.github/workflows/conflict_labeler.yml
```diff
@@ -0,0 +1,24 @@
...
+          commentOnDirty: 'This pull request has merge conflicts. Please resolve the conflicts so the PR can be successfully reviewed and merged.'
```
markus101, line 23: Is it possible to exclude the comment? I don't think it is necessary as authors will get notifications when the label is added.

.github/workflows/build.yml
```diff
@@ -236,7 +236,7 @@ jobs:
 
     steps:
       - name: Notify
-        uses: tsickert/discord-webhook@v5.3.0
+        uses: tsickert/discord-webhook@v6.0.0
```
markus101, line 239: Please do these separate version bumps in a different PR.

.github/workflows/conflict_labeler.yml
```diff
@@ -0,0 +1,24 @@
+name: Merge Conflict Labeler
+
+on:
+  push:
+    branches:
+      - develop
+  pull_request_target:
+  issue_comment:
```
markus101, line 8: Is this necessary?

markus101 (changes requested):
I agree with what Stevie and Bodgan mentioned on Discord, but I do like the automation of labeling and being able to filter out PRs that need reviews.

@stevietv @mynameisbogdan what do you think about having it label and exclude the comment?

.github/workflows/build.yml
Gauvino: will do it no problem

.github/workflows/conflict_labeler.yml
Gauvino: I don't think people get the notification when their PR is labeled, the only notification you get on a PR is comment and commit

.github/workflows/conflict_labeler.yml
Gauvino: I don't think so this is the best I found to do it, let me know if you know other way to do it

.github/workflows/conflict_labeler.yml
Gauvino: the name is issue_comment but it work also for PR see screenshot 

chrome_IgQXjJIRmF

.github/workflows/conflict_labeler.yml
Gauvino: So you would prefer no comment at all ?

.github/workflows/conflict_labeler.yml
```diff
@@ -0,0 +1,23 @@
...
+    name: Labeling
+    runs-on: ubuntu-latest
+    if: ${{ github.repository == 'Sonarr/Sonarr' }}
+    steps:
+      - name: Apply label
+        uses: eps1lon/actions-label-merge-conflict@v3
+        if: ${{ github.event_name == 'push' || github.event_name == 'pull_request_target'}}
+        with:
+          dirtyLabel: 'merge-conflict'
+          repoToken: "${{ secrets.GITHUB_TOKEN }}"
```
markus101, line 23: To match other files in the repo:

```suggestion
          repoToken: "${{ secrets.GITHUB_TOKEN }}"
          
```

markus101 (approved):
Thanks for the cleanup.

PR #6724, opened 2024-04-14
Implement equality checks for providers

mynameisbogdan:
#### Description
Comparing via JSON string doesn't work properly in Prowlarr, so I worked on implementing in the rest as well since it allows to ignore fields without breaking the API.

- [X] Indexers (extending directly `PropertywiseEquatable` in the settings class)
- [X] Notifications (implementing `NotificationSettingsBase`)
- [x] Download Clients (implementing `DownloadClientSettingsBase`)
- [x] Import Lists (implementing `ImportListSettingsBase`)

I'm uncertain if I should implement for import lists since many rely on the save to trigger a single list sync.


markus101:
> I'm uncertain if I should implement for import lists since many rely on the save to trigger a single list sync.

I think so, if nothing has changed they shouldn't kick off a new sync.

PR #6750, opened 2024-04-22
Fix wrongly named field in TransmissionProxy

mikabytes:
There is no such field `fileCount` in [Transmission RPC](https://github.com/transmission/transmission/blob/main/docs/rpc-spec.md#33-torrent-accessor-torrent-get), however, there is one called `file-count`. The naming is inconsistent, so the mistake is understandable. Also, it doesn't help that Transmission will ignore any fields it doesn't recognize.

I stumbled upon this while testing Sonarr with a client I'm developing which aims to be Transmission RPC compatible. Only, my code throws an error when asked for invalid fields (oops).

You may verify by running these commands:

_(this returns the field)_

```
curl -H "X-Transmission-Session-Id: <session-id>" <host>/transmission/rpc --data-raw '{ "method": "torrent-get", "arguments": { "fields": ["file-count"] } }'
```

_(this does not return field)_

```
curl -H "X-Transmission-Session-Id: <session-id>" <host>/transmission/rpc --data-raw '{ "method": "torrent-get", "arguments": { "fields": ["fileCount"] } }'
```

### DIFF
--- a/src/NzbDrone.Core/Download/Clients/Transmission/TransmissionProxy.cs
+++ b/src/NzbDrone.Core/Download/Clients/Transmission/TransmissionProxy.cs
@@ -178,7 +178,7 @@ namespace NzbDrone.Core.Download.Clients.Transmission
                 "seedRatioMode",
                 "seedIdleLimit",
                 "seedIdleMode",
-                "fileCount"
+                "fileCount",
                 "file-count"
             };
 
--- a/src/NzbDrone.Core/Download/Clients/Transmission/TransmissionProxy.cs
+++ b/src/NzbDrone.Core/Download/Clients/Transmission/TransmissionProxy.cs
@@ -179,6 +179,7 @@ namespace NzbDrone.Core.Download.Clients.Transmission
                 "seedIdleLimit",
                 "seedIdleMode",
                 "fileCount"
+                "file-count"
             };
 
             var arguments = new Dictionary<string, object>();
--- a/src/NzbDrone.Core/Download/Clients/Transmission/TransmissionTorrent.cs
+++ b/src/NzbDrone.Core/Download/Clients/Transmission/TransmissionTorrent.cs
@@ -1,4 +1,6 @@
-﻿namespace NzbDrone.Core.Download.Clients.Transmission
+using Newtonsoft.Json;
+
+namespace NzbDrone.Core.Download.Clients.Transmission
 {
     public class TransmissionTorrent
     {
@@ -20,6 +22,12 @@
         public int SeedRatioMode { get; set; }
         public long SeedIdleLimit { get; set; }
         public int SeedIdleMode { get; set; }
-        public int FileCount { get; set; }
+        public int FileCount => TransmissionFileCount ?? VuzeFileCount ?? 0;
+
+        [JsonProperty(PropertyName = "file-count")]
+        public int? TransmissionFileCount { get; set; }
+
+        [JsonProperty(PropertyName = "fileCount")]
+        public int? VuzeFileCount { get; set; }
     }
 }


markus101:
Does the property come back as `file-count` or `fileCount`? If it's the former then https://github.com/Sonarr/Sonarr/blob/develop/src/NzbDrone.Core/Download/Clients/Transmission/TransmissionTorrent.cs#L23 also needs to be addressed.

`fileCount` was used to fix an issue with Vuze (uses a similar API). Offhand I'm not sure if Vuze actually uses `fileCount`, but that change has been place for 6 years. https://github.com/Sonarr/Sonarr/commit/1467c52e03ca711f54f202340bb68eabe2e5c50d

mikabytes:
Hi markus101,

The behavior is like this (tested on Transmission 4.0.5):

| Request arguments                            | Reponse arguments                                     |
| -------------------------------------------- | ----------------------------------------------------- |
| `{ "ids": 445, "fields": ["fileCount"] } }`  | `{"torrents":[{}]},"result":"success"}`               |
| `{ "ids": 445, "fields": ["file-count"] } }` | `{"torrents":[{"file-count":3}]},"result":"success"}` |

Transmission silently ignores fields it does not recognize. This means that for 6 years Sonarr has asked for `fileCount` but hasn't received it.

I find it concerning that this field is used in the Vuze fix you mentioned. So I looked at the changes in RPC spec but could find no mention of `fileCount` ever existing. 

We might be better off removing this field altogether. That would keep the same behavior as previously, whereas this PR would make it return real data which could have side effects.

What do you think?

markus101:
I've confirmed that Vuze uses `fileCount`, not `file-count`, but also ignores things it doesn't know about. We'll need to handle both, but shouldn't be too difficult.

This diff is working with both Vuze and Transmission and consumers can just use `FileCount` as they did before.

```
diff --git a/src/NzbDrone.Core/Download/Clients/Transmission/TransmissionTorrent.cs b/src/NzbDrone.Core/Download/Clients/Transmission/TransmissionTorrent.cs
index e72fa55a7..3552c36e9 100644
--- a/src/NzbDrone.Core/Download/Clients/Transmission/TransmissionTorrent.cs
+++ b/src/NzbDrone.Core/Download/Clients/Transmission/TransmissionTorrent.cs
@@ -1,4 +1,6 @@
-﻿namespace NzbDrone.Core.Download.Clients.Transmission
+using Newtonsoft.Json;
+
+namespace NzbDrone.Core.Download.Clients.Transmission
 {
     public class TransmissionTorrent
     {
@@ -20,6 +22,12 @@ public class TransmissionTorrent
         public int SeedRatioMode { get; set; }
         public long SeedIdleLimit { get; set; }
         public int SeedIdleMode { get; set; }
-        public int FileCount { get; set; }
+        public int FileCount => TransmissionFileCount ?? VuzeFileCount ?? 0;
+
+        [JsonProperty(PropertyName = "file-count")]
+        public int? TransmissionFileCount { get; set; }
+
+        [JsonProperty(PropertyName = "fileCount")]
+        public int? VuzeFileCount { get; set; }
     }
 }
```

mikabytes:
Thanks for the patch @markus101 

I've applied it to this PR, and re-added `fileCount` to the requested fields. This way both `fileCount` and `file-count` will be requested from the the torrent client, and then your code should pick the right one as they come back.

I don't have the environment set up to test the code. Can you or someone else verify that it works as expected?

markus101 (approved):

src/NzbDrone.Core/Download/Clients/Transmission/TransmissionProxy.cs
```diff
@@ -179,6 +179,7 @@ private TransmissionResponse GetTorrentStatus(IEnumerable<string> hashStrings, T
                 "seedIdleLimit",
                 "seedIdleMode",
                 "fileCount"
```
markus101, line 181:
```suggestion
                "fileCount",
```

PR #6756, opened 2024-04-25
Fixed: Limit titles in task name to 10 series

mynameisbogdan:
#### Description
Limiting titles in task names to max 20 series since on big libraries you need to scroll a lot to get to the next tasks.


markus101:
20 is still a lot of series titles to list out, even 5-10 would be sufficient.

PR #6766, opened 2024-04-28
Forward X-Forwarded-Host header

### LINKED ISSUE #6764: Reverse Proxy x-forwarded-host header ignored

### Current Behavior

A reverse proxy taking in traffic on `my.domain.com` (HTTPS) forwards traffic to `sonarr.internal.local` (HTTP) while adding the appropriate `x-forwarded-for`, `x-forwarded-proto` and `x-forwarded-host` HTTP headers. All good so far!

The forwarded HTTP request does also specify the `host` header as `sonarr.internal.local` but that should be okay, as `x-forwarded-host` (set to `my.domain.com`) should be honoured by Sonarr (but it isn't - this is the core of the issue).

What happens then is that things like redirects and resource loading all reference the `host` header value instead of the `x-forwarded-host` value. This shows up right away with a 302 redirect to a `location` of `https://sonarr.internal.local/login?returnUrl=%2F` rather than `https://my.domain.com/login?returnUrl=%2F`. Which is extra fun because it should be the http version, not the https (as `sonarr.internal.local` is http not https).

### Expected Behavior

Sonarr should be honouring the `x-forwarded-host` header and treating that as the domain that the site is on. It honours the `x-forwarded-proto` header only - which is just bad because if that's different from what's actually supported then you can get ssl errors etc.

### Steps To Reproduce

Set up a reverse proxy to forward traffic to Sonarr. No need for base URLs, but that doesn't change the result.
I used a test website in asp.net core using [YARP](https://github.com/microsoft/reverse-proxy).
Notice that the Sonarr will always refer to the `host` header value it was passed when doing redirects - rather than any `x-forwarded-host` value it was given. But it will honour the `x-forwarded-proto` value (HTTP/HTTPS)!
I guess a simpler way to reproduce this is just use curl / postman / whatever to make a request that includes the forwarded headers? Do that with no auth information and the redirect 302 to `whatever/login..` will be wrong.

### Environment

```markdown
- OS: Doesn't matter
- Sonarr: 4.0.4.1491
- Docker Install: No
- Using Reverse Proxy: Yes
- Browser: Any
- Database: Sqlite 3.36.0
```

Main

### Trace Logs?

No logs show the issue?

### Anything else?

I found the problem in [src\NzbDrone.Host\Startup.cs](https://github.com/Sonarr/Sonarr/blob/c81ae6546118e954e481894d0b3fa6e9a20359c7/src/NzbDrone.Host/Startup.cs#L59) where it says:
```
options.ForwardedHeaders = ForwardedHeaders.XForwardedFor | ForwardedHeaders.XForwardedProto;
```
It should be:
```
options.ForwardedHeaders = ForwardedHeaders.XForwardedFor | ForwardedHeaders.XForwardedProto | ForwardedHeaders.XForwardedHost;
```
Or, if you're feeling fancy:
```
options.ForwardedHeaders = ForwardedHeaders.All;
```
I've cloned locally and built with that change in and works great! Seems like this was just configured wrong?

I can make a PR with that simple change, if that's desired? Not sure on how to write an integration test for such a thing though, but happy to give that a go too if that's a thing.

I can just make the reverse proxy lie on the `host` header to make it work and I suspect that's what everyone does? Or maybe I'm doing something very silly xD Hope not! Sorry if I am! ♥

markus101:
#### Description
Include `X-Forwarded-Host` header in forwarded headers.

* Closes #6764

### DIFF
--- a/src/NzbDrone.Host/Startup.cs
+++ b/src/NzbDrone.Host/Startup.cs
@@ -56,7 +56,7 @@ namespace NzbDrone.Host
 
             services.Configure<ForwardedHeadersOptions>(options =>
             {
-                options.ForwardedHeaders = ForwardedHeaders.XForwardedFor | ForwardedHeaders.XForwardedProto;
+                options.ForwardedHeaders = ForwardedHeaders.XForwardedFor | ForwardedHeaders.XForwardedProto | ForwardedHeaders.XForwardedHost;
                 options.KnownNetworks.Clear();
                 options.KnownProxies.Clear();
             });


bakerboy448:
Should `proxy_set_header Host $host;` be removed from the nginx conf in the docs now that this has landed in main.... believe this makes that line redundant?

markus101:
Looks like it's still necessary at least in some cases, so leaving the wiki as-is.

PR #6767, opened 2024-04-28
New: Use absolute timestamps for series and episode history

mynameisbogdan:
#### Description
Switching to absolute dates to make it easier to debug upgrades for users. Bonus making the series history modal wider.

image

### DIFF
--- a/frontend/src/Episode/History/EpisodeHistory.js
+++ b/frontend/src/Episode/History/EpisodeHistory.js
@@ -66,6 +66,8 @@ class EpisodeHistory extends Component {
       isPopulated,
       error,
       items,
+      timeFormat,
+      shortDateFormat,
       onMarkAsFailedPress
     } = this.props;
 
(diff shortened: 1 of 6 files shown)


markus101:
Relative dates have an option to disable already, we could change the default for that to not be enabled, but I don't think we should remove relative dates from these history items completely.

stevietv:
From a support point of view I am enthusiastic about this change. Will help when discussing screenshots

markus101:
Adding an option to make relative date cells include a time stamp should cover it and another for second accuracy, to include seconds since that's missing today. Relative with timestamp it could be something like `{relativeDate} at {timestamp}` (which will be today/yesterday as long as people have their dates set correctly) to be properly translated, if it's something not rendered as a relative date use the short date format + timestamp.

This way we don't introduce a breaking change because relative dates are stripped out completely and we can still improve the support experience.

markus101:
Closing in favour of https://github.com/Sonarr/Sonarr/pull/6811

PR #6779, opened 2024-05-01
New: Blocklist Custom Filters

### LINKED ISSUE #6763: Filters for Blocklist

In some situations, you may want to remove certain items from the block list, but it becomes difficult to find them. For example, you add new NZB hosters to your Sabnzbd want to remove the block listed items from that protocol, you simply can't.

Similar filters as there are on searching individual episodes and the filter is available on the download list.

Manually going through and checking the indexer used on each individual block listed item across over 500 pages.

### Anything else?

_No response_

stevietv:
#### Description
Adds custom filters to Blocklist for Series and Protocol

What I thought might be an easy PR was somewhat tricky for me :)
I puzzled a bit here with the controller, and had to tweak the CF calculation service as the changes I made to the blocklist controller during my testing meant i was getting `null` for the `series` property being sent by the blocklist endpoint. It resulted in working custom filters, but please do shoot at me if it needs work.

image

image

* Closes #6763


src/Sonarr.Api.V3/Blocklist/BlocklistController.cs
```diff
@@ -23,12 +28,81 @@ public class BlocklistController : Controller
...
-            return pagingSpec.ApplyToPage(_blocklistService.Paged, model => BlocklistResourceMapper.MapToResource(model, _formatCalculator));
+            return pagingSpec.ApplyToPage(spec => GetBlocklist(spec, seriesIds?.ToHashSet(), protocol), model => BlocklistResourceMapper.MapToResource(model, _formatCalculator));
+        }
+
+        private PagingSpec<NzbDrone.Core.Blocklisting.Blocklist> GetBlocklist(PagingSpec<NzbDrone.Core.Blocklisting.Blocklist> pagingSpec, HashSet<int> seriesIds, DownloadProtocol? protocol)
+        {
+            var ascending = pagingSpec.SortDirection == SortDirection.Ascending;
+            var orderByFunc = GetOrderByFunc(pagingSpec);
+
+            var blocklist = _blocklistService.GetBlocklist();
```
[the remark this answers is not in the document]
stevietv: It may be, I followed along with how custom filters were implemented on the Queue I think 🤔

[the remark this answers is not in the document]
markus101: Queue is a bit of an outcast since it doesn't actually hit the DB, mirroring what History does is likely much closer to what we'll want to do as it will do filtering at the DB level.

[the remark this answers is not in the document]
stevietv: Think I figured that out :)

src/Sonarr.Api.V3/Blocklist/BlocklistController.cs
```diff
@@ -23,12 +25,17 @@ public class BlocklistController : Controller
...
         {
             var pagingResource = new PagingResource<BlocklistResource>(paging);
             var pagingSpec = pagingResource.MapToPagingSpec<BlocklistResource, NzbDrone.Core.Blocklisting.Blocklist>("date", SortDirection.Descending);
 
-            return pagingSpec.ApplyToPage(_blocklistService.Paged, model => BlocklistResourceMapper.MapToResource(model, _formatCalculator));
+            if (seriesIds != null && seriesIds.Any())
+            {
+                pagingSpec.FilterExpressions.Add(b => seriesIds.Contains(b.SeriesId));
+            }
+
```
markus101, line 37: Looks like for protocols we can do the same as series ID filtering:

```

if (protocols != null && protocols.Any())
{
    pagingSpec.FilterExpressions.Add(b => protocols.Contains(b.Protocol));
}
```

Which saves us passing protocol through, unless there is a reason why we wouldn't want to do that?

src/Sonarr.Api.V3/Blocklist/BlocklistController.cs
```diff
@@ -23,12 +25,17 @@ public class BlocklistController : Controller
...
         [HttpGet]
         [Produces("application/json")]
-        public PagingResource<BlocklistResource> GetBlocklist([FromQuery] PagingRequestResource paging)
+        public PagingResource<BlocklistResource> GetBlocklist([FromQuery] PagingRequestResource paging, [FromQuery] int[] seriesIds = null, [FromQuery] DownloadProtocol[] protocols = null)
         {
             var pagingResource = new PagingResource<BlocklistResource>(paging);
             var pagingSpec = pagingResource.MapToPagingSpec<BlocklistResource, NzbDrone.Core.Blocklisting.Blocklist>("date", SortDirection.Descending);
 
-            return pagingSpec.ApplyToPage(_blocklistService.Paged, model => BlocklistResourceMapper.MapToResource(model, _formatCalculator));
+            if (seriesIds != null && seriesIds.Any())
```
markus101, line 33: Not sure if `seriesIds?.Any() == true` is better, but thoughts?

markus101 (commented):
Looking good, just some minor things/questions.

src/Sonarr.Api.V3/Blocklist/BlocklistController.cs
stevietv: switched over to this. looks cleaner for sure.

src/Sonarr.Api.V3/Blocklist/BlocklistController.cs
stevietv: makes sense, I didn't consider this but have moved this to the pagingspec filter instead and removed ther work from the repo

PR #6806, opened 2024-05-09
New: Root folder exists validation for import lists

mynameisbogdan:
#### Description
Adding new validator for optional root folder existence.


src/NzbDrone.Core/Validation/Paths/SeriesExistsValidator.cs
```diff
@@ -25,7 +25,7 @@ protected override bool IsValid(PropertyValidatorContext context)
 
             var tvdbId = Convert.ToInt32(context.PropertyValue.ToString());
 
-            return !_seriesService.AllSeriesTvdbIds().Any(s => s == tvdbId);
+            return _seriesService.AllSeriesTvdbIds().All(s => s != tvdbId);
```
markus101, line 28: Is this faster or just for readability? If it's the latter we have `return _seriesService.AllSeriesTvdbIds().None(s => s == tvdbId)` for this purpose.

PR #6809, opened 2024-05-09
Fixed: Notifications with only On Rename enabled

mynameisbogdan:
#### Description
As per title, consider notifications with only `On Renamed` as enabled.


stevietv:
related to this I suppose? https://old.reddit.com/r/sonarr/comments/1cnr4hh/windows_cant_get_custom_script_to_work_ps1_any/

PR #6820, opened 2024-05-14
Plex Watchlist Can Be Empty

nopoz:
#### Description
Allows for Plex Watchlists to be empty at any time. With this code change, when they are empty an info logger message is displayed, rather than an error being thrown. It shouldn't be a requirement for the list to be populated at all times as they will likely always be in flux with entries being removed and added. 

This overcomes a few issues with the current implementation:
1. The UI throws an error and will not allow you to add a Plex Watchlist when it is empty. 
2. If you managed to add the list and it becomes empty later on, you will get errors in the logs and UI noting that the list is empty. 

A small improvement I would like to make is to have the logger message contain the list name so it's more obvious to the user which list is empty if multiples are used. I need some help with figuring out how to call the list name as it's shown in the UI Watchlist "name" field like:

```
_logger.Info($"No results were returned from Plex Watchlist '{listName}'.");
```

### DIFF
--- a/src/NzbDrone.Core/ImportLists/Plex/PlexImport.cs
+++ b/src/NzbDrone.Core/ImportLists/Plex/PlexImport.cs
@@ -128,7 +128,7 @@ namespace NzbDrone.Core.ImportLists.Plex
                 return new ValidationFailure(string.Empty, $"Unable to connect to Plex Watchlist: {ex.Message}. Check the log for details.");
             }
 
-            return null; // Indicate no fatal errors even if the list is empty.
+            return null;
         }
     }
 }
--- a/src/NzbDrone.Core/ImportLists/Plex/PlexImport.cs
+++ b/src/NzbDrone.Core/ImportLists/Plex/PlexImport.cs
@@ -7,6 +7,8 @@ using NzbDrone.Common.Extensions;
 using NzbDrone.Common.Http;
 using NzbDrone.Core.Configuration;
 using NzbDrone.Core.Exceptions;
+using NzbDrone.Core.ImportLists.Exceptions;
+using NzbDrone.Core.Indexers.Exceptions;
 using NzbDrone.Core.Localization;
 using NzbDrone.Core.Notifications.Plex.PlexTv;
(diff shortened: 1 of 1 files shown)


markus101:
> The UI throws an error and will not allow you to add a Plex Watchlist when it is empty.

It's a warning, so pressing save a second time will correctly save it, if that's not working then that should be fixed, but the current behaviour is what is expected.

> If you managed to add the list and it becomes empty later on, you will get errors in the logs and UI noting that the list is empty.

These changes wouldn't address that, unless you're only seeing it when testing/saving in which case that's expected.

nopoz:
> These changes wouldn't address that, unless you're only seeing it when testing/saving in which case that's expected.

Really? Could you point out the current code snippet which effect errors for scheduled imports? I was pretty sure it was from src/NzbDrone.Core/ImportLists/HttpImportListBase.cs

markus101:
Your changes only affect testing, which isn't used when fetching an already configured import list.

What error are you getting during scheduled imports?

nopoz:
I did some more testing and the "All lists are unavailable due to failures" error is produced in the UI at "System -> Status -> Health" when you edit and click the test button for a pre-existing plexwatch list which is empty and then click the cancel button or close out the edit window. I thought it was appearing automatically after a default scheduled task triggered it, but now I'm not sure. 

Health state at "System -> Status -> Health": 
system_status_health_error_state

Warning in the events list from the test:
pipeline_error

The above mentioned user action to trigger this condition is closely tied to the previous point you made:

> It's a warning, so pressing save a second time will correctly save it, if that's not working then that should be fixed, but the current behaviour is what is expected.

There is nothing in the UI to indicate that clicking through twice on the save button will bypass this warning. I've been using Sonarr for years and I had no idea this was a thing.

markus101:
> There is nothing in the UI to indicate that clicking through twice on the save button will bypass this warning.

Offhand I'm not sure how best to show that in the UI, but definitely something that could be improved.

Closing this since the the warning is valid and we need to come up with a better way to explain that in the UI.

PR #6849, opened 2024-05-31
New: Refresh cache for tracked queue on series add

mynameisbogdan:
#### Description
This should basically trigger a reimport if you decide to add back a series that exists in the cached queue without needing a restart to do the same thing.

Added same filtering by TvdbId in deleted series event.


src/NzbDrone.Core/Download/TrackedDownloads/TrackedDownloadService.cs
```diff
@@ -278,12 +279,29 @@ public void Handle(EpisodeInfoRefreshedEvent message)
             }
         }
 
+        public void Handle(SeriesAddedEvent message)
+        {
+            var cachedItems = _cache.Values
```
markus101, line 284: Do we need to do the same thing if the cached item doesn't have a series or are those not affected?

PR #6855, opened 2024-06-03
Fixed: Skip invalid series paths during validation

markus101:
#### Description
This currently prevents people from fixing invalid paths for a series, but we don't need to validate the existing stored series paths as they are validated elsewhere.


markus101:
Hmm yeah, it'd be nice to be able to do the move as well, might need an option to ignore the extra spaces for that purpose.

PR #6870, opened 2024-06-10
Fallback to remote url for series images

mynameisbogdan:
#### Description
This is needed for places where the series resource's image might not be mapped to local.


markus101:
> This is needed for places where the series resource's image might not be mapped to local.

Where does this happen currently?

Unless these images are proxied through Sonarr's backend then we're likely to see issues with hotloading being blocked, but in general I don't think we shouldn't try to load images remotely.

markus101:
I know add series is a special cases. TheTVDB does block hotlinking (or at least did, which is why those images get proxied through Sonarr).

PR #6875, opened 2024-06-12
New: Quality limits are part of Quality Profile

### LINKED ISSUE #613: Min/Max sizes at the profile level
With animated and standard releases having vast size differences (in some cases animated releases are half the size) it would be beneficial to have size limits enforced per profile instead of globally, or alternatively enforced via tags like delay profiles.

markus101:
#### Description

Moves size limits from Quality settings into individual quality profiles.

image
image
image

#### Database Migration
YES - 207

* Closes #613


Hannah-GBS:
I know it was discussed somewhat in the #development channel, but it would be nice to have the *option* of global settings for those that don't care to change things on every profile, and just have per-profile settings override the defaults.

markus101:
That was discussed previously, but we do not have plans to maintain both. We'll likely look to make it possible to copy size settings from other profiles though.

davidnewhall:
The additional insight is appreciated. Thanks!

rcdailey:
Hi Markus. I'm very out of the loop on any discussions had about this change, but I hope you won't mind filling me in on the API aspect of this. I have a couple of questions about your response to David:

> if the library leaves those fields 

Could you expand on this a bit? What do you mean?

My main concern is, as you'd expect, API backward compatibility. I see in David's linked issue that there is talk about a v4 API. Does this mean that v3 endpoints will continue to function (but be deprecated) in newer versions of Sonarr for some time?

I just want to make sure I have time to adopt these changes in Recyclarr prior to these changes being merged and/or released to stable channels, while still supporting older versions of Sonarr to give users time to upgrade.

Just looking at your diff, I see removal of fields in the quality definition DTOs here:
https://github.com/Sonarr/Sonarr/pull/6875/files?w=1#diff-55c98eda7357971a29a5110f901b227b4c6d95d37cc99dcf9660ce89037e1593

I think changes like this will create a lot of trouble for downstream tooling like Recyclarr, Notifiarr, etc. Is there any way you could implement a more backward-compatible change before merging this (or at least before going stable)?

Thanks for giving me an opportunity to participate!

markus101:
Thanks for reaching out Robert.

> Could you expand on this a bit? What do you mean?

These fields have always been nullable and from the serialization perspective `MaxSize = null` ends up being omitted from the JSON, with this change they just won't be returned, ever because they're effectively null. An API consumer trying to set these values would result in no errors, but nothing would be updated.

> I see in David's linked issue that there is talk about a v4 API. Does this mean that v3 endpoints will continue to function (but be deprecated) in newer versions of Sonarr for some time?

That is referring to a Radarr change, we don't have a v4 API planned for Sonarr at this time, but when we do, the previous API version will exist for at least one major release, same as the unversioned (v2) API still existed in v3 and was removed in v4.

> I think changes like this will create a lot of trouble for downstream tooling like Recyclarr, Notifiarr, etc

This is definitely my concern, I'd much prefer to not make it a breaking change at all, but short of having global and override size settings I think it'll need to be a breaking change. I don't think it makes sense long term because it'd be confusing, but I could see it being used as an interim solution.

> Is there any way you could implement a more backward-compatible change before merging this (or at least before going stable)?

I think the only backwards compatible change that has potential would be to still allow the API to update these values on all profiles, but I don't think we can reliably return that information because we'd only be able to choose a single profile's values or try to merge them together and get an average. This would allow consumers to use the same API, but would result in any individual customization done by the user being wiped out.

markus101:
Going to land this directly in the v5 work when it's ready.

rcdailey:
@markus101 I see this was closed but not merged. Will you be merging directly or through another PR later?

rcdailey:
Ah ok, so you're basically working on building an initial `v5` branch or similar; where the effort made here will be a subset of that branch.

Thanks for letting me know. As a client of the API I'm trying to stay prepared for the breaking changes I'll have to address.

PR #6879, opened 2024-06-14
New: Ignore inaccessible folders with getting folders

mynameisbogdan:
#### Description
Pretty much based on e5aa8584100d96a2077c57f74ae5b2ceab63de19 and d493f8762fcb1684b44e182753c21d7a493db787 to skip inaccessible folders that breaks manual import if the selected folder contains such folders.

https://github.com/Radarr/Radarr/issues/10085


markus101:
Since this has two separate commits, please change the first commit to `New: Ignore inaccessible folders when getting folders`

PR #6882, opened 2024-06-15
New: Display stats for all series in index footer

mynameisbogdan:
#### Description
Showing `$filtered/$total` in series index footer when a filter is selected.

image

### DIFF
--- a/frontend/src/Series/Index/SeriesIndexFooter.tsx
+++ b/frontend/src/Series/Index/SeriesIndexFooter.tsx
@@ -1,18 +1,40 @@
 import classNames from 'classnames';
-import React from 'react';
+import React, { useMemo } from 'react';
 import { useSelector } from 'react-redux';
 import { createSelector } from 'reselect';
 import { ColorImpairedConsumer } from 'App/ColorImpairedContext';
 import SeriesAppState from 'App/State/SeriesAppState';
 import DescriptionList from 'Components/DescriptionList/DescriptionList';
 import DescriptionListItem from 'Components/DescriptionList/DescriptionListItem';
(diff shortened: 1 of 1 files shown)


frontend/src/Series/Index/SeriesIndexFooter.tsx
```diff
@@ -31,47 +53,77 @@ function createUnoptimizedSelector() {
...
+    createUnoptimizedSeriesSelector(),
     (series) => series
   );
 }
 
+function generateSeriesStatistics(series: SeriesStatistics[]) {
+  return series.reduce(
+    (acc, s) => {
+      const {
+        statistics = { episodeCount: 0, episodeFileCount: 0, sizeOnDisk: 0 },
```
markus101, line 65: Can we avoid setting the defaults twice?

markus101 (commented):
I like the direction, but I'm not sure about the final result, it ends up feeling more cluttered and I'm not sure it's clear enough that it's showing the filtered counts vs total counts.

markus101:
I saw that discussion on mobile, but didn't look when I was back on a computer so missed the full context.
Tooltips work well for desktop, less good for mobile, though maybe we could do something like so it's more obvious it's filtered without needing to go the whole way with it.

`Total Series 5`
`Filtered Series 2`

markus101:
I don't like that either, definitely word overload there.

What if we showed an extra field when filtered that shows a popover with all the information via something like the `(i)` icon. By default we could either show the information for all series or we could show the filtered information and show all in the popup or vice versa, but I think I'd lean a bit more to showing the filtered values in the popover.

PR #6930, opened 2024-07-04
New: Replaced HasUnmonitored with Seasons Monitored Status

### LINKED ISSUE #6896: Create New Filter for Percentage of Seasons Monitored

Currently you are able to filter for shows that have an unmointored season, but not for shows that have a monitored season or what percentage of seasons are monitored. This would be helpful for showing shows that the currents seasons you have "completed". By unmonitoring seasons for which you don't want to search for any more downloads, but keeping the show monitored, you can track your progress.

Create a Percent of seasons monitored filter to replace HasUnmonitoredSeason

Would be able to see if all seasons are monitored by using =100
Would be able see if any season is unmonitored by using <100
Would be able to see if all seasons are unmonitored by using = 0
Would also allow for any number in between (like 50%)

Migrate any filter using HasUnmonitoredSeason to the new version:

<code>HasUnmonitoredSeason is True</code> and <code>HasUnmonitoredSeason is not False </code> becomes <code>SeasonsMonitored Less than 100</code>
<code>HasUnmonitoredSeason is False</code> and <code>HasUnmonitoredSeason is not True</code> becomes <code>SeasonsMonitored equal 100</code>

Two other options I came up with were:
Create a HasMonitoredSeason filter

The first is by far the easiest as it is just duplicating the HasUnmonitoredSeason code but changing the != to =. This would also make it unnecessary to delete HasUnminotredSeason as they would not overlap entirely, but some logic would be able to do be done with either

Create a SeasonMonitorState filter with options of All Seasons Monitored, Seasons Mixed, and All Seasons Unmonitored

The second works well, and might be easier to understand than my preferred solution but doesn't offer as much flexibility

### Anything else?

Preferred outcome
image

Edited because I messed up the formatting

jbstark:
#### Description
Replaced hasUnmonitoredSeason with seasonsMonitoredStatus filter with options of All, Partial, and None. This replacement allows for all logic found in hasUnmonitoredSeason, as well as new options. 

Filter options under hasUnmonitoredSeason

- All monitored
- Not all monitored

Filter options under seasonsMonitoredStatus

- All monitored
- Not all monitored
- some seasons monitored, some unmonitored
- Not all unmonitored
- All unmonitored
- All monitored or unmonitored

This makes it easier to track what shows have been "completed". By unmonitoring seasons you can easily filter to shows that don't need any more downloads, or are still in progress without having to unmonitor the show in the case of future seasons.

image

#### Database Migration
YES - 210

* Closes #6896


src/NzbDrone.Core/Datastore/Migration/210_add_monitored_seasons_filter.cs
```diff
@@ -0,0 +1,78 @@
...
+            Execute.WithConnection(ChangeHasUnmonitoredSeason);
+        }
+
+        private void ChangeHasUnmonitoredSeason(IDbConnection conn, IDbTransaction tran)
+        {
+            var updated = new List<object>();
+            using (var getUnmonitoredSeasonFilter = conn.CreateCommand())
+            {
+                getUnmonitoredSeasonFilter.Transaction = tran;
+                getUnmonitoredSeasonFilter.CommandText = "SELECT \"Id\", \"Filters\" FROM \"CustomFilters\" WHERE EXISTS (SELECT 1 FROM json_each(\"Filters\") AS EACH WHERE json_extract(EACH.value, '$.key') = 'hasUnmonitoredSeason')";
```
markus101, line 26: For Postgres, the syntax is different.

For SQLite, `json_extract` is not guaranteed to exist in all versions that Sonarr currently functions on because it wasn't included by default until 3.38, instead perform the JSON operations within C#.

frontend/src/Store/Actions/seriesActions.js
```diff
@@ -194,20 +194,30 @@ export const filterPredicates = {
...
+      }
+      return acc;
+    }, { monitoredCount: 0, unmonitoredCount: 0 });
+
+    let seasonsMonitoredStatus = 'partial';
+    if (monitoredCount === 0) {
+      seasonsMonitoredStatus = 'none';
+    } else if (unmonitoredCount === 0) {
+      seasonsMonitoredStatus = 'all';
+    }
```
markus101, line 218:
```suggestion
    let seasonsMonitoredStatus = 'partial';

    if (monitoredCount === 0) {
      seasonsMonitoredStatus = 'none';
    } else if (unmonitoredCount === 0) {
      seasonsMonitoredStatus = 'all';
    }
```

frontend/src/Store/Actions/seriesActions.js
```diff
@@ -194,20 +194,30 @@ export const filterPredicates = {
...
+    const { monitoredCount, unmonitoredCount } = seasons.reduce((acc, { seasonNumber, monitored }) => {
+      if (seasonNumber <= 0) {
+        return acc;
+      }
+      if (monitored) {
+        acc.monitoredCount++;
+      } else {
+        acc.unmonitoredCount++;
+      }
+      return acc;
```
markus101, line 210:
```suggestion
      if (seasonNumber <= 0) {
        return acc;
      }

      if (monitored) {
        acc.monitoredCount++;
      } else {
        acc.unmonitoredCount++;
      }

      return acc;
```

markus101 (changes requested):
Looking good, haven't tested yet, but a couple nits on the UI formatting and we'll want to avoid `extract_json` for the time being.

src/NzbDrone.Core/Datastore/Migration/210_add_monitored_seasons_filter.cs
jbstark: I just removed the extra portion and made it SELECT * FROM CustomFilters
Dont really think the limiting is needed as all filters will need to be looped over either way, and it is already being checked in the for each loop.

src/NzbDrone.Core/Datastore/Migration/210_add_monitored_seasons_filter.cs
```diff
@@ -0,0 +1,76 @@
...
+            Execute.WithConnection(ChangeHasUnmonitoredSeason);
+        }
+
+        private void ChangeHasUnmonitoredSeason(IDbConnection conn, IDbTransaction tran)
+        {
+            var updated = new List<object>();
+            using (var getUnmonitoredSeasonFilter = conn.CreateCommand())
+            {
+                getUnmonitoredSeasonFilter.Transaction = tran;
+                getUnmonitoredSeasonFilter.CommandText = "SELECT \"Id\", \"Filters\" FROM \"CustomFilters\"";
```
markus101, line 25: No need to process Custom Filters for things that won't have this particular filter:

```suggestion
                getUnmonitoredSeasonFilter.CommandText = "SELECT \"Id\", \"Filters\" FROM \"CustomFilters\" WHERE \"Type\" = 'series'";
```

src/NzbDrone.Core/Datastore/Migration/210_add_monitored_seasons_filter.cs
```diff
@@ -0,0 +1,76 @@
...
+                            Filters = filters.ToJson(),
+                            Id = id
+                        });
+                    }
+                }
+            }
+
+            var updateSql = "UPDATE \"CustomFilters\" SET \"Filters\" = @Filters WHERE \"Id\" = @Id";
+            conn.Execute(updateSql, updated, transaction: tran);
+        }
```
markus101, line 74: Move this outside the reader block

```suggestion
        }

        var updateSql = "UPDATE \"CustomFilters\" SET \"Filters\" = @Filters WHERE \"Id\" = @Id";
        conn.Execute(updateSql, updated, transaction: tran);
```

markus101 (approved):
Thanks for the fixups, just a couple minor suggestions.

src/NzbDrone.Core/Datastore/Migration/210_add_monitored_seasons_filter.cs
```diff
@@ -0,0 +1,76 @@
...
+                            Filters = filters.ToJson(),
+                            Id = id
+                        });
+                    }
+                }
+            }
+        }
+
+        var updateSql = "UPDATE \"CustomFilters\" SET \"Filters\" = @Filters WHERE \"Id\" = @Id";
+        conn.Execute(updateSql, updated, transaction: tran);
```
markus101, line 74: Whoops, misread this

```suggestion

            var updateSql = "UPDATE \"CustomFilters\" SET \"Filters\" = @Filters WHERE \"Id\" = @Id";
            conn.Execute(updateSql, updated, transaction: tran);
        }
```

PR #6941, opened 2024-07-07
New: Custom colon replacement option

### LINKED ISSUE #6898: Custom colon replacement option

No problem. Just a request.

Currently the only options to handle colons are to delete them or replace them with hyphens. I was curious if it could be considered to replace the standard Colon (Punctuation) symbol with the Colon (Letter) symbol? I know this may sound odd to anyone unfamiliar but the standard colon symbol we have bound to our key on our keyboard is the Colon (Punctuation) symbol however, there exists another symbol for the Colon which is recognized by filesystems as a letter. Windows and MacOS which block the punctuation version in file names, allow the letter version in filenames. I'd like to see if it would be possible to allow the option to replace the punctuation colon with the letter colon. If this is not an option that's fine. Just putting it out there.

Alternatives considered? Just the good old trusty " - ". lol

### Anything else?

For those interested in knowing more here are some links about this:
https://en.wikipedia.org/wiki/Colon_(letter)
https://en.wikipedia.org/wiki/Colon_(punctuation)
https://en.wikipedia.org/wiki/Filename

A lot of web apps and search engines allow you to use these interchangeably. Plex being one of them. You can actually search in Plex using the Colon (Letter) or Colon (Punctuation) in the movie or show title and it will find the same results. Plex will scan movie files with Colon (Letter)'s in the names and recognize them just fine.

At the same time, you could even implement this interchangeability in to Sonarr as well. Right now using the search feature in Sonarr with the Colon (Letter) does not work in movie titles.... But it could... I don't know enough about the complexities of how everything is handled in the background of Sonarr so I don't know if implementing the ability to use them interchangeably is a easy task or not. I'm assuming there's already a function that exists to validate and clean up data from input fields. This could just be updated to swap the Colon (Letter) with the Colon (Punctuation) when searching or querying indexers etc... This way it doesn't matter what the user enters, it will still return the same results.

If you want to try for yourself,
This is a Colon (Letter) - ꞉
This is a Colon (Punctuation) - : 
While they look identical, they are not. Give it a try. Copy the first one and paste it in a file name. You'll see that it allows it and looks identical to the standard colon we typically use.

As far is being concerned with how this could impact any code or things like that, it should have no impact on this at all. No coding languages should misinterpret this as the punctuation version and shouldn't even need to be escaped. While we see them as identical. Coding languages see them as totally separate characters. 

I currently do this in FileBot today. I replace the regular colon with the Colon (Letter) and it works just fine. Plex sees the file and associates it with no issues. I know nothing about the impact this has in Jellyfin though because I don't use Jellyfin.

markus101:
#### Description

Option to use a custom character for colon replacement, which allows the letter colon or similar to be used.

#### Database Migration
YES - 211

* Closes #6898


davidnewhall:
The docs say this field is nullable, but if I make a PUT to this endpoint without a `customColonReplacementFormat` field in the json payload, I get a 409 error that says the field failed the NOT NULL constraint. 
Screenshot 2024-07-16 at 4 01 22 PM

Is this an unintentional bug, or is this field supposed to be required?

markus101:
Looks like every string in that resource is nullable, but that will fail validation because it's wrong. My best guess because every string could be null in C# it makes it that way and we'd need to find a way to make that not the case. I've opened up [aasda ](https://github.com/Sonarr/Sonarr/pull/6978) to resolve this from the API side, but the documentation will still incorrectly say it's nullable, but that's a larger issue to resolve.

PR #6943, opened 2024-07-08
Fixed: add non-numeric codec tags to SourceRegex in QualityParser

martylukyy:
#### Description
This PR improves the SourceRegex in QualityParser.cs to also account for non-numeric codec tags such as AVC or HEVC  
which would previously cause the QualityParser to fallback to HDTV, due to the regex not accounting for those tags.

ref https://discord.com/channels/383686866005917708/383688189941907459/1259748157345107978

### DIFF
--- a/src/NzbDrone.Core.Test/ParserTests/QualityParserFixture.cs
+++ b/src/NzbDrone.Core.Test/ParserTests/QualityParserFixture.cs
@@ -254,6 +254,7 @@ namespace NzbDrone.Core.Test.ParserTests
         [TestCase("Series Title Season 2 (WEB 1080p HEVC Opus) [Netaro]", false)]
         [TestCase("Series.Title.S01E01.Erste.Begegnungen.German.DD51.Synced.DL.1080p.HBOMaxHD.AVC-TVS", false)]
         [TestCase("Series.Title.S01E05.Tavora.greift.an.German.DL.1080p.DisneyHD.h264-4SF", false)]
+        [TestCase("Series.Title.S02E04.German.Dubbed.DL.AAC.1080p.WEB.AVC-GROUP", false)]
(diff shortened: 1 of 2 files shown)


src/NzbDrone.Core.Test/ParserTests/QualityParserFixture.cs
```diff
@@ -279,6 +279,7 @@ public void should_parse_webrip1080p_quality(string title, bool proper)
         [TestCase("[HorribleSubs] Series Title! S01 [Web][MKV][h264][2160p][AAC 2.0][Softsubs (HorribleSubs)]", false)]
         [TestCase("Series Title S02 2013 WEB-DL 4k H265 AAC 2Audio-HDSWEB", false)]
         [TestCase("Series.Title.S02E02.This.Year.Will.Be.Different.2160p.WEB.H.265", false)]
+        [TestCase("Series.Title.S02E04.German.Dubbed.DL.AAC.2160p.DV.HDR.WEB.HEVC-GROUP", false)]
```
markus101, line 282: Do you have an example for AVC as well? If not then I think just HEVC makes sense.

martylukyy: I currently do not, but I will check if i can find an AVC example aswell!

markus101: That'd be great, just want to eliminate a potential regression in the future if this regex changes and all tests pass, but it breaks something for an untested case.

martylukyy: Found and added a AVC test example!

markus101:
Thanks for the fix!

PR #6948, opened 2024-07-09
New: Match only custom formats with score

mynameisbogdan:
#### Description
This would be useful when you don't desire specific custom formats to be matched, for example anime related to be matched for standard shows and vice-versa.

Empty value or `null` is needed for the custom format to be skipped.

### DIFF
--- a/frontend/src/Settings/Profiles/Quality/QualityProfileFormatItem.js
+++ b/frontend/src/Settings/Profiles/Quality/QualityProfileFormatItem.js
@@ -56,13 +56,13 @@ class QualityProfileFormatItem extends Component {
 QualityProfileFormatItem.propTypes = {
   formatId: PropTypes.number.isRequired,
   name: PropTypes.string.isRequired,
-  score: PropTypes.number,
+  score: PropTypes.number.isRequired,
   onScoreChange: PropTypes.func
 };
 
 QualityProfileFormatItem.defaultProps = {
   // To handle the case score is deleted during edit
-  score: null
+  score: 0
 };
 
 export default QualityProfileFormatItem;
--- a/frontend/src/Settings/Profiles/Quality/QualityProfileFormatItems.js
+++ b/frontend/src/Settings/Profiles/Quality/QualityProfileFormatItems.js
@@ -18,14 +18,6 @@ function calcOrder(profileFormatItems) {
 
   return [...profileFormatItems].sort((a, b) => {
     if (b.score !== a.score) {
-      if (a.score === null) {
-        return 1;
-      }
-
-      if (b.score === null) {
-        return -1;
-      }
-
       return b.score - a.score;
     }
 
--- a/src/NzbDrone.Core/CustomFormats/CustomFormatCalculationService.cs
+++ b/src/NzbDrone.Core/CustomFormats/CustomFormatCalculationService.cs
@@ -149,7 +149,7 @@ namespace NzbDrone.Core.CustomFormats
             var matches = new List<CustomFormat>();
 
             var profileFormatItems = input.Series?.QualityProfile?.Value?.FormatItems
-                .Where(f => f.Score.HasValue)
+                .Where(f => f.Enabled)
                 .Select(f => f.Format)
                 .ToList();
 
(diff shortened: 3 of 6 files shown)


rcdailey:
From a design standpoint, I think it makes sense to question why all CFs must be directly associated with all profiles. Instead of turning null into a functional value, why not design profiles so that users can configure which CFs are included in profiles? I realize this is probably a much bigger change, but feels more correct.

To be clear, one observable difference between my idea and the solution proposed here is that CFs not associated with a profile won't even be returned in the `formatItems` array, whereas with this solution it seems they will still be returned by the API, but with a value of `null` instead of `0`.

rcdailey:
Thanks for getting back to me so quickly.

I was trying to approach this in a way unspecific to any particular tooling, since obviously I'll have a bias toward Recyclarr. I wanted to focus on what I feel is "correct" versus "quick and easy" with only consideration to Sonarr itself. I figured that approach would be more fair to you and also let us focus more on the merits of the change itself. However it's hard to ignore third party tooling since they will ultimately be impacted.

In general, as an engineer, I've always viewed giving `null` a functional meaning or behavior as a code smell. There's just a lot of situations that make using null for anything but basically "optional" really challenging and (importantly) confusing. From a tooling perspective, I believe there's a lot more to it than just making a DTO have a nullable property -- semantics and behavior have changed. I'm not saying it would be a *lot* of work (I haven't given this deep thought yet), but it's confidently more than is being claimed with regards to the nullable int.

> recyclarr to always apply a non-nullable value for custom formats found in the config, but null for the rest.

Could you go into more detail here? Are you basically saying that if Recyclarr "touches" a CF score, either due to configuration the user provides to Recyclarr or something that comes from the guide, that the score is set to something non-null, but CFs that are not touched, they get null values?

If I have that right, one reason that won't work is because (at least for Recyclarr), we don't want to trample on manual changes a user might have made. If I were to set "untouched" CF scores to null, we potentially overwrite values the user intended to not be touched. But I think the logic here wouldn't change; if the user didn't tell Recyclarr to touch something, it won't touch it. Doesn't matter if it is `0` or `null`.

I realize you probably aren't interested in design feedback since 1) you've already implemented the change and 2) you've made it clear you aren't interested in big changes. My main concern isn't with what changes in Recyclarr, but with the chosen solution. With that, I'll list the main points of concern for me:

- I'm opposed to "null" having specific business rules tied to it (in this case, a score of null representing an optional format item). 
- I'm opposed to choosing simpler changes over correct changes (by your own admission, if I understood you correctly).

I have no issues with what you're trying to do -- that is, making CFs have no impact in both search behavior *and* naming. I think that's a fantastic idea and allows people to get one step closer to maybe cutting down the number of separate Sonarr instances they have.

However, there's several aspects of the current design of CFs that I think should be revisited. I think there's a better representation of CFs as they relate to profiles. Again, if we look at this from the perspective of the API, it's poor design (in my view) to send data with a profile that isn't wanted for that profile. From a database perspective, why would I join a profile with custom formats that have nothing to do with it? And bringing it all together: A user experience that asks a user to type in arbitrary values to make CFs impact quality profiles differently just feels "bad".

If it were me, I would do the additional legwork for a better technical design and ultimately a better user experience. We can talk through what that looks like, like all things in software, and probably come up with something a lot more appealing.

I've rambled enough. I'm not trying to get on your bad side or insult your efforts, by the way. I just wanted to be frank about my thoughts on what's being done here. I've been doing software engineering professionally for over 20 years and I've seen too many instances of shortcuts coming back to bite later; but I also recognize this is an open source project and people aren't paid to work on this. So I can't be as forceful, but I at least want my opinion on the record :-)

Thanks for taking the time to hear me out and I hope something I've said has been helpful.

markus101:
Making the scores nullable does make things quick to implement, but I worry the UX side of things will also suffer and users will struggle to understand the difference between `null` and `0` when configuring and it'll turn into one of those you need to know to know situations.

I don't have a lot of knowledge about why profiles were done as they've been done in regards to CFs, but as-is they virtually necessitate having multiple instances if you want to similar CFs that differ slightly between Anime and non-Anime series, so finding a solution we're all happy with would be great.

> why not design profiles so that users can configure which CFs are included in profiles

This is pretty much where I landed as well, this would give us the flexibility to make including/excluding a lot more obvious to users. There's some potential for some confusion on the API side when things suddenly aren't all included, but it doesn't break backwards compatibility.

I lean towards making this a v5 change (ideally at release) so we get the shock factor of that out of the way in a major upgrade (and very likely, a API version bump), plus we have some things like quality sizes and .net 8 that really need a major version bump.

PR #6949, opened 2024-07-09
Fixed: {Episode CleanTitle} Does Not Remove Apostrophe, Backtick and Most Contractions

diamondpete:
#### Description
Updating the regex for removing characters to the proposed allows episode titles to be more thoroughly cleaned.

When using {Episode CleanTitle}, add the backtick and apostrophe to the regex for contraction cleaning when they are used instead of a single quote. As they are now, both the backtick and the apostrophe are being used in the filename, even though {Episode CleanTitle} option is set.

When using {Episode CleanTitle} only quotes from contractions that end in s or m (i.e. world's > worlds, i'm > im, etc.) are being cleaned. This adds contractions that end in t, ve, re, ll or d (i.e. don't > dont, i've > ive, they're > theyre, i'll > ill, i'd > id, etc.).


markus101 (changes requested):
Please add some test cases that validate these changes.

diamondpete:
Added Test Cases based on CleanTitle in the FileNameBuilder.

markus101 (approved):
Thanks!

PR #6952, opened 2024-07-10
fix: Remove unnecessary tabs in RemoveFileExtension method

eagnoor:
#### Description
Removed unnecessary tabs to improve code style.

### DIFF
--- a/src/NzbDrone.Core/Parser/Parser.cs
+++ b/src/NzbDrone.Core/Parser/Parser.cs
@@ -946,6 +946,7 @@ namespace NzbDrone.Core.Parser
                 {
                     return string.Empty;
                 }
+
                 return m.Value;
             });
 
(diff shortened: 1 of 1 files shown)


src/NzbDrone.Core/Parser/Parser.cs
```diff
@@ -940,15 +940,14 @@ public static string ParseReleaseGroup(string title)
...
-                    if (MediaFiles.MediaFileExtensions.Extensions.Contains(extension) || new[] { ".par2", ".nzb" }.Contains(extension))
-                    {
-                        return string.Empty;
-                    }
-
-                    return m.Value;
-                });
+                    return string.Empty;
+                }
+                return m.Value;
```
markus101, line 949: Keep the new line above the return to separate them visually.

```suggestion

                return m.Value;
```

PR #6978, opened 2024-07-17
Set default value for CustomColonReplacementFormat if not provided

markus101:
#### Description

Prevents clients that aren't sending this value as null from have an issue saving to the DB to maintain backwards compatibility, eventually we'll probably want to update the API docs to correctly indicate that some fields are not nullable.

https://stackoverflow.com/a/70365325


davidnewhall:
Really appreciate this quick fix. Thank you!

PR #6979, opened 2024-07-17
Fixed: Assume category path from qBittorent starting with '//' is a Windows path

markus101:
#### Description

For category paths qbit returns UNC paths with forward slashes instead of back slashes, this will convert those to backslashes so they can be handled properly by Sonarr. Anyone that incorrectly puts `//` at the start of their *nix path will need to fix them and should feel shame.

* Closes Radarr/Radarr#10162

### DIFF
--- a/src/NzbDrone.Core/Download/Clients/QBittorrent/QBittorrent.cs
+++ b/src/NzbDrone.Core/Download/Clients/QBittorrent/QBittorrent.cs
@@ -381,6 +381,7 @@ namespace NzbDrone.Core.Download.Clients.QBittorrent
 
                     if (savePath.StartsWith("//"))
                     {
+                        _logger.Trace("Replacing slashes in probable Windows UNC path '{0}'. If this is not meant to be a Windows UNC path fix your qBittorrent configuration", savePath);
                         savePath = savePath.Replace('/', '\\');
                     }
 
--- a/src/NzbDrone.Core/Download/Clients/QBittorrent/QBittorrent.cs
+++ b/src/NzbDrone.Core/Download/Clients/QBittorrent/QBittorrent.cs
@@ -381,7 +381,7 @@ namespace NzbDrone.Core.Download.Clients.QBittorrent
 
                     if (savePath.StartsWith("//"))
                     {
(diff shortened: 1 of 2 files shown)


markus101:
It's called in 5 different checks and not cached, so could be a little annoying, but added.

src/NzbDrone.Core/Download/Clients/QBittorrent/QBittorrent.cs
```diff
@@ -381,6 +381,7 @@ public override DownloadClientInfo GetStatus()
 
                     if (savePath.StartsWith("//"))
                     {
+                        _logger.Trace("Replacing slashes in probable Windows UNC path '{0}'. If this is not meant to be a Windows UNC path fix your qBittorrent configuration", savePath);
```
bakerboy448, line 384: User hat: fix how?

src/NzbDrone.Core/Download/Clients/QBittorrent/QBittorrent.cs
```diff
@@ -377,7 +377,15 @@ public override DownloadClientInfo GetStatus()
             {
                 if (Proxy.GetLabels(Settings).TryGetValue(Settings.TvCategory, out var label) && label.SavePath.IsNotNullOrWhiteSpace())
                 {
-                    var labelDir = new OsPath(label.SavePath);
+                    var savePath = label.SavePath;
+
+                    if (savePath.StartsWith("//"))
+                    {
+                        _logger.Trace("Replacing slashes in probable Windows UNC path '{0}'. If this is not meant to be a Windows UNC path fix your qBittorrent configuration", savePath);
```
bakerboy448, line 384:
```suggestion
                        _logger.Trace("Replacing double back slashes in probable Windows UNC path '{0}'. If this is not meant to be a Windows UNC path fix your qBittorrent configuration and don't use // for paths on non-Windows systems in Qbit", savePath);
```

wordy but can likely tweak

markus101: Wrong slash, but definitely too wordy regardless.

markus101: I updated it to point them at the Save Path for the category, if they can't get there from them then there isn't much we can do to spell it out much more clearly.

PR #6981, opened 2024-07-17
Fixed: Display tag list when sort by tags on series Posters

mynameisbogdan:
#### Description
As per title, and moving tag list after next airing.

### DIFF
--- a/frontend/src/Series/Index/Posters/SeriesIndexPoster.tsx
+++ b/frontend/src/Series/Index/Posters/SeriesIndexPoster.tsx
@@ -211,14 +211,6 @@ function SeriesIndexPoster(props: SeriesIndexPosterProps) {
         </div>
       ) : null}
 
-      {showTags && tags.length ? (
-        <div className={styles.tags}>
-          <div className={styles.tagsList}>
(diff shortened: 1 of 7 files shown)


frontend/src/Series/Index/Posters/SeriesIndexPoster.tsx
```diff
@@ -62,7 +62,7 @@ function SeriesIndexPoster(props: SeriesIndexPosterProps) {
     added,
     statistics = {} as Statistics,
     images,
-    tags,
+    tags = [],
```
markus101, line 65: No need to set a default, as they'll never be undefined per the typings.

frontend/src/Series/Index/Posters/SeriesIndexPosterInfo.tsx
```diff
@@ -35,11 +38,13 @@ function SeriesIndexPosterInfo(props: SeriesIndexPosterInfoProps) {
     seasonCount,
     path,
     sizeOnDisk,
+    tags = [],
```
markus101, line 41: Same thing here

PR #6983, opened 2024-07-17
NET 8 Upgrade

danielvschoor:
#### Description
Upgrading Sonarr to .NET 8

#### Database Migration
No


danielvschoor:
This PR is the absolute minimum required to update to .NET 8. I've made all the new .NET 8 analyzer warnings "suggestions" instead, as fixing the warnings would include code changes, and I'm not sure if that should be part of this PR.

I'm also not sure whether the Nuget package updates should be part of this PR either, so any guidance is welcome

stevietv:
Have you followed the discussions on the previous attempt at a PR for this? https://github.com/Sonarr/Sonarr/pull/6580

danielvschoor:
> Have you followed the discussions on the previous attempt at a PR for this? #6580

I have, but #6580 seem to be a bigger change than I think is necessary. I also saw that it was closed, which prompted me to create a new PR.





> There's already #6776 👹

Whoops, I missed this one somehow. Will close

PR #6989, opened 2024-07-19
Refresh series with recently aired episodes with TBA titles

mynameisbogdan:
#### Description
Initially was thinking of adding the same command as a schedule task, but seems like I can't add it with custom attributes. So I took this approach instead for `ShouldRefresh` to give a little chance for TBA episodes if possible.

### DIFF
--- a/src/NzbDrone.Core/Tv/ShouldRefreshSeries.cs
+++ b/src/NzbDrone.Core/Tv/ShouldRefreshSeries.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Linq;
 using NLog;
+using NzbDrone.Common.Extensions;
 
 namespace NzbDrone.Core.Tv
 {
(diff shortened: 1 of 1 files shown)


src/NzbDrone.Core/Tv/ShouldRefreshSeries.cs
```diff
@@ -28,6 +29,19 @@ public bool ShouldRefresh(Series series)
...
+            var episodes = _episodeService.GetEpisodeBySeries(series.Id);
+
+            var atLeastOneAiredEpisodeWithoutTitle = episodes.Any(e =>
+                e.SeasonNumber > 0 &&
+                e.AirDateUtc.HasValue && e.AirDateUtc.Value.Before(DateTime.UtcNow) &&
+                e.Title.Equals("TBA", StringComparison.Ordinal));
+
+            if (atLeastOneAiredEpisodeWithoutTitle && series.LastInfoSync < DateTime.UtcNow.AddHours(-1))
+            {
+                _logger.Trace("Series {0} with at least one aired episode with TBA title last updated more than 1 hour ago, should refreshed.", series.Title);
```
stevietv, line 41:
```suggestion
                _logger.Trace("Series {0} with at least one aired episode with TBA title last updated more than 1 hour ago, should refresh.", series.Title);
```

src/NzbDrone.Core/Tv/ShouldRefreshSeries.cs
```diff
@@ -28,6 +29,19 @@ public bool ShouldRefresh(Series series)
...
             }
 
+            var episodes = _episodeService.GetEpisodeBySeries(series.Id);
+
+            var atLeastOneAiredEpisodeWithoutTitle = episodes.Any(e =>
+                e.SeasonNumber > 0 &&
+                e.AirDateUtc.HasValue && e.AirDateUtc.Value.Before(DateTime.UtcNow) &&
+                e.Title.Equals("TBA", StringComparison.Ordinal));
+
+            if (atLeastOneAiredEpisodeWithoutTitle && series.LastInfoSync < DateTime.UtcNow.AddHours(-1))
```
markus101, line 39: I think we could just do the refresh regardless if an aired episode has a TBA title, that way at least the automatic update tries to refresh even if they just refreshed manually. Should be a small window where it was manually refreshed < 1 hour ago anyways.

markus101: Just remove the `series.LastInfoSync < DateTime.UtcNow.AddHours(-1)` and update the log message

PR #7001, opened 2024-07-24
Making status messages for meta downloading consistent 

ManiMatter:
#### Description
Currently, when downloading metadata the status is "queued" and the message is `DownloadClientQbittorrentTorrentStateMetadata`

When the download is forced, though, the status is "downloading" and the isn't a message. 
I think that is inconsitent, and in both cases the status and he message should be `queued` / `DownloadClientQbittorrentTorrentStateMetadata`


ManiMatter:
@markus101 
Thanks for the review and merge. Will your team propagate the same change to the other -arr apps?

PR #7007, opened 2024-07-25
chore: fix some comments

shangchenglumetro:
#### Description

fix some comments

#### Database Migration
YES - ###

* Closes #

### DIFF
--- a/src/NzbDrone.Core/Download/Clients/Deluge/Deluge.cs
+++ b/src/NzbDrone.Core/Download/Clients/Deluge/Deluge.cs
@@ -200,7 +200,7 @@ namespace NzbDrone.Core.Download.Clients.Deluge
 
             if (ignoredCount > 0)
             {
-                _logger.Warn("{0} torrent(s) were ignored becuase they did not have a title, check Deluge and remove any invalid torrents");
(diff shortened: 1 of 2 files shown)


distribution/debian/install.sh
```diff
@@ -59,7 +59,7 @@ app_guid=$(echo "$app_guid" | tr -d ' ')
 app_guid=${app_guid:-media}
 
 echo "This will install [${app^}] to [$bindir] and use [$datadir] for the AppData Directory"
-echo "${app^} will run as the user [$app_uid] and group [$app_guid]. By continuing, you've confirmed that that user and group will have READ and WRITE access to your Media Library and Download Client Completed Download directories"
+echo "${app^} will run as the user [$app_uid] and group [$app_guid]. By continuing, you've confirmed that user and group will have READ and WRITE access to your Media Library and Download Client Completed Download directories"
```
markus101, line 62: while `that that` appears to be a typo it is actually more correct than what it has been replaced with.

markus101 (changes requested):

distribution/debian/install.sh
Thirrian: `that this` ?

markus101:
What about the other change in `install.sh` that was removed when you reverted the other change?

PR #7031, opened 2024-07-30
Fixed: Import with copy from download clients with remove completed disabled

mynameisbogdan:
#### Description
After I got moved a bunch of files from qbit resulting in missing files that need to be redownloaded, this option is nothing less than destructive. If the download client has Remove Completed Downloads disabled, we should not touch the files no matter if the seeding goals are reached.

### DIFF
--- a/src/NzbDrone.Core/MediaFiles/EpisodeImport/ImportApprovedEpisodes.cs
+++ b/src/NzbDrone.Core/MediaFiles/EpisodeImport/ImportApprovedEpisodes.cs
@@ -140,7 +140,7 @@ namespace NzbDrone.Core.MediaFiles.EpisodeImport
                     {
                         default:
                         case ImportMode.Auto:
-                            copyOnly = downloadClientItem != null && !downloadClientItem.CanMoveFiles;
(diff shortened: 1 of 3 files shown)


src/NzbDrone.Core/MediaFiles/EpisodeImport/ImportApprovedEpisodes.cs
```diff
@@ -140,7 +140,7 @@ public List<ImportResult> Import(List<ImportDecision> decisions, bool newDownloa
                     {
                         default:
                         case ImportMode.Auto:
-                            copyOnly = downloadClientItem != null && !downloadClientItem.CanMoveFiles;
```
markus101, line 143: Instead of splitting the logic here and in the download client, `CanMoveFiles` should not be set to true if remove is false

markus101: Actually, we already do a similar check in `DownloadEventHub`, no point tackling changing this, maybe something to consider if we refactor the import process.

PR #7038, opened 2024-07-31
Fixed: Persist indexer flags when manual importing failed imports from queue

mynameisbogdan:
#### Description
Ref https://github.com/Radarr/Radarr/issues/10237

FYI for `ProcessFolder` it's important to note that if you have other files not part off the download, they'll get the same indexer flags as well. This should not happen, but it's good to keep in mind that with some weird download clients like FreeBox who don't have support for single files as output paths (like qbit has since 813f8869202b54048db8d792ee6c3f0d060dbd72), it will process the full downloads folder not just the file in question.


markus101:
Good call out, hopefully the one Freebox user is used to finding the one of many files in the list, but if they start importing a bunch of other random files then they'll figure it out eventually 😄

PR #7058, opened 2024-08-04
Add special character to search criteria

RaZaSB:
Removes special character ’ from series & episode titles when searching.

### DIFF
--- a/src/NzbDrone.Core.Test/IndexerSearchTests/SearchDefinitionFixture.cs
+++ b/src/NzbDrone.Core.Test/IndexerSearchTests/SearchDefinitionFixture.cs
@@ -15,6 +15,10 @@ namespace NzbDrone.Core.Test.IndexerSearchTests
         [TestCase("Franklin & Bash", "Franklin+and+Bash")]
         [TestCase("Chicago P.D.", "Chicago+PD")]
         [TestCase("Kourtney And Khlo\u00E9 Take The Hamptons", "Kourtney+And+Khloe+Take+The+Hamptons")]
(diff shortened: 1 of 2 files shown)


RaZaSB:
Hey, looks like that would fix my problem :+1: I hadn't seen the Lidarr fix before this, so just added to what was already here. Would you prefer I updated it to that?

A side-effect of the Lidarr fix if I understand is that single-quotes get treated as non-words (replaced with `+`) rather than removed, e.g. for the test case added `Betty White’s Off Their Rockers` would become `Betty+White+s+Off+Their+Rockers`. I think indexers would be fine with that, but it'd be a breaking change to the existing single-quote handling for Sonarr. Is that ok?

RaZaSB:
Ahh, I see. I'd mis-read the `NonWord` regex & thought it was including the regular quote, but it's negated so it's excluding it.

RaZaSB:
Pushed change to follow Lidarr's approach. 

Note that this has the side-effect of no longer treating single quotes as special characters, so they will now be sent to indexers, but standardized to `'`. See tests for examples. That's consistent with Lidarr, is that ok here?

RaZaSB:
In that case, I don’t see why we’d copy the change from Lidarr. We’re not standardizing single quotes, but removing them. If we want to remove them, it seems simplest to add the same 4 special quotes to the existing regex. That would also save an additional replace needing to be executed. What do you think?

RaZaSB:
I see, so you were asking for the two additional special single-quotes to be added & no other change?
Hopefully we're on the same page now :+1: pushed the change.

PR #7060, opened 2024-08-05
Improve Title mismatch messaging for searches

bakerboy448:
#### Description
Make it clear what the searched title was when titles don't match and id is used. Improves clarity of alias needed messaging

### DIFF
--- a/src/NzbDrone.Core/Parser/ParsingService.cs
+++ b/src/NzbDrone.Core/Parser/ParsingService.cs
@@ -384,6 +384,7 @@ namespace NzbDrone.Core.Parser
                     return new FindSeriesResult(searchCriteria.Series, SeriesMatchType.Title);
                 }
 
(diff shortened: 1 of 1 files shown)


src/NzbDrone.Core/Parser/ParsingService.cs
```diff
@@ -384,6 +384,7 @@ private FindSeriesResult FindSeries(ParsedEpisodeInfo parsedEpisodeInfo, int tvd
                     return new FindSeriesResult(searchCriteria.Series, SeriesMatchType.Title);
                 }
 
+                _logger.Debug("Could not match Searched Series [{0}] to Parsed Series [{1}], attempting id match", searchCriteria.Series.CleanTitle, parsedEpisodeInfo.SeriesTitle);
```
markus101, line 387: `CleanTitle` is not the only title searched for, nor is it the only title used when parsing, so I don't think this makes things any clearer.

Besides not really adding clarity, I don't think a second log message to convey this is necessary and the mix of title case doesn't fit.

markus101 (changes requested):

PR #7061, opened 2024-08-05
New: Match search releases using IMDb ID

mynameisbogdan:
#### Description
Since Sonarr doesn't have support for `AKA` titles, some results are rejected due to unknown series. Noticed this mostly with anime titles, but seem to have IMDb IDs so I thought we should match them like we do for tvdb and tvrage.

### DIFF
--- a/src/NzbDrone.Core/Indexers/Newznab/NewznabRssParser.cs
+++ b/src/NzbDrone.Core/Indexers/Newznab/NewznabRssParser.cs
@@ -90,6 +90,7 @@ namespace NzbDrone.Core.Indexers.Newznab
 
             releaseInfo.TvdbId = GetTvdbId(item);
             releaseInfo.TvRageId = GetTvRageId(item);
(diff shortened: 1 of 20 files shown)


src/NzbDrone.Core/Indexers/Newznab/NewznabRssParser.cs
```diff
@@ -182,6 +183,18 @@ protected virtual int GetTvRageId(XElement item)
             return 0;
         }
 
+        protected virtual string GetImdbId(XElement item)
+        {
+            var imdbIdString = TryGetNewznabAttribute(item, "imdb");
```
markus101, line 188: Is this actually part of the Newznab spec for TV shows?

markus101: If we rejected everything that wasn't in their spec we'd be screwed.

PR #7065, opened 2024-08-05
Correcting download status when fetching metadata

ManiMatter:
When qBit is **downloading** metadata, the corresponding queue item download status should be '**`Downloading`**', and not `Queued` as it is currently implemented.

qBit queued-status would - to my understanding - refer to a situation where qBit is waiting for a download slot to become available, which is currently occupied by other downloads. This is not the case here and quite the opposite is happening:the downloading (of metadata) is in progress, and thus I'd suggest `Queued` is incorrect and I propose to change it to `Downloading` with this PR.


markus101:
Until the client is actually downloading content of the torrent client then it's status should not be downloading it should be queued, regardless of what qbit or another client is doing on their side.

ManiMatter:
I think I understand. Queued in this context is an arr status, not a torrent client status and expresses that a queue item is in pre-downloading  stage. Whether that is because of metadl or queuing or some other reason is irrelevant.

happy to discard this PR

PR #7069, opened 2024-08-07
Refactoring: Moved First run components to Typescript function components

Treycos:
#### Description
Refactoring of the 3 UI components of `FirstRun` to modern standards
- AuthenticationRequiredModal
- AuthenticationRequiredModalContent
- AuthenticationRequiredModalContentConnector

No behavior changes expected

The diff might be unreadable since the file extension was changed


frontend/src/FirstRun/AuthenticationRequiredModalContentConnector.tsx
```diff
@@ -0,0 +1,43 @@
...
+    return () => {
+      dispatch(clearPendingChanges());
+    };
+  }, [dispatch]);
+
+  return (
+    <AuthenticationRequiredModalContent
+      dispatchFetchStatus={() => dispatch(fetchStatus())}
+      onInputChange={(args: any) => {
+        dispatch((setGeneralSettingsValue as any)(args));
```
Treycos, line 35: It seems that this dispatcher wasn't called before  
It now triggers a call to the redux devtools on each value change

The typing given by `react-actions` doesn't match the doc usage

frontend/.eslintrc.js
```diff
@@ -403,7 +403,7 @@ module.exports = {
             eventHandlerPropPrefix: 'on'
           }
         ],
-        'react/jsx-no-bind': ['error', { ignoreRefs: true }],
+        'react/jsx-no-bind': 'off',
```
Treycos, line 406: This rule is somewhat inefficient with functional component

Respecting it would mean using `useCallback` everywhere

frontend/src/FirstRun/AuthenticationRequiredModal.tsx
```diff
@@ -8,27 +7,21 @@ function onModalClose() {
...
     <Modal
       size={sizes.MEDIUM}
       isOpen={isOpen}
       closeOnBackgroundClick={false}
       onModalClose={onModalClose}
     >
-      <AuthenticationRequiredModalContentConnector
-        onModalClose={onModalClose}
-      />
+      <AuthenticationRequiredModalContentConnector />
```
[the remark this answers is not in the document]
Treycos: The prop was passed down to `AuthenticationRequiredModalContent` by the connector and never used in it

frontend/src/FirstRun/AuthenticationRequiredModalContentConnector.tsx
```diff
@@ -0,0 +1,43 @@
+import React, { useEffect } from 'react';
```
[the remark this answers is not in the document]
Treycos: I'd recommend doing it in another PR  to avoid adding complexity to this one

This PR is mostly made to have a modern 1:1 adaptation of a few components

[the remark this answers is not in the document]
Treycos: I've moved the behavior of the connector into the modal itself

frontend/.eslintrc.js
markus101: You should be using `useCallback` in those instances though. Otherwise every render the function is recreated.

frontend/src/App/State/AppSectionState.ts
```diff
@@ -53,6 +53,7 @@ export interface AppSectionItemSchemaState<T> {
 export interface AppSectionItemState<T> {
   isFetching: boolean;
   isPopulated: boolean;
+  isSaving: boolean;
```
markus101, line 56: `isSaving` is not present for all sections, this should be added to a more specific type instead.

frontend/src/FirstRun/AuthenticationRequiredModalContent.tsx
```diff
@@ -0,0 +1,188 @@
...
+    dispatch(fetchGeneralSettings());
+  }, [dispatch]);
+
+  useEffect(() => {
+    return () => {
+      dispatch(clearPendingChanges());
+    };
+  }, [dispatch]);
+
+  const onInputChange = (args: any) => {
```
markus101, line 51: Some input change events are already typed, but this should be typed instead of using `any`.

frontend/src/FirstRun/AuthenticationRequiredModalContent.tsx
```diff
@@ -0,0 +1,188 @@
...
+    dispatch((setGeneralSettingsValue as any)(args));
```
markus101, line 52: Does this really need to be asserted to `any`?

frontend/src/FirstRun/AuthenticationRequiredModalContent.tsx
```diff
@@ -0,0 +1,188 @@
...
+    passwordConfirmation,
+  } = settings;
+
+  const authenticationEnabled =
+    authenticationMethod && authenticationMethod.value !== 'none';
+
+  const didMount = useRef(false);
+
+  useEffect(() => {
+    if (!isSaving && didMount.current) {
```
markus101, line 69: I realize this was existing, but this has been solved in more recent changes by having a `wasSaving` value stored with `usePrevious`, which is generally better than checking if it was mounted.

frontend/src/FirstRun/AuthenticationRequiredModalContent.tsx
```diff
@@ -0,0 +1,188 @@
...
+const SECTION = 'general';
+
+const selector = createSettingsSectionSelector(SECTION);
+
+function onModalClose() {
+  // No-op
+}
+
+export default function AuthenticationRequiredModalContent() {
+  const { isPopulated, error, isSaving, settings } = useSelector(selector);
```
markus101, line 38: Does this prevent `useSelector` from reprocessing or can we just use it directly?

```suggestion
  const { isPopulated, error, isSaving, settings } = useSelector(createSettingsSectionSelector(SECTION));
```

frontend/src/FirstRun/AuthenticationRequiredModalContent.tsx
```diff
@@ -0,0 +1,188 @@
...
+
+  useEffect(() => {
+    dispatch(fetchGeneralSettings());
+  }, [dispatch]);
+
+  useEffect(() => {
+    return () => {
+      dispatch(clearPendingChanges());
+    };
+  }, [dispatch]);
```
markus101, line 49: Should be safe to do this as a single `useEffect` hook

```suggestion
  useEffect(() => {
    dispatch(fetchGeneralSettings());
    
    return () => {
      dispatch(clearPendingChanges());
    };
  }, [dispatch]);
```

frontend/src/FirstRun/AuthenticationRequiredModalContent.tsx
```diff
@@ -0,0 +1,188 @@
...
+
+        {!isPopulated && !error ? <LoadingIndicator /> : null}
+      </ModalBody>
+
+      <ModalFooter>
+        <SpinnerButton
+          kind={kinds.PRIMARY}
+          isSpinning={isSaving}
+          isDisabled={!authenticationEnabled}
+          onPress={() => {
```
markus101, line 179: This will recreate the function each time the component renders, whereas `useCallback` will not.

markus101 (changes requested):
Made a quick pass to give some initial feedback.

frontend/src/FirstRun/AuthenticationRequiredModalContent.tsx
Treycos: Ideally, it would inherit the type of `FormInputGroup` `onChange`, but its still in JS for now  

I can create an temporary type in the meantime

frontend/src/FirstRun/AuthenticationRequiredModalContent.tsx
Treycos: It avoids unnecessary computation since the function itself doesn't use state or prop variables, might as well call it outside the component

frontend/src/FirstRun/AuthenticationRequiredModalContent.tsx
Treycos: I'll trust you on that one  
Keep in mind that this is an anti-pattern though

frontend/src/Store/Actions/Settings/general.ts
```diff
@@ -3,6 +3,7 @@ import createFetchHandler from 'Store/Actions/Creators/createFetchHandler';
 import createSaveHandler from 'Store/Actions/Creators/createSaveHandler';
```
Treycos, line 3: The file extension change triggered different eslint formatting rules

frontend/src/typings/FormChange.ts
```diff
@@ -0,0 +1,4 @@
+export type FormChangePaylod = {
+  name: string;
+  value: unknown;
+};
```
Treycos, line 4: Not sure if this type already existed

frontend/src/App/State/AppSectionState.ts
Treycos: Reverted

frontend/src/FirstRun/AuthenticationRequiredModalContent.tsx
Treycos: Ideally, it would inherit the type of FormInputGroup onChange, but its still in JS for now

I can create an temporary type in the meantime

frontend/src/FirstRun/AuthenticationRequiredModal.tsx
```diff
@@ -0,0 +1,27 @@
...
+import { sizes } from 'Helpers/Props';
+import AuthenticationRequiredModalContent from './AuthenticationRequiredModalContent';
+
+function onModalClose() {
+  // No-op
+}
+
+export type AuthenticationRequiredModalProps = {
+  isOpen: boolean;
+};
```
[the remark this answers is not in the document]
Treycos: Is there a reason for the props type to not be accessible ?

frontend/src/typings/FormChange.ts
markus101: I think there is just `CheckInputChanged`.

frontend/src/Store/Actions/Settings/general.ts
markus101: Eventually the plan is to nuke all the redux, so you can drop these changes, we've only been typing the state so we can verify types on things we're consuming.

markus101 (changes requested):
You should be able to drop the changes related to the action handling, like we've done in other places since redux will eventually get replaced.

frontend/src/typings/FormChange.ts
Treycos: Seems to be made for checkboxes/switches  
It only accepts booleans

frontend/src/Store/Actions/Settings/general.ts
Treycos: Does that revert include the file renaming ? that would bring the type used back to `any`

And if the revert is only about the eslint changes, the checkup script would stop passing

I wouldn't mind dropping redux, but that MR is already doing way more than it should

frontend/src/typings/FormChange.ts
markus101: Yeah, just pointing out that's the only one, though this typing should be moved to that file and probably renamed to something like `InputChanged` until we can be type all of them.

frontend/src/Store/Actions/Settings/general.ts
markus101: I didn't realize that using `payloadCreator`  in `setGeneralSettingsValue` lead to the type being changed, which leads to the issue you're seeing. For now I think we go with not typing these actions at all. and just use a expect error to ignore typescript.

frontend/src/FirstRun/AuthenticationRequiredModalContent.tsx
```diff
@@ -0,0 +1,191 @@
...
+    dispatch(fetchGeneralSettings());
+
+    return () => {
+      dispatch(clearPendingChanges());
+    };
+  }, [dispatch]);
+
+  const onInputChange = useCallback(
+    (args: FormChangePaylod) => {
+      dispatch(setGeneralSettingsValue(args));
```
markus101, line 63:
```suggestion
      // @ts-expect-error Actions aren't typed
      dispatch(setGeneralSettingsValue(args));
```

markus101 (changes requested):
Just a few more things to match the existing code style and avoid typing the actions, which we can revert and ignore.

markus101 (approved):
Thanks for the fixups

PR #7073, opened 2024-08-09
New: Bulk import list exclusions removal

mynameisbogdan:


markus101 (commented):
The actions column is looking abnormally wide now

image

markus101 (approved):

PR #7066, opened 2024-08-05
Fix indexer multi-languages for release/push

kephasdev:
#### Description
Use the indexer default languages configuration for multi-languages when a release with `MULTI` is pushed using the API `/release/push`.
The Interactive search and RSS sync seems to be already correct.

I had to change `IndexerBase.cs` to put the regex expression for MULTI in `FileNameBuilder.cs`, not quite sure if it's the right place.

Note : this is my first PR for Sonarr, I tried to look around at the function, tests naming convention to use it. If something is not good enough for a merge, feel free to say it I and will change it.

#### Issues Fixed or Closed by this PR
* Closes #7059

PR #7082, opened 2024-08-12
Fix indexer multi-languages for tracked download

kephasdev:
#### Description
Related to PR https://github.com/Sonarr/Sonarr/pull/7066, the indexer was not set before the `_aggregationService.Augment(...)` for the tracked downloads. This cause incorrect custom format score.

### DIFF
--- a/src/NzbDrone.Core/Download/Aggregation/Aggregators/AggregateLanguages.cs
+++ b/src/NzbDrone.Core/Download/Aggregation/Aggregators/AggregateLanguages.cs
@@ -76,18 +76,17 @@ namespace NzbDrone.Core.Download.Aggregation.Aggregators
                 languages = languages.Except(languagesToRemove).ToList();
             }
 
-            if ((languages.Count == 0 || (languages.Count == 1 && languages.First() == Language.Unknown)) &&
-                releaseInfo?.Title?.IsNotNullOrWhiteSpace() == true &&
-                (releaseInfo is { IndexerId: > 0 } || releaseInfo.Indexer?.IsNotNullOrWhiteSpace() == true))
+            if ((languages.Count == 0 || (languages.Count == 1 && languages.First() == Language.Unknown)) && releaseInfo?.Title?.IsNotNullOrWhiteSpace() == true)
             {
-                IndexerDefinition indexer;
+                IndexerDefinition indexer = null;
+
                 if (releaseInfo is { IndexerId: > 0 })
                 {
                     indexer = _indexerFactory.Get(releaseInfo.IndexerId);
                 }
-                else
+                else if (releaseInfo.Indexer?.IsNotNullOrWhiteSpace() == true)
                 {
-                    indexer = _indexerFactory.All().FirstOrDefault(v => v.Name.EqualsIgnoreCase(releaseInfo.Indexer));
+                    indexer = _indexerFactory.FindByName(releaseInfo.Indexer);
                 }
 
                 if (indexer?.Settings is IIndexerSettings settings && settings.MultiLanguages.Any() && Parser.Parser.HasMultipleLanguages(releaseInfo.Title))
--- a/src/NzbDrone.Core/Download/Aggregation/Aggregators/AggregateLanguages.cs
+++ b/src/NzbDrone.Core/Download/Aggregation/Aggregators/AggregateLanguages.cs
@@ -76,9 +76,19 @@ namespace NzbDrone.Core.Download.Aggregation.Aggregators
                 languages = languages.Except(languagesToRemove).ToList();
             }
 
-            if ((languages.Count == 0 || (languages.Count == 1 && languages.First() == Language.Unknown)) && releaseInfo is { IndexerId: > 0 } && releaseInfo.Title.IsNotNullOrWhiteSpace())
+            if ((languages.Count == 0 || (languages.Count == 1 && languages.First() == Language.Unknown)) &&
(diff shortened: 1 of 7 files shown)


src/NzbDrone.Core/Download/Aggregation/Aggregators/AggregateLanguages.cs
```diff
@@ -76,9 +76,14 @@ public RemoteEpisode Aggregate(RemoteEpisode remoteEpisode)
...
-            if ((languages.Count == 0 || (languages.Count == 1 && languages.First() == Language.Unknown)) && releaseInfo is { IndexerId: > 0 } && releaseInfo.Title.IsNotNullOrWhiteSpace())
+            if ((languages.Count == 0 || (languages.Count == 1 && languages.First() == Language.Unknown)) &&
+                releaseInfo?.Title?.IsNotNullOrWhiteSpace() == true &&
+                (releaseInfo is { IndexerId: > 0 } || releaseInfo.Indexer?.IsNotNullOrWhiteSpace() == true))
             {
-                var indexer = _indexerFactory.Get(releaseInfo.IndexerId);
+                var indexer = _indexerFactory.All()
+                    .FirstOrDefault(v =>
+                        (releaseInfo is { IndexerId: > 0 } && releaseInfo.IndexerId == v.Id) ||
+                        (releaseInfo.Indexer?.IsNotNullOrWhiteSpace() == true && v.Name.EqualsIgnoreCase(releaseInfo.Indexer)));
```
markus101, line 86: With release/push can we end up with a scenario where the ID and the Name are for two different indexers?

src/NzbDrone.Core/Download/Aggregation/Aggregators/AggregateLanguages.cs
```diff
@@ -76,9 +76,14 @@ public RemoteEpisode Aggregate(RemoteEpisode remoteEpisode)
                 languages = languages.Except(languagesToRemove).ToList();
             }
 
-            if ((languages.Count == 0 || (languages.Count == 1 && languages.First() == Language.Unknown)) && releaseInfo is { IndexerId: > 0 } && releaseInfo.Title.IsNotNullOrWhiteSpace())
+            if ((languages.Count == 0 || (languages.Count == 1 && languages.First() == Language.Unknown)) &&
+                releaseInfo?.Title?.IsNotNullOrWhiteSpace() == true &&
+                (releaseInfo is { IndexerId: > 0 } || releaseInfo.Indexer?.IsNotNullOrWhiteSpace() == true))
             {
-                var indexer = _indexerFactory.Get(releaseInfo.IndexerId);
+                var indexer = _indexerFactory.All()
```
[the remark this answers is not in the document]
markus101: Saving just the name there makes sense, but we can also resolve that name to an indexer so anything that needs the indexer will have the indexer.

src/NzbDrone.Core/Download/Aggregation/Aggregators/AggregateLanguages.cs
```diff
@@ -76,9 +76,19 @@ public RemoteEpisode Aggregate(RemoteEpisode remoteEpisode)
                 languages = languages.Except(languagesToRemove).ToList();
             }
 
-            if ((languages.Count == 0 || (languages.Count == 1 && languages.First() == Language.Unknown)) && releaseInfo is { IndexerId: > 0 } && releaseInfo.Title.IsNotNullOrWhiteSpace())
+            if ((languages.Count == 0 || (languages.Count == 1 && languages.First() == Language.Unknown)) &&
+                releaseInfo?.Title?.IsNotNullOrWhiteSpace() == true &&
+                (releaseInfo is { IndexerId: > 0 } || releaseInfo.Indexer?.IsNotNullOrWhiteSpace() == true))
```
markus101, line 81: `(releaseInfo is { IndexerId: > 0 } || releaseInfo.Indexer?.IsNotNullOrWhiteSpace() == true)` these checks can be moved to inside this statement to reduce the duplication.

`indexer?.Settings is IIndexerSettings` below takes care of null checking against `indexer`

src/NzbDrone.Core/Download/Aggregation/Aggregators/AggregateLanguages.cs
```diff
@@ -76,9 +76,19 @@ public RemoteEpisode Aggregate(RemoteEpisode remoteEpisode)
...
             {
-                var indexer = _indexerFactory.Get(releaseInfo.IndexerId);
+                IndexerDefinition indexer;
```
markus101, line 83:
```suggestion
                IndexerDefinition indexer;

```

src/NzbDrone.Core/Download/Aggregation/Aggregators/AggregateLanguages.cs
```diff
@@ -76,9 +76,19 @@ public RemoteEpisode Aggregate(RemoteEpisode remoteEpisode)
...
+                if (releaseInfo is { IndexerId: > 0 })
+                {
+                    indexer = _indexerFactory.Get(releaseInfo.IndexerId);
+                }
+                else
+                {
+                    indexer = _indexerFactory.All().FirstOrDefault(v => v.Name.EqualsIgnoreCase(releaseInfo.Indexer));
```
markus101, line 90: Instead of getting all indexers then filtering by name we should be able to add a lookup in the DB by name, which could be made case sensitive.

markus101 (changes requested):

markus101 (approved):
Thanks for all the fixups, looks great.

PR #7085, opened 2024-08-12
Fixed: Updating existing import list exclusions

mynameisbogdan:
#### Description
Prevent `This exclusion has already been added` errors when editing an existing exclusion.

### DIFF
--- a/src/Sonarr.Api.V3/ImportLists/ImportListExclusionController.cs
+++ b/src/Sonarr.Api.V3/ImportLists/ImportListExclusionController.cs
@@ -20,7 +20,10 @@ namespace Sonarr.Api.V3.ImportLists
         {
             _importListExclusionService = importListExclusionService;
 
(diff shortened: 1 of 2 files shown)


src/Sonarr.Api.V3/ImportLists/ImportListExclusionController.cs
```diff
@@ -20,7 +20,10 @@ public class ImportListExclusionController : RestController<ImportListExclusionR
         {
             _importListExclusionService = importListExclusionService;
 
-            SharedValidator.RuleFor(c => c.TvdbId).NotEmpty().SetValidator(importListExclusionExistsValidator);
+            SharedValidator.RuleFor(c => c.TvdbId).Cascade(CascadeMode.Stop)
```
markus101, line 23: If the `TvdbId` and title are missing won't this will lead to subsequent failures instead of reporting all issues in the initial request?

markus101: Ahh right, only within that property, never mind me. 😄

PR #7091, opened 2024-08-12
Add Last Search Time to /episode API and to the Frontend (Cut-off Unmet & Missing)

ManiMatter:
#### Description
Adds ability to check when user has added/last searched for a given episode.
Reason to chose /episode over /episodeFile is because /episodeFile is not returned until the first file is available.
Since here we deal with items that are in process of being searched, chance is high that no file is there yet, hence /episode rather than /episodeFile

### DIFF
--- a/src/NzbDrone.Core/Localization/Core/en.json
+++ b/src/NzbDrone.Core/Localization/Core/en.json
@@ -1063,6 +1063,7 @@
   "Large": "Large",
   "LastDuration": "Last Duration",
   "LastExecution": "Last Execution",
+  "LastSearched": "Last Searched",
   "LastUsed": "Last Used",
   "LastWriteTime": "Last Write Time",
   "LatestSeason": "Latest Season",
--- a/frontend/src/Episode/Episode.ts
+++ b/frontend/src/Episode/Episode.ts
@@ -9,6 +9,7 @@ interface Episode extends ModelBase {
   episodeNumber: number;
   airDate: string;
   airDateUtc?: string;
+  lastSearchTime?: string;
   runtime: number;
   absoluteEpisodeNumber?: number;
   sceneSeasonNumber?: number;
--- a/frontend/src/Store/Actions/wantedActions.js
+++ b/frontend/src/Store/Actions/wantedActions.js
@@ -51,6 +51,12 @@ export const defaultState = {
         isVisible: true
       },
       {
+        name: 'episodes.lastSearchTime',
+        label: () => translate('LastSearched'),
+        isSortable: true,
+        isVisible: false
+      },
+      {
         name: 'status',
         label: () => translate('Status'),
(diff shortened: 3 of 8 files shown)


ManiMatter:
Thanks for the inputs @mynameisbogdan 
The only piece missing now I think is that when we click "search selected" or "search all" that the "Last Search" column updates. 

It currently does not update until the page is refreshed, which looks a bit weird. 

I think you proposed on discord whether we could add support for episodeUpdated in wanted pages. Would appreciate any pointers how to do that, or if there's a better way to accomplish this.

.vscode/settings.json
```diff
@@ -1,3 +1,4 @@
 {
-  "typescript.tsdk": "node_modules\\typescript\\lib"
+  "typescript.tsdk": "node_modules\\typescript\\lib",
+  "workbench.editor.enablePreviewFromCodeNavigation": true
```
markus101, line 3: This should be reverted, you'll want this to be a personal setting, not a workspace one.

src/NzbDrone.Core/Localization/Core/en.json
```diff
@@ -1064,6 +1064,7 @@
   "LastDuration": "Last Duration",
   "LastExecution": "Last Execution",
   "LastUsed": "Last Used",
+  "LastSearchTime": "Last Search",
```
markus101, line 1067: If this is for width reasons then the key should be renamed as well.

markus101 (changes requested):
Looking good. Does repopulating correctly keep checked items?

.vscode/settings.json
ManiMatter: Apols, will revert

src/NzbDrone.Core/Localization/Core/en.json
ManiMatter: The value is called LastSearchTime consistently on the backend, hence why I kept the key in this way.

On the frontend we show a time if the search is from today, or a date if it‘s before today. I felt that the column name „Last Search Time“ doesn‘t do it justice since when showing dates there is no time.

It‘s a detail and I can totally live with „Last Search Time“ as column header - what do you think?

ManiMatter:
> Looking good. Does repopulating correctly keep checked items?

Yes

src/NzbDrone.Core/Localization/Core/en.json
markus101: Yeah, but this is purely for the translation in the UI, makes sense to show `Last Search` in the UI, but the key for the translation should match in this case.

ManiMatter: Will amend the key 👍

markus101 (approved):
Nice work!

PR #7097, opened 2024-08-14
Refactoring: Label component

Treycos:
#### Description

Upgrades the Label component to Typescript

No behavior changes expected


frontend/src/Components/Label.tsx
```diff
@@ -0,0 +1,33 @@
+import classNames from 'classnames';
+import React, { ComponentProps } from 'react';
+import { kinds, sizes } from 'Helpers/Props';
+import styles from './Label.css';
+
+export type LabelProps = {
+  kind?: Extract<(typeof kinds.all)[number], keyof typeof styles>;
+  size?: Extract<(typeof sizes.all)[number], keyof typeof styles>;
```
Treycos, line 8: Matches types that appear both in the options list and in styles keys

frontend/src/Components/Label.tsx
```diff
@@ -0,0 +1,33 @@
...
+  outline?: boolean;
+} & ComponentProps<'span'>;
```
Treycos, line 10: Passes every native props down with their correct typings

Also removes the need for `children` and `title` prop definitions

frontend/src/Components/Label.tsx
```diff
@@ -0,0 +1,33 @@
+import classNames from 'classnames';
+import React, { ComponentProps } from 'react';
+import { kinds, sizes } from 'Helpers/Props';
+import styles from './Label.css';
+
+export type LabelProps = {
```
Treycos, line 6: Exporting types makes component overrides easier

markus101: An interface works fine as well

```
export interface LabelProps extends ComponentProps<'span'> {
  kind?: Extract<(typeof kinds.all)[number], keyof typeof styles>;
  size?: Extract<(typeof sizes.all)[number], keyof typeof styles>;
  outline?: boolean;
}
```

Though reducing it to only props that we actually expect makes the interface clearer, which I believe is limited to `title`.

frontend/src/Components/Label.tsx
markus101: `children` should be required though, currently it's optional.

markus101 (changes requested):

frontend/src/Components/Label.tsx
Treycos: Be sure to keep in mind that since the interface is exported, any other interface with the same name will be merged

From experience, keeping props to the bare minimum leads to countless other unnecessary PRs do add other props one by one (title, name, id, autofocus.... etc)
Keeping the props open makes the component more flexible and allows it to be independent of the project it is used in

markus101 (approved):

PR #7111, opened 2024-08-17
New: Expose log size limit to API and WebUI

martylukyy:
#### Description
This PR exposes the newly added LogSizeLimit from this commit https://github.com/Sonarr/Sonarr/commit/813965e6a20edef2772d68eaa7646af33028425a to the API and the WebUI.

image

#### Proposals

While fiddling this together i noticed that the log size limit has a hardcoded upper limit of 10 MB.
https://github.com/Sonarr/Sonarr/blob/cd3a1c18ab4f8b644ddea3e6358ec23eb3069526/src/NzbDrone.Core/Configuration/ConfigFileProvider.cs#L245
Is there a specific reason for that limit or could this be subject to change?

### DIFF
--- a/frontend/src/Settings/General/LoggingSettings.js
+++ b/frontend/src/Settings/General/LoggingSettings.js
@@ -30,12 +30,14 @@ const logLevelOptions = [
 
 function LoggingSettings(props) {
   const {
+    advancedSettings,
     settings,
     onInputChange
   } = props;
 
   const {
-    logLevel
+    logLevel,
+    logSizeLimit
   } = settings;
 
   return (
@@ -52,11 +54,30 @@ function LoggingSettings(props) {
           {...logLevel}
         />
       </FormGroup>
+
+      <FormGroup
+        advancedSettings={advancedSettings}
+        isAdvanced={true}
+      >
+        <FormLabel>{translate('LogSizeLimit')}</FormLabel>
+
+        <FormInputGroup
+          type={inputTypes.NUMBER}
+          name="logSizeLimit"
+          min={1}
+          max={10}
+          unit="MB"
+          helpText={translate('LogSizeLimitHelpText')}
+          onChange={onInputChange}
+          {...logSizeLimit}
+        />
+      </FormGroup>
     </FieldSet>
   );
 }
 
 LoggingSettings.propTypes = {
+  advancedSettings: PropTypes.bool.isRequired,
   settings: PropTypes.object.isRequired,
   onInputChange: PropTypes.func.isRequired
 };
--- a/src/NzbDrone.Core/Localization/Core/en.json
+++ b/src/NzbDrone.Core/Localization/Core/en.json
@@ -1097,6 +1097,8 @@
   "LogLevel": "Log Level",
   "LogLevelTraceHelpTextWarning": "Trace logging should only be enabled temporarily",
   "LogOnly": "Log Only",
+  "LogSizeLimit": "Log Size Limit",
+  "LogSizeLimitHelpText": "Maximum log file size in MB before archiving. Default is 1MB.",
   "Logging": "Logging",
   "Logout": "Logout",
   "Logs": "Logs",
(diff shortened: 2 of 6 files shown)


martylukyy:
> You should undo the non-en translations as it's going to break weblate and would require a reset to make it work again.
> 
> Considering we only support values from 1 to 10, we should have validation for this?

Done and done. Thanks for your help. Appreciate it!

Do you have any insights on the consensus to raise the log size limit?

martylukyy:
So more than a few times i found myself looking through my trace logs to debug something only to find out the thing i am looking for is not in the most recent trace log but in the one before.
Funny enough i even needed both logs once to see what was the matter.

I can see your point though and maybe this should be an advanced setting (as bogdan said) to prevent the situation you mentioned.

Personally, i don't see a reason not to raise the hard limit to something bigger if it's within reason and keep the default.
I wouldn't set the hard limit so high that your editor / IDE crashes from loading one file but big enough so i don't need to put all my recent logs files in a folder and CTRL SHIFT F through it with an IDE.
Somewhere from 30 to 50 for the upper limit would be reasonable if you want my opinion.
If you are worried about space constrictions maybe decrease the number of logs kept? (Just brainstorming here)

To add to the "Why do we need this in the UI?" part:
Otherwise i would need to go edit the file via SSH/FTP in case the above mentioned situation happens and it's kinda a pity you need to do this since sonarr has all the things you need to debug from the WebUI.

bakerboy448:
Definitely should be an advanced option and need to expand the description and maybe we should show the total usage based on the selected log level and log file quantity? Or just show total usage for input value denoted by each possibly log level and the log file quantity?

Use case would be as stated not having to comb multiple trace logs to get all relevant pieces, primarily during scans, rss, or searches for debugging

src/NzbDrone.Core/Localization/Core/en.json
```diff
@@ -1097,6 +1097,8 @@
   "LogLevel": "Log Level",
   "LogLevelTraceHelpTextWarning": "Trace logging should only be enabled temporarily",
   "LogOnly": "Log Only",
+  "LogSizeLimit": "Log Size Limit",
+  "LogSizeLimitHelpText": "Sets the size in MB above which log files will be automatically archived.",
```
markus101, line 1101: Maybe something like:
```suggestion
  "LogSizeLimitHelpText": "Maximum log file size in MB before archiving. Default is 1MB",
```

frontend/src/Settings/General/LoggingSettings.js
```diff
@@ -52,6 +53,19 @@ function LoggingSettings(props) {
           {...logLevel}
         />
       </FormGroup>
+      <FormGroup>
+
+        <FormLabel>{translate('LogSizeLimit')}</FormLabel>
+
+        <FormInputGroup
+          type={inputTypes.NUMBER}
+          name="logSizeLimit"
```
markus101, line 62:
```suggestion
          max={10}
          min={1}
          name="logSizeLimit"
```

frontend/src/Settings/General/LoggingSettings.js
```diff
@@ -52,6 +53,19 @@ function LoggingSettings(props) {
           {...logLevel}
         />
       </FormGroup>
+      <FormGroup>
+
+        <FormLabel>{translate('LogSizeLimit')}</FormLabel>
+
```
markus101, line 59:
```suggestion

      <FormGroup>
        <FormLabel>{translate('LogSizeLimit')}</FormLabel>
```

markus101 (changes requested):
Definitely an advanced setting. For now let's stick with the 10 MB limit, in a future version we can consider changing the default/maximum/total number of files.

martylukyy:
Implemented your suggestions.
Need to look into why i cannot make it an advanced setting...
When i set the advancedSettings prop i just disappears 🤔

martylukyy:
Thanks - now i feel stupid 😂 
I was missing the part in GeneralSettings.js 😅

markus101 (approved):
Thanks for the fixes

PR #7113, opened 2024-08-17
Improve context for UI language validation

mynameisbogdan:
#### Description
Switching to `Must` to simplifying the validation for UI language and for more context. 

before
before

after
after


markus101:
IMO, the output after the change is too verbose and there is a bunch of noise in there that doesn't add anything to the error message.

PR #7115, opened 2024-08-18
Add to blocklist without queue removal

ManiMatter:
#### Description
When users download multi-pack series, they want to mark certain files in the download client as "Not Download" and rely on a different torrent for those items. Thus the download has 2 content types:
A. Items that should should finish downloading and then be imported
B. Items for which other torrents should be found

Currently, there is no way to handle this. Either, the user waits till the type A. items have finished, and then starts searching for the others. If then the same torrent is proposed again, the user can blocklist it, and try another one. 

The addition I propose here is to be able to add an ongoing queue item to the blocklist, but without by doing so removing the queue item from the queue.

This would allow a re-search, and for the type B. items a the current torrent would already be ignored (since blocklisted).

The user path I envision is the following:
- User choses to "Keep" queue item
- User choses to "Blocklist" (with or without re-scan) 
Already coded:
image


ManiMatter:
I added this PR already now (as DRAFT), to 
1) get your feedback whether this makes sense
2) to get any pointers on the backend; here concretely I'm stuck with two things. 
a) the keepInQueue seems always to come through as "false" - I think I need to change smth in the openapi.json
b) When keepInQueue is true and we want to blocklist, I am stuck in coding up how we add the item to the blocklist, and then trigger a re-scan if needed (without marking the download as failed)

augustuen:
If I'm understanding you correctly, you see this as being used in cases where a user downloads a season pack, but only wants to import some of the episodes from that pack, and instead grab the rest of the episodes individually? And so they set some of the files from the first pack not download?

I'm not a maintainer, but I doubt this'll be accepted, as my understanding is that that kind of behaviour is bad torrenting and even banned by some private trackers.

>This would allow a re-search, and for the type B. items a the current torrent would already be ignored (since blocklisted).

Except it wouldn't, since the episodes from the first download would still be in the queue, and so would be cause for rejecting any new release, unless that new release is an upgrade on the old one, in which case this change is entirely unnecessary.

markus101:
Third on not seeing the need for this. If a season pack is in queue then every episode in that season will be rejected unless the release you're trying to get exceeds that release in some way, but unless you searched multiple times you wouldn't get multiple items added.

Marking a file in the download client as do not download won't change how Sonarr behaves, it's still always going to see the import as incomplete and require manual intervention.

ManiMatter:
> If I'm understanding you correctly, you see this as being used in cases where a user downloads a season pack, but only wants to import some of the episodes from that pack, and instead grab the rest of the episodes individually? And so they set some of the files from the first pack not download?

I was thinking of using this in cases when some of the episodes don't have 100% availability. Knowing that for those the download might never finish, I would mark them as "Do not download". In the queue, I'm aware that respective episodes would still show up as being downloaded. Then would like to trigger another search for the missing episodes. Here I wanted to use the block list to ensure we would not re-add the same torrent once again, but look for different torrents.

My understanding from the following comment is that marking an episode as "don't download" in the torrent client would mean that once the torrent reaches 100%  and is marked as completed, Sonarr would still not import it (as it is waiting for all files to become available belonging to the same torrent), thus requiring manual intervention. Is that correct? 
> Marking a file in the download client as do not download won't change how Sonarr behaves, it's still always going to see the import as incomplete and require manual intervention.

After marking a specific episode in qbit as "do not download", I would want to re-trigger a search (which I'd do over the entire series, not specific to the episode, since the file I marked as don't download in the torrent client is not yet associated with a specific queue item). During this re-search, as I mentioned above, I want to avoid that the same torrent gets fetched again. Based on the the below feedback I understand that the problem of triggering this search would actually be that I'd get the same torrent once more. Rather, the re-search would not do anything at all, since in the queue the previous torrent is still present, for which Sonarr believes it will receive files and thus won't re-add (since not an upgrade in quality etc). That right? 
> If a season pack is in queue then every episode in that season will be rejected unless the release you're trying to get exceeds that release in some way, but unless you searched multiple times you wouldn't get multiple items added.


In a nutshell, the challenge I'm trying to solve is:
- There is a season pack, some episodes are available, some are not (no peer with 100% availabiltiy)
- I want to fetch whatever I can from that season pack (which has 100% availabiltiy), and get the rest from somewhere else
- If I can avoid it, I'd want to not spend bandwith on the season pack on the items that have less than 100% availabiltiy

Based on above feedbacks, my take is that there's two challenges
- If not downloading the entire season pack, Sonarr won't import the "healthy" files after the torrent has completed
- The re-search for the "unhealthy" items won't find new torrents since Sonarr believes that the season pack will grab these files

You are probably right and the approach I tried here won't work. Any ideas how this challenge could be solved differently?

markus101:
The only way you'll be able to handle this is manually. Sonarr won't know that you don't want specific episodes in the season pack and that it needs to find them elsewhere. At best you could wait for the downloaded episodes to import, and then tell Sonarr to remove the item and blocklist it, but the blocklisting wouldn't really matter since Sonarr wouldn't grab a season pack for less than the full season being needed.

PR #7119, opened 2024-08-19
Refactoring: Icon component Typescript

Treycos:
#### Description

Typescript refactoring of the `Icon `component

No behavior changes expected


frontend/src/InteractiveImport/Interactive/InteractiveImportRow.tsx
```diff
@@ -525,7 +525,7 @@ function InteractiveImportRow(props: InteractiveImportRowProps) {
             <>
               {indexerFlags ? (
                 <Popover
-                  anchor={<Icon name={icons.FLAG} kind={kinds.PRIMARY} />}
```
Treycos, line 528: I seems this kind of icon doesn't exist anymore

frontend/src/System/Tasks/Queued/QueuedTaskRow.tsx
```diff
@@ -41,14 +41,14 @@ function getStatusIconProps(status: string, message: string | undefined) {
         name: icons.CHECK,
         kind: kinds.SUCCESS,
         title: message === 'Completed' ? title : `${title}: ${message}`,
-      };
+      } as const;
```
Treycos, line 44: Typing the function returning this works the same if you prefer this syntax

markus101: Creating a type for what is being returned here would be preferred.

PR #7122, opened 2024-08-19
Shift quality definition limits management to the backend

rcdailey:
This update moves the minimum, maximum, and preferred quality limits to
the backend, accessible via the new `/qualitydefinition/limits`
endpoint. This change improves support for unofficial Sonarr API clients
and enables a more flexible frontend.
  
The Quality Definitions settings page has been significantly refactored
to allow for asynchronous and dynamic loading of these limits.

Additional changes:

* Add .idea directory to gitignore

  For users of the Jetbrains IDEs, the `.idea` directory isn't strictly
  necessary for version control. It's better to ignore it than tie a repo
  to specific tooling.

* Add --output-pathinfo to start & watch scripts

  The motivation for this change is to get React debugging working.

### DIFF
--- a/src/NzbDrone.Core/Qualities/QualityDefinitionLimits.cs
+++ b/src/NzbDrone.Core/Qualities/QualityDefinitionLimits.cs
@@ -1,0 +1,11 @@
+﻿using System.Diagnostics.CodeAnalysis;
+
+namespace NzbDrone.Core.Qualities;
+
+[SuppressMessage("Performance", "CA1822:Mark members as static", Justification =
+    "Serializable properties of a DTO")]
+public record QualityDefinitionLimits
+{
+    public int MinLimit => 0;
+    public int MaxLimit => 1000;
+}
--- a/src/NzbDrone.Core/Qualities/QualityDefinitionLimits.cs
+++ b/src/NzbDrone.Core/Qualities/QualityDefinitionLimits.cs
@@ -1,7 +1,7 @@
-﻿namespace NzbDrone.Core.Qualities;
+namespace NzbDrone.Core.Qualities;
 
 public static class QualityDefinitionLimits
 {
-    public const int MinLimit = 0;
-    public const int MaxLimit = 1000;
+    public const int Min = 0;
+    public const int Max = 1000;
 }
--- a/src/NzbDrone.Core/Qualities/QualityDefinitionLimits.cs
+++ b/src/NzbDrone.Core/Qualities/QualityDefinitionLimits.cs
@@ -1,11 +1,7 @@
-﻿using System.Diagnostics.CodeAnalysis;
+﻿namespace NzbDrone.Core.Qualities;
 
-namespace NzbDrone.Core.Qualities;
-
-[SuppressMessage("Performance", "CA1822:Mark members as static", Justification =
-    "Serializable properties of a DTO")]
-public record QualityDefinitionLimits
+public static class QualityDefinitionLimits
 {
-    public int MinLimit => 0;
-    public int MaxLimit => 1000;
+    public const int MinLimit = 0;
+    public const int MaxLimit = 1000;
 }
--- a/src/Sonarr.Api.V3/Qualities/QualityDefinitionController.cs
+++ b/src/Sonarr.Api.V3/Qualities/QualityDefinitionController.cs
@@ -55,6 +55,12 @@ namespace Sonarr.Api.V3.Qualities
                 .ToResource());
         }
 
+        [HttpGet("limits")]
+        public ActionResult<QualityDefinitionLimits> GetLimits()
+        {
+            return Ok(_qualityDefinitionService.GetLimits());
+        }
+
         [NonAction]
         public void Handle(CommandExecutedEvent message)
         {
--- a/src/Sonarr.Api.V3/Qualities/QualityDefinitionController.cs
+++ b/src/Sonarr.Api.V3/Qualities/QualityDefinitionController.cs
@@ -25,18 +25,12 @@ namespace Sonarr.Api.V3.Qualities
         {
             _qualityDefinitionService = qualityDefinitionService;
 
-            SetupValidation();
-        }
-
-        private void SetupValidation()
-        {
             SharedValidator.RuleFor(c => c)
                 .SetValidator(new QualityDefinitionResourceValidator());
         }
 
         [RestPutById]
-        public ActionResult<QualityDefinitionResource> Update(
-            [FromBody] QualityDefinitionResource resource)
+        public ActionResult<QualityDefinitionResource> Update([FromBody] QualityDefinitionResource resource)
         {
             var model = resource.ToModel();
             _qualityDefinitionService.Update(model);
@@ -70,8 +64,8 @@ namespace Sonarr.Api.V3.Qualities
         public ActionResult<QualityDefinitionLimitsResource> GetLimits()
         {
             return Ok(new QualityDefinitionLimitsResource(
-                QualityDefinitionLimits.MinLimit,
-                QualityDefinitionLimits.MaxLimit));
+                QualityDefinitionLimits.Min,
+                QualityDefinitionLimits.Max));
         }
 
         [NonAction]
--- a/src/Sonarr.Api.V3/Qualities/QualityDefinitionController.cs
+++ b/src/Sonarr.Api.V3/Qualities/QualityDefinitionController.cs
@@ -20,6 +20,15 @@ namespace Sonarr.Api.V3.Qualities
             : base(signalRBroadcaster)
         {
             _qualityDefinitionService = qualityDefinitionService;
+
+            SetupValidation(qualityDefinitionService);
+        }
+
+        private void SetupValidation(IQualityDefinitionService qualityDefinitionService)
+        {
+            var limits = qualityDefinitionService.GetLimits();
+            SharedValidator.RuleFor(c => c)
+                .SetValidator(new QualityDefinitionResourceValidator(limits));
         }
 
         [RestPutById]
--- a/src/Sonarr.Api.V3/Qualities/QualityDefinitionController.cs
+++ b/src/Sonarr.Api.V3/Qualities/QualityDefinitionController.cs
@@ -12,27 +12,31 @@ using Sonarr.Http.REST.Attributes;
 namespace Sonarr.Api.V3.Qualities
 {
     [V3ApiController]
-    public class QualityDefinitionController : RestControllerWithSignalR<QualityDefinitionResource, QualityDefinition>, IHandle<CommandExecutedEvent>
+    public class QualityDefinitionController :
+        RestControllerWithSignalR<QualityDefinitionResource, QualityDefinition>,
+        IHandle<CommandExecutedEvent>
     {
         private readonly IQualityDefinitionService _qualityDefinitionService;
 
-        public QualityDefinitionController(IQualityDefinitionService qualityDefinitionService, IBroadcastSignalRMessage signalRBroadcaster)
+        public QualityDefinitionController(
+            IQualityDefinitionService qualityDefinitionService,
+            IBroadcastSignalRMessage signalRBroadcaster)
             : base(signalRBroadcaster)
         {
             _qualityDefinitionService = qualityDefinitionService;
 
-            SetupValidation(qualityDefinitionService);
+            SetupValidation();
         }
 
-        private void SetupValidation(IQualityDefinitionService qualityDefinitionService)
+        private void SetupValidation()
         {
-            var limits = qualityDefinitionService.GetLimits();
             SharedValidator.RuleFor(c => c)
-                .SetValidator(new QualityDefinitionResourceValidator(limits));
+                .SetValidator(new QualityDefinitionResourceValidator());
         }
 
         [RestPutById]
-        public ActionResult<QualityDefinitionResource> Update([FromBody] QualityDefinitionResource resource)
+        public ActionResult<QualityDefinitionResource> Update(
+            [FromBody] QualityDefinitionResource resource)
         {
             var model = resource.ToModel();
             _qualityDefinitionService.Update(model);
@@ -54,9 +58,7 @@ namespace Sonarr.Api.V3.Qualities
         public object UpdateMany([FromBody] List<QualityDefinitionResource> resource)
         {
             // Read from request
-            var qualityDefinitions = resource
-                                                 .ToModel()
-                                                 .ToList();
+            var qualityDefinitions = resource.ToModel().ToList();
 
             _qualityDefinitionService.UpdateMany(qualityDefinitions);
 
@@ -65,9 +67,11 @@ namespace Sonarr.Api.V3.Qualities
         }
 
         [HttpGet("limits")]
-        public ActionResult<QualityDefinitionLimits> GetLimits()
+        public ActionResult<QualityDefinitionLimitsResource> GetLimits()
         {
-            return Ok(_qualityDefinitionService.GetLimits());
+            return Ok(new QualityDefinitionLimitsResource(
+                QualityDefinitionLimits.MinLimit,
+                QualityDefinitionLimits.MaxLimit));
         }
 
         [NonAction]
--- a/src/Sonarr.Api.V3/Qualities/QualityDefinitionLimitsResource.cs
+++ b/src/Sonarr.Api.V3/Qualities/QualityDefinitionLimitsResource.cs
@@ -1,9 +1,6 @@
-﻿using System.Diagnostics.CodeAnalysis;
-
 namespace Sonarr.Api.V3.Qualities;
 
-[SuppressMessage("StyleCop.CSharp.NamingRules",
-    "SA1313:Parameter names should begin with lower-case letter",
-    Justification =
-        "False positive for record types since params should follow property naming rules")]
-public record QualityDefinitionLimitsResource(int MinLimit, int MaxLimit);
+// SA1313 still applies to records until https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/3181 is available in a release build.
+#pragma warning disable SA1313
+public record QualityDefinitionLimitsResource(int Min, int Max);
+#pragma warning restore SA1313
--- a/src/Sonarr.Api.V3/Qualities/QualityDefinitionLimitsResource.cs
+++ b/src/Sonarr.Api.V3/Qualities/QualityDefinitionLimitsResource.cs
@@ -1,0 +1,9 @@
+﻿using System.Diagnostics.CodeAnalysis;
+
+namespace Sonarr.Api.V3.Qualities;
+
+[SuppressMessage("StyleCop.CSharp.NamingRules",
+    "SA1313:Parameter names should begin with lower-case letter",
+    Justification =
+        "False positive for record types since params should follow property naming rules")]
+public record QualityDefinitionLimitsResource(int MinLimit, int MaxLimit);
--- a/src/Sonarr.Api.V3/Qualities/QualityDefinitionResourceValidator.cs
+++ b/src/Sonarr.Api.V3/Qualities/QualityDefinitionResourceValidator.cs
@@ -7,22 +7,25 @@ public class QualityDefinitionResourceValidator : AbstractValidator<QualityDefin
 {
     public QualityDefinitionResourceValidator()
     {
-        When(c => c.MinSize is not null, () =>
-        {
-            RuleFor(c => c.MinSize)
-                .GreaterThanOrEqualTo(QualityDefinitionLimits.MinLimit)
-                .WithErrorCode("GreaterThanOrEqualTo")
-                .LessThanOrEqualTo(c => c.PreferredSize ?? QualityDefinitionLimits.MaxLimit)
-                .WithErrorCode("LessThanOrEqualTo");
-        });
+        RuleFor(c => c.MinSize)
+            .GreaterThanOrEqualTo(QualityDefinitionLimits.MinLimit)
+            .WithErrorCode("GreaterThanOrEqualTo")
+            .LessThanOrEqualTo(c => c.PreferredSize ?? QualityDefinitionLimits.MaxLimit)
+            .WithErrorCode("LessThanOrEqualTo")
+            .When(c => c.MinSize is not null);
 
-        When(c => c.MaxSize is not null, () =>
-        {
-            RuleFor(c => c.MaxSize)
-                .GreaterThanOrEqualTo(c => c.PreferredSize ?? QualityDefinitionLimits.MinLimit)
-                .WithErrorCode("GreaterThanOrEqualTo")
-                .LessThanOrEqualTo(QualityDefinitionLimits.MaxLimit)
-                .WithErrorCode("LessThanOrEqualTo");
-        });
+        RuleFor(c => c.PreferredSize)
+            .GreaterThanOrEqualTo(c => c.MinSize ?? QualityDefinitionLimits.MinLimit)
+            .WithErrorCode("GreaterThanOrEqualTo")
+            .LessThanOrEqualTo(c => c.MaxSize ?? QualityDefinitionLimits.MaxLimit)
+            .WithErrorCode("LessThanOrEqualTo")
+            .When(c => c.PreferredSize is not null);
+
+        RuleFor(c => c.MaxSize)
+            .GreaterThanOrEqualTo(c => c.PreferredSize ?? QualityDefinitionLimits.MinLimit)
+            .WithErrorCode("GreaterThanOrEqualTo")
+            .LessThanOrEqualTo(QualityDefinitionLimits.MaxLimit)
+            .WithErrorCode("LessThanOrEqualTo")
+            .When(c => c.MaxSize is not null);
     }
 }
--- a/src/Sonarr.Api.V3/Qualities/QualityDefinitionResourceValidator.cs
+++ b/src/Sonarr.Api.V3/Qualities/QualityDefinitionResourceValidator.cs
@@ -1,0 +1,28 @@
+﻿using FluentValidation;
+using NzbDrone.Core.Qualities;
+
+namespace Sonarr.Api.V3.Qualities;
+
+public class QualityDefinitionResourceValidator : AbstractValidator<QualityDefinitionResource>
+{
+    public QualityDefinitionResourceValidator(QualityDefinitionLimits limits)
+    {
+        When(c => c.MinSize is not null, () =>
+        {
+            RuleFor(c => c.MinSize)
+                .GreaterThanOrEqualTo(limits.MinLimit)
+                .WithErrorCode("GreaterThanOrEqualTo")
+                .LessThanOrEqualTo(c => c.PreferredSize ?? limits.MaxLimit)
+                .WithErrorCode("LessThanOrEqualTo");
+        });
+
+        When(c => c.MaxSize is not null, () =>
+        {
+            RuleFor(c => c.MaxSize)
+                .GreaterThanOrEqualTo(c => c.PreferredSize ?? limits.MinLimit)
+                .WithErrorCode("GreaterThanOrEqualTo")
+                .LessThanOrEqualTo(limits.MaxLimit)
+                .WithErrorCode("LessThanOrEqualTo");
+        });
+    }
+}
--- a/src/Sonarr.Api.V3/Qualities/QualityDefinitionResourceValidator.cs
+++ b/src/Sonarr.Api.V3/Qualities/QualityDefinitionResourceValidator.cs
@@ -8,23 +8,23 @@ public class QualityDefinitionResourceValidator : AbstractValidator<QualityDefin
     public QualityDefinitionResourceValidator()
     {
         RuleFor(c => c.MinSize)
-            .GreaterThanOrEqualTo(QualityDefinitionLimits.MinLimit)
+            .GreaterThanOrEqualTo(QualityDefinitionLimits.Min)
             .WithErrorCode("GreaterThanOrEqualTo")
-            .LessThanOrEqualTo(c => c.PreferredSize ?? QualityDefinitionLimits.MaxLimit)
+            .LessThanOrEqualTo(c => c.PreferredSize ?? QualityDefinitionLimits.Max)
             .WithErrorCode("LessThanOrEqualTo")
             .When(c => c.MinSize is not null);
 
         RuleFor(c => c.PreferredSize)
-            .GreaterThanOrEqualTo(c => c.MinSize ?? QualityDefinitionLimits.MinLimit)
+            .GreaterThanOrEqualTo(c => c.MinSize ?? QualityDefinitionLimits.Min)
             .WithErrorCode("GreaterThanOrEqualTo")
-            .LessThanOrEqualTo(c => c.MaxSize ?? QualityDefinitionLimits.MaxLimit)
+            .LessThanOrEqualTo(c => c.MaxSize ?? QualityDefinitionLimits.Max)
             .WithErrorCode("LessThanOrEqualTo")
             .When(c => c.PreferredSize is not null);
 
         RuleFor(c => c.MaxSize)
-            .GreaterThanOrEqualTo(c => c.PreferredSize ?? QualityDefinitionLimits.MinLimit)
+            .GreaterThanOrEqualTo(c => c.PreferredSize ?? QualityDefinitionLimits.Min)
             .WithErrorCode("GreaterThanOrEqualTo")
-            .LessThanOrEqualTo(QualityDefinitionLimits.MaxLimit)
+            .LessThanOrEqualTo(QualityDefinitionLimits.Max)
             .WithErrorCode("LessThanOrEqualTo")
             .When(c => c.MaxSize is not null);
     }
--- a/src/Sonarr.Api.V3/Qualities/QualityDefinitionResourceValidator.cs
+++ b/src/Sonarr.Api.V3/Qualities/QualityDefinitionResourceValidator.cs
@@ -5,23 +5,23 @@ namespace Sonarr.Api.V3.Qualities;
 
 public class QualityDefinitionResourceValidator : AbstractValidator<QualityDefinitionResource>
 {
-    public QualityDefinitionResourceValidator(QualityDefinitionLimits limits)
+    public QualityDefinitionResourceValidator()
     {
         When(c => c.MinSize is not null, () =>
         {
             RuleFor(c => c.MinSize)
-                .GreaterThanOrEqualTo(limits.MinLimit)
+                .GreaterThanOrEqualTo(QualityDefinitionLimits.MinLimit)
                 .WithErrorCode("GreaterThanOrEqualTo")
-                .LessThanOrEqualTo(c => c.PreferredSize ?? limits.MaxLimit)
+                .LessThanOrEqualTo(c => c.PreferredSize ?? QualityDefinitionLimits.MaxLimit)
                 .WithErrorCode("LessThanOrEqualTo");
         });
 
         When(c => c.MaxSize is not null, () =>
         {
             RuleFor(c => c.MaxSize)
-                .GreaterThanOrEqualTo(c => c.PreferredSize ?? limits.MinLimit)
+                .GreaterThanOrEqualTo(c => c.PreferredSize ?? QualityDefinitionLimits.MinLimit)
                 .WithErrorCode("GreaterThanOrEqualTo")
-                .LessThanOrEqualTo(limits.MaxLimit)
+                .LessThanOrEqualTo(QualityDefinitionLimits.MaxLimit)
                 .WithErrorCode("LessThanOrEqualTo");
         });
     }
--- a/.gitignore
+++ b/.gitignore
@@ -162,3 +162,6 @@ src/.idea/
 
 # API doc generation
 .config/
+
+# Ignore Jetbrains IntelliJ Workspace Directories
+.idea/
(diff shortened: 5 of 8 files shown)


frontend/src/Settings/Quality/Definition/QualityDefinitions.js
```diff
@@ -3,6 +3,7 @@ import React, { Component } from 'react';
 import FieldSet from 'Components/FieldSet';
 import PageSectionContent from 'Components/Page/PageSectionContent';
 import translate from 'Utilities/String/translate';
+import { sizeLimitsShape } from '../Shapes/SizeLimitsShape';
```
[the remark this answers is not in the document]
rcdailey: I honestly have been letting WebStorm auto-add and manage the imports. I'll see if that's configurable, but I'm not sure.

I'm not really experienced with React. Is there a reason for using the absolute path in this case?

frontend/src/Store/Actions/Settings/qualityDefinitions.js
```diff
@@ -44,15 +44,45 @@ export default {
...
 
   //
   // Action Handlers
 
   actionHandlers: {
     [FETCH_QUALITY_DEFINITIONS]: createFetchHandler(section, '/qualitydefinition'),
-    [SAVE_QUALITY_DEFINITIONS]: createSaveHandler(section, '/qualitydefinition'),
+
+    [FETCH_QUALITY_DEFINITION_LIMITS]: function(getState, payload, dispatch) {
+      dispatch(set({ section, isLimitsFetching: true }));
```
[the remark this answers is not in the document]
rcdailey: I'm not seeing the difference here.

rcdailey:
I am also adding some validation to the backend with unit tests. I will push a commit for that soon.

frontend/src/Settings/Quality/Definition/QualityDefinition.js
```diff
@@ -48,20 +36,29 @@ class QualityDefinition extends Component {
...
 
-    this._forceUpdateTimeout = null;
+    const { sizeLimits } = this.props;
 
-    this.state = {
-      sliderMinSize: getSliderValue(props.minSize, slider.min),
-      sliderMaxSize: getSliderValue(props.maxSize, slider.max),
-      sliderPreferredSize: getSliderValue(props.preferredSize, (slider.max - 3))
-    };
+    sizeLimits.clampMax = (inputValue) =>
```
markus101, line 41: `sizeLimits` comes from the redux store and this mutates that state. Mutating the store outside of an action in a way that will break and is not reactive.

These should be instance functions instead.

frontend/src/Store/Actions/Settings/qualityDefinitions.js
```diff
@@ -44,15 +44,45 @@ export default {
...
+          section,
+          isLimitsFetching: false,
+          isLimitsPopulated: true,
+          limitsError: null,
+          limits: data
+        }));
+      });
+
+      promise.fail((xhr) => {
+        dispatch(set({
```
markus101, line 78: If this fails to fetch we treat it as populated, but anything trying to use the limits will fail.

src/NzbDrone.Core/Qualities/QualityDefinitionLimits.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System.Diagnostics.CodeAnalysis;
+
+namespace NzbDrone.Core.Qualities;
+
+[SuppressMessage("Performance", "CA1822:Mark members as static", Justification =
```
markus101, line 5: We should be able to serialize the static values, probably just need to tell the serializer how, though just having these as static values and translating to a resource in the API would be the typical way we've been doing this.

src/Sonarr.Api.V3/Qualities/QualityDefinitionController.cs
```diff
@@ -55,6 +55,12 @@ public object UpdateMany([FromBody] List<QualityDefinitionResource> resource)
                 .ToResource());
         }
 
+        [HttpGet("limits")]
+        public IActionResult GetLimits()
+        {
+            return Ok(_qualityDefinitionService.GetLimits());
```
markus101, line 61: I don't think this is going to play nice with API docs.

markus101 (changes requested):
Given the major effort of converting JS to TS and the conflicts with https://github.com/Sonarr/Sonarr/pull/6875 I don't think landing the UI changes makes sense at this time. Adding the API response would be beneficial though.

package.json
```diff
@@ -6,8 +6,8 @@
     "build": "webpack --config ./frontend/build/webpack.config.js",
     "prebuild": "yarn clean",
     "clean": "rimraf ./_output/UI && rimraf --glob \"**/*.js.map\"",
-    "start": "webpack --watch --config ./frontend/build/webpack.config.js",
-    "watch": "webpack --watch --config ./frontend/build/webpack.config.js",
+    "start": "webpack --output-pathinfo --watch --config ./frontend/build/webpack.config.js",
+    "watch": "webpack --output-pathinfo --watch --config ./frontend/build/webpack.config.js",
```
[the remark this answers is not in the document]
rcdailey: I can remove this. I added it based on some findings from my research on Google. I had some difficulty debugging in WebStorm for a bit, and I thought maybe this helped.

frontend/src/Settings/Quality/Definition/QualityDefinition.js
```diff
@@ -48,20 +36,29 @@ class QualityDefinition extends Component {
...
+      clampRangeAndTranslateNull(inputValue, sizeLimits.minLimit, sizeLimits.preferredLimit);
+
+    this.sizeLimits = sizeLimits;
+    this._forceUpdateTimeout = null;
+    this.state = this.buildStateUsingSize(props.minSize, props.maxSize, props.preferredSize);
   }
 
   componentDidMount() {
-    // A hack to deal with a bug in the slider component until a fix for it
+    // A hack to deal with a bug in the this.slider component until a fix for it
```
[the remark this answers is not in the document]
rcdailey: Correct, and good catch. I will restore this line.

frontend/src/Settings/Quality/Shapes/SizeLimitsShape.js
```diff
@@ -0,0 +1,7 @@
+import PropTypes from 'prop-types';
```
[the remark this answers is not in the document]
rcdailey: I appreciate and respect that; however I'm not very familiar with the code base, the specific best practices being encouraged for its future, nor the technology it is based on. With this much uncertainty, I've been trying to "go with the flow" and keep things similar to what I'm already seeing.

If you can provide more specific guidance, I am happy to comply, otherwise I'm not sure what it is you want me to do.

frontend/src/Settings/Quality/Definition/QualityDefinition.js
rcdailey: Could you provide an example, for my educational benefit and to ensure I make the changes you're expecting?

frontend/src/Store/Actions/Settings/qualityDefinitions.js
rcdailey: What is the corrective action you'd like me to take here?

src/NzbDrone.Core/Qualities/QualityDefinitionLimits.cs
rcdailey: I do not believe they should be static. Perhaps I could enhance the wording of my justification here, but my goal was to silence the false positive here. I don't personally like leaving analysis issues for the next person...

src/Sonarr.Api.V3/Qualities/QualityDefinitionController.cs
rcdailey: You mean swagger/OpenAPI? I wasn't aware you were using that, so I can make the necessary changes for that and verify the output is properly documented.

Having said that, if there's a specific pattern you'd like me to follow, I'd appreciate some clearer direction. I, of course, want to follow current standards required by the project.

rcdailey:
> Given the major effort of converting JS to TS and the conflicts with #6875 I don't think landing the UI changes makes sense at this time. Adding the API response would be beneficial though.

I'll admit this is a little frustrating to hear, especially after having put in significant effort to first visit the Discord server and thoroughly discuss my idea prior to starting work. One of the primary reasons for doing that was to avoid wasted work, which this will now become.

Please understand that I've spent a significant amount of my free time working on this, probably more than justified, simply because the frontend work is not my area of expertise. While I understand and appreciate your concerns regarding conflicts, and was fully aware going into this that this would be throw-away work, I was still expecting that this would live for some period of time prior to your PR getting merged. There's at least some benefit in that, IMO.

I would ask that you reconsider. I'll verify by looking at your PR again, but my assumption up until this point has been that these conflicts should be fairly easy to resolve, since you're probably just deleting the Quality Definitions page anyway, right?

frontend/src/Settings/Quality/Shapes/SizeLimitsShape.js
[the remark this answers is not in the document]
rcdailey: Removing the shape seems to me to make the code more error prone, as now there's zero context as to the schema of the object. Removing the shape wasn't one of the interpretations I had after reading his feedback. If anything, it sounded more like he was asking for a typescript version of this instead of adding more `js` files.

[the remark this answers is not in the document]
markus101: Removing the shape is certainly the easiest solution, but my original thought was adding this shape to `QualityDefinitions` instead of a separate file.

frontend/src/Settings/Quality/Definition/QualityDefinition.js
markus101: Since you need to reference `this.props` in these functions you'll want to do something similar to how `getSelectedIds` is implemented here:

https://github.com/Sonarr/Sonarr/blob/14005d8d1054eafaba808337a109d5812f3e79e6/frontend/src/Wanted/CutoffUnmet/CutoffUnmet.js#L65-L67

```
clampMaxAndTranslateNull = (inputValue) => {
	const { minLimit, maxLimit } = this.props.sizeLimits;

   return clampRangeAndTranslateNull(inputValue, sizeLimits.minLimit, sizeLimits.maxLimit);
}

the caller of the functions would call `this.clampMaxAndTranslateNull(inputValue)`

File: frontend/src/Store/Actions/Settings/qualityDefinitions.js (hunk 4 above)
Comment (markus101) [reply frontend/src/Store/Actions/Settings/qualityDefinitions.js:78]: Either the definitions shouldn't be displayed and and error message is shown instead or the limits on the server side need to be the defaults here, which really defeats the purpose.

File: src/NzbDrone.Core/Qualities/QualityDefinitionLimits.cs (hunk 5 above)
Comment (markus101) [reply src/NzbDrone.Core/Qualities/QualityDefinitionLimits.cs:5]: Why shouldn't they be static? They're not going to be changed during run time and assuming validation happens on the API side they'd be referenced there as well instead of just serialized to JSON.

Perhaps though they'd be better as constants and the API can generate the expected object.

File: src/Sonarr.Api.V3/Qualities/QualityDefinitionController.cs (hunk 6 above)
Comment (markus101) [reply src/Sonarr.Api.V3/Qualities/QualityDefinitionController.cs:61]: Yeah, they're automatically generated with a GHA.

Not sure if something like https://github.com/Sonarr/Sonarr/blob/develop/src/Sonarr.Api.V3/Series/SeriesImportController.cs#L19-L24 with a proper return type should do it. From the API typically we'd return a "Resource" so something like `QualityLimitsResource`

**Review by markus101 — COMMENTED** [review]
> I'll admit this is a little frustrating to hear, especially after having put in significant effort to first visit the Discord server and thoroughly discuss my idea prior to starting work. One of the primary reasons for doing that was to avoid wasted work, which this will now become.

Your original plan was to add it to the API which then evolved into adding it to the UI. I mentioned doing this work against the PR that moves size limits to the profiles instead of what was always destined to be throw away work adding limits to support sizes in quality definitions.

> Please understand that I've spent a significant amount of my free time working on this, probably more than justified, simply because the frontend work is not my area of expertise. While I understand and appreciate your concerns regarding conflicts, and was fully aware going into this that this would be throw-away work, I was still expecting that this would live for some period of time prior to your PR getting merged. There's at least some benefit in that, IMO.

There is definitely value in the API returning this information, even beyond #6875, but that PR already converts some of this to TypeScript.

> I would ask that you reconsider. I'll verify by looking at your PR again, but my assumption up until this point has been that these conflicts should be fairly easy to resolve, since you're probably just deleting the Quality Definitions page anyway, right?

It's not being deleted outright and the conflicts won't be hard to resolve, but reviewing the UI changes here is not a zero-cost task either.

File: src/NzbDrone.Core/Qualities/QualityDefinitionLimits.cs [hunk 10]
```diff
@@ -0,0 +1,12 @@
...
+    "Required to be non-static for serialization purposes")]
+public record QualityDefinitionLimits
+{
+    public int MinLimit => 0;
+    public int PreferredLimit => 995;
```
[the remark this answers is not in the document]
rcdailey: Could you help me understand the issue with preferred being an absolute number as opposed to an offset? So far in my coding and testing, it seems to work just fine. The distance between min & max is purely a UI concern, from what I can tell.

If the max value changes in the future, we simply adjust the preferred along with it as needed. Making it a distance means more math/work that clients have to repeat, which is one of the issues I'm trying to solve here.

src/NzbDrone.Core/Qualities/QualityDefinitionLimits.cs
rcdailey: I'm thinking of it the other way around:

- Why *should* they be static?
- Does the difference matter?
- What's wrong with how they currently are?

I'd rather spend cycles working on your other feedback which is far more important, IMO.

frontend/src/Settings/Quality/Definition/QualityDefinition.js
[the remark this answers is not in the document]
rcdailey: I have reverted all frontend changes and this will be resolved in my next push.

package.json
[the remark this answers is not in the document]
rcdailey: This has been reverted and will be available in my next push.

frontend/src/Settings/Quality/Shapes/SizeLimitsShape.js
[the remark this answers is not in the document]
rcdailey: I have reverted all frontend changes and this will be resolved in my next push.

frontend/src/Settings/Quality/Definition/QualityDefinition.js
rcdailey: I have reverted all frontend changes and this will be resolved in my next push.

frontend/src/Store/Actions/Settings/qualityDefinitions.js
rcdailey: I have reverted all frontend changes and this will be resolved in my next push.

rcdailey:
> Your original plan was to add it to the API which then evolved into adding it to the UI.

I thought it was clear that the UI work was implied, as having the limits hard-coded in two locations seems impractical. I never would have assumed you would be okay with that. However, it seems we had a misunderstanding, and there's no use dwelling on it now.

> what was always destined to be throw away work

While this is true, I wouldn’t be pursuing this work if the throwaway aspect had been determined before any actual usage. I acknowledge that this work was always meant to be temporary, but the timing of when that decision is made is important. Given how long your PR has been open and the size of the changes, I anticipated that the UI work would be useful for some time before your merge.

Overall, I see an opportunity for better engagement in discussions among core maintainers on Discord, as well as a need for fewer assumptions and clearer outlines of the work to be performed. These are goals I think we can all strive toward.

That said, my next push will remove all of the work under `frontend/` and focus solely on the API changes. I also plan to add some test coverage and validation; most of that work is already completed, but I want to ensure we can address some of the ongoing (C# specific) disagreements before I invest more time.

rcdailey:
> > Overall, I see an opportunity for better engagement in discussions among core maintainers on Discord, as well as a need for fewer assumptions and clearer outlines of the work to be performed. These are goals I think we can all strive toward.
> 
> I think we already tried that but we had no clue what are your intentions fully. Personally I don't think anyone would've any issues with this PR being merged before v5 (since that's why the other PR is on stand-by) if the UI changes were simple and not making use of some reactjs anti-patterns like the state manipulation and arrow functions right in the constructor.
> 
> I understand you're a newcomer to this codebase and reactjs, happy to help like I already did, but there's no point in getting emotional over this. I closed enough PRs to acknowledge that once merged the maintainers need to deal with long after the author is gone.

I'm not emotional at all. Not sure why you're asserting otherwise. I'm stating constructive action for the future, which I think is perfectly fine. The frontend code will be removed from this PR, like I stated, so we can move on to more important matters.

src/Sonarr.Api.V3/Qualities/QualityDefinitionController.cs
rcdailey: I changed the return type to `ActionResult<QualityDefinitionLimits>` and I get this output when I generate docs from the exported Open API JSON file:
image

This will be available in my next push, with the rest of the changes.

src/NzbDrone.Core/Qualities/QualityDefinitionLimits.cs
rcdailey: I'm going to decline the request to change these to static. Out of respect for your request, I did do a little more research just to make sure my opinion is informed. Here are the reasons I'm going to keep it as-is:

- These properties define the surface of a public API. The fact that the values are constant is an implementation detail. They may not always be constant. From a design standpoint, these properties make more sense as non-static given the intent and their meaning, as well as the context in which they are used (DTO, API contract).
- There doesn't seem to be an easy way to serialize static members. I'm not saying it's impossible, but it's just not easy. It's just quite frankly not the right solution for serializable types (i.e. DTOs). The pattern I've adopted here (instance properties) is very typical for DTOs.

Having said that, I have made the justification message less confusing, so hopefully that helps. That will be available in my next push.

rcdailey: > Because they never change during run time, regardless of them changing in the future they will never change during run time, which a static value is perfect for. It would also allow another consumption of these values without creating another instance of these limits.

I appreciate and understand your reasoning, however for the reasons previously mentioned I'm going to keep it as-is. Consumption of the limits elsewhere is currently not a requirement and so is not something I'm willing to consider here; it would be a simple matter to refactor this code later, should that become a requirement.

> Defining a class within Core and then using that directly as a resource in the API is not something that we typically do because the API contract doesn't necessarily follow the internal structure; albeit unlikely in this particular instance.

This reframes the discussion a bit. My goal was to reduce maintenance overhead by avoiding duplicating this class. If you insist on this anyway, then your argument for static makes more sense -- in that case, the `Resource` version would be non-static (for the reasons I mentioned in my last reply) and the non-resource version in Core would probably be static since I can see how no instance is really needed there, although I'll admit because I skipped this pattern I did not look close enough to ensure that's an informed opinion.

> You were talking about validation previously, in validation we wouldn't create a new instance of this DTO in order to validate that each quality is within the correct bounds.

I'm not quite sure what your specific expectations are regarding tests for validation; however in what I've started, the DTO is indeed created multiple times for validation purposes, but in the context of a test I'm not sure how or why this matters. Semantically, a new instance is only created once per test method, but that's required because I do not typically share state at class scope in test fixtures as a personal policy. Again, I'm willing to follow the direction of the project here; what I've started really was just in lieu of understanding a desired approach.

I think if you can let me know how strongly you feel about duplicating the class (to have a separate resource version for the API and another for the core domain), that will ultimately determine the outcome of the static discussion (which isn't the main focus as I'm coming to understand).

src/NzbDrone.Core/Qualities/QualityDefinitionLimits.cs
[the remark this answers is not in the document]
rcdailey: Thanks for the clarification.

> Because while preferred size has a finite maximum, the preferred size should never exceed the maximum allowed size for a quality. Currently the only thing that enforces this is the slider, the slider won't allow things to cross over or get within 5 points of each other. Likewise preferred shouldn't be allowed to be below the minimum size and the slider enforces that it should be 5 points above the minimum size.

I agree, and the API validation that I have yet to add to this PR should address this concern. That by itself doesn't require the API itself to be aware of any kind of relative offsets. Granted, I have not implemented the validation in terms of a specific distance from the min or max points, which I don't feel is necessary for the API side of things. Limiting the distance of preferred from its endpoints makes more sense in the UI for UX reasons; for example the user needs to be able to see & interact with the preferred circle control when it is really close to one of its endpoints, since if they're on top of each other one can't be clicked... this is why I was saying it's a UI concern.

> I can't think of a scenario where the preferred size should be less than the minimum or greater than the maximum, so enforcing the same behaviour as the slider on all inputs as well as in API validation.

I've removed all of my frontend changes, so whatever is being done there won't matter anymore for the purposes of this discussion. And in my text above I discuss validation on the API side, I explain that I don't see a reason for the API validation to necessarily need the same exact logic.

> Thinking about the UX, either the inputs would need to enforce those limits while changing or enforce the limits on the other values, such as pearling [zillow.github.io/react-slider](https://zillow.github.io/react-slider/)

I think with the focus now changed to the UI work to be done on the quality profiles page related to sizes, this is a separate concern. Obviously any UI validation/limits enforcement would benefit UX, but is no longer a hard requirement once API validation gets added. It'll be harder for not only the UI, but more importantly third party clients to submit invalid sizes.

I'll go ahead and resolve this conversation since this far into the discussion I still don't see a reason for the backend to concern itself with relative distance for preferred. As long as preferred <= max (for example), that's mathematically valid. Any stricter limits are up to the client (the UI in this scenario) to enforce. Again if I'm missing important functional considerations let me know, but otherwise I'll keep what I have.

rcdailey:
@markus101 I know there's been a lot of discussion and churn on this PR. I think there's two things I'm waiting on at the moment:

1. Clarification needed on the DTO/Resource class duplication (second paragraph in [this comment](https://github.com/Sonarr/Sonarr/pull/7122#discussion_r1730040193)), that'll help direct where we go as far as that specific piece of feedback is concerned.

2. If you could sign off on the Preferred discussion, I can get those validation rules implemented and pushed up to the PR for your review.

Thanks again for taking the time!

src/NzbDrone.Core/Qualities/QualityDefinitionLimits.cs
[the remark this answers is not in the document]
markus101: If the UI and API have different rules on what is valid then it'd be possible for an API client to set a value to something that isn't possible in the UI, while it's possible to correct that in the UI it's pretty janky and makes me wonder why enforce it in the API at all?

That said the more I think about it, the more I'd rather not have the difference at all, just enforce that Min <= Preferred and Max >= Preferred, but in that case I think the preferred maximum just disappears, since it doesn't need to be 5 less than the maximum, it just needs to not exceed the maximum limit or max size.

src/NzbDrone.Core/Qualities/QualityDefinitionLimits.cs
markus101: I don't feel too strongly about separating the core from API in this instance, but once it's used in multiple places I think it being a static variable will make more sense. If you have a solution in for validation I'm fine to defer this until you have an example for that.

src/NzbDrone.Core/Qualities/QualityDefinitionLimits.cs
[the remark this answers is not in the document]
rcdailey: > If the UI and API have different rules on what is valid then it'd be possible for an API client to set a value to something that isn't possible in the UI, while it's possible to correct that in the UI it's pretty janky and makes me wonder why enforce it in the API at all?

I agree with you, but that wasn't what I was trying to say. I apologize for the confusion; let me attempt to clarify. Let's say that the API and the UI both enforced the same rule. Let's just use min and max for example. So the rule would be `Min <= Max`. The UI-specific concern here is, how do we visually separate the two circle widgets so that the user can click each one? At a resolution of 1000, the spacing between say 999 and 1000 isn't very big. So we probably want some artificial gap that the slider uses to keep them separated enough for user interaction. This is the UI-only concern I was talking about. We might say that a spacing of 5 units is enough to get that spacing we want -- but that's not necessarily something the API enforces. The UI could, for example, do something like this:

```
slider.min = min(quality.min, quality.max - 5)
```

Users could still input a manual value that is closer than 5 units, but the slider will always render the circles a certain distance from each other. This means that if the user wants a high level of control or granularity, they probably need to stick with inputting manual numbers instead of using the slider itself.

This is just an example to demonstrate my point -- the UI can be *more restrictive* than the API, but in a way that doesn't really demand a difference in behavior from other clients, because the reasons for the stricter limit may not exist in those third party clients (e.g. UI elements).

> That said the more I think about it, the more I'd rather not have the difference at all, just enforce that Min <= Preferred and Max >= Preferred, but in that case I think the preferred maximum just disappears, since it doesn't need to be 5 less than the maximum, it just needs to not exceed the maximum limit or max size.

I agree here and this is the approach I was attempting to take. The more straightforward comparisons you mentioned are what I currently have in my validation rules. I also agree with your point about the preferred maximum. Removing this would also greatly simplify validation. I will go ahead and remove it in favor of the rule you described.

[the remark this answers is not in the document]
markus101: I think even at a distance of 5 they become too close to really see which is which and part of the reason I'd like to come up with some different UX there, but I agree enforcing that gap at the API level becomes tricky and would be annoying to communicate clearly.

markus101:
Moving to draft while you work on validation.

rcdailey:
Limits validation has been pushed up. I also added support for validating collection types in `RestController`.

There's some important behavior differences on the qualities page that I messaged you about in Discord. Hopefully we can iron that out there.

src/Sonarr.Api.V3/Qualities/QualityDefinitionController.cs
```diff
@@ -20,6 +20,15 @@ public QualityDefinitionController(IQualityDefinitionService qualityDefinitionSe
             : base(signalRBroadcaster)
         {
             _qualityDefinitionService = qualityDefinitionService;
+
+            SetupValidation(qualityDefinitionService);
+        }
+
+        private void SetupValidation(IQualityDefinitionService qualityDefinitionService)
+        {
+            var limits = qualityDefinitionService.GetLimits();
```
markus101, line 29: This is what I was talking about before, you need to make an instance of `QualityDefinitionLimits` to use in the validator instead of just accessing the limits from a static class.

Since `QualityDefinitionResourceValidator` will never actually use different limits within the same version of the application it makes sense to use a static value instead of initializing with a dynamic value.

src/NzbDrone.Core/Qualities/QualityDefinitionLimits.cs
markus101: As commented on the usage of the validator I think it makes sense to move to static values here and a resource class to handle the serialization on the API side.

This will also allow the limits to be used in quality profiles without refactoring.

src/NzbDrone.Api.Test/Qualities/QualityDefinitionResourceValidatorTest.cs
```diff
@@ -0,0 +1,99 @@
+﻿using FluentValidation.TestHelper;
+using NUnit.Framework;
+using NzbDrone.Core.Qualities;
+using Sonarr.Api.V3.Qualities;
+
+namespace NzbDrone.Api.Test.Qualities;
```
markus101, line 6: This should be moved under `v3`

markus101 (commented):
I haven't had a chance to whip up a diff for the UI, but wanted to drop some feedback. If multiple changes are made it can be hard to tell which item submitted is wrong, not sure if there is anything other than `WithState` that could be used to add that.

src/Sonarr.Api.V3/Qualities/QualityDefinitionResourceValidator.cs
```diff
@@ -0,0 +1,28 @@
...
+public class QualityDefinitionResourceValidator : AbstractValidator<QualityDefinitionResource>
+{
+    public QualityDefinitionResourceValidator(QualityDefinitionLimits limits)
+    {
+        When(c => c.MinSize is not null, () =>
+        {
+            RuleFor(c => c.MinSize)
+                .GreaterThanOrEqualTo(limits.MinLimit)
+                .WithErrorCode("GreaterThanOrEqualTo")
+                .LessThanOrEqualTo(c => c.PreferredSize ?? limits.MaxLimit)
```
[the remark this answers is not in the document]
rcdailey: In your test, what was `MinSize` set to?

src/Sonarr.Api.V3/Qualities/QualityDefinitionController.cs
rcdailey: Is creating that instance actually a problem here, though? In the grander scheme of things, I don't think it really matters. Please remember my original points as to why these members are not static. This isn't me just being hard-headed about it.

I completely understand what you're saying, but on the other hand, if you look at this objectively it really doesn't matter if they're static or not from a performance standpoint. There's just not enough overhead here for it to matter, IMO.

In our previous discussion, I did clarify with you whether or not you wanted to duplicate the structure (once for serialization and once as more of a domain object) but it seemed like you agreed at the time that the duplication (maintenance overhead) wasn't worth it, which I agree with. But if you decide to change your mind on that, it may open the opportunity to use static values here.

I'm still 100% ok with this code given:

1. The points I made in our earlier discussion about why they are not static
2. The performance overhead / difference is negligible.
3. Code can always be changed and refactored later should the first two points ever change.

I'll go ahead and resolve this conversation for now for the sake of making progress and getting things merged. This PR has already been open quite a while and I want to try to come to a compromise as much as possible to avoid this turning into something akin to the level of paid work.

src/NzbDrone.Api.Test/Qualities/QualityDefinitionResourceValidatorTest.cs
rcdailey: Rider is very intrusive about how it wants namespaces to be done. When I created the test fixture, it assumed the namespace should follow the directory structure. I'll take a look at some of the other tests and try to deduce what the namespace should be here.

I do wish there was a way to tell Rider to use a specific namespace for all files; I haven't quite found a good solution to that.

src/Sonarr.Api.V3/Qualities/QualityDefinitionResourceValidator.cs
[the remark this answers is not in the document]
rcdailey: Maybe your point here was: "I changed Preferred but it complained about Min instead". In that case, I don't think the API can really know what the user's *intent* was; especially given how the payload comes in right now. We get 3 min,preferred,max values. If I add validation to Preferred, you'll get two validation errors: One saying that Min must be lower than preferred, and one saying Preferred cannot be lower than Min (or something along those lines).

I originally had a validation rule for `PreferredSize` but this made the logic confusing *and* the test more complicated, without really adding any additional safeguards over just the min & max validation.

I agree that the verbiage is confusing; I just don't know how to make that better without overcomplicating the validation itself.

src/NzbDrone.Api.Test/Qualities/QualityDefinitionResourceValidatorTest.cs
rcdailey: Ah I misunderstood -- you meant move it under the directory named `v3`. Namespaces following the directory structure *does* seem to be the desired pattern for at least this test project. I will get it done!

src/Sonarr.Api.V3/Qualities/QualityDefinitionController.cs
rcdailey: I just saw your other comment on `QualityDefinitionLimits.cs` (but I can't comment on that directly, for some reason...)

If you're OK with the two separate classes I will go ahead and make that change out of respect for your request, although I personally do not see an issue here. Hopefully after this we can get this merged without much fuss.

src/Sonarr.Api.V3/Qualities/QualityDefinitionResourceValidator.cs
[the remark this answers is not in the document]
markus101: What happened was MinSize was 20, I changed MaxSize to 5 to see how things would fail, but because the slider kept preferred -5 from max it actually set preferred to 0.

Perhaps an easier comparison would be.

```
Min >=0
Max <= 1000
Preferred >Min or 0
Preferred< Max or 1000
```

It's certainly not perfect and we don't know the intent of the user so they could submit something that gives a weird result. but at least it shouldn't try to return an error that would lead to another error.

src/NzbDrone.Api.Test/Qualities/QualityDefinitionResourceValidatorTest.cs
markus101: Ahh yeah, I was unclear, move to v3 folder + update namespace.

src/Sonarr.Api.V3/Qualities/QualityDefinitionResourceValidator.cs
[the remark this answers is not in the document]
rcdailey: This use case confuses me, I'm sorry. Can you walk me through the actual steps you took? If I set Min size to 20 in the UI, and try to manually type in `5` for Max, it caps it to `21`. So I don't actually what exactly you're testing.

What exact min/preferred/max payload for that specific quality came into the API?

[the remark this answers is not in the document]
markus101: I think you can do it by setting min 0, max to 5, then preferred to (if it doesn't do it automatically), then increase min and max. It's definitely a little funky doing it in the UI, but the same is true if you just hit the API directly as well.

src/Sonarr.Api.V3/Qualities/QualityDefinitionController.cs
markus101: This is what I was thinking: https://github.com/Sonarr/Sonarr/commit/954903812197d0fad3bc091e6e469342680f7776

markus101:
For the UI we could do something like this to just show the error, but if multiple are updated and have error it could be hard to map the errors to the items that are having the issue without more details https://github.com/Sonarr/Sonarr/commit/974c4a601b6b6871f8283a1b9472a987d32ff924

src/Sonarr.Api.V3/Qualities/QualityDefinitionResourceValidator.cs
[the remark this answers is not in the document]
rcdailey: In my last comment I asked for the payload that was sent to the API in your test case. If you can get me that, I'll just test directly to the API instead of messing with the UI.

src/Sonarr.Api.V3/Qualities/QualityDefinitionController.cs
rcdailey: So basically you're OK with no separation of concerns between the controller and the logic. Not that there's much logic there to begin with, but I'm more interested in the principles in this case. I'll go ahead and adjust my code to remove logic from the service class.

rcdailey:
> For the UI we could do something like this to just show the error, but if multiple are updated and have error it could be hard to map the errors to the items that are having the issue without more details [974c4a6](https://github.com/Sonarr/Sonarr/commit/974c4a601b6b6871f8283a1b9472a987d32ff924)

This looks like a fantastic idea; I'll go ahead and enable edits to my PR by maintainers if you'd like to add that change here, or otherwise it looks like a good candidate for a second PR later.

src/Sonarr.Api.V3/Qualities/QualityDefinitionController.cs
rcdailey: Resolved in latest push

src/Sonarr.Api.V3/Qualities/QualityDefinitionResourceValidator.cs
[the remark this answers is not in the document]
rcdailey: I went ahead and added validation for `PreferredSize` (I originally added that to begin with but removed it thinking it was redundant). I still didn't fully understand your scenario, but I do recognize that if min & max are `null` and we have an invalid preferred size, such as `-20`, the validation wouldn't catch that without this change.

rcdailey:
I believe at this point all concerns have been addressed. I've moved this PR out of draft status. I'd very much appreciate a merge!

I have also enabled the option to allow edits by maintainers if you'd like to make further adjustments on your own.

src/Sonarr.Api.V3/Qualities/QualityDefinitionLimitsResource.cs
```diff
@@ -0,0 +1,6 @@
+namespace Sonarr.Api.V3.Qualities;
+
+// SA1313 still applies to records until https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/3181 is available in a release build.
+#pragma warning disable SA1313
```
markus101, line 4: Made this consistent with other places where we've disabled this.

src/NzbDrone.Core/Qualities/QualityDefinitionLimits.cs
```diff
@@ -0,0 +1,7 @@
+namespace NzbDrone.Core.Qualities;
+
+public static class QualityDefinitionLimits
+{
+    public const int Min = 0;
```
markus101, line 5: Ended up dropping `Limit` from these to avoid the duplication.

markus101 (approved):
Thanks for all the work on this! I just made a small tweak in the naming to reduce duplication.

PR #7133, opened 2024-08-23
Fixed: Handle Spaces in ProxyBypassList

bakerboy448:
#### Description
title for when users `use, spaces, to, delimit,rather,than,no,space`

https://github.com/Prowlarr/Prowlarr/issues/2203

### DIFF
--- a/src/NzbDrone.Common/Http/Proxy/HttpProxySettings.cs
+++ b/src/NzbDrone.Common/Http/Proxy/HttpProxySettings.cs
@@ -30,7 +30,7 @@ namespace NzbDrone.Common.Http.Proxy
             {
                 if (!string.IsNullOrWhiteSpace(BypassFilter))
                 {
-                    var hostlist = BypassFilter.Split(',');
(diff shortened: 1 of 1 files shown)


mynameisbogdan (changes requested):

src/NzbDrone.Common/Http/Proxy/HttpProxySettings.cs
```diff
@@ -30,7 +30,7 @@ public string[] BypassListAsArray
             {
                 if (!string.IsNullOrWhiteSpace(BypassFilter))
                 {
-                    var hostlist = BypassFilter.Split(',');
+                    var hostlist = BypassFilter.Split(',', StringSplitOptions.RemoveEmptyEntries | StringSplitOptions.TrimEntries);
                     for (var i = 0; i < hostlist.Length; i++)
```
markus101, line 34:
```suggestion
                    var hostlist = BypassFilter.Split(',', StringSplitOptions.RemoveEmptyEntries | StringSplitOptions.TrimEntries);

                    for (var i = 0; i < hostlist.Length; i++)
```

PR #7142, opened 2024-08-26
Convert Components to TypeScript

markus101:
#### Description

Converts a bunch of components to TypeScript.


Treycos (commented):
I think I'll stop there since most comments are more or less about the same design recommendations   

In almost every components, I'd recommend extending other interfaces (native or not) if possible, to avoid duplicating types and facilitate core components updates.

A minimal component that implements these recommendations would be `TableRowCellprops` in this PR :
https://github.com/Sonarr/Sonarr/pull/7145/files

TL:DR
- Extend any native or already existing types
- Export the interface to allow other components to do the same
- Spread the whole thing in the component JSX whenever it makes sense
- Use `Pick` when you want a subset of an existing interface
- Use `FooProps['<name>']` when you want to import a prop under another name

frontend/src/Components/HeartRating.tsx
```diff
@@ -1,22 +1,22 @@
...
 import Tooltip from 'Components/Tooltip/Tooltip';
 import { icons, kinds, tooltipPositions } from 'Helpers/Props';
 import translate from 'Utilities/String/translate';
 import styles from './HeartRating.css';
 
-function HeartRating({ rating, votes, iconSize }) {
+interface HeartRatingProps {
+  rating: number;
+  votes?: number;
+  iconSize?: number;
```
Treycos, line 11:
```suggestion
  iconSize?: IconProps['size'];
```

frontend/src/Components/Scroller/OverlayScroller.tsx
```diff
@@ -0,0 +1,128 @@
...
+import { OnScroll } from './Scroller';
+import styles from './OverlayScroller.css';
+
+const SCROLLBAR_SIZE = 10;
+
+interface OverlayScrollerProps {
+  className?: string;
+  trackClassName?: string;
+  scrollTop?: number;
+  scrollDirection: unknown | unknown | unknown;
```
Treycos, line 12: Typo ?

frontend/src/Components/SpinnerIcon.tsx
```diff
@@ -0,0 +1,30 @@
+import { FontAwesomeIconProps } from '@fortawesome/react-fontawesome';
+import React from 'react';
+import { icons } from 'Helpers/Props';
+import Icon from './Icon';
+
+interface SpinnerIconProps {
+  className?: string;
+  name: FontAwesomeIconProps['icon'];
+  spinningName?: FontAwesomeIconProps['icon'];
```
Treycos, line 9:
```suggestion
  spinningName?: IconProps['name'];
```

frontend/src/Components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,215 @@
...
+import dimensions from 'Styles/Variables/dimensions';
+import { isMobile as isMobileUtil } from 'Utilities/browser';
+import styles from './Tooltip.css';
+
+export interface TooltipProps {
+  className?: string;
+  bodyClassName?: string;
+  anchor: React.ReactNode;
+  tooltip: string | React.ReactNode;
+  kind?: Extract<(typeof kinds.all)[number], keyof typeof styles>;
```
Treycos, line 21: I didn't anticipate the type `(typeof kinds.all)[number]` would be used so much (same for sizes)

Exporting it from `Kinds` under a simpler name might be a good idea now

frontend/src/Components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,215 @@
...
+    } else if (windowWidth >= parseInt(dimensions.breakpointSmall)) {
+      return 500;
+    }
+
+    return 450;
+  }, []);
+
+  const computeMaxSize = useCallback(
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    (data: any) => {
```
Treycos, line 89: Only type I could get working is this this `react-popper` doesn't export much from `popper`

Not great 🙁 

```suggestion
  const computeMaxSize = useCallback<
    NonNullable<
      NonNullable<
        NonNullable<PopperProps['modifiers']>['computeMaxHeight']
      >['fn']
    >
  >(data) => {
```

frontend/src/Components/Tooltip/Tooltip.tsx
markus101: Good call, did the same for the others. Ended up extending the types a bit to avoid issues with them being treated as strings in some places.

frontend/src/Components/Tooltip/Tooltip.tsx
markus101: Yeah, saw there wasn't much exported and gave up. I ran into issues with this change and reverted back, but not too worried, eventually popper will get updated/replaced and it'll improve.

markus101 (commented):
Thanks for the feedback, addressed most of the comments/added some replies.

frontend/src/Components/Portal.tsx
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+
+interface PortalProps {
+  children: React.ReactNode;
+  target?: Element | DocumentFragment;
+}
```
Treycos, line 7:
```suggestion
interface PortalProps {
  children: Parameters<typeof ReactDOM.createPortal>[0];
  target?: Parameters<typeof ReactDOM.createPortal>[1];
}
```

frontend/src/Components/Portal.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+function Portal(props: PortalProps) {
+  const { children, target = document.getElementById('portal-root') } = props;
```
Treycos, line 10: You might want to move the `getElementById` call above the component to avoid calling it for every portal

frontend/src/Components/SpinnerIcon.tsx
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
```
Treycos, line 1: Did the same refactor on another branch
Here's my version

```suggestion
import React from 'react';
import { icons } from 'Helpers/Props';
import Icon, { IconProps } from './Icon';

export interface SpinnerIconProps extends IconProps {
  spinningName?: IconProps['name'];
  isSpinning: Required<IconProps['isSpinning']>;
}

export default function SpinnerIcon({
  name,
  spinningName = icons.SPINNER,
  ...otherProps
}: SpinnerIconProps) {
  return (
    <Icon
      name={(otherProps.isSpinning && spinningName) || name}
      {...otherProps}
    />
  );
}
```

frontend/src/Helpers/Props/align.ts
```diff
@@ -3,3 +3,5 @@ export const CENTER = 'center';
 export const RIGHT = 'right';
 
 export const all = [LEFT, CENTER, RIGHT];
+
+export type Align = ('left' | 'center' | 'right') & (typeof all)[number];
```
Treycos, line 7: Is the type redundancy intentional ?

These would be the same type

```suggestion
export type Align = 'left' | 'center' | 'right;
```

```suggestion
export type Align = (typeof all)[number];
```

Treycos: Same question for Kinds, Sizes, directions... etc

frontend/src/Components/Tooltip/Popover.tsx
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+import Tooltip, { TooltipProps } from './Tooltip';
+import styles from './Popover.css';
+
+interface PopoverProps extends Omit<TooltipProps, 'tooltip' | 'bodyClassName'> {
+  title: string;
+  body: string | React.ReactNode;
```
Treycos, line 7: Redundant typing

```suggestion
  body: React.ReactNode;
```

frontend/src/Components/Scroller/OverlayScroller.tsx
```diff
@@ -0,0 +1,129 @@
...
+  autoHide: boolean;
+  autoScroll: boolean;
+  children?: React.ReactNode;
+  onScroll?: (payload: OnScroll) => void;
+}
+
+interface ScrollbarTrackProps {
+  style: React.CSSProperties;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  props: any;
```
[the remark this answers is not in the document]
markus101: Not sure if I replied to the wrong thing, but yeah that definitely works 🤔

frontend/src/Components/SpinnerIcon.tsx
markus101: Totally spaced that you did this one already, making a small tweak to not use `otherProps` inside of the component at the cost of a couple extra lines 😄

frontend/src/Helpers/Props/align.ts
markus101: Yes, the indexed strings were causing issues in some places so I extended the type, but forgot how TS resolves the types/strings in that scenario. I'll clean them up

src/NzbDrone.Core/Localization/Core/es.json
```diff
@@ -1,2108 +1,2108 @@
 {
```
Treycos, line 1: Are all the formatting changes wanted or is it your formater derping 🤔

markus101: Formatter went rogue 😞

PR #7155, opened 2024-08-29
Color impaired improvements

mynameisbogdan:

- [progressbar](https://github.com/user-attachments/assets/cfa83df1-96ca-4a6f-b054-d8f43a23a54b)

- [calendar full color](https://github.com/user-attachments/assets/76987658-8493-4219-8a6c-3cae91d9ab82)

- [calendar not full color](https://github.com/user-attachments/assets/189fdf88-67c7-4c76-ae7d-60c864e7b8ab)

- [series index legend](https://github.com/user-attachments/assets/a758b143-f1ca-4a3c-bf8e-23ccae9bb0ae)

### DIFF
--- a/frontend/src/Components/ProgressBar.css
+++ b/frontend/src/Components/ProgressBar.css
@@ -19,34 +19,50 @@
 
   &.primary {
     background-color: var(--primaryColor);
+
+    &:global(.colorImpaired) {
+      background: repeating-linear-gradient(25deg, color(#5d9cec shade(5%)), color(#5d9cec shade(5%)) 5px, color(#5d9cec shade(70%)) 5px, color(#5d9cec shade(70%)) 10px);
+    }
+  }
+
+  &.success {
+    background-color: var(--successColor);
+
+    &:global(.colorImpaired) {
+      background: repeating-linear-gradient(50deg, color(#00853d shade(5%)), color(#00853d shade(5%)) 5px, color(#00853d shade(70%)) 5px, color(#00853d shade(70%)) 10px);
+    }
   }
 
   &.danger {
     background-color: var(--dangerColor);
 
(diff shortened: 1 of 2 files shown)


stevietv:
looking beautiful! I wonder if instead of different angles of the banding, maybe different thickness would read better?

stevietv:
> Might be. At least a mix of both IMO, but we need someone color impaired to yay or nay which is better.

would be better than the non-blind leading the non-blind!

PsychoManiac189:
Hi, I am R/G color impared and here is my opinion on these links:

- progress bar: if these are all the ones from the calendar could you maybe swap the colors of stripes from the 2nd from bottom and the top one? 

- calendar full color: looks good, but too bad it didnt have all examples so I can really know. I have to see stuff next to each other to see a problem/difference 

- calendar not full color: I dont like this. Then I just have to memorize the paterns and ill probably never remember them all colors/strips is easier to remember. 

- series index legend: looks good other than my comment about swapping those 2 stripes. 


other comments: I wish I could run a beta or something to really see it in action cuz I'm just staring at images and that is easy for me, but when my eyes are moving and im doing multiple things that is where the real test is  as im distracted.

PsychoManiac189:
this is what i see for sonarr image
this is what i see for radarr image
radarr stripes seem darker already?

markus101:
With these changes Red/Green both have angled lines and appear more similar than they were before, is this an overall improvement? To my eye with a protanomaly (red) filter on I find `Missing (Unmonitored)` and `Downloaded (Monitored)` quite similar, though I have no idea how accurate this filter is.

markus101:
They were "missing" intentionally, so that like colours had different effects.

PsychoManiac189:
Thank you!

PR #7161, opened 2024-08-31
Minimum Upgrade Score for Custom Formats

### LINKED ISSUE #6800: Add a minimum step for custom formats score upgrades

I use a custom format to add a +1 score to any release that has a freeleech flag, so that if multiple indexers have the same release, Sonarr will always grab it from an indexer that has it freeleech, if any. 

Here is the issue that can arise from this method. 
Let's say Sonarr searches for a show and grabs a release from RlsGroupA, with a score of 1000. Multiple indexers have that release but none of them have it freeleech.
Later on, an indexer releases the same show by RlsGroupB, which by default has the same CF score as RlsGroupA (1000); however, this release is freeleech, so it will get the CF bonus and have a score of 1001. Sonarr will thus see it as an upgrade from the previously grabbed release by RlsGroupA and replace it, even though both RlsGroup are otherwise treated as equivalent and shouldn't be considered upgrades from one another.

I realize that what I describe above is Sonarr working as intended, but I believe Sonarr could benefit from a tweak that could resolve this issue without fundamentally impacting how upgrades work.

I think it could be nice to have the ability to set a minimum step in score upgrades before Sonarr considers one release an upgrade from another. For instance, in the situation described above, I could tell Sonarr to only consider upgrading a release if another one gets a CF score of at least 5 more points; i.e., the release from RlsGroupA with a CF score of 1000 wouldn't be upgraded by the release from RlsGroupB with its CF score of 1001, but it would get an upgrade from a release by RlsGroupC that has a score of 1005 or above. 

I can't think of any alternative that doesn't involve giving up on the freeleech CF method.

### Anything else?

_No response_

somniumV:
#### Description
Allows user to restrict upgrades based on custom formats until minimum score has been met. Set to 1 by default.

image

#### Database Migration
YES - 212

* Closes #6800


src/NzbDrone.Core/Localization/Core/en.json
```diff
@@ -1176,6 +1176,8 @@
   "MinimumFreeSpace": "Minimum Free Space",
   "MinimumFreeSpaceHelpText": "Prevent import if it would leave less than this amount of disk space available",
   "MinimumLimits": "Minimum Limits",
+  "MinimumUpgradeCustomFormatScore": "Minimum Upgrade Custom Format Score",
```
markus101, line 1179: I'm worried this name is too close to `Minimum Custom Format Score`, but more so that this particular verbiage is a little awkward, moving the `Upgrade` close to the end would help.

some ideas, but also curious what @stevietv might have in mind for ideas.

- `Custom Format Score Upgrade Differential`
- `Custom Format Score Upgrade Difference`
- `Minimum Custom Format Score Upgrade Differential`
- `Minimum Custom Format Score Upgrade Difference`

markus101 (changes requested):
Looking good, just naming seems a bit awkward at the moment.

src/NzbDrone.Core/Localization/Core/en.json
stevietv: I think right now its too similar in naming too. Id think something like `Minimum Custom Format Score Increase` - its like differential, just more simple.

augustuen: What about "Minimum Custom Format Improvement"?

somniumV: I'm in favor of using `Minimum Custom Format Improvement`. Would that be acceptable?

somniumV: Noticed that renaming to `Minimum Custom Format Score Improvement` will result in the name being displayed in three seperate rows, which looks too busy imo. Changed `Improvement` to `Increase` for now.

image

bakerboy448: > Minimum Custom Format Score Upgrade Increment

This seems the clearest

nuxencs: I think `Minimum Custom Format Score Upgrade Difference` fits better 🤔

nuxencs:
This seems like a really nice feature, but I think using `Repack` Custom Formats, like quite a lot of people do, would limit the usability quite a lot because those are usually scored quite low. That would lead to them "bottlenecking" this feature.

### Simple example

Release on disk Score: 100
Minimum Upgrade Score: 10
Repack CF Score: 5
Repacked Release Score: 105

This example could be an issue when using the TRaSH Guides Custom Formats and the recommended scoring and would unfortunately prevent you from getting any repack upgrades. Setting the Minimum Upgrade Score any lower would just make the feature mostly obsolete.

One way to mitigate this would be some sort of bypass or exception for certain custom formats, but I'm not sure if that would be the best way to handle it.

I would appreciate feedback on this 😅

augustuen:
> One way to mitigate this would be some sort of bypass or exception for certain custom formats, but I'm not sure if that would be the best way to handle it.
> 
> I would appreciate feedback on this 😅

The way to mitigate this would be to modify the scores/values, just like for any thing else. TRaSH's guide is at the end of the day a guide. If you agree with his goals it'll usually work good out or the box, but many will adjust it to fit their needs. 

And if he saw it necessary, he'd update his guide after this feature is added.

And if the feature doesn't fit your usecase, setting it to 0 will disable it.

nuxencs:
> This would come in handy even with Minimum Upgrade Score 2 to prevent useless upgrades just because the new release is freeleech.

that's why I said mostly 😄 

> The way to mitigate this would be to modify the scores/values, just like for any thing else.

increasing the score for repacks doesn't make much sense, because then you would need to increase all the other scores just so repack doesn't outscore better release and then you end up in the exact same situation again.

frontend/src/Settings/Profiles/Quality/EditQualityProfileModalContent.js
```diff
@@ -244,6 +245,24 @@ class EditQualityProfileModalContent extends Component {
...
+
+                              <FormInputGroup
+                                type={inputTypes.NUMBER}
+                                name="minUpgradeFormatScore"
+                                min={1}
+                                {...minUpgradeFormatScore}
+                                helpText={translate('MinimumCustomFormatScoreIncrementHelpText')}
+                                onChange={onInputChange}
+                              />
+                            </FormGroup>
```
markus101, line 263:
```suggestion
                          upgradeAllowed.value && formatItems.value.length > 0 ?
                            <FormGroup size={sizes.EXTRA_SMALL}>
                              <FormLabel size={sizes.SMALL}>
                                {translate('MinimumCustomFormatScoreIncrement')}
                              </FormLabel>

                              <FormInputGroup
                                type={inputTypes.NUMBER}
                                name="minUpgradeFormatScore"
                                min={1}
                                {...minUpgradeFormatScore}
                                helpText={translate('MinimumCustomFormatScoreIncrementHelpText')}
                                onChange={onInputChange}
                              />
                            </FormGroup> :
                            null
```

markus101 (approved):
Just a couple things

PR #7163, opened 2024-08-31
New: Except language option for Language Custom Formats

### LINKED ISSUE #7120: Add option in custom formats to decrease the score of multi audio releases

Currently we have the original language preset that works great. But I would prefer an option to download only original language and ignore releases with multiple languages

Currently we have the original language preset that works great. But I would prefer an option to download only original language and ignore releases with multiple releases

I tried using multi as block but some of the trackers like avistaz list all the language in the search results so sonarr know what the languages are but it is not in the filename so is not blocked

### Anything else?
Screenshot from 2024-08-19 16-36-42

markus101:
#### Description

Adds an Except Language option for Language CF that allows matching on languages other than the selected language. This will allow Except Original Language to match if a language other than the original language is present and negating that will allow matching if none of the other languages are present.

* Closes #7120

### DIFF
--- a/src/NzbDrone.Core.Test/CustomFormats/Specifications/LanguageSpecification/MultiLanguageFixture.cs
+++ b/src/NzbDrone.Core.Test/CustomFormats/Specifications/LanguageSpecification/MultiLanguageFixture.cs
@@ -43,6 +43,26 @@ namespace NzbDrone.Core.Test.CustomFormats.Specifications.LanguageSpecification
         }
 
         [Test]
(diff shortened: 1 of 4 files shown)


markus101:
@mynameisbogdan thoughts on this instead of doing it based on language count?

markus101:
It'd allow for rejecting releases that contain languages other than the wanted language which should solve the problem, but knowing releases names don't always contain all the languages it'll probably lead to some imports being blocked due to different CF scores. 🤷 can't really solve that hole though.

Issue #1682, opened 2017-02-15
Multiple User Support


markus101:
Multiple users would allow user's to allow family members and friends to access Sonarr without giving them full access.

Role based access control would allow different roles to be assigned to each user that would give them different levels of access.

- Superuser - Master user, can add/remove users, alter any settings
- Admin - Master user, can add/remove users, alter settings (limited by Superuser)
- Show Manager - Can add new shows, but can't alter settings
- Read only - Can only view existing shows, can't edit or add series nor can they access settings

There could be more roles as well, making it flexible, but somewhat guided so it's not overwhelming and remains functional. Any other ideas around multiple users is welcomed, nothing is currently set in stone and this is only a proposal, it may not be something we end up implementing, but it's been something I've been thinking about.

maxkoryukov:
What you think about some kind of **custom policies** ? I mean, it could not be sufficient to have only four predefined roles: `[Superuser, Admin , Show manager, Read Only]`.

I would like to suggest the following solution:

We could add **meta** options for each setting. I suggest to use these meta':

* visibility
* editable

If these meta-settings will depend on role (for *role1* - visible and editable, for *role2* - only visible), then the proposed system will have an ability to configure custom roles (and listed predefined roles too)

markus101:
Yeah, I don't think there needs to be a limitation on which profiles exist, just some defaults to start it off.

I think there will need to be some limitations on things though, for sanity and ease of implementation, probably in some logical groups for settings outside of providers (download clients, indexers, etc) and the entire provider would be editable if you had edit rights. Going further than that would be to difficult to maintain and would likely lead to support issues all over the place, especially if new settings are added.

maxkoryukov:
Previously I have implemented simple framework, there were implemented such things, like:

* boolean checker
* dropdown element
* text input
* number input
* combobox input
* switch (boolean with dependence subfields)
* ....

Each component have:
* property `readonly`
* property `visible`
* property `id`
* property `value`

Most properties are set outside the component, but each component should correctly handle all properties

theCheek:
This is an excellent suggestion as I have multiple users requesting shows inside Plex which are sent to Sonarr, and they currently have no way of seeing the status of the downloads because I don't want to give them access to deluge or sabnzbd as they will probably break stuff. To have a user class who can only see the queue would solve this.

gold007eye:
[...]

PR #7186, opened 2024-09-06
Multiple User Support 

Alexk2309:
#### Description

This is a work in progress and is a draft to get feedback from the Sonarr team to make sure I am going in the right direction.

There are most likely many bugs that need to be fixed before this can be properly merged.

Here are the features that were added.

- A new settings page for managing users, where users can be added, changed, and deleted. 
- Creation of multiple API keys for different users.
- Added roles for users and implemented RBAC (Role Based Access Control), so far the only supported endpoint is UserController, but this is very easy to add for the other classes.
- Added protected routing for frontend, to ensure only certain roles can access certain routes (settings/users is the only protected route so far).
- Renders the UI differently depending on the user's role.
- Refactored HostConfigController, to only create a user on initial startup (user editing has now moved to user settings page).
- Refactored Authentication schema, to support multiple API keys for different users.

Screenshot 2024-09-06 150142

Screenshot 2024-09-06 150147

Screenshot 2024-09-06 150157

Screenshot 2024-09-06 150209

UI looks different in this scenario because the user doesn't have the right access level.

image

#### Database Migration

- Added Role collum in the Users table.
- Added ApiKey collum in the Users table.

* Attempts to close https://github.com/Sonarr/Sonarr/issues/1682


Alexk2309:
Please help review this PR.

@markus101 @joshka

bakerboy448:
Showing empty / blank data if permissions restrict is poor UX and will be confusing to users, create additional support issues.

Rather if access denied messages noting as such should be displayed

thezak48:
You do realise this will result in Sonarr getting banned by every private tracker out there right for account sharing

Alexk2309:
> You do realise this will result in Sonarr getting banned by every private tracker out there right for account sharing

I'm sorry this was an open issue which did not mention the point above, so I thought it was okay.

bakerboy448:
Usenet indexers as well likely, due to said sharing

Thirrian:
As if you can't run any of the apps like overseerr on top of sonarr, add their plex list, or any other number of ways to "share" your instance with other people.

Disclaimer: I don't, I just think it's weird to bring this point up now that someone is actually working on it, while the FR had been open for 7+ years...

Alexk2309:
> As if you can't run any of the apps like overseerr on top of sonarr, add their plex list, or any other number of ways to "share" your instance with other people.
> 
> Disclaimer: I don't, I just think it's weird to bring this point up now that someone is actually working on it, while the FR has been open for 7+ years...

I honestly agree with this point here. I even did some reading up on [reddit](https://www.reddit.com/r/trackers/comments/mh2wkz/do_private_trackers_care_if_you_share_their/) and I think it should be fine.

bakerboy448:
> As if you can't run any of the apps like overseerr on top of sonarr, add their plex list, or any other number of ways to "share" your instance with other people.

none of these request apps or watchlist would provide direct access to hitting indexers / trackers.

I.e. request apps can require approval of requests so a user can't trigger 100 seasons being searched at once 

These concerns don't prevent this type of work - but just means roles and permissions need to be well thought out and consensus with larger trackers and indexers to ensure they are fine with it should occur as well.


Similar to the reasons behind removing the indexer api credentials from the UI and forcing authentication

src/Sonarr.Http/Authentication/ApiKeyAuthenticationHandler.cs
```diff
@@ -61,11 +66,15 @@ protected override Task<AuthenticateResult> HandleAuthenticateAsync()
                 return Task.FromResult(AuthenticateResult.NoResult());
             }
 
-            if (_apiKey == providedApiKey)
+            // If the authentication method is set to None, this means the app has not been setup yet and allow all requests until setup is complete.
```
joshka, line 69: Is this actually desired behavior, or should we make the setup mandatory?

src/Sonarr.Http/Frontend/InitializeJsonController.cs
```diff
@@ -16,18 +19,28 @@ public class InitializeJsonController : Controller
...
             _analyticsService = analyticsService;
-
-            _apiKey = configFileProvider.ApiKey;
             _urlBase = configFileProvider.UrlBase;
+
+            _httpContextAccessor = httpContextAccessor;
+            _userService = userService;
+
+            var currentUser = GetCurrentUser();
+            _apiKey = currentUser != null ? currentUser.ApiKey : "None";
```
joshka, line 43: Is "None" the right value to use here, or would null be better? (Is this just a display property perhaps?)

joshka (commented):
A few comments / questions. I mostly didn't review the frontend parts of this as it's outside of my expertise, so I'd request someone else take a look at that.

The backend parts look fairly sane to me (bearing in mind, I haven't written any dotnet code since ~2016)

I note that there's a TODO about not allowing duplicate user names, which seems reasonable to make sure is done.

What sort of tests should be written for this?

Right now the roles are Admin and ReadOnly, but I'm not certain that the things accessible by a read only user are actually read only. Am I misunderstanding this change?

From a RBAC perspective, perhaps the real distinction is whether the user can administer users or not.

From a process perspective, this is a bit of a large PR at 1365 lines added because it touches a lot of code paths. It's possible the sonarr owners prefer larger PRs, but I know as an OSS project maintainer I generally find a few shorter PRs easier to review and get over the line than one large one. I wonder if there's some smaller chunks that can be extracted, reviewed, and merged independently. I'm not sure quite how I'd carve that out specifically here, but it's worth asking from a perspective of getting this merged easily. If this was possible, the chunks would have to be independently useful in some way. Like perhaps an initial PR might be "add a Default Role for all users set to Admin and protect the routes that need it" or something like that. Maybe then "Add user management" and "Support multi-user login" or something like that?

joshka:
I'd suggest moving the discussion about the whether this is bad for trackers over to the issue rather than the PR, as it's not really anything to do with the implementation details, and would be useful to bring up with those subscribed to the issue instead of the more limited amount of people looking at this PR.

src/NzbDrone.Core/Datastore/Migration/212_add_role_to_users.cs
```diff
@@ -0,0 +1,14 @@
+using FluentMigrator;
+using NzbDrone.Core.Datastore.Migration.Framework;
+
+namespace NzbDrone.Core.Datastore.Migration
+{
+    [Migration(212)]
```
markus101, line 6: Another PR is already using migration 212, this should be done in the same migration as adding the API key to the user.

src/NzbDrone.Core/Datastore/Migration/213_add_api_key_to_users.cs
```diff
@@ -0,0 +1,14 @@
...
+using NzbDrone.Core.Datastore.Migration.Framework;
+
+namespace NzbDrone.Core.Datastore.Migration
+{
+    [Migration(213)]
+    public class add_api_key_to_users : NzbDroneMigrationBase
+    {
+        protected override void MainDbUpgrade()
+        {
+            Alter.Table("Users").AddColumn("ApiKey").AsString().WithDefaultValue("Test");
```
markus101, line 11: Since the only user would be the admin user when this runs it should set the API key that already exists.

src/Sonarr.Http/Authentication/ApiKeyAuthenticationHandler.cs
markus101: From the UI it will be mandatory, but we should have some sort of default user created with an API key that the user can change in that setup instead of ignoring the API key.

frontend/src/FirstRun/AuthenticationRequiredModalContent.tsx
```diff
@@ -28,16 +29,18 @@ import { InputChanged } from 'typings/inputs';
 import translate from 'Utilities/String/translate';
 import styles from './AuthenticationRequiredModalContent.css';
 
-const SECTION = 'general';
+const GENERAL_SECTION = 'general';
 
-const selector = createSettingsSectionSelector(SECTION);
+const general_selector = createSettingsSectionSelector(GENERAL_SECTION);
```
markus101, line 34:
```suggestion
const generalSelector = createSettingsSectionSelector(GENERAL_SECTION);
```

src/NzbDrone.Core/Authentication/User.cs
```diff
@@ -3,12 +3,21 @@
...
     public class User : ModelBase
     {
         public Guid Identifier { get; set; }
         public string Username { get; set; }
         public string Password { get; set; }
         public string Salt { get; set; }
         public int Iterations { get; set; }
+        public UserRole Role { get; set; }
+
+        public string ApiKey { get; set; }
```
markus101, line 21:
```suggestion
        public UserRole Role { get; set; }
        public string ApiKey { get; set; }
```

src/NzbDrone.Core/Authentication/UserRepository.cs
```diff
@@ -9,6 +9,8 @@ public interface IUserRepository : IBasicRepository<User>
     {
         User FindUser(string username);
         User FindUser(Guid identifier);
+
+        User FindByApiKey(string apiKey);
```
markus101, line 13:
```suggestion
        User FindByApiKey(string apiKey);
```

src/NzbDrone.Core/Authentication/UserService.cs
```diff
@@ -13,12 +14,22 @@ namespace NzbDrone.Core.Authentication
...
         User Upsert(string username, string password);
         User FindUser();
         User FindUser(string username, string password);
         User FindUser(Guid identifier);
+        User UpdateByModel(User updatedUser);
+        public User FindUser(int id);
+
+        void Delete(int id);
+
+        public User FindUserFromApiKey(string apiKey);
```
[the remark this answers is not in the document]
markus101: The interface doesn't need to define that they are public and should be removed to match the existing methods.

src/NzbDrone.Core/Authentication/UserService.cs
```diff
@@ -13,12 +14,22 @@ namespace NzbDrone.Core.Authentication
...
+
+        List<User> All();
+
+        bool hasUsers();
```
markus101, line 32:
```suggestion
        bool HasUsers();
```

though as commented elsewhere we should have a default user instead of a wide open API with zero authentication.

src/NzbDrone.Core/Authentication/UserService.cs
```diff
@@ -30,22 +41,25 @@ public class UserService : IUserService, IHandle<ApplicationStartedEvent>
         private readonly IUserRepository _repo;
         private readonly IAppFolderInfo _appFolderInfo;
         private readonly IDiskProvider _diskProvider;
+        private readonly IEventAggregator _eventAggregator;
 
-        public UserService(IUserRepository repo, IAppFolderInfo appFolderInfo, IDiskProvider diskProvider)
+        public UserService(IUserRepository repo, IEventAggregator eventAggregator, IAppFolderInfo appFolderInfo, IDiskProvider diskProvider)
```
markus101, line 46:
```suggestion
        public UserService(IUserRepository repo, IAppFolderInfo appFolderInfo, IDiskProvider diskProvider, IEventAggregator eventAggregator)
```

src/NzbDrone.Core/Authentication/UserService.cs
```diff
@@ -56,13 +70,32 @@ public User Update(User user)
             return _repo.Update(user);
         }
 
+        // TODO: Do not allow to change username to something that already exists.
```
markus101, line 73: This can be done in API validation, seem how it's done for indexers/custom formats/etc

src/NzbDrone.Core/Authentication/UserService.cs
```diff
@@ -77,7 +110,7 @@ public User Upsert(string username, string password)
 
         public User FindUser()
         {
-            return _repo.SingleOrDefault();
+            return _repo.All().FirstOrDefault();
         }
```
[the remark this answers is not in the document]
markus101: I think this just needs to be removed in favour of `public User FindUser(int id)` below

src/Sonarr.Api.V3/Users/UserResource.cs
```diff
@@ -0,0 +1,81 @@
...
+    public class UserResource : RestResource
+    {
+        public Guid Identifier { get; set; }
+        public string Username { get; set; }
+
+        public string Password { get; set; }
+
+        public string PasswordConfirmation { get; set; }
+
+        public UserRole Role { get; set; }
```
markus101, line 18:
```suggestion
        public Guid Identifier { get; set; }
        public string Username { get; set; }
        public string Password { get; set; }
        public string PasswordConfirmation { get; set; }
        public UserRole Role { get; set; }
```

src/Sonarr.Http/Frontend/InitializeJsonController.cs
markus101: I don't think this is needed, it's only going to be called once (singleton) and instead should be looked up during the request.

markus101 (commented):
Gave it a quick review, tried not to focus on the styling too much since it's early, but let some that apply generally and will go towards an easier review.

Overall it's looking pretty good, but definitely some new concerns about how this will be perceived by indexers.

src/NzbDrone.Core/Authentication/UserService.cs
joshka: Perhaps generating a random initial password into the app.config would work (with the assumption that if you have access there then you're the superuser)

src/NzbDrone.Core/Authentication/UserService.cs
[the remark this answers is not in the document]
Alexk2309: Hi, the public accessors are needed because they are used by other classes. For example in ApiKeyAuthenticationHandler.cs

`var user = _userService.FindUserFromApiKey(providedApiKey);`

I'm getting a compile error If I remove this remove these public addresses. Unless you mean something else I am missing  @markus101.

src/NzbDrone.Core/Authentication/UserService.cs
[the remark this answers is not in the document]
Alexk2309: Hi this change was made inorder for the code to compile this was only a temporary fix for the problem.

The only time findUser() is ever used in is the HostConfigController class.

I'm thinking about refactoring, HostConfigController to remove anything to do with users and delegate the user management stuff to the new UserController class.

What do you think @markus101 , @joshka.

bakerboy448:
Additional implementation thought: Should service users be part of this? A toggle to enable / disable login at creation? So different third party api consumers can use their own distinct keys?

src/NzbDrone.Core/Authentication/UserService.cs
Alexk2309: Just realized this was already implemented by default just tested it now, getting this error which is expected `constraint failed
UNIQUE constraint failed: Users.Username`.

Would most likely need to implement frontend change to show this error.

Marking this as resolved for now.

markus101: The validations till needs to be done in the API to return an error message to the UI, relying on on the DB unique check will result in a 500 response instead of 400.

Alexk2309: Ah right sure, sounds good thanks @markus101

frontend/src/Components/Page/Sidebar/PageSidebar.js
```diff
@@ -173,10 +177,18 @@ const links = [
   }
 ];
 
+const userRole = window.Sonarr.role;
+const filteredLinks = links.filter((link) => {
+  if (link.iconName === icons.SYSTEM || link.iconName === icons.SETTINGS) {
+    return userRole === 'Admin';
+  }
+  return true;
+});
```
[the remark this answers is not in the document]
Alexk2309: Sure this was a quick solution, just wanted to see the working prototype. 

Will work on this when once I can get the more important points finshed.

frontend/src/App/ProtectedRoute.tsx
```diff
@@ -0,0 +1,36 @@
...
+  ...rest
+}: ProtectedRouteProps) {
+  const userRole: Role = window.Sonarr.role;
+
+  const renderComponent = useCallback(
+    (props: RouteComponentProps) => {
+      return allowedRoles.includes(userRole) ? (
+        <Component {...props} />
+      ) : (
+        <Redirect to="/" />
```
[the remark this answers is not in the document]
Alexk2309: I will work on this later, once I finish the more important tasks at hand.

src/NzbDrone.Core/Authentication/UserService.cs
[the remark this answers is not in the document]
joshka: I'll defer to @markus101 on this

src/Sonarr.Http/Authentication/ApiKeyAuthenticationHandler.cs
Alexk2309: Hi, @markus101 is this the direction you want to go towards?

Correct me if I'm wrong but isn't the authentication method only ever set to None before the app has been setup. So if this is because of security vulnerabilities, don't see how is any different from my method. 

In conclusion, i don't really see the point to doing this.

> default user created with an API key that the user can change in that setup instead of ignoring the API key

Could you please provide some reasons why your method is preferred compared to mine? Thanks!

joshka:
> Additional implementation thought: Should service users be part of this? A toggle to enable / disable login at creation? So different third party api consumers can use their own distinct keys?

It's probably worth keeping that out of this PR to avoid things getting too complex, but I think it's a good idea to explore once this lands. Making specific service users would make it possible to scope usage to just the necessary parts of Sonarr, which is a good idea in general.

Alexk2309:
Hi @markus101 !

I just wanted to confirm if the current method to restrict API access based on the user's role is acceptable.

Thanks!

src/NzbDrone.Core/Authentication/UserService.cs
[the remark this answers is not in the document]
markus101: Yeah, doesn't make sense to have these pieces in `HostConfigController` in a multi-user world.

src/Sonarr.Http/Authentication/ApiKeyAuthenticationHandler.cs
markus101: Ideally we'd never have Sonarr publicly exposed at all, this just makes that more obvious. `External` is going to have the same problem, there isn't an internal Sonarr configured, but that's a larger problem to solve.

markus101:
I believe so, looks like it's doing RBAC which we want.

src/NzbDrone.Core/Authentication/UserService.cs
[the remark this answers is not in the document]
joshka: Can ApiKeyAuthenticationHandler rely on IUserService instead of the concrete type?

quorn23:
This basically enables direct (usage) sharing of api/passkeys, something i can't recall being allowed by any indexer/tracker. It's a slippery slope into a lot of issues, which haven't been ones before. 

It's not real multi user either, as the datasets aren't unique and config is being shared. If that will be the case, you will still run into the issue of indexer/tracker not allowing multiple accounts access from the same origin. 

I'm genuinely curious what the advantages of this kind of multi user support would be, that is worth the risks.

Alexk2309:
Will be holding off from development until the concern of account sharing is resolved. Will start development again once the green light is given.

joshka:
> This basically enables direct (usage) sharing of api/passkeys, something i can't recall being allowed by any indexer/tracker. It's a slippery slope into a lot of issues, which haven't been ones before.
> 
> It's not real multi user either, as the datasets aren't unique and config is being shared. If that will be the case, you will still run into the issue of indexer/tracker not allowing multiple accounts access from the same origin.
> 
> I'm genuinely curious what the advantages of this kind of multi user support would be, that is worth the risks.

Can you post this over on the issue instead of here?

markus101:
Locking this for the time being while reuse is considered for scoped API keys per https://github.com/Sonarr/Sonarr/issues/1682#issuecomment-2341796770

Thank you for your efforts @Alexk2309 and @joshka

markus101:
Opened an issue for scoped API keys https://github.com/Sonarr/Sonarr/issues/7549

PR #7216, opened 2024-09-16
New: Allowed sort keys for paginated resources

mynameisbogdan:
#### Description
Prevent fatal exceptions due to bad sorting keys that require locale storage resetting to fix the issue.

### DIFF
--- a/src/Sonarr.Http/PagingResource.cs
+++ b/src/Sonarr.Http/PagingResource.cs
@@ -38,7 +38,11 @@ namespace Sonarr.Http
 
     public static class PagingResourceMapper
     {
-        public static PagingSpec<TModel> MapToPagingSpec<TResource, TModel>(this PagingResource<TResource> pagingResource, string defaultSortKey = "Id", SortDirection defaultSortDirection = SortDirection.Ascending)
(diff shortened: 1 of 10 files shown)


src/Sonarr.Http/PagingResource.cs
```diff
@@ -38,7 +38,11 @@ public PagingResource(PagingRequestResource requestResource)
 
     public static class PagingResourceMapper
     {
-        public static PagingSpec<TModel> MapToPagingSpec<TResource, TModel>(this PagingResource<TResource> pagingResource, string defaultSortKey = "Id", SortDirection defaultSortDirection = SortDirection.Ascending)
+        public static PagingSpec<TModel> MapToPagingSpec<TResource, TModel>(
+            this PagingResource<TResource> pagingResource,
+            string defaultSortKey = "Id",
+            SortDirection defaultSortDirection = SortDirection.Ascending,
+            HashSet<string> allowedSortKeys = null)
```
markus101, line 45: To avoid future issues should this be a required parameter? Otherwise passing in nothing/null/empty hash set will use whatever key is passed in (which could be anything from an API client).

PR #7225, opened 2024-09-18
LanguageParser: Support both 'ger' and 'ger dub' in language regex

NextWork123:
#### Description
Hello, I have noticed that the current regex handling for the German language is a bit restrictive. Previously, we could only use "GER" when accompanied by dubbing. However, to improve flexibility, I propose modifying the regex so that "GER" can be used both with and without dubbing.

This adjustment would ensure consistency and better alignment with the original project goals.

Please let me know if this update is feasible or if further clarification is required.

Best regards,
NextWorks


thezoggy:
it was done to prevent false positives..

NextWork123:
> it was done to prevent false positives..

To make Sonarr recognize GER, do I need to create a custom format? 

{55367738-D317-45B2-AF3B-865EC2D927EF}

markus101:
Each of those releases has audio tracks for each of those listed languages? Or are they subtitle languages?

If it's the former, then tests need to be added to validate these changes.
If it's the latter, then _not_ matching is intentional, languages are audio languages, not subtitle languages.

NextWork123:
> Each of those releases has audio tracks for each of those listed languages? Or are they subtitle languages?
> 
> If it's the former, then tests need to be added to validate these changes. If it's the latter, then _not_ matching is intentional, languages are audio languages, not subtitle languages.

How should we handle the situation when both the audio and subtitle languages are in GER?

markus101:
Subtitle languages aren't a concern for Sonarr, they're not tracked.

NextWork123:
> Subtitle languages aren't a concern for Sonarr, they're not tracked.

Got it, thank you for your time! 👍

PR #7237, opened 2024-09-22
Fixed: Sorting queue by columns

mynameisbogdan:
#### Description
Adding list of allowed sorting keys  for queue since fca8c36156da1d9591483a08beddd115c3cb2c31 made the allowed list mandatory.

### DIFF
--- a/src/Sonarr.Api.V3/Queue/QueueController.cs
+++ b/src/Sonarr.Api.V3/Queue/QueueController.cs
@@ -139,7 +139,31 @@ namespace Sonarr.Api.V3.Queue
         public PagingResource<QueueResource> GetQueue([FromQuery] PagingRequestResource paging, bool includeUnknownSeriesItems = false, bool includeSeries = false, bool includeEpisode = false, [FromQuery] int[] seriesIds = null, DownloadProtocol? protocol = null, [FromQuery] int[] languages = null, int? quality = null)
         {
             var pagingResource = new PagingResource<QueueResource>(paging);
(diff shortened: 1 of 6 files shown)


src/Sonarr.Api.V3/Queue/QueueController.cs
```diff
@@ -139,7 +139,28 @@ public object RemoveMany([FromBody] QueueBulkResource resource, [FromQuery] bool
...
+                    "languages",
+                    "quality",
+                    "protocol",
+                    "indexer",
+                    "downloadClient",
+                    "title",
+                    "size",
+                    "estimatedCompletionTime",
+                    "added",
+                    "progress"
```
markus101, line 160: If requiring them makes the most sense to fix this then I think they should be sorted, to make it easier to read.

```suggestion
                    "added",
                    "downloadClient",
                    "episode",
                    "episodes.airDateUtc",
                    "episodes.title",
                    "estimatedCompletionTime",
                    "indexer",
                    "languages",
                    "progress",
                    "protocol",
                    "quality",
                    "series.sortTitle",
                    "size",
                    "status",
                    "timeleft",
                    "title"
```

markus101 (commented):
Could do something like this in `PagingResource` instead:

```
            pagingSpec.SortKey = pagingResource.SortKey != null &&
                                 (allowedSortKeys == null || allowedSortKeys.Contains(pagingResource.SortKey))
                ? pagingResource.SortKey
                : defaultSortKey;
```

This would ensure if the `HashSet` is not provided it'll not check whether the key is present and if it's provided it must contain the key which would guard against an empty `HashSet` being passed in.

PR #7250, opened 2024-09-27
[DiskTransferService] WaitForIO after Copy/Move file

jaredledvina:
#### Description

I'm encountering an issue where the file size right after a copy is reported as smaller than the original file size. I believe this is because the remote copy destination is over NFS which takes a moment to sync the most recent writes. This change adds `WaitForIO();` after the copy and move actions before the file size check to give time for the writes to sync. `WaitForIO()` will simply sleep for 3 seconds as is done in `RollbackMove` and `RollbackPartialMove` already:  https://github.com/Sonarr/Sonarr/blob/e196c1be69593b26c49e9844efd56d4322e3b334/src/NzbDrone.Common/Disk/DiskTransferService.cs#L466-L470

### DIFF
--- a/src/NzbDrone.Common/Disk/DiskTransferService.cs
+++ b/src/NzbDrone.Common/Disk/DiskTransferService.cs
@@ -507,7 +507,7 @@ namespace NzbDrone.Common.Disk
             var targetSize = _diskProvider.GetFileSize(targetPath);
             if (targetSize != originalSize)
             {
-                _logger.Info("File {0} incomplete, waiting in case filesystem is not synchronized. [{1}] was {2} bytes long instead of the expected {3}.", action, targetPath, targetSize, originalSize));
+                _logger.Debug("File {0} incomplete, waiting in case filesystem is not synchronized. [{1}] was {2} bytes long instead of the expected {3}.", action, targetPath, targetSize, originalSize));
                 WaitForIO();
             }
             targetSize = _diskProvider.GetFileSize(targetPath);
--- a/src/NzbDrone.Common/Disk/DiskTransferService.cs
+++ b/src/NzbDrone.Common/Disk/DiskTransferService.cs
@@ -511,7 +511,7 @@ namespace NzbDrone.Common.Disk
                 return;
             }
 
-            _logger.Debug("File {0} incomplete, waiting in case filesystem is not synchronized. [{1}] was {2} bytes long instead of the expected {3}.", action, targetPath, targetSize, originalSize));
+            _logger.Debug("File {0} incomplete, waiting in case filesystem is not synchronized. [{1}] was {2} bytes long instead of the expected {3}.", action, targetPath, targetSize, originalSize);
             WaitForIO();
             targetSize = _diskProvider.GetFileSize(targetPath);
 
--- a/src/NzbDrone.Common/Disk/DiskTransferService.cs
+++ b/src/NzbDrone.Common/Disk/DiskTransferService.cs
@@ -475,6 +475,8 @@ namespace NzbDrone.Common.Disk
             {
                 _diskProvider.CopyFile(sourcePath, targetPath);
 
+                WaitForIO();
+
                 var targetSize = _diskProvider.GetFileSize(targetPath);
                 if (targetSize != originalSize)
                 {
@@ -493,6 +495,8 @@ namespace NzbDrone.Common.Disk
             try
             {
                 _diskProvider.MoveFile(sourcePath, targetPath);
+
+                WaitForIO();
 
                 var targetSize = _diskProvider.GetFileSize(targetPath);
                 if (targetSize != originalSize)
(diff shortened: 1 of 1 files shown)


markus101 (changes requested):
Instead of always delaying for 3 seconds the verification portion should be pulled out to a separate method that checks the size, if it doesn't match, call `WaitForIO()` then check again and throw if it's still a mismatch. This way users with systems not affected by this issue aren't punished for the ones that are.

jaredledvina:
@markus101 - Thanks for the review, I'm new to C# but, I have given your suggestion an shot! I added a simple `VerifyFile` that tries once, and then if it fails does the sleep & logs, and tries again before throwing the exception. Lemme know what you think!

src/NzbDrone.Common/Disk/DiskTransferService.cs
```diff
@@ -511,6 +502,21 @@ private void TryMoveFileVerified(string sourcePath, string targetPath, long orig
             }
         }
 
+        private void VerifyFile(string sourcePath, string targetPath, long originalSize, string action)
+        {
+            var targetSize = _diskProvider.GetFileSize(targetPath);
+            if (targetSize != originalSize)
+            {
+                _logger.Info("File {0} incomplete, waiting in case filesystem is not synchronized. [{1}] was {2} bytes long instead of the expected {3}.", action, targetPath, targetSize, originalSize));
```
bakerboy448, line 510: discussion comment: should this be debug instead of info?

markus101: I think Debug is more appropriate here since we're throwing if it's still an issue after the wait. I don't think we need to log how long it's being delayed for though.

```suggestion
                _logger.Debug("File {0} incomplete, waiting in case filesystem is not synchronized. [{1}] was {2} bytes long instead of the expected {3}.", action, targetPath, targetSize, originalSize));
```

src/NzbDrone.Common/Disk/DiskTransferService.cs
```diff
@@ -511,6 +502,21 @@ private void TryMoveFileVerified(string sourcePath, string targetPath, long orig
             }
         }
 
+        private void VerifyFile(string sourcePath, string targetPath, long originalSize, string action)
+        {
+            var targetSize = _diskProvider.GetFileSize(targetPath);
+            if (targetSize != originalSize)
+            {
```
markus101, line 509: New line before if block, but invert to return early if it's equal for less nesting overall.

```suggestion

            if (targetSize == originalSize)
            {
                return;
            }
```

src/NzbDrone.Common/Disk/DiskTransferService.cs
```diff
@@ -511,6 +502,21 @@ private void TryMoveFileVerified(string sourcePath, string targetPath, long orig
...
+                _logger.Info("File {0} incomplete, waiting in case filesystem is not synchronized. [{1}] was {2} bytes long instead of the expected {3}.", action, targetPath, targetSize, originalSize));
+                WaitForIO();
+            }
+            targetSize = _diskProvider.GetFileSize(targetPath);
+            if (targetSize != originalSize)
```
markus101, line 514: Similar thing here, with a new line after the if block and an early return if equal

jaredledvina: 👍 Thanks for the advice, updated!

markus101 (approved):
Thanks for the fixes, looks good once CI is happy.

src/NzbDrone.Common/Disk/DiskTransferService.cs
```diff
@@ -511,6 +502,27 @@ private void TryMoveFileVerified(string sourcePath, string targetPath, long orig
...
+        private void VerifyFile(string sourcePath, string targetPath, long originalSize, string action)
+        {
+            var targetSize = _diskProvider.GetFileSize(targetPath);
+
+            if (targetSize == originalSize)
+            {
+                return;
+            }
+
+            _logger.Debug("File {0} incomplete, waiting in case filesystem is not synchronized. [{1}] was {2} bytes long instead of the expected {3}.", action, targetPath, targetSize, originalSize));
```
markus101, line 514: Extra bracket

```suggestion
            _logger.Debug("File {0} incomplete, waiting in case filesystem is not synchronized. [{1}] was {2} bytes long instead of the expected {3}.", action, targetPath, targetSize, originalSize);
```

src/NzbDrone.Common/Disk/DiskTransferService.cs
```diff
@@ -493,12 +488,8 @@ private void TryMoveFileVerified(string sourcePath, string targetPath, long orig
             try
             {
                 _diskProvider.MoveFile(sourcePath, targetPath);
+                VerifyFile(sourcePath, targetPath, originalSize, "move");
```
markus101, line 492:
```suggestion
```

jaredledvina:
Awesome, thanks so much @markus101 !

PR #7253, opened 2024-09-27
Convert FormInputButton to TypeScript

mynameisbogdan:
#### Description
Mostly to fix: 

```
Warning: Failed prop type: Invalid prop `children` of type `string` supplied to `FormInputButton`, expected a single ReactElement.
FormInputButton@webpack-internal:///61334:23:7
Naming@webpack-internal:///15067:38:5
NamingConnector@webpack-internal:///6921:46:5
```


frontend/src/Components/Form/FormInputButton.tsx
```diff
@@ -0,0 +1,45 @@
...
+import { LinkProps } from 'Components/Link/Link';
+import SpinnerButton from 'Components/Link/SpinnerButton';
+import { kinds } from 'Helpers/Props';
+import styles from './FormInputButton.css';
+
+export interface FormInputButtonProps extends Omit<LinkProps, 'children'> {
+  children: Required<LinkProps['children']>;
+  canSpin?: boolean;
+  isLastButton?: boolean;
+}
```
markus101, line 13: `SpinnerButton` builds on top of `Button` so this can be simplified to:

```suggestion
export interface FormInputButtonProps extends ButtonProps {
  canSpin?: false;
  isLastButton?: boolean;
}
```

frontend/src/Components/Form/FormInputButton.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+function FormInputButton({
+  className = styles.button,
+  children,
```
markus101, line 17: Can pass this through via `otherProps`

markus101 (changes requested):

frontend/src/Components/Form/FormInputButton.tsx
markus101: It's not because it was passed in via `otherProps`, it's because `children` was defined as `children: PropTypes.element`, `ReactNode` (which is inherited from `ComponentPropsWithoutRef` allows for `string`

PR #7256, opened 2024-09-27
Fixed: Use Id from route and check for existence on provider update

mynameisbogdan:
#### Description
Prevent NullRef when someone tries to update an non-existing provider.

While working on this I would the fact that `id` from the body is used, and not the one from the path.

I think we should switch to the id from route eventually.

### DIFF
--- a/src/Sonarr.Api.V3/ProviderControllerBase.cs
+++ b/src/Sonarr.Api.V3/ProviderControllerBase.cs
@@ -86,9 +86,16 @@ namespace Sonarr.Api.V3
         [RestPutById]
         [Consumes("application/json")]
         [Produces("application/json")]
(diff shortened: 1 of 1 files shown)


src/Sonarr.Api.V3/ProviderControllerBase.cs
```diff
@@ -89,6 +89,12 @@ public ActionResult<TProviderResource> CreateProvider([FromBody] TProviderResour
         public ActionResult<TProviderResource> UpdateProvider([FromBody] TProviderResource providerResource, [FromQuery] bool forceSave = false)
         {
             var existingDefinition = _providerFactory.Find(providerResource.Id);
+
+            if (existingDefinition == null)
```
markus101, line 93: To your point it'd be worth checking for the ID in the route first, then the body, in API v4 we could drop the fallback.

PR #7280, opened 2024-10-06
Reordered Series Details, and split onto two rows (except for small s…

ManiMatter:
Reordered the series details.
Advantage:
- Information about the series are together (Language, Network, etc)
- Information that are user defined are together (such as save path, whether monitored, etc)
- Above break into two rows (except for small screens)
- Path is at the very end (since it is dynamic and kept pushing the other labels around, which I found distracting for the eye)

**New order:**
image

**Old order:**
image

**Compatibility:**
If ever, we wanted to move to a different way of showing the details, for instance, like in radarr, this should still work:
image

### DIFF
--- a/frontend/src/Series/Details/SeriesDetails.js
+++ b/frontend/src/Series/Details/SeriesDetails.js
@@ -199,6 +199,7 @@ class SeriesDetails extends Component {
       isSearching,
       isFetching,
       isPopulated,
+      isSmallScreen,
       episodesError,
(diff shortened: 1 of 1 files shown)


markus101:
I'm not opposed to splitting into multiple lines, since there is a lot of vertical space on the farnart that could be taken up on most items, but this makes the UI feel a little zig zaggy and the width of different content changes.

PR #7303, opened 2024-10-16
New: Labels support for Transmission

### LINKED ISSUE #7300: Proper support for labels/categories in transmission

Bittorrent download clients other than transmission support using actual labels within the client to identify sonarr torrents. in the past, Transmission did not support labels, so sonarr just set a download directory as a "fake" category. In my experience,  adding torrents manually to my download client for wanted episodes is not as successful with transmission. I believe it may be due to a lack of label support.

For some time now transmission has supported labels (some time in v3.x, it is now v4.x). Please add proper support for labels to bring support for transmission into parity with qbittorent, rtorrent and deluge. 

---

### Anything else?

_No response_

mynameisbogdan:
#### Description
Adding support for labels starting with Transmission v3.0.0 with some caveats:
- ~~v3.0.0 has torrent labels support only on `torrent-set`, labels support was added on `torrent-add` since v4.0.0 so for < v4.0.0 an extra request is mandatory.~~
- Adding support for `Post-Import Category`
- Changing the default value for `Category` to `tv-sonarr`.
- To maintain BC I didn't touched the destination directory logic so it would still create a sub-folder with the `Category`. I think it's best anyway to have a sub-folder anyway since the labels aren't like in Qbit with a save path.

* Closes #7300

### DIFF
--- a/src/NzbDrone.Common/Extensions/IEnumerableExtensions.cs
+++ b/src/NzbDrone.Common/Extensions/IEnumerableExtensions.cs
@@ -148,10 +148,5 @@ namespace NzbDrone.Common.Extensions
         {
             return string.Join(separator, source.Select(predicate));
         }
-
-        public static HashSet<T> ToHashSet<T>(this IEnumerable<T> source, IEqualityComparer<T> comparer = null)
-        {
-            return new HashSet<T>(source, comparer);
-        }
     }
 }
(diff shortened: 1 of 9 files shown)


markus101:
Given how long Transmission v4 has been released I don't think we should bother supporting v3 labels, if people want real label support they can upgrade to v4.

markus101 (approved):

PR #7304, opened 2024-10-17
Inherit trigger from pushed command models

mynameisbogdan:
#### Description
I had to debug why all `CommandQueueManager.Push` where having a `unspecified` command trigger while scheduled tasks had the right trigger, and it was because for scheduled tasks it was set [here](https://github.com/Sonarr/Sonarr/blob/86446a768627b98909389eac230720fcd00cb127/src/NzbDrone.Core/Messaging/Commands/CommandQueueManager.cs#L145). 

Same for the pushes via API due to this [part](https://github.com/Sonarr/Sonarr/blob/2f87d52a962df65d59fd5cafbfc144e78fd2e52f/src/Sonarr.Api.V3/Commands/CommandController.cs#L71).

We can set Trigger on pushed commands and would be fine, but like [this](https://github.com/Sonarr/Sonarr/blob/0e95ba2021b23cc65bce0a0620dd48e355250dab/src/NzbDrone.Core/Update/InstallUpdateService.cs#L283) the CommandTrigger is lost as we have access only to the Command body and not the CommandModel wrapper.

This would also ensure the command trigger is the same across both objects.

### DIFF
--- a/src/NzbDrone.Core/Messaging/Commands/CommandQueueManager.cs
+++ b/src/NzbDrone.Core/Messaging/Commands/CommandQueueManager.cs
@@ -106,6 +106,8 @@ namespace NzbDrone.Core.Messaging.Commands
             _logger.Trace("Publishing {0}", command.Name);
             _logger.Trace("Checking if command is queued or started: {0}", command.Name);
 
+            command.Trigger = trigger;
+
             lock (_commandQueue)
             {
                 var existingCommands = QueuedOrStarted(command.Name);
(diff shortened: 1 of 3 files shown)


markus101:
One potential source of confusion is processing downloads would be "scheduled", even though there isn't actually a scheduled task for it, but the only real way around that would be to introduce a different trigger for that and not sure it's necessary. 🤷

markus101 (approved):
Ahh, misunderstood what you were trying to solve, makes sense.

PR #7306, opened 2024-10-18
Improve message for grab errors due to no matching tags

mynameisbogdan:
#### Description
For the users using tags on everything returning the actual reason why there's no download client available.

### DIFF
--- a/src/NzbDrone.Core/Download/DownloadClientProvider.cs
+++ b/src/NzbDrone.Core/Download/DownloadClientProvider.cs
@@ -41,18 +41,23 @@ namespace NzbDrone.Core.Download
             var blockedProviders = new HashSet<int>(_downloadClientStatusService.GetBlockedProviders().Select(v => v.ProviderId));
             var availableProviders = _downloadClientFactory.GetAvailableProviders().Where(v => v.Protocol == downloadProtocol).ToList();
 
-            if (tags != null)
+            if (!availableProviders.Any())
+            {
+                return null;
+            }
+
+            if (tags is { Count: > 0 })
             {
                 var matchingTagsClients = availableProviders.Where(i => i.Definition.Tags.Intersect(tags).Any()).ToList();
 
                 availableProviders = matchingTagsClients.Count > 0 ?
                     matchingTagsClients :
                     availableProviders.Where(i => i.Definition.Tags.Empty()).ToList();
-            }
 
-            if (!availableProviders.Any())
-            {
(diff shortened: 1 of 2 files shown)


src/NzbDrone.Core/Download/DownloadClientProvider.cs
```diff
@@ -41,18 +41,23 @@ public IDownloadClient GetDownloadClient(DownloadProtocol downloadProtocol, int
...
                     matchingTagsClients :
                     availableProviders.Where(i => i.Definition.Tags.Empty()).ToList();
-            }
 
-            if (!availableProviders.Any())
-            {
-                return null;
+                if (!availableProviders.Any())
+                {
+                    throw new DownloadClientUnavailableException("No download client matching the assigned series tags nor download client without tags available. Please check your settings.");
```
zakkarry, line 59:
```suggestion
                    throw new DownloadClientUnavailableException("No download client was found with either no assigned tags or a matching series tags. Please check your settings.");
```

This was the best I could come up with, alternatives are...

No download client with applicable or no series tags was found. Please check your settings.

No download client was found without tags or associated with a series tag.  Please check your settings.

No download client could be found associated with or without series tags. Please check your settings.

No download client matched via assigned series tag or universally. Please check your settings.

Let me know if any of these seem to vibe with you @mynameisbogdan @bakerboy448 - we can tweak them further, this was just off the top.

markus101: The double `download client` ends up being quite a mouthful, maybe something closer to Zakkarry's second suggestion fits a bit better?

```suggestion
                    throw new DownloadClientUnavailableException("No download client was found without tags or a matching series tag. Please check your settings.");
```

zakkarry: Just to clarify, are you referring to the second in my body, or the second (first in the body, counting the suggestion as the first)?

Meaning

No download client was found without tags or associated with a series tag. Please check your settings.

markus101: Second in the body, `No download client was found without tags or associated with a series tag. Please check your settings.`

zakkarry: Yea, I think bogdan was saying the same, I agree that one and the suggestion's are both similar, just glad you were happy with one of them.

PR #7344, opened 2024-10-31
Fixed: Loading queue with pending releases for deleted series

mynameisbogdan:
#### Description
While we have `CleanupOrphanedPendingReleases` and a clean up on SeriesDeletedEvent, we still get cases where we get pending releases with non-existing series ids.

We should load only the pending releases with existing series by using an INNER JOIN to prevent queue loading errors.

* Closes https://github.com/Radarr/Radarr/issues/10616

### DIFF
--- a/src/NzbDrone.Core/Datastore/Extensions/BuilderExtensions.cs
+++ b/src/NzbDrone.Core/Datastore/Extensions/BuilderExtensions.cs
@@ -64,19 +64,25 @@ namespace NzbDrone.Core.Datastore
         public static SqlBuilder Join<TLeft, TRight>(this SqlBuilder builder, Expression<Func<TLeft, TRight, bool>> filter)
         {
             var wb = GetWhereBuilder(builder.DatabaseType, filter, false, builder.Sequence);
-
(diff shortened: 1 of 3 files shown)


src/NzbDrone.Core/Datastore/Extensions/BuilderExtensions.cs
```diff
@@ -79,6 +79,15 @@ public static SqlBuilder LeftJoin<TLeft, TRight>(this SqlBuilder builder, Expres
             return builder.LeftJoin($"\"{rightTable}\" ON {wb.ToString()}");
         }
 
+        public static SqlBuilder InnerJoin<TLeft, TRight>(this SqlBuilder builder, Expression<Func<TLeft, TRight, bool>> filter)
+        {
+            var wb = GetWhereBuilder(builder.DatabaseType, filter, false, builder.Sequence);
+
+            var rightTable = TableMapping.Mapper.TableNameMapping(typeof(TRight));
```
markus101, line 86:
```suggestion
            var wb = GetWhereBuilder(builder.DatabaseType, filter, false, builder.Sequence);
            var rightTable = TableMapping.Mapper.TableNameMapping(typeof(TRight));
```

PR #7345, opened 2024-11-01
Fixed: Telegram notification link text

aviadlevy:
#### Description
Small fix to the links label on telegram notification so the label would match the link itself


markus101:
Thanks for the fix!

PR #7357, opened 2024-11-04
Show a series path as example in Mount Health Check

mynameisbogdan:
#### Description
Adding a series path as example for MountCheck to make it easier to debug if the path it's in /home. 

Should help support with recent Archlinux hardening with `ProtectHome=true`.

image


markus101 (approved):

markus101:
Yeah, definitely doesn't make sense to fill the screen with all series paths.

PR #7361, opened 2024-11-04
Console warnings for missing translations on development builds

mynameisbogdan:
#### Description
I have been using ```if (!(key in translations)) { console.warn(`Missing translation for key: ${key}`); }``` lately too many times, so might be a good idea to add it to avoid missing translations.


markus101 (approved):
🙈 be nice to catch these earlier for sure.

PR #7378, opened 2024-11-10
fix: Allow `GetFileSize` to follow symlinks

eliasbenb:
#### Description

Some of the videos in my library are symlinks because they are too large to store on the disk space I have, so I download it to the cloud and mount it using [rclone](https://rclone.org) instead.

The problem with using symlinks is that Sonarr does not follow symlinks when reading file sizes. So, the videos will report as being only a few hundred KBs.

### DIFF
--- a/src/NzbDrone.Common/Disk/DiskProviderBase.cs
+++ b/src/NzbDrone.Common/Disk/DiskProviderBase.cs
@@ -189,6 +189,17 @@ namespace NzbDrone.Common.Disk
             }
 
             var fi = new FileInfo(path);
+
+            // If the file is a symlink, resolve the target path and get the size of the target file.
+            if ((fi.Attributes & FileAttributes.ReparsePoint) == FileAttributes.ReparsePoint)
+            {
+                var targetPath = fi.ResolveLinkTarget(true)?.FullName;
+                if (targetPath != null)
+                {
+                    fi = new FileInfo(targetPath);
+                }
+            }
+
             return fi.Length;
         }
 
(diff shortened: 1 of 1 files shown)


eliasbenb:
Honestly, I was trying to merge a branch in my forked repo but ended up accidentally opening a PR to the upstream Sonarr repo.

I'm not even sure if the fix works yet since I haven't fully tested it nor do I know if this is something the Sonarr community would want since symlinks are unusual.

Anyways, since I already opened the PR and can't delete it, I'll leave it open for now.

src/NzbDrone.Common/Disk/DiskProviderBase.cs
```diff
@@ -189,6 +189,17 @@ public long GetFileSize(string path)
             }
 
             var fi = new FileInfo(path);
+
+            // If the file is a symlink, resolve the target path and get the size of the target file.
+            if ((fi.Attributes & FileAttributes.ReparsePoint) == FileAttributes.ReparsePoint)
```
markus101, line 194: This can be simplified

```suggestion
            if (fi.Attributes.HasFlag(FileAttributes.ReparsePoint))
```

src/NzbDrone.Common/Disk/DiskProviderBase.cs
```diff
@@ -189,6 +189,17 @@ public long GetFileSize(string path)
...
+            {
+                var targetPath = fi.ResolveLinkTarget(true)?.FullName;
+                if (targetPath != null)
```
markus101, line 197: New line before if block

```suggestion

                if (targetPath != null)
```

markus101 (commented):
Not opposed to following the symlink here, there are some places where they are handled. Our primary reason to not handle symlinks is for the purpose of importing.

PR #7385, opened 2024-11-14
Ignore "Percona Distribution" in Postgres version string

aronsky:
Drop the ` - Percona Distribution` suffix before passing the version string to `System.Version`.

#### Description
I tried using Sonarr with Percona Everest (a k8s operator that provisions various databases, including PostgreSQL). However, Percona Everest provisions a custom version of PostgreSQL, that reports its version as "16.3 - Percona Distribution". This causes an exception when that version is passed to the `System.Version` constructor.

### DIFF
--- a/src/NzbDrone.Core/Datastore/Database.cs
+++ b/src/NzbDrone.Core/Datastore/Database.cs
@@ -52,7 +52,7 @@ namespace NzbDrone.Core.Datastore
             {
                 using var db = _datamapperFactory();
                 var dbConnection = db as DbConnection;
-                var version = Regex.Replace(dbConnection.ServerVersion, @"\(.*?\)", "");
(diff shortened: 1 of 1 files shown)


aronsky:
The full string is `16.3 - Percona Distribution`.

I agree that a long-term solution would be to properly extract the major and minor version numbers in a generic way (which would hopefully be future-proof against similar cases). However, I wanted a quick solution with minimal impact on the current (working) code - partially because I don't have a build setup for Sonarr 🙄, so I couldn't properly test more advanced changes :)

markus101:
It may be slightly more complicated, but it's the solution that we would accept instead of hard coding to remove text that could change.

markus101:
Closing in favour of https://github.com/Sonarr/Sonarr/pull/7415

PR #7397, opened 2024-11-20
Option to treat downloads with non-media extensions as failed

### LINKED ISSUE #7369: Treat Downloads with Executable Files as Failed

Undesirable files being downloaded eg .zipx and .lnk files

Looking for the ablility to define a list of file extensions e.g "*.zipx, *.lnk"to be blacklisted from being downloaded by the client. Recent spike in dodgy releases would make this very handy.

Have looked at handling this at p2p client level but seems nothing avalible there either.

### Anything else?

_No response_

markus101:
#### Description
Adds enums for decision engine rejections and the ability to handle certain rejections during import as rejections.

image

* Closes #7369


markus101:
> IMO seems too much to be an option per indexer, since most of the time you would want this for all. Adding this as a single option in settings > indexers wouldn't make the masses happy?

I considered making it global, butI think there is room to grow this feature into things that would be more appropriate as per-indexer, such as stalled torrent detection leading to failed download. Not stuck on per-indexer and I'd be curious what other people think.

> Currently it's checking just the extensions, so how about checking the mime-type too?

I don't see why we couldn't but at the moment extensions are what we have.

markus101:
The consensus on Discord was keep it per-indexer and in the future we can consider related things that will live along-side.

https://discord.com/channels/383686866005917708/383687854057979906/1311149767287509012

BeardMatt:
Since you've done this per-indexer, how does this affect people using prowlarr to manage indexers?

adude00:
I understand that having the granularity of chosing the behaviour per-index is cool, but don't most users use the software as part of a somwhat-automated ARR suite?

Personally I was waiting for an option like this to come out and I'm thankful that the devs that took their time to code it, but having it only in the indexer kinda defeats the purpose of having an automated system (not having to think about it).

I've never had to mess with indexers before, it's Prowlarr's job. It's not a big deal and I just did to enable this option, but won't it get lost once an indexer is automatically removed and re-added by prowlarr based on their avaiability?

markus101:
Agreed, bulk editing indexers (RSS, search, interactive search, plus fail download options) would be a good addition.

PR #7403, opened 2024-11-22
Deluge communication improvements

### LINKED ISSUE #7318: Improve Deluge Client API Usage

Missing functionality and API Usage improvements for Deluge's client implementation

1. https://github.com/Sonarr/Sonarr/blob/86446a768627b98909389eac230720fcd00cb127/src/NzbDrone.Core/Download/Clients/Deluge/DelugeProxy.cs#L52-L73

You can use {"method":"system.listMethods","params":[],"id":0} even when unauthenticated to the web daemon to get the available methods. By doing this, you can check for daemon.get_version or daemon.info, and determine the version without a failing RPC call.
Should assume 2.x as 1.x has not been updated at all recently, and should not even be used. Whether this is removed/deprecated is up to the team to decide. Some seedbox providers only provide 1.x, unfortunately, but should definitely not be doing this for a variety of reasons.

This can also be used to determine if label.* methods are available, rather than listing the plugins outright, I don't see any other reason why plugin listing is used.

Return is a string[] with methods listed (core.* daemon.* label.* web.* etc)

2. https://github.com/Sonarr/Sonarr/blob/develop/src/NzbDrone.Core/Download/Clients/Deluge/Deluge.cs#L215-L246

Assumes that the label is the only source for move_completed_path to be set, this can be set globally as well.

Using web.get_torrent_status will return an JSON object in result, this will return that torrents specific move_completed_path and move_completed values.

Relying on the label for this alone is not reliable, and for specific torrent information, web.get_torrent_status can give specifics, which after being stored
do not need to be polled for repeatedly. 

state will be "Moving" upon completion and once finished will return to "Seeding" or "Paused" depending on add_paused settings.

3. ~~https://github.com/Sonarr/Sonarr/blob/develop/src/NzbDrone.Core/Download/Clients/Deluge/Deluge.cs#L168-L188~~

~~https://github.com/Sonarr/Sonarr/blob/develop/src/NzbDrone.Core/Download/Clients/Deluge/DelugeTorrentStatus.cs~~

~~State of "Moving" should be accounted for and treated as "Downloading" perhaps - as this state, in my opinion, should not start an import until complete~~

4. https://github.com/Sonarr/Sonarr/blob/develop/src/NzbDrone.Core/Download/Clients/Deluge/Deluge.cs#L132-L142

This error would often be associated with a desync of the web daemon/plugin and the RPC daemon. Simply reconnecting the WebUI to the RPC daemon will remedy this.

Issuing a web.disconnect and then https://github.com/Sonarr/Sonarr/blob/develop/src/NzbDrone.Core/Download/Clients/Deluge/DelugeProxy.cs#L353 calling this should fix the issue.

We experienced this with cross-seed, and I wrote a script to accomplish a simple reconnect.

5. https://github.com/Sonarr/Sonarr/blob/develop/src/NzbDrone.Core/Download/Clients/Deluge/Deluge.cs#L38-L55

Using this https://github.com/Sonarr/Sonarr/blob/develop/src/NzbDrone.Core/Download/Clients/Deluge/DelugeProxy.cs#L176

label.get_labels can be used to get all current labels. If a label is missing it can be added back. an error for "Unknown Label" will return RPC_FAIL - ("code: 4" where a bad method will return "code: 2") 
<hr>
To summarize the reasoning behind these changes:

1. isnt really critical, but would be a better approach - avoids relying on failures to get version information
2. definitely would be better and more comprehensive way to do this - default completed paths can be set globally, this currently does not address that
3. ~~needs to be done for 2 to work properly - "Moving" state is currently not accounted for at all.~~
4. this was something @markus101 and I spoke about before, there is a function you could extract for connecting the web daemon, just needs web.disconnect for these errors to "fix" themselves
5. would just result in less errors and more concise logging as to issues, i think this would then only happen if label plugin was unloaded from deluge after startup?

N/A

### Anything else?

If any assistance is needed in testing or API usage, examples or responses, I can help with that. Just let me know here or on Discord.

markus101:
#### Description

Improves `GetVersion` and reconnecting after the daemon is disconnected.

* Closes #7318

### DIFF
--- a/src/NzbDrone.Core/Download/Clients/Deluge/Deluge.cs
+++ b/src/NzbDrone.Core/Download/Clients/Deluge/Deluge.cs
@@ -128,14 +128,9 @@ namespace NzbDrone.Core.Download.Clients.Deluge
 
             foreach (var torrent in torrents)
             {
-                // Silently ignore torrents with no hash
-                if (torrent.Hash.IsNullOrWhiteSpace())
-                {
-                    continue;
-                }
-
-                // Ignore torrents without a name, but track to log a single warning for all invalid torrents.
-                if (torrent.Name.IsNullOrWhiteSpace())
+                // Ignore torrents without a hash or name, but track to log a single warning
+                // for all invalid torrents as well as reconnect to the Daemon.
+                if (torrent.Hash.IsNullOrWhiteSpace() || torrent.Name.IsNullOrWhiteSpace())
                 {
                     ignoredCount++;
                     continue;
@@ -201,7 +196,8 @@ namespace NzbDrone.Core.Download.Clients.Deluge
 
             if (ignoredCount > 0)
             {
-                _logger.Warn("{0} torrent(s) were ignored because they did not have a title. Check Deluge and remove any invalid torrents");
+                _logger.Warn("{0} torrent(s) were ignored because they did not have a hash or title. Check Deluge and remove any invalid torrents");
+                _proxy.ReconnectToDaemon(Settings);
             }
 
             return items;
@@ -322,9 +318,9 @@ namespace NzbDrone.Core.Download.Clients.Deluge
                 return null;
             }
 
-            var enabledPlugins = _proxy.GetEnabledPlugins(Settings);
+            var methods = _proxy.GetMethods(Settings);
 
-            if (!enabledPlugins.Contains("Label"))
+            if (!methods.Any(m => m.StartsWith("label.")))
             {
                 return new NzbDroneValidationFailure("TvCategory", _localizationService.GetLocalizedString("DownloadClientDelugeValidationLabelPluginInactive"))
                 {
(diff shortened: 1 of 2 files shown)


src/NzbDrone.Core/Download/Clients/Deluge/Deluge.cs
```diff
@@ -201,7 +196,8 @@ public override IEnumerable<DownloadClientItem> GetItems()
 
             if (ignoredCount > 0)
             {
-                _logger.Warn("{0} torrent(s) were ignored because they did not have a title. Check Deluge and remove any invalid torrents");
+                _logger.Warn("{0} torrent(s) were ignored because they did not have a hash or title. Check Deluge and remove any invalid torrents");
```
Thirrian: Sorry, I don't know how to do a fancy "replace this with that" comment, but `ignoredCount` seems to be missing. (Was already missing in the past)

markus101: @zakkarry based on the warning never showing the count maybe no one ever looked at it 😆

zakkarry: lol that tracks

zakkarry: > Sorry, I don't know how to do a fancy "replace this with that" comment, but `ignoredCount` seems to be missing. (Was already missing in the past)

You'd use
````
```suggestion
what-to-replace-with
```
````

PR #7426, opened 2024-11-27
Fixed: Avoid default category on existing Transmission configurations 

### LINKED ISSUE #7424: New Tramission Default settings break existing configurations

### Current Behavior

the new default value for category on [TransmissionSettings.cs#L35](https://github.com/Sonarr/Sonarr/blob/main/src/NzbDrone.Core/Download/Clients/Transmission/TransmissionSettings.cs#L35) creates a config conflict for those who use the Directory setting
image
as the config tests fail the download client will be deactivated, likely leaving users without an active download client

### Expected Behavior

Download client keeps working without intervention

### Steps To Reproduce

1. install older sonarr (lets go with previous, 4.0.10.2544)
2. setup transmission download client
  a. don't set a Category
  b. set the Directory setting (Advanced option)
3. verify download client config works (pressing the test button should be enough)
4. save download client
5. upgrade to 4.0.11.2680
6. test download client again (you'll run into a config conflict)

### Environment

```markdown
- OS: Debian Bullseye
- Sonarr: 4.0.11.2680
- Docker Install: no
- Using Reverse Proxy: no
- Browser: n/a
- Database: Sqlite 3.34.1
```

Main

### Trace Logs?

https://privatebin.net/?234aa264f9014101#FwRnK2S9EFfXfr1Yb2iRjm3EACPNrzYtXeVka2Zv3rqh
search for "request Validation failed"

### Anything else?

_No response_

mynameisbogdan:
#### Description
Reverting the default category that was introduced in 675e3cd38a14ea33c27f2d66a4be2bf802e17d88. Sadly while this was a regression due to the fact `TvCategory` was null for some configurations and using the new default value, at this point I can't imagine a solution to revert this without manual intervention on the user side.

* Closes #7424

### DIFF
--- a/src/NzbDrone.Core/Download/Clients/Transmission/TransmissionSettings.cs
+++ b/src/NzbDrone.Core/Download/Clients/Transmission/TransmissionSettings.cs
@@ -1,3 +1,4 @@
+using System.Text.Json.Serialization;
 using System.Text.RegularExpressions;
 using FluentValidation;
 using NzbDrone.Common.Extensions;
@@ -27,12 +28,25 @@ namespace NzbDrone.Core.Download.Clients.Transmission
     {
         private static readonly TransmissionSettingsValidator Validator = new ();
 
+        // This constructor is used when creating a new instance, such as the user adding a new Transmission client.
         public TransmissionSettings()
         {
             Host = "localhost";
             Port = 9091;
             UrlBase = "/transmission/";
             TvCategory = "tv-sonarr";
+        }
+
+        // TODO: Remove this in v5
+        // This constructor is used when deserializing from JSON, it will set the
+        // category to the deserialized value, defaulting to null.
+        [JsonConstructor]
+        public TransmissionSettings(string tvCategory = null)
+        {
+            Host = "localhost";
+            Port = 9091;
+            UrlBase = "/transmission/";
+            TvCategory = tvCategory;
         }
 
         [FieldDefinition(0, Label = "Host", Type = FieldType.Textbox)]


markus101:
Give this diff a test instead of reverting the default completely

```
diff --git a/src/NzbDrone.Core/Download/Clients/Transmission/TransmissionSettings.cs b/src/NzbDrone.Core/Download/Clients/Transmission/TransmissionSettings.cs
index 17fc43bd3..bc7a7c868 100644
--- a/src/NzbDrone.Core/Download/Clients/Transmission/TransmissionSettings.cs
+++ b/src/NzbDrone.Core/Download/Clients/Transmission/TransmissionSettings.cs
@@ -1,3 +1,4 @@
+using System.Text.Json.Serialization;
 using System.Text.RegularExpressions;
 using FluentValidation;
 using NzbDrone.Common.Extensions;
@@ -27,6 +28,7 @@ public class TransmissionSettings : DownloadClientSettingsBase<TransmissionSetti
     {
         private static readonly TransmissionSettingsValidator Validator = new ();
 
+        // This constructor is used when creating a new instance, such as the user adding a new Transmission client.
         public TransmissionSettings()
         {
             Host = "localhost";
@@ -35,6 +37,17 @@ public TransmissionSettings()
             TvCategory = "tv-sonarr";
         }
 
+        // TODO: Remove this in v5
+        // This constructor is used when deserializing from JSON, it will set the
+        // category to the deserialized value, defaulting to null.
+        [JsonConstructor]
+        public TransmissionSettings(string tvCategory = null)
+        {
+            Host = "localhost";
+            Port = 9091;
+            UrlBase = "/transmission/";
+            TvCategory = tvCategory;
+        }
+
         [FieldDefinition(0, Label = "Host", Type = FieldType.Textbox)]
         public string Host { get; set; }
```

It should behave the way we want it to, existing clients will be ignored to avoid the potential conflict with `Directory`

markus101:
Ahh, pretty sure I know what I did there 😬

PR #7427, opened 2024-11-27
Do not allow upgrades if the upgrade checkbox is unnchecked, regardless of anything else

CeruleanRed:
#### Description
#7412 had fixed disallowing upgrades in cases where the CF score was higher, however I noticed that in my case the upgrade was triggered by the potential options being tagged as a revision upgrade. I changed the IsUpgradeAllowed method to always return false in the case where the 'Upgrades Allowed' checkbox is unchecked. 

If this is acceptable, I would also question having every path returning true in that method, it might perhaps be more prudent to only allow upgrades if there is something to upgrade, and changing the default path to false.

While I have added an extra test to check that false is returned in the revision case as well, I do not think that with this solution some of the tests make sense, since they would basically be checking the same path. I would say that one test that simply checks that false is returned when the checkbox is unchecked would be enough in this case

* If 'Upgrades Allowed' is unchecked, upgrades will always be denied, regardless of what else is true

### DIFF
--- a/src/Sonarr.Api.V3/openapi.json
+++ b/src/Sonarr.Api.V3/openapi.json
@@ -6970,6 +6970,29 @@
         }
       }
     },
+    "/api/v3/series/{id}/folder": {
+      "get": {
+        "tags": [
+          "SeriesFolder"
+        ],
+        "parameters": [
+          {
+            "name": "id",
+            "in": "path",
+            "required": true,
+            "schema": {
+              "type": "integer",
+              "format": "int32"
+            }
+          }
+        ],
+        "responses": {
+          "200": {
+            "description": "OK"
+          }
+        }
+      }
+    },
     "/api/v3/series/import": {
       "post": {
         "tags": [
@@ -10848,15 +10871,6 @@
             "type": "string",
             "nullable": true
           },
-          "sizeleft": {
-            "type": "number",
-            "format": "double"
-          },
-          "timeleft": {
-            "type": "string",
-            "format": "date-span",
-            "nullable": true
-          },
           "estimatedCompletionTime": {
             "type": "string",
             "format": "date-time",
@@ -10911,6 +10925,17 @@
           },
           "episodeHasFile": {
             "type": "boolean"
+          },
+          "sizeleft": {
+            "type": "number",
+            "format": "double",
+            "deprecated": true
+          },
+          "timeleft": {
+            "type": "string",
+            "format": "date-span",
+            "nullable": true,
+            "deprecated": true
           }
         },
         "additionalProperties": false


bakerboy448:
Prefer Propers/Repacks is the setting Markus refers to.

most disable it and manage with custom formats and quality simply

CeruleanRed:
I understand, would it then be acceptable to change the "Do not prefer" to also not upgrade automatically when repacks are available? This is currently achieved with "Do not upgrade", however that doesn't allow for CF scoring as far as I could find via testing/wiki. Basically turning "Do not prefer" into a sort of "customize it however you want" kind of setting even in regards to revisions

CeruleanRed:
With "Do not prefer" and "Upgrades allowed" = false, Sonarr will still auto grab revision updates/repacks, which to me feels unintuitive

markus101:
Setting it to `DoNotPrefer` means only Custom Formats will change the behaviour, which will fall under the changes in #7412.

I've added a test to confirm and it's passing

```
[Test]
public void should_return_false_if_quality_profile_does_not_allow_upgrades_and_is_proper_for_existing_quality()
{
    Mocker.GetMock<IConfigService>()
          .SetupGet(s => s.DownloadPropersAndRepacks)
          .Returns(ProperDownloadTypes.DoNotPrefer);

    GivenProfile(new QualityProfile
    {
        Cutoff = Quality.RAWHD.Id,
        Items = Qualities.QualityFixture.GetDefaultQualities(),
        UpgradeAllowed = false
    });

    GivenFileQuality(new QualityModel(Quality.WEBDL1080p, new Revision(version: 1)));
    GivenNewQuality(new QualityModel(Quality.WEBDL1080p, new Revision(version: 2)));

    Subject.IsSatisfiedBy(_parseResultSingle, null).Accepted.Should().BeFalse();
}

```

Setting the same test to `PreferAndUpgrade` or `DoNotUpgrade` fails this test, but that's expected.

CeruleanRed:
I believe in this case the unit test might not be seeing the whole picture. Here is an excerpt from my debug logs that illustrate the issue

```
2024-11-28 08:50:30.1|Debug|UpgradeDiskSpecification|Comparing file quality with report. Existing file is WEBDL-1080p v1.
2024-11-28 08:50:30.1|Debug|UpgradableSpecification|New quality is a better revision for existing quality. Existing: WEBDL-1080p v1. New: WEBDL-1080p v2
2024-11-28 08:50:30.1|Debug|UpgradableSpecification|New item's custom formats [AMZN, Repack/Proper, WEB Tier 01] (1775) improve on [WEB Tier 01] (1700), accepting
```

As you can see in my case, the current CF scoring of the file on disk is 1700. For the profile, I'm generally using trash guides, but for testing purposes I turned "Upgrades allowed" off. I also have my repack settings set to "Do not prefer". When doing an interactive search we can see that the repack is the only version that would be accepted, even though there are other versions with a higher CF score than the file on disk, those however, are rejected.

Sonarr issues no hover
Sonarr issues hover

From what I can see, this is due to how UpgradeDiskSpecification is checking for upgrades. It checks if the cutoff is not met via `CutoffNotMet`, which in the revision case is false, as `IsRevisionUpgrade` is the first check and returns true. Because of this, the `IsUpgradable` method is then asked to reject the candidate, which it does not because here we no longer check if upgrades are allowed, therefore completely bypassing weather upgrades are allowed or not. It then simply checks for a higher CF score, which exists, and the candidate is marked as acceptable. I think a similar approach to #7412 should fix it, where it simply checks against the CF minimum if upgrades are not allowed.

PR #7433, opened 2024-11-28
Bug: If a revision upgrade exists and CF scoring is higher than on disk, "upgrade allowed" is completely bypassed

CeruleanRed:
#### Description
If a revision upgrade exists and CF scoring is higher than on disk, "upgrade allowed" is completely bypassed, even if repack setting is "Do not prefer".

Continuing from #7427 with added tests to showcase the bug. The 2 tests I added are exact copies of each other. The only difference in the dummy second one, is that I made it not be a revision upgrade. If there is a revision upgrade, then the test always fails, if there isn't, it always passes.

* Revision upgrades no longer bypass "Upgrade allowed" even if set to "Do no prefer"

### DIFF
--- a/src/NzbDrone.Core.Test/DecisionEngineTests/UpgradeDiskSpecificationFixture.cs
+++ b/src/NzbDrone.Core.Test/DecisionEngineTests/UpgradeDiskSpecificationFixture.cs
@@ -5,6 +5,7 @@ using FluentAssertions;
 using Moq;
 using NUnit.Framework;
 using NzbDrone.Common.Serializer;
+using NzbDrone.Core.Configuration;
 using NzbDrone.Core.CustomFormats;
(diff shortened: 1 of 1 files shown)


markus101:
Thanks for the unit test, I completely missed the custom formats part of the problem, but see what the issue is and where it can be fixed in a more general way after https://github.com/Sonarr/Sonarr/pull/7397 lands

CeruleanRed:
Glad I could help. Should I close this and make a bug ticket instead, or would you rather keep this open and add any changes here? I'd also be happy to make the necessary changes if you let me know the general direction of the fix

markus101:
I ended up making some additional changes to `QueueSpecification` as well, but closing in favour of https://github.com/Sonarr/Sonarr/pull/7460

PR #7434, opened 2024-11-28
New: Additional language fields in webhook notification

### LINKED ISSUE #7421: Add Detected and Original Language to Webhook Notification

Currently, the webhook notification does not send the parsed/detected language or the original language in its webhook notifications

The detected language be added to the `release` object of the grab and the `episodeFile` object of the download/upgrade payloads, with original language available within the meta

You can run a release title through the parser to determine the parsed language for a grab, but with import, the logic ends up utilizing the Media Info which makes it harder.

### Anything else?

_No response_

stevietv:
#### Description
Adds OriginalLanguage to the series object, and Languages string to the episode file and release objects

* Closes #7421

### DIFF
--- a/src/NzbDrone.Core/Notifications/Webhook/WebhookEpisodeFile.cs
+++ b/src/NzbDrone.Core/Notifications/Webhook/WebhookEpisodeFile.cs
@@ -1,4 +1,5 @@
 using System;
+using NzbDrone.Common.Extensions;
 using NzbDrone.Core.MediaFiles;
 
 namespace NzbDrone.Core.Notifications.Webhook
@@ -20,6 +21,7 @@ namespace NzbDrone.Core.Notifications.Webhook
             SceneName = episodeFile.SceneName;
             Size = episodeFile.Size;
             DateAdded = episodeFile.DateAdded;
+            Languages = episodeFile.Languages.SelectList(language => language.Name).ConcatToString("/");
 
             if (episodeFile.MediaInfo != null)
             {
@@ -36,6 +38,7 @@ namespace NzbDrone.Core.Notifications.Webhook
         public string SceneName { get; set; }
         public long Size { get; set; }
         public DateTime DateAdded { get; set; }
+        public string Languages { get; set; }
         public WebhookEpisodeFileMediaInfo MediaInfo { get; set; }
         public string SourcePath { get; set; }
         public string RecycleBinPath { get; set; }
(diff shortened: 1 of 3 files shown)


src/NzbDrone.Core/Notifications/Webhook/WebhookEpisodeFile.cs
```diff
@@ -20,6 +21,7 @@ public WebhookEpisodeFile(EpisodeFile episodeFile)
             SceneName = episodeFile.SceneName;
             Size = episodeFile.Size;
             DateAdded = episodeFile.DateAdded;
+            Languages = episodeFile.Languages.SelectList(language => language.Name).ConcatToString("/");
```
austinwbest, line 24: CustomFormats, AudioLanguages, Subtitles, Images, etc are all lists so just a thought to make this a list to be consistent

markus101: Agreed, we're not limited to simple strings like we are for scripts

markus101 (changes requested):

src/NzbDrone.Core/Notifications/Webhook/WebhookEpisodeFile.cs
stevietv: thanks for the feedback, swapped these over to Language lists

stevietv: Makes sense to also remove the ToString for Original language too.

I would also kinda agree that we could pass all these languages as list of language names (strings) instead of lists of language objects, like we do for custom formats for example.
Interested what the others think

src/NzbDrone.Core/Notifications/Webhook/WebhookRelease.cs
```diff
@@ -31,5 +33,6 @@ public WebhookRelease(QualityModel quality, RemoteEpisode remoteEpisode)
         public long Size { get; set; }
         public int CustomFormatScore { get; set; }
         public List<string> CustomFormats { get; set; }
+        public List<Language> Languages { get; set; }
```
markus101, line 36: Just making sure, but this serializes the whole Language object right and not just the ID like we have when storing in the DB?

stevietv: yep, it is a list of Language objects, so `id` and `name`

eg:

```json

{
  "series": {
    ...
    "originalLanguage": {
      "id": 1,
      "name": "English"
    }
  },
   ...
  "episodeFile": {
   ...
    "languages": [
      {
        "id": 1,
        "name": "English"
      }
    ],
   ...
  ],
  "customFormatInfo": {
    "customFormats": [],
    "customFormatScore": 0
  },
...
}
```

markus101 (approved):

Issue #6534, opened 2024-02-21
Match local networks when determining if a client is local


brianjmurrell:
### Is there an existing issue for this?

- [X] I have searched the existing open and closed issues

### Is your feature request related to a problem? Please describe

When I connect to my Sonarr instance on my LAN using it's Internet-routable (IPv6, but would apply equally to IPv4 -- I just don't have an Internet routable IPv4 network assigned to me) address I am required to authenticate even though I have opted to not require authentication for _local addresses_.

### Describe the solution you'd like

https://github.com/Sonarr/Sonarr/blob/main/src/NzbDrone.Common/Extensions/IpAddressExtensions.cs should be expanded to match the client address against all address/netmask pairs assigned to interfaces on the Sonarr machine for a match.

In fact once that is done, the matching of unroutable and link-local networks can be eliminated because any of those that apply will be assigned to an interface on the machine and the above proposed matching will apply to them automagically.

### Describe alternatives you've considered

There are none.

markus101:
If you're connecting via an internet routable IPv6 address then Sonarr's detection is correct as it's not a _local_ IP address. If Sonarr included all IP addresses assigned to network interfaces on the machine it's running on then it'd effectively require no authentication for requests made to a public IP address if it was directly on the internet, like a seedbox, VPS or dedicated machine.

brianjmurrell:
> If you're connecting via an internet routable IPv6 address then Sonarr's detection is correct as it's not a _local_ IP address.

I would beg to differ.  You seem to be conflating two different concepts:
- routable/non-routable addressing
- local network determination

They are not the same thing.  The former can be used as a proxy for the latter but even that is not really precise.

The definition of a _local address_ is an address that is in the same logical network.  I.e. determining if the address of the client is within any of the ip/netmask pairs assigned on the server regardless of whether they are routable or not because those addresses **have** to be on the same subnet (i.e. local LAN) to even work.

> If Sonarr included all IP addresses assigned to network interfaces on the machine it's running on then it'd effectively require no authentication for requests made to a public IP address if it was directly on the internet, like a seedbox, VPS or dedicated machine.

This is completely un-true.  Even if the Sonarr machine were on a VPS with a publicly routable IP address, any client that is not on the same subnet (i.e. that were somewhere else on the Internet) would not match any of the ip-address/netmask pairs on the Sonarr machine and would therefore be subject to authentication.

[...]

PR #7436, opened 2024-12-01
New: Add opt-in CGNAT support for authentication bypass

s0up4200:
#### Description

Added support for CGNAT IP ranges (100.64.0.0/10) to be optionally recognized as trusted addresses, allowing authentication bypass when accessing Sonarr through CGNAT networks (like Tailscale), consistent with other local network behavior.

Changes include:

- Added CGNAT validation
- Made CGNAT support configurable via config or `SONARR__AUTH__TRUSTCGNATIPADDRESSES` environment variable
- Updated test cases
- Maintains existing behavior for local IP ranges

The CGNAT support is disabled by default and must be explicitly enabled through configuration or environment variable.

- None

### DIFF
--- a/src/NzbDrone.Common.Test/ExtensionTests/IPAddressExtensionsFixture.cs
+++ b/src/NzbDrone.Common.Test/ExtensionTests/IPAddressExtensionsFixture.cs
@@ -8,26 +8,11 @@ namespace NzbDrone.Common.Test.ExtensionTests
     [TestFixture]
     public class IPAddressExtensionsFixture
     {
-        [OneTimeSetUp]
-        public void Setup()
-        {
-            Environment.SetEnvironmentVariable("SONARR_TRUST_CGNAT", "true");
-        }
-
-        [OneTimeTearDown]
-        public void Cleanup()
-        {
-            Environment.SetEnvironmentVariable("SONARR_TRUST_CGNAT", null);
-        }
-
         [TestCase("::1")]
         [TestCase("10.64.5.1")]
         [TestCase("127.0.0.1")]
         [TestCase("172.16.0.1")]
         [TestCase("192.168.5.1")]
-        [TestCase("100.64.0.1")]
-        [TestCase("100.127.255.254")]
-        [TestCase("100.100.100.100")]
         public void should_return_true_for_local_ip_address(string ipAddress)
         {
             IPAddress.Parse(ipAddress).IsLocalAddress().Should().BeTrue();
@@ -36,9 +21,9 @@ namespace NzbDrone.Common.Test.ExtensionTests
         [TestCase("1.2.3.4")]
         [TestCase("172.55.0.1")]
         [TestCase("192.55.0.1")]
-        [TestCase("100.63.255.255")]
-        [TestCase("100.128.0.0")]
-        [TestCase("100.0.0.1")]
+        [TestCase("100.64.0.1")]
+        [TestCase("100.127.255.254")]
+        [TestCase("100.100.100.100")]
         public void should_return_false_for_public_ip_address(string ipAddress)
         {
             IPAddress.Parse(ipAddress).IsLocalAddress().Should().BeFalse();
--- a/src/NzbDrone.Common.Test/ExtensionTests/IPAddressExtensionsFixture.cs
+++ b/src/NzbDrone.Common.Test/ExtensionTests/IPAddressExtensionsFixture.cs
@@ -23,10 +23,26 @@ namespace NzbDrone.Common.Test.ExtensionTests
         [TestCase("192.55.0.1")]
         [TestCase("100.64.0.1")]
         [TestCase("100.127.255.254")]
-        [TestCase("100.100.100.100")]
         public void should_return_false_for_public_ip_address(string ipAddress)
         {
             IPAddress.Parse(ipAddress).IsLocalAddress().Should().BeFalse();
+        }
+
+        [TestCase("100.64.0.1")]
+        [TestCase("100.127.255.254")]
+        [TestCase("100.100.100.100")]
+        public void should_return_true_for_cgnat_ip_address(string ipAddress)
+        {
+            IPAddress.Parse(ipAddress).IsCgnatIpAddress().Should().BeTrue();
+        }
+
+        [TestCase("1.2.3.4")]
+        [TestCase("192.168.5.1")]
+        [TestCase("100.63.255.255")]
+        [TestCase("100.128.0.0")]
+        public void should_return_false_for_non_cgnat_ip_address(string ipAddress)
+        {
+            IPAddress.Parse(ipAddress).IsCgnatIpAddress().Should().BeFalse();
         }
     }
 }
--- a/src/NzbDrone.Common.Test/ExtensionTests/IPAddressExtensionsFixture.cs
+++ b/src/NzbDrone.Common.Test/ExtensionTests/IPAddressExtensionsFixture.cs
@@ -8,6 +8,18 @@ namespace NzbDrone.Common.Test.ExtensionTests
     [TestFixture]
     public class IPAddressExtensionsFixture
     {
+        [OneTimeSetUp]
+        public void Setup()
+        {
+            Environment.SetEnvironmentVariable("SONARR_TRUST_CGNAT", "true");
+        }
+
+        [OneTimeTearDown]
+        public void Cleanup()
+        {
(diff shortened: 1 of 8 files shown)


src/NzbDrone.Common/Extensions/IpAddressExtensions.cs
```diff
@@ -50,7 +50,10 @@ private static bool IsLocalIPv4(byte[] ipv4Bytes)
             // Class C private range: 192.168.0.0 – 192.168.255.255 (192.168.0.0/16)
             bool IsClassC() => ipv4Bytes[0] == 192 && ipv4Bytes[1] == 168;
 
-            return IsLinkLocal() || IsClassA() || IsClassC() || IsClassB();
+            // Tailscale CGNAT range: 100.64.0.0 - 100.127.255.255 (100.64.0.0/10)
+            bool IsTailscaleCGNAT() => ipv4Bytes[0] == 100 && ipv4Bytes[1] >= 64 && ipv4Bytes[1] <= 127;
```
Gylesie, line 54: Why use a non-static lambda? AFAIK, this would result in an unnecessary heap allocation for the variable capture inside a closure. Using a simple bool local variable would suffice.

A tiny note: maybe check the IP ranges with bit masking? As that feels more natural for IP ranges, and they have been defined based on a binary representation.

More of a general review question, not scoped to this PR particularly, as I can see the previous code is done in the same way.

bakerboy448:
Previously denied.

https://github.com/Sonarr/Sonarr/issues/6534#issuecomment-1957930557

s0up4200:
> Previously denied.
> 
> [#6534 (comment)](https://github.com/Sonarr/Sonarr/issues/6534#issuecomment-1957930557)

While they are somewhat similar, it's not really comparable in my opinion. Your tailnet is private to you by design, until you invite someone to join it.

markus101:
Tailscale may be private, but ISPs using CGNAT aren't.

Since CGNAT is used by ISPs, someone behind CGNAT opening up their firewall/port forwarding from the outside into their network would suddenly allow anyone in their IP range to connect to their Sonarr, without authentication the same way it'd happen with that previous discussion and a publicly routable IP.

The CGNAT IPs themselves aren't publicly routable, so people on different ISPs can't talk to each other via those IPs, but your local router/firewall will pass requests through from the public side, with a source IP of a CGNAT IP and a nefarious IP-space neighbour could be running scans, like you see with this user testing things. https://www.reddit.com/r/HomeNetworking/comments/oz7l57/is_my_routerdevices_exposed_to_other_users_within/

s0up4200:
You make a good point there actually.

We could use an environment variable to explicitly opt-in to trusting the CGNAT range. Is that something you'd be willing to allow as an alternative?

If not, feel free to close this, unless someone has a better suggestion 👍

markus101:
A config file setting (which also allows environment variables to be used) seems reasonable, though it should not explicitly be for Tailscale because it cannot differentiate between Tailscale or another CGNAT source.

src/NzbDrone.Common/Extensions/IpAddressExtensions.cs
```diff
@@ -5,6 +5,15 @@ namespace NzbDrone.Common.Extensions
 {
     public static class IPAddressExtensions
     {
+        private static readonly IConfigService _configService;
+        private static readonly bool TrustCGNAT;
+
+        static IPAddressExtensions()
+        {
+            _configService = ServiceFactory.Instance.GetInstance<IConfigService>();
```
markus101, line 13: Extensions shouldn't refer to services.

src/NzbDrone.Common/Extensions/IpAddressExtensions.cs
```diff
@@ -50,7 +59,10 @@ private static bool IsLocalIPv4(byte[] ipv4Bytes)
             // Class C private range: 192.168.0.0 – 192.168.255.255 (192.168.0.0/16)
             bool IsClassC() => ipv4Bytes[0] == 192 && ipv4Bytes[1] == 168;
 
-            return IsLinkLocal() || IsClassA() || IsClassC() || IsClassB();
+            // CGNAT range: 100.64.0.0 - 100.127.255.255 (100.64.0.0/10)
+            bool IsCGNAT() => ipv4Bytes[0] == 100 && ipv4Bytes[1] >= 64 && ipv4Bytes[1] <= 127;
+
+            return IsLinkLocal() || IsClassA() || IsClassC() || IsClassB() || (TrustCGNAT && IsCGNAT());
```
markus101, line 65: This is no longer checking if it's a local IP and should be done elsewhere, probably where `IsLocalAddress` is called from.

src/NzbDrone.Core/Configuration/IConfigService.cs
```diff
@@ -94,5 +94,8 @@ public interface IConfigService
 
         CertificateValidationType CertificateValidation { get; }
         string ApplicationUrl { get; }
+
+        // Network
+        bool TrustCGNAT { get; set; }
```
markus101, line 99: For config file/environment variable this should be in `IConfigFileProvider`, the name should follow existing naming, `TrustCgnat`

src/Sonarr.Api.V3/Config/HostConfigResource.cs
```diff
@@ -89,7 +90,8 @@ public static HostConfigResource ToResource(this IConfigFileProvider model, ICon
                 BackupFolder = configService.BackupFolder,
                 BackupInterval = configService.BackupInterval,
                 BackupRetention = configService.BackupRetention,
-                ApplicationUrl = configService.ApplicationUrl
+                ApplicationUrl = configService.ApplicationUrl,
+                TrustCGNAT = configService.TrustCGNAT
```
markus101, line 94: At this time it shouldn't be part of the API/UI.

markus101 (changes requested):

src/Sonarr.Http/Authentication/UiAuthorizationHandler.cs
```diff
@@ -27,16 +27,35 @@ protected override Task HandleRequirementAsync(AuthorizationHandlerContext conte
...
+                    {
+                        context.Succeed(requirement);
+                    }
                 }
             }
 
             return Task.CompletedTask;
         }
 
+        private bool IsCGNATAddress(IPAddress ipAddress)
```
markus101, line 43: This should be part of `IPAddressExtensions` as `IsCgnatIpAddress`

src/NzbDrone.Common.Test/ExtensionTests/IPAddressExtensionsFixture.cs
```diff
@@ -21,6 +21,9 @@ public void should_return_true_for_local_ip_address(string ipAddress)
         [TestCase("1.2.3.4")]
         [TestCase("172.55.0.1")]
         [TestCase("192.55.0.1")]
+        [TestCase("100.64.0.1")]
+        [TestCase("100.127.255.254")]
```
markus101, line 25: 2 new tests for CGNAT IPs should be added, one to make sure they're not considered local and another for testing that they are CGNAT.

src/NzbDrone.Core/Configuration/ConfigFileProvider.cs
```diff
@@ -65,6 +65,7 @@ public interface IConfigFileProvider : IHandleAsync<ApplicationStartedEvent>,
         string PostgresPassword { get; }
         string PostgresMainDb { get; }
         string PostgresLogDb { get; }
+        bool TrustCgnat { get; }
```
markus101, line 68: I think `TrustCgnatIpAddresses` to be extra explicit would be good.

src/Sonarr.Http/Authentication/UiAuthorizationHandler.cs
```diff
@@ -27,16 +27,35 @@ protected override Task HandleRequirementAsync(AuthorizationHandlerContext conte
...
+        {
+            if (ipAddress.IsIPv4MappedToIPv6)
+            {
+                ipAddress = ipAddress.MapToIPv4();
+            }
+
+            if (ipAddress.AddressFamily != AddressFamily.InterNetwork)
```
markus101, line 50: I think you can skip this check and just check the IP address bytes

markus101 (changes requested):
Looking better, just some cleanup and some more tests and it should be good.

markus101 (approved):
Thanks for the clean up, looks good.

s0up4200:
Thank you for the assist and approval👍

PR #7437, opened 2024-12-01
Fixed: Log hardlink creation errors as warnings in Mono environment

Darion:
#### Description

It is confusing when hardlink is not created, setting for hardlink is set, and you see nothing in Info level logs. It's like everything works like expected.
I spend a lot of time debugging my setup, it would be much easier with right warnings. Hope it helps someone else.

### DIFF
--- a/src/NzbDrone.Mono/Disk/DiskProvider.cs
+++ b/src/NzbDrone.Mono/Disk/DiskProvider.cs
@@ -452,14 +452,14 @@ namespace NzbDrone.Mono.Disk
                 }
                 else
                 {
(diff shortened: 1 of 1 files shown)


markus101:
It's best effort hardlinking and enabled by default, which means a ton of users are going to have their log files suddenly filling with warnings, which is why it's intentionally a debug level event, not warning currently.

If people are debugging why their hard links aren't working it's appropriate for them to login  at debug logs.

markus101:
It is intentional at this time, though in the future they could be a hardlink-only mode that would make sense to log errors about.

PR #7443, opened 2024-12-01
New: Bump to .NET 8

mynameisbogdan:


markus101:
Do we just need to wait for 8.0.12 to be available for BSD?

niklaskorz:
> I understand the desire to make your build work on NixOS, but please understand that the .NET bump might be a breaking change on some older operating systems and personally I would like avoid doing that in a minor version if possible.
> 
> While Sonarr doesn't build for linux-x86 anymore, Radarr and others still do. Currently the patched dotnet-sdk 8 for linux-x86 has memory leaks and we plan to remove support for this platform in the next major version.

To clarify, you mean x86 as in 32-bit, not on x86_64 / 64-bit Linux? From a package distributors perspective we are fine with maintaining this patch inside nixpkgs ontop of v4 ourselves, so you do you. But I still have to point out that it's a bad idea to only have a stable version around that depends on an insecure .NET release.

markus101:
This will not be backported to v4, upgrading .net (2 major versions) is a breaking change and while .net 6 no longer supports some older Windows versions they are functional today, we do not know the outcome for that with .net 8 and will not be throwing any semblance of semver out the window to test that out.

PR #7447, opened 2024-12-02
Cleanup unused metadatas connector

mynameisbogdan:
#### Description
Removing leftover connector and translating root folder tooltip for edit series path.


markus101:
I'd be good with an overall increase there.

PR #7462, opened 2024-12-07
Refine localization string for IndexerSettingsFailDownloadsHelpText

stevietv:
The current string didn't read very nicely, tried to refine it to clarify what the purpose is


bakerboy448:
can add a wiki link helptext to https://wiki.servarr.com/sonarr/settings#extension-details

stevietv:
Thanks for adding it to the wiki, we don't actually link to the wiki from any of the settings pages or modals, not sure we would want to deviate from that for this setting only to add a link here.

PR #7477, opened 2024-12-11
Convert Calendar to TypeScript

markus101:
#### Description

Converts the Calendar and all it's pieces to TypeScript along with upgrades to TypeScript and core-js so `Object.groupBy` is available.


markus101:
So not polyfilled automatically because ESR has been updated recently and no longer falls under the automatic polyfilling?

Is there an additional rule we can add to browserlist so it polyfills for Firefox ESR 115 as well? Eventually that's going to start failing when a new ESR is available, but would give people some time to move to a newer browser.

A manual polyfill is going to get stale eventually and not likely to be cleaned up

markus101:
Old issue about ESR, but a discussion about last two versions was here: https://github.com/browserslist/browserslist/issues/148

Not sure if we can still do that though.

markus101:
Ahh too bad, hardcoding it is then I guess. Should be able to conditionally include it so it won't actually apply if present.

PR #7489, opened 2024-12-15
Upgrade babel and Font Awesome

markus101:
#### Description

A few package updates


markus101:
Don't need to run lint manually very often, but I do see that warning. I've upgraded those packages as well.

PR #7517, opened 2024-12-24
Raw search: Fix non-anime seasons and episodes not searching with clean titles

meharryp:
#### Description
#6844 attempted to fix this previously but it looks like it was only added to searches for Anime. This change extends that fix to episodes and seasons.


markus101:
Thanks for the fix!

PR #7525, opened 2024-12-28
Suggest adding IP to RPC whitelist for on failed Transmission auth

mynameisbogdan:
#### Description
Suggest adding the source IP to RPC whitelist on `403` since some consider it to be an issue with the arrs and not Transmission.

### DIFF
--- a/src/NzbDrone.Core/Download/Clients/Transmission/TransmissionProxy.cs
+++ b/src/NzbDrone.Core/Download/Clients/Transmission/TransmissionProxy.cs
@@ -7,6 +7,7 @@ using System.Net;
 using Newtonsoft.Json.Linq;
 using NLog;
 using NzbDrone.Common.Cache;
+using NzbDrone.Common.EnvironmentInfo;
 using NzbDrone.Common.Extensions;
 using NzbDrone.Common.Http;
 using NzbDrone.Common.Serializer;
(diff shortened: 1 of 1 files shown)


src/NzbDrone.Core/Download/Clients/Transmission/TransmissionProxy.cs
```diff
@@ -273,24 +273,26 @@ private void AuthenticateClient(HttpRequestBuilder requestBuilder, TransmissionS
...
-                else
+                switch (response.StatusCode)
                 {
-                    throw new DownloadClientAuthenticationException("Failed to authenticate with Transmission.");
+                    case HttpStatusCode.MovedPermanently:
+                        var url = response.Headers.GetSingleValue("Location");
+
+                        throw new DownloadClientException("Remote site redirected to " + url);
+                    case HttpStatusCode.Forbidden:
+                        throw new DownloadClientException("Failed to authenticate with Transmission. It may be necessary to add your IP address to RPC whitelist.");
```
bakerboy448, line 284:
```suggestion
                        throw new DownloadClientException("Failed to authenticate with Transmission. It may be necessary to add your IP address to RPC whitelist.");
```

Should we specify sonarr's/apps's IP?

markus101: I think narrowing down which IP should be added would be good, not sure what option we have to keep it generic between different arr apps.

PR #7529, opened 2024-12-30
Check if backup folder is writable on backup

mynameisbogdan:
#### Description
Adding a little more information about the backup folder and path on creating a new backup.

As reference https://github.com/Radarr/Radarr/issues/10792, the backup folder seems to be `/home/(removed)/.config/Radarr/Backups` but SharpZipLib is seeing a difference between in and out.

### DIFF
--- a/src/NzbDrone.Common/ArchiveService.cs
+++ b/src/NzbDrone.Common/ArchiveService.cs
@@ -42,17 +42,18 @@ namespace NzbDrone.Common
 
         public void CreateZip(string path, IEnumerable<string> files)
         {
-            using (var zipFile = ZipFile.Create(path))
+            _logger.Debug("Creating archive {0}", path);
+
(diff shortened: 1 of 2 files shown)


src/NzbDrone.Common/ArchiveService.cs
```diff
@@ -42,17 +42,18 @@ public void Extract(string compressedFile, string destination)
...
-                    zipFile.Add(file, Path.GetFileName(file));
-                }
+            using var zipFile = ZipFile.Create(path);
 
-                zipFile.CommitUpdate();
+            zipFile.BeginUpdate();
+
+            foreach (var file in files)
+            {
+                zipFile.Add(file, Path.GetFileName(file));
```
bakerboy448, line 53: should we trace log each file?

Issue #6495, opened 2024-02-09
Improve Searching For Anime Seasons


Ravencentric:
### Is there an existing issue for this?

- [X] I have searched the existing open and closed issues

### Is your feature request related to a problem? Please describe

`Series Type: Anime` makes it so Sonarr searches every episode times every alias for every season of an Anime. This was probably necessary long long ago but in modern anime scene, this does nothing but slow down searches and waste api hits. For a 24 episode anime, it took Sonarr 10 mins on Nyaa and 4 mins on AB, this number usually varies depending on the number of episodes and alias but you get the gist of the problem.

This behavior is based on the philosophy that "anime has no seasons" which isn't true, atleast hasn't been true for years. Anime has seasons, They usually will either use SXX/Season X or more commonly, "Season Titles" to declare seasons. For example, a season pack may use a season title to differentiate itself, i.e,  `[group] Dr. Stone: Stone Wars [...]` ([xem has this info](https://thexem.info/xem/show/5773), https://github.com/Sonarr/Sonarr/issues/5195) where `Stone Wars` is the title of the second season.

To better demonstrate my point, consider the two "main" trackers for anime:

1. AnimeBytes does not allow anything that isn't a pack at all, even airing episodes are all under a single group. Searching every episode of an anime on AnimeBytes is such a massive waste because of their strict grouping and season pack policies. The [SysOp of AB himself addressed how AB only allows season packs](https://github.com/Prowlarr/Prowlarr/pull/1573#issuecomment-1497869996). Prowlarr already realizes this and strips out the episode number from the search and caches the results so when Sonarr hits AB for the 100th time, it gives the same cached result again and again till Sonarr exhausts all the episode and alias combos but this still ends up adding a weird artificial slowness to the whole process. Sonarr could have had done the same thing by just searching the season aliases in a few seconds.

2. Nyaa, the biggest anime public tracker (and possibly the only relevant one), is the source of most anime. Nyaa is where most content originates and then gets cross posted to AB. So practically, Nyaa has most of the season packs that AB has, albeit with worse retention. Once again, Sonarr searching every alias of every episode ends up being a massive waste of both API hits and anime (which can sometimes get you 429'd). This takes even longer than AB due to no Prowlarr magic behind it, causing a 24 episode anime to take 10 whole mins to search.

Q. But One Piece/Pokemon/{insert old and long anime here}?
[...]

PR #7556, opened 2025-01-08
Don't search individual episodes when searching for anime seasons

felizk:
#### Description
Changes the anime season search implementation to no longer search for all episodes in a season individually. Searching for each episode makes searching for anime seasons prohibitively slow and doesn't result in any better search results (see issue #6495)

This PR is a relatively naive fix for the issue, which simply removes the episode searching without fixing any of the other things mentioned in #6495

### DIFF
--- a/src/NzbDrone.Core.Test/IndexerSearchTests/ReleaseSearchServiceFixture.cs
+++ b/src/NzbDrone.Core.Test/IndexerSearchTests/ReleaseSearchServiceFixture.cs
@@ -332,23 +332,6 @@ namespace NzbDrone.Core.Test.IndexerSearchTests
         }
 
         [Test]
-        public async Task season_search_for_anime_should_search_for_each_monitored_episode()
-        {
-            WithEpisodes();
-            _xemSeries.SeriesType = SeriesTypes.Anime;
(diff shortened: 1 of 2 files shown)


bakerboy448:
Anime does not have seasons, as explained in the docs

markus101:
The likelihood of this leading to other issues and breaking the known behaviour is too high at this time so it's not something we're going to land. Potentially something to look at for v5 down the road, but we'd need to weigh the pros/cons vs doing it properly.

felizk:
While this does look too simple to be the fix, I thought the people in the issue thread weighed up the pros and cons pretty well. Without this change season searching for anime is all but useless as it takes too long. 

The loss of functionality is seen primarily in very long running anime (such as One Piece) in which  tvdb does assign seasons, but releases would not use those seasons but rather just "One piece - 311".

As explained in the issue, that is not the regular use case anymore for anime as most anime now follow a 12 or 24 episode season structure.

I think everyone in that thread would agree having a good season search is better than catering for a now corner case use case. 

That said, I don't have the full visibility on how and when season search is executed. For interactive search, this change is miles better.

Issue #5470, opened 2023-03-04
Add reflink support for other Arches


Jannik2099:
### Is there an existing issue for this?

- [X] I have searched the existing open and closed issues

### Is your feature request related to a problem? Please describe

Sonarr supports reflinks since commit 2c286f7, but it uses hardcoded ioctl numbers and only supports x86_64 - Sonarr is seeing increased use on other ISAs, namely arm and arm64 aswell. Could reflink support be extended to those?

### Describe the solution you'd like

Adding proper logic to look up the number of the ioctl would be the "clean" solution, but frankly just extending the hardcoded lookup to arm and arm64 would suffice.

### Describe alternatives you've considered

.

Jannik2099:
Better yet, perhaps Sonarr should just use the `copy_file_range` syscall? This works for btrfs, xfs, soon zfs.

markus101:
Not likely we'll expand the support to other architectures as-is, this was originally for mono and with the move to .net in v4 there may be better alternatives.

https://github.com/microsoft/CopyOnWrite
https://github.com/dotnet/runtime/pull/64264

This will require someone that is really motivated to implement this and will also need a lot of testing.

calebj:
> This works for btrfs, xfs, soon zfs.
It'll also be accelerated by offload on BcacheFS, CIFS, NFS 4.2+, and even CephFS. Since the application doesn't need to see the data, the kernel has the opportunity to offload the copy however it can, and on these specific filesystems, the kernel doesn't even need to see the data to copy it. That's what copy_file_range is all about.

The implementation markus101 linked in the .NET runtime looks like it'll attempt the right thing in `File.Copy()`, including on OSX, and on Linux i386, ARM and ARM64. It even handles the buggy implementation for some kernel versions. Win32 `CopyFile()` probably does the right thing on ReFS, but I couldn't find any information on it.

Since Sonarr v4 is moving to .NET anyway, I think dropping in `File.Copy()` should Just Work... but only users running .NET 7 and up would see the improvements that come with `copy_file_range()`.

johnpyp:
For reference wrt ZFS, ZFS 2.2.0 finally released last week with support for block cloning/reflinks, so this would now be super helpful for all us ZFS users 😄 

I can't wait to be able to hard modify subs without affecting the source cloned file!

calebj:
Just updating this with my latest findings.

[...]

PR #7566, opened 2025-01-10
Add reflink support for ZFS

### LINKED ISSUE #6183: Support reflinks on zfs 2.2.0+

Of course hard links are the best links. But they do have one vulnerability, if you modify any "copy" it modifies *all* copies. This will cause a torrent to fail its hash check, for example.

> Block cloning (https://github.com/openzfs/zfs/pull/13392) - Block cloning is a facility that allows a file (or parts of a file) to be "cloned", that is, a shallow copy made where the existing data blocks are referenced rather than copied. Later modifications to the data will cause a copy of the data block to be taken and that copy modified. This facility is used to implement "reflinks" or "file-level copy-on-write". Many common file copying programs, including newer versions of /bin/cp on Linux, will try to create clones automatically.

- [zfs-2.2.0](https://github.com/openzfs/zfs/releases/tag/zfs-2.2.0)

We already support reflinks on btrfs, lets add zfs?

This is what seems to do it, mainly limited to linux.
https://github.com/Sonarr/Sonarr/blob/develop/src/NzbDrone.Mono/Disk/RefLinkCreator.cs#L15

This is what limits it to btrfs.
https://github.com/Sonarr/Sonarr/blob/develop/src/NzbDrone.Common/Disk/DiskTransferService.cs#L347

The tricky part would be limiting it to zfs will have a lot of failures since 2.2.0 is new. But maybe we could spew something like "need to use 2.2.0+ for reflinks" or maybe reflinks could be a toggle paired w/ hard links?

I could use btrfs, but I'd rather not lose my data! ;)

### Anything else?

_No response_

jcassette:
#### Description
This allows for fast copy on ZFS volumes using the ZFS block cloning feature, aka reflink.

* Closes #6183

### DIFF
--- a/src/NzbDrone.Common/Disk/DiskTransferService.cs
+++ b/src/NzbDrone.Common/Disk/DiskTransferService.cs
@@ -341,10 +341,11 @@ namespace NzbDrone.Common.Disk
 
             var isCifs = targetDriveFormat == "cifs";
             var isBtrfs = sourceDriveFormat == "btrfs" && targetDriveFormat == "btrfs";
+            var isZfs = sourceDriveFormat == "zfs" && targetDriveFormat == "zfs";
 
             if (mode.HasFlag(TransferMode.Copy))
             {
-                if (isBtrfs)
+                if (isBtrfs || isZfs)
                 {
                     if (_diskProvider.TryCreateRefLink(sourcePath, targetPath))
                     {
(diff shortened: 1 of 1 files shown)


markus101:
With .net 7+ is this still necessary? Sounds like from https://github.com/Sonarr/Sonarr/issues/5470#issuecomment-2099536884 we should get it without extra work arounds.

PR #7572, opened 2025-01-12
Fix translation key for RSS in History Details

mynameisbogdan:
#### Description
Just a minor typo in translation token.


stevietv:
Good catch! I missed it as the key and translation were equal!

PR #7591, opened 2025-01-23
Fix Exclude Folder Regex for Torrent BlackHole

YvesPa:
#### Description

I'm using Sonarr with a Synology Nas and I use Torrent BlackHole for some case.
In tab Activity > Queue, I always have one element found and it's the @eaDir folder.

I did see that the current Regex is not taking the folder so I update it.

None

#### Database Migration
None

None

### DIFF
--- a/src/NzbDrone.Core/MediaFiles/DiskScanService.cs
+++ b/src/NzbDrone.Core/MediaFiles/DiskScanService.cs
@@ -69,8 +69,8 @@ namespace NzbDrone.Core.MediaFiles
             _logger = logger;
         }
 
-        private static readonly Regex ExcludedExtrasSubFolderRegex = new Regex(@"(?:\\|\/|^)(?:extras|extrafanart|behind the scenes|deleted scenes|featurettes|interviews|other|scenes|samples|shorts|trailers)(?:\\|\/)", RegexOptions.Compiled | RegexOptions.IgnoreCase);
-        private static readonly Regex ExcludedSubFoldersRegex = new Regex(@"(?:\\|\/|^)(?:@eadir|\.@__thumb|plex versions|\.[^\\/]+)(?:\\|\/)", RegexOptions.Compiled | RegexOptions.IgnoreCase);
+        private static readonly Regex ExcludedExtrasSubFolderRegex = new Regex(@"(?:\\|\/|^)(?:extras|extrafanart|behind the scenes|deleted scenes|featurettes|interviews|other|scenes|samples|shorts|trailers)(?:\\|\/|$)", RegexOptions.Compiled | RegexOptions.IgnoreCase);
+        private static readonly Regex ExcludedSubFoldersRegex = new Regex(@"(?:\\|\/|^)(?:@eaDir|\.@__thumb|plex versions|\.[^\\/]+)(?:\\|\/|$)", RegexOptions.Compiled | RegexOptions.IgnoreCase);
         private static readonly Regex ExcludedExtraFilesRegex = new Regex(@"(-(trailer|other|behindthescenes|deleted|featurette|interview|scene|short)\.[^.]+$)", RegexOptions.Compiled | RegexOptions.IgnoreCase);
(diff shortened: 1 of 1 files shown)


src/NzbDrone.Core/MediaFiles/DiskScanService.cs
```diff
@@ -69,8 +69,8 @@ public DiskScanService(IDiskProvider diskProvider,
             _logger = logger;
         }
 
-        private static readonly Regex ExcludedExtrasSubFolderRegex = new Regex(@"(?:\\|\/|^)(?:extras|extrafanart|behind the scenes|deleted scenes|featurettes|interviews|other|scenes|samples|shorts|trailers)(?:\\|\/)", RegexOptions.Compiled | RegexOptions.IgnoreCase);
-        private static readonly Regex ExcludedSubFoldersRegex = new Regex(@"(?:\\|\/|^)(?:@eadir|\.@__thumb|plex versions|\.[^\\/]+)(?:\\|\/)", RegexOptions.Compiled | RegexOptions.IgnoreCase);
+        private static readonly Regex ExcludedExtrasSubFolderRegex = new Regex(@"(?:\\|\/|^)(?:extras|extrafanart|behind the scenes|deleted scenes|featurettes|interviews|other|scenes|samples|shorts|trailers)(?:\\|\/|$)", RegexOptions.Compiled | RegexOptions.IgnoreCase);
+        private static readonly Regex ExcludedSubFoldersRegex = new Regex(@"(?:\\|\/|^)(?:@eaDir|\.@__thumb|plex versions|\.[^\\/]+)(?:\\|\/|$)", RegexOptions.Compiled | RegexOptions.IgnoreCase);
```
bakerboy448, line 73: `RegexOptions.IgnoreCase);`
the case change does nothing here

bakerboy448 (commented):
so the scope is to add `$` / end of line ? since there is no reported trailing `/` presumably from the file system?

src/NzbDrone.Core/MediaFiles/DiskScanService.cs
YvesPa: Oh I didn't see I commit the case change It was a try

YvesPa:
> so the actual change is to add `$` / end of line ? since there is no reported trailing `/` presumably from the file system?

For the Torent BlackHole in my tests on my nas and on my windows it doesn't have any trailing '/'


(Sorry for the force pushes didn't correctly configure git account on this new project)

bakerboy448:
Defer to Markus - but may be prudent to ensure the blackhole test cases are updated for this?

https://github.com/search?q=repo%3ASonarr%2FSonarr+eadir+path%3A%2F%5Esrc%5C%2FNzbDrone%5C.Core%5C.Test%5C%2F%2F&type=code

doesn't look like it's a current test case

markus101:
I don't think changing this regex will make a difference because it's used to filter a list of files where either `/` or `\` will be present, but a test that validates the behaviour will confirm.

YvesPa:
> I don't think changing this regex will make a difference because it's used to filter a list of files where either `/` or `\` will be present, but a test that validates the behaviour will confirm.
> 
> Looks like this does break an existing test though.

Sorry I did launch the test but didn't see it failed last time, but now I see it.

Currently when the torrentBlackhole use the FilterPath command the subforlder don't have any trailing slash, so the filter is not working.
In my example the paths list is P:\Sonar\@aeDir, with the current regex it except a trailing slash.
As you can see the @aeDir folder is returned, you can easealy reproduce.
image

For the test that is failed, in fact with the current filter is filtering all subdirectory starting with a dot.


In this use case, I'm not sure If we need to

markus101:
I was attempting to answer your question and ended up concluding neither option made sense and we are kind of abusing a file filter for a folder and ended up solving it another way.

I've opened up https://github.com/Sonarr/Sonarr/pull/7612 to resolve. Thanks for bringing this to our attention and your initial work on fixing it.

YvesPa:
Ok thank you, don't forget the same apply for Radarr @markus101

PR #7599, opened 2025-01-24
Fixed: Format bitrate for primary streams in media info

### LINKED ISSUE #7585: media info - add video bitrate / show bitrate in human readable suffix

Currently sonarr shows the following metadata:

Image

1.  While there is Audio Bitrate, Video Bitrate is not included. Please add.

This of course comes in very handy at times to know if a release has lower than expected bitrate/a fixed bitrate or something that might be an indication of quality of lack of.

2. Audio Bitrate is shown in Bytes. Please use scale suffix to make things human readable. 
Every other tool instead shows a much more human readable, for example 768 KB/s not 768000. 
Cant recall the term for this, basically you dont want 4+ leading digits, so you move to next magnitude to keep things readable.. 

example:
Image

3. ~~Possibly look at restructuring the layout to show the data in a more condense format..~~  moved to https://github.com/Sonarr/Sonarr/issues/7607

provided examples above inline

none

### Anything else?

_No response_

mynameisbogdan:
#### Description
Fixing the value for video bitrate by using the tag `BPS`, and using filesize in base 10 to properly format both bitrates in a human readable output.

* Closes #7585

### DIFF
--- a/frontend/src/Episode/Summary/MediaInfo.tsx
+++ b/frontend/src/Episode/Summary/MediaInfo.tsx
@@ -2,6 +2,7 @@ import React from 'react';
 import DescriptionList from 'Components/DescriptionList/DescriptionList';
 import DescriptionListItem from 'Components/DescriptionList/DescriptionListItem';
 import MediaInfoProps from 'typings/MediaInfo';
+import formatBitrate from 'Utilities/Number/formatBitrate';
 import getEntries from 'Utilities/Object/getEntries';
 
 function MediaInfo(props: MediaInfoProps) {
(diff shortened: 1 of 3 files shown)


frontend/src/Utilities/Number/formatBitrate.ts
```diff
@@ -0,0 +1,16 @@
...
+  const size = Number(input);
+
+  if (isNaN(size)) {
+    return '';
+  }
+
+  return `${filesize(size, {
+    base: 10,
+    round: 1,
+  })}`;
```
markus101, line 13: I think we should add ` /s` at the end as well

markus101:
That's fine, something for v5 I think anyways.

markus101:
Outside of bigger errors everything will be in v5 (will be changing the default branch soon™)

frontend/src/Episode/Summary/MediaInfo.tsx
```diff
@@ -16,9 +17,11 @@ function MediaInfo(props: MediaInfoProps) {
           return null;
         }
 
-        return (
-          <DescriptionListItem key={key} title={title} data={props[key]} />
-        );
+        if (key === 'audioBitrate' || key === 'videoBitrate') {
+          value = formatBitrate(value);
+        }
```
markus101, line 22: I think this should avoid the assignment to `value` even though it's not treated as a `const`.

```suggestion
        if (key === 'audioBitrate' || key === 'videoBitrate') {
          return (
            <DescriptionListItem
              key={key}
              title={title}
              data={formatBitrate(value)}
            />
          );
        }
```

PR #7611, opened 2025-01-30
Fix builds with v5 API

markus101:
#### Description

v5 API wasn't included in the release build 🙈


markus101:
Whoops, too much copy pasta on that one, I've fixed the `'default value'` and passing in the appropriate value for CI as well.

PR #7656, opened 2025-02-18
Improvement: Pushcut notifications

denisgabriel5:
#### Description
* Add series' poster image in the notification
* Add button to the series' IMDb page in the notification


src/NzbDrone.Core/Notifications/Pushcut/PushcutProxy.cs
```diff
@@ -29,7 +30,7 @@ public PushcutProxy(IHttpClient httpClient, ILocalizationService localizationSer
             _logger = logger;
         }
 
-        public void SendNotification(string title, string message, PushcutSettings settings)
+        public void SendNotification(string title, string message, string imageLink, PushcutSettings settings, string imdbId)
```
markus101, line 33: Keep settings last

src/NzbDrone.Core/Notifications/Pushcut/Pushcut.cs
```diff
@@ -28,52 +30,52 @@ public override ValidationResult Test()
 
         public override void OnGrab(GrabMessage grabMessage)
         {
-            _proxy.SendNotification(EPISODE_GRABBED_TITLE, grabMessage?.Message, Settings);
+            _proxy.SendNotification(EPISODE_GRABBED_TITLE, grabMessage?.Message, grabMessage.Series.Images.FirstOrDefault(x => x.CoverType == MediaCoverTypes.Poster)?.RemoteUrl, Settings, grabMessage.Series.ImdbId);
```
markus101, line 33: Instead of duplicating this lookup multiple times, extract to a function.

src/NzbDrone.Core/Notifications/Pushcut/PushcutProxy.cs
```diff
@@ -12,12 +12,13 @@ namespace NzbDrone.Core.Notifications.Pushcut
...
     public interface IPushcutProxy
     {
-        void SendNotification(string title, string message, PushcutSettings settings);
+        void SendNotification(string title, string message, string imageLink, PushcutSettings settings, string imdbId);
         ValidationFailure Test(PushcutSettings settings);
     }
 
     public class PushcutProxy : IPushcutProxy
     {
+        private const string _imdbLink = "https://www.imdb.com/title/";
```
markus101, line 21: We also have this in Gotify and Telegram, it'd be good to de-dupe these.

markus101 (changes requested):
What if I don't want an image sent or an IMDB button?
What if I want a button to TheTVDB instead?

Please add options for these

denisgabriel5:
> What if I don't want an image sent or an IMDB button? What if I want a button to TheTVDB instead?
> 
> Please add options for these

I don't feel like there are that many Pushcut users in order to add all this complexity. I am thinking that if anyone will ever need these, then they will open an issue.

markus101:
They shouldn't have to open an issue for what would effectively be a breaking change to their notifications.

If the numbers for Pushcat users is that low should we even bother supporting it at all? It's a non-zero amount of work to maintain it when additional notifications are added for minimal gains.

denisgabriel5:
> They shouldn't have to open an issue for what would effectively be a breaking change to their notifications.
> 
> 
> 
> If the numbers for Pushcat users is that low should we even bother supporting it at all? It's a non-zero amount of work to maintain it when additional notifications are added for minimal gains.

I use Pushcut because I like it. I added support for Pushcut notifications by myself in the first place. 

There are no breaking changes. Right now the notification shows just the event type and episode/show name. These changes just add an icon in the notification and the option to open IMDb if desired.

Sometimes a new show is grabbed and I forgot what was it about. So since I'm lazy and I don't want to google search it, I want with 2 taps to get the information I want.

markus101 (approved):
I ended up adding the options suggested here.

PR #7657, opened 2025-02-19
add .arj & .lzh to dangerousExtensions

v3DJG6GL:
#### Description
two file archive file extensions that are frequently used for spreading malware


v3DJG6GL:
I thought since `zipx` is also in this list, this might be the right place?

I've never seen these archives as legitimate archive files, especially when it comes to TV shows. They are mainly used to spread malware, as unpacking these files - I assume - is usually not supported by antivirus software, making it easier to bypass scanners.

Both archive files were apparently last widespread in the 90s.
I have never seen these extensions before sonarr grabbed fake/malware episodes from some trackers a few weeks/months (?) ago.

markus101:
I agree, I think the biggest consideration here is whether it's an archive that we expect should be unarchived, rar, zip or something pretending to be an achieve file.

PR #7673, opened 2025-02-24
Recommend against using uTorrent

mynameisbogdan:
#### Description
To be inline with our wiki.

image


src/NzbDrone.Core/Localization/Core/en.json
```diff
@@ -545,6 +545,7 @@
   "DownloadClientStatusSingleClientHealthCheckMessage": "Download clients unavailable due to failures: {downloadClientNames}",
   "DownloadClientTransmissionSettingsDirectoryHelpText": "Optional location to put downloads in, leave blank to use the default Transmission location",
   "DownloadClientTransmissionSettingsUrlBaseHelpText": "Adds a prefix to the {clientName} rpc url, eg {url}, defaults to '{defaultUrl}'",
+  "DownloadClientUTorrentProviderMessage": "Because uTorrent is known for cryptoware, malware and ads we'd suggest switching to a better client like qBittorrent, Deluge or ruTorrent.",
```
markus101, line 548: I think we should remain neutral on which client and something like:

```suggestion
  "DownloadClientUTorrentProviderMessage": "uTorrent has a history of including cryptominers, malware and ads, we strongly encourage you to choose a different client.",
```

PR #7684, opened 2025-02-27
enhance failed download warning for items not grabbed by sonarr

stevietv:
#### Description
enhance failed download warning for items not grabbed by sonarr

### DIFF
--- a/src/NzbDrone.Core/Download/FailedDownloadService.cs
+++ b/src/NzbDrone.Core/Download/FailedDownloadService.cs
@@ -83,7 +83,7 @@ namespace NzbDrone.Core.Download
 
                 if (grabbedItems.Empty())
                 {
-                    trackedDownload.Warn("Download wasn't grabbed by sonarr, skipping");
+                    trackedDownload.Warn("Download wasn't grabbed by Sonarr, skipping automatic download handling");
                     return;
                 }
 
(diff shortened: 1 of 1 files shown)


src/NzbDrone.Core/Download/FailedDownloadService.cs
```diff
@@ -83,7 +83,7 @@ public void Check(TrackedDownload trackedDownload)
 
                 if (grabbedItems.Empty())
                 {
-                    trackedDownload.Warn("Download wasn't grabbed by sonarr, skipping");
+                    trackedDownload.Warn("Download wasn't grabbed by Sonarr, skipping automatic download handling");
```
markus101, line 86: should we add in `failed` as well, so someone seeing this doesn't assume any download not grabbed by Sonarr will be skipped since it's just the failed ones?

stevietv: Its not only failed downloads at this warning, also encrypted downloads hit this warning if not grabbed by Sonarr. Can split it out if you want?

markus101: We do end up treating them both as failed, but handling the messaging differently seems reasonable. To avoid duplicating the whole block, rechecking `trackedDownload.DownloadItem.IsEncrypted` here would probably be easiest.

stevietv: done

PR #7713, opened 2025-03-11
Improve logging when login fails due to CryptographicException

markus101:
#### Description

Adds some logging and a suggestion if the XML is likely corrupt.


stevietv (approved):
assuming the using is fixed, looks like a nice addition

PR #7715, opened 2025-03-11
New: Selecting multiple indexers per release profile

### LINKED ISSUE #7706: Apply Release Profiles to multiple Indexers

I only just realized I can configure indexers to only be used for specific types of content without using tags - i.e. anime / tv can be 'filtered(?)' by emptying out the indexer's categories/anime categories dropdown lists.
Probably because I haven't changed my configuration since I set the service, my current setup involves tagging every indexer, release profile and show as 'anime' or 'tv'.

The problem I am facing is that I have one release profile for anime and one release profile for tv. 
If I were to remove my tagging, the release profiles would either apply to all indexers or only one indexer I specify.
It does not seem like I can set a release profile to multiple indexers at the moment.

I'd like to be able to apply a release profile to multiple indexers.
Perhaps in the same way that indexer categories work - with a dropdown list?
The aesthetics of how to go about doing this are not really of concern.

Alternatively, giving users the option to clone release profiles like custom formats can be cloned, would also make this process more efficient - so I can just duplicate release profiles and apply them to each indexer specifically. 

I thought about combining the two release profiles and while I think it would work, I think it depends on the release profile's parameters. Since tv and anime do not always share the same language during searches (English / Romanji), it's possible that short release profile restrictions could affect search results.

The other alternative I considered was to deploy another instance of sonarr specifically for the purpose of managing one type of content.
I'm not too sure about how much effort this approach might be in the short/long run.
I haven't really considered the up/down stream changes.

### Anything else?

_No response_

mynameisbogdan:
#### Description
Adding support for selecting multiple indexers per release profile.

#### Database Migration
YES - 224

* Closes #7706

### DIFF
--- a/src/NzbDrone.Core/Datastore/Migration/224_release_profile_indexer_ids.cs
+++ b/src/NzbDrone.Core/Datastore/Migration/224_release_profile_indexer_ids.cs
@@ -1,0 +1,59 @@
+using System.Collections.Generic;
+using System.Data;
+using Dapper;
+using FluentMigrator;
+using NzbDrone.Common.Serializer;
+using NzbDrone.Core.Datastore.Migration.Framework;
+
+namespace NzbDrone.Core.Datastore.Migration
+{
+    [Migration(224)]
+    public class release_profile_indexer_ids : NzbDroneMigrationBase
+    {
+        protected override void MainDbUpgrade()
+        {
+            Alter.Table("ReleaseProfiles").AddColumn("IndexerIds").AsString().WithDefaultValue("[]");
+
+            Execute.WithConnection(MigrateIndexerIds);
+
+            Delete.Column("IndexerId").FromTable("ReleaseProfiles");
+        }
+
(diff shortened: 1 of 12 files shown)


mynameisbogdan:
Since we have a v5 endpoint now, I'm only showing the first indexer Id on v3 API.

src/NzbDrone.Core/Datastore/Migration/224_release_profile_indexer_ids.cs
```diff
@@ -0,0 +1,59 @@
...
+using System.Data;
+using Dapper;
+using FluentMigrator;
+using NzbDrone.Common.Serializer;
+using NzbDrone.Core.Datastore.Migration.Framework;
+
+namespace NzbDrone.Core.Datastore.Migration
+{
+    [Migration(224)]
+    public class release_profile_indexer_ids : NzbDroneMigrationBase
```
markus101, line 11: Please add some tests to validate this migration.

PR #7718, opened 2025-03-12
Add history-based custom format scoring to decision logic

### LINKED ISSUE #5298: Cache Custom Format Calculations

Currently the backend needs to calculate Custom Formats for many different cases whenever the UI needs to display them, which means the same things are calculated over and over.

Cache Custom Format calculations and scores in memory.

When Custom Formats change the stored calculations need to be recalculated or wiped out and when quality profiles change the scores need to be redone as well.

It'd be easiest to wipe out all cached items whenever CFs change and let them be recalculated when requested again and same for scores when quality profiles change.

Not caching and returning the updated values on every request.

### Anything else?

Follow up to https://github.com/Sonarr/Sonarr/pull/5292

ImRidingNow:
Integrates history scores into download decision-making by comparing the custom format score of existing history entries with that of incoming releases. Releases are rejected if a higher score already exists in the history, ensuring better quality selections.

Database Migration
NO

Description
The goal is to check the history for the latest entry to determine if a better version already exists.

Let's discuss any necessary changes or potential concerns regarding this pull request here.

#### Attention
I have never touched the Sonarr code before, so please carefully review this for any disadvantages or issues I may have overlooked.

Technically, this method works. I have tested it locally, and I am currently using this version.

Maybe this should be behind a checkbox?

This is parallel to https://github.com/Radarr/Radarr/pull/10924

### DIFF
--- a/src/NzbDrone.Core.Test/HistoryTests/HistorySpecificationFixture.cs
+++ b/src/NzbDrone.Core.Test/HistoryTests/HistorySpecificationFixture.cs
@@ -62,7 +62,6 @@ namespace NzbDrone.Core.Test.HistoryTests
 
         private void SetupHistoryServiceMock(List<EpisodeHistory> history)
         {
-            // Setup mock for IHistoryService
             Mocker.GetMock<IHistoryService>()
                 .Setup(s => s.FindByEpisodeId(It.IsAny<int>()))
                 .Returns(history);
@@ -70,7 +69,6 @@ namespace NzbDrone.Core.Test.HistoryTests
 
         private void SetupCdh(bool cdhEnabled)
         {
-            // Setup mock for IHistoryService
             Mocker.GetMock<IConfigService>()
                 .Setup(s => s.EnableCompletedDownloadHandling)
                 .Returns(cdhEnabled);
@@ -191,7 +189,6 @@ namespace NzbDrone.Core.Test.HistoryTests
             // Arrange
             var betterQuality = new QualityModel(Quality.Bluray1080p);
             var history = GivenFileHistory(DateTime.UtcNow.AddHours(-1), betterQuality, EpisodeHistoryEventType.Grabbed);
-
             var newQuality = new QualityModel(Quality.Bluray1080p);
             var remoteEpisode = MockUpOrDownGrade(3000, 8000, newQuality, Quality.Bluray1080p.Id);
             SetupHistoryServiceMock(history);
--- a/src/NzbDrone.Core.Test/HistoryTests/HistorySpecificationFixture.cs
+++ b/src/NzbDrone.Core.Test/HistoryTests/HistorySpecificationFixture.cs
@@ -133,7 +133,6 @@ namespace NzbDrone.Core.Test.HistoryTests
         [Test]
         public void should_reject_when_grabbed_history_is_old_and_cdh_enabled_when_no_quality_update()
         {
-            // Arrange
             var newQuality = new QualityModel(Quality.HDTV720p);
             var remoteEpisode = MockUpOrDownGrade(3000, 3000, newQuality, Quality.HDTV720p.Id, true);
             var history = GivenFileHistory(DateTime.UtcNow.AddHours(-13), _quality, EpisodeHistoryEventType.Grabbed);
@@ -141,34 +140,28 @@ namespace NzbDrone.Core.Test.HistoryTests
             SetupHistoryServiceMock(history);
             SetupCdh(true);
 
-            // Act
             var decision = Subject.IsSatisfiedBy(remoteEpisode, null);
 
-            // Assert
             decision.Accepted.Should().BeFalse();
         }
 
         [Test]
         public void should_accept_when_grabbed_history_is_old_and_cdh_enabled_when_quality_update()
         {
-            // Arrange
             var newQuality = new QualityModel(Quality.Bluray1080p);
             var remoteEpisode = MockUpOrDownGrade(3000, 3000, newQuality, Quality.WEBRip2160p.Id);
             var history = GivenFileHistory(DateTime.UtcNow.AddHours(-13), _quality, EpisodeHistoryEventType.Grabbed);
             SetupHistoryServiceMock(history);
             SetupCdh(true);
 
-            // Act
             var decision = Subject.IsSatisfiedBy(remoteEpisode, null);
 
-            // Assert
             decision.Accepted.Should().BeTrue();
         }
 
         [Test]
         public void should_accept_when_grabbed_history_is_old_and_cdh_enabled_when_custom_format_score_update()
         {
-            // Arrange
             var newQuality = new QualityModel(Quality.Bluray1080p);
             var remoteEpisode = MockUpOrDownGrade(3000, 8000, newQuality, Quality.WEBRip1080p.Id);
             var history = GivenFileHistory(DateTime.UtcNow.AddHours(-13), _quality, EpisodeHistoryEventType.Grabbed);
@@ -176,27 +169,22 @@ namespace NzbDrone.Core.Test.HistoryTests
             SetupHistoryServiceMock(history);
             SetupCdh(true);
 
-            // Act
             var decision = Subject.IsSatisfiedBy(remoteEpisode, null);
 
-            // Assert
             decision.Accepted.Should().BeTrue();
         }
 
         [Test]
         public void should_reject_when_grabbed_history_meets_cutoff_and_is_recent()
         {
-            // Arrange
             var betterQuality = new QualityModel(Quality.Bluray1080p);
             var history = GivenFileHistory(DateTime.UtcNow.AddHours(-1), betterQuality, EpisodeHistoryEventType.Grabbed);
             var newQuality = new QualityModel(Quality.Bluray1080p);
             var remoteEpisode = MockUpOrDownGrade(3000, 8000, newQuality, Quality.Bluray1080p.Id);
             SetupHistoryServiceMock(history);
 
-            // Act
             var decision = Subject.IsSatisfiedBy(remoteEpisode, null);
 
-            // Assert
             decision.Accepted.Should().BeFalse();
             decision.Reason.Should().Be(DownloadRejectionReason.HistoryRecentCutoffMet);
         }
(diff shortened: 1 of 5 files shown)


bakerboy448:
>  Releases are rejected if a higher score already exists in the history

all of history? If so - that seems problematic and will break workflow for anyone who deletes media and or is otherwise trying to upgrade? 

Based on the code it appears to only look at **Series Import** (I assume this is an internal hidden history equivalent to Library Import or post-add scanning) and **Grabs** - however it appears it only applies if the most recent history is a grab?

or what's the layman's speak?

devinSpitz:
> > Releases are rejected if a higher score already exists in the history
> 
> ~all of history? If so - that seems problematic and will break workflow for anyone who deletes media and or is otherwise trying to upgrade?~
> 
> ~Based on the code it appears to only look at **Series Import** (I assume this is an internal hidden history equivalent to Library Import or post-add scanning) and **Grabs** - however it appears it only applies if the most recent history is a grab?~
> 
> ~or what's the layman's speak?~
> 
> edit; answered by your edit; applies to most recent grab vs. current grab.
> 
> need to ensure this does not erroneously block quality upgrade grabs nor impact season pack grabs

'm trying to find a way to validate that, to be honest. However, I think it shouldn't be an issue, as it simply doesn't reject the download when no custom format score is present. Like what the other code does :D

bakerboy448:
custom format scores will almost always be present for prior grabs...so you're saying it'd only allow a quality upgrade if there was no custom format score now or if and only if the new CF score is better even tho quality is?

that wouldn't work and would be a major breaking change for basic functionality.

devinSpitz:
> custom format scores will almost always be present for prior grabs...so you're saying it'd only allow a quality upgrade if there was no custom format score now or if and only if the new CF score is better even tho quality is?
> 
> that wouldn't work and would be a major breaking change for basic functionality.

No. It just creates another path to reject a download.

If the old logic applies, nothing has changed.

If there is an EpisodeHistory that is a Grab or SeriesFolderImported, it takes the most recent one and compares the custom format score.

If there is no custom format score, it just continues like before. So no code change at all if these conditions are not met.

But if they are met and the new custom score would be lower, it would reject the download.

bakerboy448:
Added some edits to my other message - it appears existing logic gets the current custom formats? https://github.com/Sonarr/Sonarr/blob/c8cb74a976b937cf2b737d2c5f1b8deecaddbf12/src/NzbDrone.Core/DecisionEngine/Specifications/UpgradableSpecification.cs#L83

this would change it to compare to just the grab.... and the grab may or may not reflect what is on disk? Thus this may cause loops or otherwise prevent upgrades given material differences in formats exist between release title and on-disk after probing. this is more of a concern for radarr
https://github.com/Sonarr/Sonarr/issues/5598

devinSpitz:
> Added some edits to my other message - it appears existing logic gets the current custom formats?
> 
> https://github.com/Sonarr/Sonarr/blob/c8cb74a976b937cf2b737d2c5f1b8deecaddbf12/src/NzbDrone.Core/DecisionEngine/Specifications/UpgradableSpecification.cs#L83
> 
> this would change it to compare to just the grab.... and the grab may or may not reflect what is on disk? Thus this may cause loops or otherwise prevent upgrades given material differences in formats exist between release title and on-disk after probing. this is more of a concern for radarr #5598

No it would only get it if the **most recent history entry is a grab** otherwise it will **not**. Mine will when there **is a file and latest grab or SeriesFolderImported exists**

which is not the same :D

devinSpitz:
In simpler words it does add another chance to not lower the custom score of your library.

src/NzbDrone.Core/DecisionEngine/Specifications/RssSync/HistorySpecification.cs
```diff
@@ -116,6 +124,41 @@ public virtual DownloadSpecDecision IsSatisfiedBy(RemoteEpisode subject, SearchC
                             return DownloadSpecDecision.Reject(DownloadRejectionReason.HistoryUpgradesNotAllowed, "{0} grab event in history and Quality Profile '{1}' does not allow upgrades", rejectionSubject, qualityProfile.Name);
                     }
                 }
+
+                if (episode is { HasFile: true })
+                {
+                    EpisodeHistory availableUsableEpisodeHistoryForCustomFormatScore;
+
+                    var episodeHistoryEventTypeForHistoryComparison = new List<EpisodeHistoryEventType>
```
markus101, line 132: Use a HashSet, that is defined outside of the function so it's not re-created every time.

src/NzbDrone.Core/DecisionEngine/Specifications/RssSync/HistorySpecification.cs
```diff
@@ -116,6 +124,41 @@ public virtual DownloadSpecDecision IsSatisfiedBy(RemoteEpisode subject, SearchC
...
+                    {
+                        EpisodeHistoryEventType.Grabbed,
+                        EpisodeHistoryEventType.SeriesFolderImported,
+                    };
+
+                    if (episodeHistoryEventTypeForHistoryComparison.Contains(mostRecent.EventType))
+                    {
+                        availableUsableEpisodeHistoryForCustomFormatScore = mostRecent;
```
markus101, line 140: You can use a ternary for this:

```
var availableUsableEpisodeHistoryForCustomFormatScore = episodeHistoryEventTypeForHistoryComparison.Contains(mostRecent.EventType) ? mostRecent : _historyService.MostRecentForEpisodeInEventCollection(
                            episode.Id,
                            new ReadOnlyCollection<EpisodeHistoryEventType>(episodeHistoryEventTypeForHistoryComparison));
```

I'd say that these variables are too verbose in naming though, they're monstrosities to read.

src/NzbDrone.Core/DecisionEngine/Specifications/RssSync/HistorySpecification.cs
```diff
@@ -116,6 +124,41 @@ public virtual DownloadSpecDecision IsSatisfiedBy(RemoteEpisode subject, SearchC
...
+                    {
+                        continue;
+                    }
+
+                    var mostRecentCustomFormat = _formatService.ParseCustomFormat(availableUsableEpisodeHistoryForCustomFormatScore, subject.Series);
+                    var mostRecentCustomFormatScore = qualityProfile.CalculateCustomFormatScore(mostRecentCustomFormat);
+
+                    if (mostRecentCustomFormatScore > subject.CustomFormatScore)
+                    {
+                        return DownloadSpecDecision.Reject(DownloadRejectionReason.HistoryCustomFormatScore, "Quality Profile '{0}' has a higher Custom Format score than the report: {1}", qualityProfile.Name, subject.CustomFormatScore);
```
markus101, line 159: None of our other messaging indicates the quality profile name and instead indicates which item is being compared, be it file or release in queue, we should maintain that here as well.

src/NzbDrone.Core/DecisionEngine/Specifications/RssSync/HistorySpecification.cs
```diff
@@ -51,7 +54,12 @@ public virtual DownloadSpecDecision IsSatisfiedBy(RemoteEpisode subject, SearchC
                 _logger.Debug("Checking current status of episode [{0}] in history", episode.Id);
                 var mostRecent = _historyService.MostRecentForEpisode(episode.Id);
 
-                if (mostRecent != null && mostRecent.EventType == EpisodeHistoryEventType.Grabbed)
+                if (mostRecent == null)
```
markus101, line 57: Instead of only getting the most recent item, why not get all the events for the episode? If the latest is grab then it can handle that like normal. Otherwise you can loop until you find an event type that you're looking for.

src/NzbDrone.Core/History/HistoryRepository.cs
```diff
@@ -132,6 +134,11 @@ public PagingSpec<EpisodeHistory> GetPaged(PagingSpec<EpisodeHistory> pagingSpec
             return pagingSpec;
         }
 
+        public EpisodeHistory MostRecentForEpisodeInEventCollection(int id, ReadOnlyCollection<EpisodeHistoryEventType> episodeHistoryEventTypes)
```
markus101, line 137: WIth my suggestions I don't think this is necessary.

markus101 (changes requested):

src/NzbDrone.Core/History/HistoryRepository.cs
devinSpitz: Thanks for the Input i will check every thing as soon as i have time :D

devinSpitz:
I also add tests. It will take some more days :D

devinSpitz:
I have now also added tests.

These are the results before my code changes on the v5-develop branch today:
image

I have hopefully added all your concerns to the tests. Otherwise, we can always add more to make sure we get what we want :D

devinSpitz:
The main concern this fixes is when you already have a big library and import it. Before the behavior was that your old library would always get overwritten with whatever it gets via RSS cause you had no EpisodeHistoryEventType.Grabbed in the past 12 hours. Now it will also check for EpisodeHistoryEventType.SeriesFolderImported and consider them.

devinSpitz:
The tests should be merged with src/NzbDrone.Core.Test/DecisionEngineTests/RssSync/HistorySpecificationFixture.cs

devinSpitz:
> > The tests should be merged with [src/NzbDrone.Core.Test/DecisionEngineTests/RssSync/HistorySpecificationFixture.cs](https://github.com/Sonarr/Sonarr/blob/v5-develop/src/NzbDrone.Core.Test/DecisionEngineTests/RssSync/HistorySpecificationFixture.cs)
> 
> Why don't you merge your src/NzbDrone.Core.Test/HistoryTests/HistorySpecificationFixture.cs changes with src/NzbDrone.Core.Test/DecisionEngineTests/RssSync/HistorySpecificationFixture.cs?

I just saw it an hour ago xD Will do it :)

src/NzbDrone.Core.Test/HistoryTests/HistorySpecificationFixture.cs
```diff
@@ -0,0 +1,273 @@
...
+                new()
+                {
+                    Date = date,
+                    Quality = quality,
+                    EventType = eventType
+                }
+            };
+        }
+
+        private RemoteEpisode MockUpOrDownGrade(int customFormatScoreHistory, int customFormatScoreNew, QualityModel newQuality, int cutoff, bool hasFile = false)
```
markus101, line 90: This is doing too much and should be split into  `GivenProfile` that does the profile + mock and then a separate function to create the `RemoteEpisode`

markus101 (changes requested):
What happens if you download a file, it imports then it's later removed for some reason (manually deleted, lost a disk, etc)?

markus101:

PR #7735, opened 2025-03-20
Add series progress tag to header

stevietv:
#### Description
Add an extra tag to the header showing total episode files / total episodes

image

image

### DIFF
--- a/frontend/src/Series/Details/SeriesDetails.css
+++ b/frontend/src/Series/Details/SeriesDetails.css
@@ -153,6 +153,13 @@
   padding: 20px;
 }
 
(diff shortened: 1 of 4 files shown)


stevietv:
> How about after the running years? Looks too much where you put it. 🫣

I did consider that, but thought the box fitted better here since its in a row of boxes already, whereas the row with years is all plain text.

markus101:
Not sure if it's better, but what about a progress bar under the poster?

stevietv:
> Not sure if it's better, but what about a progress bar under the poster?

I was gonna try this last night, but thinking about it felt like it looked odd

PR #7737, opened 2025-03-21
Improve Series Details loading

mynameisbogdan:
#### Description
Improve series details loading since `localeCompare` is quite taxing.


frontend/src/Series/Details/SeriesDetails.tsx
```diff
@@ -126,7 +126,10 @@ function createEpisodeFilesSelector() {
 
 function createSeriesSelector(seriesId: number) {
   return createSelector(createAllSeriesSelector(), (allSeries) => {
-    const sortedSeries = [...allSeries].sort(sortByProp('sortTitle'));
+    const sortedSeries = useMemo(
```
markus101, line 129: The selector should only be triggered when the input changes, same as how memo operates. Hooks should also not be used outside of other hooks or components.

PR #7771, opened 2025-03-26
Expose total space in root folder endpoint

### LINKED ISSUE #7769: Add total space for root folder endpoint

This was originally filled here #2390

But is seemingly gone now

Original or similar implementation to before

none exist

### Anything else?

https://github.com/Sonarr/Sonarr/issues/2390

Josh-Archer:
I thought about adding a check that the field was there in the integration tests, since the check is just that it's not 0. But seemed like overkill since we didn't do that elsewhere. 
#### Description
Expose the total space field in the `RootFolder` endpoint. Also, update the ChromeDriver package to support latest. 

* Closes #7769

### DIFF
--- a/src/NzbDrone.Automation.Test/Sonarr.Automation.Test.csproj
+++ b/src/NzbDrone.Automation.Test/Sonarr.Automation.Test.csproj
@@ -5,7 +5,7 @@
   <ItemGroup>
     <PackageReference Include="GitHubActionsTestLogger" Version="2.4.1" />
     <PackageReference Include="Selenium.Support" Version="3.141.0" />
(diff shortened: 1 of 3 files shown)


Josh-Archer:
Sounds like from the Issue that it was accidentally removed. I kept to the scope of the ticket, but can add the FE piece of it as well.

Josh-Archer:
Yeah, I noticed that while I was going through. Seems redundant to have in 2 apis. Ill close and comment back on the issue.

markus101:
I considered that the disk space endpoints include it, but there isn't necessarily a 1:1 mapping between the storage stored there and root folders so I can see why someone would want it from the root folder endpoint. We also still get the total space for root folders, we just stopped returning it via the API.

PR #7822, opened 2025-04-17
add validation to CF spec for max size

stevietv:
#### Description
add validation to CF spec for max size. Avoid any overflow exceptions.


src/NzbDrone.Core/CustomFormats/Specifications/SizeSpecification.cs
```diff
@@ -11,6 +11,7 @@ public SizeSpecificationValidator()
         {
             RuleFor(c => c.Min).GreaterThanOrEqualTo(0);
             RuleFor(c => c.Max).GreaterThan(c => c.Min);
+            RuleFor(c => c.Max).LessThanOrEqualTo(400);
```
markus101, line 14: Actually I was thinking this was size per minute limits, not for CF, I think `double.MaxValue` is best after that realization.

```suggestion
            RuleFor(c => c.Max).LessThanOrEqualTo(double.MaxValue);
```

markus101 (approved):

PR #7847, opened 2025-05-07
feat: paginate interactive search results

darmiel:
#### Description
When interactive search returns _many_ results (for example for Season Search with the _Season Pack_ filter disabled), the browser tab may crash (or the UI slows down a lot). This Pull Request adds client-side "virtual" pagination: the full result set is still fetched in one request, but only a subset is rendered in the DOM.

This Pull Request can be seen more as a PoC and to get some feedback of this change and if we want to go forward paginating the interactive search 🙂.

image

#### Database Migration
NO

n/a

### DIFF
--- a/frontend/src/InteractiveSearch/InteractiveSearch.tsx
+++ b/frontend/src/InteractiveSearch/InteractiveSearch.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useEffect } from 'react';
+import React, { useCallback, useEffect, useMemo, useState } from 'react';
 import { useDispatch, useSelector } from 'react-redux';
 import ClientSideCollectionAppState from 'App/State/ClientSideCollectionAppState';
 import ReleasesAppState from 'App/State/ReleasesAppState';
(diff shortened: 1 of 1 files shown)


markus101:
An issue is a good place for a discussion, but this isn’t something we want to do. Virtualization would be a much better approach.

darmiel:
You're right - virtual scrolling could be a better solution. I just wanted to experiment and fix the slowdown I had on my instance and this was my first shot. I opened the PR mainly to get your thoughts  (we can also turn it into an issue if that's easier).

One more idea for the list would be a quick-search box that filters the titles on the fly without the need to create a filter. This was actually in my prototype _before_ opening this PR and looked like this:

https://github.com/user-attachments/assets/bf8f22c1-8889-4816-9cb0-cb8cdfa960b6

This also eliminates the "browser-based search problem" we have with virtual scrolling and a paginated table.

markus101:
I think if the UX can be improved something like that could be useful in addition to the virtualization.

Moving this to draft while this is iterated on.

markus101:

Issue #7625, opened 2025-02-03
Add support for subtitle attribute and custom format condition


PCJones:
### Is there an existing issue for this?

- [x] I have searched the existing open and closed issues

### Is your feature request related to a problem? Please describe

There is currently no option to automatically filter for releases with specific subtitle languages or to give them priority.

### Describe the solution you'd like

- Add support for the newznab/torznab "subs" [attribute](https://inhies.github.io/Newznab-API/attributes/). There is already existing support for the "language" attribute (and possibly others?).
- Add a custom format to filter for subtitles for specific languages, similar to audio. In addition to the newznab/torznab attribute it could also be triggered by certain strings like "Ger.Sub", "Eng.Sub", "Eng-Subs", "Multi-Subs"/"MultiSub" (might have to guess here, but the original language is probably a given), just as the condition for the audio language is influenced by both the newznab attribute and certain keywords in the publication title.
- Just like for audio once the file is downloaded MediaInfo should be used to extract this information as the final source of truth

### Describe alternatives you've considered

The only alternative right now is to create custom format conditions for the release title. However, except for some anime releases, almost no release contains the available subtitles in it's title. It would be much better to use information provided by mediainfo and indexers/trackers trough their API.

markus101:
Given Bogdan's concerns and myself having the same reservations this isn't going to be something we commit to doing, but will tag for future consideration.

PCJones:
There are more and more indexers who support the "subs" attribute: https://www.reddit.com/r/usenet/comments/1iyr7ci/any_other_indexers_apart_from_geek_that_publish/

PR #7858, opened 2025-05-15
 Add Subtitle Information to Interactive Search

Fabzien:
#### Description

Hi team,

This is my first contribution to the project, and I’d appreciate your feedback.
Motivation:

I often need to find releases that include subtitles in multiple languages. Since I’m using a private tracker with private releases, I can’t rely on external tools like Bazarr to fetch subtitles.

However, my private tracker provides subtitle information directly through Jackett. To make use of this, I’ve added support for displaying subtitle availability in the interactive search results.

Notes:

- This is an initial implementation intended as a proposal.
- I’d love your thoughts on whether this approach aligns with the project’s goals.
- Could this feature be extended to other indexers or sources in the future?

Looking forward to your feedback and suggestions for improvement!

image

#### Releated issue
*  #7625


src/NzbDrone.Core/Indexers/Torznab/TorznabRssParser.cs
```diff
@@ -124,6 +124,24 @@ protected override List<Language> GetLanguages(XElement item)
             return results;
         }
 
+        protected override List<Language> GetSubtitles(XElement item)
```
markus101, line 127: The same should be added to `Newznab` parsing.

markus101 (changes requested):
This should also support newznab parsing if it's already supported or later added.

This new column should also be optional, perhaps hidden by default (but I don't feel too strongly about that), this will allow users that don't care about subtitle languages to hide it and potentially other columns as well.

Fabzien:
@markus101 Regarding Newznab, I’ll take a look to see if I can add it.

I agree with you this new column should be optional. How do you think it should be handled? As far as I know, in interactive search there’s no option to add or remove columns like in "Queue" or "History" UI. Should we add an option in the interactive search UI?

src/NzbDrone.Core/Indexers/Torznab/TorznabRssParser.cs
Fabzien: I did it.

Fabzien:
Meanwhile, I added subtitles to the filter options. A possible temporary workaround could be to exclude subtitles from the interactive search display, but still allow users to filter by them.

markus101:
Yeah a similar option should be added to the UI. Logs is probably the easiest example of that in the new UI work (no longer using redux for that).

I don't thinking hiding them, but allowing filtering is worth the confusion. Going to convert this to draft while effort towards optional columns is made.

markus101:

PR #7864, opened 2025-05-17
New: Include Special option for Wanted Missing

mynameisbogdan:
#### Description
New `Special Seasons` custom filter for wanted missing. 

Enabled by default in the UI and v3 API, but we should make it disabled by default in v5 API @ MissingController.

Screenshot 1
Screenshot 2

### DIFF
--- a/frontend/src/Wanted/Missing/Missing.tsx
+++ b/frontend/src/Wanted/Missing/Missing.tsx
@@ -20,6 +20,7 @@ import TablePager from 'Components/Table/TablePager';
 import Episode from 'Episode/Episode';
 import { useToggleEpisodesMonitored } from 'Episode/useEpisode';
 import { Filter } from 'Filters/Filter';
+import { useCustomFiltersList } from 'Filters/useCustomFilters';
 import { align, icons, kinds } from 'Helpers/Props';
 import { SortDirection } from 'Helpers/Props/sortDirections';
 import InteractiveImportModal from 'InteractiveImport/InteractiveImportModal';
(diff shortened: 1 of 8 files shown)


markus101:
Should just target this against v5, by the time it releases nothing should continue using the v3 API.

mynameisbogdan:
Rebased for the v5 API and react-query. 💪

frontend/src/Wanted/Missing/MissingOptions.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+function MissingOptions() {
+  const includeSpecials = useMissingOption('includeSpecials');
+  const { goToPage } = useMissing();
+
+  const handleOptionChange = useCallback(
+    ({ name, value }: OptionChanged<MissingOptionstype>) => {
+      setMissingOption(name, value);
+
+      console.debug({ name });
```
markus101: Whoops.

```suggestion
```

markus101 (changes requested):
I think this should be implemented via Custom Filters instead of an option, that'd fit more inline with what we're already doing for unmonitored episodes and be more flexible in the future.

mynameisbogdan:
Do you plan to do the same for the queue? Because I took inspiration from there, and it's the same as here.

markus101:
I hadn't considered that, but yeah, that makes sense. For the calendar I went straight to custom filter for excluding specials and didn't consider a separate option and forgot about this special case for the queue.

mynameisbogdan:
I did a rebase + refactor to use custom filters, but I have a few mentions to make:
- when you add a new custom filter with the new react-query, you get a `Matching filter not found` error in the console.
- when you remove a custom filter it defaults to [`selectedFilterKey: 'all'` ](https://github.com/Sonarr/Sonarr/blob/7d2e01d5161f9f6610d073574e46418987183e9c/frontend/src/Components/Filter/CustomFilters/CustomFilter.tsx#L52)
- monitored is true by default, so if you create a new filter with just `includeSpecials` it's still going to filter by monitored.

markus101 (approved):

PR #7873, opened 2025-05-22
Prevent should refresh series from failing

markus101:
#### Description
Prevent error checking if series should be refreshed from failing refresh series task, which came up on the fourms.


markus101:
Yeah the update may fail, but it won't fail the whole task like it's doing now.

markus101:
Yeah, the single series can still fail, currently the entire update fails when the affected series fails, so any series in the list after the affected series never gets updated.

PR #7881, opened 2025-05-25
added 'libicu72' as a required package

Cybertinus:
Sonarr refushed to start on Debian 12.11 without this package present. So, I added it to the required packages.

#### Description
When I ran install.sh on my pretty bare installed Debian 12.11 VM Sonarr failed to start afterwards with this error:
Process terminated. Couldn't find a valid ICU package installed on the system. Please install libicu using your package manag
er and try again. Alternatively you can set the configuration flag System.Globalization.Invariant to true if you want to run with no globalization support. Please see https://aka.ms/dotnet-missing-libicu for more information.

So, I installed the package `libicu72` and it started without issue. So to make life easier for other people, I added this to the prereq list in the install script.

### DIFF
--- a/distribution/debian/install.sh
+++ b/distribution/debian/install.sh
@@ -33,7 +33,7 @@ fi
 
 app="sonarr"
 app_port="8989"
-app_prereq="curl sqlite3 wget"
+app_prereq="curl sqlite3 wget libicu72"
 app_umask="0002"
 branch="main"
 


stevietv:
@markus101 

As i suspected, this script fails on an ubuntu server as the package is not available. I suggest that we revert the change and potentially add a note that on the webpage instead, or we need to build in a way for checking if the package exists for each package before installing.

image


we also need to consider that changes to the script on v5-develop branch aren't in production on the website as that links to develop

markus101:
Thanks for checking Stevie, I have reverted it in `v5-develop`. I agree, it'd be nice if we could properly detect if it's available / only install on Debian instead of all debian-flavoured systems that this script can run on.

My worry with only having it on the website is people will overlook it or we have a bunch of cases that people will need to see if they apply to them.

bachrc:
I just installed the latest version of Sonarr for debian 14 with the one-line command in the docs, and I had this message telling me I didn't have a libICU library.

In debian 14 repos, the package libicu72 isn't available.

I had to install the debian `libicu-dev` package and everything went smooth again.

PR #7907, opened 2025-06-07
New: allow user to define additional rejected extension types

### LINKED ISSUE #7721: User-editable list of dangerous extensions

Currently, it looks like "dangerous extensions" are defined statically in this file: https://github.com/Sonarr/Sonarr/blob/develop/src/NzbDrone.Core/MediaFiles/FileExtensions.cs#L24

However, whoever has been pushing malware releases with `.exe` and `.lnk`, has recently switched to using `.arj`.

Right now, in order to get that extension added to the list, a user needs to file an issue (like [this one](https://github.com/Sonarr/Sonarr/issues/7588)) and/or submit a PR (like [this one](https://github.com/Sonarr/Sonarr/pull/7596)) and wait for the change to make it's way through. Additionally, there may be some disagreement about which extensions belong on this list.

Create a setting in the UI where the end-user can populate additional extensions to be considered dangerous, in addition to the hardcoded list.

This would allow the end-user to be more reactive to new malware tactics as they come up, and it would allow them to include additional extensions beyond what maintainers might consider appropriate for the hardcoded list.

One _could_ maintain a fork with an extended list of dangerous extensions, but a user-editable setting for this would provide a better experience

### Anything else?

_No response_

stevietv:
#### Description
Adds an option to Media Management settings for user to define their own rejected extension types. 
If assigned to an indexer this will then fail downloads just as Executables or Potentially Dangerous would.

image

image

* Closes #7721


frontend/src/typings/Settings/MediaManagement.ts
```diff
@@ -18,5 +18,6 @@ export default interface MediaManagement {
   scriptImportPath: string;
   importExtraFiles: boolean;
   extraFileExtensions: string;
+  userRejectedExtensions: string;
```
mynameisbogdan, line 21: Why not a string[]? Input type tag supports arrays.

stevietv: > Why not a string[]? Input type tag supports arrays.

kept it the same as extra file extensions

mynameisbogdan:
Should we have validation to prevent adding valid extensions like mkv?

stevietv:
> Should we have validation to prevent adding valid extensions like mkv?

Was wondering that myself. I figured first leave it like this, if a user wants to block mkvs then its on them! And then its also a question of what to validate against, given that there are technically a lot of MediaFileExtensions defined

markus101:
Yeah, we should not allow media file or valid archive extensions to be rejected as they aren't dangerous and someone is just going to paste in a list and have issues at some point.

stevietv:
added some validation based on ArchiveExtensions (had to make this class public) and MediaExtensions

image

src/NzbDrone.Core/Localization/Core/en.json
```diff
@@ -2138,6 +2138,9 @@
   "UsenetDelayTime": "Usenet Delay: {usenetDelay}",
   "UsenetDisabled": "Usenet Disabled",
   "UserInvokedSearch": "User Invoked Search",
+  "UserRejectedExtensions": "Additional Rejected File Extensions",
+  "UserRejectedExtensionsHelpText": "Comma separated list of files extensions to fail (Fail Downloads also needs to be enabled per indexer)",
+  "UserRejectedExtensionsTextsExamples": "Examples: '.001, .dog' or '001,dog'",
```
markus101: Should we treat `001` is an archive file as that's typically the use of them?

markus101 (commented):
Just the question about `001` potentially being an archive file.

src/NzbDrone.Core/Localization/Core/en.json
stevietv: > Should we treat `001` is an archive file as that's typically the use of them?

I think it would make sense to include the 001 in the list of archive extensions, but nothing more than that. want separate PR for that?
I'll update the helptext here

src/NzbDrone.Core/MediaFiles/FileExtensions.cs
```diff
@@ -3,10 +3,11 @@
 
 namespace NzbDrone.Core.MediaFiles
 {
-    internal static class FileExtensions
+    public static class FileExtensions
     {
         private static List<string> _archiveExtensions = new List<string>
         {
+            ".001",
```
markus101: I think for the time being we shouldn't include it as an archive file (probably it's being used for junk/malicious files) based on some reports, but I wouldn't use it in the example.

stevietv: removed this commit. makes sense.

GLotsapot:
> Yeah, we should not allow media file or valid archive extensions to be rejected as they aren't dangerous and someone is just going to paste in a list and have issues at some point.

I agree with not allowing the to block media file extensions to help save themself from their own stupidity, but I wouldn't stop them from blocking valid compressed files.
Many of the fake torrents out there currently do actually use a valid compressed file extension (ZIPX, ARJ, 001, etc.), and even if they were potentially valid compressed media, the end user would have to have a 3rd party setup that extracts the media for Sonarr to actually be able to manage the file.

markus101:
> Many of the fake torrents out there currently do actually use a valid compressed file extension (ZIPX, ARJ, 001, etc.)

None of those are treated as compressed files by Sonarr.

> and even if they were potentially valid compressed media, the end user would have to have a 3rd party setup that extracts the media for Sonarr to actually be able to manage the file.

That's correct, but treating them as malicious because that's not set up isn't a reason that we're going to cover.

PR #7908, opened 2025-06-07
Fixed: Handling series with empty IMDB IDs in lists clean library

mpeleshenko:
#### Description
Fixed import list sync cleanup which was not cleaning up all series without an ImdbId if the import lists contained series without an ImdbId as well, despite no match on other criteria.


markus101:
Thanks for the fix!

PR #7923, opened 2025-06-16
fix: select all on series import

### LINKED ISSUE #7909: Select All in Library Import

### Current Behavior

The select all toggle button at the top of the Library Import table doesn't function, select boxes don't toggle on and off.

It seems to work when all shows are not in sonarr, but when one has been matched to an existing show it appears unable to toggle properly.

trying to toggle on and off when one show has been marked as existing:

https://github.com/user-attachments/assets/35b42b3c-6567-4b8a-ba78-893b0b4de087

when none of the shows are matched to existing:

https://github.com/user-attachments/assets/07f738b5-6344-4d23-ad1f-2077cdab4591

### Expected Behavior

toggle on and off all importable shows, skipping the Exisiting ones

### Steps To Reproduce

1. Go to library import for a folder with some unmatched files
2. let the processing run and attempt to match
3. if one is matched to an existing show attempt to toggle.

### Environment

```markdown
- OS: Windows 11 & ubuntu
- Sonarr: 4.0.14.2938
- Docker Install: no & yes
- Using Reverse Proxy: no & yes
- Browser: Firefox & Chrome
- Database: Sqlite 3.49.2
```

Develop

### Trace Logs?

Frontend issue, no trace logs generated for it

### Anything else?

_No response_

ricecrackerfiend:
Change the behaviour of the SelectAll checkbox input in the ImportSeriesHeader to properly account for selectable rows only.

#### Description

When rows in the ImportSeriesList are a mix of selected and unselected, a null value gets passed to the CheckboxInput in the list header. Null is treated as falsy, and so when you click on the input element, the browser tries to flip the checkbox value to true which will then select all rows.

Under normal circumstances where every item is selectable, this is fine. You can click the checkbox in the list header to select everything. The checkbox value will then become true, and you can click it again to set it to false. However, in the import series list, some rows are not selectable because the series might already exist in the user's series library or because it might not match any known series.

So if the checkbox value in the header is null, and you click on it, the browser sends an event saying that it's being flipped to true. However, it can't be set to true because some items aren't selectable. So then it gets set back to null. But wait, it was null before we clicked it and null again after we clicked it. So nothing happened as a result of the user click.

To fix it, we need to make it so the value of the checkbox in the list header only accounts for the items that are selectable. Meaning, if everything that's selectable in the import series list is selected, then the checkbox in the header should be checked with a value of true.

This PR does just that.

##### Before (taken from the linked issue)

https://private-user-images.githubusercontent.com/1117625/452624456-35b42b3c-6567-4b8a-ba78-893b0b4de087.mp4?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3NTAwMzU2MTcsIm5iZiI6MTc1MDAzNTMxNywicGF0aCI6Ii8xMTE3NjI1LzQ1MjYyNDQ1Ni0zNWI0MmIzYy02NTY3LTRiOGEtYmE3OC04OTNiMGI0ZGUwODcubXA0P1gtQW16LUFsZ29yaXRobT1BV1M0LUhNQUMtU0hBMjU2JlgtQW16LUNyZWRlbnRpYWw9QUtJQVZDT0RZTFNBNTNQUUs0WkElMkYyMDI1MDYxNiUyRnVzLWVhc3QtMSUyRnMzJTJGYXdzNF9yZXF1ZXN0JlgtQW16LURhdGU9MjAyNTA2MTZUMDA1NTE3WiZYLUFtei1FeHBpcmVzPTMwMCZYLUFtei1TaWduYXR1cmU9OTA1ODZiZWZkNmFkOGJjYzY3YTdkOTc5NWVlYzZlNzJiMjlkYmNiZTY0MzIyZmVjZmYyZmY5YTgzMjEwYWNiOCZYLUFtei1TaWduZWRIZWFkZXJzPWhvc3QifQ.gNIFSnLkeEhTQb6Vo4qJqFLgtnKQlas2Pxa2ASxArBk

##### After

https://github.com/user-attachments/assets/89dcba57-9670-47e1-8c4b-b28214d09433

* Closes #7909

### DIFF
--- a/frontend/src/AddSeries/ImportSeries/Import/ImportSeriesTable.tsx
+++ b/frontend/src/AddSeries/ImportSeries/Import/ImportSeriesTable.tsx
@@ -13,6 +13,7 @@ import {
 } from 'Store/Actions/importSeriesActions';
 import createAllSeriesSelector from 'Store/Selectors/createAllSeriesSelector';
 import createDimensionsSelector from 'Store/Selectors/createDimensionsSelector';
+import createImportSeriesAllSelectedSelector from 'Store/Selectors/createImportSeriesAllSelectedSelector';
 import { CheckInputChanged } from 'typings/inputs';
 import { SelectStateInputProps } from 'typings/props';
 import { UnmappedFolder } from 'typings/RootFolder';
@@ -54,6 +55,7 @@ function Row({ index, style, data }: ListChildComponentProps<RowItemData>) {
   );
 }
 
+const allSelectedSelector = createImportSeriesAllSelectedSelector();
 function ImportSeriesTable({
   unmappedFolders,
   scrollerRef,
@@ -78,7 +80,10 @@ function ImportSeriesTable({
   const listRef = useRef<FixedSizeList<RowItemData>>(null);
   const initialUnmappedFolders = useRef(unmappedFolders);
   const previousItems = usePrevious(items);
-  const { allSelected, allUnselected, selectedState } = selectState;
+  const { allUnselected, selectedState } = selectState;
+  const allSelected = useSelector((state: AppState) =>
+    allSelectedSelector(state, selectedState)
+  );
 
   const handleSelectAllChange = useCallback(
     ({ value }: CheckInputChanged) => {
--- a/frontend/src/Store/Selectors/createExistingSeriesSelector.ts
+++ b/frontend/src/Store/Selectors/createExistingSeriesSelector.ts
@@ -1,5 +1,17 @@
 import { createSelector } from 'reselect';
+import Series from 'Series/Series';
 import createAllSeriesSelector from './createAllSeriesSelector';
+
+export function isExistingSeries(
+  series: Series[],
+  tvdbId: number | undefined
+): boolean {
+  if (tvdbId == null) {
+    return false;
+  }
+
+  return series.some((s) => s.tvdbId === tvdbId);
+}
 
 function createExistingSeriesSelector(tvdbId: number | undefined) {
   return createSelector(createAllSeriesSelector(), (series) => {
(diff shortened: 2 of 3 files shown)


frontend/src/AddSeries/ImportSeries/Import/ImportSeriesTable.tsx
```diff
@@ -78,7 +80,10 @@ function ImportSeriesTable({
   const listRef = useRef<FixedSizeList<RowItemData>>(null);
   const initialUnmappedFolders = useRef(unmappedFolders);
   const previousItems = usePrevious(items);
-  const { allSelected, allUnselected, selectedState } = selectState;
+  const { allUnselected, selectedState } = selectState;
+  const allSelected = useSelector((state: AppState) =>
```
markus101, line 84: Instead of a separate selector this should be made more generic so the logic can be reused in other places we use this state. Probably something similar to how `toggleSelected` works, then uncheck it and add it to record / state that has the ID and `true` so we can look through all items and confirm it's either checked or disabled.

ricecrackerfiend: Just to confirm I understand your suggestion, is your idea to maintain allSelected as local state that gets re-computed when either when:
- The list of selectable items changes
- toggleSelected is invoked

And have the logic that computes allSelected be a re-usable function rather than a selector?

markus101: `allSelected` would still come from `useSelect`, `useSelect` would also include disabled items and `allSelected` would derive it's value from items being selected as well as items that are disabled. Similar to selecting an item items that are disabled would dispatch that to the state in the context.

markus101:

PR #7924, opened 2025-06-16
fix: prevent series synopsis from being cutoff

### LINKED ISSUE #7875: Long Anime Titles Prevent Synopsis from Being Seen

### Current Behavior

If you open a series page and the name of that series is too long and in windows your scale and layout is set to 125% the summary/synopsis will get cut off like in the image below:

Image

### Expected Behavior

The summary/synopsis should not be getting cut off and should be shown.

### Steps To Reproduce

1. In windows display settings, set your scale to 125%.
2. Open a series in sonarr with an extremely long name and the synopsis will be shown cut-off.

### Environment

```markdown
- OS: Windows 10 22H2
- Sonarr: 4.0.14.2939
- Docker Install: No
- Using Reverse Proxy: No
- Browser: Brave/Chrome
- Database: Default I guess? (Sorry I don't know this one.)
```

Main

### Trace Logs?

I don't believe trace-logs are required for this.

### Anything else?

_No response_

ricecrackerfiend:
#### Description
This PR addresses issues with long texts in the series title and synopsis being cutoff.

The series overview page was making use of rigid layout patterns that prevented containers from flowing to fit their content. Most notably, it was making use of JavaScript based text truncation, a technique typically used to achieve multi-line ellipses that often leads to bizarre UI behaviour (often a result of height calculations being off), slower UI responsiveness, and poorer accessibility (particularly for screen readers).

For standard prose (such as synopsis), text truncation should be considered as a last resort. It's often better to re-think how you should present such information (e.g. you could make the container scrollable or use a pop-up modal).

Rather than making that choice as part of this PR, I decided to just let the content flow for now.

I also removed a fixed-height constraint for the header and an unnecessary flex container.

##### Before (taken from issue)
A long synopsis being cut off

##### After
A long synopsis and a long title that are both fully visible

* Closes #7875

### DIFF
--- a/frontend/src/Series/Details/SeriesDetails.css
+++ b/frontend/src/Series/Details/SeriesDetails.css
@@ -5,7 +5,6 @@
 .header {
   position: relative;
   width: 100%;
-  height: 425px;
 }
 
 .backdrop {
@@ -27,22 +26,20 @@
 .headerContent {
   display: flex;
   padding: 30px;
+  gap: 35px;
   width: 100%;
   height: 100%;
   color: var(--white);
+  align-items: center;
 }
 
 .poster {
   flex-shrink: 0;
-  margin-right: 35px;
   width: 250px;
   height: 368px;
 }
 
 .info {
-  display: flex;
-  flex-direction: column;
-  flex-grow: 1;
   overflow: hidden;
 }
 
--- a/frontend/src/Series/Details/SeriesDetails.css
+++ b/frontend/src/Series/Details/SeriesDetails.css
@@ -56,14 +56,13 @@
 }
 
 .title {
-  display: -webkit-box;
   overflow: auto;
+  max-height: calc(3 * 50px);
   text-wrap: balance;
   font-weight: 300;
   font-size: 50px;
   line-height: 50px;
-  -webkit-line-clamp: 3;
-  -webkit-box-orient: vertical;
+  line-clamp: 3;
 }
 
 .toggleMonitoredContainer {
@@ -172,6 +171,7 @@
 
   .title {
     overflow: hidden;
+    max-height: calc(3 * 30px);
     font-weight: 300;
     font-size: 30px;
     line-height: 30px;
--- a/frontend/src/Series/Details/SeriesDetails.css
+++ b/frontend/src/Series/Details/SeriesDetails.css
@@ -30,7 +30,6 @@
   width: 100%;
   height: 100%;
   color: var(--white);
-  align-items: center;
 }
 
 .poster {
--- a/frontend/src/Series/Details/SeriesDetails.css
+++ b/frontend/src/Series/Details/SeriesDetails.css
@@ -26,10 +26,10 @@
 .headerContent {
   display: flex;
   padding: 30px;
-  gap: 35px;
   width: 100%;
   height: 100%;
   color: var(--white);
+  gap: 35px;
 }
 
 .poster {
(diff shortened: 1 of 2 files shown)


markus101:
I think in addition to removing the fixed height we should enforce a fixed height for the title, limiting it to 2-3 rows, either truncating the text and showing the full title in a tooltip or shrinking the text to accommodate the ridiculously long title featured here.

ricecrackerfiend:
@markus101 that's a good idea. Small breakpoints in particular benefit much more from that. I do allow the title to scroll if you exceed that 3 line limit and you're on anything bigger than a phone. Otherwise, on a phone, the overflow is hidden.


https://github.com/user-attachments/assets/5f438459-f317-44a0-a12c-a570c276c7d9

markus101:
Looking pretty good, the top of title should align with top of the poster though (like it does for a single line title currently).

ricecrackerfiend:
That makes sense, especially with the line clamp on the title text. Here's what it looks like now:

image

frontend/src/Series/Details/SeriesDetails.css
```diff
@@ -27,22 +26,19 @@
...
   flex-shrink: 0;
-  margin-right: 35px;
   width: 250px;
   height: 368px;
 }
 
 .info {
-  display: flex;
-  flex-direction: column;
-  flex-grow: 1;
```
markus101, line 45: Width is not maintained any more.

image

Reverting this change fixes it.

frontend/src/Series/Details/SeriesDetails.css
```diff
@@ -27,22 +26,19 @@
 .headerContent {
   display: flex;
   padding: 30px;
+  gap: 35px;
```
markus101: stylelint wants this below `color`

markus101 (changes requested):
Looking great for the most part now, but a couple issues to address.

frontend/src/Series/Details/SeriesDetails.css
ricecrackerfiend: I ended up just using width: 100% instead because it doesn't seem like there is a need for a flex box here. Blocks naturally stack vertically, so in this case there's no need for a vertical flex box.

frontend/src/Series/Details/SeriesDetails.css
ricecrackerfiend: Thanks for pointing this out. I didn't even realize stylelint was setup. I fixed this and some other errors as well.

ricecrackerfiend:
@markus101 there was another stylelint issue around vendor prefixing that gets in the way of using line-clamping as a universal solution. As such, I implemented a complementary "legacy" solution by setting a max height based on 3x the line height.

markus101 (approved):
Thanks for the fixes on this!

mynameisbogdan:
The `.title` changes introduced a regression on Firefox (currently using version 143) on macOS.

Screenshot

markus101:
I noticed that as well, looks like FF doesn't support the `line-clamp` (appears to be supported, but wasn't liking something with it).

mynameisbogdan:
I don't think it's `line-clamp`. Without line-height the text has a height of 76px, and barely at 60px it doesn't get a scroll.

markus101:
Yeah, it's not `line-clamp`, but the line height, descenders were also cut off (any title with a lower-case `y`). 60px does seem to be enough, but we could adjust further.

I opened up https://github.com/Sonarr/Sonarr/pull/8094

PR #7925, opened 2025-06-16
New: Support for multiple seriesIds in Rename API endpoint

nuxencs:
#### Description
This adds functionality to the API to retrieve any number of series that need renames at once. This removes the need to make a call for each movie individually when you want to know the rename status for multiple series.

### DIFF
--- a/src/NzbDrone.Core/MediaFiles/MediaFileRepository.cs
+++ b/src/NzbDrone.Core/MediaFiles/MediaFileRepository.cs
@@ -8,7 +8,9 @@ namespace NzbDrone.Core.MediaFiles
     public interface IMediaFileRepository : IBasicRepository<EpisodeFile>
     {
         List<EpisodeFile> GetFilesBySeries(int seriesId);
+        List<EpisodeFile> GetFilesBySeries(List<int> seriesIds);
         List<EpisodeFile> GetFilesBySeason(int seriesId, int seasonNumber);
+        List<EpisodeFile> GetFilesBySeason(List<int> seriesIds, int seasonNumber);
         List<EpisodeFile> GetFilesWithoutMediaInfo();
         List<EpisodeFile> GetFilesWithRelativePath(int seriesId, string relativePath);
         void DeleteForSeries(List<int> seriesIds);
@@ -26,9 +28,19 @@ namespace NzbDrone.Core.MediaFiles
             return Query(c => c.SeriesId == seriesId).ToList();
         }
 
+        public List<EpisodeFile> GetFilesBySeries(List<int> seriesIds)
+        {
+            return Query(c => seriesIds.Contains(c.SeriesId)).ToList();
+        }
+
         public List<EpisodeFile> GetFilesBySeason(int seriesId, int seasonNumber)
         {
             return Query(c => c.SeriesId == seriesId && c.SeasonNumber == seasonNumber).ToList();
+        }
+
+        public List<EpisodeFile> GetFilesBySeason(List<int> seriesIds, int seasonNumber)
+        {
+            return Query(c => seriesIds.Contains(c.SeriesId) && c.SeasonNumber == seasonNumber).ToList();
         }
 
         public List<EpisodeFile> GetFilesWithoutMediaInfo()
--- a/src/NzbDrone.Core/MediaFiles/MediaFileRepository.cs
+++ b/src/NzbDrone.Core/MediaFiles/MediaFileRepository.cs
@@ -10,7 +10,6 @@ namespace NzbDrone.Core.MediaFiles
         List<EpisodeFile> GetFilesBySeries(int seriesId);
         List<EpisodeFile> GetFilesBySeries(List<int> seriesIds);
         List<EpisodeFile> GetFilesBySeason(int seriesId, int seasonNumber);
-        List<EpisodeFile> GetFilesBySeason(List<int> seriesIds, int seasonNumber);
         List<EpisodeFile> GetFilesWithoutMediaInfo();
         List<EpisodeFile> GetFilesWithRelativePath(int seriesId, string relativePath);
         void DeleteForSeries(List<int> seriesIds);
@@ -36,11 +35,6 @@ namespace NzbDrone.Core.MediaFiles
         public List<EpisodeFile> GetFilesBySeason(int seriesId, int seasonNumber)
         {
             return Query(c => c.SeriesId == seriesId && c.SeasonNumber == seasonNumber).ToList();
-        }
-
-        public List<EpisodeFile> GetFilesBySeason(List<int> seriesIds, int seasonNumber)
-        {
-            return Query(c => seriesIds.Contains(c.SeriesId) && c.SeasonNumber == seasonNumber).ToList();
         }
 
         public List<EpisodeFile> GetFilesWithoutMediaInfo()
--- a/src/NzbDrone.Core/MediaFiles/MediaFileRepository.cs
+++ b/src/NzbDrone.Core/MediaFiles/MediaFileRepository.cs
@@ -8,7 +8,7 @@ namespace NzbDrone.Core.MediaFiles
     public interface IMediaFileRepository : IBasicRepository<EpisodeFile>
     {
         List<EpisodeFile> GetFilesBySeries(int seriesId);
-        List<EpisodeFile> GetFilesBySeries(List<int> seriesIds);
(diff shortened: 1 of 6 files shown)


src/Sonarr.Api.V3/Episodes/RenameEpisodeController.cs
```diff
@@ -17,14 +18,19 @@ public RenameEpisodeController(IRenameEpisodeFileService renameEpisodeFileServic
 
         [HttpGet]
         [Produces("application/json")]
-        public List<RenameEpisodeResource> GetEpisodes(int seriesId, int? seasonNumber)
+        public List<RenameEpisodeResource> GetEpisodes([FromQuery(Name = "seriesId")] List<int> seriesIds, int? seasonNumber)
```
markus101: Instead of merging this as one query param, keep `seriesIds` and `seriesId` separate. I think with v5 API it would make sense to only have `seriesIds`

nuxencs: I did it like this to make it backwards compatible and to be in line with the change that got merged into Radarr 😅  

separate them in the single function or separate them into two functions?
so like this:
```csharp
public List<RenameEpisodeResource> GetEpisodes(int seriesId, List<int> seriesIds, int? seasonNumber)
```
or like this:
```csharp
public List<RenameEpisodeResource> GetEpisodes(int seriesId, int? seasonNumber)
public List<RenameEpisodeResource> GetEpisodes(List<int> seriesIds, int? seasonNumber)
```

nuxencs:
Would you be fine with me doing some refatoring for some of the function names of the episode/files service/repo to make them more consistent and easier to understand?
https://github.com/Sonarr/Sonarr/blob/4b50861a6b458f11b1cf4c173df83bf4efe1cd1e/src/NzbDrone.Core/Tv/EpisodeService.cs#L25
for example I would change to
```csharp
List<Episode> GetEpisodesBySeriesId(int seriesId);
```
and change the names I chose to indicate using multiple ids and not having to overload the function

```csharp
List<Episode> GetEpisodesBySeriesIds(List<int> seriesIds);
```

src/Sonarr.Api.V3/Episodes/RenameEpisodeController.cs
markus101: I think it'd need to be a single function that branches off, otherwise I don't think it'll know how to handle the route.

I think it'd also make sense to have the `seriesIds` case not allow for season number because the season number may not exist on all seasons.

markus101:
I think it's fine to leave them as is, but to my other comment, you should use `List<Episode> GetEpisodesBySeriesIds(List<int> seriesIds);` instead of the newly added `List<Episode> GetEpisodes(List<int> seriesId, int seasonNumber);`

nuxencs:
that function doesn't take the wanted season as an argument, so wouldn't I have to filter for the correct season after the call? 🤔

markus101:
With multiple series I don't think you should accept season, so there would be three cases:

1. `seriesIds`: multiple series, all seasons
2. `seriesId`: single series, no season number
3. `seriesId` and `seasonNumber`: specific season for a single series

nuxencs:
so something like this then, or should `seriesId` be nullable?
```csharp
        [HttpGet]
        public List<RenameEpisodeResource> GetEpisodes(List<int> seriesIds, int seriesId, int? seasonNumber)
        {
            if (seriesIds is not { Count: 0 })
            {
                return _renameEpisodeFileService.GetRenamePreviews(seriesIds).ToResource();
            }

            if (seasonNumber.HasValue)
            {
                return _renameEpisodeFileService.GetRenamePreviews(seriesId, seasonNumber.Value).ToResource();
            }

            return _renameEpisodeFileService.GetRenamePreviews(seriesId).ToResource();
        }
```

markus101:
Gave this some more thought and I think we'd actually want two separate routes, if it's just one the documentation is going to be confusing.

Something like this:

```
[HttpGet]
public List<RenameEpisodeResource> GetEpisodes(int seriesId, int? seasonNumber)
{
    if (seasonNumber.HasValue)
    {
        return _renameEpisodeFileService.GetRenamePreviews(seriesId, seasonNumber.Value).ToResource();
    }

    return _renameEpisodeFileService.GetRenamePreviews(seriesId).ToResource();
}

[HttpGet("multiple")]
public List<RenameEpisodeResource> GetEpisodes(List <int> seriesIds)
{
    return new List<RenameEpisodeResource>();
}
```

The downside is it will have a separate route, but the upside is the docs are more accurate and we son't need to do as much manual validation. Not in love with the name `multiple` if you have something better in mind.

nuxencs:
I can't really think of anything better either, only thing that came to mind was `bulk`

src/NzbDrone.Core/MediaFiles/MediaFileRepository.cs
```diff
@@ -26,6 +27,11 @@ public List<EpisodeFile> GetFilesBySeries(int seriesId)
             return Query(c => c.SeriesId == seriesId).ToList();
         }
 
+        public List<EpisodeFile> GetFilesBySeries(List<int> seriesIds)
```
markus101: This will better align with `EpisodeRepository`

```suggestion
        public List<EpisodeFile> GetFilesBySeriesIds(List<int> seriesIds)
```

src/NzbDrone.Core/MediaFiles/MediaFileService.cs
```diff
@@ -17,6 +17,7 @@ public interface IMediaFileService
         void Update(List<EpisodeFile> episodeFiles);
         void Delete(EpisodeFile episodeFile, DeleteMediaFileReason reason);
         List<EpisodeFile> GetFilesBySeries(int seriesId);
+        List<EpisodeFile> GetFilesBySeries(List<int> seriesIds);
```
markus101: Same thing here.

src/Sonarr.Api.V3/Episodes/RenameEpisodeController.cs
```diff
@@ -26,5 +27,17 @@ public List<RenameEpisodeResource> GetEpisodes(int seriesId, int? seasonNumber)
 
             return _renameEpisodeFileService.GetRenamePreviews(seriesId).ToResource();
         }
+
+        [HttpGet("multiple")]
```
markus101: I think `bulk` fits better, we've used it elsewhere already too.

src/Sonarr.Api.V3/Episodes/RenameEpisodeController.cs
```diff
@@ -26,5 +27,17 @@ public List<RenameEpisodeResource> GetEpisodes(int seriesId, int? seasonNumber)
...
+        [Produces("application/json")]
+        public List<RenameEpisodeResource> GetEpisodes(List<int> seriesIds)
```
markus101: I was getting a 415 error without the `[FromQuery]]`

```suggestion
        public List<RenameEpisodeResource> GetEpisodes([FromQuery] List<int> seriesIds)
```

src/Sonarr.Api.V3/Episodes/RenameEpisodeController.cs
```diff
@@ -26,5 +27,17 @@ public List<RenameEpisodeResource> GetEpisodes(int seriesId, int? seasonNumber)
...
+        {
+            if (seriesIds is not { Count: not 0 })
```
markus101: Lot of negatives here, but looks like this can be simplified to
```suggestion
            if (seriesIds is { Count: 0 })
```

src/Sonarr.Api.V3/Episodes/RenameEpisodeController.cs
```diff
@@ -26,5 +27,17 @@ public List<RenameEpisodeResource> GetEpisodes(int seriesId, int? seasonNumber)
...
+            {
+                throw new BadRequestException("seriesIds must be provided");
+            }
+
```
markus101, line 40: Another check that none of the IDs are `0` would also be good, otherwise it leads to a 500 response.

markus101 (changes requested):
This direction looks a lot better, just some smaller issues.

src/Sonarr.Api.V3/Episodes/RenameEpisodeController.cs
nuxencs: Added a check for both negative and zero values, as ids should also never be negative 😄

markus101 (approved):

PR #7940, opened 2025-06-25
New: save queue removal options between sessions

### LINKED ISSUE #7938: Store Last Used Queue Removal Method

When I need to remove torrents from my queue for whatever reason, most of the time I do not want the torrent to be immediately removed. I am on private indexers where I need to invest time in seeding, regardless of whether I used all or any of the episodes within the torrent Sonarr chose.

Image

This means it is imperative that I do not choose 'Remove from Download Client'.

Instead, it would be best if I chose 'Change category'. But I often forget to do this, as I primarily use Usenet, where I can redownload things all day, remove them at will, and it makes no difference how the download is ejected from Sonarr.

While yes, this boils down to me just needing to remember to pick the appropriate option, I think it would be best if Sonarr allowed me to customize this, or perhaps had smarter behavior when concerning the default option for removal. See below for my elaboration on this.

The general aim is for the Removal Method to become 'Change Category' by default. Not necessarily for everyone, but at least for men, in the scenario I've already described.

There's a number of ways one could do implement this. Here are some of my ideas:

- 'Change Category' could just become the new default.
    - While it has very little complexity in the implementation, I think more people would be affected by this change than not. I also am not sure that it is 100% possible in every scenario. I would only do this if a significant number of people stated that they too were having problems with the current default.
- The default could be globally configured within Sonarr's settings (Media Management section, perhaps), where the Removal Method is set server-side, and the client is informed upon load.
- This could be a download-client-based, or indexer-based option, where the download's related download client or indexer could provide a default Removal Method.
    - If you went this route, I would recommend that the download client hold this option over the indexer, that would seem less confusing.
- The default could be configured client-side, but I doubt that's ideal, as it of course would not synchronize between clients, and Sonarr has very little persistent client-side state as far as I know.
    - If you went this route, perhaps it could simply 'remember' the last setting, and attempt to switch to that option when possible.

Image

- The default state could become dynamically informed by the existence of the 'Post-Import Category' being set on the download client. This might be one of the easiest ideas to implement, with the least amount of friction.
    - Essentially, the client associated with the download has the 'Post-Import Category' is not empty and configured, then the default Removal Method becomes 'Change Category'
    - Otherwise, it is simply 'Remove From Client' like it is now.

Given the nature of my problem, this feature is more of sparing me trouble from unintentional deletes due to mental exhaustion in the UI.

Ostensibly, I could make copies, or perhaps hard links of torrents upon completion, with some sort of custom script. But I'm not convinced, that is way too complex, and would cause me a lot more work and worry.

I'm not sure what alternatives exist that I haven't already described in the above solution description; I've already come up with a bunch of ideas, I just want to throw them all at the wall and see which one gets picked by developers, if anyone likes my idea.

### Anything else?

_No response_

stevietv:
#### Description
Saves the state of the queue removal options and don't reset them every time modal closes.

* Closes #7938

### DIFF
--- a/frontend/src/Activity/Queue/RemoveQueueItemModal.tsx
+++ b/frontend/src/Activity/Queue/RemoveQueueItemModal.tsx
@@ -1,4 +1,6 @@
-import React, { useCallback, useMemo, useState } from 'react';
+import React, { useCallback, useMemo } from 'react';
+import { useDispatch, useSelector } from 'react-redux';
+import AppState from 'App/State/AppState';
 import FormGroup from 'Components/Form/FormGroup';
 import FormInputGroup from 'Components/Form/FormInputGroup';
 import FormLabel from 'Components/Form/FormLabel';
@@ -10,6 +12,8 @@ import ModalContent from 'Components/Modal/ModalContent';
 import ModalFooter from 'Components/Modal/ModalFooter';
 import ModalHeader from 'Components/Modal/ModalHeader';
 import { inputTypes, kinds, sizes } from 'Helpers/Props';
+import { setQueueRemovalOption } from 'Store/Actions/queueActions';
+import { InputChanged } from 'typings/inputs';
 import translate from 'Utilities/String/translate';
 import styles from './RemoveQueueItemModal.css';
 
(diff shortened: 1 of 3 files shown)


markus101:
This won't be long lived in redux (been working on migrating the queue to zustand/query, but updating it on changes is still mostly the same.

stevietv:
> This won't be long lived in redux (been working on migrating the queue to zustand/query, but updating it on changes is still mostly the same.

haha timing indeed, this one has been on my mind to do for a long time!
hopefully I will understand zustand more than I do redux (which shouldn't be difficult!)

markus101:
There is a lot less boilerplate and things are typed so in theory we should write less bugs due to typos, etc... in theory.

PR #7947, opened 2025-06-29
feat: add reflink support for cifs

### LINKED ISSUE #7946: Remove Special Handling of Reflinks for BTRFS and ZFS (now handled by .net)

Like BTRFS and ZFS, CIFS also supports reflinks. Can confirm using `cp --reflink=auto /mnt/smb/foo /mnt/smb/bar` will use CoW if the underlying storage supports it, in my case BTRFS.

We should add `cifs` [here](https://github.com/Sonarr/Sonarr/blob/4b50861a6b458f11b1cf4c173df83bf4efe1cd1e/src/NzbDrone.Common/Disk/DiskTransferService.cs#L348) and [here](https://github.com/Sonarr/Sonarr/blob/4b50861a6b458f11b1cf4c173df83bf4efe1cd1e/src/NzbDrone.Common/Disk/DiskTransferService.cs#L362) like the changes made in #7566.

n/a

### Anything else?

_No response_

tabrezm:
#### Description
Adds reflink support for CIFS volumes.

* Closes #7946

### DIFF
--- a/src/NzbDrone.Common/Disk/DiskTransferService.cs
+++ b/src/NzbDrone.Common/Disk/DiskTransferService.cs
@@ -345,7 +345,7 @@ namespace NzbDrone.Common.Disk
 
             if (mode.HasFlag(TransferMode.Copy))
             {
-                if (isBtrfs || isZfs)
+                if (isBtrfs || isZfs || isCifs)
                 {
                     if (_diskProvider.TryCreateRefLink(sourcePath, targetPath))
                     {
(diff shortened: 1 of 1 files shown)


markus101:
Is this actually necessary now that v5 targets .net 8?

Are there any limitations on CIFS versions that needs to be accounted for?

Does NFS support the same, so we don't have to do this one at a time?

tabrezm:
> Is this actually necessary now that v5 targets .net 8?
> 
> Are there any limitations on CIFS versions that needs to be accounted for?
> 
> Does NFS support the same, so we don't have to do this one at a time?

Will respond on the linked issue because I think it's a bit more complex than I originally thought.

tabrezm:
Not needed with .NET 8

PR #7954, opened 2025-07-05
New: indexer option for Season Pack Seed Ratio

stevietv:
#### Description
Adds an option to define a different seed ratio for season packs. When left empty the normal ratio will be used.

#### Database Migration
YES - 220

image


src/NzbDrone.Core/Localization/Core/en.json
```diff
@@ -1017,6 +1017,8 @@
   "IndexerSettingsRejectBlocklistedTorrentHashesHelpText": "If a torrent is blocked by hash it may not properly be rejected during RSS/Search for some indexers, enabling this will allow it to be rejected after the torrent is grabbed, but before it is sent to the client.",
   "IndexerSettingsRssUrl": "RSS URL",
   "IndexerSettingsRssUrlHelpText": "Enter to URL to an {indexer} compatible RSS feed",
+  "IndexerSettingsSeasonPackSeedRatio": "Season Pack Seed Ratio",
+  "IndexerSettingsSeasonPackSeedRatioHelpText": "The ratio a season-pack torrent should reach before stopping, empty uses the download client's default. Ratio should be at least 1.0 and follow the indexers rules",
   "IndexerSettingsSeasonPackSeedTime": "Season-Pack Seed Time",
   "IndexerSettingsSeasonPackSeedTimeHelpText": "The time a season-pack torrent should be seeded before stopping, empty uses the download client's default",
```
nuxencs:
```suggestion
  "IndexerSettingsSeasonPackSeedRatio": "Season Pack Seed Ratio",
  "IndexerSettingsSeasonPackSeedRatioHelpText": "The ratio a season pack torrent should reach before stopping, empty uses the download client's default. Ratio should be at least 1.0 and follow the indexers rules",
  "IndexerSettingsSeasonPackSeedTime": "Season Pack Seed Time",
  "IndexerSettingsSeasonPackSeedTimeHelpText": "The time a season pack torrent should be seeded before stopping, empty uses the download client's default",
```
for the sake of consistency 😄

stevietv: thanks, nice catch

src/NzbDrone.Core/Indexers/SeedConfigProvider.cs
```diff
@@ -51,7 +51,9 @@ public TorrentSeedConfiguration GetSeedConfiguration(int indexerId, bool fullSea
 
             var seedConfig = new TorrentSeedConfiguration
             {
-                Ratio = seedCriteria.SeedRatio
+                Ratio = (fullSeason && seedCriteria.SeasonPackSeedRatio != null)
```
markus101: I think this could be a little confusing, if the user doesn't want to seed a ratio then it'd automatically use the normal seed ratio. I think instead we should set the value to the `SeedRatio`, probably via a migration.

Alternatively we could always use `SeasonPackSeedRatio` for season packs, but I think that's confusing in a different way (initially at least).

stevietv: You're right indeed, if its empty then it will fall back to the `SeedRatio` instead of falling back to client default. We don't want it falling back to `SeedRatio` when empty, but we also don't want it to set no `SeedRatio` after introducing this setting and it is empty.

If we set it via a migration then that would handle the case during introduction of this setting, however we might then also need to incorporate a way to ensure that going forward its also handled when user adds an indexer. If its through Prowlarr then Prowlarr needs to be in line when v5 launches and set `SeasonPackSeedRatio` when it adds an indexer, or build in some logic that when a indexer is added with out this field that it takes whatever is in `SeedRatio`. However that seems to introduce an odd step where the user would then need to open the indexer settings to explicitly clear if they want no ratio.

Another option I considered was allowing user to enter `-1` in this field to fall back to client setting, or an additional indexer setting for `use season pack ratio`.

markus101: > however we might then also need to incorporate a way to ensure that going forward its also handled when user adds an indexer

If they add it via Sonarr then it should be set in the UI. If they add it via Prowlarr then Prowlarr will need to know about the setting or it will just leave it blank. At the API level we could assume null is excluded thus we use `SeedRatio`, but we then have the same problem that we fall back if it's left out, which may be intentional.

> Another option I considered was allowing user to enter -1 in this field to fall back to client setting, or an additional indexer setting for use season pack ratio.

What's the use case for allowing fallback to the download client's setting? We don't do that else where.

An additional setting may actually make sense here, if it was an enum we could start with `Use Seed Ratio`, `Use Season Pack Seed Ratio` and potentially a do not set / use download client's default.

It'd be extra nice if we could have the default from Sonarr UI be `Use Season Pack Seed Ratio`, but the default value actually be `Use Seed Ratio`, so Prowlarr not knowing about the setting would just default to the existing setting.

stevietv: > What's the use case for allowing fallback to the download client's setting? We don't do that else where.

The idea for falling back to DL client when blank would be for example to set no ratio at all (if none is also set in the DL client). We don't have a way in Sonarr to set no ratio otherwise for season packs.

> An additional setting may actually make sense here, if it was an enum we could start with Use Seed Ratio, Use Season Pack Seed Ratio and potentially a do not set / use download client's default..It'd be extra nice if we could have the default from Sonarr UI be Use Season Pack Seed Ratio, but the default value actually be Use Seed Ratio, so Prowlarr not knowing about the setting would just default to the existing setting.

This makes sense I think and is likely the best way to handle it.

stevietv: OK, so I added this as an enum to choose whether to use season seed goals or not. I don't think a third enum is required for use DL clients defaults, this can be achieved by just leaving that goal entry blank.

I also included a migration for indexers that have a season seed time set to set the enum to the use season pack goals. It might not be the cleanest migration, i puzzled with it for a while, but it works! comments welcome!

stevietv:
image

markus101 (approved):

PR #7981, opened 2025-07-23
Fixed: Improve parsing of multi-episode with episodes in brackets

Spell6inder:
#### Description

Improves multi-episode filename parsing by expanding support for additional episode range formats.

#### Changes

- Updated regex patterns for parsing multi-episode releases in square `[]` and round `()` brackets.
- Added support for episode ranges such as `S01E1-3 of 12`, commonly used on trackers like **toloka.to**.
- Modified episode and season group patterns to accept 1–2 digits (`\d{1,2}` instead of `\d{2}`) for better compatibility with releases that use single-digit numbers.

#### Before

These formats were **not matched**:
- `Series Title [S01E1-3 of 12]`
- `Series Title (S01E1-2 of 10)`

#### After

These formats are now properly matched and parsed:
- `Series.Title.[S01E1-3 of 12]`
- `Series.Title.(S01E1-2 of 10)`

### Motivation

Some public trackers use episode ranges that include total episode counts (e.g., `1-3 of 12`). Prior regex patterns failed to extract episode numbers in such cases, causing episodes not to be recognized or imported correctly. This change improves interoperability with such naming styles.

Example
image

expected result
image


thezoggy:
do you have example release names.. and if they can only be found on one site.. prob not worth the risk to others to support such non-standard naming scheme

Spell6inder:
@thezoggy 
> do you have example release names.. and if they can only be found on one site.. prob not worth the risk to others to support such crappy naming scheme

origin https://toloka.to/t688614
`Дандадан (Сезон 2, серії 1-3 з 12) / Dandadan (Season 2) (2025) WEBDLRip 1080p H.265 Ukr/Jap | sub Ukr`

processed by prowlarr
`Дандадан (S2E1-3 of 12) / Dandadan (S2) (2025) WEB-DL 1080p H.265 Ukr/Jap | sub Ukr`

thezoggy:
with that scheme the series title wouldnt get parsed, so doesnt really matter what the season/ep info is reported at this point. as well as the quality string there..  
image

Spell6inder:
it's raw title. try processed by prowlarr, please.

image

thezoggy:
> `Дандадан (S2E1-3 of 12) / Dandadan (S2) (2025) WEB-DL 1080p H.265 Ukr/Jap | sub Ukr`

so you can see prowlarr tries to fixup some stuff..  have you tried seeing if the definition can just be updated to take this further for further cleanup of this sites format? vs trying to add site specific rules to sonarr

Spell6inder:
same for jackett
image

So it may occur not only on this site and not only with Prowlarr. That's why I decided to start from here. If this approach doesn't work, I'll try to address it on the Prowlarr and Jackett side.

src/NzbDrone.Core/Parser/Parser.cs
```diff
@@ -289,12 +289,12 @@ public static class Parser
                 new Regex(@"(?:.*(?:\""|^))(?<title>.*?)(?:[-_\W](?<![()\[]))+(?:\W?Season\W?)(?<season>(?<!\d+)\d{1,2}(?!\d+))(?:\W|_)+(?:Episode\W)(?:[-_. ]?(?<episode>(?<!\d+)\d{1,2}(?!\d+)))+",
                           RegexOptions.IgnoreCase | RegexOptions.Compiled),
 
-                // Multi-episode with episodes in square brackets (Series Title [S01E11E12] or Series Title [S01E11-12])
-                new Regex(@"(?:.*(?:^))(?<title>.*?)[-._ ]+\[S(?<season>(?<!\d+)\d{2}(?!\d+))(?:[E-]{1,2}(?<episode>(?<!\d+)\d{2}(?!\d+)))+\]",
```
markus101: Is this change actually necessary? There are no related tests for it.

Instead of `\s` `[._ ]` would be better to handle spaces, periods or underscores and matches the format already used.

Spell6inder: This is not a necessary change. I added it just because regexps look similar. Should I add tests or remove this change?

Spell6inder: adjusted and added tests

markus101: If there are release names that actually use it then let's keep it, otherwise we shouldn't add them for the sake of it.

Spell6inder: I removed changes related to square brackets

assada:
one more example `Rick and Morty (S8E1-9) (2025) WEBRip 1080p Ukr/Eng | Sub Ukr/Eng (S8E1-9)`
image

Spell6inder:
I have synced my fork

markus101:
Thanks for rebasing, looks like there are some tests failing now.

markus101:
The failing tests were due to the single digit season number, that had to be separated to a different regex to avoid stomping on all the other cases.

PR #7982, opened 2025-07-24
Fixed: Prevent using Original names with other episode file tokens

bakerboy448:
#### Description
title
* UX

### DIFF
--- a/src/NzbDrone.Core/Organizer/FileNameValidation.cs
+++ b/src/NzbDrone.Core/Organizer/FileNameValidation.cs
@@ -118,9 +118,10 @@ namespace NzbDrone.Core.Organizer
                 return false;
             }
 
-            return FileNameBuilder.SeasonEpisodePatternRegex.IsMatch(value) ||
+            return (FileNameBuilder.SeasonEpisodePatternRegex.IsMatch(value) ||
                    (FileNameBuilder.SeasonRegex.IsMatch(value) && FileNameBuilder.EpisodeRegex.IsMatch(value)) ||
-                   FileNameBuilder.AbsoluteEpisodePatternRegex.IsMatch(value) ||
+                   FileNameBuilder.AbsoluteEpisodePatternRegex.IsMatch(value) &&
+                   !FileNameValidation.OriginalTokenRegex.IsMatch(value)) ||
                    FileNameValidation.OriginalTokenRegex.IsMatch(value);
         }
     }


markus101:
Why?

bakerboy448:
sidebar where does `useCurrentFilenameAsFallback` get set as this appears to differ
- https://github.com/Sonarr/Sonarr/blob/6f1d461dad8eb146db739191d21390a34e3ba2bc/src/NzbDrone.Core/Organizer/FileNameBuilder.cs#L1129-L1132

vs downstream

https://github.com/Radarr/Radarr/blob/e29be26fc9a5570bdf37a1b9504b3c0162be7715/src/NzbDrone.Core/Organizer/FileNameBuilder.cs#L660-L663

as far as why:

https://github.com/Radarr/Radarr/issues/11174

And repeated commentary from downtream (B) and I believe Q etc in the past that OriginalTitle and OriginalFilename are not intended / designed to be used with other tokens.
https://discord.com/channels/264387956343570434/264388019585286144/1320841774398177300

SF4L:
Please don't break Sonarr's manual importing too... at least until "SceneName" is populated in Manual Imports... 

https://github.com/Radarr/Radarr/issues/11174
https://github.com/Radarr/Radarr/issues/6547

It's a seriously valid usecase (less so for TV shows than movies, I'll admit).

There are good reasons to have Movie (Year) - [Filename].mkv as your filename, as outlined in Radarr support today (https://discordapp.com/channels/264387956343570434/264388019585286144/1398252059865780286 and https://discordapp.com/channels/264387956343570434/264388019585286144/1398045700062969933).

I get the Recursive Rename issue, but there's a better solution than just making all manual imports have empty filenames if there are other tokens...

Preserving filenames when using the Manual Import functionality is important. The logic for not allowing Tokens + OriginalTitle has never really been enumerated, has it? I see B and PearsonFlyer mentioned it, but never really explained why. It always worked, until one day it didn't (in Radarr). It's still fine in Sonarr, thankfully.

There's often good information in Filenames that can't be captured other ways, and it's the only thing that is no longer retrievable from Metadata/MediaInfo once files get renamed. There are reasons to preserve it (edition information, source, etc.) that isn't always captured by TRaSH Guide's naming schemes.

Please consider fixing Radarr support, instead of breaking Sonarr support for this. Thank you.

bakerboy448:
This has nothing to do with manual import. Please stop conflating and confusing issues.

SF4L:
> This has nothing to do with manual import. Please stop conflating and confusing issues.

Here is the impact it has on Manual Import:
https://discordapp.com/channels/264387956343570434/264387956343570434/1398277850531889256

A practical downside of the implementation of this is that it breaks a manual import where `OriginalTitle` or `OriginalFilename` are used in a naming scheme. I'm sorry if it muddies the waters, but IMO it should be considered.

Unless this bug is fixed: https://github.com/Radarr/Radarr/issues/6547 , implementing this today has immediate downsides to anybody who has a perfectly reasonable naming scheme that has worked for 10+ years, similar to https://github.com/Radarr/Radarr/commit/7142d1f2249c1cbea1a1035476216c4b8350cb19 a few years ago.

SF4L:
Putting aside practical consequences of the implementation...

Strictly from a code perspective, do we think `GetOriginalFilename` should return an Empty string? An if statement to override an original filename to an empty string because of an unrelated condition. The code becomes harder to follow.

I understand why it was done (to fix an edge case recursive issue with renaming where people added weirdly stuff outside of Radarr, scanned it in, and then used Radarr to rename stuff that never had a filename in Radarr's DB to begin with), but perhaps the better solution is to try fixing _that_ scenario...

Oh well. I've said everything I can on this topic. If the devs _really_ believe that there's no valid reason why one would want to have `Movie (Year) - [Filename].mkv` as their preferred Naming Scheme (I've tried to highlight reasons for it in the Discord and GitHub issues), I'll just drop it.

mynameisbogdan:
> If the devs _really_ believe that there's no valid reason why one would want to have `Movie (Year) - [Filename].mkv` as their preferred Naming Scheme

It might lead to recursion issues on reimports.

> Please don't break Sonarr's manual importing too... at least until "SceneName" is populated in Manual Imports...

https://github.com/Sonarr/Sonarr/blob/6f1d461dad8eb146db739191d21390a34e3ba2bc/src/NzbDrone.Core/Parser/SceneChecker.cs#L7-L29

Yours with spaces isn't a valid scene name anyway.

markus101:
Sonarr has supported this for a long time, though there are limitations in how it can be used.

> It might lead to recursion issues on reimports.

This is why we prevent adding it in some cases, but the idea is you can include it when importing a file from outside the series folder and Soanrr will try to include it, otherwise it's only included when renaming and the information is available.

> Different Behavior Between Manual Import and Regular Import when {Original Title} or {Original Filename} are in Naming Scheme

Depending how this is manifesting (I haven't looked at the issue deeply, nor the code) then I'd say that should be fixed instead of ripping out a feature.

SF4L:
This comment summed it up as best I could:
https://github.com/Radarr/Radarr/issues/11174#issuecomment-3118244464

> It might lead to recursion issues on reimports.

Let's fix that!

thezoggy:
for as long as i can recall its never been advised to use other tokens with the Original* ones (and seems to be the understanding by everyone else that does support). Logically I could understand hey use this naming of the various tokens if original* ones arent populated since you dont want to rename it to nothing..  but there is always a filename? and original title generally is just assumed to be the parent folder/release name so i can only assume this then leads to funky results..

anyways where this all came about, user had funky naming on import due to fallout of mixing the tokens (hes since deleted original posts about it) but now its mainly just been him fighting about wanting to use the tokens with other stuff after we told him to stop doing that and his argument is that hes been doing this forever and thinks it should just work

image
~
image

if original* tokens may or maynot actually be used depending on the scenario might be good to call out that on the key (link to wiki page or something if its verbose) or perhaps offer a better structure supported in the naming scheme that makes it easy to understand the condition of the if/else (if empty) to be used.

something like:
`{Original.Title}||{Movie CleanTitle} {(Release Year)} - {{Edition Tags}} {[MediaInfo 3D]}{[Custom Formats]}{[Quality Full]}{[Mediainfo AudioCodec}{ Mediainfo AudioChannels]}{[MediaInfo VideoDynamicRangeType]}{[Mediainfo VideoCodec]}{-Release Group}`

SF4L:
> It might lead to recursion issues on reimports.

Tried to address potential recursion issues: https://github.com/Radarr/Radarr/pull/11179

SF4L:
Last thing I'll say: zoggy, I'm not really sure why you're editing context to try to make people look worse, but I guess you're more than welcome to do so. It seems I'm doing a pretty good job of that on my own :( . None of those posts have been deleted (you did conveniently crop out a handful, including the start of the conversation...), and everything has been pretty transparently laid out.

Just so we're on the same page on the timeline: I was advised to use that "funky" naming scheme ~7 years ago in the Sonarr support channel. About 2.5 years ago, a decision was made to stop supporting it in Radarr, as I found today as a catch-all to fix those funky renaming/recursion issues.

I apologize if things got heated and out of hand these past few days. It was never the intention to make things personal, and I've tried to avoid doing so, but it can be a bit frustrating when you're being 4v1'ed by people who don't answer _why_ (until bogdan finally did, in this PR today). Especially when one of those people calls you illiterate (speaking of deleted posts...) and someone else tells you to STFU and calls you dumb, which I know for a fact I never did to anybody.

If my way of addressing this issue came across as entitled and arrogant, I do genuinely apologize. That was never the intention. All I wanted was for someone to remind me _why_ this `Movie (Year) - [Original Filename].mkv` was considered bad now (when years ago, it was recommended), and see if we can fix it.

Anyways, I don't want to fight with anyone, especially not over free software that you guys all work so hard to support and maintain. It's all peace and love from me, happy Friday and have a great weekend. I'm sorry.

PR #8007, opened 2025-08-06
Add HasSubtitles indexer flag to BTN

oxfordllama:
#### Description

Allows prioritization of BTN torrents that have subtitles in them.

Although https://apidocs.broadcasthe.net/ does not list `Tags` in the response, a test requests against the API will demonstrate that it is present.

b
a

### DIFF
--- a/src/NzbDrone.Core.Test/IndexerTests/BroadcastheNetTests/BroadcastheNetFixture.cs
+++ b/src/NzbDrone.Core.Test/IndexerTests/BroadcastheNetTests/BroadcastheNetFixture.cs
@@ -65,7 +65,7 @@ namespace NzbDrone.Core.Test.IndexerTests.BroadcastheNetTests
             torrentInfo.Codec.Should().Be("x264");
             torrentInfo.Resolution.Should().Be("SD");
 
-            torrentInfo.Tags.Should().Contain("Subtitles");
+            torrentInfo.IndexerFlags.Should().HaveFlag(IndexerFlags.IncludesSubtitles);
         }
 
         private void VerifyBackOff()
--- a/src/NzbDrone.Core.Test/IndexerTests/BroadcastheNetTests/BroadcastheNetFixture.cs
+++ b/src/NzbDrone.Core.Test/IndexerTests/BroadcastheNetTests/BroadcastheNetFixture.cs
@@ -64,6 +64,8 @@ namespace NzbDrone.Core.Test.IndexerTests.BroadcastheNetTests
             torrentInfo.Container.Should().Be("MP4");
             torrentInfo.Codec.Should().Be("x264");
             torrentInfo.Resolution.Should().Be("SD");
+
+            torrentInfo.Tags.Should().Contain("Subtitles");
         }
 
         private void VerifyBackOff()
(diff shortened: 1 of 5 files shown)


oxfordllama:
Sorry - took me a minute to get a functional Windows environment for Visual Studio. There were a couple syntax errors which are resolved now. Screenshots included.

markus101 (approved):
Thanks for contributing.

src/NzbDrone.Core.Test/IndexerTests/BroadcastheNetTests/BroadcastheNetFixture.cs
```diff
@@ -64,6 +64,8 @@ public async Task should_parse_recent_feed_from_BroadcastheNet()
             torrentInfo.Container.Should().Be("MP4");
             torrentInfo.Codec.Should().Be("x264");
             torrentInfo.Resolution.Should().Be("SD");
+
+            torrentInfo.Tags.Should().Contain("Subtitles");
```
markus101: I think you want `torrentInfo.IndexerFlags.Should().HaveFlag(IndexerFlags.HasSubtitles);` instead.

src/NzbDrone.Core/Parser/Model/IndexerFlags.cs
```diff
@@ -43,6 +43,11 @@ public enum IndexerFlags
         /// <summary>
         /// The release is nuked
         /// </summary>
-        Nuked = 128
+        Nuked = 128,
+
+        /// <summary>
+        /// The release contains subtitles
+        /// </summary>
+        HasSubtitles = 256
```
markus101: After looking at this again I think `Subtitles` or `IncludesSubtitles` may be a better name here, the former matches the existing naming, but may be too generic

markus101 (changes requested):

src/NzbDrone.Core.Test/IndexerTests/BroadcastheNetTests/BroadcastheNetFixture.cs
johnmaguire: Thank you - that's correct. I believe I figured out how to run the tests and they are now passing.

oxfordllama: Thank you. I think I figured out how to run the tests and they are now passing.

src/NzbDrone.Core/Parser/Model/IndexerFlags.cs
oxfordllama: Renamed to `IncludesSubtitles` but can update with whatever you like. I did consider "Subbed" (a la "Nuked") but I feared this may cause confusion on foreign titles. I attempted to use language that described the torrent, matching "Nuked", rather than using a noun for what was inside, but "Subtitles" is shorter which is compelling.

markus101: Yeah, the naming for such a simple thing is awkward, I think `IncludeSubtitles` might be the best we can get.

oxfordllama: The other tags use descriptive language, so the imperative mood is a bit unexpected, but ultimately I do not care what the tag is named as long as I can promote subtitled releases :)

oxfordllama: done

markus101: We ended up discussing a bit more on Discord and just `Subtitles` was came up a few times, after thinking on it a bit more, I've switched it to that and should land later today. Thanks again for contributing and the discussion.

markus101 (approved):

PR #8014, opened 2025-08-11
Adding excluded tags to release profiles

Tro95:
#### Description
This PR extends the existing tags feature of release profiles to additionally offer excluded tags, allowing release profiles to specifically not apply to shows that are tagged with one of the excluded tags. This is necessary to create a strict default release profile, and to override it with less strict release profiles by tagging a show with an excluded tag of the default release profile.

Primary motivation is https://github.com/Lidarr/Lidarr/issues/5563, but it makes sense for Sonarr too.

Screenshot 2025-08-11 at 14 08 59
Screenshot 2025-08-11 at 14 09 25

#### Database Migration
YES - Adding an excluded tag field to the release profile table.

Primary motivation is https://github.com/Lidarr/Lidarr/issues/5563, but it makes sense for Sonarr too.

### DIFF
--- a/frontend/src/Settings/Profiles/Release/EditReleaseProfileModalContent.tsx
+++ b/frontend/src/Settings/Profiles/Release/EditReleaseProfileModalContent.tsx
@@ -77,7 +77,8 @@ function EditReleaseProfileModalContent({
   const { item, isFetching, isSaving, error, saveError, ...otherProps } =
     useSelector(createReleaseProfileSelector(id));
 
-  const { name, enabled, required, ignored, tags, excludedTags, indexerId } = item;
+  const { name, enabled, required, ignored, tags, excludedTags, indexerId } =
+    item;
 
   const dispatch = useDispatch();
   const previousIsSaving = usePrevious(isSaving);


stevietv:
Interesting idea, i have a few initial questions/comments
* I would expect some validation to not allow the tag to be in both fields
* UI wise, I'd expect the excluded tags to be red on the profiles box (final screenshot), like we do for negated CFs for example
* no need to edit the openapi.json - this will be generated on PR merge
* ~migration 221 is already claimed by an open PR, please adjust~
*

markus101:
> migration 221 is already claimed by an open PR, please adjust

It is currently, but that migration is not needed for that PR, so it's safe to use here and will be dropped from that PR

Tro95:
I have removed the changes to openapi.json, and I have updated the UI to have the red boxes for excluded tags (see screenshots).

Screenshot 2025-08-11 at 18 12 10
Screenshot 2025-08-11 at 18 12 17

I will work on validation next, but my C# is not very good. Is adding another SharedValidator rule inside ReleaseProfileController the correct way to do this on the backend?

Tro95:
I have added validation to prevent the same tag being applied as both inclusion and exclusion criteria.
Screenshot 2025-08-11 at 22 33 52
Screenshot 2025-08-11 at 22 34 46
Screenshot 2025-08-11 at 22 34 58

Tro95:
I have added some tests, and I believe this PR is ready for review.

src/NzbDrone.Core.Test/Profiles/Release/ReleaseProfileServiceFixture.cs
```diff
@@ -0,0 +1,76 @@
...
+            _excludedReleaseProfile = _releaseProfiles[2];
+            _excludedReleaseProfile.ExcludedTags.Add(2);
+
+            Mocker.GetMock<IRestrictionRepository>()
+                  .Setup(s => s.All())
+                  .Returns(_releaseProfiles);
+        }
+
+        [Test]
+        public void untagged_artist_should_match_default_and_excluded_release_profiles()
```
stevietv:
```suggestion
        public void untagged_series_should_match_default_and_excluded_release_profiles()
```

stevietv (changes requested):

src/NzbDrone.Core.Test/Profiles/Release/ReleaseProfileServiceFixture.cs
Tro95: Thanks, I have changed it to `all_for_tags_should_match_default_and_excluded_release_profiles_for_untagged_series` to better match the other test names.

stevietv (approved):

src/Sonarr.Api.V3/Profiles/Release/ReleaseProfileController.cs
```diff
@@ -43,6 +43,11 @@ public ReleaseProfileController(IReleaseProfileService profileService, IIndexerF
                 {
                     context.AddFailure(nameof(ReleaseProfile.IndexerId), "Indexer does not exist");
                 }
+
+                if (restriction.ExcludedTags.Intersect(restriction.Tags).Any())
+                {
+                    context.AddFailure(nameof(ReleaseProfile.ExcludedTags), "Release profile should not have the same tag as an include and an exclude");
+                }
```
markus101: This reads a little awkward to me and doesn't need `Release profile`.

I think something like `Tag cannot be in 'Tags' and 'Excluded Tags'` would be better, but I think it should go a step further and list the tags that are actually intersecting, will be a little more involved, perhaps like `RootFolderValidator` is done, but it'll be a lot more friendly to users.

markus101: Also having it outside the single custom validator would be good.

Tro95: I have improved the error message (see screenshot), and refactored the validation. I believe this addresses your feedback, but I wasn't quite sure what you meant by moving outside of the single custom validator. I'm new to C# and dotnet; I also tried to make a validator class for tags similar to the various path validators, but I wasn't successful in making it work and it seemed a bit unnecessary unless you expect to have excluded tags on other entities than release profiles?

Screenshot 2025-08-21 at 14 34 21

Tro95:
I have realized that I am not properly handling the excluded tags property in the Tags settings page (see screenshot), so I'd suggest I need to address that before this PR should be merged. I can't progress this PR for the next 2+ weeks, so anyone else is welcome to take it over if desired.
Screenshot 2025-08-21 at 14 41 43

Tro95:
I've added support for excluded release profiles in the Tags settings, so I think this PR is ready for another review and merge (if approved).
Screenshot 2025-08-21 at 18 35 01

Screenshot 2025-08-21 at 18 35 30

stevietv:
Wouldn't it be better in the UI just to add the uses together on release profiles and group as `X Release profiles` whether its in the excluded field or the included field.
The wording now implies that the release profiles are excluded.

Tro95:
> Wouldn't it be better in the UI just to add the uses together on release profiles and group as `X Release profiles` whether its in the excluded field or the included field. The wording now implies that the release profiles are excluded.

I think it makes sense to distinguish between whether the release profile is included or excluded based on the tag. I could put both under "Release Profiles", but I'd still need subheadings for included/excluded or some other visual distinction. The current UI for release profiles associated with a Tag could definitely be improved, such as using the name of the release profiles, but I think those improvements should be in a separate PR. I kept things simple for this PR, with a view to a later PR improving the UI.

markus101:
I agree with Stevie, the way it's displayed incorrectly associates them with something that doesn't exist since there is no such thing as a `Excluded Release Profile`. I think just keeping it under Release Profile is sufficient and looks better.

Tro95:
I have moved them all under Release Profiles. New UI:

Screenshot 2025-09-05 at 21 14 25

Screenshot 2025-09-05 at 21 14 31

Screenshot 2025-09-05 at 21 14 35

stevietv:
I still don't think there needs to be a distinction between included and excluded in this UI. 
An indication that a tag is used on a release profile should be sufficient.

Tro95:
Ok, I reverted to the original UI. Now excluded release profiles simply show up as another release profile resource under tags.

Screenshot 2025-09-10 at 20 07 26

src/NzbDrone.Core/Tags/TagDetails.cs
```diff
@@ -10,6 +10,7 @@ public class TagDetails : ModelBase
         public List<int> SeriesIds { get; set; }
         public List<int> NotificationIds { get; set; }
         public List<int> RestrictionIds { get; set; }
+        public List<int> ExcludedReleaseIds { get; set; }
```
markus101: I think `ReleaseProfile` makes more sense, in v5 API we could also rename `RestrictionIds` to `ReleaseProfileIds`.

```suggestion
        public List<int> ExcludedReleaseProfileIds { get; set; }
```

src/NzbDrone.Core/Tags/TagService.cs
```diff
@@ -92,6 +92,7 @@ public TagDetails Details(int tagId)
             var importLists = _importListFactory.AllForTag(tagId);
             var notifications = _notificationFactory.AllForTag(tagId);
             var restrictions = _releaseProfileService.AllForTag(tagId);
+            var excludedReleases = _releaseProfileService.AllExcludedForTag(tagId);
```
markus101:
```suggestion
            var excludedReleaseProfiles = _releaseProfileService.AllExcludedForTag(tagId);
```

src/Sonarr.Api.V3/Profiles/Release/ReleaseProfileController.cs
```diff
@@ -44,6 +47,11 @@ public ReleaseProfileController(IReleaseProfileService profileService, IIndexerF
                     context.AddFailure(nameof(ReleaseProfile.IndexerId), "Indexer does not exist");
                 }
             });
+
+            SharedValidator.RuleFor(d => d.Tags.Intersect(d.ExcludedTags))
+                .Empty()
+                .WithName("ExcludedTags")
+                .WithMessage(d => string.Format("'{0}' cannot be in both 'Tags' and 'Excluded Tags'", string.Join(", ", _tagService.GetTags(d.Tags.Intersect(d.ExcludedTags)).ConvertAll(t => t.Label))));
```
markus101: Use string interpolation instead of `string.Format` and `Select` instead of `ConvertAll`

src/NzbDrone.Core/Datastore/Migration/221_add_exclusion_tags_to_release_profiles.cs
```diff
@@ -0,0 +1,14 @@
...
+using NzbDrone.Core.Datastore.Migration.Framework;
+
+namespace NzbDrone.Core.Datastore.Migration
+{
+    [Migration(221)]
+    public class add_exclusion_tags_to_release_profiles : NzbDroneMigrationBase
+    {
+        protected override void MainDbUpgrade()
+        {
+            Alter.Table("ReleaseProfiles").AddColumn("ExcludedTags").AsString().NotNullable();
```
markus101: As-is this can't be NotNullable as the default value is `Null`, you probably want `.WithDefaultValue("[]")` too.

src/NzbDrone.Core.Test/Profiles/Release/ReleaseProfileServiceFixture.cs
```diff
@@ -0,0 +1,76 @@
...
+namespace NzbDrone.Core.Test.Profiles
+{
+    [TestFixture]
+    public class ReleaseProfileServiceFixture : CoreTest<ReleaseProfileService>
+    {
+        private List<ReleaseProfile> _releaseProfiles;
+        private ReleaseProfile _defaultReleaseProfile;
+        private ReleaseProfile _includedReleaseProfile;
+        private ReleaseProfile _excludedReleaseProfile;
+        [SetUp]
```
markus101, line 26:
```suggestion

        [SetUp]
```

src/NzbDrone.Core.Test/Profiles/Release/ReleaseProfileServiceFixture.cs
```diff
@@ -0,0 +1,76 @@
...
+        public void Setup()
+        {
+            _releaseProfiles = Builder<ReleaseProfile>.CreateListOfSize(3).TheFirst(1)
+                .With(r => r.Required = ["required_one"])
+                .TheNext(1)
+                .With(r => r.Required = ["required_two"])
+                .TheNext(1)
+                .With(r => r.Required = ["required_three"])
+                .Build()
+                .ToList();
```
markus101: Assign the tags directly, but you need one more as well to ensure you don't get a release profile

```suggestion
            _releaseProfiles = Builder<ReleaseProfile>.CreateListOfSize(4)
                .TheFirst(1)
                .With(r => r.Required = ["required_one"])
                .TheNext(1)
                .With(r => r.Required = ["required_two"])
                .With(r => r.Tags = [1])
                .TheNext(1)
                .With(r => r.Required = ["required_three"])
                .With(r => r.ExcludedTags = [2])
                .TheNext(1)
                .With(r => r.Required = ["required_four"])
                .With(r => r.Tags = [3])
                .With(r => r.ExcludedTags = [4])
                .Build()
                .ToList();
```

src/NzbDrone.Core.Test/Profiles/Release/ReleaseProfileServiceFixture.cs
```diff
@@ -0,0 +1,76 @@
...
+
+            _defaultReleaseProfile = _releaseProfiles[0];
+            _includedReleaseProfile = _releaseProfiles[1];
+            _includedReleaseProfile.Tags.Add(1);
+            _excludedReleaseProfile = _releaseProfiles[2];
+            _excludedReleaseProfile.ExcludedTags.Add(2);
```
markus101:
```suggestion
            _defaultReleaseProfile = _releaseProfiles[0];
            _includedReleaseProfile = _releaseProfiles[1];
            _excludedReleaseProfile = _releaseProfiles[2];
            _includedAndexcludedReleaseProfile = _releaseProfiles[3];
```

src/NzbDrone.Core/Profiles/Releases/ReleaseProfileService.cs
```diff
@@ -35,14 +36,19 @@ public List<ReleaseProfile> All()
...
+
         public List<ReleaseProfile> AllForTag(int tagId)
         {
             return _repo.All().Where(r => r.Tags.Contains(tagId)).ToList();
         }
 
         public List<ReleaseProfile> AllForTags(HashSet<int> tagIds)
         {
-            return _repo.All().Where(r => r.Tags.Intersect(tagIds).Any() || r.Tags.Empty()).ToList();
+            return _repo.All().Where(r => r.Tags.Intersect(tagIds).Any() || (r.Tags.Empty() && !r.ExcludedTags.Intersect(tagIds).Any())).ToList();
```
markus101: This doesn't correctly exclude profiles that contain a wanted tag & and excluded tag, I think you want:

```suggestion
            return _repo.All().Where(r => (r.Tags.Intersect(tagIds).Any() || r.Tags.Empty()) && !r.ExcludedTags.Intersect(tagIds).Any()).ToList();
```

src/NzbDrone.Core.Test/Profiles/Release/ReleaseProfileServiceFixture.cs
```diff
@@ -0,0 +1,76 @@
...
+
+        [Test]
+        public void all_for_tags_should_match_default_and_excluded_release_profiles_for_untagged_series()
+        {
+            var releaseProfiles = Subject.AllForTags([]);
+            releaseProfiles.Should().Equal([_defaultReleaseProfile, _excludedReleaseProfile]);
+        }
+
+        [Test]
+        public void all_for_tags_should_match_default_and_included_and_not_excluded_release_profiles_for_tagged_series()
```
markus101: Too many `and` in here to make it easy to understand. Worth taking another pass at how these are named.

src/NzbDrone.Core/Tags/TagDetails.cs
Tro95: I went and made them generally more consistent without changing the API for restrictionIds in https://github.com/Sonarr/Sonarr/pull/8014/commits/92a5509c2e74bfbdbd0eafb768d8459fac6b19d7. It should be simple to change the name for V5.

src/NzbDrone.Core/Profiles/Releases/ReleaseProfileService.cs
Tro95: Either wanted tags or excluded tags have to take priority over the other. The way I originally implemented this feature makes the wanted tags take precendence over the excluded tags, i.e., a release profile that matches both a wanted tag and an excluded tag would be applied, rather than skipped. Your proposed change would reverse this logic, i.e., a release profile that matches both a wanted tag and an excluded tag would be skipped, rather than applied. I don't mind which way round this should function, so please let me know which you prefer.

PR #8016, opened 2025-08-12
Correctly parse backslashes from Transmission

JackDanger:
#### Description

Attempting to fix the issue where the Transmission download client responds with backslashes to escape characters and we mistake it for mixed platform paths.

Example from the system status page:
```
Unable to communicate with Transmission. Cannot combine OsPaths of different platforms ('/mnt/downloads/Sonarr' + 'Don\'t+Fail+Me+Now')
```

#### Database Migration
NO

### DIFF
--- a/src/NzbDrone.Core.Test/Download/DownloadClientTests/TransmissionTests/TransmissionFixture.cs
+++ b/src/NzbDrone.Core.Test/Download/DownloadClientTests/TransmissionTests/TransmissionFixture.cs
@@ -273,6 +273,35 @@ namespace NzbDrone.Core.Test.Download.DownloadClientTests.TransmissionTests
             items.First().OutputPath.Should().Be(@"C:\Downloads\Finished\transmission\" + _title);
         }
 
+        [Test]
+        public void should_fix_backward_slashes()
+        {
+            WindowsOnly();
+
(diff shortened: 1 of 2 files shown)


markus101:
Transmission seems to do weird things, depending how things are encoded when submitted to it. With https://github.com/Sonarr/Sonarr/pull/7937 this should be better based on the discussion in the linked issue, but still going to be at the mercy of Transmission for existing torrents and potentially ones added elsewhere.

src/NzbDrone.Core.Test/Download/DownloadClientTests/TransmissionTests/TransmissionFixture.cs
```diff
@@ -273,6 +273,35 @@ public void should_fix_forward_slashes()
...
+                        LeftUntilDone = 100,
+                        DownloadDir = "Can\'t touch this"
+                    };
+                });
+
+            var items = Subject.GetItems().ToList();
+
+            items.Should().HaveCount(1);
+            items.First().OutputPath.Should().Be(
+                    @"C:\Downloads\Finished\transmission\" + \
```
rumplin, line 301: I don't undestand this test. Should the fix not work on all parts of the string?
like that: 
`@"C:\Downloads\Finished\transmission\Can't touch this\Don't Look Back in Anger");`

PR #8023, opened 2025-08-18
Library import progress bar

ghost:
#### Description
Added a progress bar during library import
Added a counter while mapping files before import

Related to [Radarr PR #11208
](https://github.com/Radarr/Radarr/pull/11208)

https://github.com/user-attachments/assets/0439a52a-472e-41e7-a370-aa00bd42598d

https://github.com/user-attachments/assets/2439726c-5b42-48e6-86c0-5bb41bc7aeeb

#### Database Migration
NO


frontend/src/AddSeries/ImportSeries/Import/ImportSeriesFooter.tsx
```diff
@@ -39,9 +40,35 @@ function ImportSeriesFooter() {
...
+  const {
+    isLookingUpSeries,
+    isImporting,
+    items,
+    importError,
+    importingCount = 0,
+    importedCount = 0,
+  } = useSelector(
+    (
+      state: AppState & {
```
markus101, line 52: This needs to be part of the `importSeries` state in `AppState`

markus101 (commented):
I don't like the approach here, visually the text on the progress bar moving as the progress does doesn't look very good, but having to send `n` requests for each series is unnecessary when it can be accomplished in a single request.

The API response alone doesn't indicate that the series has been fully added, since there is a lot more work for refreshing and scanning that happens in the background so tracking the progress like this doesn't really add a lot of value.

I don't think the work here makes sense with a looming work to move to API v5 and refactor the UI functionality.

ghost:
> I don't like the approach here, visually the text on the progress bar moving as the progress does doesn't look very good, but having to send `n` requests for each series is unnecessary when it can be accomplished in a single request.

I can either stick it in the middle or remove it, whichever you think would be best. I personally don't find it very useful since there's an import count above, but I thought most people probably like percentages.
Regarding the `n` requests, I tried multiple approaches, but that's the only way I found to have proper import progress tracking without adding any backend code

> The API response alone doesn't indicate that the series has been fully added, since there is a lot more work for refreshing and scanning that happens in the background so tracking the progress like this doesn't really add a lot of value.

That was my concern while working on the Radarr import, but I wasn't sure because even tho I know updating the metadatas is the heavier part, it was almost seemless on my tests, so I asked in the radarr-testing discord channel. It seems that tracking the import without the metadatas update was sufficient, but I wouldn't mind trying to implement full process tracking

bakerboy448:
> It seems that tracking the import without the metadatas update was sufficient, but I wouldn't mind trying to implement full process tracking

assumption being full process would be too difficult to track and from the UX viewpoint the movie is added. Any refresh is behind the scenes after adding.

markus101:
With SignalR the UI receives updates whenever a new series is added, so it could be tracked that way, would just need to track which series are added and confirm that they're in the list of series to be added, but again that doesn't really track much, the refresh (pretty quick) and then the scan (could be quite long) are the important aspects that aren't really going to be accounted for in this approach or one that checks for the series being added.

markus101:
Closing as this current approach isn't something we're going to implement. I'll keep progress in mind during the refactor and see if there is something useful that could be added.

PR #5194, opened 2022-10-16
Additional funny loading messages

ChristinMilloy:
#### Database Migration
NO

#### Description
Got bored seeing the same loading messages over and over again so I wrote 41 more.

PR #8025, opened 2025-08-20
Update LoadingMessage.tsx

Schmidsfeld:
Some extra loading messages. to increase the selection at startup.

#### Description
I simly added some extra loading messages that i hope that they are funny. if you smile at least 2-3 times while reading trough them please consider this PR.


stevietv:
ref - https://github.com/Sonarr/Sonarr/pull/5194

markus101:
If we end up adding some they'll probably come from things that have been said (like the python one) or from the community in general.

PR #8033, opened 2025-08-27
Allow Auto Redirects when Fetching Custom Lists

kenkangxgwe:
#### Description
A useful scenario: Using Google Sheets to maintain a list and leveraging Apps Script to provide Restful API. Due to security reasons, Google Apps Script will redirects the client to another URL in order to fetch the correct JSON results.

### DIFF
--- a/src/NzbDrone.Common/Http/HttpRequestBuilder.cs
+++ b/src/NzbDrone.Common/Http/HttpRequestBuilder.cs
@@ -390,5 +390,12 @@ namespace NzbDrone.Common.Http
 
             return this;
         }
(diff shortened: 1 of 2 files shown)


markus101 (approved):
Thanks!

kenkangxgwe:
Hi, do you think this non-breaking, small improvement can be cherry-picked to v4, so it can take effect earlier?

markus101:
At this time only critical bug or security issues will be back ported to v4. If one arises additional changes may be brought back, but nothing is planned at this time.

PR #8046, opened 2025-09-02
New: Allow timezone adjustment

Audionut:
#### Description
Allow user to manually specify a timezone, for related UI such as calendar.
This allows users with intentionally obscure browser timezone to display accurate dates/times in Sonarr.

Screenshot 2025-09-02 210908

I asked an LLM to build a key list for the timezones, and it returned a complete list, to say the least. I had considered condensing the list, but since it's a set and forget.....

### DIFF
--- a/frontend/src/Settings/UI/UISettings.tsx
+++ b/frontend/src/Settings/UI/UISettings.tsx
@@ -21,9 +21,9 @@ import createLanguagesSelector from 'Store/Selectors/createLanguagesSelector';
 import createSettingsSectionSelector from 'Store/Selectors/createSettingsSectionSelector';
 import themes from 'Styles/Themes';
 import { InputChanged } from 'typings/inputs';
+import timeZoneOptions from 'Utilities/Date/timeZoneOptions';
 import titleCase from 'Utilities/String/titleCase';
 import translate from 'Utilities/String/translate';
-import timeZoneOptions from 'Utilities/Date/timeZoneOptions';
 
 const SECTION = 'ui';
 
--- a/frontend/src/Utilities/Date/convertToTimezone.ts
+++ b/frontend/src/Utilities/Date/convertToTimezone.ts
@@ -1,19 +1,24 @@
 import moment from 'moment-timezone';
 
-export const convertToTimezone = (date: moment.MomentInput, timeZone: string) => {
+export const convertToTimezone = (
+  date: moment.MomentInput,
+  timeZone: string
+) => {
   if (!date) {
     return moment();
   }
-  
+
   if (!timeZone) {
     return moment(date);
   }
-  
+
   try {
-    // Use moment-timezone to convert to the specified timezone
     return moment.tz(date, timeZone);
   } catch (error) {
-    console.warn(`Error converting to timezone ${timeZone}. Using system timezone.`);
+    console.warn(
+      `Error converting to timezone ${timeZone}. Using system timezone.`,
+      error
+    );
     return moment(date);
   }
 };
--- a/frontend/src/Utilities/Date/timeZoneOptions.ts
+++ b/frontend/src/Utilities/Date/timeZoneOptions.ts
@@ -8,10 +8,10 @@ export const timeZoneOptions: EnhancedSelectInputValue<string>[] = [
       return translate('SystemDefault');
     },
   },
-  
+
   // UTC
   { key: 'UTC', value: 'UTC' },
-  
+
   // Africa (Major cities and unique timezones)
   { key: 'Africa/Abidjan', value: 'Africa/Abidjan' },
   { key: 'Africa/Algiers', value: 'Africa/Algiers' },
@@ -21,7 +21,7 @@ export const timeZoneOptions: EnhancedSelectInputValue<string>[] = [
   { key: 'Africa/Lagos', value: 'Africa/Lagos' },
   { key: 'Africa/Nairobi', value: 'Africa/Nairobi' },
   { key: 'Africa/Tripoli', value: 'Africa/Tripoli' },
-  
+
   // America - North America (Major US/Canada zones)
   { key: 'America/New_York', value: 'America/New_York (Eastern)' },
   { key: 'America/Chicago', value: 'America/Chicago (Central)' },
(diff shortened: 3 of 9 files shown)


frontend/src/Utilities/Date/convertToTimezone.ts
```diff
@@ -0,0 +1,21 @@
+import moment from 'moment-timezone';
+
+export const convertToTimezone = (date: moment.MomentInput, timeZone: string) => {
+  if (!date) {
+    return moment(date);
+  }
+  
+  if (!timeZone || timeZone === '') {
```
markus101: `!timeZone` covers all nullish values, including, `null`, `undefined` or empty string

frontend/src/Utilities/Date/convertToTimezone.ts
```diff
@@ -0,0 +1,21 @@
...
+    return moment(date);
+  }
+  
+  try {
+    // Use moment-timezone to convert to the specified timezone
+    return moment.tz(date, timeZone);
+  } catch (error) {
+    console.warn('Error converting to timezone', timeZone, '- using system timezone');
```
markus101: Use string interpolation instead, it makes it way easier to read. The last piece should be it's own sentence like is done elsewhere in error messaging.

frontend/src/Settings/UI/UISettings.tsx
```diff
@@ -67,6 +67,458 @@ export const timeFormatOptions: EnhancedSelectInputValue<string>[] = [
   { key: 'HH:mm', value: '17:00/17:30' },
 ];
 
+const timeZoneOptions: EnhancedSelectInputValue<string>[] = [
```
markus101: image

Do we need this to be a hardcoded list? If yes, put it somewhere else, this file doubled in size for a single drop down list, it'd be a good thing to have it's it's own separate setting.

> I asked an LLM to build a key list for the timezones, and it returned a complete list, to say the least. I had considered condensing the list, but since it's a set and forget.....

Is it actually a complete list? Not that AI has been known to completely make up a fact or anything...
I disagree that it's set and forget, it mostly is, but there will be eventually be changes.

src/Sonarr.Api.V5/Config/UiConfigController.cs
```diff
@@ -0,0 +1,39 @@
+using System.Reflection;
```
markus101: The v5 API endpoints will be added when the UI work for it starts, no need for this to be added.

frontend/src/Settings/UI/UISettings.tsx
Thirrian: Does it need to be a "complete" list or just a list ranging from -12 to +14 and the specials in between, for something your're going to set once?
https://en.m.wikipedia.org/wiki/List_of_UTC_offsets
I'm in UTC+1 which is about 50 entries of countries/cities for example. Honestly couldn't care less to have to scroll past hundreds of entries in the entire list just to set an offset :wink:

Maybe I'm not getting the issue, in that case my apologies.

Audionut: `moment` is timezone aware when using identifiers like `America/New_York`.

It's nice for people to have a list of cities within their region, instead of trying to decide if some city thousands of kilometers away is in their timezone. It's also nicer than `UTC+` imo.

frontend/src/Utilities/Date/convertToTimezone.ts
```diff
@@ -0,0 +1,21 @@
+import moment from 'moment-timezone';
+
+export const convertToTimezone = (date: moment.MomentInput, timeZone: string) => {
+  if (!date) {
+    return moment(date);
```
markus101: Since we don't have a date, should this be `return moment()` or `return ''`?

frontend/src/Utilities/Date/convertToTimezone.ts
```diff
@@ -0,0 +1,21 @@
+import moment from 'moment-timezone';
+
+export const convertToTimezone = (date: moment.MomentInput, timeZone: string) => {
```
markus101: I don't love this name, but not sure what would be better 🤷

frontend/src/Utilities/Date/timeZoneOptions.ts
```diff
@@ -0,0 +1,183 @@
+import { EnhancedSelectInputValue } from 'Components/Form/Select/EnhancedSelectInput';
+
+export const timeZoneOptions: EnhancedSelectInputValue<string>[] = [
+  { key: '', value: 'System Default' },
```
markus101: `System Default` should be translated:

```suggestion
  {
    key: '',
    get value(): translate('SystemDefault'),
  },
```

frontend/src/Settings/UI/UISettings.tsx
```diff
@@ -67,6 +68,8 @@ export const timeFormatOptions: EnhancedSelectInputValue<string>[] = [
   { key: 'HH:mm', value: '17:00/17:30' },
 ];
 
+
+
```
markus101: These extra lines should be removed.

frontend/src/Utilities/Date/timeZoneOptions.ts
```diff
@@ -0,0 +1,189 @@
...
+  { key: 'Pacific/Fiji', value: 'Pacific/Fiji' },
+  { key: 'Pacific/Guam', value: 'Pacific/Guam' },
+  { key: 'Pacific/Tahiti', value: 'Pacific/Tahiti' },
+  { key: 'Pacific/Apia', value: 'Pacific/Apia' },
+  { key: 'Pacific/Tongatapu', value: 'Pacific/Tongatapu' },
+  { key: 'Pacific/Port_Moresby', value: 'Pacific/Port_Moresby' },
+  { key: 'Pacific/Noumea', value: 'Pacific/Noumea' },
+];
+
+export default timeZoneOptions;
```
markus101: Just an empty line at the end.

```suggestion
export default timeZoneOptions;

```

markus101 (approved):

PR #8055, opened 2025-09-07
Bump FluentMigrator to official 6.2.0

mynameisbogdan:
#### Description
Removing the dependency on a custom build of FluentMigrator by using a custom SQLiteTypeMap that contains a custom type map for DateTimeOffset.

- Why FM 6.2.0 and not latest 7.x? 6.x is the latest one working with Microsoft.Extensions.DependencyInjection 8.x, as 7.x releases require Microsoft.Extensions.DependencyInjection 9.x.
- Since 5.x CompatibilityMode is set to STRICT, so [no more silent failures](https://github.com/fluentmigrator/fluentmigrator/blob/1cb851622ab6f8e11f8fdf67a9804d0a75a8546b/src/FluentMigrator.Runner.Core/Generators/CompatibilityModeExtension.cs#L35-L43). 
- Sonarr doesn't require `AlterDefaultConstraintExpression` yet, but Radarr does so it's best to have it in upstream.

### DIFF
--- a/src/NuGet.Config
+++ b/src/NuGet.Config
@@ -8,6 +8,5 @@
     <add key="SQLite" value="https://pkgs.dev.azure.com/Servarr/Servarr/_packaging/SQLite/nuget/v3/index.json" />
     <add key="coverlet-nightly" value="https://pkgs.dev.azure.com/Servarr/coverlet/_packaging/coverlet-nightly/nuget/v3/index.json" />
     <add key="FFMpegCore" value="https://pkgs.dev.azure.com/Servarr/Servarr/_packaging/FFMpegCore/nuget/v3/index.json" />
-    <add key="FluentMigrator" value="https://pkgs.dev.azure.com/Servarr/Servarr/_packaging/FluentMigrator/nuget/v3/index.json" />
   </packageSources>
 </configuration>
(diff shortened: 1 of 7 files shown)


markus101:
Going to be looking at .net 10 soonish since it's hit RC, no point targeting .net 8 when it's going to be EOL within a year of release.

mynameisbogdan:
Already played with 10 rc1 yesterday, it works but it's little buggy on Rider. 

It should be easier to bump from FM 6 to 7, than it was from 3 to 6. Do you plan to merge this PR soon or wait for .NET 10?

Personally I think the switch to nullable is going to be a much biggest change.

markus101:
Yeah, nullable is a bit of a trip when everything has been designed before that existed, already have some of it in API v5 though.

I think it makes sense to land this instead of playing the waiting game on .net 10 / rider and everything.

mynameisbogdan:
Good then.

screenshot

I'm already running this PR and #8056 on Radarr, and so far I have no issues with any of them.

stevietv:
We may need to move to FM 7 to get the [fix for this](https://github.com/fluentmigrator/fluentmigrator/issues/2063)
Currently [migration 9](https://github.com/Sonarr/Sonarr/blob/v5-develop/src/NzbDrone.Core/Datastore/Migration/009_fix_renameEpisodes.cs) fails on at least Norwegian and Finnish locales

We may otherwise be able to force it by casting the `-1` to an int.

mynameisbogdan:
Sadly that fix isn't yet released and we don't know when it's going to be tagged, and FM 7 requires a .NET 9 dependency which conflicts with the current 8 Sonarr is using. 

> We may otherwise be able to force it by casting the `-1` to an int.

You mean a string in this case?

I didn't think too much on the issue, but I would assume MigrationController should be using InvariantCulture in this case as a possible fix to prevent any further `ToString` issues.

PR #8073, opened 2025-09-16
Fix external restart handling

TypNull:
#### Description
This commit prevents internal restart service conflicts when running under external process managers (systemd or containerized environments). Previously it would always attempt to spawn a new process on restart, causing boot loops when external systems were already managing the service lifecycle.

The implementation adds a new "Use External Restart Service" setting in containerized environments, allowing users to choose between internal restart handling.

The checkbox appears only in containerized environments.

#### Database Migration
NO, but settings are stored in the configuration file.

#### Issues Fixed 
Resolves restart-related boot loops when a external process managers conflict with Sonarr's internal restart mechanism. This issue is particularly noticeable in the Lidarr plugin branch during plugin installations or updates.

### DIFF
--- a/src/NzbDrone.Core/Configuration/ConfigFileProvider.cs
+++ b/src/NzbDrone.Core/Configuration/ConfigFileProvider.cs
@@ -34,7 +34,6 @@ namespace NzbDrone.Core.Configuration
         int SslPort { get; }
         bool EnableSsl { get; }
         bool LaunchBrowser { get; }
-        bool ExternalRestart { get; }
         AuthenticationType AuthenticationMethod { get; }
         AuthenticationRequiredType AuthenticationRequired { get; }
         bool AnalyticsEnabled { get; }
@@ -181,8 +180,6 @@ namespace NzbDrone.Core.Configuration
         public bool EnableSsl => _serverOptions.EnableSsl ?? GetValueBoolean("EnableSsl", false);
 
         public bool LaunchBrowser => _appOptions.LaunchBrowser ?? GetValueBoolean("LaunchBrowser", true);
-
-        public bool ExternalRestart => _appOptions.ExternalRestart ?? GetValueBoolean("ExternalRestart", true);
 
         public string ApiKey
         {
--- a/src/NzbDrone.Host/AppLifetime.cs
+++ b/src/NzbDrone.Host/AppLifetime.cs
@@ -71,11 +71,28 @@ namespace NzbDrone.Host
         {
             if (_runtimeInfo.RestartPending && !_runtimeInfo.IsWindowsService)
             {
-                if (_runtimeInfo.IsSystemdService || (_runtimeInfo.IsContainerized && _configFileProvider.ExternalRestart))
+                if (_runtimeInfo.IsSystemdService)
                 {
-                    var restartHandler = _runtimeInfo.IsSystemdService ? "systemd" : "container runtime";
-                    _logger.Info("Skipping process spawn, letting {0} handle restart", restartHandler);
+                    _logger.Info("Skipping process spawn, letting systemd handle restart");
                     return;
+                }
+
+                if (_runtimeInfo.IsContainerized)
+                {
+                    _logger.Info("Waiting for external restart in containerized environment...");
+                    const int maxWaitSeconds = 10;
+                    for (var i = 0; i < maxWaitSeconds; i++)
+                    {
+                        Thread.Sleep(1000);
+
+                        if (_processProvider.Exists(ProcessProvider.SONARR_PROCESS_NAME))
+                        {
+                            _logger.Info("Sonarr was restarted by external process.");
+                            return;
+                        }
+                    }
+
+                    _logger.Info("External restart did not occur within {0} seconds, spawning process.", maxWaitSeconds);
                 }
 
                 var restartArgs = GetRestartArgs();
--- a/src/NzbDrone.Common/Options/AppOptions.cs
+++ b/src/NzbDrone.Common/Options/AppOptions.cs
@@ -5,5 +5,4 @@ public class AppOptions
     public string InstanceName { get; set; }
     public string Theme { get; set; }
     public bool? LaunchBrowser { get; set; }
-    public bool? ExternalRestart { get; set; }
 }
--- a/src/Sonarr.Api.V3/Config/HostConfigResource.cs
+++ b/src/Sonarr.Api.V3/Config/HostConfigResource.cs
@@ -14,7 +14,6 @@ namespace Sonarr.Api.V3.Config
         public int SslPort { get; set; }
         public bool EnableSsl { get; set; }
         public bool LaunchBrowser { get; set; }
-        public bool ExternalRestart { get; set; }
         public AuthenticationType AuthenticationMethod { get; set; }
         public AuthenticationRequiredType AuthenticationRequired { get; set; }
         public bool AnalyticsEnabled { get; set; }
@@ -61,7 +60,6 @@ namespace Sonarr.Api.V3.Config
                 SslPort = model.SslPort,
                 EnableSsl = model.EnableSsl,
                 LaunchBrowser = model.LaunchBrowser,
-                ExternalRestart = model.ExternalRestart,
                 AuthenticationMethod = model.AuthenticationMethod,
                 AuthenticationRequired = model.AuthenticationRequired,
                 AnalyticsEnabled = model.AnalyticsEnabled,


markus101:
Can we just detect this automatically?

Even if we can't I don't see value in having this as a setting in the UI.

TypNull:
I couldn't find a way to detect Docker restart policies automatically. Docker doesn't expose these settings to containers, and detecting them would require mounting the Docker socket (`/var/run/docker.sock`), which isn't feasible in many deployment scenarios for security reasons.

Removing this option entirely would break functionality for users who run containers without restart policies (the default is `restart: no`). In those cases, internal application restarts would cause the container to exit completely instead of restarting, even when the restart was intentional.

While most users likely rely on the external restart mechanisms, some prefer or require internal restart handling. The manual setting accommodates both use cases.

I experienced this firsthand during development. I initially forgot to specify `--restart=always` when running my container, which meant the application would exit on restart attempts rather than actually restarting. This highlighted how common it is to run containers without proper restart policies configured.

PS: Not sure why the diff shows such large changes in en.json.

stevietv:
Nice idea to skip the restart for these environments.

I also don't see the need to expose this in the UI. Restart should be skipped for all docker/containerized environments so we could just remove all the UI changes in this PR and skip the restart if the environment shouldn't do it.

also, the en.json change needs reverting, json tabs are two, and normally we only edit the en.json in the PR and let the weblate pull it in and expose the key for other languages, editing the other language jsons in GH can cause conflicts that need resolving,

Not sure if changing isDocker to isContainerized in the API should be considered a breaking change.

TypNull:
It’s not my idea, it’s an issue within Lidarr. I’m not sure whether changing from `isDocker` to `isContainerized` counts as a breaking change, but adding detection for Podman and the .NET environment variable felt like an improvement. The UI button only appears in those environments. However, if you think it’s better to remove them, I’ll revert all frontend changes and either remap `isContainerized` back to `isDocker`, or remove support for these additional systems entirely.

Also, does modifying the other localized JSON files actually cause problems? Honestly, some of them could need a bit of sorting.

markus101:
Yeah, it'd be a breaking change in the API.

I think What we'd want to do here is add `isPodman` as well as `isContainerized` to keep the two consistent and then in UI things in the future we can just use `isContainerized`.

> I experienced this firsthand during development. I initially forgot to specify `--restart=always` when running my container, which meant the application would exit on restart attempts rather than actually restarting. This highlighted how common it is to run containers without proper restart policies configured.

What's the experience if we do restart, does docker properly detect that it's running the process or does it think it's abandoned and nothing is running?

mynameisbogdan:
> the .NET environment variable felt like an improvement. 

That environment variable is only used on some official dotnet images because they have some conditional code based on that, and it's not the norm. 

https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-environment-variables#dotnet_running_in_container-and-dotnet_running_in_containers
https://redirect.github.com/dotnet/dotnet-docker/issues/3997

Thus it requires changes in the image building pipelines by 3rd parties since Sonarr doesn't build its own image.

TypNull:
I'll add `isDocker` back to maintain API compatibility and introduce both `isContainerized` and `isPodman` as new properties, ensuring the v3 API doesn't break.

Regarding the restart behavior: Docker and systemd do consider the application stopped when the PID changes that's exactly the issue this PR addresses. When Sonarr restarts itself internally, the PID changes, and external process managers see this as the application terminating. Sonarr restarts and Docker/systemd react too or if neither restarts, the result is effectively a container shutdown rather than a restart. This is why I added the checkbox specifically for containerized environments.

For the .NET environment variable: You're correct that it's primarily used in official .NET container images. I tested this with Hotio's Lidarr images and confirmed they don't set this variable. I haven't tested LinuxServer images yet, but my reasoning was that anyone new building images would likely start from official base images. Additionally, if a containerized environment doesn't use Docker, the images could still set the environment variable to ensure it's detected as a managed environment. I thought using a standardized approach was better than relying on completely custom detection methods, especially since it provides an additional detection layer without breaking existing functionality.

The current builds don't require any changes since the Docker detection continues to work exactly as it did before.

TypNull:
I just wanted to ask, is there still anything I'm missing? 
As the bug is a bit frustrating.

mynameisbogdan:
Any purpose for IsPodman to be exposed in the API when it's used in the UI? Shouldn't be enough for internal use? Personally I would deprecate IsDocker and only use `isContainerized` in v5. 🤷 

FYI DOTNET_RUNNING_IN_CONTAINER can be either `true` and `1`.

TypNull:
No, `IsPodman` isn’t actually used, I just assumed Markus wanted it included. If you’d prefer to deprecate `IsDocker`, that’s entirely up to you. I can leave the property in place but remove all references to it or do nothing, since this affects the v3 API and I’m not sure whether you plan to migrate it to v5 later. I’ll update the `DOTNET_RUNNING_IN_CONTAINER` check to account for the value being `1` as well.

src/NzbDrone.Common/Instrumentation/Sentry/SentryTarget.cs
```diff
@@ -168,9 +168,13 @@ public void InitializeScope()
 
         public void UpdateScope(IOsInfo osInfo)
         {
+            // SentrySdk.ConfigureScope(scope =>
```
markus101: Remove this

frontend/src/Settings/General/HostSettings.tsx
```diff
@@ -184,6 +188,20 @@ function HostSettings({
           />
         </FormGroup>
       )}
+
+      {isContainerized ? (
```
markus101: I don't think it's necessary to add this to the UI, I imagine a lot of docker/podman users would just set it as an environment variable and could be confusing to others.

I also don't think it's limited to containerized setups, could also be restarted by system.d or the like.

TypNull: Sorry, I'm not quite sure I follow. Are you saying that if someone uses Docker, they rely entirely on an external restart mechanism and if they don’t, they should? From my perspective, systemd shouldn't be the issue. It’s designed to handle automatic restarts; otherwise, it wouldn’t make sense to configure it in the first place. That’s just my take, which differs from how I approach things with Docker.

What I don’t understand is how an environment variable would help in this scenario. Should I just remove the container check, even though it wouldn’t make a difference outside of Docker? Or are you suggesting I modify the C# code to treat it as an override setting?

A bit more elaboration would really help 😅

markus101: > What I don’t understand is how an environment variable would help in this scenario. 

Settings in the config file can be set in the config file or via environment variables.

This is a pretty advanced setting, thus making it config file only would be better. I'm saying this shouldn't be an option in the UI and people can either set it in the config file directly, or more likely in a containerized environment like docker, they'd set it in via an environment variable.

markus101:
I didn't think of this before, but regarding my original comment on detecting it, instead could we have a wrapper that waits to see whether Sonarr starts up again automatically, if it does, it just exits, if it doesn't it could start Sonarr again. This is how the update process is handled for things running under systemd.

TypNull:
You could implement it this way, but it would require launching an external service and monitoring it. It's feasible, though I believe it's beyond the scope of this PR

src/NzbDrone.Core/Configuration/ConfigFileProvider.cs
```diff
@@ -181,6 +182,8 @@ public string BindAddress
 
         public bool LaunchBrowser => _appOptions.LaunchBrowser ?? GetValueBoolean("LaunchBrowser", true);
 
+        public bool ExternalRestart => _appOptions.ExternalRestart ?? GetValueBoolean("ExternalRestart", true);
```
mynameisbogdan: Are we going to make this option enabled by default? I think it's supposed to be false here.

```suggestion
        public bool ExternalRestart => _appOptions.ExternalRestart ?? GetValueBoolean("ExternalRestart", false);
```

frontend/src/System/useIsContainerized.ts
```diff
@@ -0,0 +1,9 @@
+import useSystemStatus from './useSystemStatus';
+
+function useIsContainerized() {
+  const { isContainerized } = useSystemStatus();
+
+  return isContainerized ?? false;
```
mynameisbogdan: Personally I would use destructing with default instead here, so we can catch any bad value and keep it in line with the rest of the code base.

```suggestion
  const { isContainerized = false } = useSystemStatus();

  return isContainerized;
```

src/NzbDrone.Core/Configuration/ConfigFileProvider.cs
TypNull: I need clarification on the intended behavior and hierarchy of this setting:

**1. Override hierarchy:** Should `ExternalRestart` act as the master override that supersedes all automatic detection (Docker/systemd/Podman)? Or should it work alongside the containerized environment detection?

**2. Default behavior:** If `ExternalRestart` defaults to `false`, does this mean:
- All Docker users must explicitly set `ExternalRestart=true` via environment variable or config file to fix their boot loop issues?
- Or will the automatic `isContainerized` detection handle this automatically?

For a Docker user experiencing the boot loop problem this PR aims to solve:
- Do they need to manually enable this setting, or
- Will running in a containerized environment automatically trigger the external restart behavior?
- How should it work with systemd?

If `ExternalRestart` defaults to `false` but requires manual activation, users experiencing boot loops won't get automatic fixes, they'll need to know about this setting first. This seems counterintuitive for a fix.

Could you clarify the intended logic flow?
I think `true` would be the correct default value.

mynameisbogdan: Looking again at the changes, it's basically always enabled for systemd and opt-out for `isContainerized=true`. If I understood correctly.

Then neglect my suggestion to make it opt-in if it only affects docker installations and this is needed to avoid the multiple instances running issue for which we removed the shutdown/reboot buttons from the UI. 

Don't forget to revert the non-en.json localization changes, as it's going to be a nightmare with cherry-picking downstream since Weblate handles those and they might be stale already when someone decides to cherry pick them.

markus101:
> You could implement it this way, but it would require launching an external service and monitoring it. It's feasible, though I believe it's beyond the scope of this PR

Less that it needs an external service, more that the Restart action would start an external application/script that would wait a predetermined amount of time before starting Sonarr again, if a new process for it didn't start within X seconds.

It would replace the PR as it's done now because it wouldn't need to have a setting. The logic for how this is done in the update process is [here](https://github.com/Sonarr/Sonarr/blob/v5-develop/src/NzbDrone.Update/UpdateEngine/InstallUpdateService.cs#L148-L169), though `StartNzbDrone` has additional pieces to ensure it starts with the same arguments.

Though ther

src/NzbDrone.Host/AppLifetime.cs
```diff
@@ -71,6 +71,30 @@ private void OnAppStopped()
...
+                {
+                    _logger.Info("Skipping process spawn, letting systemd handle restart");
+                    return;
+                }
+
+                if (_runtimeInfo.IsContainerized)
+                {
+                    _logger.Info("Waiting for external restart in containerized environment...");
+                    const int maxWaitSeconds = 10;
+                    for (var i = 0; i < maxWaitSeconds; i++)
```
markus101, line 84:
```suggestion
                    _logger.Info("Waiting for external restart in containerized environment...");
                    const int maxWaitSeconds = 10;

                    for (var i = 0; i < maxWaitSeconds; i++)
```

frontend/src/System/useIsContainerized.ts
markus101, line 1: This can be removed

frontend/src/typings/Settings/General.ts
```diff
@@ -27,6 +27,7 @@ export default interface General {
   urlBase: string;
   instanceName: string;
   applicationUrl: string;
+  externalRestart: boolean;
```
markus101, line 30: `externalRestart` can be cleaned up now

TypNull:
I wanted to test this, but I was unable to reproduce the issue with my Docker container. To properly test this, I would need to:

1. Build an old Lidarr version (since Backer disabled the automatic restart as a temporary fix)
2. Build a local Docker container and verify that it shows the restart bug
3. Apply my fix to see if it resolves the issue
4. Switch between the two versions to confirm the issue is completely reproducible

All of this would need to be done on a quite slow VM, as I don't have access to a fully configured Linux desktop. This process is time-consuming, and unfortunately, I haven't had the time to complete it.

As it stands, the "Revert localisation" version has been tested, but the "Testout internal restart" has not. I'm still uncertain whether this type of monitoring system will work when running from within the process itself, or if I need to trigger an external script (e.g., a Bash script) to monitor the application. Additionally, as I mentioned before, I still believe this is outside the scope of this PR.

I'm marking this PR as closed and will continue working on it when I have time, since you're requesting a monitor solution instead of a setting. I'll need some free time and much more important some motivation to do this properly.

PR #8091, opened 2025-09-28
New: Support PEM format for SSL certificates

### LINKED ISSUE #8087: Configure SSL with PEM certificate and key

Currently SSL can only be configured using a PFX (PKCS 12) file.
This prevents integration with systems that use the more common PEM file format.
For example:
https://github.com/truenas/apps/pull/3253
> I will only consider it if it can accept the cert using the certificates from the truenas system (Like we do on other apps).
> From a quick look it accepts only pfx (ie no PEM).

1. make the "SSL Cert Path" option accept a PEM-encoded certificate (in addition to PFX).
2. add a "SSL Key Path" option that accepts a PEM-encoded key (unused if using PFX).
3. use those files to configure SSL

Running a reverse proxy but this adds complexity and would not allow to access Radarr from the TrueNAS Apps Dashboard.

### Anything else?

It could be implemented in the `ValidateSslCertificate` method by calling `X509Certificate2.CreateFromEncryptedPem`
https://github.com/Sonarr/Sonarr/blob/1610e546500a0278862b43ae2bb966a11a646536/src/NzbDrone.Host/Bootstrap.cs#L280

markus101:
#### Description
SSL Key Path is a new option for use with PEM based certificates, ignore if you're using a PFX certificate.

If the SSL Cert Path ends in `.pem` or SSL Key Path is set then Sonarr will attempt to load the certificate as a PEM-based certificate

image

* Closes #8087


src/Sonarr.Api.V3/Config/CertificateValidator.cs
```diff
@@ -33,16 +34,31 @@ protected override bool IsValid(PropertyValidatorContext context)
                 return true;
             }
 
+            var certPath = resource.SslCertPath;
+            var keyPath = resource.SslKeyPath;
+            var certPassword = resource.SslCertPassword;
+            var isPem = certPath.GetPathExtension() == ".pem" || keyPath.IsNotNullOrWhiteSpace();
```
mynameisbogdan: I think you could make use of `X509Certificate2.GetCertContentType()` in this case. `X509ContentType.Cert` for pem files and `Pkcs12` (or `Pfx` since it's an alias) for pfx files.

markus101: Awesome, that's much nicer.

jcassette:
Thank you!

PR #8093, opened 2025-09-28
New Feature: adding an ability to downgrade based of how long the media has been held for.

Zippy-boy:
#### Description
WIP!

Adding the ability to downgrade to a different profile after x amount of days since the media was added.

There is a couple files that I have accidently added, but i will remove them in the final commit. 

image

### DIFF
--- a/frontend/src/Settings/Profiles/Quality/EditQualityProfileModalContent.tsx
+++ b/frontend/src/Settings/Profiles/Quality/EditQualityProfileModalContent.tsx
@@ -651,10 +651,7 @@ function EditQualityProfileModalContent({
                         type={inputTypes.QUALITY_PROFILE_SELECT}
                         name="downgradeToProfileId"
                         {...downgradeToProfileId}
-<<<<<<< HEAD
                         value={downgradeToProfileId.value ?? ''}
-=======
->>>>>>> cfedfaae4952b3aa5a2b0df3c5afd1defdd034b6
                         onChange={handleInputChange}
                       />
                     </FormGroup>


markus101:
Just basing this on the screenshot, but I don't think this makes sense to have at the profile level, a profile should cause the series to make changes to the profile based on something in the profile.

This feels more like a separate entity or profile that can make changes to series based on different parameters, kind of like how Auto Tagging works. This would allow for more flexibility and avoid people asking for other things to be factored into the decision to change the profile.

This is something that warrants more discussion and planning to ensure it's something we want to support and maintain.

markus101:
Just basing this on the screenshot, but I don't think this makes sense to have at the profile level, a profile should cause the series to make changes to the profile based on something in the profile.

This feels more like a separate entity or profile that can make changes to series based on different parameters, kind of like how Auto Tagging works. This would allow for more flexibility and avoid people asking for other things to be factored into the decision to change the profile.

This is something that warrants more discussion and planning to ensure it's something we want to support and maintain.

Zippy-boy:
That makes sense. This is only a quick mock up. I could see people asking for more variables to switching profiles and how that wouldn't be easy to maintain. Im just proposing a solution, but I think it may make more sense as an external application.

markus101:
Could definitely be a standalone application, but really depends how many different knobs people want.

framerate:
@Zippy-boy @markus101 what's the status/thinking on this in 2026? This is a feature I want (After I finish a series, downgrade from 4k -> 1080p for my family to watch at their leisure) and I was thinking about doing a similar PR.

Is someone working on a standalone app for this instead?

markus101:
Nothing moving on this particular PR. I think the biggest concern here is number of things that will feed into reducing the quality. The original work here was number of days, but you mentioned wanting to do it after it's watched, which Sonarr has no knowledge of and to throw my own wrench, what about season 1 watched when season 2 is announced / released, which profile would be selected then?


I'm going to close this PR, but it'd be worth a new GHI to discuss and decide how it should work before any further effort is put towards a solution I think.

Zippy-boy:
@framerate @markus101 The problem was having separate quantities for different seasons of shows as markus said. I don't think that implementing the ability for know when a user has watched something would get too hard, as there is already a jellyfin integration. The problem is with how quality profiles are designed the the inability to have a different quality per episode or season. This also stunts any external tools in this area (as decussed befor) and would require a rework of sonarr. This would be possible in radarr though, but that is not my interest.

PR #8110, opened 2025-10-04
Fix Queue v5 API resource mapping when no episodes present

stevietv:
#### Description
API can throw a null error when no episodes are present when mapping to a resource, causing a queue load error.

### DIFF
--- a/src/Sonarr.Api.V5/Queue/QueueResource.cs
+++ b/src/Sonarr.Api.V5/Queue/QueueResource.cs
@@ -53,10 +53,10 @@ namespace Sonarr.Api.V5.Queue
             {
                 Id = model.Id,
                 SeriesId = model.Series?.Id,
-                EpisodeIds = model.Episodes.Select(e => e.Id).ToList(),
+                EpisodeIds = model.Episodes?.Select(e => e.Id).ToList() ?? new List<int>(),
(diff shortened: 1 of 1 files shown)


src/Sonarr.Api.V5/Queue/QueueResource.cs
```diff
@@ -53,10 +53,10 @@ public static QueueResource ToResource(this NzbDrone.Core.Queue.Queue model, boo
             {
                 Id = model.Id,
                 SeriesId = model.Series?.Id,
-                EpisodeIds = model.Episodes.Select(e => e.Id).ToList(),
+                EpisodeIds = model.Episodes?.Select(e => e.Id).ToList() ?? new List<int>(),
                 SeasonNumbers = model.SeasonNumber.HasValue ? new List<int> { model.SeasonNumber.Value } : new List<int>(),
```
mynameisbogdan: How about this new syntax? 😄 Looks like stylecop unstable doesn't complain about them now.

```suggestion
                EpisodeIds = model.Episodes?.Select(e => e.Id).ToList() ?? [],
                SeasonNumbers = model.SeasonNumber.HasValue ? [model.SeasonNumber.Value] : [],
```

PR #8112, opened 2025-10-04
Fixed: qBittorrent /login API success check

Polgonite:
#### Description
qBittorrent has recently pushed an update to their /login API which caused Radarr authentication check to fail. This change handles both the old and the new authentication response.

More info of qBittorrent change here:
https://github.com/qbittorrent/qBittorrent/pull/23202

Discussion on this fix here:
https://github.com/Radarr/Radarr/pull/11258

### DIFF
--- a/src/NzbDrone.Core/Download/Clients/QBittorrent/QBittorrentProxyV2.cs
+++ b/src/NzbDrone.Core/Download/Clients/QBittorrent/QBittorrentProxyV2.cs
@@ -434,8 +434,8 @@ namespace NzbDrone.Core.Download.Clients.QBittorrent
                 }
                 catch (HttpException ex)
                 {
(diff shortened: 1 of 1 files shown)


DrStrange:
I updated to Sonarr 4.0.16.2942 and Radarr 6.0.4.10281 which includes this PR. It works with qBittorrent 5.1.2. It works with qBittorrent 5.2.0alpha1 CI workflow builds earlier than 16th October 2025. However, it doesn't seem to work on CI workflow builds of qBittorrent 5.20alpha1 from job `8951 WebAPI:add support for authenticating via API key` (16th October 2025) onwards. Sonarr & Radarr show an authentication failure for the Download Clients>qBittorrent>username setting. I tested a few qBittorrent 5.2.0alpha1 builds from 16th October onwards, including the latest job build. I did not test every build.

PR #8124, opened 2025-10-11
Fix: Validate `release/push` download client configuration

s0up4200:
#### Description
Sonarr now double-checks `release/push` payloads and immediately turns away anything that references a non-existing or disabled download client, returning a clear 400 instead of stumbling later in the pipeline.

N/A

#### Database Migration
NO

None


markus101:
Thanks for your original pass on this. I ended up taking a different direction to implement this primarily so it could be shared between different API versions by keeping this logic out of the controller (what we should have done originally). Also makes the implementation a lot easier to build upon later including testing.

https://github.com/Sonarr/Sonarr/pull/8302

Issue #1128, opened 2016-02-07
Integrate generic aggregated search capability


Taloth:
Just so the user can search for arbitrary text on the configured indexers (tv categories only) and send it to the download client.
Sonarr wouldn't have to associate it with a series/episode, it's just so the user doesn't have to go to each indexer's website.

croon:
I noticed this being added to the 3.x milestone. I'm however unsure about the request:

Is this asking for a "front page" search functionality not associated with anything? "Sonarr wouldn't have to associate it with a series/episode"

Because if so, issue #1736 is asking for something else, namely Show X, episode Y manual search. When I can't find anything with manual search, I can input free text that I know will yield a result not normally associated with that episode (often the case with specials), and since I found it through that episode's manual/free text search that download would automatically be associated with that episode.

It would be tremendously useful over the current case of:

1) Going through several indexers manually to find the episode.
2) Download it.
3) Rename it to fit Sonarrs/theTVDB's convention.
4) Put it in a folder where Sonarr can pick it up.

New flow when regular auto/manual search fails:
1) Free text search "Show Y special 2017".
2) Click download on the correct match.
( 3) Sonarr handles it like a regular match.)

This is what I'm imagining would be a good implementation for me, but is this what the original request meant? I think this is what #1736 meant at least.

master0044657:
Is someone working in this.? It would be an awsome feature

xinseo:
Any progress on this? This would be extremely helpful for power users.

CTalvio:
I want this so bad as I have several shows that do not have up to date info on alternate titles on thexem.de (some really old animes) and the fact that shows like Panty & Stocking get searched with the word "and" instead of "&" which screws with trackers that only accept exact search terms.

Neolel:
I would like this as well. Just today, I have encountered a problem which could have been solved easily by having this exact feature.

bakerboy448:
a simple +1 on the OP would suffice

comments indicating "me too" or "I want this too" and other not productive comments are extremely frowned upon

louman:
Just like yours? Your comment has created two more useless comments so far ;)

markus101:
No short term plans for this one.

FazalAAli:
Would it be possible to add it to a long term roadmap if it is not possible in the short term? Clearly there is a huge need for this as it is not always possible to update the indexers or tv databases

FazalAAli:
[...]

PR #8132, opened 2025-10-21
feat: Add support for manual searching

AgeManning:
#### Description

I'm a sonarr user and I found having a manual search option very useful. I noticed a number of other users also agree this would be a good feature. See eg: https://github.com/Sonarr/Sonarr/issues/1128. 

I noticed that there were no plans to implement it, so I have made a version. This PR could possibly be reduced, however it was made with the vision that Sonarr could also automatically import the manually selected episode, however going down that path lead me to a much larger PR. Consider this a pre-cursor and if there is interest a future PR could be made that auto-imports the resulting download.

## Overview

This adds an option to a series called "Enable Manual Search" (and thus requires a schema upgrade). I did this to have this feature disabled by default so it impacts as little users as possible):

If enabled for a series, a user can perform a "Manual Search" for an episode (via the spanner icon:

Once clicked it opens the modal:

Which allows users to enter a search string for trackers to return a list of results. The user can select their chosen episode and manually download it. It gets tied to the episode they manually searched for. 

Sonarr tracks the download and associates it with the episode that was chosen. 

### Manual Import Still Required

This PR still requires the user to then manually import the resulting file. There is no guarantee that Sonar can parse the manually downloaded search files. 

Potentially a future PR can modify the importing logic to tie the specific episode to the manually searched file, but this is left for future work. 

Note: I understand that this PR is quite large and may not want to be included into Sonarr. If there is enough users wanting this I can maintain a small fork, as I'll likely be doing this for my own personal use. I just figured may potentially be worth upstreaming given the support on the linked issue.


markus101:
Not opposed to the idea, but Prowlarr offers a similar solution, so I'm curious what benefit this would add over that.

I don't like the approach here of having an option to enable a search instead of integrating it at the season/episode level into the existing Interactive Search. That'd have to be rethought and reworked to be a little less bolted on like it appears to be now.

Keep in mind there is a quite a lot of refactoring going on in the app right now on the front end and no new features are going to be added to the v3 API.

AgeManning:
I haven't used Prowlarr, if it solves the issue, maybe I'll switch to that. 

Issue currently being that sonarr cannot index some series, or rather the producers of the series use a non-conventional naming convention. I can of course download manually outside of Sonarr, but it is a bit annoying, and figured being able to just click, search and download inside sonarr is more useful than making users have a second path. 

I avoided adding to the interactive search because I wanted to have as minimal impact as possible to current users (i.e they would have to enable a non-default option). But this then invokes a db migration, so perhaps just adding it to interactive search is the simpler approach. 

My language is rust, cs and type-script are not my go-to languages, so there's that. If there's a bunch of refactoring going on, perhaps its best to close this until v3 is released. If there is still interest, and a better design approach, I might have another crack putting it into the interactive search.

markus101:
Prowlarr isn't a replacement for Sonarr, but more of a indexer aggregator, it can push indexers to Sonarr/Radarr and has a built-in search function. Not exactly what you did here, but already supports this functionality.

I'm going to close for now, but potentially something after v5 releases.

startswithaj:
Such a great feature. Would love to see this. 

>I don't like the approach here of having an option to enable a search instead of integrating it at the season/episode level into the existing Interactive Search. That'd have to be rethought and reworked to be a little less bolted on like it appears to be now.

This makes sense to me. 

It would also be great to have the current search term displayed in the interactive search modal. Often I will need to check the logs to see what Sonarr is searching for.

Issue #7534, opened 2024-12-31
Support for IPV6 addresses via Host field on Add Download Client


Ttfgggf:
### Is there an existing issue for this?

- [X] I have searched the existing open and closed issues

### Current Behavior

So if I add an IPv4 address or subdomain, it seems to work.
If I add an IPv6 address even with square brackets or without, it somehow cannot connect to my download client.

### Expected Behavior

IPv6 addresses (whether ULA or GUA) can be used within the host field for Add Download Client.

### Steps To Reproduce

1. Set up Sonarr as usual.
2. Add Download Client with IPv6 address and it will say it cannot connect to qBittorrent.

### Environment

```markdown
- OS: Fedora 41
- Sonarr: Sonarr 4.0.11.2680
- Docker Install: Yes
- Using Reverse Proxy: No
- Browser: Firefox 90
- Database: SQlite
```

### What branch are you running?

Main

### Trace Logs?

Here is a log of what happened
https://privatebin.net/?0c1ae5131ad8756a#72kGb62wkvgtaDC4RpUD8394E2bQLuRybcNsmXrUqboj

### Anything else?

![Screenshot 2024-12-31 170453](https://github.com/user-attachments/assets/c85b2940-41d8-490e-80f1-dc9b618e37b3)

Ttfgggf:
Also I am facing similar issue to this raised issue (https://github.com/Prowlarr/Prowlarr/issues/2157) in regards to anything IPv6.
I have switched to a Global Unicast Addresses, but I seem to have DNS issue after issue in regards to the *arr stack.
I have manual mappings for my DNS on router, which I though would solve this but they haven't.

Ttfgggf:
> According to [RFC2732](https://www.ietf.org/rfc/rfc2732.txt), you should use square brackets for the host like`[fd3a:9a29:b136:eeee:aaaa::6]`.

I have tried that already. Sonarr doesn't seem to like it.

Ttfgggf:
> What doesn't seems to like it? Share errors, details etc.

I have shared a log above already. But let me share it again. (Please be aware I have omitted part of my IPv6 address from the logs provided, but you should be able to see the errors.

Here is a privatebin link (Although I have omitted the address and changed over to Global Unicast Address, it still the same problem)

https://privatebin.net/?f64bc135b94ce504#GJfXaqMTKnTTMTp1WYhyhF2DNTjy8aKN7Uq7ctEJXiMc

[...]

PR #8153, opened 2025-10-30
Use Happy Eyeballs for HTTP socket address selection

tie:
#### Description

This change adds Happy Eyeballs implementation for HTTP, so the connection would be established concurrently to interleaved IP addresses (IPv6 → IPv4 → IPv6 → IPv4 → …). This should also improve connectivity to IPv4 hosts where some of the addresses are unreachable (previous implementation attempted connection only to a single address).

Fixes IPv6 getting permanently disabled if an error happens with a single connection. Based on
https://slugcat.systems/post/24-06-16-ipv6-is-hard-happy-eyeballs-dotnet-httpclient/ but with tests and factored out generic Happy Eyeballs algorithm.

- Towards #7534

### DIFF
--- a/src/NzbDrone.Common/Http/HappyEyeballs/HttpHappyEyeballs.cs
+++ b/src/NzbDrone.Common/Http/HappyEyeballs/HttpHappyEyeballs.cs
@@ -38,7 +38,7 @@ public class HttpHappyEyeballs
 
         var happyEyeballs = CreateHappyEyeballs(endPoint);
         var socket = await happyEyeballs.Connect(resolvedAddresses, cancellationToken).ConfigureAwait(false);
-        _logger.Trace("Successfully connected {0} to address: {1}", endPoint.HostPort(), socket.RemoteEndPoint);
+        _logger.Trace("Successfully connected {0} to address: {1}", endPoint.HostPort, socket.RemoteEndPoint);
 
         return new NetworkStream(socket, ownsSocket: true);
     }
@@ -54,7 +54,7 @@ public class HttpHappyEyeballs
     {
         cancellationToken.ThrowIfCancellationRequested();
         var timeSpan = TimeSpan.FromMilliseconds(ConnectionAttemptDelay);
-        _logger.Trace("Waiting on {0} connection attempt delay for {1}", endPoint.HostPort(), timeSpan);
+        _logger.Trace("Waiting on {0} connection attempt delay for {1}", endPoint.HostPort, timeSpan);
 
         await Task.Delay(timeSpan, cancellationToken).ConfigureAwait(false);
     }
@@ -75,7 +75,7 @@ public class HttpHappyEyeballs
             NoDelay = true
         };
 
-        _logger.Trace("Trying Happy Eyeballs connection to {0} for host {1}", ipAddress, endPoint.HostPort());
+        _logger.Trace("Trying Happy Eyeballs connection to {0} for host {1}", ipAddress, endPoint.HostPort);
 
         try
         {
@@ -84,7 +84,7 @@ public class HttpHappyEyeballs
         catch (Exception e)
         {
             socket.Dispose();
-            _logger.Trace(e, "Happy Eyeballs connection to {0} for host {1} failed", ipAddress, endPoint.HostPort());
+            _logger.Trace(e, "Happy Eyeballs connection to {0} for host {1} failed", ipAddress, endPoint.HostPort);
             throw;
         }
 
--- a/src/NzbDrone.Common/Extensions/DnsEndPointExtensions.cs
+++ b/src/NzbDrone.Common/Extensions/DnsEndPointExtensions.cs
@@ -4,6 +4,9 @@ namespace NzbDrone.Common.Extensions
 {
     public static class DnsEndPointExtensions
     {
-        public static string HostPort(this DnsEndPoint endPoint) => $"{endPoint.Host}:{endPoint.Port}";
+        extension(DnsEndPoint endPoint)
+        {
+            public string HostPort => $"{endPoint.Host}:{endPoint.Port}";
+        }
     }
 }


tie:
Note that [Whisparr](https://github.com/Whisparr/Whisparr) uses .NET 6 but this change requires .NET 8.

src/NzbDrone.Common/Http/HappyEyeballs/HappyEyeballs.cs
```diff
@@ -0,0 +1,182 @@
+/*
+Until .NET implements Happy Eyeballs natively, use third-party implementation from
+https://slugcat.systems/post/24-06-16-ipv6-is-hard-happy-eyeballs-dotnet-httpclient/#the-implementation
+This issue is being tracked at https://github.com/dotnet/runtime/issues/26177.
+
+Below is a slightly modified Happy Eyeballs implementation from the post above.
+We’ve factored out HTTP-specific implementation into HttpHappyEyeballs class to
+make testing easier.
+*/
+
```
markus101: For consistency it'd be better to move this into the namespace, still prominent, but keeps imports at the top.

src/NzbDrone.Common/Http/HappyEyeballs/HttpHappyEyeballs.cs
```diff
@@ -0,0 +1,89 @@
...
+        if (resolvedAddresses.Length == 0)
+        {
+            throw new WebException(
+                $"The remote name {endPoint.Host} could not be resolved",
+                WebExceptionStatus.NameResolutionFailure);
+        }
+
+        var happyEyeballs = CreateHappyEyeballs(endPoint);
+        var socket = await happyEyeballs.Connect(resolvedAddresses, cancellationToken).ConfigureAwait(false);
+        _logger.Debug("Successfully connected {DnsEndPoint} to address: {RemoteEndPoint}", endPoint, socket.RemoteEndPoint);
```
markus101: All these debug messages are _incredibly_ chatty, even at Trace I think they're too much.

Also while the tokens are replaced correctly, I think it makes more sense to stick with either numeric tokens or using ones that match the variables, the former being what the majority of the code base does right now.

src/NzbDrone.Common/Http/HappyEyeballs/HttpHappyEyeballs.cs
```diff
@@ -0,0 +1,89 @@
...
+        // The following socket constructor will create a dual-mode socket on
+        // systems where IPv6 is available.
+        var socket = new Socket(SocketType.Stream, ProtocolType.Tcp)
+        {
+            // Turn off Nagle's algorithm since it degrades performance in most
+            // HttpClient scenarios.
+            NoDelay = true
+        };
+
+        _logger.Debug("Trying Happy Eyeballs connection to {IPEndPoint} for host {DnsEndPoint}", ipAddress, endPoint);
```
markus101: `DnsEndPoint.ToString()` ends up being `Unspecified/hostname` for everything that gets logged, instead I think it'd make sense to be more specific and log `endPoint.Host` (if the logging is kept at all).

src/NzbDrone.Common.Test/Http/HappyEyeballsFixture.cs
```diff
@@ -0,0 +1,222 @@
...
+            _connectSocketMock.InSequence(sequence)
+                .Setup(x => x(ipv6Address1, It.IsAny<CancellationToken>()))
+                .ReturnsAsync(Mock.Of<IDisposable>(MockBehavior.Strict));
+            _taskDelayMock.InSequence(sequence)
+                .Setup(x => x(It.IsAny<CancellationToken>()))
+                .Returns(TaskFromCancellationToken);
+
+            var result = await _happyEyeballs.Connect(addresses, cancellationToken);
+
+            Assert.NotNull(result);
```
markus101: Use `FluentAssertions` instead:

```suggestion
            result.Should().NotBeNull();
```

src/NzbDrone.Common.Test/Http/HappyEyeballsFixture.cs
```diff
@@ -0,0 +1,222 @@
...
+                .Setup(x => x(It.IsAny<CancellationToken>()))
+                .Returns(TaskFromCancellationToken);
+            _connectSocketMock.InSequence(sequence)
+                .Setup(x => x(ipv4Address2, It.IsAny<CancellationToken>()))
+                .ThrowsAsync(new Exception());
+
+            var ex = Assert.ThrowsAsync<AggregateException>(async () =>
+                await _happyEyeballs.Connect(addresses, cancellationToken));
+
+            Assert.That(ex.InnerExceptions, Has.Count.EqualTo(4));
```
markus101: Same for this and the rest of the test cases.

markus101 (changes requested):
The approach looks good, but the logging looks to be too chatty, even for trace I think.

src/NzbDrone.Common.Test/Http/HappyEyeballsFixture.cs
tie: Done.

src/NzbDrone.Common/Http/HappyEyeballs/HttpHappyEyeballs.cs
tie: For reference, [DnsEndPoint.ToString implementation](https://github.com/dotnet/runtime/blob/9d5a6a9aa463d6d10b0b0ba6d5982cc82f363dc3/src/libraries/System.Net.Primitives/src/System/Net/DnsEndPoint.cs#L48).

I’ve added `HostPort()` extension method that returns a string without address family.

src/NzbDrone.Common/Http/HappyEyeballs/HttpHappyEyeballs.cs
tie: Changed to numeric tokens, lowered log level to trace.

src/NzbDrone.Common/Http/HappyEyeballs/HappyEyeballs.cs
tie: Done.

src/NzbDrone.Common/Http/HappyEyeballs/HappyEyeballs.cs
```diff
@@ -0,0 +1,182 @@
...
+                using var delayCts = CancellationTokenSource.CreateLinkedTokenSource(successCts.Token);
+
+                // If we have another one to queue, wait for a timeout instead of *just* waiting for a connection task.
+                var timeoutTask = _taskDelay(delayCts.Token);
+                var whenAnyOrTimeout = await Task.WhenAny(whenAnyDone, timeoutTask).ConfigureAwait(false);
+                if (whenAnyOrTimeout != whenAnyDone)
+                {
+                    // Timeout finished. Go to next iteration so we queue another one.
+                    continue;
+                }
```
markus101, line 128: new line before `if` and drop the unnecessary comment, unless there is value to expanding the comment to why we're continuing.

```suggestion

                if (whenAnyOrTimeout != whenAnyDone)
                {
                    continue;
                }
```

src/NzbDrone.Common/Http/HappyEyeballs/HappyEyeballs.cs
```diff
@@ -0,0 +1,182 @@
...
+            else
+            {
+                completedTask = await whenAnyDone.ConfigureAwait(false);
+            }
+
+            tasks.Remove(completedTask);
+
+            if (completedTask.IsCompletedSuccessfully)
+            {
+                // We did it. We have success.
```
markus101: Comment doesn't really add value.

src/NzbDrone.Common/Http/HappyEyeballs/HappyEyeballs.cs
```diff
@@ -0,0 +1,182 @@
...
+        {
+            if (task.IsCompletedSuccessfully)
+            {
+                task.Result.Dispose();
+            }
+        }
+
+        if (successTask == null)
+        {
+            // We didn't get a single successful connection. Well heck.
```
markus101: Same here, comment doesn't add to the code.

src/NzbDrone.Common/Http/HappyEyeballs/HttpHappyEyeballs.cs
```diff
@@ -0,0 +1,89 @@
...
+        var endPoint = context.DnsEndPoint;
+
+        var ipHostEntry = await Dns.GetHostEntryAsync(endPoint.Host, endPoint.AddressFamily, cancellationToken).ConfigureAwait(false);
+        var resolvedAddresses = ipHostEntry.AddressList;
+        if (resolvedAddresses.Length == 0)
+        {
+            throw new WebException(
+                $"The remote name {endPoint.Host} could not be resolved",
+                WebExceptionStatus.NameResolutionFailure);
+        }
```
markus101, line 37: Group the variables and separate the `if` block.

```suggestion
        var endPoint = context.DnsEndPoint;
        var ipHostEntry = await Dns.GetHostEntryAsync(endPoint.Host, endPoint.AddressFamily, cancellationToken).ConfigureAwait(false);
        var resolvedAddresses = ipHostEntry.AddressList;
        
        if (resolvedAddresses.Length == 0)
        {
            throw new WebException(
                $"The remote name {endPoint.Host} could not be resolved",
                WebExceptionStatus.NameResolutionFailure);
        }
```

src/NzbDrone.Common/Http/HappyEyeballs/HttpHappyEyeballs.cs
```diff
@@ -0,0 +1,89 @@
...
+
+        var happyEyeballs = CreateHappyEyeballs(endPoint);
+        var socket = await happyEyeballs.Connect(resolvedAddresses, cancellationToken).ConfigureAwait(false);
+        _logger.Trace("Successfully connected {0} to address: {1}", endPoint.HostPort(), socket.RemoteEndPoint);
+        return new NetworkStream(socket, ownsSocket: true);
```
markus101: Separate the log and return

```suggestion
        _logger.Trace("Successfully connected {0} to address: {1}", endPoint.HostPort(), socket.RemoteEndPoint);

        return new NetworkStream(socket, ownsSocket: true);
```

src/NzbDrone.Common/Http/HappyEyeballs/HttpHappyEyeballs.cs
```diff
@@ -0,0 +1,89 @@
...
+        // systems where IPv6 is available.
+        var socket = new Socket(SocketType.Stream, ProtocolType.Tcp)
+        {
+            // Turn off Nagle's algorithm since it degrades performance in most
+            // HttpClient scenarios.
+            NoDelay = true
+        };
+
+        _logger.Trace("Trying Happy Eyeballs connection to {0} for host {1}", ipAddress, endPoint.HostPort());
+        try
```
markus101, line 80: New line before try/catch block

src/NzbDrone.Common/Http/HappyEyeballs/HttpHappyEyeballs.cs
```diff
@@ -0,0 +1,89 @@
...
+            (ipAddress, cancel) => ConnectSocket(ipAddress, endPoint, cancel),
+            cancel => TaskDelay(endPoint, cancel));
+    }
+
+    private async Task TaskDelay(DnsEndPoint endPoint, CancellationToken cancellationToken)
+    {
+        cancellationToken.ThrowIfCancellationRequested();
+        var timeSpan = TimeSpan.FromMilliseconds(ConnectionAttemptDelay);
+        _logger.Trace("Waiting on {0} connection attempt delay for {1}", endPoint.HostPort(), timeSpan);
+        await Task.Delay(timeSpan, cancellationToken).ConfigureAwait(false);
```
markus101: Similar here to separate the await
```suggestion
        _logger.Trace("Waiting on {0} connection attempt delay for {1}", endPoint.HostPort(), timeSpan);

        await Task.Delay(timeSpan, cancellationToken).ConfigureAwait(false);
```

src/NzbDrone.Common/Http/HappyEyeballs/HappyEyeballs.cs
```diff
@@ -0,0 +1,182 @@
...
+            cancellationToken.ThrowIfCancellationRequested();
+            Debug.Assert(innerExceptions.Count > 0);
```
markus101: Do we need the assert?

src/NzbDrone.Common/Http/HappyEyeballs/HappyEyeballs.cs
```diff
@@ -0,0 +1,182 @@
...
+            throw new AggregateException(innerExceptions);
```
markus101, line 176: Separate the throw
```suggestion

            throw new AggregateException(innerExceptions);
```

src/NzbDrone.Common/Http/HappyEyeballs/HappyEyeballs.cs
```diff
@@ -0,0 +1,182 @@
...
+            {
+                break;
+            }
+
+            if (taskIndex < totalTasks)
+            {
+                // We have to queue another task this iteration.
+                var newTask = taskBuilder(taskIndex, successCts.Token);
+                tasks.Add(newTask);
+                taskIndex++;
```
markus101: Drop the comment and add the task directly instead of creating the variable first.

```suggestion
                tasks.Add(taskBuilder(taskIndex, successCts.Token);
                taskIndex++;
```

src/NzbDrone.Common/Http/HappyEyeballs/HappyEyeballs.cs
```diff
@@ -0,0 +1,182 @@
...
+        var innerExceptions = new List<Exception>();
+
+        // The general loop here is as follows:
+        // 1. Add a new task for the next IP to try.
+        // 2. Wait until any task completes OR the delay happens.
+        // If an error occurs, we stop checking that task and continue checking the next.
+        // Every iteration we add another task, until we're full on them.
+        // We keep looping until we have SUCCESS, or we run out of attempt tasks entirely.
+        Task<TSocket> successTask = null;
+        while (taskIndex < totalTasks || tasks.Count > 0)
```
markus101, line 101: Separate the long comment from the variable the the while loop block.
```suggestion
        // The general loop here is as follows:
        // 1. Add a new task for the next IP to try.
        // 2. Wait until any task completes OR the delay happens.
        // If an error occurs, we stop checking that task and continue checking the next.
        // Every iteration we add another task, until we're full on them.
        // We keep looping until we have SUCCESS, or we run out of attempt tasks entirely.

        Task<TSocket> successTask = null;

        while (taskIndex < totalTasks || tasks.Count > 0)
```

src/NzbDrone.Common/Http/HappyEyeballs/HappyEyeballs.cs
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    public async ValueTask<TSocket> Connect(
+        IPAddress[] addresses,
+        CancellationToken cancellationToken)
+    {
+        ArgumentOutOfRangeException.ThrowIfNegativeOrZero(addresses.Length);
+
+        var ips = SortInterleaved(addresses);
+        return await ParallelTask(
```
markus101, line 43: new line before return

```suggestion

        return await ParallelTask(
```

src/NzbDrone.Common/Http/HappyEyeballs/HappyEyeballs.cs
```diff
@@ -0,0 +1,182 @@
...
+        // Interleave returned addresses so that they are IPv6 -> IPv4 -> IPv6 -> IPv4.
+        // Assuming we have multiple addresses of the same type that is.
+        // As described in the RFC.
+        var ipv6 = addresses.Where(x => x.AddressFamily == AddressFamily.InterNetworkV6).ToArray();
+        var ipv4 = addresses.Where(x => x.AddressFamily == AddressFamily.InterNetwork).ToArray();
+
+        var commonLength = Math.Min(ipv6.Length, ipv4.Length);
+
+        var result = new IPAddress[addresses.Length];
+        for (var i = 0; i < commonLength; i++)
```
markus101, line 61: new line before loop
```suggestion

        for (var i = 0; i < commonLength; i++)
```

src/NzbDrone.Common/Http/HappyEyeballs/HappyEyeballs.cs
tie: Hm, probably not, this seems to be already asserted in [`Task.Exception` property](https://learn.microsoft.com/en-us/dotnet/api/system.threading.tasks.task.exception?view=net-10.0).
https://github.com/dotnet/dotnet/blob/c93d94a7e854e38eb560698df4b1b3313e4efc13/src/runtime/src/libraries/System.Private.CoreLib/src/System/Threading/Tasks/TaskExceptionHolder.cs#L256

FWIW, this condition is always true because
1. we throw if we stopped because of cancellation
2. otherwise, we always have a non-empty tasks list

So at this point we know that at least one of the tasks is faulted (but not canceled), so `innerExceptions` is not empty.

mynameisbogdan:
Is this still needed on .NET 10 after so many years? 🤔

tie:
@mynameisbogdan, yes, seems so, unfortunately. Upstream issue: https://github.com/dotnet/runtime/issues/26177

mynameisbogdan:
@tie might be a little too verbose

```
[Trace] ConfigService: Using default config value for 'proxyenabled' defaultValue:'False' 
[Trace] ManagedHttpDispatcher: Trying Happy Eyeballs connection to 2606:4700:20::ac43:4662 for host services.sonarr.tv:443 
[Trace] ManagedHttpDispatcher: Happy Eyeballs connection to 2606:4700:20::ac43:4662 for host services.sonarr.tv:443 failed 

[v10.0.0.19513] System.Net.Sockets.SocketException (65): No route to host
   at System.Net.Sockets.Socket.AwaitableSocketAsyncEventArgs.CreateException(SocketError error, Boolean forAsyncThrow)
   at System.Net.Sockets.Socket.AwaitableSocketAsyncEventArgs.ConnectAsync(Socket socket, CancellationToken cancellationToken)
   at System.Net.Sockets.Socket.ConnectAsync(EndPoint remoteEP, CancellationToken cancellationToken)
   at NzbDrone.Common.Http.HappyEyeballs.HttpHappyEyeballs.ConnectSocket(IPAddress ipAddress, DnsEndPoint endPoint, CancellationToken cancellationToken) in ./NzbDrone.Common/Http/HappyEyeballs/HttpHappyEyeballs.cs:line 82
   at System.Runtime.CompilerServices.AsyncMethodBuilderCore.Start[TStateMachine](TStateMachine& stateMachine)
   at NzbDrone.Common.Http.HappyEyeballs.HttpHappyEyeballs.ConnectSocket(IPAddress ipAddress, DnsEndPoint endPoint, CancellationToken cancellationToken)
   at NzbDrone.Common.Http.HappyEyeballs.HttpHappyEyeballs.<>c__DisplayClass4_0.<CreateHappyEyeballs>b__0(IPAddress ipAddress, CancellationToken cancel) in ./NzbDrone.Common/Http/HappyEyeballs/HttpHappyEyeballs.cs:line 49
   at NzbDrone.Common.Http.HappyEyeballs.HappyEyeballs`1.<>c__DisplayClass3_0.<Connect>b__0(Int32 i, CancellationToken cancel) in ./NzbDrone.Common/Http/HappyEyeballs/HappyEyeballs.cs:line 45
   at NzbDrone.Common.Http.HappyEyeballs.HappyEyeballs`1.ParallelTask(Int32 totalTasks, Func`3 taskBuilder, CancellationToken cancellationToken) in ./NzbDrone.Common/Http/HappyEyeballs/HappyEyeballs.cs:line 110
   at System.Runtime.CompilerServices.AsyncMethodBuilderCore.Start[TStateMachine](TStateMachine& stateMachine)
   at NzbDrone.Common.Http.HappyEyeballs.HappyEyeballs`1.ParallelTask(Int32 totalTasks, Func`3 taskBuilder, CancellationToken cancellationToken)
   at NzbDrone.Common.Http.HappyEyeballs.HappyEyeballs`1.Connect(IPAddress[] addresses, CancellationToken cancellationToken) in ./NzbDrone.Common/Http/HappyEyeballs/HappyEyeballs.cs:line 43
   at System.Runtime.CompilerServices.AsyncMethodBuilderCore.Start[TStateMachine](TStateMachine& stateMachine)
   at NzbDrone.Common.Http.HappyEyeballs.HappyEyeballs`1.Connect(IPAddress[] addresses, CancellationToken cancellationToken)
   at NzbDrone.Common.Http.HappyEyeballs.HttpHappyEyeballs.OnConnect(SocketsHttpConnectionContext context, CancellationToken cancellationToken) in ./NzbDrone.Common/Http/HappyEyeballs/HttpHappyEyeballs.cs:line 40
   at System.Runtime.CompilerServices.AsyncTaskMethodBuilder`1.AsyncStateMachineBox`1.ExecutionContextCallback(Object s)
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state)
   at System.Runtime.CompilerServices.AsyncTaskMethodBuilder`1.AsyncStateMachineBox`1.MoveNext(Thread threadPoolThread)
   at System.Threading.Tasks.AwaitTaskContinuation.RunOrScheduleAction(IAsyncStateMachineBox box, Boolean allowInlining)
   at System.Threading.Tasks.Task.RunContinuations(Object continuationObject)
   at System.Threading.Tasks.Task.FinishSlow(Boolean userDelegateExecute)
   at System.Threading.Tasks.Task.ExecuteWithThreadLocal(Task& currentTaskSlot, Thread threadPoolThread)
   at System.Threading.ThreadPoolWorkQueue.Dispatch()
   at System.Threading.PortableThreadPool.WorkerThread.WorkerThreadStart()
   at System.Threading.Thread.StartCallback()
--- End of stack trace from previous location ---
   at NzbDrone.Common.Http.HappyEyeballs.HttpHappyEyeballs.ConnectSocket(IPAddress ipAddress, DnsEndPoint endPoint, CancellationToken cancellationToken) in ./NzbDrone.Common/Http/HappyEyeballs/HttpHappyEyeballs.cs:line 82

[Trace] ManagedHttpDispatcher: Waiting on services.sonarr.tv:443 connection attempt delay for 00:00:00.2500000 
[Trace] ManagedHttpDispatcher: Trying Happy Eyeballs connection to 104.26.1.163 for host services.sonarr.tv:443 
[Trace] ManagedHttpDispatcher: Waiting on services.sonarr.tv:443 connection attempt delay for 00:00:00.2500000 
[Trace] ManagedHttpDispatcher: Successfully connected services.sonarr.tv:443 to address: [::ffff:104.26.1.163]:443
```

Issue #8096, opened 2025-09-29
Add support for additional Postgres options


solidDoWant:
### Is there an existing issue for this?

- [x] I have searched the existing open and closed issues

### Is your feature request related to a problem? Please describe

See https://github.com/Sonarr/Sonarr/issues/7943 - I'd comment on there but the issue has been locked.

I filed [a small PR against Radarr](https://github.com/Radarr/Radarr/pull/11255) to add support for all postgres options before being informed that Sonarr is upstream of Radarr. Would a PR for this against Sonarr be accepted? The other issue did not specify whether the feature would not be implemented because existing maintainers had no interest in putting the work in for it (but would accept a PR for it), or because maintainers do not want this feature at all.

### Describe the solution you'd like

Additional postgres options implemented, see https://github.com/Radarr/Radarr/pull/11255

### Describe alternatives you've considered

Maintain a fork with these changes I guess... not really something I want to do

markus101:
If it's properly tested and addresses the issues already mentioned in that PR and anything else then I don't see a reason not to accepted.

Closing as we're not going to track this.

Issue #7943, opened 2025-06-28
support PostgreSQL connection options


gclawes:
### Is there an existing issue for this?

- [x] I have searched the existing open and closed issues

### Is your feature request related to a problem? Please describe

Currently there connection string used for PostgreSQL supports only database/host/username/password/port fields.  This limits the ability to use other PostgreSQL features like SSL and client certificate authentication, or pass performance tuning parameters.

### Describe the solution you'd like

Support for arbitrary PostgreSQL connection options would enable users to use advanced PostgreSQL settings (at their own risk) without requiring additional configuration options to be encoded in the configuration schema (or official support for these modes beyond the ability to pass the options).

### Describe alternatives you've considered

Utilizing an external proxy to wrap the psql connection with the additional features.  This may work for SSL and client cert authentication, but has performance implications and would not necessarily help with other use cases like performance tuning options in Sonarr's DB connection.

markus101:
At this time we do not have plans to expand support for additional Postgres options.

PR #8177, opened 2025-11-14
New: Extend Postgres support to enable all connection string options

solidDoWant:
#### Description

This PR enables support for all Postgres connection string options supported by the underlying Npgsql library. It does this by allowing users to specify a `PostgresConnectionString` parameter. If this parameter is provided, it is passed to the Npgsql connection string builder directly. Backwards compatibility is maintained by allowing existing config values to override any values in the connection string.

This will allow users to (if they explicitly choose to):
* Require connections to use TLS
* Configure the trusted CA for the server's serving cert
* Specify a client certificate to use for mTLS (this is what I personally want this for)
* Configure database pooling

I've tested this patch both against Radarr, and against Sonarr (v5).

I initially opened [a PR against Radarr](https://github.com/Radarr/Radarr/pull/11255) before realizing the Sonarr is the upstream. This went through a few rounds of review and I think it is ready for merge against Radarr, should this get merged into Sonarr.

#### Database Migration
NO

* Closes https://github.com/Sonarr/Sonarr/issues/8096, https://github.com/Sonarr/Sonarr/issues/7943

### DIFF
--- a/src/NzbDrone.Core/Datastore/ConnectionStringFactory.cs
+++ b/src/NzbDrone.Core/Datastore/ConnectionStringFactory.cs
@@ -24,6 +24,7 @@ namespace NzbDrone.Core.Datastore
             _configFileProvider = configFileProvider;
 
             var connectionStringType = GetConnectionStringType();
+
             switch (connectionStringType)
             {
                 case ConnectionStringType.PostgreSqlVars:


src/NzbDrone.Core/Datastore/ConnectionStringFactory.cs
```diff
@@ -74,5 +90,61 @@ private DatabaseConnectionInfo GetPostgresConnectionString(string dbName)
...
+            };
+
+            return new DatabaseConnectionInfo(DatabaseType.PostgreSQL, connectionBuilder.ConnectionString);
+        }
+
+        /// <summary>
+        /// Validates that either Postgres connection strings are both set or neither are set, and that either connection strings or
+        /// other Postgres settings are set, but not both.
+        /// </summary>
+        /// <exception cref="ArgumentException">Thrown when configuration is invalid.</exception>
```
markus101: Please remove the XML documentation.

src/NzbDrone.Core/Datastore/ConnectionStringFactory.cs
```diff
@@ -74,5 +90,61 @@ private DatabaseConnectionInfo GetPostgresConnectionString(string dbName)
...
+                }
+
+                if (isLogDBConnectionStringSet && !isMainDBConnectionStringSet)
+                {
+                    throw new ArgumentException("Postgres LogDbConnectionString is set but MainDbConnectionString is not. Both must be set or neither.");
+                }
+            }
+
+            // At this point either all required connection strings are set or neither, so only one needs to be checked
+            var areConnectionStringConfigsSet = isMainDBConnectionStringSet;
```
markus101: Just use `isMainDBConnectionStringSet` instead of assigning to another variable.

src/NzbDrone.Core/Datastore/ConnectionStringFactory.cs
```diff
@@ -74,5 +90,61 @@ private DatabaseConnectionInfo GetPostgresConnectionString(string dbName)
...
+
+            // This one _must_ be set if connection strings are not being used, so it is used as a test to see if the user attempted configuration via individual settings
+            var areOtherPostgresConfigsSet = _configFileProvider.PostgresHost.IsNotNullOrWhiteSpace();
+
+            if (areConnectionStringConfigsSet && areOtherPostgresConfigsSet)
```
markus101: Similar here, can just check `isMainDBConnectionStringSet && _configFileProvider.PostgresHost.IsNotNullOrWhiteSpace()`

src/NzbDrone.Core/Datastore/ConnectionStringFactory.cs
```diff
@@ -60,9 +76,9 @@ private static DatabaseConnectionInfo GetConnectionString(string dbPath)
             return new DatabaseConnectionInfo(DatabaseType.SQLite, connectionBuilder.ConnectionString);
         }
 
-        private DatabaseConnectionInfo GetPostgresConnectionString(string dbName)
+        private DatabaseConnectionInfo GetPostgresConnectionInfoFromIndividualValues(string dbName)
```
markus101: I think `BuildPostgresConnectionString` would be clear enough without being too verbose.

src/NzbDrone.Core/Datastore/ConnectionStringFactory.cs
```diff
@@ -74,5 +90,61 @@ private DatabaseConnectionInfo GetPostgresConnectionString(string dbName)
...
+
+            return new DatabaseConnectionInfo(DatabaseType.PostgreSQL, connectionBuilder.ConnectionString);
+        }
+
+        /// <summary>
+        /// Validates that either Postgres connection strings are both set or neither are set, and that either connection strings or
+        /// other Postgres settings are set, but not both.
+        /// </summary>
+        /// <exception cref="ArgumentException">Thrown when configuration is invalid.</exception>
+        private void ValidatePostgresOptions()
```
markus101: I think instead of setting class variables this should be renamed to something like `GetDatabaseType` and return a enum with 3 possible values, Postgres, PostgresConnectionString or Sqlite, that'd eliminate the private variables that are set by a "validate" function.

src/NzbDrone.Core/Datastore/ConnectionStringFactory.cs
```diff
@@ -74,5 +90,61 @@ private DatabaseConnectionInfo GetPostgresConnectionString(string dbName)
...
+        {
+            var isMainDBConnectionStringSet = !string.IsNullOrWhiteSpace(_configFileProvider.PostgresMainDbConnectionString);
+            var isLogDBConnectionStringSet = !string.IsNullOrWhiteSpace(_configFileProvider.PostgresLogDbConnectionString);
+            var isHostSet = !string.IsNullOrWhiteSpace(_configFileProvider.PostgresHost);
```
markus101: Use the extension for not null/whitespace.

```suggestion
            var isMainDBConnectionStringSet = _configFileProvider.PostgresMainDbConnectionString.IsNotNullOrWhiteSpace(;
            var isLogDBConnectionStringSet = _configFileProvider.PostgresLogDbConnectionString.IsNotNullOrWhiteSpace(;
            var isHostSet = _configFileProvider.PostgresHost.IsNotNullOrWhiteSpace();
```

markus101 (changes requested):
Looks pretty good, but some suggestions for clean up/consistency.

src/NzbDrone.Core/Datastore/ConnectionStringFactory.cs
```diff
@@ -62,7 +74,7 @@ private static DatabaseConnectionInfo GetConnectionString(string dbPath)
 
         private DatabaseConnectionInfo GetPostgresConnectionString(string dbName)
         {
-            var connectionBuilder = new NpgsqlConnectionStringBuilder
+            var connectionBuilder = new NpgsqlConnectionStringBuilder()
```
markus101:
```suggestion
            var connectionBuilder = new NpgsqlConnectionStringBuilder
```

src/NzbDrone.Core/Datastore/ConnectionStringFactory.cs
```diff
@@ -74,5 +86,55 @@ private DatabaseConnectionInfo GetPostgresConnectionString(string dbName)
...
+
+            return new DatabaseConnectionInfo(DatabaseType.PostgreSQL, connectionBuilder.ConnectionString);
+        }
+
+        private enum ConnectionStringType
+        {
+            SQLite,
+            PostgreSQLVars,
+            PostgreSQLConnectionString
+        }
```
markus101: Move to the bottom of the class and update names to match of naming standard.

```suggestion
        private enum ConnectionStringType
        {
            Sqlite,
            PostgreSqlVars,
            PostgreSqlConnectionString
        }
```

src/NzbDrone.Core/Datastore/ConnectionStringFactory.cs
```diff
@@ -22,11 +22,23 @@ public ConnectionStringFactory(IAppFolderInfo appFolderInfo, IConfigFileProvider
...
+                case ConnectionStringType.PostgreSQLConnectionString:
+                    MainDbConnection = GetPostgresConnectionInfoFromConnectionString(_configFileProvider.PostgresMainDbConnectionString);
+                    LogDbConnection = GetPostgresConnectionInfoFromConnectionString(_configFileProvider.PostgresLogDbConnectionString);
+                    break;
+                case ConnectionStringType.SQLite:
+                    MainDbConnection = GetConnectionString(appFolderInfo.GetDatabase());
+                    LogDbConnection = GetConnectionString(appFolderInfo.GetLogDatabase());
+                    break;
+                default:
+                    throw new ArgumentOutOfRangeException();
```
markus101: This should never hit default, unless someone is added a new DB type, potentially SQLite just is the default or it should throw a `SonarrStartupException` indicating what the problem is.

src/NzbDrone.Core/Datastore/ConnectionStringFactory.cs
```diff
@@ -74,5 +86,55 @@ private DatabaseConnectionInfo GetPostgresConnectionString(string dbName)
...
+            {
+                // No Postgres settings are set, so nothing to validate
+                return ConnectionStringType.SQLite;
+            }
+
+            if (_configFileProvider.LogDbEnabled)
+            {
+                if (!isMainDBConnectionStringSet && isLogDBConnectionStringSet)
+                {
+                    throw new ArgumentException("Postgres MainDbConnectionString is set but LogDbConnectionString is not. Both must be set or neither.");
```
markus101: These should either be `SonarrStartupException` or an exception that implements from it, similar to how `CorruptDatabaseException` is done since it affects start up and will get handled properly by the app.

markus101 (commented):
Thanks for the fixes so far, just a handful more and it'll be good to land.

solidDoWant:
@markus101 let me know if this looks good to you, and if so I'll cut another local/testing build from my fork and verify that I didn't break anything prior to merge.

BTW, so far this has been working on both v4 and v5 with the same exact patch. Any chance that this would be accepted as a v4 change as well? This is a new feature, but a relatively small one.

markus101 (approved):
Thanks for the fixes, will land after CI is happy.

We're not adding new features to v4, only fixes and security issues will be backported.

src/NzbDrone.Core/Datastore/ConnectionStringFactory.cs
```diff
@@ -22,11 +23,24 @@ public ConnectionStringFactory(IAppFolderInfo appFolderInfo, IConfigFileProvider
         {
             _configFileProvider = configFileProvider;
 
-            MainDbConnection = _configFileProvider.PostgresHost.IsNotNullOrWhiteSpace() ? GetPostgresConnectionString(_configFileProvider.PostgresMainDb) :
-                GetConnectionString(appFolderInfo.GetDatabase());
-
-            LogDbConnection = _configFileProvider.PostgresHost.IsNotNullOrWhiteSpace() ? GetPostgresConnectionString(_configFileProvider.PostgresLogDb) :
-                GetConnectionString(appFolderInfo.GetLogDatabase());
+            var connectionStringType = GetConnectionStringType();
+            switch (connectionStringType)
```
markus101, line 28:
```suggestion
            var connectionStringType = GetConnectionStringType();

            switch (connectionStringType)
```

Issue #2477, opened 2018-03-26
Authelia / Authentik Login Support


rickatnight11:
## Feature Request

Support for centralized authentication (i.e. LDAP) and SSO (i.e. SAML/OATH) provides several benefits:

1. Multiple user accounts without password sharing
2. Don't store admin credentials in plaintext config
3. Use same credentials/sessions across apps in this space ( i.e Radarr, NZBGet, Tautulli, etc)
4. Opens the door for user permissions (authorization) down the road
5. Improve security by offloading AuthN/AuthZ to services designed for these purposes

rickatnight11:
I opened a similar request for Radarr: https://github.com/Radarr/Radarr/issues/2680

markus101:
1. https://github.com/Sonarr/Sonarr/issues/1682
2. They're not stored in plaintext

We do not have plans to support a different authentication method, at least at this time.

kragebein:
i guess its been a day

neonspectra:
+1 on this issue. Why is this closed, anyway? Especially now that Sonarr requires auth be turned on in wake of that one Defcon talk.

If you're not going to let me big a big girl and disable auth at my own risk, then can you at least let me hook it up to a SAML provider and call it a day so that I don't have an additional cred to manage? 😅

joshka:
@markus101 would you mind reopening this based on my comments in https://github.com/Sonarr/Sonarr/issues/1682#issuecomment-2046483923 and an intent to implement this?

joshka:
I'm about to start working on the multi-user parts of Sonarr in https://github.com/Sonarr/Sonarr/issues/1682 and wanted to add some clarity around the SSO parts that I intend to follow on with:

My current thinking on this (feel free to disagree with these)

0. SSO happens after the multi-user stuff in #1682
1. It makes sense to focus on OIDC rather than LDAP / SAML / OAuth for this. LDAP is too finicky for integrating directly with. SAML is more difficult to implement well compared to OIDC. OAuth is authorization, not authentication. There might be parts of this that you could bolt onto just OAuth, and there are parts of the interaction between the *arrs that might work better with OAuth, but it's not too helpful for the SSO part.
2. I intend to use Authelia as a test backend (though will transition to Authentik / Keycloak if necessary)
3. I think probably mapping an OIDC "group" claim to roles in Sonarr is likely the right way to go (wouldn't mind discussing this further if anyone has specific guidance)
[...]

PR #8178, opened 2025-11-14
Fix: incompatibility with reverse proxy forward auth providers

solidDoWant:
#### Description

Sonarr is currently partially incompatible with reverse proxy forward auth providers, like [Authentik](https://github.com/goauthentik/authentik), which need the client/front-end to submit cookies via CORS for credential renewal. Reverse proxies like Envoy Gateway and Istio respond to client requests with expired credentials with a redirect to the forward auth provider, which then handles credential renewal. Typically the redirected request to the forward auth provider will need to provide cookies (under the auth provider domain, but with an origin of the Sonarr domain). For forward auth providers that support this, renewal can be handled without any user interaction (like refreshing the page).

The problem is that for this to work properly, any requests made via XMLHTTPRequest needs to allow for sending the forward auth provider's cookie to the forward auth provider, when the XMLHTTPRequest is redirected to it. This requires setting `XMLHTTPRequest.withCredentials = true`.

When this field is set to `true`, CORS preflight responses must include the HTTP `Origin` domain name rather than the `*` that is currently set. This should only be set if the provided `Origin` header matches an application-configured value. They also need to respond with the `Access-Control-Allow-Credentials` header.

I've added support for all of this, and I've made the application-configured CORS supported origin value configurable. This has not been plumbed up to the UI, but can be set via env var. Please let me know what, if anything, should be presented in the UI and where.

Unless a value is explicitly set, this should have no impact on users.

#### Database Migration
NO

### DIFF
--- a/src/NzbDrone.Common/Options/ServerOptions.cs
+++ b/src/NzbDrone.Common/Options/ServerOptions.cs
@@ -4,7 +4,7 @@ public class ServerOptions
 {
     public string UrlBase { get; set; }
     public string BindAddress { get; set; }
-    public string AllowedCORSOrigins { get; set; }
+    public string AllowedCorsOrigins { get; set; }
     public int? Port { get; set; }
     public bool? EnableSsl { get; set; }
     public int? SslPort { get; set; }
--- a/src/NzbDrone.Host/CorsOriginConfigurator.cs
+++ b/src/NzbDrone.Host/CorsOriginConfigurator.cs
@@ -1,0 +1,60 @@
+using System;
+using System.Linq;
+using Microsoft.AspNetCore.Cors.Infrastructure;
+using Microsoft.Extensions.Options;
+using NzbDrone.Core.Configuration;
+using Sonarr.Http;
+
+namespace NzbDrone.Host
+{
+    public class CorsOriginConfigurator(IConfigFileProvider configFileProvider, NLog.Logger logger) : IPostConfigureOptions<CorsOptions>
+    {
+        private readonly IConfigFileProvider _configFileProvider = configFileProvider;
+        private readonly NLog.Logger _logger = logger;
+
+        public void PostConfigure(string name, CorsOptions options) => PostConfigure(options);
+
+        public void PostConfigure(CorsOptions options)
+        {
+            _logger.Info("Configuring CORS. Allowed origins: {0}", _configFileProvider.AllowedCorsOrigins);
+            options.GetPolicy(VersionedApiControllerAttribute.API_CORS_POLICY).IsOriginAllowed = CorsOriginCheck;
+            options.GetPolicy("AllowGet").IsOriginAllowed = CorsOriginCheck;
+        }
+
+        /// <summary>
+        /// Check if the request origin is allowed based on the configured allowed origins in the config file.
+        /// </summary>
+        /// <param name="requestOrigin">The origin to check against the list of allowed cross-origin domains.</param>
+        /// <returns>True if the request origin matches a configured origin (based upon the CORS spec), or if the allowed origins contains '*', false otherwise.</returns>
+        protected bool CorsOriginCheck(string requestOrigin)
+        {
+            var allowedOrigins = _configFileProvider.AllowedCorsOrigins.Split(';', StringSplitOptions.RemoveEmptyEntries | StringSplitOptions.TrimEntries);
+            return allowedOrigins.Any(
+                allowedOrigin => allowedOrigin.Equals(CorsConstants.AnyOrigin, StringComparison.Ordinal) ||
+                IsSameOrigin(allowedOrigin, requestOrigin));
+        }
+
+        /// <summary>
+        /// Check if two origins are the same based on scheme, host and port.
+        /// </summary>
+        /// <param name="configuredOrigin">The first origin to compare.</param>
+        /// <param name="requestOrigin">The second origin to compare.</param>
+        /// <returns>True if the scheme, host, and port match, false otherwise.</returns>
+        protected bool IsSameOrigin(string configuredOrigin, string requestOrigin)
+        {
+            if (!Uri.TryCreate(configuredOrigin, UriKind.Absolute, out var configuredOriginUri))
+            {
+                _logger.Warn("Invalid origin configured: {0}", configuredOrigin);
+                return false;
+            }
+
+            if (!Uri.TryCreate(requestOrigin, UriKind.Absolute, out var requestOriginUri))
+            {
+                _logger.Debug("Invalid request origin: {0}", requestOrigin);
+                return false;
+            }
+
+            return Uri.Compare(configuredOriginUri, requestOriginUri, UriComponents.SchemeAndServer, UriFormat.Unescaped, StringComparison.OrdinalIgnoreCase) == 0;
+        }
+    }
+}
--- a/src/NzbDrone.Host/Startup.cs
+++ b/src/NzbDrone.Host/Startup.cs
@@ -5,7 +5,6 @@ using System.Linq;
 using DryIoc;
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Builder;
-using Microsoft.AspNetCore.Cors.Infrastructure;
 using Microsoft.AspNetCore.DataProtection;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.HttpOverrides;
@@ -13,7 +12,6 @@ using Microsoft.AspNetCore.Mvc.Controllers;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
-using Microsoft.Extensions.Options;
 using Microsoft.OpenApi.Models;
 using NLog.Extensions.Logging;
 using NzbDrone.Common.EnvironmentInfo;
@@ -247,7 +245,7 @@ namespace NzbDrone.Host
 
             services.AddAppAuthentication();
 
-            services.ConfigureOptions<CORSOriginConfigurator>();
+            services.ConfigureOptions<CorsOriginConfigurator>();
         }
 
         public void Configure(IApplicationBuilder app,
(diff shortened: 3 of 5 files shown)


src/NzbDrone.Host/Startup.cs
```diff
@@ -359,4 +366,53 @@ private void EnsureSingleInstance(bool isService, IStartupContext startupContext
             }
         }
     }
+
+    public class CORSOriginConfigurator : IPostConfigureOptions<CorsOptions>
```
markus101: This should be in it's own file, not part of `Startup.cs`

frontend/src/Utilities/createAjaxRequest.js
```diff
@@ -27,6 +27,7 @@ function addContentType(ajaxOptions) {
 
 export default function createAjaxRequest(originalAjaxOptions) {
   const requestXHR = new window.XMLHttpRequest();
+  requestXHR.withCredentials = true; // Needed for CORS requests with cookies, which some reverse proxies with forward auth require
```
markus101, line 30: Everything already using v5 endpoints is using fetch, this will be removed before v5 ships. `fetchJson` is where that's happening currently.

src/NzbDrone.Common/Options/ServerOptions.cs
```diff
@@ -4,6 +4,7 @@ public class ServerOptions
 {
     public string UrlBase { get; set; }
     public string BindAddress { get; set; }
+    public string AllowedCORSOrigins { get; set; }
```
markus101: Rename to follow our naming standard.

```suggestion
    public string AllowedCorsOrigins { get; set; }
```

src/NzbDrone.Host/Startup.cs
```diff
@@ -359,4 +366,53 @@ private void EnsureSingleInstance(bool isService, IStartupContext startupContext
...
+    {
+        private readonly IConfigFileProvider _configFileProvider;
+        private readonly NLog.Logger _logger;
+
+        public CORSOriginConfigurator(IConfigFileProvider configFileProvider, NLog.Logger logger)
```
markus101: Same thing with the name, `CORS` -> `Cors`

src/NzbDrone.Host/Startup.cs
```diff
@@ -359,4 +366,53 @@ private void EnsureSingleInstance(bool isService, IStartupContext startupContext
...
+
+        protected bool CorsOriginCheck(string requestOrigin)
+        {
+            var allowedOrigin = _configFileProvider.AllowedCORSOrigins;
+            if (allowedOrigin.Equals(CorsConstants.AnyOrigin, StringComparison.Ordinal))
+            {
+                return true;
+            }
+
+            if (string.IsNullOrWhiteSpace(requestOrigin))
```
markus101: Use the extension instead.

```suggestion
            if (requestOrigin.IsNullOrWhiteSpace())
```

src/NzbDrone.Host/Startup.cs
```diff
@@ -359,4 +366,53 @@ private void EnsureSingleInstance(bool isService, IStartupContext startupContext
...
+            {
+                return false;
+            }
+
+            var allowedOriginAsUri = new Uri(allowedOrigin);
+            Uri requestOriginAsUri;
+            try
+            {
+                requestOriginAsUri = new Uri(requestOrigin);
```
markus101: `Uri.TryCreate` is probably a better choice here.

src/NzbDrone.Host/Startup.cs
```diff
@@ -359,4 +366,53 @@ private void EnsureSingleInstance(bool isService, IStartupContext startupContext
...
+        public void PostConfigure(CorsOptions options)
+        {
+            _logger.Info("Configuring CORS. Allowed origins: {0}", _configFileProvider.AllowedCORSOrigins);
+            options.GetPolicy(VersionedApiControllerAttribute.API_CORS_POLICY).IsOriginAllowed = CorsOriginCheck;
+            options.GetPolicy("AllowGet").IsOriginAllowed = CorsOriginCheck;
+        }
+
+        protected bool CorsOriginCheck(string requestOrigin)
+        {
+            var allowedOrigin = _configFileProvider.AllowedCORSOrigins;
```
markus101: Is this supposed to be origin or origins? The setting is plural, but the usage seems to imply it's singular.

solidDoWant: Plural. I forgot to commit something when I wrote this a few weeks ago. I'll add it back, sorry for the confusion.

frontend/src/Utilities/createAjaxRequest.js
solidDoWant: Thanks for the heads up. I've added this to `fetchJson`.

It looks like there are at least three other `fetch()` calls throughout the frontend that don't use `fetchJson`. Will those be centralized at some point, or do they need to be updated with the equivalent of the above as well?

solidDoWant:
@markus101 After you and any other reviewers are satisfied with this, I'll test again prior to merging. I especially need to retest with the `fetch` stuff specifically.

src/NzbDrone.Host/CorsOriginConfigurator.cs
```diff
@@ -0,0 +1,60 @@
...
+            _logger.Info("Configuring CORS. Allowed origins: {0}", _configFileProvider.AllowedCorsOrigins);
+            options.GetPolicy(VersionedApiControllerAttribute.API_CORS_POLICY).IsOriginAllowed = CorsOriginCheck;
+            options.GetPolicy("AllowGet").IsOriginAllowed = CorsOriginCheck;
+        }
+
+        /// <summary>
+        /// Check if the request origin is allowed based on the configured allowed origins in the config file.
+        /// </summary>
+        /// <param name="requestOrigin">The origin to check against the list of allowed cross-origin domains.</param>
+        /// <returns>True if the request origin matches a configured origin (based upon the CORS spec), or if the allowed origins contains '*', false otherwise.</returns>
```
markus101, line 28: Drop the XML docs here

src/NzbDrone.Host/CorsOriginConfigurator.cs
```diff
@@ -0,0 +1,60 @@
...
+                allowedOrigin => allowedOrigin.Equals(CorsConstants.AnyOrigin, StringComparison.Ordinal) ||
+                IsSameOrigin(allowedOrigin, requestOrigin));
+        }
+
+        /// <summary>
+        /// Check if two origins are the same based on scheme, host and port.
+        /// </summary>
+        /// <param name="configuredOrigin">The first origin to compare.</param>
+        /// <param name="requestOrigin">The second origin to compare.</param>
+        /// <returns>True if the scheme, host, and port match, false otherwise.</returns>
```
markus101, line 42: And here.

frontend/src/Utilities/createAjaxRequest.js
markus101: I think it's just the one in `src/index`, the others look to be things with the same name, but not actually using the `fetch` API.

I'm not sure if it will also be updated, probably safest to add it there directly as well, worst case it'll get cleaned up during a refactor.

src/NzbDrone.Host/Startup.cs
```diff
@@ -72,15 +72,18 @@ public void ConfigureServices(IServiceCollection services)
...
                     .AllowAnyHeader());
 
                 options.AddPolicy("AllowGet",
                     builder =>
-                    builder.AllowAnyOrigin()
+                    builder
+                    .AllowCredentials()
                     .WithMethods("GET", "OPTIONS")
                     .AllowAnyHeader());
             });
```
solidDoWant, line 89: I got to thinking about this and I think this may introduce a security issue _with users of this new feature specifically_. Looking for some feedback here.

Here is what the XHR flow looks like for the authentication middleware (Istio/envoy external auth + Authentik) that I'd like to use this with:

1. Front-end sends HTTP request to backend, which has a reverse proxy in the middle (`frontend -> reverse proxy -> backend`)
2. If the request has a specific valid credential cookie attached, it is considered "authenticated" by the reverse proxy and passed through to the backend (`frontend -> reverse proxy -> backend`, with a sidechannel between `reverse proxy <-> authentik` prior to forwarding the request to the backend)
3. If the cookie is missing or invalid, the reverse proxy forwards the request to Authentik, which returns a 302 redirect. This redirect changes the domain from `sonarr.domain.com` to `authentik.domain.com`, as well as the path.
4. The front-end receives the redirect and executes it.
5. Authentik receives this request and does some processing, returning another 302 redirect. This redirects back to the original domain name (`sonarr.domain.com`), but with an Authentik-specific `/outpost.goauthentik.io` path.
6. The front-end receives the redirect and executes it. PRIOR TO THIS POINT all requests had the `Origin` header set. AT AND AFTER THIS POINT, the `Origin` header is set to a literal string value of `null` for XHR requests. This is a browser security feature.
7. The reverse proxy intercepts this request and forwards it to Authentik. Authentik returns another 302 response, and includes the `Set-Cookie` header to set the cookie mentioned in (2). This 302 response redirects back `//application.domain.com/<originally requested path from (1)>`.
8. The front-end executes this redirect, going back to (1) but now with the authentication cookie.

Because the browser passes `Origin: null` in part of this flow, the backend _must_ set `Access-Control-Allow-Origin: null` for this to work. The problem I see is that this basically bypasses all CORS protections. If an attacker can get the user's browser to execute a request to `sonarr.domain.com` from a separate `attacker.web.site` while setting `Origin: null` (which is trivial), then request will succeed.

This could potentially be worked around via a CSRF token, but that adds complexity to the entire application (every page would need one, not sure if middleware can solve this) and doesn't entirely fix the problem.

**TL;DR:** this approach works but has an authentication bypass vulnerability that can be triggered if an attacker can get the user's browser to make a request to Sonarr. @markus101 Is it better to support this auth flow with this known security vulnerability, or not support it at all?

solidDoWant:
I'll get the comments you left fixed. Looking for some feedback on https://github.com/Sonarr/Sonarr/pull/8178#discussion_r2531165714, so I'm requesting another review.

src/NzbDrone.Host/Startup.cs
markus101: Good catch, I think it'd be better to not support it, since we'd be knowingly creating a security issue. I think it'd be better instead of implement OIDC properly (so Sonarr instigates the login) instead of relying on it being proxied (there is an issue open for this).

CSRF token would also be useful security wise, but likely a pretty large undertaking.

solidDoWant: Yea that's fair. I'll close this PR. I've also raised an issue about this with Authentik to see if their "proxy provider" approach to this is fundamentally incompatible with application-initiated requests (fetch, xhr, etc) in hopes that there's a way around this.

Is this the issue you're referring to: https://github.com/Sonarr/Sonarr/issues/2477? I'm willing to take a look at what it would require to add oidc support if work hasn't started already, though I can't promise that I'll get to the point where I open a PR for it.

solidDoWant:
Closing due to a security issue with this, see https://github.com/Sonarr/Sonarr/pull/8178#discussion_r2531165714 for details.

PR #8186, opened 2025-11-15
Fix changing sort direction

markus101:
#### Description

Previously you couldn't change the sort direction for things converted to React Query, you can now and added a `setSort` helper to handle setting it. Ideally, `setOption` and `setOptions` would prevent setting `sortKey` and `sortDirection` by typings, but that would get complicated with more generic `handleOptionChange` like `QueueOptions` has. Eventually we may be able to improve that, but for the time being those generic settings will throw.


mynameisbogdan:
Secondary sorting keys are still supported?

markus101:
Yeah, it's still used for sorting series index.

Issue #6960, opened 2024-07-14
Return Additional Information With Status / Rejection Messages


markus101:
### Is there an existing issue for this?

- [X] I have searched the existing open and closed issues

### Is your feature request related to a problem? Please describe

Currently status messages such as https://github.com/Sonarr/Sonarr/blob/e97e5bfe8f3a7f01f7188f786eb7b51f1d8ef8b5/src/NzbDrone.Core/Download/Clients/QBittorrent/QBittorrent.cs#L248 are translated while processing, which means all API consumers get the same translation. Additionally, rejections for Interactive Search or Manual Import are just text rejections that can't be used for filtering.

### Describe the solution you'd like

We need to come up with a solution that allows for statuses and rejections to be returned via the API as an object instead of a string.

For rejections the API should return the rejection type (enum) and the translated status message. For status messages the message should be returned in addition to an enum representing the message.

Internally we should only translate when we're returning the result from the API, which would allow the API consumer to request a specific language instead of an app-wide setting that controls translations. We can pass around the translation token and any arguments to it to achieve this.

### Describe alternatives you've considered

Nothing else is really worth considering.

ManiMatter:
hi @markus101 

Given you linked #4132 (which refers to the interactive search) as well as #6959 and #6865, is it correct that with this task you aim to add the `sub status` across all relevant APIs (eg. `/queue`, `/release` for those cases)?

For `/queue`, I guess we have to come up with some `sub status` that are somewhat applicable across all clients.
Looking at the qbit client, would this first proposal be helpful?

| #  | Status  | Sub-Status |
| -- | ------------- | ------------- |
| 1 | warning  | error |
| 2 | paused  | paused  |
| 3 | queued  | queued  |
| 4 | completed  | completed  |
| 5 | warning  | stalled  |
| 6 | warning  | missing_files  |
| 7 | queued  | meta_dl  |
| 8 | warning  | dht_disabled  |
| 9 | downloading  | downloading  |
| 10 | downloading  | unkown  |

### 1. Status
Status: warning
Sub-Status: error
https://github.com/Sonarr/Sonarr/blob/b1527f9abbe4903c89058d5e351b914bae6a7162/src/NzbDrone.Core/Download/Clients/QBittorrent/QBittorrent.cs#L249-L252

### 2. Status
Status: paused
Sub-Status: paused
[...]

PR #8188, opened 2025-11-16
V5 Interactive Search 

### LINKED ISSUE #3955: Indicate which results have been previously grabbed
Having the results in the search tab colour coded would make it easier to determine what release was snatched and which files are preferred over the others.

Pull from Radarr PR
https://github.com/Radarr/Radarr/pull/5008

markus101:
#### Description

Adds new endpoints for v5 releases (entirely new structure, including full rejection information) and updates the UI to use it with React Query. Filtering by rejection reason has been added to Custom Filters as well. Also added in which releases were grabbed.

Rejection filtering:
image

Previously grabbed:
image

* Towards #6960
* Closes #3955


frontend/src/InteractiveSearch/InteractiveSearchRow.tsx
```diff
@@ -206,6 +213,43 @@ function InteractiveSearchRow(props: InteractiveSearchRowProps) {
...
+        {isBlocklisted ? (
+          <Icon
+            containerClassName={
+              history ? styles.blocklistIconContainer : undefined
+            }
+            name={icons.BLOCKLIST}
+            kind={kinds.DANGER}
+            title={
+              history?.failed
+                ? `${translate('Blocklisted')}: ${formatDateTime(
```
mynameisbogdan: I would recommend using `"BlocklistedAt": "Blocklisted at {date}"` instead for better localization.

src/Sonarr.Api.V5/Release/ReleaseHistoryResource.cs
```diff
@@ -0,0 +1,40 @@
...
+    public DateTime? Failed { get; set; }
+}
+
+public static class ReleaseHistoryResourceMapper
+{
+    public static ReleaseHistoryResource? ToResource(this ReleaseInfo release, List<EpisodeHistory> history)
+    {
+        var grabbedHistory = history.FirstOrDefault(h => h.EventType == EpisodeHistoryEventType.Grabbed &&
+                                                         h.Data.TryGetValue("guid", out var guid) &&
+                                                         guid == release.Guid);
```
mynameisbogdan: I've been using a more primitive variant of this where the logic is done in the frontend, and from experience I noticed the `guid` is not as reliable as some sites rotate authkeys. A possible fallback to indexer and source title matching should be enough to catch old grabs with a different guid.

frontend/src/InteractiveSearch/InteractiveSearchRow.tsx
mynameisbogdan: I noticed you're using history failed for blocklisted data, but sadly provides the wrong datetime since the necessary data is in the blocklist table.

markus101: Failed and blocklist are going to be added within seconds of each in the majority of cases, more reliable than guessing which release might be blocklisted I think.

Two timestamps from when I tested this:

History: `2025-11-17 00:19:11.15835Z`
Blocklist: `2025-11-17 00:19:11.170017Z`

src/Sonarr.Api.V5/Release/ReleaseHistoryResource.cs
markus101: That makes sense, I think you pretty much have to do the same logic that is done for Blocklist then, unique ID (Guid/torrent hash), then title + indexer + published date.

frontend/src/InteractiveSearch/InteractiveSearchRow.tsx
mynameisbogdan: I guess that might work for info purposes. I thought blocklisting isn't cleared on series deletion but it is.

https://github.com/Sonarr/Sonarr/blob/a5ea19ddfb20106a70ba0eca2f82f9a4587ef65a/src/NzbDrone.Core/Blocklisting/BlocklistService.cs#L209-L212

src/Sonarr.Api.V5/Release/ReleaseHistoryResource.cs
mynameisbogdan: Published date is going to be problematic with trackers that don't send a date and `now` is being used.

markus101: I've updated it to follow the same logic that Blocklisting does (refactored to pull out the comparisons), should be as accurate as we can get.

frontend/src/InteractiveSearch/InteractiveSearchRow.tsx
markus101: Ideally I'd like to just use Blocklist directly, but processing the blocklist a second time to get information we've grabbed and tossed seems wasteful, but refactoring to be able to to include additional information via rejections is a bit of an undertaking for minimal gain.

mynameisbogdan: Indeed, I was trying to find ways where you could have blocklist with missing history but unless there's a race condition or corruption you should have always both.

src/Sonarr.Api.V5/Release/ReleaseHistoryResource.cs
mynameisbogdan: A huge improvement, I might say. 👍

PR #8193, opened 2025-11-18
Update SDK version to 6.0.428

supergonzales:
#### Description

Update SDK version and therefore updating runtime included with released version to get latest security patches available.


stevietv:
Sonarr V5 has migrated to dotnet 8

supergonzales:
Indeed but it's not released yet.

The version 6.0.405 dates from January 2023 and the last version 6.0.428 dates from November 2024. Several security patches has been included in between.

mynameisbogdan:
There's no .NET 6 build for freebsd and later builds of .NET 6 fail to run on synologies.

markus101:
Unfortunately due to the issues Bogdan mentioned this isn't something we can land as builds will be unsuccessful and would break some existing installations.

PR #8194, opened 2025-11-20
V5 Episode Files

markus101:
#### Description

Adds API endpoints and converts UI to use react-query. Also two smaller changes, title for Manual Import from Queue and the "import" action for Manage Episodes from Series Details.

### DIFF
--- a/frontend/src/InteractiveImport/Interactive/InteractiveImportModalContent.tsx
+++ b/frontend/src/InteractiveImport/Interactive/InteractiveImportModalContent.tsx
@@ -956,7 +956,7 @@ function InteractiveImportModalContentInner(
             isDisabled={!selectedIds.length || !!invalidRowsSelected.length}
             onPress={onImportSelectedPress}
           >
-            {translate('Import')}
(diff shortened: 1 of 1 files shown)


src/Sonarr.Api.V5/EpisodeFiles/EpisodeFileController.cs
```diff
@@ -0,0 +1,202 @@
...
+            var series = _seriesService.GetSeries(seriesId.Value);
+            var files = _mediaFileService.GetFilesBySeries(seriesId.Value);
+
+            if (files == null)
+            {
+                return new List<EpisodeFileResource>();
+            }
+
+            return files.ConvertAll(e => e.ToResource(series, _upgradableSpecification, _formatCalculator))
+                        .ToList();
```
mynameisbogdan: Not needed since ConvertAll already returns a list.

```suggestion
            return files.ConvertAll(e => e.ToResource(series, _upgradableSpecification, _formatCalculator));
```

PR #8202, opened 2025-11-25
Add v5 root folder endpoints

### LINKED ISSUE #8196: Add notes about root folder needs to be non-empty if "Create Empty Series Folders" is enabled

From https://github.com/Sonarr/Sonarr/issues/4473, I understand that this is a fail safe, not a bug.

This issue took me a while to resolve, so it would be nice if users know when they want to enable the option.

Can we add an extra comment to this option?

Something like "Make sure the root folder contains at least a file so Sonarr can create new empty folders"?

`$ touch this_is_here_so_sonarr_knows_this_folder_exists.txt`

### Anything else?

_No response_

markus101:
#### Description
Converts Root Folders to use React Query and v5 API, also added a label and tool tip for empty root folders and a health check for empty root folders (warning).

Root folder list
image

Health check
image

* Closes #8196


src/Sonarr.Api.V5/RootFolders/RootFolderController.cs
```diff
@@ -0,0 +1,73 @@
...
+using NzbDrone.SignalR;
+using Sonarr.Http;
+using Sonarr.Http.Extensions;
+using Sonarr.Http.REST;
+using Sonarr.Http.REST.Attributes;
+
+namespace Sonarr.Api.V5.RootFolders;
+
+[V5ApiController]
+public class RootFolderController : RestControllerWithSignalR<RootFolderResource, RootFolder>
```
mynameisbogdan, line 14: A rare opportunity to fix the endpoint name from singular to plural? `/rootfolders{/$id}` would more appropriate than `/rootfolder{/$id}`.

```suggestion
public class RootFoldersController : RestControllerWithSignalR<RootFolderResource, RootFolder>
```

markus101: I'm not sure plural or singular really matters, so long as we're consistent, though I'm not opposed to it, but I think it warrants a bit larger discussion and if we decide to, do it as a change once everything has been migrated before shipping v5 to make sure we get everything.

mynameisbogdan: I was under the impression that for collections plural is the norm since it's more intuitive. [MS suggests the same thing](https://learn.microsoft.com/en-us/azure/architecture/best-practices/api-design#resource-uri-naming-conventions).

PR #8204, opened 2025-11-25
Bump Sentry to 5.16.2

mynameisbogdan:
4.0.2 is deprecated. See https://www.nuget.org/packages/Sentry/4.0.2


mynameisbogdan:
Since sentry is not being configured for CI builds, disabling SentryUploadSymbols is needed for a successful build.

PR #8218, opened 2025-11-30
v5 Series

markus101:
#### Description

Lots of changes since series are so integrated in things (unsurprisingly), but now moved to v5 and React Query.


frontend/src/Series/useSeries.ts
```diff
@@ -1,19 +1,968 @@
...
+
+export const useHasSeries = () => {
+  const { data: seriesData = [] } = useSeries();
+
+  return useMemo(() => {
+    return seriesData.length > 0;
+  }, [seriesData]);
+};
+
+export const useSeriesWithSSlug = (titleSlug: string) => {
```
mynameisbogdan:
```suggestion
export const useSeriesWithSlug = (titleSlug: string) => {
```

PR #8222, opened 2025-12-01
Prevent command-line data paths with multiple consecutive slashes from causing backups to fail

### LINKED ISSUE #8221: Command-line data path causing backups to fail

### Current Behavior

In my environment I run Sonarr specifying a data dir on the command line, for example `/usr/bin/open /Applications/Sonarr.app --args /nobrowser -data=$APP_HOME/appdata/Sonarr/`.

When elements of my environment changed, $APP_HOME ended up expanding to a directory with a trailing slash, so the the parameter became `-data=/opt/example//appdata/Sonarr/`.

Most Sonarr features work just fine, but I noticed that backups started failing, and looking at the debug logs, other file-handling features were too. The backups were trying to back up the non-existent config file `/opt/example/appdata/Sonarrr/config.xml`. Note the single slashes and three Rs in Sonarrr.

Please note this is a duplicate of [Radarr issue 11300](https://github.com/Radarr/Radarr/issues/11300).

### Expected Behavior

I'd expect Sonarr's backups to work with a config file path like /opt/example//appdata/Sonarr/config.xml

### Steps To Reproduce

1. Start Sonarr using a command-line parameter to specify the data dir with multiple slashes within the path (not the final path separator). For example, `-data=/opt/example///appdata/Sonarr/` but not `-data=/opt/example/appdata/Sonarr///`.
2. In the System menu select the Backup page.
3. Click 'Backup Now' to initiate a manual backup.

The backup will fail and the log/debug log will show that the backup file failed because the malformed config file path couldn't be found.

### Environment

```markdown
- OS: macOS Sequoia 15.7.2
- Sonarr: 4.0.16.2944
- Docker Install: No
- Using Reverse Proxy: No
- Browser: n/a
- Database: Sqlite 3.36.0
```

Develop

### Trace Logs

[over-slashed-directory-trace-log-for-sonarr](https://gist.github.com/Touchstone64/adfdcdfaa433414a596dcb95062af410)

### Anything else?

Please note this is a duplicate of [Radarr issue 11300](https://github.com/Radarr/Radarr/issues/11300).

Touchstone64:
#### Description
This PR addresses [Issue 8221](https://github.com/Sonarr/Sonarr/issues/8221) by checking a path is 'clean' before resolving its parent path. Specifically, in this case, that the path doesn't include any multiple path separators. If it does, a warning is issued to the logger and a 'clean' copy of the path is used instead.

#### Database Migration
NO

* Closes #8221

### DIFF
--- a/src/NzbDrone.Common.Test/DiskTests/DiskTransferServiceFixture.cs
+++ b/src/NzbDrone.Common.Test/DiskTests/DiskTransferServiceFixture.cs
@@ -580,6 +580,21 @@ namespace NzbDrone.Common.Test.DiskTests
         }
 
         [Test]
+        public void TransferFile_should_find_files_with_multiple_slashes_within_their_path()
+        {
+            WithRealDiskProvider();
+
+            var root = GetFilledTempFolder();
+            var rootDir = root.FullName;
+            var from = Path.Combine(rootDir, "source-file");
+            var toRootDir = rootDir.Replace(Path.DirectorySeparatorChar.ToString(), new string(Path.DirectorySeparatorChar, 3));
+            var to = Path.Combine(toRootDir, "destination-file");
+            File.WriteAllText(from, "Source file");
+            var mode = Subject.TransferFile(from, to, TransferMode.Copy);
+            mode.Should().Be(TransferMode.Copy);
+        }
+
+        [Test]
         public void should_throw_if_destination_is_readonly()
         {
             Mocker.GetMock<IDiskProvider>()
--- a/src/NzbDrone.Common/Disk/DiskTransferService.cs
+++ b/src/NzbDrone.Common/Disk/DiskTransferService.cs
@@ -28,15 +28,15 @@ namespace NzbDrone.Common.Disk
 
         private string ResolveRealParentPath(string path)
         {
-            var parentPath = path.GetParentPath();
-            if (!_diskProvider.FolderExists(parentPath))
+            if (!_diskProvider.FolderExists(path.GetParentPath()))
             {
                 return path;
             }
 
+            var cleanPath = path.GetCleanPath();
+            var parentPath = cleanPath.GetParentPath();
             var realParentPath = parentPath.GetActualCasing();
-
-            var partialChildPath = path.Substring(parentPath.Length);
+            var partialChildPath = cleanPath.Substring(parentPath.Length);
 
             return realParentPath + partialChildPath;
         }
--- a/src/NzbDrone.Core.Test/Download/DownloadClientTests/RQBitTests/RQBitFixture.cs
+++ b/src/NzbDrone.Core.Test/Download/DownloadClientTests/RQBitTests/RQBitFixture.cs
@@ -7,6 +7,7 @@ using NUnit.Framework;
 using NzbDrone.Core.Download;
 using NzbDrone.Core.Download.Clients.RQBit;
 using NzbDrone.Core.MediaFiles.TorrentInfo;
+using NzbDrone.Test.Common;
 
 namespace NzbDrone.Core.Test.Download.DownloadClientTests.RQBitTests
 {
@@ -212,6 +213,8 @@ namespace NzbDrone.Core.Test.Download.DownloadClientTests.RQBitTests
 
             items.Should().HaveCount(1);
             items.First().Title.Should().Be("Test2");
+
+            ExceptionVerification.ExpectedWarns(1);
         }
 
         [Test]
@@ -229,6 +232,8 @@ namespace NzbDrone.Core.Test.Download.DownloadClientTests.RQBitTests
 
             items.Should().HaveCount(1);
             items.First().Title.Should().Be("Test2");
+
+            ExceptionVerification.ExpectedWarns(1);
         }
     }
 }
--- a/src/NzbDrone.Core.Test/MediaFiles/UpgradeMediaFileServiceFixture.cs
+++ b/src/NzbDrone.Core.Test/MediaFiles/UpgradeMediaFileServiceFixture.cs
@@ -145,6 +145,8 @@ namespace NzbDrone.Core.Test.MediaFiles
             Subject.UpgradeEpisodeFile(_episodeFile, _localEpisode);
 
             Mocker.GetMock<IMediaFileService>().Verify(v => v.Delete(_localEpisode.Episodes.Single().EpisodeFile, DeleteMediaFileReason.Upgrade), Times.Once());
+
+            ExceptionVerification.ExpectedWarns(1);
         }
 
         [Test]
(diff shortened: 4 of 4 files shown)


Touchstone64:
The first commit in the PR, while not strictly linked to the issue in hand, was necessary to get a clean run of unit tests using `./scripts/test.sh Mac Unit Test`.

Once that commit is in place, the basic confidence test is:

1. Checkout the test code changes but not the source code changes
2. Build the solution
3. Run the tests (./scripts/test.sh Mac Unit Test) to prove the new test fails with the existing source
4. Checkout the source code changes
5. Build the solution
6. Run the tests (./scripts/test.sh Mac Unit Test) to prove the new test passes with the new source

src/NzbDrone.Common/Disk/DiskTransferService.cs
```diff
@@ -28,15 +28,21 @@ public DiskTransferService(IDiskProvider diskProvider, Logger logger)
 
         private string ResolveRealParentPath(string path)
         {
-            var parentPath = path.GetParentPath();
-            if (!_diskProvider.FolderExists(parentPath))
+            var testExists = path.GetParentPath();
+            if (!_diskProvider.FolderExists(testExists))
```
markus101: No need for a variable if it's only going to be used once.

```suggestion
            if (!_diskProvider.FolderExists(path.GetParentPath()))
```

src/NzbDrone.Common/Disk/DiskTransferService.cs
```diff
@@ -28,15 +28,21 @@ public DiskTransferService(IDiskProvider diskProvider, Logger logger)
...
             {
                 return path;
             }
 
-            var realParentPath = parentPath.GetActualCasing();
+            var cleanPath = path.GetCleanPath();
+            if (cleanPath != path)
```
markus101: I don't think the warning adds much value, either we should silently fix it or we should expect errors in the logs.

src/NzbDrone.Common.Test/DiskTests/DiskTransferServiceFixture.cs
```diff
@@ -579,6 +579,29 @@ public void MoveFolder_should_rename_casesensitive_folder()
...
+            var to = Path.Combine(rootDir, "destination-file");
+
+            var oneSlash = new string(Path.DirectorySeparatorChar, 1);
+            var threeSlashes = new string(Path.DirectorySeparatorChar, 3);
+            var overSlashed = Path.Combine(rootDir.Replace(oneSlash, threeSlashes), fromFileName);
+
+            File.WriteAllText(from, "Source file");
+
+            var mode = Subject.TransferFile(overSlashed, to, TransferMode.Copy);
+            mode.Should().Be(TransferMode.Copy);
```
markus101, line 594: No need to store a bunch of variables and do string replacement, in testing this can be simplified to the following:

```suggestion
            var root = GetFilledTempFolder();
            var rootDir = root.FullName;
            var from = rootDir + Path.DirectorySeparatorChar + Path.DirectorySeparatorChar + "source-file";
            var to = Path.Combine(rootDir, "destination-file");

            File.WriteAllText(from, "Source file");

            var mode = Subject.TransferFile(from, to, TransferMode.Copy);
            mode.Should().Be(TransferMode.Copy);
```

src/NzbDrone.Common.Test/DiskTests/DiskTransferServiceFixture.cs
```diff
@@ -579,6 +579,29 @@ public void MoveFolder_should_rename_casesensitive_folder()
...
+
+            ExceptionVerification.ExpectedWarns(1);
```
markus101: Should be removed per my other comment.

markus101 (changes requested):
> Fixed: add warning traps for unit tests that fail without them

The `Fixed:` prefix should only be used for user facing changes, this isn't one that needs user visibility.

> Fixed: Command-line data path causing backups to fail

This needs to be more specific because it's not all `--data` paths, just ones with double slashes (potentially others, but that's the one we know 100% it happens with).

src/NzbDrone.Common.Test/DiskTests/DiskTransferServiceFixture.cs
Touchstone64: I committed this change to begin with, then realised that it doesn't test the underlying fix. Multiple trailing slashes already worked, it's multiple slashes seperating directories that are the problem. So `/one/two///file.txt` would be fine but `/one///two/file.txt` would break. Happy to accept any optimisations of course but I've reinstated the test code.

Touchstone64:
> This needs to be more specific because it's not all `--data` paths, just ones with double slashes (potentially others, but that's the one we know 100% it happens with).

Title changed :)

src/NzbDrone.Common.Test/DiskTests/DiskTransferServiceFixture.cs
markus101: The string replacement in the path feels a little janky overall, but still other ways to clean up the extra variables created here. Could even drop `toRootDir` as another variable, but does make it a bit more readable.

```suggestion
            var root = GetFilledTempFolder();
            var rootDir = root.FullName;
            var from = Path.Combine(rootDir, "source-file");
            var toRootDir = rootDir.Replace(Path.DirectorySeparatorChar.ToString(), new string(Path.DirectorySeparatorChar, 3));
            var to = Path.Combine(toRootDir, "destination-file");
            File.WriteAllText(from, "Source file");
            var mode = Subject.TransferFile(from, to, TransferMode.Copy);
            mode.Should().Be(TransferMode.Copy);
```

markus101 (changes requested):
Just the test that needs some cleanup still I think.

markus101 (approved):

markus101:
Thanks for the work on this!

Touchstone64:
> Thanks for the work on this!

No worries, thanks for all your work on Sonarr!

Touchstone64:
Hey folks, I notice this PR isn't included in the latest [pre-release](https://github.com/Sonarr/Sonarr/releases/tag/v4.0.17.2967), and the backups still fail as per the original issue (#8221).

Would it be possible to include these commits into a release? Thanks!

markus101:
Likely not, v4 isn't getting most fixes, mainly critical bug fixes. If we end up adding more changes there we may consider it.

Touchstone64:
> Likely not, v4 isn't getting most fixes...

How did I miss v4 vs. v5?!? Looking forward to the new version 👍

markus101:
It's been in development for a while, but haven't talked a ton about it, but didn't make a lot of sense to try and maintain multiple versions in development.

PR #8225, opened 2025-12-03
New: Bump FFMpegCore to 5.4.0 and FFprobe to 8.0.1

### LINKED ISSUE #8214: ffprobe failing with modern AAC profiles

### Current Behavior

The bundled version of ffprobe (5.1.4) doesn't recognize xHE-AAC audio streams. (A newish version of ffprobe (7.1.2) does.) This is causing Sonarr to not be able to detect if a downloaded file is a sample and thus to not import it (even manually), leaving it "stuck" in the queue.

### Expected Behavior

I expect Sonarr to handle (or at least safely ignore) such audio streams and to import the file as normal (that is, without manual intervention or getting "stuck").

### Steps To Reproduce

1. In the linuxserver docker image (v4.0.16.2944)
2. Attempt to import a downloaded file with an xHE-AAC (sometimes called USAC) audio stream
   - Creative Commons example files can be downloaded from [this page](https://www2.iis.fraunhofer.de/AAC/xhe-aac-abr.html). I confirmed that I experience the same error with the bundled version of ffprobe via CLI
3. Web UI reports "Unable to determine if file is a sample" and cannot be made to import the file

### Environment

```markdown
- OS: The host is a maintained Gentoo Linux sysyem
- Sonarr: 4.0.16.2944-ls299 by linuxserver.io
- Docker Install: Yes
- Using Reverse Proxy: No
- Browser: Various, but the core issue is demonstrable via CLI
- Database: Sqlite 3.49.2
```

Main

### Trace Logs?

[sonarr.trace.txt](https://github.com/user-attachments/files/23833761/sonarr.trace.txt)

### Anything else?

Here's a comparison between the outputs of the bundled ffprobe and the one on my host system (which is much newer)

[bundled_ffprobe.txt](https://github.com/user-attachments/files/23833743/bundled_ffprobe.txt)
[modern_ffprobe.txt](https://github.com/user-attachments/files/23833742/modern_ffprobe.txt)

mynameisbogdan:
#### Description
- Updating FFprobe to 8.0.1
- Support for interlaced scan type
- Support for `HE-AAC` and `xHE-AAC`
- Pixel formats dictionary is cached on library side.
- Removed RawFrameData and DoviConfigurationRecord from media info model are they are unused and not needed.

FFprobe changes:
- `DTS:X` is now `dts` + `DTS-HD MA + DTS:X IMAX`
- `ec+3` is now `eac3` + `Dolby Digital Plus + Dolby Atmos`
- `thd+` is now `truehd` + `Dolby TrueHD + Dolby Atmos`

* Closes #8214

### DIFF
--- a/src/NzbDrone.Core/MediaFiles/MediaInfo/VideoFileInfoReader.cs
+++ b/src/NzbDrone.Core/MediaFiles/MediaInfo/VideoFileInfoReader.cs
@@ -2,6 +2,7 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Text.Json.Nodes;
 using FFMpegCore;
 using NLog;
 using NzbDrone.Common.Disk;
@@ -19,10 +20,9 @@ namespace NzbDrone.Core.MediaFiles.MediaInfo
     {
         private readonly IDiskProvider _diskProvider;
         private readonly Logger _logger;
-        private readonly List<FFProbePixelFormat> _pixelFormats;
 
-        public const int MINIMUM_MEDIA_INFO_SCHEMA_REVISION = 12;
-        public const int CURRENT_MEDIA_INFO_SCHEMA_REVISION = 12;
+        public const int MINIMUM_MEDIA_INFO_SCHEMA_REVISION = 13;
+        public const int CURRENT_MEDIA_INFO_SCHEMA_REVISION = 13;
 
         private static readonly string[] ValidHdrColourPrimaries = { "bt2020" };
         private static readonly string[] HlgTransferFunctions = { "arib-std-b67" };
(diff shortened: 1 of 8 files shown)


src/NzbDrone.Core/MediaFiles/MediaInfo/VideoFileInfoReader.cs
```diff
@@ -19,7 +20,6 @@ public class VideoFileInfoReader : IVideoFileInfoReader
     {
         private readonly IDiskProvider _diskProvider;
         private readonly Logger _logger;
-        private readonly List<FFProbePixelFormat> _pixelFormats;
 
         public const int MINIMUM_MEDIA_INFO_SCHEMA_REVISION = 12;
         public const int CURRENT_MEDIA_INFO_SCHEMA_REVISION = 12;
```
markus101: Since we're upgrading all the things we should bump these I think, or at least the latter so it'll upgrade for people. Thoughts?

mynameisbogdan: That was the plan, but ATM only the support for (x)HE-AAC and scan type is making it necessary.

PR #8229, opened 2025-12-04
Prevent missing media info from failing script import

markus101:
#### Description

Guards against missing media info from failing to the import script.

* Towards #8214

### DIFF
--- a/src/NzbDrone.Core/MediaFiles/ScriptImportDecider.cs
+++ b/src/NzbDrone.Core/MediaFiles/ScriptImportDecider.cs
@@ -160,15 +160,31 @@ namespace NzbDrone.Core.MediaFiles
             environmentVariables.Add("Sonarr_Download_Client", downloadClientInfo?.Name ?? string.Empty);
             environmentVariables.Add("Sonarr_Download_Client_Type", downloadClientInfo?.Type ?? string.Empty);
             environmentVariables.Add("Sonarr_Download_Id", downloadId ?? string.Empty);
-            environmentVariables.Add("Sonarr_EpisodeFile_MediaInfo_AudioChannels", MediaInfoFormatter.FormatAudioChannels(localEpisode.MediaInfo).ToString());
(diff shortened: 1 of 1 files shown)


src/NzbDrone.Core/MediaFiles/ScriptImportDecider.cs
```diff
@@ -160,15 +160,31 @@ public ScriptImportDecision TryImport(string sourcePath, string destinationFileP
...
-            environmentVariables.Add("Sonarr_EpisodeFile_MediaInfo_Languages", localEpisode.MediaInfo.AudioLanguages.ConcatToString(" / "));
-            environmentVariables.Add("Sonarr_EpisodeFile_MediaInfo_Height", localEpisode.MediaInfo.Height.ToString());
-            environmentVariables.Add("Sonarr_EpisodeFile_MediaInfo_Width", localEpisode.MediaInfo.Width.ToString());
-            environmentVariables.Add("Sonarr_EpisodeFile_MediaInfo_Subtitles", localEpisode.MediaInfo.Subtitles.ConcatToString(" / "));
-            environmentVariables.Add("Sonarr_EpisodeFile_MediaInfo_VideoCodec", MediaInfoFormatter.FormatVideoCodec(localEpisode.MediaInfo, null));
-            environmentVariables.Add("Sonarr_EpisodeFile_MediaInfo_VideoDynamicRangeType", MediaInfoFormatter.FormatVideoDynamicRangeType(localEpisode.MediaInfo));
+
+            if (localEpisode.MediaInfo == null)
+            {
+                _logger.Trace("MediaInfo is null for episode file import. This may cause issues with the import script.");
```
mynameisbogdan, line 166: I think warning or even error would be more suitable in this case.

```suggestion
                _logger.Warn("MediaInfo is null for episode file import. This may cause issues with the import script.");
```

markus101: The issue is it's going to loop the same message over and over (in most cases) until ffprobe is updated, so I don't see a ton of value in making it a warning. In other places we don't warn either, going to keep this trace for the time being.

PR #8234, opened 2025-12-06
Calculate custom score using renamed filename before importing.

felizk:
#### Description
This change determines the renamed filename for a recently downloaded file and uses that name to calculate the custom score before importing.

This fixes an issue where the custom score that can be determined from the release name doesn't match score after renaming. For example, if you have custom format's that rely on MediaInfo tokens in the renamed file.

This could cause undesirable files to overwrite existing good files.

This change empowers the use of Release Title custom formats to check for MediaInfo features such as whether subtitles are included with a specific language or the file has the correct HDR video format.

This is accomplished by calling a new function on IBuildFileNames that takes a LocalEpisode, before calculating the custom score for the localepisode. 

### Next Steps
I have not written any tests yet as I wanted to check whether this is a change you're even interested in.


src/NzbDrone.Core/CustomFormats/CustomFormatCalculationService.cs
```diff
@@ -114,7 +114,7 @@ public List<CustomFormat> ParseCustomFormat(EpisodeHistory history, Series serie
             return ParseCustomFormat(input);
         }
 
-        public List<CustomFormat> ParseCustomFormat(LocalEpisode localEpisode)
+        public List<CustomFormat> ParseCustomFormat(LocalEpisode localEpisode, string postRenameFileName)
```
markus101: I think instead of a second argument, a new property for the future filename be a better approach.

src/NzbDrone.Core/Organizer/FileNameBuilder.cs
```diff
@@ -14,13 +14,15 @@
...
 using NzbDrone.Core.MediaFiles.MediaInfo;
+using NzbDrone.Core.Parser.Model;
 using NzbDrone.Core.Qualities;
 using NzbDrone.Core.Tv;
 
 namespace NzbDrone.Core.Organizer
 {
     public interface IBuildFileNames
     {
+        string BuildFileName(LocalEpisode localEpisode, string extension = "", NamingConfig namingConfig = null, List<CustomFormat> customFormats = null);
```
markus101: I think a lot of these changes could be eliminated by converting `LocalEpisode` to `EpisodeFile` similar to what is done in `ImportApprovedEpisodes` (probably extracted to a separate function so that logic can be re-used).

src/NzbDrone.Core/MediaFiles/MediaInfo/UpdateMediaInfoService.cs
```diff
@@ -85,7 +85,11 @@ public bool UpdateMediaInfo(EpisodeFile episodeFile, Series series)
             }
 
             episodeFile.MediaInfo = updatedMediaInfo;
-            _mediaFileService.Update(episodeFile);
+            if (episodeFile.Id != 0)
```
felizk, line 89: Because we're creating EpisodeFile instances from LocalEpisodes without registering them with the database, there is a risk we'll hit this branch with Id == 0, which would cause an exception.

src/NzbDrone.Core/MediaFiles/EpisodeImport/LocalEpisodeCustomFormatCalculationService.cs
```diff
@@ -0,0 +1,41 @@
...
+
+namespace NzbDrone.Core.MediaFiles.EpisodeImport;
+
+public interface ILocalEpisodeCustomFormatCalculationService
+{
+    public List<CustomFormat> ParseEpisodeCustomFormats(LocalEpisode localEpisode, out string fileNameUsedForCustomFormatCalculation);
+    public void UpdateEpisodeCustomFormats(LocalEpisode localEpisode);
+}
+
+public class LocalEpisodeCustomFormatCalculationService : ILocalEpisodeCustomFormatCalculationService
```
felizk, line 15: I tried many different patterns for encapsulating the use of IBuildFileNames and ICustomFormatCalculationService as well as updating multiple fields on LocalEpisode.

Using a separate service to deal with all of it seemed to be the way that fits the rest of the code base best. Rather than extension methods, or adding methods to LocalEpisode that take services as arguments.

src/NzbDrone.Core/CustomFormats/CustomFormatCalculationService.cs
```diff
@@ -19,7 +19,7 @@ public interface ICustomFormatCalculationService
         List<CustomFormat> ParseCustomFormat(EpisodeFile episodeFile);
         List<CustomFormat> ParseCustomFormat(Blocklist blocklist, Series series);
         List<CustomFormat> ParseCustomFormat(EpisodeHistory history, Series series);
-        List<CustomFormat> ParseCustomFormat(LocalEpisode localEpisode);
+        List<CustomFormat> ParseCustomFormat(LocalEpisode localEpisode, string fileName);
```
felizk, line 22: Because we want to get two sets of CustomFormat from a local episode (before and after rename), it wasn't possible to simply add the filename property to LocalEpisode without ending up with a Method here that mutates LocalEpisode.

Mutating LocalEpisode here would be odd and there is a single instance in ManualImportService where we want to get the CustomFormats for a LocalEpisode but we're assigning the result to a ManualImportItem instead of LocalEpisode.

src/NzbDrone.Core/MediaFiles/EpisodeImport/LocalEpisodeCustomFormatCalculationService.cs
markus101: Yeah I think it makes sense, though the the `out string` makes it a bit weird since it's only really used internally to the file.

I think this would make more sense at the cost of one line being duplicated:

```
public List<CustomFormat> ParseEpisodeCustomFormats(LocalEpisode localEpisode)
    {
        var fileNameUsedForCustomFormatCalculation = _fileNameBuilder.BuildFileName(localEpisode.Episodes, localEpisode.Series, localEpisode.ToEpisodeFile());

        return ParseEpisodeCustomFormats(localEpisode, fileNameUsedForCustomFormatCalculation);
    }

    public void UpdateEpisodeCustomFormats(LocalEpisode localEpisode)
    {
        var fileNameUsedForCustomFormatCalculation = _fileNameBuilder.BuildFileName(localEpisode.Episodes, localEpisode.Series, localEpisode.ToEpisodeFile());

        localEpisode.CustomFormats = ParseEpisodeCustomFormats(localEpisode, fileNameUsedForCustomFormatCalculation);
        localEpisode.FileNameUsedForCustomFormatCalculation = fileNameUsedForCustomFormatCalculation;
        localEpisode.CustomFormatScore = localEpisode.Series.QualityProfile?.Value.CalculateCustomFormatScore(localEpisode.CustomFormats) ?? 0;

        localEpisode.OriginalFileNameCustomFormats = _formatCalculator.ParseCustomFormat(localEpisode, Path.GetFileName(localEpisode.Path));
        localEpisode.OriginalFileNameCustomFormatScore = localEpisode.Series.QualityProfile?.Value.CalculateCustomFormatScore(localEpisode.OriginalFileNameCustomFormats) ?? 0;
    }

    private List<CustomFormat> ParseEpisodeCustomFormats(LocalEpisode localEpisode, string fileNameUsedForCustomFormatCalculation)
    {
        return _formatCalculator.ParseCustomFormat(localEpisode, fileNameUsedForCustomFormatCalculation);
    }
```

felizk: That out param follow me around for a while and I didn't realize I didn't need it any more when the local service came into being. Good catch!

src/NzbDrone.Core/MediaFiles/MediaInfo/UpdateMediaInfoService.cs
```diff
@@ -85,7 +85,11 @@ public bool UpdateMediaInfo(EpisodeFile episodeFile, Series series)
...
+            {
+                _mediaFileService.Update(episodeFile);
+            }
```
markus101, line 92: New line before if block:
```suggestion

            if (episodeFile.Id != 0)
            {
                _mediaFileService.Update(episodeFile);
            }
```

felizk: Hey Markus, sorry for the late response.

When I test this locally the import is already rejected when the already imported file has a higher custom score than the newly downloaded file.

In the case where the custom score is higher before you do the local rename, it might still be better than what you already have imported and you might not want to reject just on that basis.

We wouldn't want to reject just because the two numbers don't match, that's to be expected in many cases, because the filename doesn't have as much info as the downloaded file with the mediainfo added to the file.

felizk:
Without the latest changes, the import would be rejected correctly, but it would not prevent the same release from being downloaded again.

I updated the `DownloadedEpisodesImportService` to mark a download failed using `IFailedDownloadService` if any import from the download was rejected specifically because custom format was expected to be higher based on the original filename.

This should prevent the same release from being downloaded again.

markus101:
That shouldn't result in the download being failed. The file should not import, but it should remain in the queue for the user to decide what to do.

felizk:
OK, then I'll revert this latest change and it'll do what we expect! I'll keep the custom error message and reject reason

src/NzbDrone.Core/MediaFiles/MediaInfo/UpdateMediaInfoService.cs
markus101: > When I test this locally the import is already rejected when the already imported file has a higher custom score than the newly downloaded file.

Right, but that's expected, otherwise it'd be a downgrade (assuming equal quality).

> In the case where the custom score is higher before you do the local rename, it might still be better than what you already have imported and you might not want to reject just on that basis.
> We wouldn't want to reject just because the two numbers don't match, that's to be expected in many cases, because the filename doesn't have as much info as the downloaded file with the mediainfo added to the file.

The media info would be the same, so file name + media info CF score should either be the same or the renamed file higher, shouldn't it? Otherwise we're kind of setting things up for failure.

Perhaps comparing the file before import with the post-import file isn't the right choice, but if we're checking the renamed file then we should expect that it's equal or higher to the release name at time of grab otherwise we are going to end up with a download loop or at least unnecessary grabs won't we?

felizk: In my particular case, the one that prompted me to start this work, my renamed files contain a bunch of media info, audio languages, subtitles, dolby vision/hdr info. 

This is used to check if the downloaded file has English subtitles for example and I have a custom format thats scored very negative if English is missing. 

Its a very useful and flexible way to validate downloaded files. The trouble is without this change, I dont find out the newly downloaded file has a bad score until its already imported the file.

Because we dont know the media info before the download has completed, there will always be cases where the release looks like it will have a good custom format score before download, that then drops once the media info is added. 

The loop issue is what prompted me to add the download failed feature I reverted back out. I want to prevent sonarr from downloading the same release again now that the mediainfo has exposed the file to be bad.

felizk: So the question is, do we go ahead with this change? Then figure out how to reject bad releases. Or should I bring back the rejection code?

markus101: Yeah, I think it makes sense, worst case we run into issues and either revert or find another solution.

Issue #424, opened 2014-03-11
Restarting on Mono should not cause service wrapper to restart Sonarr


SonarrBot:
This causes launchd (OS X) and upstart (Linux) to have a fit.

Taloth:
Upstart and other external restart during update should be fixed by commit 7d23f24854854b41faaf3e7574dd64f89e3b1ad6.
So that leaves UI restart.

bakerboy448:
systemd  not upstart these days

markus101:
Resolved by 6cd1ab3764587f1a6784f0d46efc2fc749149056

Issue #1194, opened 2016-03-21
System should have a function to Cancel/Suspend all Tasks


Taloth:
It's fairly 1%, but in those odd cases where you mess up and cause a huge indexer search and or back/forth process with the download client, it can be useful to be able to suspend sonarr while trying to fix stuff up. Kinda like a 'panic' button.
I've run into something similar myself a long time ago where some issue caused repeated failed downloads going back and forth between sonarr and the download client, had to disable CDH/FDH, terminate sonarr and other drastic steps to stop it from repeatedly searching indexers.

Some criteria:
- It should cause a clear health check Error.
- It should prevent any new scheduled tasks to be executed.
- It should prevent any non-Manual commands to be performed. 
- If possible, it should cancel any running command without corrupting state.

serialoverflow:
If this feature is implemented, i would appreciate it if it wasn't just a panic button but that each task would separately be suspendable. I run multiple versions of sonarr and it is very inconvenient that i have no control over the library update schedule.

markus101:
Being able to cancel a current or upcoming task without data loss would also be useful, to abort a missing or series search for example. There would be some additional work to handle cancelling one instead of the panic button, but I think the majority of the work would come from making the cancellation graceful, so adding it here and we can open a new GHI as required.

bakerboy448:
@Taloth  any plans or close?

Paul5656:
Any word on this? I've also come across this situation lots of times where sonarr will just go on an infinite recursive search for individual episodes. A Stop All Active Searches option would be nice in such situations instead of having to Stop and Restart the container.

markus101:
No update = no change. We'll update the issue when there is something.

Open a thread in one of the support channels with the relevant debug/trace logs, that isn't expected nor something we've heard of happening before and this feature certainly isn't a fix for such an issue.

Paul5656:
Well, I don't usually have trace logs enabled but it would be convenient to be able to cancel any ongoing searches without having to shut down the container for the reasons OP mentioned.

ajax-lives:
Over half a decade...

markus101:
@ajax-lives Contributor's time to work on Sonarr is limited as such issues may be open for an extended period of time. If you are interested in a feature that has not yet been implemented and you're able to contribute yourself please look at our [documentation](https://github.com/Sonarr/Sonarr/blob/develop/CONTRIBUTING.md) for more information on doing so.

[...]

PR #8239, opened 2025-12-07
Fix external restart handling

TypNull:
#### Description

Fixes application restart failures and boot loops across all environments by implementing an internal restart loop instead of spawning new processes. When a restart is requested, the application restarts now within the same parent process, eliminating conflicts with external process managers.

**Commit 1 (Core Fix):** Adds restart loop in `Bootstrap.cs` that monitors `RestartPending` flag, removes process spawning logic from `AppLifetime.cs` and `LifecycleService.cs`.

**Commit 2 (Optional API):** Adds automatic detection for Podman and systemd environments and exposes `IsContainerized` property in System API. Maintains backward compatibility with existing `IsDocker` property.

**Commit 3 (Optional UI)**: Re-enables restart button. Since the restart mechanism now works, this button no longer needs to be hidden in Docker.

#### Database Migration
NO

* Fixes restart-related boot loops in managed environments. Should be related to: #424

---

**Note:** This solution mirrors how Jellyfin handles restarts, keeping the process under the original PID and letting the application manage its own restart cycle internally. It was easy after thinking why does Jellyfin not have the same problem.

Frontend calls /system/restart but backend expects /api/v3/system/restart, possible incomplete API version migration

### DIFF
--- a/src/NzbDrone.Core/Messaging/Events/EventAggregator.cs
+++ b/src/NzbDrone.Core/Messaging/Events/EventAggregator.cs
@@ -7,6 +7,7 @@ using NzbDrone.Common;
 using NzbDrone.Common.EnsureThat;
 using NzbDrone.Common.EnvironmentInfo;
 using NzbDrone.Common.Messaging;
+using NzbDrone.Common.Reflection;
 using NzbDrone.Common.TPL;
 
 namespace NzbDrone.Core.Messaging.Events
@@ -56,9 +57,9 @@ namespace NzbDrone.Core.Messaging.Events
 
             var eventName = GetEventName(@event.GetType());
 
-            if (_runtimeInfo.IsExiting && Attribute.IsDefined(@event.GetType(), typeof(LifecycleEventAttribute)))
+            if (_runtimeInfo.IsExiting && @event.GetType().HasAttribute<LifecycleEventAttribute>())
             {
-                _logger.Debug("Event {0} blocked due to application shutdown", eventName);
+                _logger.Warn("Event {0} blocked due to application shutdown", eventName);
                 return;
             }
 
@@ -86,7 +87,15 @@ namespace NzbDrone.Core.Messaging.Events
             {
                 if (!_eventSubscribers.TryGetValue(eventName, out var target))
                 {
-                    _eventSubscribers[eventName] = target = new EventSubscribers<TEvent>(_serviceFactory);
+                    try
+                    {
+                        _eventSubscribers[eventName] = target = new EventSubscribers<TEvent>(_serviceFactory);
+                    }
+                    catch (Exception ex)
+                    {
+                        _logger.Warn(ex, "Unable to resolve event subscribers for {0}, container may be disposed", eventName);
+                        return;
+                    }
                 }
 
                 subscribers = target as EventSubscribers<TEvent>;


frontend/src/typings/SystemStatus.ts
```diff
@@ -9,7 +9,7 @@ interface SystemStatus {
   instanceName: string;
   isAdmin: boolean;
   isDebug: boolean;
-  isDocker: boolean;
+  isContainerized: boolean;
```
markus101, line 12: Makes sense to have both.

src/NzbDrone.Core/Lifecycle/LifecycleService.cs
```diff
@@ -18,41 +16,28 @@ public class LifecycleService : ILifecycleService, IExecute<ShutdownCommand>, IE
...
-            _serviceProvider = serviceProvider;
             _logger = logger;
         }
 
         public void Shutdown()
         {
             _logger.Info("Shutdown requested.");
-            _eventAggregator.PublishEvent(new ApplicationShutdownRequested());
-
-            if (_runtimeInfo.IsWindowsService)
```
markus101: Is this handled somewhere else now?

src/NzbDrone.Core/Lifecycle/LifecycleService.cs
```diff
@@ -18,41 +16,28 @@ public class LifecycleService : ILifecycleService, IExecute<ShutdownCommand>, IE
...
         }
 
         public void Restart()
         {
             _logger.Info("Restart requested.");
-
+            _runtimeInfo.RestartPending = true;
             _eventAggregator.PublishEvent(new ApplicationShutdownRequested(true));
-
-            if (_runtimeInfo.IsWindowsService)
```
markus101: What about restarting the Windows service?

frontend/src/Components/Page/Header/PageHeaderActionsMenu.tsx
```diff
@@ -45,24 +45,22 @@ function PageHeaderActionsMenu(props: PageHeaderActionsMenuProps) {
...
-              <MenuItem onPress={handleShutdownPress}>
-                <Icon
-                  className={styles.itemIcon}
-                  name={icons.SHUTDOWN}
-                  kind={kinds.DANGER}
-                />
-                {translate('Shutdown')}
-              </MenuItem>
-            </>
+          {!isContainerized && (
```
markus101: Keep the ternary, it avoids non-boolean truthy values causing unexpected rendering.

```suggestion
          {isContainerized ? null : (
```

frontend/src/typings/SystemStatus.ts
TypNull: Why? The `isDocker` variable is exposed by the backend API but is not currently being used. For what case holding on to it?

src/NzbDrone.Core/Lifecycle/LifecycleService.cs
TypNull: I have my problems with these lines since I made a mistake with the bootstrap implementation. However, I'm not clear on what they actually do, the Windows Service restarts and shuts down properly even without them, at least in my testing. 
I'll run additional tests to confirm.
As mentioned, I made an error with the restart loop, but I'm not seeing any difference in behavior whether these lines are included or not during shutdown or restart. The same applies when using the `net stop Sonarr` command.
Or could you explain to me what function they actually have?

TypNull: Ah, got it! The problem is that after a restart, `net stop Sonarr` sees the service as "not running" and forcibly kills the process instead of gracefully shutting down.

I think I have a workaround. I'll push a commit, please let me know if it works or if you see any issues with the approach.

TypNull:
I’ve implemented a fix for the Windows Service shutdown process.

- Introduced a singleton `WindowsServiceWrapper` to preserve the same `ServiceBase` instance across restarts, ensuring consistent registration with Windows SCM.
- Added `NzbDroneLogger.ResetAllTargets()` to clear logging between restart cycles.

Has been tested on:
- Windows (Console and Service mode)
- Linux (Ubuntu x86_64 24.04.3 LTS) in Console, systemd and Docker environments.

For validation, I recommend testing the shutdown and restart scenarios on all environments yourself. 
I don’t have access to a macOS environment, so I couldn’t test it there, let me know if you encounter any issues.

src/NzbDrone.Common/EnvironmentInfo/OsInfo.cs
```diff
@@ -79,11 +81,15 @@ public OsInfo(IEnumerable<IOsVersionAdapter> versionAdapters, Logger logger)
...
+            if (IsLinux)
             {
-                IsDocker = true;
+                IsDocker = File.Exists("/.dockerenv") ||
+                           (File.Exists("/proc/1/cgroup") && File.ReadAllText("/proc/1/cgroup").Contains("/docker/"));
+                IsPodman = File.Exists("/run/.containerenv") ||
+                           Environment.GetEnvironmentVariable("container") != null;
+
+                var dotNetEnvironmentVariable = Environment.GetEnvironmentVariable("DOTNET_RUNNING_IN_CONTAINER");
+                IsContainerized = IsDocker || IsPodman || dotNetEnvironmentVariable == "true" || dotNetEnvironmentVariable == "1";
```
mynameisbogdan:
```suggestion
                IsContainerized = IsDocker || IsPodman || Environment.GetEnvironmentVariable("DOTNET_RUNNING_IN_CONTAINER") is "true" or "1";
```

mynameisbogdan (changes requested):

src/Sonarr.Api.V3/openapi.json
```diff
@@ -12097,9 +12097,15 @@
           "isWindows": {
```
mynameisbogdan, line 12097: You should leave these to the API docs workflow to generate them.

TypNull: There is a workflow for this? I’ll remove these changes.

Does the rest of the implementation work for you, or do you have any concerns about the current restart design?

mynameisbogdan: There is https://github.com/Sonarr/Sonarr/blob/v5-develop/.github/workflows/api_docs.yml

Sadly I wasn't able to test it, git apply complains about whitespacing and even if doesn't throw an error it fails to apply the changes in full.

TypNull: The question about the workflow was more rhetorical, but thanks anyway.
I hope someone will be able to test the restart fix.
Maybe I'm missing something, but I normally create a new local branch and hard reset it to the PR. Also, doesn't StyleCop enforce strict rules and prevent trailing whitespace at the end of lines?
How should I configure VS to prevent this from happening? I thought the `.editorconfig` would ensure consistency?

src/NzbDrone.Host/Bootstrap.cs
```diff
@@ -130,6 +79,64 @@ public static void Start(string[] args, Action<IHostBuilder> trayCallback = null
...
+                    services.Configure<AppOptions>(config.GetSection("Sonarr:App"));
+                    services.Configure<AuthOptions>(config.GetSection("Sonarr:Auth"));
+                    services.Configure<ServerOptions>(config.GetSection("Sonarr:Server"));
+                    services.Configure<LogOptions>(config.GetSection("Sonarr:Log"));
+                    services.Configure<UpdateOptions>(config.GetSection("Sonarr:Update"));
+                })
+                .Build();
+        }
+
+        private static void RunApplicationWithRestarts(string[] args, StartupContext startupContext, ApplicationModes appMode, Action<IHostBuilder> trayCallback)
```
markus101: I think this could use a better name, plural `Restarts` seems like a typo

src/NzbDrone.Host/Bootstrap.cs
```diff
@@ -130,6 +79,64 @@ public static void Start(string[] args, Action<IHostBuilder> trayCallback = null
...
+        {
+            Logger.Debug(appMode == ApplicationModes.Service ? "Service mode" : trayCallback != null ? "Tray mode" : "Console mode");
```
markus101: Please don't nest the ternary, also this message doesn't seem to add a lot of value, why are we just logging the mode?

src/NzbDrone.Host/Bootstrap.cs
```diff
@@ -130,6 +79,64 @@ public static void Start(string[] args, Action<IHostBuilder> trayCallback = null
...
+            do
+            {
+                var builder = CreateConsoleHostBuilder(args, startupContext);
+                trayCallback?.Invoke(builder);
+
+                shouldRestart = RunWithRestartCheck(builder.Build());
+
+                if (shouldRestart)
+                {
+                    Logger.Info("Restarting now");
```
markus101: Similar thought with the logging here too, more information as to why we're restarting would be helpful.

src/NzbDrone.Host/RestartableServiceLifetime.cs
```diff
@@ -0,0 +1,247 @@
...
+using Microsoft.Extensions.Options;
+using NLog;
+using NzbDrone.Common.Instrumentation;
+
+namespace NzbDrone.Host
+{
+    /// <summary>
+    /// Custom IHostLifetime for Windows Services with in-process restart support.
+    /// </summary>
+    public sealed class RestartableServiceLifetime : IHostLifetime, IDisposable
```
markus101: No need for the comment or sealing the class

```suggestion
    public class RestartableServiceLifetime : IHostLifetime, IDisposable
```

src/NzbDrone.Host/RestartableServiceLifetime.cs
```diff
@@ -0,0 +1,247 @@
...
+    {
+        private static readonly Lock _staticLock = new();
```
markus101: This is the suggested naming from ReSharper/Rider:

```suggestion
        private static readonly Lock StaticLock = new();
```

src/NzbDrone.Host/RestartableServiceLifetime.cs
```diff
@@ -0,0 +1,247 @@
...
+            {
+                _stoppedRegistration.Dispose();
+                if (_singletonService?.CurrentLifetime != this)
+                {
+                    _hostStoppedEvent?.Dispose();
+                }
+            }
+        }
+
+        internal static void ResetForTesting()
```
markus101: Doesn't look like this is used.

src/NzbDrone.Host/RestartableServiceLifetime.cs
```diff
@@ -0,0 +1,247 @@
...
+        {
+            lock (_staticLock)
+            {
+                _singletonService?.Dispose();
+                _singletonService = null;
+                _serviceBaseRunCalled = false;
+            }
+        }
+
+        internal void OnServiceStart()
```
markus101: Can be private

src/NzbDrone.Host/RestartableServiceLifetime.cs
```diff
@@ -0,0 +1,247 @@
...
+            }
+            catch (Exception ex)
+            {
+                Logger.Error(ex, "Service start failed");
+                _waitForStartTask.TrySetException(ex);
+                throw;
+            }
+        }
+
+        internal void OnServiceStop()
```
markus101: Same here

src/NzbDrone.Host/RestartableServiceLifetime.cs
```diff
@@ -0,0 +1,247 @@
...
+                _applicationLifetime?.StopApplication();
+            }
+            catch (Exception ex)
+            {
+                Logger.Error(ex, "Error during service stop");
+                throw;
+            }
+        }
+
+        private void RegisterStoppedCallback() => _stoppedRegistration = _applicationLifetime.ApplicationStopped.Register(
```
markus101: The assignment via arrow function is clunkier to read.

```suggestion
        private void RegisterStoppedCallback()
        {
            _stoppedRegistration = _applicationLifetime.ApplicationStopped.Register(
                state => ((RestartableServiceLifetime)state)._hostStoppedEvent.Set(), this);
        }
```

src/NzbDrone.Host/RestartableServiceLifetime.cs
```diff
@@ -0,0 +1,247 @@
...
+                state => ((RestartableServiceLifetime)state)._hostStoppedEvent.Set(), this);
+
+        private sealed class WindowsServiceWrapper : ServiceBase
+        {
+            public RestartableServiceLifetime CurrentLifetime { get; private set; }
+
+            public WindowsServiceWrapper(string serviceName) => ServiceName = serviceName;
```
markus101: Same here
```suggestion
            public WindowsServiceWrapper(string serviceName)
            {
                ServiceName = serviceName;
            }
```

src/NzbDrone.Host/RestartableServiceLifetime.cs
```diff
@@ -0,0 +1,247 @@
...
+
+            public void PrepareForRestart(RestartableServiceLifetime newLifetime, IHostApplicationLifetime newApplicationLifetime)
+            {
+                CurrentLifetime?._stoppedRegistration.Dispose();
+
+                CurrentLifetime = newLifetime;
+                CurrentLifetime._applicationLifetime = newApplicationLifetime;
+
+                CurrentLifetime._hostStoppedEvent.Reset();
+                CurrentLifetime._waitForStartTask.TrySetResult(null);
```
markus101: While this is functional it seems really weird to access private properties this way and should be renamed and their modifiers updated to match.

src/NzbDrone.Host/Bootstrap.cs
TypNull: No clue what you mean with why it restarted, but I changed the log message.

src/NzbDrone.Host/RestartableServiceLifetime.cs
TypNull: It was there for if you want to add a unit test, but it seems not like this, so I removed it now.

src/NzbDrone.Host/Bootstrap.cs
TypNull: Why should that be a typo? But I renamed it and hope it's more to your liking now.

src/NzbDrone.Host/Bootstrap.cs
TypNull: The logging was there before too, just merged it. I now use the enum name hope this is better.

src/NzbDrone.Host/RestartableServiceLifetime.cs
TypNull: You are right I split it.

TypNull:
> If I restart from within the app, what's the expected behaviour? Internally it restarts, but the process itself doesn't (same PID) or the whole process should restart?
> 
> I tested on macOS and got the following error:
> 
>
> 2025-12-20 16:09:25.8|Error|TaskExtensions|Task Error
> .......

Good to hear it works on mac too.

Yes, the current behavior is correct and expected. When you restart from within the app:
- The process ID (PID) stays the same
- It's an internal restart, not spawning a new process
- The application rebuilds the host and DI container within the same process

### About the Error

This is not a bug, it's just informational noise. Here's what's happening:

The error occurs because the EventAggregator uses a "fire and forget" pattern for async event handlers. When you trigger a restart:
1. Background event handlers are still processing
2. The host/container gets disposed as part of the restart cycle
3. Any still-running handlers that try to resolve services hit the disposed container and fail

Previously, these background tasks would just be terminated abruptly during restart. Now they attempt to complete, and if they fail (like this), they log the error and move on.

Since this doesn't actually break anything, I've left it as-is, the tasks just hard fail on restart. However, if you'd prefer cleaner shutdown behavior, you have two options:

1. Implement graceful shutdown in EventAggregator, Add a check to stop accepting/processing events when `ApplicationShutdownRequested` is fired
2. Wait for event queue completion, Make the app wait for all pending async event handlers to complete before shutdown

markus101:
Thanks for the details. The error makes sense, I do think it'd be good to handle it more gracefully, especially since it's a rather scary looking error. From a safety perspective waiting for all the events to fire is probably the best, but if that's significantly more effort I'd be fine with maintaining more or less the current behaviour and just prevent more from firing to avoid the error.

TypNull:
I’ll give it some thought after Christmas, since right now I don’t have enough time and understanding of the event handling system to figure out a way to wait on all fire-and-forget operations.
I did consider an idea for pausable ScheduledTasks, so if what I have in mind for that could work here as well, that would be neat. But probably not.
For now though, I simply wish you a Merry Christmas.

markus101:
Merry Christmas to you as well!

Pausable tasks would probably satisfy https://github.com/Sonarr/Sonarr/issues/1194 which has been around forever, but probably worthy of another PR to avoid growing the scope too much here. We can discuss more after the holidays and figure out a solution thouugh!

mynameisbogdan:
Works fine while there are no tasks running, but not while you're running a longer task. 

```
[Error] CommandExecutor: Error occurred while executing task RefreshSeries 

[v10.0.0.28493] DryIoc.ContainerException: code: Error.ContainerIsDisposed;
message: Container is disposed and should not be used: "container without scope is disposed!
 You may include Dispose stack-trace into the message via:
container.With(rules => rules.WithCaptureContainerDisposeStackTrace())"
   at DryIoc.Throw.It(Int32 error, Object arg0, Object arg1, Object arg2, Object arg3) in /_/src/DryIoc/Container.cs:line 14774
   at DryIoc.Container.ResolveAndCache(Int32 serviceTypeHash, Type serviceType, IfUnresolved ifUnresolved) in /_/src/DryIoc/Container.cs:line 397
   at DryIoc.Container.System.IServiceProvider.GetService(Type serviceType) in /_/src/DryIoc/Container.cs:line 344
   at Microsoft.Extensions.DependencyInjection.ServiceProviderServiceExtensions.GetRequiredService[T](IServiceProvider provider)
   at NzbDrone.Common.ServiceFactory.BuildAll[T]() in ./NzbDrone.Common/ServiceFactory.cs:line 36
   at NzbDrone.Core.Messaging.Events.EventAggregator.EventSubscribers`1..ctor(IServiceFactory serviceFactory) in ./NzbDrone.Core/Messaging/Events/EventAggregator.cs:line 31
   at NzbDrone.Core.Messaging.Events.EventAggregator.PublishEvent[TEvent](TEvent event) in ./NzbDrone.Core/Messaging/Events/EventAggregator.cs:line 89
   at NzbDrone.Core.Messaging.Commands.CommandExecutor.ExecuteCommand[TCommand](TCommand command, CommandModel commandModel) in ./NzbDrone.Core/Messaging/Commands/CommandExecutor.cs:line 104
   at System.Dynamic.UpdateDelegates.UpdateAndExecuteVoid3[T0,T1,T2](CallSite site, T0 arg0, T1 arg1, T2 arg2)
   at NzbDrone.Core.Messaging.Commands.CommandExecutor.ExecuteCommands() in ./NzbDrone.Core/Messaging/Commands/CommandExecutor.cs:line 42
```

src/NzbDrone.Core/Messaging/Events/EventAggregator.cs
```diff
@@ -53,6 +56,12 @@ public void PublishEvent<TEvent>(TEvent @event)
 
             var eventName = GetEventName(@event.GetType());
 
+            if (_runtimeInfo.IsExiting && Attribute.IsDefined(@event.GetType(), typeof(LifecycleEventAttribute)))
+            {
+                _logger.Debug("Event {0} blocked due to application shutdown", eventName);
```
mynameisbogdan: Might be more suitable for a warning, due to it's nature.
```suggestion
                _logger.Warn("Event {0} blocked due to application shutdown", eventName);
```

src/NzbDrone.Core/Messaging/Events/EventAggregator.cs
```diff
@@ -53,6 +56,12 @@ public void PublishEvent<TEvent>(TEvent @event)
 
             var eventName = GetEventName(@event.GetType());
 
+            if (_runtimeInfo.IsExiting && Attribute.IsDefined(@event.GetType(), typeof(LifecycleEventAttribute)))
```
mynameisbogdan: `Attribute.IsDefined(@event.GetType(), typeof(LifecycleEventAttribute)))` is just to check if the attribute exists on the event?

We already an extension for `HasAttribute` aka `@event.GetType().HasAttribute<LifecycleEventAttribute>()`, maybe it's more useful to you here. 

One big difference that comes to mind is that HasAttribute checks for inherited attributes while Attribute.IsDefined doesn't.

TypNull:
> Works fine while there are no tasks running, but not while you're running a longer task.

I’m still working through the best solution. Right now, the easiest fix is to suppress and log the error, but the real improvement would be to add more async support to Sonarr, ideally a lot more. That way, tasks could be canceled or awaited properly. The current `TaskFactory` feels outdated, stuck in the .NET 4.5 and the same goes for all I/O operations, where async would prevent blocking.

The challenge is that simply adding `Task.WaitAll` could cause restart issues. A better approach might be a three-tiered event system (Sync/Async/Background) and adding a `MaxConcurrency` for background tasks. Still, rewriting all I/O-bound work to be async is a huge job, even though it would make everything more modern and efficient. I'm just unsure.

src/NzbDrone.Core/Messaging/Events/EventAggregator.cs
TypNull: Ah, got it, I’ll use `HasAttribute`. Didn’t realize that was an option. I don’t need inheritance but it’s better to stay consistent.

mynameisbogdan:
> I’m still working through the best solution. Right now, the easiest fix is to suppress and log the error, but the real improvement would be to add more async support to Sonarr, ideally a lot more.

Yeah, I agree with you with adding more async support but sadly that's a lot of work refactoring.

PR #8253, opened 2025-12-14
fix(connection): use ISO 8601 format specifier for Discord timestamps

TheColorman:
#### Description
If Sonarr is running on a system where the local DateTime format does not use colons (:) as the delimiter for time, the Discord webhook notification does not work, and are rejected by Discord with the cryptic `{"embeds": ["0"]}` error. This is because colons (:) have a special meaning when used within DateTime.ToString format strings in .NET, being converted to the time delimiter of the CurrentCulture, which may be a period. This causes the resulting strings to not follow the ISO 8601 specification, causing Discord to reject them.

An example of such a locale is da-DK. An interactive example of this problem can be viewed here: https://dotnetfiddle.net/NY2vrG

This is fixed by using `.ToString("O")`, which is a format specifier designed for the ISO 8601 standard.

- The `":"` format specifier is documented here: https://learn.microsoft.com/en-us/dotnet/standard/base-types/custom-date-and-time-format-strings#timeSeparator
- The `"O"` format specifier is documented here: https://learn.microsoft.com/en-us/dotnet/standard/base-types/standard-date-and-time-format-strings#Roundtrip

No open issue as far as I can tell.

### DIFF
--- a/src/NzbDrone.Core/Notifications/Discord/Discord.cs
+++ b/src/NzbDrone.Core/Notifications/Discord/Discord.cs
@@ -48,7 +48,7 @@ namespace NzbDrone.Core.Notifications.Discord
                 Title = GetTitle(series, episodes),
                 Color = (int)DiscordColors.Standard,
                 Fields = new List<DiscordField>(),
-                Timestamp = DateTime.UtcNow.ToString("yyyy-MM-ddTHH:mm:ss.fffZ")
+                Timestamp = DateTime.UtcNow.ToString("O")
             };
 
             if (Settings.GrabFields.Contains((int)DiscordGrabFieldType.Poster))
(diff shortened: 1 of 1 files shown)


mynameisbogdan:
I remember this issue, and of course it's because of the localization. 😂

TheColorman:
> I remember this issue, and of course it's because of the localization. 😂

Was certainly annoying to debug this :sweat_smile:

stevietv:
Nice catch!
I'm wondering, given that in all the notifications we only ever set Timestamp to be Now() is it really necessary to add the timestamp at all? According to discord [docs ](https://discord.com/developers/docs/resources/message#embed-object-embed-structure)its not a required field

I guess if we don't add a timestamp then it won't appear at the bottom of the embed, but surely the date and time of the message already says that?!

TheColorman:
It's entirely for aesthetics, but I personally do like it. When many messages get grouped because they're sent close together, the time is only visible on the first message. Having it at the bottom of every embed is both useful, and looks nice.

stevietv:
> It's entirely for aesthetics, but I personally do like it. When many messages get grouped because they're sent close together, the time is only visible on the first message. Having it at the bottom of every embed is both useful, and looks nice.

yeah makes perfect sense!

markus101 (approved):
Thanks!

TheColorman:
Is there a way to backport this to v4? Should I have targeted `develop` instead of `v5-develop`? I've been compiling a patched version of v4 that includes this change for the past 8 months and it feels like a bit of a waste...

markus101:
This is in develop now as well.

PR #8292, opened 2025-12-27
A bit of async

TypNull:
#### Description

This pull request begins the async migration process for Sonarr. Currently, it shifts blocking operations from external libraries into the Sonarr source code. With the transition from .NET 6 to .NET 10, incorporating more async programming patterns will make Sonarr more modern, responsive, and resource-efficient.

**Current scope:**
- **Event Aggregator**: Migrated from a fire-and-forget approach to a channel-based implementation
- **Repository Layer**: Migrated to use async operations, with blocking shifted to the service layer for database operations

**Future work:**
Migrating the service layer would be the next logical step, though this represents a substantial undertaking. Best practice would be using async operations throughout the entire stack from controllers down to disk writes and HTTP calls.

**Review needed:**
Before proceeding further, I'm requesting a review to ensure this approach aligns with the project's direction. Given the significant time investment required for the next migration, I want to confirm this is the desired path forward and avoid potential rework if the repository layer requires a different approach or if this will not be merged after the service transition, as I don't want to continuously rebase this PR.

### DIFF
--- a/src/NzbDrone.Core.Test/AutoTagging/AutoTaggingServiceFixture.cs
+++ b/src/NzbDrone.Core.Test/AutoTagging/AutoTaggingServiceFixture.cs
@@ -2,6 +2,7 @@ using System.Collections.Generic;
 using System.Linq;
 using FizzWare.NBuilder;
 using FluentAssertions;
+using Moq;
 using NUnit.Framework;
 using NzbDrone.Core.AutoTagging;
 using NzbDrone.Core.AutoTagging.Specifications;
@@ -24,9 +25,9 @@ namespace NzbDrone.Core.Test.AutoTagging
                                      .Build();
 
             _tag = new AutoTag
-                           {
-                               Name = "Test",
-                               Specifications = new List<IAutoTaggingSpecification>
+            {
+                Name = "Test",
+                Specifications = new List<IAutoTaggingSpecification>
                                                 {
                                                     new GenreSpecification
                                                     {
@@ -37,16 +38,16 @@ namespace NzbDrone.Core.Test.AutoTagging
                                                                 }
                                                     }
                                                 },
-                               Tags = new HashSet<int> { 1 },
-                               RemoveTagsAutomatically = false
-                           };
+                Tags = new HashSet<int> { 1 },
+                RemoveTagsAutomatically = false
+            };
         }
 
         private void GivenAutoTags(List<AutoTag> autoTags)
         {
             Mocker.GetMock<IAutoTaggingRepository>()
-                  .Setup(s => s.All())
-                  .Returns(autoTags);
+                  .Setup(s => s.AllAsync())
+                  .ReturnsAsync(autoTags.AsEnumerable());
         }
 
         [Test]
@@ -107,10 +108,10 @@ namespace NzbDrone.Core.Test.AutoTagging
         public void should_have_tags_to_add_if_series_does_not_have_match_tag_and_series_matches_all_rules()
         {
             _tag.Specifications.Add(new SeriesTypeSpecification
-                                    {
-                                        Name = "Series Type",
-                                        Value = (int)_series.SeriesType
-                                    });
+            {
+                Name = "Series Type",
+                Value = (int)_series.SeriesType
+            });
 
             GivenAutoTags(new List<AutoTag> { _tag });
 
--- a/src/NzbDrone.Core.Test/Blocklisting/BlocklistRepositoryFixture.cs
+++ b/src/NzbDrone.Core.Test/Blocklisting/BlocklistRepositoryFixture.cs
@@ -24,14 +24,14 @@ namespace NzbDrone.Core.Test.Blocklisting
         public void Setup()
         {
             _blocklist = new Blocklist
-                     {
-                         SeriesId = 12345,
-                         EpisodeIds = new List<int> { 1 },
-                         Quality = new QualityModel(Quality.Bluray720p),
-                         Languages = new List<Language> { Language.English },
-                         SourceTitle = "series.title.s01e01",
-                         Date = DateTime.UtcNow
-                     };
+            {
+                SeriesId = 12345,
+                EpisodeIds = new List<int> { 1 },
+                Quality = new QualityModel(Quality.Bluray720p),
+                Languages = new List<Language> { Language.English },
+                SourceTitle = "series.title.s01e01",
+                Date = DateTime.UtcNow
+            };
 
             _series1 = Builder<Series>.CreateNew()
                                       .With(s => s.Id = 7)
@@ -46,7 +46,8 @@ namespace NzbDrone.Core.Test.Blocklisting
         public async Task should_be_able_to_write_to_database()
         {
             await Subject.InsertAsync(_blocklist);
-            (await Subject.AllAsync()).Should().HaveCount(1);
+            var enumerable = await Subject.AllAsync();
+            enumerable.Should().HaveCount(1);
         }
 
         [Test]
@@ -54,7 +55,8 @@ namespace NzbDrone.Core.Test.Blocklisting
         {
             await Subject.InsertAsync(_blocklist);
 
-            (await Subject.AllAsync()).First().EpisodeIds.Should().Contain(_blocklist.EpisodeIds);
+            var enumerable = await Subject.AllAsync();
+            enumerable.First().EpisodeIds.Should().Contain(_blocklist.EpisodeIds);
         }
 
         [Test]
@@ -62,7 +64,8 @@ namespace NzbDrone.Core.Test.Blocklisting
         {
             await Subject.InsertAsync(_blocklist);
 
-            (await Subject.BlocklistedByTitleAsync(_blocklist.SeriesId, _blocklist.SourceTitle.ToUpperInvariant())).Should().HaveCount(1);
+            var blocklists = await Subject.BlocklistedByTitleAsync(_blocklist.SeriesId, _blocklist.SourceTitle.ToUpperInvariant());
+            blocklists.Should().HaveCount(1);
         }
 
         [Test]
@@ -79,7 +82,7 @@ namespace NzbDrone.Core.Test.Blocklisting
                 .With(c => c.EpisodeIds = new List<int> { 1 })
                 .BuildListOfNew();
 
-            Db.InsertMany(blocklistItems);
+            await Db.InsertManyAsync(blocklistItems);
 
             await Subject.DeleteForSeriesIdsAsync(new List<int> { _series1.Id });
 
--- a/src/NzbDrone.Core.Test/Blocklisting/BlocklistServiceFixture.cs
+++ b/src/NzbDrone.Core.Test/Blocklisting/BlocklistServiceFixture.cs
@@ -42,7 +42,7 @@ namespace NzbDrone.Core.Test.Blocklisting
             await Subject.HandleAsync(_event, CancellationToken.None);
 
             Mocker.GetMock<IBlocklistRepository>()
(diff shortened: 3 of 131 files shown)


src/NzbDrone.Core/Datastore/Extensions/SqlMapperExtensions.cs
```diff
@@ -1,20 +1,26 @@
...
     {
-        public static IEnumerable<T> Query<T>(this IDatabase db, string sql, object param = null)
+        public static async Task<IEnumerable<T>> QueryAsync<T>(
+            this IDatabase db,
+            string sql,
+            object param = null,
+            CancellationToken cancellationToken = default)
         {
-            using (var conn = db.OpenConnection())
+            using (var conn = await db.OpenConnectionAsync(cancellationToken).ConfigureAwait(false))
```
mynameisbogdan, line 18: Sadly Dapper sucks at making use of the cancellation token for queries, so `ThrowIfCancellationRequested` should be used in these circumstances which is the perfect use-case for it IMO.

The using must await-ed too. If you aren't already, you should switch to Rider as it's showing suggestions for these cases.

```suggestion
            cancellationToken.ThrowIfCancellationRequested();

            await using var conn = await db.OpenConnectionAsync(cancellationToken).ConfigureAwait(false);
```

mynameisbogdan (changes requested):

mynameisbogdan:
Overall .NET 10 has a lot of improvements for async, like ToListAsync/FirstOrDefaultAsync/SingleOrDefaultAsync but they apply to IAsyncEnumerable.

Irony enough, `QueryUnbufferedAsync` is sync with some async magic under the hood.

Here's an example with IAsyncEnumerable:
```csharp
public static async Task<IAsyncEnumerable<T>> QueryAsync<T>(this IDatabase db, string sql, object param = null, CancellationToken cancellationToken = default)
{
    cancellationToken.ThrowIfCancellationRequested();

    await using var conn = await db.OpenConnectionAsync(cancellationToken).ConfigureAwait(false);

    IAsyncEnumerable<T> items;
    try
    {
        items = conn.QueryUnbufferedAsync<T>(sql, param);
    }
    catch (Exception e)
    {
        e.Data.Add("SQL", SqlBuilderExtensions.GetSqlLogString(sql, param));
        throw;
    }

    if (TableMapping.Mapper.LazyLoadList.TryGetValue(typeof(T), out var lazyProperties))
    {
        await foreach (var item in items)
        {
            ApplyLazyLoad(db, item, lazyProperties);
        }
    }

    return items;
}
```

Overall, these changes should be made with care as there's a high risk of breaking something.

src/NzbDrone.Core/Datastore/Extensions/SqlMapperExtensions.cs
TypNull: Sure, I’ll add the cancellation check. I’m deeply used to VS, it’s just where I’m most comfortable. When I’ve tried Rider on Linux, it always felt a bit off, and my Git workflow isn’t as smooth there. I’ll at least consider giving it another look, though.

But in general, is this the right direction? As when I prepare to transition the services, I want to avoid constantly rebasing onto all the ongoing changes. For example, the mini profiler is also moving from `IDbConnection` to `DbConnection` isn’t a big deal since it’s a small change. However, it’s a different story with the services they’re much more involved. I would like it to be merged in smaller PR's instead of one large PR.

TypNull:
I understand why these changes need to be handled carefully, but I'm certain something will break regardless. That's exactly why I believe the v5 rewrite is the ideal timeframe to tackle this.
Neither I nor anyone else can write perfect code, as much as we'd all like to. This is clearly a task that requires more than just my own review. Even with additional oversight, there will likely still be issues that slip through.

src/NzbDrone.Core/Datastore/Extensions/SqlMapperExtensions.cs
mynameisbogdan: Sadly Rider is more annoying at making this kind of suggestions. 

Personally I consider these changes too big of a task, and should be instead in increments having sync and async methods available. Like I implemented async support for HttpClient just to add HTTP2 support, sync just calls the async due to the limitation.

Taking EventAggregator for example, you have one big issue as you have both Handle and HandleAsync, but isn't really async. And the TaskFactory without the Wait() makes me thing it's actually an attempt to make it a background handle actually. 

I noticed you added `IHandleBackgroundAsync` to address this, but at the moment I can't think of any suggestions in order to handle sync and real async handlers, maybe we should address it in the caller like wolverine does, PublishAsync doesn't wait since it's processed in the background, while InvokeAsync waits.

FYI in NotificationService the behavior is changed when you made all of them `IHandleBackgroundAsync`, while it builds Rider is throwing errors:
Screenshot 2025-12-27 at 17 35 46

markus101:
> That's exactly why I believe the v5 rewrite is the ideal timeframe to tackle this.

v5 isn't meant as a rewrite for all the things, it's a .net upgrade, converting the front end to more modern tech and adding some long asked for features, even though it's still in progress it's getting near the end and I think layering on a massive backend change at the same time and in a single PR is just flying in the face of failure.

I'd love to see things converted to async, but it needs to be done piecemeal, much like the API v5 and UI side of things has been doing. At over 7000 total changes this is realistically something that won't get reviewed properly to catch even the most obvious issues. I'm not sure what makes sense of async to be part of v5, since it's getting pretty late in the effort to add on a large refactor and I'd like to see it get out to alpha/beta as soon as the UI migration is complete (save for some things awaiting those).

TypNull:
I don't think Rider is more aggressive with suggestions than VS. My Visual Studio instance is probably just overworked with the whole stack. I can try switching to Rider to address the remaining build issues.

I'm not sure what you mean by referencing Wolverine. I could rename the interfaces, but they're designed exactly as they should be. `TaskFactory` should only be used for long-running tasks that aren't background work, and everything else should be done with `Task.Run`. The background handling is exactly that, a limited parallel system that still holds references to all events to eventually cancel them with Channel plus `Parallel.Async`, even if 50 is just a magic number right now. The event system is just a bit of a showcase how I would imagine it.

For the notification service, I made it this way intentionally because it's not a critical system. It can run completely in the background without waiting. If the build errors are a problem, I can address them when doing a PR. A lot of methods already use `GetAwaiter().GetResult()` like `IProviderStatusServiceBase.RecordSuccess` after the Repository changes. I opened this as a draft just to test reactions, not as a PR.

I know this is a big change, but you have to start somewhere. I never said I'd do it as one big PR with multiple commits. If I had to do it as one large PR, I'd stop here. Though I'd probably still replace all database repository operations in one PR since they're closely related.

I understand your concern about the timing and scope, but I disagree about waiting until after v5. The .NET upgrade is exactly why this makes sense now. A lot of methods getting old even if an async migration wouldn't help with them. It's easier to implement modern features on a modern foundation. You're hopping from .NET 6 to 10, but the code is still the same. I get you want to get it into alpha/beta and the UI was a big focus, but if there's no official beta yet, now seems like the right time. The changes sound massive, but most of that is mechanical conversion. The actual logic changes are concentrated in the database layer (and the event system, which acts as its own isolated playground.) 

This is the window where breaking changes are acceptable and expected, but if you're already planning v6 for the backend migration, that makes sense too. If this isn't something you want to tackle now, I'm fine with that. It's a lot of work and even these few parts took considerable effort. If you have a better approach in mind, I'll just go with that instead. It's not like I'm looking forward to a huge workload like this anyway when it mainly needs time and effort.

mynameisbogdan:
> I don't think Rider is more aggressive with suggestions than VS. My Visual Studio instance is probably just overworked with the whole stack. I can try switching to Rider to address the remaining build issues.

I haven't used VS to have something to compare, as I found it awful. But Rider shows suggestions for missing awaits, etc.

> I'm not sure what you mean by referencing Wolverine. I could rename the interfaces, but they're designed exactly as they should be. `TaskFactory` should only be used for long-running tasks that aren't background work, and everything else should be done with `Task.Run`. The background handling is exactly that, a limited parallel system that still holds references to all events to eventually cancel them with Channel plus `Parallel.Async`, even if 50 is just a magic number right now. The event system is just a bit of a showcase how I would imagine it.

[Wolverine](https://github.com/JasperFx/wolverine) is basically EventAggregator [done right with a lot of other overkill features](https://wolverinefx.net/guide/handlers). 

Regarding parallelism with batches of 50, I don't think it's quite a good idea with sqlite if multiple events are doing writes in the same time IMO. 

> For the notification service, I made it this way intentionally because it's not a critical system. It can run completely in the background without waiting. If the build errors are a problem, I can address them when doing a PR. A lot of methods already use `GetAwaiter().GetResult()` like `IProviderStatusServiceBase.RecordSuccess` after the Repository changes. I opened this as a draft just to test reactions, not as a PR.

Ah, you're correct, notifications should be done in background 100%. I just thought it was a mistake as I was looking for feature parity, my bad. 

The issue with those errors is that dotnet build won't throw errors, only my IDE seems to show them.

> I understand your concern about the timing and scope, but I disagree about waiting until after v5. The .NET upgrade is exactly why this makes sense now. A lot of methods getting old even if an async migration wouldn't help with them. It's easier to implement modern features on a modern foundation. You're hopping from .NET 6 to 10, but the code is still the same. I get you want to get it into alpha/beta and the UI was a big focus, but if there's no official beta yet, now seems like the right time. The changes sound massive, but most of that is mechanical conversion. The actual logic changes are concentrated in the database layer (and the event system, which acts as its own isolated playground.)
> 
> This is the window where breaking changes are acceptable and expected, but if you're already planning v6 for the backend migration, that makes sense too. If this isn't something you want to tackle now, I'm fine with that. It's a lot of work and even these few parts took considerable effort. If you have a better approach in mind, I'll just go with that instead. It's not like I'm looking forward to a huge workload like this anyway when it mainly needs time and effort.

But changing sync to async should not be a breaking change, changing the behavior would be a breaking change and we don't want behavior changes. Adding async support could and should be done even in a patch version IMO. 

The main issue would be if something is broken by mistake and it's undetected till stable hits.

Fun fact, async HttpClient was a little faster (due to concurrency I assume) and right after sqlite locking issues started to arise and we had to raise the busy timeout. 😄

markus101:
> I know this is a big change, but you have to start somewhere. I never said I'd do it as one big PR with multiple commits. If I had to do it as one large PR, I'd stop here.

Is the plan to go separate this PR into several smaller PRs then? Because as-is it's not really feasible to review and more importantly test all the things that are changes.

> Though I'd probably still replace all database repository operations in one PR since they're closely related.

I disagree for two reasons, it's a review nightmare and it's too much surface area to test and if there are issues diagnose where something went wrong. Separating things out to sections of the app makes a lot more sense for both those points and allows things to be tested in smaller chunks and when there are issues you can pinpoint them like B did here: https://github.com/Sonarr/Sonarr/pull/8276

> You're hopping from .NET 6 to 10, but the code is still the same.

Sure, but we're getting new features, security patches and the like. The v5 API project is taking advantage of some new things, but we're not going to implement everything everywhere all at once.

> I get you want to get it into alpha/beta and the UI was a big focus, but if there's no official beta yet, now seems like the right time.

Sure, but that's an immense amount of scope being added to an already large release and trying to do too much at the same time is how we'll not get any of it finished. I personally don't want have the bandwidth to commit to bringing the backend into the async world in addition to everything else I've committed to before v5 hit beta. As much as I'd like to just get it done I think this makes more sense for v6 or as B mentioned throughout the lifecycle of v5, though there are some risks to that of course.

TypNull:
Thank you for your feedback. I’ll hold off until v6 or until you need assistance. While I still believe now is a good time for this, I don’t have many items in Sonarr, so there’s no urgency. Lidarr, however, would likely benefit more from async support.

mynameisbogdan:
Sadly Lidarr is still far away from reaching parity with upstream anyway.

PR #8298, opened 2025-12-29
v5 UI Settings

markus101:
#### Description

Moves advanced settings over to a zustand store and converts UI settings over to React Query. v5 API for UI settings is now `settings/ui` instead of `config/ui` that was in v3.


frontend/src/Settings/UI/useUiSettings.ts
```diff
@@ -0,0 +1,54 @@
...
+  timeZone: string;
+  firstDayOfWeek: number;
+  enableColorImpairedMode: boolean;
+  calendarWeekColumnHeader: string;
+  uiLanguage: number;
+}
+
+const PATH = '/settings/ui';
+
+export const useUiSettingsValue = () => {
```
mynameisbogdan: Would Values be more suitable?

markus101: I was on the fence before, but I think it makes more sense.

PR #8316, opened 2026-01-02
Fixed: Login with credentials on Qbittorrent 5.2

mynameisbogdan:
#### Description
According to https://github.com/qbittorrent/qBittorrent/blob/9d5a8270cf8fb72775c6e5e1a96925a7c1ab2773/src/webui/webapplication.cpp#L648, now the API key is sent via the Authorization header which fails while we send the credentials as basic auth along with the form data.

The basic auth seems to be implement a decade ago with 0552b56 even after the split to the 2 proxies, so might be safe to remove with the new versions of the API.

* Closes https://github.com/Radarr/Radarr/issues/11339

### DIFF
--- a/src/NzbDrone.Core/Download/Clients/QBittorrent/QBittorrentProxyV2.cs
+++ b/src/NzbDrone.Core/Download/Clients/QBittorrent/QBittorrentProxyV2.cs
@@ -422,13 +422,12 @@ namespace NzbDrone.Core.Download.Clients.QBittorrent
                 _authCookieCache.Remove(authKey);
 
                 var authRequestBuilder = BuildRequest(settings);
-                authRequestBuilder.NetworkCredential = new BasicNetworkCredential(settings.Username, settings.Password);
 
                 var authLoginRequest = authRequestBuilder.Resource("/api/v2/auth/login")
-                                                             .Post()
-                                                             .AddFormParameter("username", settings.Username ?? string.Empty)
-                                                             .AddFormParameter("password", settings.Password ?? string.Empty)
-                                                             .Build();
+                    .Post()
+                    .AddFormParameter("username", settings.Username ?? string.Empty)
+                    .AddFormParameter("password", settings.Password ?? string.Empty)
+                    .Build();
 
                 HttpResponse response;
                 try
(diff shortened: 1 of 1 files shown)


src/NzbDrone.Core/Download/Clients/QBittorrent/QBittorrentProxyV2.cs
```diff
@@ -422,13 +422,12 @@ private void AuthenticateClient(HttpRequestBuilder requestBuilder, QBittorrentSe
                 _authCookieCache.Remove(authKey);
 
                 var authRequestBuilder = BuildRequest(settings);
-                authRequestBuilder.NetworkCredential = new BasicNetworkCredential(settings.Username, settings.Password);
```
markus101, line 425: This isn't necessary for older qbit versions?

mynameisbogdan: How old you mean? Because there's 2 proxies that split a while ago.

markus101: I mean does anything using the v2 proxy require the authentication sent as a basic auth header?

If it was only ever needed for versions that the v1 proxy handles then 👍 otherwise we should tread a little more carefully and only stop sending it when we're sure it's not used.

mynameisbogdan: I don't think so, since it should use the cookies. The basic auth was kept from a decade ago so v1 can still handle those. 

I'm check some older versions to be sure.

markus101: Ahh gotcha, that makes sense, I forgot it's all cookie auth now.

mynameisbogdan: Yea, I'm quite certain whatever uses v2 proxy is all cookies, the only reason now is broken is due to the use of the same header for api key. See #8317.

mynameisbogdan: https://redirect.github.com/qbittorrent/qBittorrent/issues/23270#issuecomment-3707142266

Confirmed by upstream, it never supported basic auth but there's a PR opened.

PR #8317, opened 2026-01-02
New: API key support for qBittorrent

### LINKED ISSUE #8282: QBittorrent: API Key Setting

When adding a download client, there is presently no way to add an API key for authentication purposes. This means that for download clients such as qBitTorrent,auto needs completely disabled within the subnet running the applications, either localhost or the Docker network which the containers are in. While this isn’t a huge deal per se, [qBitTorrent does now support API key authentication](https://github.com/qbittorrent/qBittorrent/issues/13238#issuecomment-3592720284). For enhanced security, users can now generate an API key and re-enable authentication within the subnet, provided that tools such as Sonarr, etc. enable this API key auth on their end!

When configuring the download client, there should an optional and advanced configuration for the API key for the client itself. Then this API key should be sent during communications with the download client as needed to provide authentication.

The alternative is to simply continue sending traffic without auth, though this means that all traffic from the application is essentially authenticated and valid.

### Anything else?

_No response_

mynameisbogdan:
#### Description
Adding support for API keys supported since qBittorrent 5.2.0.

* Closes #8282

### DIFF
--- a/src/NzbDrone.Core/Localization/Core/en.json
+++ b/src/NzbDrone.Core/Localization/Core/en.json
@@ -562,7 +562,7 @@
   "DownloadClientValidationApiKeyIncorrect": "API Key Incorrect",
   "DownloadClientValidationApiKeyRequired": "API Key Required",
   "DownloadClientValidationAuthenticationFailure": "Authentication Failure",
-  "DownloadClientValidationAuthenticationFailureDetail": "Please verify your username and password. Also verify if the host running {appName} isn't blocked from accessing {clientName} by WhiteList limitations in the {clientName} configuration.",
+  "DownloadClientValidationAuthenticationFailureDetail": "Please verify your credentials. Also verify if the host running {appName} isn't blocked from accessing {clientName} by WhiteList limitations in the {clientName} configuration.",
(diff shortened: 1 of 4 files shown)


src/NzbDrone.Core/Localization/Core/en.json
```diff
@@ -562,6 +562,7 @@
   "DownloadClientValidationApiKeyIncorrect": "API Key Incorrect",
   "DownloadClientValidationApiKeyRequired": "API Key Required",
   "DownloadClientValidationAuthenticationFailure": "Authentication Failure",
+  "DownloadClientValidationAuthenticationFailureApiKeyDetail": "Please verify your API key. Also verify if the host running {appName} isn't blocked from accessing {clientName} by WhiteList limitations in the {clientName} configuration.",
   "DownloadClientValidationAuthenticationFailureDetail": "Please verify your username and password. Also verify if the host running {appName} isn't blocked from accessing {clientName} by WhiteList limitations in the {clientName} configuration.",
```
mynameisbogdan: Here I'm inclined between this one, or changing DownloadClientValidationAuthenticationFailureDetail to `Please verify your credentials.`

PR #8318, opened 2026-01-02
v5 Connections

markus101:
#### Description

v5 Notifications (Connections) and UI changes with React Query. Also a small fix for folder validation I ran into when testing Custom Script with an empty path.

Maybe time to rename Notifications to Connections in the API?


stevietv:
> Maybe time to rename Notifications to Connections in the API?

I'd agree, we've called it connect in the UI for a long time

PR #8321, opened 2026-01-03
Bump BusyTimeout for SQLite to 1000ms

mynameisbogdan:
#### Description
Waiting 1s might be a saner value for slower setups like Synology where config is stored on the same medium as the downloads.


src/NzbDrone.Core/Datastore/ConnectionStringFactory.cs
```diff
@@ -64,7 +64,7 @@ private static DatabaseConnectionInfo GetConnectionString(string dbPath)
                 JournalMode = OsInfo.IsOsx ? SQLiteJournalModeEnum.Truncate : SQLiteJournalModeEnum.Wal,
                 Pooling = true,
                 Version = 3,
-                BusyTimeout = 100
+                BusyTimeout = 1000
```
markus101, line 67: The commit/PR mention 500ms, but this is 1000ms, which one did you intend it to be?

mynameisbogdan: 1000ms :D

PR #8326, opened 2026-01-04
Remove Custom Filters from fetch server collection

markus101:
#### Description

The last thing to use Server Side collections is Import List Exclusions, which looks for Custom Filters which no longer exist in Redux state. Import list exclusions don't have custom filters so nothing of value was lost and will get ported to React Query soon enough.


frontend/src/Store/Actions/Creators/createFetchServerSideCollectionHandler.js
mynameisbogdan:
```suggestion
```

Issue #733, opened 2015-08-07
Prioritize Download Decisions


waldnercharles:
In `PrioritizeDecisions` inside of the `DownloadDecisionPriorizationService `, number of seeds has a lower priority than number of episodes & average episode size. Shouldn't this be the other way around? Or weighted?

I'm also not sure why we order by the lowest episode number either. Correct me if I'm wrong, but won't this always force a season pack to have higher priority?

Or even better yet, why not allow the user to decide the order of importance? Something like:
-- Priority
:: # Of Seeds ::
:: Season Packs     ::
:: Average Episode Size ::
:: Age of Torrent ::

(Those :: are handles for users to drag to specify the order.)

markus101:
Yes, season packs are prioritized, I don't see that changing, though the option to exclude season packs entirely might be something we consider.

I don't see a customizable order being viable, its useful for 1% of the users, adds a lot of complexity and keeps a strict order, a weighted system would be more flexible and useful for usenet and torrents.

We've had some internal discussions about how we'd want to handle this scenario, though nothing concrete has been decided.

waldnercharles:
I guess my question is: If I am looking for s01e03 and i find a season pack with 1 seed, it will prefer the season pack over the single episode, even if that episode has 1,000 seeds. This seems incorrect to me.

Taloth:
Define 'looking'. Did you do an Automatic Search, or a Manual Search.
Did Sonarr actually grab that season pack?

waldnercharles:
Automatic Search. Even if it finds 2 single episode torrents, the one with the smaller size (given a 200mb difference) will be chosen over one with more seeds.

Taloth:
We were talking about it grabbing season packs over single eps. Not grabbing one single episode over another.

waldnercharles:
I'm confused. The logic used for both is the same isn't it? Contained within `DownloadDecisionPriorizationService.PrioritizeDecisions`.

Taloth:
Yes and No, the logic isn't limited to that function. That function merely prioritizes it, the rejection logic is performed elsewhere.

A season pack will only be grabbed during a season search when that season pack is an upgrade for all episodes in the season. Otherwise it will be rejected, regardless of prioritization.
It always takes precedence over single episodes for this very reason. Otherwise Sonarr might grab one  single episode which would cause the season pack to be rejected entirely, with you ending up with a single episode in your collection instead of all of them.

[...]

PR #8354, opened 2026-01-19
Add Seeders Preference setting for torrent prioritization

JustMaier:
#### Description
Adds a Seeders Preference setting to Indexer Options (advanced) that lets users control where seeder count is evaluated in the release comparison chain.

Three options:
- Default: existing behavior (seeders as low-priority tiebreaker)
- High: seeders evaluated after quality
- Highest: seeders evaluated before quality

I know similar requests have come up before (#733) and there were concerns about complexity. I tried to keep this simple with just three presets, but I'm happy to adjust the approach if you'd prefer something different.

If a fully customizable reorderable list for the comparison chain would be more useful, I'm open to reworking this. Just let me know what direction you'd prefer - this has been a pretty common request from folks in the torrent ecosystem and I'd like to find an implementation that works for the project.

* Related to #733


src/NzbDrone.Core/DecisionEngine/DownloadDecisionComparer.cs
```diff
@@ -27,18 +27,34 @@ public DownloadDecisionComparer(IConfigService configService, IDelayProfileServi
...
+                comparers.Add(ComparePeersIfTorrent);
+            }
+
+            comparers.Add(CompareCustomFormatScore);
+            comparers.Add(CompareProtocol);
+            comparers.Add(CompareEpisodeCount);
+            comparers.Add(CompareEpisodeNumber);
+            comparers.Add(CompareIndexerPriority);
+
+            if (seedersPreference == SeedersPreferenceType.Default)
```
mynameisbogdan, line 51: You could make use of https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.list-1.insert?view=net-10.0 simplifying all of this.

mynameisbogdan (changes requested):

src/NzbDrone.Core/Localization/Core/zh_CN.json
```diff
@@ -1665,6 +1665,11 @@
     "SecretToken": "密钥令牌",
     "Security": "安全",
     "Seeders": "种子",
+    "SeedersPreference": "种子偏好",
```
mynameisbogdan, line 1668: Non-en localization changes should be addressed in weblate as there's a high chance of conflicts when done in a PR and weblate has pending changes.

mynameisbogdan (changes requested):

markus101:
> If a fully customizable reorderable list for the comparison chain would be more useful, I'm open to reworking this.

I think this would make more sense than the mess that would certainly arise if we started doing this for multiple things at the same time. I'm not 100% sure that quality/custom format score should be beaten by anything though, so I think at least initially those should be in a fixed position at highest and second highest order respectively. Given you can group all qualities and CFs aren't mandatory you can still achieve seeders being more preferred, just not via a single setting.

PR #8363, opened 2026-01-26
Fix: Fix ffprobe call causing unnecessary disk reads

realzombee:
#### Description
- We currently pass the global stream index of the primary video stream with v:index.
- ffprobe expects the index to be filtered by the stream type. If the video appears at index 2 but is the only video stream in the file, we should pass v:0. Passing v:2 will cause ffprobe to unnecessarily read the entire file during analysis to find a video stream at index 2 which doesn't exist

#### Steps to reproduce bug
1. Find a video file where the video stream is at a non-zero global index, let's say it's at index 2.
2. Current behavior: Run `ffprobe -loglevel error -print_format json -show_frames -v verbose -sexagesimal -read_intervals %+#1 -select_streams v:2 <filename>`. Notice the total bytes read equals the size of the file.
4. Repeat with `v:0` assuming the file only has one video stream. The total bytes read will be a lot lower (<100KB usually)

### DIFF
--- a/src/NzbDrone.Core/MediaFiles/MediaInfo/VideoFileInfoReader.cs
+++ b/src/NzbDrone.Core/MediaFiles/MediaInfo/VideoFileInfoReader.cs
@@ -108,7 +108,9 @@ namespace NzbDrone.Core.MediaFiles.MediaInfo
                 // if it looks like PQ10 or similar HDR, do a frame analysis to figure out which type it is
                 if (PqTransferFunctions.Contains(mediaInfoModel.VideoTransferCharacteristics))
                 {
-                    frames = FFProbe.GetFrames(filename, customArguments: $"-read_intervals \"%+#1\" -select_streams v:{primaryVideoStream?.Index ?? 0}");
+                    var videoStreamIndex = analysis.VideoStreams.Count == 1 ? 0
+                        : analysis.VideoStreams.FindIndex(s => Equals(s?.Index, primaryVideoStream?.Index));
+                    frames = FFProbe.GetFrames(filename, customArguments: $"-read_intervals \"%+#1\" -select_streams v:{videoStreamIndex}");
                 }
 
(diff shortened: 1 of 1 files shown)


mynameisbogdan:
Did you checked this behaviour with the latest ffprobe static build version 8.0.1 or the previous 5.1.x available in Sonarr v4?

realzombee:
I've tested with the version available in Sonarr v4 as well as version 7.0

mynameisbogdan:
Can you test as well with sonarr v5-develop which comes with ffprobe 8.0.1 at the moment?

realzombee:
Just tested with 8.0.1 and a file containing this stream:
```
Stream #0:2: Video: hevc (Main 10), 1 reference frame, yuv420p10le(tv, bt2020nc/bt2020/smpte2084, topleft), 3840x1920 [SAR 1:1 DAR 2:1], 23.98 fps, 23.98 tbr, 1k tbn (default)
```

ffprobe stats:

Proposed change
```
ffprobe -loglevel error -print_format json -show_frames -v verbose -sexagesimal -read_intervals %+#1 -select_streams v:0
[AVIOContext @ 0x558226451940] Statistics: 66708 bytes read, 2 seeks
    ]
```

Existing behavior
```
ffprobe -loglevel error -print_format json -show_frames -v verbose -sexagesimal -read_intervals %+#1 -select_streams v:2
[AVIOContext @ 0xc64c100] Statistics: 4931248800 bytes read, 3 seeks
    ]
```

markus101:
Thanks for the fix and verifying it works with newer ffprobe versions.

realzombee:
Thanks, any chance this can get merged into v4 develop? I can submit another PR for that

markus101:
If there ends up being a need to backport a high priority fix this could be included, but otherwise no plans for that at this time.

PR #8401, opened 2026-02-09
New: Add option to not download before air date to Release Profiles

### LINKED ISSUE #969: Prevent Grabbing if Episode Hasn't aired

markus101:
#### Description

Adds an two settings to Release Profiles to prevent grabbing episodes before their air date as well as a grace period (in days). Release Profiles can be included via `Tags` and excluded via `Excluded Tags` to include/exclude series with matching tags.

image

* Closes #969


frontend/src/Settings/Profiles/Release/EditReleaseProfileModalContent.tsx
```diff
@@ -131,6 +140,32 @@ function EditReleaseProfileModalContent({
...
+          </FormGroup>
+
+          {doNotGrabBeforeAirDate.value ? (
+            <FormGroup>
+              <FormLabel>{translate('AirDateGracePeriod')}</FormLabel>
+
+              <FormInputGroup
+                {...airDateGracePeriod}
+                type={inputTypes.NUMBER}
+                name="airDateGracePeriod"
```
mynameisbogdan, line 163:
```suggestion
                name="airDateGracePeriod"
                unit="days"
```

mynameisbogdan (changes requested):

src/NzbDrone.Core/Datastore/Migration/226_add_air_date_filtering_to_release_profiles.cs
```diff
@@ -0,0 +1,14 @@
...
+using NzbDrone.Core.Datastore.Migration.Framework;
+
+namespace NzbDrone.Core.Datastore.Migration;
+
+[Migration(226)]
+public class add_air_date_filtering_to_release_profiles : NzbDroneMigrationBase
+{
+    protected override void MainDbUpgrade()
+    {
+        Alter.Table("ReleaseProfiles").AddColumn("DoNotGrabBeforeAirDate").AsInt32().WithDefaultValue(0);
```
mynameisbogdan:
```suggestion
        Alter.Table("ReleaseProfiles").AddColumn("DoNotGrabBeforeAirDate").AsBoolean().WithDefaultValue(false);
```

mynameisbogdan (changes requested):

src/NzbDrone.Core/Profiles/Releases/ReleaseProfile.cs
```diff
@@ -9,6 +9,8 @@ public class ReleaseProfile : ModelBase
         public bool Enabled { get; set; }
         public List<string> Required { get; set; }
         public List<string> Ignored { get; set; }
+        public bool AirDateRejection { get; set; }
```
mynameisbogdan: It's `AirDateRestriction` or `AirDateRejection`? There are still usages of `DoNotGrabBeforeAirDate`.

markus101: `AirDateRestriction` was the plan, then I mangled it somewhere along the way 🙈

Sarke:
Why was this done in days? I would want to wait maybe 4-6 hours, not an entire day.  It would be much more useful if it was in hours, or even if it was a float and you could do 0.25 days.

EDIT: You could even keep the db schema change if you want and just convert it to seconds to still use the Int32.  You can show the value as a float in days

Issue #7927, opened 2025-06-16
Display Link IDs


JamesHabben:
### Is there an existing issue for this?

- [x] I have searched the existing open and closed issues

### Is your feature request related to a problem? Please describe

Various indexers, requests, and other series related searching ask for the identifiers of the shows, and many of them ask for multiple identifiers. currently, you have to either tap the link to open a new window and then highlight a portion of the url to copy, or copy the url to the clipboard and paste it somewhere else to then grab just the ID portion - and the fields asking for the ID usually aren't big enough accept the full URL to cut down in place.

### Describe the solution you'd like

I was wondering if the team would be interested in making the database link identifiers easier to access? could be simply displaying the ID to allow the viewer to highlight and copy, or perhaps even include a clicky icon that would copy each of the IDs to the clipboard?

rough mock up in developer tools to show my thought:
<img width="759" alt="Image" src="https://github.com/user-attachments/assets/13e8ebfb-b510-4fde-beed-0889ad8aa29a" />

### Describe alternatives you've considered

described in the problem section. they are accessible but it takes many steps to get each of the ID values.

### Anything else?

I am happy to take a run at the react code if this is something that would be welcomed into the tool.

JamesHabben:
additional mock up option for having the font awesome `copy` icon show beside the id.

kinda funny that the biggest label has the smallest ID and smallest label has the biggest ID. again, just a rough mock up. ideally the labels would be sized the same width as the IDs below them.

<img width="702" alt="Image" src="https://github.com/user-attachments/assets/07992797-7a16-4d44-be06-ee9180884dc5" />


I think this is a pretty small adjustment that is not intrusive to the rest of the interface and adds helpful functionality. should really only be a bit of code in the `SeriesDetailsLinks` function return, with maybe an additional component to import if there isn't anything existing to leverage. just my thoughts to help out.
https://github.com/Sonarr/Sonarr/blob/v5-develop/frontend/src/Series/Details/SeriesDetailsLinks.tsx

markus101:
That approach seems reasonable, if you or someone wants to open a PR for it that'd be great.

bakerboy448:
Reminder to ensure it's tested on mobile - iOS device (safari is weird) and then simulate size for android; but feature request and proposed fix seem reasonable and good UX

JamesHabben:
[...]

PR #7949, opened 2025-06-29
Added IDs below SeriesDetailsLinks

ghost:
#### Description
Added ID displays with copy buttons below Links in SeriesDetailsLinks

#### Screenshots for UI Changes
image
image

image
image

#### Issues Fixed or Closed by this PR
* Closes #7927

PR #8423, opened 2026-02-19
New: Display the series link IDs

Yashizzle:
#### Description
Adding a copy button with the ID so that it can be easily copied to be used in indexers and searching tools. See #7927 for details. I've put the text in a button following the conversation [here](https://github.com/Sonarr/Sonarr/pull/7949#pullrequestreview-2995595202) and tried to make it look like existing controls.

- Add a `text` field to the ClipboardButton
- Display the link ID for `SeriesDetailsLinks`

##### Browser - Light Mode
SonarrLinkCopy

##### Browser - Dark Mode
image

##### Mobile View (Emulated)
image

##### iOS 26
image

\* Note: The background image not appearing seems to be an unrelated issue

* Closes: #7927

-----

Hi all, I'm new to React + Typescript and I'm trying to contribute as a learning exercise, so please let me know if I've done anything silly. I'm happy to update as necessary. Thanks in advance!

### DIFF
--- a/frontend/src/Components/Link/ClipboardButton.css
+++ b/frontend/src/Components/Link/ClipboardButton.css
@@ -4,6 +4,10 @@
   position: relative;
 }
 
+.buttonText {
+  margin: 0 5px;
+}
+
 .stateIconContainer {
   position: absolute;
   top: 50%;
--- a/frontend/src/Components/Link/ClipboardButton.tsx
+++ b/frontend/src/Components/Link/ClipboardButton.tsx
@@ -8,6 +8,7 @@ import styles from './ClipboardButton.css';
 
 export interface ClipboardButtonProps extends Omit<ButtonProps, 'children'> {
   value: string;
+  label?: string | number;
 }
 
 export type ClipboardState = 'success' | 'error' | null;
@@ -15,6 +16,7 @@ export type ClipboardState = 'success' | 'error' | null;
 export default function ClipboardButton({
   id,
   value,
+  label,
   className = styles.button,
   ...otherProps
 }: ClipboardButtonProps) {
@@ -68,6 +70,7 @@ export default function ClipboardButton({
         ) : null}
 
         <span className={styles.clipboardIconContainer}>
+          {label ? <span className={styles.buttonText}>{label}</span> : null}
           <Icon name={icons.CLIPBOARD} />
         </span>
       </span>
--- a/frontend/src/Series/Details/SeriesDetailsLinks.css
+++ b/frontend/src/Series/Details/SeriesDetailsLinks.css
@@ -1,4 +1,6 @@
 .links {
+  display: flex;
+  flex-wrap: wrap;
   margin: 0;
 }
 
@@ -9,12 +11,22 @@
 .linkLabel {
   composes: label from '~Components/Label.css';
 
-  cursor: pointer;
+  margin: 0;
+  border-top-right-radius: 0;
+  border-bottom-right-radius: 0;
+
+  &:hover {
+    cursor: pointer;
+  }
+}
+
+.linkBlock {
+  display: flex;
+  margin: 3px;
 }
 
 @media only screen and (max-width: $breakpointExtraSmall) {
   .links {
-    display: flex;
     flex-flow: column wrap;
   }
 }
--- a/frontend/src/Series/Details/SeriesDetailsLinks.tsx
+++ b/frontend/src/Series/Details/SeriesDetailsLinks.tsx
@@ -1,8 +1,10 @@
-import React from 'react';
+import React, { useMemo } from 'react';
 import Label from 'Components/Label';
+import ClipboardButton from 'Components/Link/ClipboardButton';
 import Link from 'Components/Link/Link';
 import { kinds, sizes } from 'Helpers/Props';
 import Series from 'Series/Series';
+import translate from 'Utilities/String/translate';
 import styles from './SeriesDetailsLinks.css';
 
 type SeriesDetailsLinksProps = Pick<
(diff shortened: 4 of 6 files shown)


frontend/src/Series/Details/SeriesDetailsLinks.tsx
```diff
@@ -12,94 +14,67 @@ type SeriesDetailsLinksProps = Pick<
 
 function SeriesDetailsLinks(props: SeriesDetailsLinksProps) {
   const { tvdbId, tvMazeId, imdbId, tmdbId } = props;
+  const links = [
```
markus101: This will result in a new array created every time, this needs to be done with `useMemo` instead.

frontend/src/Components/Link/ClipboardButton.tsx
```diff
@@ -68,6 +70,7 @@ export default function ClipboardButton({
         ) : null}
 
         <span className={styles.clipboardIconContainer}>
+          {text && <span className={styles.buttonText}>{text}</span>}
```
markus101: Use a ternary instead:

```suggestion
          {text ? <span className={styles.buttonText}>{text}</span> : null}
```

This prevents checks for falsey results from rendering things incorrectly, such as the value being `0` or `false` from rendering instead of null.

frontend/src/Components/Link/ClipboardButton.tsx
```diff
@@ -8,13 +8,15 @@ import styles from './ClipboardButton.css';
 
 export interface ClipboardButtonProps extends Omit<ButtonProps, 'children'> {
   value: string;
+  text?: string;
```
markus101: I think `label` would be more consistent with other components.

frontend/src/Series/Details/SeriesDetailsLinks.css
```diff
@@ -1,20 +1,37 @@
...
   margin: 0;
 }
 
 .link {
+  display: flex;
   white-space: nowrap;
+
+  &:hover {
+    text-decoration: none;
+  }
```
markus101: Is this necessary? Looks like things display the same with or without this addition.

frontend/src/Components/Link/ClipboardButton.css
```diff
@@ -4,6 +4,10 @@
   position: relative;
 }
 
+.buttonText {
+  padding: 3px;
```
markus101: I don't think we need vertical padding and margin is probably a better choice, 3px also looks a little too thin maybe?

```suggestion
  margin: 0 5px;
```

frontend/src/Series/Details/SeriesDetailsLinks.tsx
```diff
@@ -12,94 +14,67 @@ type SeriesDetailsLinksProps = Pick<
...
+            <Link className={styles.link} to={link.url}>
+              <Label
+                className={styles.linkLabel}
+                kind={kinds.INFO}
+                size={sizes.LARGE}
+              >
+                {link.name}
+              </Label>
+            </Link>
+            <ClipboardButton
```
markus101: Nit: extra spacing between these multi-line components.

```suggestion
            </Link>

            <ClipboardButton
```

frontend/src/Series/Details/SeriesDetailsLinks.tsx
mynameisbogdan: This needs to be done with `Array.push` instead, there's no need to add invalid items to the list just to `if (!link.id) { return null; }` them IMO.

frontend/src/Series/Details/SeriesDetailsLinks.css
Yashizzle: Looks like this was only necessary because I also added `display: flex`. Removing both that and the hover seems display the same.

Yashizzle:
Updated to address the comments. Thank you @markus101 & @mynameisbogdan. 
Really appreciate you taking the time to explain why the changes were necessary too. I learnt quite a few things.
Please let me know if there's anything else.

frontend/src/Series/Details/SeriesDetailsLinks.tsx
```diff
@@ -1,105 +1,101 @@
...
+import translate from 'Utilities/String/translate';
 import styles from './SeriesDetailsLinks.css';
 
 type SeriesDetailsLinksProps = Pick<
   Series,
   'tvdbId' | 'tvMazeId' | 'imdbId' | 'tmdbId'
 >;
 
+interface SeriesDetailsLink {
+  id: string;
```
mynameisbogdan: Maybe `externalId` would be better in this context? 

```suggestion
  externalId: string | number;
```

mynameisbogdan: @Yashizzle Let's see what others think before making any changes, but this would be better as `string | number` to avoid any `toString`s IMO.

markus101: Yeah, I agree.

Yashizzle: In this case, I'll have to call `toString` later on since `ClipboardButton` only accepts strings as a `value`. 

I hesitate to update the `ClipboardButton` component to accept numbers since `navigator.clipboard.writeText` used within only accepts strings.

Converting to strings within `useMemo` feels like the best approach here. Happy to follow your guidance here though. In the meantime I'll rename the variable.

mynameisbogdan: It's okay to use `${externalId}` there.

PR #8434, opened 2026-02-25
New: Warning on add series list for Import list exclusions

### LINKED ISSUE #8433: Add a warning to the add new series tool when a series is excluded from import lists

Nope

I think being able to use import list exclusions, in this case from the add new series page, to help prevent redownloading shows you already watched years ago but maybe forgot you didn't like would be a nice feature. After I start watching a show that I'm not enjoying, or that is not good enough to keep around after having watched it, I delete it and add it to the import exclusion list. Most of my shows are added through the add new series page, however, not import lists, and it seems it does not check the import list exclusions.

It'd be nice if each series in the add new series page could have a warning next to the add button if that show is excluded from import lists. But still allow adding the series if the user so wishes, such as when they want to give the show another shot.

I don't know of any alternatives to this type of feature, aside from avoiding the add new series tool entirely or manually tracking unwanted shows.

### Anything else?

_No response_

stevietv:
#### Description
If a series is on the Import List exclusions list then show a warning Icon

image

image

* Closes #8433

### DIFF
--- a/frontend/src/AddSeries/AddNewSeries/AddNewSeriesSearchResult.tsx
+++ b/frontend/src/AddSeries/AddNewSeries/AddNewSeriesSearchResult.tsx
@@ -34,7 +34,7 @@ function AddNewSeriesSearchResult({ series }: AddNewSeriesSearchResultProps) {
     overview,
     seriesType,
     images,
-    isExcluded = false,
+    isExcluded,
   } = series;
 
   const isExistingSeries = useExistingSeries(tvdbId);
--- a/frontend/src/AddSeries/AddSeries.ts
+++ b/frontend/src/AddSeries/AddSeries.ts
@@ -2,6 +2,7 @@ import Series from 'Series/Series';
 
 interface AddSeries extends Series {
   folder: string;
+  isExcluded: boolean;
 }
 
 export default AddSeries;
--- a/frontend/src/Series/Series.ts
+++ b/frontend/src/Series/Series.ts
@@ -102,7 +102,6 @@ interface Series extends ModelBase {
   tmdbId: number;
   useSceneNumbering: boolean;
   year: number;
-  isExcluded?: boolean;
   addOptions: SeriesAddOptions;
 }
 


src/Sonarr.Api.V5/Series/SeriesLookupController.cs
```diff
@@ -45,6 +48,12 @@ private IEnumerable<SeriesResource> MapToResource(IEnumerable<NzbDrone.Core.Tv.S
 
             resource.Folder = _fileNameBuilder.GetSeriesFolder(currentSeries);
             resource.Statistics = new SeriesStatistics().ToResource(resource.Seasons);
+            var importListExclusion = _importListExclusionService.FindByTvdbId(currentSeries.TvdbId);
+
+            if (importListExclusion != null)
+            {
+                resource.ImportListExclusion = true;
+            }
```
mynameisbogdan:
```csharp
resource.ImportListExclusion = importListExclusion is not null;
```

src/Sonarr.Api.V5/Series/SeriesResource.cs
```diff
@@ -53,6 +53,7 @@ public class SeriesResource : RestResource
     public Ratings? Ratings { get; set; }
     public SeriesStatisticsResource? Statistics { get; set; }
     public bool? EpisodesChanged { get; set; }
+    public bool ImportListExclusion { get; set; }
```
mynameisbogdan:
```suggestion
    public bool? ImportListExclusion { get; set; }
```
or
```suggestion
    [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingDefault)]
    public bool ImportListExclusion { get; set; }
```

I think we should go with a hidden field here so it can be ignored on other endpoints, since otherwise it's going to be false and it's going to be misleading.

Also `ImportListExclusion` would mean the whole object. In radarr I added basic `IsExcluded` with this on the resource.

```csharp
        // Hiding this so people don't think its usable (only used for searches)
        [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingDefault)]
        [SwaggerIgnore]
        public bool IsExcluded { get; set; }
```

frontend/src/Series/Series.ts
```diff
@@ -103,6 +103,7 @@ interface Series extends ModelBase {
   useSceneNumbering: boolean;
   year: number;
   addOptions: SeriesAddOptions;
+  importListExclusion: boolean;
```
mynameisbogdan:
```suggestion
  importListExclusion: boolean;
  addOptions: SeriesAddOptions;
```

I think we should leave `addOptions` last. 😄

stevietv:
Thanks for the review B, made the tweaks as you proposed

frontend/src/AddSeries/AddNewSeries/AddNewSeriesSearchResult.tsx
```diff
@@ -34,7 +34,7 @@ function AddNewSeriesSearchResult({ series }: AddNewSeriesSearchResultProps) {
     overview,
     seriesType,
     images,
-    importListExclusion,
+    isExcluded,
```
mynameisbogdan, line 37:
```suggestion
    isExcluded = false,
```

frontend/src/Series/Series.ts
```diff
@@ -102,6 +102,7 @@ interface Series extends ModelBase {
   tmdbId: number;
   useSceneNumbering: boolean;
   year: number;
+  isExcluded: boolean;
```
mynameisbogdan:
```suggestion
  isExcluded?: boolean;
```

src/Sonarr.Api.V5/Series/SeriesResource.cs
```diff
@@ -53,6 +55,9 @@ public class SeriesResource : RestResource
     public Ratings? Ratings { get; set; }
     public SeriesStatisticsResource? Statistics { get; set; }
     public bool? EpisodesChanged { get; set; }
+    [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingDefault)]
+    [SwaggerIgnore]
+    public bool? IsExcluded { get; set; }
```
mynameisbogdan, line 60: I tested this and making this nullable was a better choice, as false it's still going to show on search results but not on other endpoints. 👍 But the interface needs to reflect this, as my last suggestions.

mynameisbogdan (changes requested):

frontend/src/Series/Series.ts
mynameisbogdan: Actually now I wonder if we should move this to https://github.com/Sonarr/Sonarr/blob/591b569bddfcbfa7024cbfef32a35470984bac86/frontend/src/AddSeries/AddSeries.ts#L3 🤔

stevietv: I'm not sure about that, it feels more like a property of a series.

mynameisbogdan: Same thing goes with `Folder`, as `AddSeries` is used only for search results.

https://github.com/Sonarr/Sonarr/blob/b09c6f08117b492b625149239d90a7e372834e2e/src/Sonarr.Api.V5/Series/SeriesResource.cs#L47
https://github.com/Sonarr/Sonarr/blob/5a47f34ef9631b88158e12f306ffc6d51e7642c2/src/Sonarr.Api.V5/Series/SeriesLookupController.cs#L46
https://github.com/Sonarr/Sonarr/blob/591b569bddfcbfa7024cbfef32a35470984bac86/frontend/src/AddSeries/AddSeries.ts#L4

Unless you think this is going to be present on other endpoints as well.

src/Sonarr.Api.V5/Series/SeriesResource.cs
stevietv: you mean remove the ignores and just leave it nullable?

mynameisbogdan: No no, I'm was just saying I suggested 2 things and you combined them, and it proved a better choice. 😄 Now there isn't anything to be changed to the resource, just the interface for the frontend (which you already committed).

frontend/src/Series/Series.ts
stevietv: was thinking that it could be useful. what if an external program that was adding things wanted to simply check that it wasn't blocked from import lists during adding.

mynameisbogdan: I think there's a language barrier without our morning coffee. 😸 

I'm suggesting to leave the seriesresource as is, as it's ignored when null, but move the `isExcluded: boolean` to `AddSeries` interface for the frontend which is the same as `folder: string`. As `isExcluded` is going to be a property shown only for series lookup endpoint, right?

src/Sonarr.Api.V5/Series/SeriesResource.cs
stevietv: oh I get you, sorry!

frontend/src/Series/Series.ts
stevietv: now I get you, coffee is indeed needed!

mynameisbogdan: Ok, let me make a few last suggestions. 😄

frontend/src/AddSeries/AddNewSeries/AddNewSeriesSearchResult.tsx
```diff
@@ -34,6 +34,7 @@ function AddNewSeriesSearchResult({ series }: AddNewSeriesSearchResultProps) {
     overview,
     seriesType,
     images,
+    isExcluded = false,
```
mynameisbogdan:
```suggestion
    isExcluded,
```

frontend/src/AddSeries/AddSeries.ts
```diff
@@ -2,6 +2,7 @@ import Series from 'Series/Series';
 
 interface AddSeries extends Series {
   folder: string;
+  isExcluded?: boolean;
```
mynameisbogdan:
```suggestion
  isExcluded: boolean;
```

As it's now mandatory on series lookup endpoint.

PR #8444, opened 2026-03-01
Mark downloads failed if local episode custom format score was lower than the remote score

felizk:
Sometimes the custom format score calculated for a remote episode does not match the reality when the episode is downloaded and media-info is analyzed for the downloaded file.

For example, if an episode appears to have English language before it is downloaded, but upon analyzing the files, it turns out to not have English language.

This could reduce the custom score for the episode and potentially could cause an import rejection based on custom format score.

If such an import is rejected, there is nothing preventing Sonarr from downloading the same release again if a search is initiated for that episode. The RemoteEpisode will still appear as having a high enough custom score to replace the existing file.

This change checks for this and marks a download failed preventing it from being downloaded again.

If a import was rejected based on it not being a custom-format upgrade AND the remote episode custom format score is *higher* than the one from the analyzed local episode, then and only then do we mark the download failed.

### DIFF
--- a/src/NzbDrone.Core.Test/MediaFiles/DownloadedEpisodesImportServiceFixture.cs
+++ b/src/NzbDrone.Core.Test/MediaFiles/DownloadedEpisodesImportServiceFixture.cs
@@ -89,6 +89,29 @@ namespace NzbDrone.Core.Test.MediaFiles
                   .Callback(() => WasImportedResponse());
         }
 
+        private LocalEpisode GivenRejectedImportDecision(ImportRejectionReason reason)
+        {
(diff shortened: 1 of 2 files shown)


markus101:
This cannot be treated as a failed download, that will result in it being removed from download clients which would result in issues where seeding is required.

felizk:
Oh! That makes sense! 

What would be the correct way to mark a download / remote episode as "do not download again"? 

My reference point was going into the episode and manually marking it failed, but I didn't realize this affects the download clients.

felizk:
Would it be better if it used the `IBlocklistService`?

PR #8445, opened 2026-03-01
Block releases if custom score rejection during import

felizk:
Sometimes the custom format score calculated for a remote episode does not match the reality when the episode is downloaded and media-info is analyzed for the downloaded file.

For example, if an episode appears to have English language before it is downloaded, but upon analyzing the files, it turns out to not have English language.

This could reduce the custom score for the episode and potentially could cause an import rejection based on custom format score.

If such an import is rejected, there is nothing preventing Sonarr from downloading the same release again if a search is initiated for that episode. The RemoteEpisode will still appear as having a high enough custom score to replace the existing file.

This change checks for this and blocklists the associated release preventing it from being downloaded again.

If a import was rejected based on it not being a custom-format upgrade AND the remote episode custom format score is higher than the one from the analyzed local episode, then and only then do we blocklist.

(Sorry this is back so similar to the previous one, but I couldn't update the other pull request)

### DIFF
--- a/src/NzbDrone.Core.Test/MediaFiles/DownloadedEpisodesImportServiceFixture.cs
+++ b/src/NzbDrone.Core.Test/MediaFiles/DownloadedEpisodesImportServiceFixture.cs
@@ -8,6 +8,7 @@ using Moq;
 using NUnit.Framework;
 using NzbDrone.Common.Disk;
 using NzbDrone.Common.Extensions;
+using NzbDrone.Core.Blocklisting;
 using NzbDrone.Core.Download;
 using NzbDrone.Core.Download.TrackedDownloads;
 using NzbDrone.Core.MediaFiles;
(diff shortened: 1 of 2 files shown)


felizk:
Feel free to tell me to buzz off with this. I just thought it might be useful. :)

markus101:
I was going to add to your previous PR, but you've opened another one (ignoring the template again).

That would prevent it from being grabbed again, but it'd still sit in the queue. This is a lot more complicated than changing what happens.

Like I said in the last PR the user needs to be able to decide what to do with the download, so you can't just flip this on globally because it solves the problem for you. Either a user needs to make a decision after it's rejected or be able to allow users to customize a whole solution on how to handle rejections. That wouldn't be a small undertaking and not something we're going to review a PR out of the blue either. There are already some external tools that do this to some degree, Declutarr is one that ones to mind

This is not a throw a quick PR together and get it landed feature.

felizk:
Thank you for taking the time to explain!

markus101:
If this is something you're super interested in adding the right way then an issue following the feature template would be a good start and we can discuss more on how to implement it before spending too much time going in directions that won't make long term sense. At a high level it'd require some sort of rules system (like CFs, Auto tagging, etc) that would allow for customization as well as proper handling for usenet and seeding torrents.

PR #8446, opened 2026-03-02
New: Delete files for Select Series

### LINKED ISSUE #5110: Delete File(s) button in Mass Editor

Lacking flexibility on the deletion of library items

I am currently trying to clean up "old" shows that I have either watched, grabbed (but have no intention of watching), or simply don't want. 

However I would like to filter or make assumptions on (meta) data based on Sonarr information. Things like ratings, release date, etc. 

Normally, I could delete an entire folder on the disk itself, but, then I wouldn't be able to make assumptions or determine specifications with the files I would want to delete (EG: it would be more work this way). 

Filebot
Removing files by hand

### Anything else?

In the mass editor, click the delete button branch out a specification of "Delete series and files" and another option of "Delete files only".

markus101:
#### Description
Adds a new option when selecting a series to remove all episode files, alongside deleting the series entirely.

image

* Closes #5110


thezoggy:
what if someone questions what is tracked/untracked on that. and then does it make sense to offer up a 'delete bypass recyclebin' option

markus101:
Tracked is things actually linked to episodes, not random other video files that may exist and aren't.

I considered it, but opted to follow the existing pattern of consistently sending to the recycle bin.

thezoggy:
> Tracked is things actually linked to episodes, not random other video files that may exist and aren't.
> 
> I considered it, but opted to follow the existing pattern of consistently sending to the recycle bin.

right, but its more of just what if the user questions what the '# episode files' means because it may be more than media files or if it leaves stuff over (thus was untracked) then people are like what. dunno if there is a good way to show what the said files are via tooltip or option to list said files under it  (collapse by default) so people could review to confirm. vs having to just do it and see per logs what it did or look in the folder after the fact to check.

markus101:
The file information is not known and not really reasonable to pull in on multiple series. Manage Episodes for an individual series would be more appropriate.

frontend/src/Series/Index/Select/Delete/useSelectedSeriesStats.ts
```diff
@@ -0,0 +1,45 @@
...
+  const { data: allSeries } = useSeries();
+  const { useSelectedIds } = useSelect<Series>();
+  const seriesIds = useSelectedIds();
+
+  const series = useMemo((): Series[] => {
+    const seriesList = seriesIds.map((id) => {
+      return allSeries.find((s) => s.id === id);
+    }) as Series[];
+
+    return orderBy(seriesList, ['sortTitle']);
```
mynameisbogdan: How would the performance be with native sorting?

```suggestion
    return seriesList.sort(sortByProp('sortTitle'));
```

markus101: Should be fine, I'll switch to that.

frontend/src/Series/Index/Select/Delete/useSelectedSeriesStats.ts
```diff
@@ -0,0 +1,45 @@
...
+  const { data: allSeries } = useSeries();
+  const { useSelectedIds } = useSelect<Series>();
+  const seriesIds = useSelectedIds();
+
+  const series = useMemo((): Series[] => {
+    const seriesList = seriesIds.map((id) => {
+      return allSeries.find((s) => s.id === id);
+    }) as Series[];
+
+    return seriesList.sort(sortByProp ('sortTitle'));
```
mynameisbogdan:
```suggestion
    return seriesList.sort(sortByProp('sortTitle'));
```

What a weird error from eslint.
```
Delete `·`
```

PR #8470, opened 2026-03-11
Fixed: First column reordering

Yashizzle:
#### Description
Minor bug fix. The first column of a table could not be re-ordered since the index `0` evaluated to `false`.

**Steps to reproduce**

1. Navigate to Wanted -> Missing -> Options
2. Select the first column and drag it down

Expected: The column is reordered
Actual: The column snaps back into first (original) place

This only seems to impact V5.

**Before**

ColumnReorder-Before

**After**

ColumnReorder-After


markus101:
Thanks!

PR #8471, opened 2026-03-11
Bump .NET to 10.0.5

mynameisbogdan:


Mitaku716:
I think this looks fine. I would just like to know if you managed to test that this worked locally. it just looks like you only upgraded a library thats about it.

mynameisbogdan:
I bumped the requirement on the dotnet-sdk too, in case you missed that, which is the most important change. 

And yes, of course it's been tested.

PR #8476, opened 2026-03-16
Fix: Invisible UI Elements

npapadacis:
#### Description

- fanart/poster was rendering behind body as z-index was set to -1 but parent z-index was unset.
- Mobile action menu for seasons would not render when changing from desktop to mobile view as maxheight defaults to 0 until updateMaxHeight call is made.

#### Poster Before

image

#### Poster After

image

#### Action button clicked but menu not visible

image

### DIFF
--- a/frontend/src/Components/Menu/Menu.tsx
+++ b/frontend/src/Components/Menu/Menu.tsx
@@ -157,7 +157,7 @@ function Menu({
           {React.cloneElement(childrenArray[1] as ReactElement, {
             forwardedRef: refs.setFloating,
             style: {
-              maxHeight,
+              maxHeight: enforceMaxHeight ? maxHeight : undefined,
               ...floatingStyles,
             },
             isOpen: isMenuOpen,
--- a/frontend/src/Series/Details/SeriesDetails.css
+++ b/frontend/src/Series/Details/SeriesDetails.css
@@ -4,6 +4,7 @@
 
 .header {
   position: relative;
+  z-index: 0;
   width: 100%;
 }
 


frontend/src/Components/Menu/Menu.tsx
```diff
@@ -157,7 +157,7 @@ function Menu({
           {React.cloneElement(childrenArray[1] as ReactElement, {
             forwardedRef: refs.setFloating,
             style: {
-              maxHeight,
+              ...(enforceMaxHeight ? { maxHeight } : {}),
```
markus101: No need to spread a single value when it can be undefined.

```suggestion
              ...enforceMaxHeight ? maxHeight : undefined,
```

markus101 (changes requested):
Please update the description to match the template, instead of what I assume is a slop driven template.

Also these fixes should be separate commits, in this one PR.

npapadacis:
> Please update the description to match the template, instead of what I assume is a slop driven template.
> 
> Also these fixes should be separate commits, in this one PR.

Apologies, didn't realise the template was prescriptive. Used it as guidance instead. I have updated the PR description now and split the squashed commit back into two separate ones.

mynameisbogdan:
Didn't even noticed the missing banner, when that got broken. 😄

frontend/src/Components/Menu/Menu.tsx
mynameisbogdan: It doesn't work.

```
TS2698: Spread types may only be created from object types.
    158 |             forwardedRef: refs.setFloating,
    159 |             style: {
  > 160 |               ...(enforceMaxHeight ? maxHeight : undefined),
        |               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    161 |               ...floatingStyles,
    162 |             },
    163 |             isOpen: isMenuOpen,
```

Works with
```js
maxHeight: enforceMaxHeight ? maxHeight : undefined,
```

mynameisbogdan (changes requested):

frontend/src/Components/Menu/Menu.tsx
```diff
@@ -157,7 +157,7 @@ function Menu({
           {React.cloneElement(childrenArray[1] as ReactElement, {
             forwardedRef: refs.setFloating,
             style: {
-              maxHeight,
+              ...enforceMaxHeight ? maxHeight : undefined,
```
markus101:
```suggestion
              maxHeight: enforceMaxHeight ? maxHeight : undefined,
```

markus101 (approved):
Thanks for the fixups

PR #8489, opened 2026-03-22
Fixed: Search input loses focus after clearing on mobile browsers

### LINKED ISSUE #8487: Focus lost after clearing search input in mobile browsers

After I've searched for a show, and then want to search for a different show, when I click or tap the "x" to clear the search input box, I have to click or tap again to get the cursor inside the input box to start typing my next search.

This is especially noticeable on mobile browser such as Safari, where it takes 1-2 extra interactions to get to a state where I can start typing.

Reproduced on the Add New Series page (/add/new) in mobile Safari and mobile Chrome on iOS 18.6.2 and on desktop Firefox 147.

When I click the "x" to clear the search input box, I would like the cursor placed inside a search input box so that I can start typing immediately my next search string.

I can tap inside the search input box, then Select All, then start typing.

As trivial as it sounds, this takes extra time and interactions to start typing the next search.

### Anything else?

As one point of comparison, when I've used the omnibar (search input box) to run a web search in mobile Safari, then tap to bring focus back to the omnibar, Safari brings up the input box with the existing text already highlighted, so that with that one tap I can start typing my next input immediately.

Happy to submit a PR if this is welcome.

MikeTheCanuck:
#### Description
On mobile browsers (Safari, Chrome), tapping the × button to clear the Add New Series search input dismisses the keyboard and drops focus. A second tap is required to resume typing — an unnecessary friction point on mobile.

The fix calls `focus()` on the search input synchronously within the clear button's press handler. Because the call happens inside a direct user gesture, mobile browsers honour it and raise the keyboard immediately.

Steps to reproduce

1. Open `/add/new` on a mobile browser
2. Type a search term
3. Tap the **×** button to clear it

**Expected:** keyboard stays up, cursor is in the input, ready to type

**Actual:** keyboard dismisses, input is unfocused, requires a second tap

**Note on `TextInput.tsx` diff**

The diff for `TextInput.tsx` looks larger than the change warrants; wrapping the component in `forwardRef` required re-indenting the entire function body. The functional change is two new imports: the `TextInputHandle` interface, and the `useImperativeHandle` call. [View without whitespace changes](https://github.com/Sonarr/Sonarr/pull/8489/files?w=1) to see the real delta.

#### Database Migration
NO

* Closes #8487

### DIFF
--- a/frontend/src/Components/Form/TextInput.tsx
+++ b/frontend/src/Components/Form/TextInput.tsx
@@ -6,7 +6,6 @@ import React, {
   SyntheticEvent,
   useCallback,
   useEffect,
-  useImperativeHandle,
   useRef,
 } from 'react';
 import { FileInputChanged, InputChanged } from 'typings/inputs';
@@ -41,11 +40,7 @@ export interface FileInputProps extends CommonTextInputProps {
   onChange: (change: FileInputChanged) => void;
 }
 
-export interface TextInputHandle {
-  focus: () => void;
-}
-
-const TextInput = forwardRef<TextInputHandle, TextInputProps | FileInputProps>(
+const TextInput = forwardRef<HTMLInputElement, TextInputProps | FileInputProps>(
   (
     {
       className = styles.input,
@@ -71,12 +66,18 @@ const TextInput = forwardRef<TextInputHandle, TextInputProps | FileInputProps>(
   ) => {
     const inputRef = useRef<HTMLInputElement>(null);
 
-    useImperativeHandle(
-      ref,
-      () => ({
-        focus: () => inputRef.current?.focus(),
-      }),
-      []
+    const setRef = useCallback(
+      (node: HTMLInputElement | null) => {
+        (inputRef as React.MutableRefObject<HTMLInputElement | null>).current =
+          node;
+        if (typeof ref === 'function') {
+          ref(node);
+        } else if (ref) {
+          (ref as React.MutableRefObject<HTMLInputElement | null>).current =
+            node;
+        }
+      },
+      [ref]
     );
 
     const selectionTimeout = useRef<ReturnType<typeof setTimeout>>();
@@ -171,7 +172,7 @@ const TextInput = forwardRef<TextInputHandle, TextInputProps | FileInputProps>(
 
     return (
       <input
-        ref={inputRef}
+        ref={setRef}
         type={type}
         readOnly={readOnly}
         autoFocus={autoFocus}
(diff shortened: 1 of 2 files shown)


MikeTheCanuck:
Opened prematurely — will resubmit once #8487 is triaged.

mynameisbogdan:
FYI you'll need to run `yarn lint-fix` as current eslint configuration isn't fully happy with `TextInput.tsx`.

stevietv:
Can't it be done with a simple `e.preventDefault()` ?

MikeTheCanuck:
> Can't it be done with a simple `e.preventDefault()` ?

Good point - `e.preventDefault()` on `mousedown` is a cleaner approach for desktop (it keeps focus in the input without needing to imperatively re-focus it). 

However, on mobile browsers, IIUC focus is lost during `touchstart`, which fires before the synthesized `mousedown`. By the time `preventDefault()` could prevent anything, focus has already moved. 

The PR's `.focus()` call in the press handler works in both environments.

Happy to reconsider if you have a different placement in mind.

frontend/src/Components/Form/TextInput.tsx
```diff
@@ -39,147 +41,165 @@ export interface FileInputProps extends CommonTextInputProps {
...
+      onFocus,
+      onCopy,
+      onChange,
+      onSelectionChange,
+    }: TextInputProps | FileInputProps,
+    ref
+  ) => {
+    const inputRef = useRef<HTMLInputElement>(null);
+
+    useImperativeHandle(
```
markus101: Can we achieve this with a combined ref instead of separate refs and `useImperativeHandle`?

markus101 (commented):
Looking good, thanks for implementing, just the one question.

MikeTheCanuck (commented):
Done — replaced `useImperativeHandle` and the `TextInputHandle` interface with a combined ref callback (`setRef`) that sets both the internal `inputRef` and the forwarded ref. `AddNewSeries` now uses `useRef<HTMLInputElement>` directly. Thanks for the direction.

frontend/src/Components/Form/TextInput.tsx
MikeTheCanuck: Done — replaced useImperativeHandle and the TextInputHandle interface with a combined ref callback (setRef) that sets both the internal inputRef and the forwarded ref. AddNewSeries now uses useRef<HTMLInputElement> directly. Thanks for the direction.

markus101:
I added a new `useCombinedRefs` hook and implemented it here instead of potentially having a bunch of `setRefs`.

PR #8507, opened 2026-04-01
docs: fix grammar and clarify search error message

Lucgecko:
#### Description
Added error message info for users and cleaned up grammar in relation to SQLite DB API generated messages duplicating.

Added troubleshooting context (slow i/o and lack of admin privileges).
Deleted "its" for linear grammar😁.

#### Database Migration
NO - ###

### DIFF
--- a/src/NzbDrone.Core/Localization/Core/en.json
+++ b/src/NzbDrone.Core/Localization/Core/en.json
@@ -1121,7 +1121,7 @@
   "InteractiveSearchGrabError": "Failed to add to download queue",
   "InteractiveSearchModalHeader": "Interactive Search",
   "InteractiveSearchModalHeaderSeason": "Interactive Search - {season}",
-  "InteractiveSearchResultsSeriesFailedErrorMessage": "Search failed because its {message}. Try refreshing the series info and verify the necessary information is present before searching again.",
+  "InteractiveSearchResultsSeriesFailedErrorMessage": "Search failed because {message}. Try refreshing the series info and verify the necessary information is present before searching again. Check disk I/O and/or admin permissions.",
   "InteractiveSearchSeason": "Interactive search for all episodes in this season",
   "Interval": "Interval",
   "InvalidFormat": "Invalid Format",


src/NzbDrone.Core/Localization/Core/en.json
```diff
@@ -1121,7 +1121,7 @@
   "InteractiveSearchGrabError": "Failed to add to download queue",
   "InteractiveSearchModalHeader": "Interactive Search",
   "InteractiveSearchModalHeaderSeason": "Interactive Search - {season}",
-  "InteractiveSearchResultsSeriesFailedErrorMessage": "Search failed because its {message}. Try refreshing the series info and verify the necessary information is present before searching again.",
+  "InteractiveSearchResultsSeriesFailedErrorMessage": "Search failed because {message}. Try refreshing the series info and verify the necessary information is present before searching again. Check disk I/O and/or admin permissions.",
```
mynameisbogdan, line 1124: > Check disk I/O and/or admin permissions.

I find it more confusing, as usually it's a networking issue or a timeout is the culprit.

stevietv: the `its` kinda works as the error message that can appear is:
* Air date is missing

Image

I don't think there are currently any other error messages that can appear, and i don't think this message appears when disk IO or admin permissions are an issue

Lucgecko: [SQLite Rescodes](https://sqlite.org/rescode.html)

[Here's a forum post](https://forums.sonarr.tv/t/search-failed-because-its-attempt-to-write-a-readonly-database-attempt-to-write-a-readonly-database-try-refreshing-the-series-info-and-verify-the-necessary-information-is-present-before-searching-again/39870)

[Another from Reddit](https://www.reddit.com/r/sonarr/comments/1ejrsnq/database_is_locked/)

I ran into this problem last night is why I submitted this, and after some research seems to have debated/multiple causes but admin privileges and failing disks seem to be the root cause for some.

markus101: it's works with `air date is missing`, but dropping `its` doesn't make the readonly DB message any clearer. Adding a message to check disk IO and.or admin permissions is just clunky, because it probably has nothing to do with the error message, it just happens to in this one instance.

Lucgecko: Makes sense, closing this to reduce PR clutter.

Issue #8133, opened 2025-10-21
Reduce data loss likelyhood for new users that are migrating existing setups


BreiteSeite:
### Is there an existing issue for this?

- [x] I have searched the existing open and closed issues

### Is your feature request related to a problem? Please describe

I had data loss of many seaons of multiple series because of the way sonarr handled multi-season torrents when importing existing data while not knowing about this limitation and being surprised about it removing torrents by default.

Concrete was the following situation:
* Already had many movies/series before setting up sonarr
* My setup before sonarr was that the downloads get moved into my media folder, hence i had my torrent client to set to never delete any downloads (also i kinda like to keep them for the possibility to seed at a later point)
* I migrated (with the help of fryfrog in Discord) my existing setup by moving the torrents back to the "incomplete" folder and manually add the series, letting sonarr discover them via the downloader and import them out of the downloader. However, importing multi-series torrents are unreliable ([source](https://discord.com/channels/264387956343570434/264387956343570434/1430270205103439882))

However, as a new user, i did not know about the limitation that multi series/movie downloads are not really supported. I also did not expect sonarr to remove the downloads after import (as this setting was opt-out, and i was expecting defaults to be sane - as in: not destructive). Additionally, the remove completed settings was not even visible in the adding-a-downloader dialog as it was scrolled out of view with no indication that i have to scroll for more options as the dialog looks like it is just the visible options (it perfectly fills) (no scrolling indicator, no "cut off text" that indicated that there is more at the bottom). So i didn't even know about the setting. For reference:

Sreenshot showing that it's not obvious that there are more settings.
<img width="729" height="758" alt="Image" src="https://github.com/user-attachments/assets/2a85de67-e0e3-4134-b830-b3ad590362b5" />

### Describe the solution you'd like

I'm not sure if all solutions have to be implemented or make the most sense (as i'm a newb to the software), but my proposals for discussions are:

1. Make the removed complete opt-in instead of opt-out (current). This way, default behavior is non-destructive, preventing data loss without explicit user choice (for users using hardlinks, this is basically free. For users where sonarr has to copy around, this would lead to more space consumption i guess). Removing torrents that were unexpectedly kept later is much easier than to recover/redownload data that was unexpectedly removed (might be impossible to get the exact copy back actually).
2. Show a warning next to the "Removed complete" setting, that torrents with multiple seasons/movies can lead to data loss when importing existing torrents
[...]

PR #8508, opened 2026-04-04
feat: add multi-season pack auto-import support

Nikamura:
#### Description
Enables auto-import of manually downloaded multi-season packs. Sonarr detected multi-season releases but refused to auto-import them, requiring manual import. This adds episode resolution for all seasons in a multi-season pack and removes the auto-import block. Multi-season packs remain blocked at the search/grab decision level.

### DIFF
--- a/src/NzbDrone.Core/Parser/Model/ParsedEpisodeInfo.cs
+++ b/src/NzbDrone.Core/Parser/Model/ParsedEpisodeInfo.cs
@@ -22,6 +22,7 @@ namespace NzbDrone.Core.Parser.Model
         public bool FullSeason { get; set; }
         public bool IsPartialSeason { get; set; }
         public bool IsMultiSeason { get; set; }
+        public int[] SeasonNumbers { get; set; }
         public bool IsSeasonExtra { get; set; }
         public bool IsSplitEpisode { get; set; }
         public bool IsMiniSeries { get; set; }
@@ -37,6 +38,7 @@ namespace NzbDrone.Core.Parser.Model
             EpisodeNumbers = Array.Empty<int>();
             AbsoluteEpisodeNumbers = Array.Empty<int>();
             SpecialAbsoluteEpisodeNumbers = Array.Empty<decimal>();
+            SeasonNumbers = Array.Empty<int>();
             Languages = new List<Language>();
         }
 
@@ -107,7 +109,7 @@ namespace NzbDrone.Core.Parser.Model
 
                 if (FullSeason)
                 {
-                    return Model.ReleaseType.SeasonPack;
+                    return IsMultiSeason ? Model.ReleaseType.MultiSeasonPack : Model.ReleaseType.SeasonPack;
                 }
 
                 return Model.ReleaseType.Unknown;
@@ -121,6 +123,10 @@ namespace NzbDrone.Core.Parser.Model
             if (IsDaily && EpisodeNumbers.Empty())
             {
                 episodeString = string.Format("{0}", AirDate);
+            }
+            else if (FullSeason && IsMultiSeason && SeasonNumbers.Length > 1)
+            {
+                episodeString = string.Format("Season {0:00}-{1:00}", SeasonNumbers[0], SeasonNumbers[SeasonNumbers.Length - 1]);
             }
             else if (FullSeason)
             {
--- a/src/NzbDrone.Core/Parser/Parser.cs
+++ b/src/NzbDrone.Core/Parser/Parser.cs
@@ -1054,10 +1054,27 @@ namespace NzbDrone.Core.Parser
                     }
                 }
 
-                // If more than 1 season was parsed set IsMultiSeason to true so it can be rejected later
+                // If more than 1 season was parsed set IsMultiSeason to true
                 if (seasons.Distinct().Count() > 1)
                 {
                     result.IsMultiSeason = true;
+
+                    var distinctSeasons = seasons.Distinct().OrderBy(s => s).ToArray();
+
+                    if (distinctSeasons.Length == 2)
+                    {
+                        // Range format (e.g., S01-S09) where regex captures only endpoints, expand to full range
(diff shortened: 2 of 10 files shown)


mynameisbogdan:
How do you plan to deal with quality upgrades considering the possibility of every season having a different quality and different custom format scoring, increasing the chances for a download loop?

Nikamura:
> download loop?

How would we end up in the download loop ? If user has S01 and we find pack with S01 and S02. We would import S02 and S01 only be imported if new download was an upgrade. So next time pack would be ignored as it didn't include anything needed.

Or do you mean that we would still download S01 even when we wouldn't be importing it? But that's kinda the case at the moment where we download full pack, but only import S01.

mynameisbogdan:
- Feel free to read the comments on #8133 on the same subject and issues with multi-season packs.

Nikamura:
> * Feel free to read the comments on [Do not Automatically Handle Completed Multi-Season Releases #8133](https://github.com/Sonarr/Sonarr/issues/8133) on the same subject and issues with multi-season packs.

I don't think my PR introduces new loop. PR doesn't change the grab side quality evaluation logic, it only unblocks the episode resolution and import paths that were previously hard rejected.

mynameisbogdan:
> I don't think my PR introduces new loop. PR doesn't change the grab side quality evaluation logic, it only unblocks the episode resolution and import paths that were previously hard rejected.

It does. ATM multi-season packs are being blocked at search decision level, but you literally removed that specification in this PR.

Let's take `How I Met Your LLM S01-S09 Complete 1080p WEB-DL Bluray H264 DDP5 1-Mixed` for example (which contains only a season of bluray and 8 seasons web-dl), the quality is being parsed as `Bluray-1080p` but imported seasons with the quality `WEB-DL 1080p` will see this pack in future RSS syncs or searches as an eligible upgrade, thus resulting in a download loop.

Nikamura:
> 
> It does. 

I have updated the PR to exclude those changes. Current ones are just for auto importing manually added ones. 

I could also add configurable option for auto importing multi seasons.

markus101:
Your PR description needs to follow the PR template, not the summary from AI.

src/Sonarr.Api.V5/Queue/QueueResource.cs
```diff
@@ -54,7 +54,9 @@ public static QueueResource ToResource(this NzbDrone.Core.Queue.Queue model, boo
                 Id = model.Id,
                 SeriesId = model.Series?.Id,
                 EpisodeIds = model.Episodes?.Select(e => e.Id).ToList() ?? [],
-                SeasonNumbers = model.SeasonNumber.HasValue ? [model.SeasonNumber.Value] : [],
+                SeasonNumbers = model.Episodes?.Select(e => e.SeasonNumber).Distinct().OrderBy(s => s).ToList() is { Count: > 0 } seasonNumbers
```
markus101, line 57: The model should store the season numbers, not a single season number.

src/NzbDrone.Core/Parser/Parser.cs
```diff
@@ -1054,10 +1054,27 @@ private static ParsedEpisodeInfo ParseMatchCollection(MatchCollection matchColle
...
+                    var distinctSeasons = seasons.Distinct().OrderBy(s => s).ToArray();
+
+                    if (distinctSeasons.Length == 2)
+                    {
+                        // Range format (e.g., S01-S09) where regex captures only endpoints, expand to full range
+                        result.SeasonNumbers = Enumerable.Range(distinctSeasons[0], distinctSeasons[1] - distinctSeasons[0] + 1).ToArray();
+                    }
+                    else
+                    {
+                        // Discrete seasons explicitly named (e.g., S01 S03 S05), keep as-is
```
markus101, line 1071: Should we even support season packs that aren't sequential? If we parse 1, 3, 5 then we should just fail to parse instead of trying to process.

src/NzbDrone.Core/Parser/Model/ParsedEpisodeInfo.cs
```diff
@@ -122,6 +124,10 @@ public override string ToString()
             {
                 episodeString = string.Format("{0}", AirDate);
             }
+            else if (FullSeason && IsMultiSeason && SeasonNumbers.Length > 1)
```
markus101, line 127: Will we ever have `IsMultiSeason` when `FullSeason` is false?

src/NzbDrone.Core/Parser/Model/ParsedEpisodeInfo.cs
```diff
@@ -107,7 +109,7 @@ public ReleaseType ReleaseType
 
                 if (FullSeason)
                 {
-                    return Model.ReleaseType.SeasonPack;
+                    return IsMultiSeason ? Model.ReleaseType.MultiSeasonPack : Model.ReleaseType.SeasonPack;
```
markus101, line 112: Related to my previous question, if the answer is no, then `IsMultiSeason` could be handled outside of this `if` block.

src/NzbDrone.Core/Parser/Model/ParsedEpisodeInfo.cs
```diff
@@ -37,6 +38,7 @@ public ParsedEpisodeInfo()
             EpisodeNumbers = Array.Empty<int>();
             AbsoluteEpisodeNumbers = Array.Empty<int>();
             SpecialAbsoluteEpisodeNumbers = Array.Empty<decimal>();
+            SeasonNumbers = Array.Empty<int>();
```
markus101, line 41: This is going to get confusing with `SeasonNumber` and `SeasonNumbers`.

If we want to keep `SeasonNumber` to limit the number of changes then we should make `SeasonNumber` return the first entry from `SeasonNumbers` instead.

Similarly, should `IsMultiSeason` be `=> SeasonNumbers.Count > 1`?

markus101 (changes requested):

markus101:
:wave: @Nikamura, this draft pull request has been closed automatically because it has not had any activity in 90 days.

If you are still working on this, please update it and let us know so we can reopen it.

**Please do not open a new PR for these changes** — comment here and we will reopen this one.

PR #8511, opened 2026-04-05
feat: Default Manual Import to Copy when copyUsingHardlinks is enabled

### LINKED ISSUE #8510: Feature Request: Configurable default Import Mode for Manual Import (and API)

- [ ] I have searched the existing open and closed issues

Currently, the Manual Import feature (both UI and API `ManualImport` command) defaults to **Move** mode. There is no server-side configuration to change this default.

When `copyUsingHardlinks` is enabled in Media Management settings, the expected behavior for most users with hardlink-based workflows is to **always** use Copy mode (which creates hardlinks). However, Manual Import silently defaults to Move, which:

1. **Removes the original file** from the download client's directory
2. **Breaks active torrents** in qBittorrent/Deluge/etc. (torrent status becomes `missingFiles`)
3. **Contradicts** the `copyUsingHardlinks: true` setting, which implies the user wants hardlinks

The `copyUsingHardlinks` setting creates a false sense of security that hardlinks are being used everywhere, but Manual Import ignores it entirely.

**Reproduction:**
- Enable `copyUsingHardlinks: true` in Settings → Media Management
- Use Manual Import via UI (default selection is "Move") or via API (`/api/v3/command` with `name: "ManualImport"` without specifying `importMode`)
- The file is **moved** (not hardlinked) from the download directory to the library
- The torrent in qBittorrent loses its files → `missingFiles` status

Note: The Completed Download Handler (automatic import) correctly uses hardlinks. Only Manual Import is affected.

**Preferred: Infer default from existing settings (Option B)**

When `copyUsingHardlinks` is `true`, automatically default Manual Import to **Copy** mode instead of Move. This is the correct behavior since the user has explicitly opted into a hardlink-based workflow.

**Alternative: Explicit setting in Media Management config (Option A)**

Add a new field to `/api/v3/config/mediamanagement`:
```json
{
  "defaultImportMode": "copy"  // "move" | "copy" | "auto"
}
```
Where `"auto"` would use Copy when `copyUsingHardlinks` is true, Move otherwise.

This would affect both the UI default selection and the API default when `importMode` is not specified.

1. **Always pass `importMode: "copy"` in API calls** — this is what I do now in my automation scripts, but it's error-prone. Forgetting to include it causes silent data loss (files moved instead of hardlinked).

2. **Remember to select "Copy" in the UI every time** — the UI remembers the last selection in browser localStorage, but this is per-browser and resets on cache clear. New browsers/devices default back to Move.

3. **Per-download-client setting** — add `defaultImportMode` to the Download Client configuration, allowing different defaults for different clients. This is more granular but adds complexity.

### Anything else?

Same feature request filed for Radarr: https://github.com/Radarr/Radarr/issues/11418

**Environment:**
- Sonarr version: latest (linuxserver/sonarr Docker image)
- OS: Linux (Ugreen DXP4800 Plus NAS)
- Download Client: qBittorrent
- Filesystem: ext4 (hardlinks supported and working for auto-import)"

smgladkovskiy:
## Summary

When `ImportMode` is `Auto` (the default) and `copyUsingHardlinks` is enabled in Media Management settings, Manual Import now resolves to **Copy** mode instead of falling through to **Move**.

Closes #8510

## Changes

**`src/NzbDrone.Core/MediaFiles/EpisodeImport/Manual/ManualImportService.cs`**:
- Added `IConfigService` dependency to check `CopyUsingHardlinks` setting
- In `Execute(ManualImportCommand)`, when `ImportMode.Auto` and `CopyUsingHardlinks` is true, resolve to `ImportMode.Copy`
- Changed `_importApprovedEpisodes.Import()` calls to use the resolved `importMode` instead of `message.ImportMode`

## Why

Currently, `ImportMode.Auto` resolves to Move when there's no download client item (the common case for manual imports):

```csharp
case ImportMode.Auto:
    copyOnly = downloadClientItem is { CanMoveFiles: false };
    break;
```

When `downloadClientItem` is null (manual import without tracked download), `copyOnly` is always `false` → **Move**. This:

1. Removes the original file from the download client's directory
2. Breaks active torrents (status becomes `missingFiles`)
3. Contradicts the `copyUsingHardlinks: true` setting

The Completed Download Handler (automatic import) correctly uses hardlinks because it has access to the download client item. Manual Import lacks this context, so it should fall back to the user's configured preference.

## Behavior

| `ImportMode` | `CopyUsingHardlinks` | Before | After |
|---|---|---|---|
| `Auto` | `true` | Move | **Copy** |
| `Auto` | `false` | Move | Move (unchanged) |
| `Move` | any | Move | Move (unchanged) |
| `Copy` | any | Copy | Copy (unchanged) |

This is backward-compatible: explicit `importMode: "move"` or `importMode: "copy"` in API calls still work as before.

Same change submitted for Radarr: https://github.com/Radarr/Radarr/pull/11419


mynameisbogdan:
RIP your tokens. 

You need to target v5-develop, not develop.

markus101:
Don't bother, we're not interested in reviewing this AI slop.

PR #8517, opened 2026-04-12
Allow "is not" on activity queue custom filters

naxhh:
#### Description
Allow to use "is not" filter for queue custom filters. [FILTER_BUILDER](https://github.com/Sonarr/Sonarr/blob/5bde9242393b988d8b5f154b0b935f6289af9f6d/frontend/src/Activity/Queue/useQueue.ts#L38)

Now is possible to make a filter that says "status is completed & series is "my favorite one" but is not possible to create one that says "status is completed & series **is not** "my favorite one".

This should allow for that.
I did a very shallow codebase exploration so if i'm issing something important please let me know.

imagen

Couldn't find any existing related to this.

### DIFF
--- a/frontend/src/Helpers/Props/filterBuilderTypes.ts
+++ b/frontend/src/Helpers/Props/filterBuilderTypes.ts
@@ -79,6 +79,10 @@ export const possibleFilterTypes: Record<
       key: 'equal',
       value: () => translate('FilterIs'),
     },
+    {
+      key: 'notEqual',
+      value: () => translate('FilterIsNot'),
+    },
   ],
 
   exact: [


frontend/src/Helpers/Props/filterBuilderTypes.ts
```diff
@@ -79,6 +79,10 @@ export const possibleFilterTypes: Record<
       key: 'equal',
       value: () => translate('FilterIs'),
     },
+    {
+      key: 'notEqual',
+      value: () => translate('FilterIsNot'),
+    },
```
mynameisbogdan, line 85: Why not just change the type in the series filter instead of making exact the same as equal?

https://github.com/Sonarr/Sonarr/blob/5bde9242393b988d8b5f154b0b935f6289af9f6d/frontend/src/Activity/Queue/useQueue.ts#L40-L43

mynameisbogdan (changes requested):

frontend/src/Helpers/Props/filterBuilderTypes.ts
mynameisbogdan: Also, the reason it's like this currently is due to having only including support in the endpoint, as it's a server side collection and not a client side one.

https://github.com/Sonarr/Sonarr/blob/5bde9242393b988d8b5f154b0b935f6289af9f6d/src/Sonarr.Api.V5/Queue/QueueController.cs#L173

So changing the filter type won't do anything unless you alter the API as well.

mynameisbogdan (changes requested):

frontend/src/Helpers/Props/filterBuilderTypes.ts
naxhh: yeah i started pulling from the `type` but i wasn't sure if there was a reason why should be equals instead of say `exact`. They seemed to do the same but i'm new to this codebase.

so it would be acceptable to move it to `exact` and then verify that the QueueController manages this correctly?

mynameisbogdan: > verify that the QueueController manages this correctly?

I just explained my earlier message that it doesn't, it's not a client side filtering but server side so short of adding a new query parameter to exclude series ids I can't see how you're going to implement this functionality.

naxhh: I understood your point, i was refering to adding support on that endpoint.
Would need to look at it but seems it should be able to support filtering conditions.

If you don't want to invest time in this as a feature just let me know

mynameisbogdan: @markus101 yay or nay?

markus101: I think it's reasonable idea, but a couple things come to mind:

- The changes actually need to be tested before opening a PR, or requesting a review now that it's already open
- You'll need to investigate what other parts of the app will be affected by this and ensure they're also working correctly, which could include both client and server side filtering

PR #8528, opened 2026-04-16
fix: Stop tracking downloads after they're removed from the download client

### LINKED ISSUE #2393: Ignore previous import events when grabbing a torrent release that was previously downloaded then deleted
As of version 2.0.0.5085, if a release is successfully imported then deleted, a new attempt to grab the same release is not properly monitored or imported by sonarr. The download is tracked internally, but it does not show up in "Activity" and never gets imported. Apparently, this is because the release already has `SonarrStage=Imported`.

This would effectively prevent any release from ever being re-downloaded, as far as I can tell.

See: https://forums.sonarr.tv/t/unable-to-add-torrent-file-to-deluge/17562/6

realzombee:
#### Description
Stop tracking downloads after they're removed from the download client. We already do these when manually removing items in the QueueController but is missed in the import workflow.

Steps to reproduce:
1. Grab and import a download
2. Delete the episode files
3. Grab the same hash again. Sonarr ignores this and removes it from the download client without importing.

If you do step 3 after sonarr is restarted, it would correctly import the download since the trackedDownloads cache is cleared out.

* Closes #2393

### DIFF
--- a/src/NzbDrone.Core/Download/DownloadEventHub.cs
+++ b/src/NzbDrone.Core/Download/DownloadEventHub.cs
@@ -31,7 +31,6 @@ namespace NzbDrone.Core.Download
             var trackedDownload = message.TrackedDownload;
 
             if (trackedDownload == null ||
-                message.TrackedDownload.DownloadItem.Removed ||
                 !trackedDownload.DownloadItem.CanBeRemoved)
             {
                 return;
@@ -56,8 +55,7 @@ namespace NzbDrone.Core.Download
 
             MarkItemAsImported(trackedDownload, downloadClient);
 
-            if (trackedDownload.DownloadItem.Removed ||
-                !trackedDownload.DownloadItem.CanBeRemoved ||
+            if (!trackedDownload.DownloadItem.CanBeRemoved ||
                 trackedDownload.DownloadItem.Status == DownloadItemStatus.Downloading)
             {
                 return;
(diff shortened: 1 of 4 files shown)


src/NzbDrone.Core/Download/DownloadEventHub.cs
```diff
@@ -91,6 +94,7 @@ private void RemoveFromDownloadClient(TrackedDownload trackedDownload, IDownload
                 _logger.Debug("[{0}] Removing download from {1} history", trackedDownload.DownloadItem.Title, trackedDownload.DownloadItem.DownloadClientInfo.Name);
                 downloadClient.RemoveItem(trackedDownload.DownloadItem, true);
                 trackedDownload.DownloadItem.Removed = true;
```
markus101: As discussed on Discord, this doesn't really seem to be doing anything and we should remove it completely as part of this fix.

realzombee: Removed `trackedDownload.DownloadItem.Removed = true;` and squashed

markus101: Outside of `Aria2` this was the only place where `Removed` was set (and it's not really doing anything there). `Removed` needs to be completely ripped out (from all places it's currently used), it's just dangling there doing nothing now.

Issue #7874, opened 2025-05-22
Trakt Lists no longer work after 24 hours


antigravity83:
### Is there an existing issue for this?

- [x] I have searched the existing open and closed issues

### Current Behavior

Trakt access tokens only last 24 hours due to a recent change made by Trakt to their API.

As a result, any import lists that use Trakt tokens don't work after 24 hours

Trakt's instructions to app developers is as follows:

```
"We're instead going with a much lower 24 hours and relying on apps to automatically use the refresh_token more often to obtain a new access_token."

Tokens should continue to work up until their expiration date. At that time, your app needs to refresh it and get a new access_token, then use the expires_in value so your app can dynamically handle when to refresh."
```

### Expected Behavior

Access tokens to be auto refreshed using the refresh_token call at the start of the import list job. This will auto refresh the token and allow trakt lists to work for longer than 24 hours

### Steps To Reproduce

Run any trakt based import lists for longer than 24 hours

### Environment

```markdown
Any environment
```

### What branch are you running?

Main

### Trace Logs?

2025-05-22 20:47:23.5|Trace|Http|Req: 1183090 [POST] /api/v3/importlist/test?forceTest=true (from 162.*.*.181 Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/136.0.0.0 Safari/537.36)
2025-05-22 20:47:23.5|Debug|Sonarr.Http.Authentication.ApiKeyAuthenticationHandler|AuthenticationScheme: API was successfully authenticated.
2025-05-22 20:47:23.5|Trace|SonarrErrorPipeline|Handling Exception
2025-05-22 20:47:23.5|Warn|SonarrErrorPipeline|Invalid request Validation failed: 
 -- SignIn: Must authenticate with Trakt
2025-05-22 20:47:23.5|Trace|Http|Res: 1183090 [POST] /api/v3/importlist/test?forceTest=true: 400.BadRequest (19 ms)
2025-05-22 20:47:23.5|Debug|Api|[POST] /api/v3/importlist/test?forceTest=true: 400.BadRequest (19 ms)

markus101:
Please fill out all the environment information you ignored and provide a full trace log of the import task syncing after 24 hours.

markus101:
The JSON deserialization is using the wrong property attributes so refreshing the token fails, trace logs of the refresh wouldn't exactly show this, but would give some indication.

[...]

PR #7878, opened 2025-05-25
Fixed: Trakt Import List authentication after 24 hours

markus101:
#### Description

Refreshing the token never worked for Trakt, it's just a lot more obvious when the token expires after 24 hours instead of a lot longer.

#### Issues Fixed or Closed by this PR
* Closes #7874

PR #8531, opened 2026-04-19
Fixed: Refresh Trakt token 24 hours before expiry instead of 5 minutes

sean-wils:
#### Description
Trakt advises apps to refresh tokens proactively rather than waiting for expiry (#7874). With import lists syncing every 12 hours, a 5-minute window gives only one attempt right at expiry. Widening to 24 hours allows a couple of retries while the token is still valid.

* Related to #7874

### DIFF
--- a/src/NzbDrone.Core/ImportLists/Trakt/TraktImportBase.cs
+++ b/src/NzbDrone.Core/ImportLists/Trakt/TraktImportBase.cs
@@ -40,7 +40,7 @@ namespace NzbDrone.Core.ImportLists.Trakt
             Settings.Validate().Filter("AccessToken", "RefreshToken").ThrowOnError();
             _logger.Trace($"Access token expires at {Settings.Expires}");
 
-            if (Settings.Expires < DateTime.UtcNow.AddMinutes(5))
+            if (Settings.Expires < DateTime.UtcNow.AddHours(24))
             {
                 RefreshToken();
             }


mynameisbogdan:
Can you share a link to where trakt is advising this? 

Might be old, as now expires is 24h and not 3 months any longer. 

https://redirect.github.com/trakt/trakt-api/discussions/495

Basically this would lead to an always refresh token call.

https://trakt.docs.apiary.io/#reference/authentication-oauth/get-token/exchange-code-for-access_token seems to show 7 days. 😕

sean-wils:
Ahh nuts you're right. I took that directly from #7874. 🤦 Looking at the docs, it looks like the real fix would be to persist `expires_in` from the refresh response and refresh at 50% of the token's lifetime.

sean-wils:
The whole purpose of this PR was to avoid ever having to manually re-authenticate Trakt.

My lists were silently failing with no visible error (#8530) and I'd resorted to a cron job as a workaround. A dynamic window based on the actual `expires_in` would be the right way to solve that properly.

Happy to close this. I'll take a look at this other approach shortly.

mynameisbogdan:
#7874 was just an issue with bad deserialization, as the DTO was using STJson instead of Newtonsoft attributes. https://github.com/Sonarr/Sonarr/pull/7878/files

Current behavior should be fine as refresh_token still has 3 months, unless I'm mistaken. 😄 

Also the trakt oauth service has some issues from time to time.

sean-wils:
Ahh okay, cheers for the info!

PR #8537, opened 2026-04-20
Fixed queue details counter

mynameisbogdan:
#### Description
Updating `Queue` interface to match the v5 API.

queue changes
before
after

queue details on series page
Screenshot

### DIFF
--- a/frontend/src/Activity/Queue/QueueRow.tsx
+++ b/frontend/src/Activity/Queue/QueueRow.tsx
@@ -29,6 +29,8 @@ import Queue, {
   QueueTrackedDownloadStatus,
   StatusMessage,
 } from 'typings/Queue';
+import formatDateTime from 'Utilities/Date/formatDateTime';
+import getRelativeDate from 'Utilities/Date/getRelativeDate';
 import formatBytes from 'Utilities/Number/formatBytes';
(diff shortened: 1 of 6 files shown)


frontend/src/Activity/Queue/QueueRow.tsx
```diff
@@ -250,12 +250,14 @@ function QueueRow(props: QueueRowProps) {
             <TableRowCell key={name}>
               <RelativeDateCell
                 key={name}
+                className=""
```
markus101: We shouldn't be rendering a cell within a cell, we should render the cell contents directly instead

markus101 (changes requested):
The fix for episode air date fixes the issue, but just masks the issue with the implementation.

frontend/src/Activity/Queue/QueueRow.tsx
mynameisbogdan: With span you're not actually doing a table cell. Do you want to change this to literal spans? Would look more verbose.

markus101: True, more verbose, but the tool tip could have the whole range instead of two separate things as well.

markus101 (approved):

PR #8539, opened 2026-04-20
Fix loading Map objects on older browsers

mynameisbogdan:
#### Description
Map reduce/some is not supported on some older phone browsers.

### DIFF
--- a/frontend/src/polyfills.js
+++ b/frontend/src/polyfills.js
@@ -47,3 +47,5 @@ if (!('contains' in String.prototype)) {
 if (!Object.groupBy) {
   import('core-js/actual/object/group-by');
 }
+
+import 'core-js/actual/iterator';


markus101:
Which older browsers don't support it?

mynameisbogdan:
> Which older browsers don't support it?

Safari on iOS 17 doesn't, as they're part of ES2025.

https://2ality.com/2025/06/ecmascript-2025.html#iterator-helper-methods

markus101:
Can we polyfill it? It's a pretty large performance hit.

image

mynameisbogdan:
How slow is the spread variant `[...map.values()]` since it just accept iterators?

markus101:
Looks like it's the same as `Array.from`

mynameisbogdan:
I added the whole list of polyfills for Iterator from core-js, as it feels more practical than dealing with every method when it's going to be used with something like:

```js
import 'core-js/actual/iterator/every';
import 'core-js/actual/iterator/reduce';
import 'core-js/actual/iterator/some';
```

PR #8540, opened 2026-04-20
Fixed: Scroll position not reset when navigating on small screens

thibaultherve:
#### Database Migration
NO

#### Description
Scroll position isn't reset when navigating to a new page on small screens.

Originally reported on Radarr ([Radarr/Radarr#10060](https://github.com/Radarr/Radarr/issues/10060)) and fixed in [Radarr/Radarr#11431](https://github.com/Radarr/Radarr/pull/11431) but the same bug exists on Sonarr.
Reproduction steps and screenshots in [this comment on the Radarr PR](https://github.com/Radarr/Radarr/pull/11431#issuecomment-4274605067).

The real issue is that the app has two scroll containers: inner div on desktop, window on mobile (via `overflow-y: hidden !important`). `withScrollPosition` only covers the first one, so mobile falls through. Unifying the two containers would be the clean fix but probably breaks the mobile URL bar auto-hide.

This PR turns `withScrollPosition` into a `useScrollPosition` hook inside `PageContentBody`. Pages opt into memorization via a `scrollPositionKey` prop, window reset becomes automatic. That replaces the HOC and centralizes the scroll memory previously handled in   `SeriesIndex`.


frontend/src/Helpers/Hooks/useScrollPosition.ts
```diff
@@ -0,0 +1,35 @@
...
+import { useHistory, useLocation } from 'react-router';
+import { OnScroll } from 'Components/Scroller/Scroller';
+import scrollPositions from 'Store/scrollPositions';
+
+function useScrollPosition(key?: string) {
+  const { pathname } = useLocation();
+  const { action } = useHistory();
+
+  // Reset window scroll on PUSH/REPLACE (mobile's scroll container).
+  // Skip POP so the browser (mobile) and memorized `initialScrollTop` (desktop inner) can restore.
```
markus101: I think this can be more clear without having to know that the `action` can only be one of three options.

```suggestion
  // Reset the scroll position unless we're going back, this will allow the scroll
  // position to reset when moving forward (PUSH/REPLACE) and restore when
  // moving backwards (POP).
```

markus101 (commented):
Looks good, just one small update to the comment.

mynameisbogdan:
Tested on custom build, seems to work as intended. 👍

PR #8541, opened 2026-04-20
Fix series ending year on details page

mynameisbogdan:
#### Description
Last aired is not part of `SeriesStatisticsResource`, but `SeriesResource`.

https://github.com/Sonarr/Sonarr/blob/1449b8152545171a6f628a0e2ce6292e4c420da8/src/Sonarr.Api.V5/Series/SeriesController.cs#L296-L297

### DIFF
--- a/frontend/src/Series/Details/SeriesDetails.tsx
+++ b/frontend/src/Series/Details/SeriesDetails.tsx
@@ -36,7 +36,7 @@ import DeleteSeriesModal from 'Series/Delete/DeleteSeriesModal';
 import EditSeriesModal from 'Series/Edit/EditSeriesModal';
 import SeriesHistoryModal from 'Series/History/SeriesHistoryModal';
 import MonitoringOptionsModal from 'Series/MonitoringOptions/MonitoringOptionsModal';
-import { Image, Statistics } from 'Series/Series';
+import { Image, SeriesStatus, Statistics } from 'Series/Series';
 import SeriesGenres from 'Series/SeriesGenres';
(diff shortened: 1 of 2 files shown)


frontend/src/Series/Details/SeriesDetails.tsx
```diff
@@ -394,18 +394,16 @@ function SeriesDetails({ seriesId }: SeriesDetailsProps) {
...
-    lastAired,
-  } = statistics;
+  const { episodeCount = 0, episodeFileCount = 0, sizeOnDisk = 0 } = statistics;
 
   const statusDetails = getSeriesStatusDetails(status);
   const runningYears =
-    status === 'ended' ? `${year}-${getDateYear(lastAired)}` : `${year}-`;
+    status === 'ended' && lastAired
+      ? `${year}-${getDateYear(lastAired)}`
+      : `${year}-`;
```
markus101: While you're fixing this, if `year` is `0` it'll show `0-`, we should guard against that as well.

mynameisbogdan: What happens when year is 0 and last aired is not empty? `-2026` or show nothing at all?

mynameisbogdan: Screenshot 2026-04-20 at 19 46 59

Pushed change for nothing shown.

markus101: It should actually be possible, at least something I've never seen.

PR #8550, opened 2026-04-21
Add TagExisting to Import Lists

### LINKED ISSUE #8551: Lists retroactively apply tags

I would like a checkbox in the importlist screen that allows for the system to retroactively apply tags to existing movies in the database.

when i select this checkbox on the config screen, i want it to go back to series already in the system and apply the tags. Meaning if i have a list for cartoons, and the tag is cartoons. When the list is processed, if a series from that list already exists in the database, it should apply the cartoons tag to that series.

thought about doing this in python with an external tool but decided against it. better to built it into Sonarr.

### Anything else?

_No response_

scphantm:
#### Description

This is similar to the PR https://github.com/Radarr/Radarr/pull/10537.  

This adds the ability to retro-actively add tags to series that already exist in your database.

If you are wondering about use case, its pretty easy. If I have 10 Netflix series already in my collection, and I add a list that tags series with "series2".  If this is selected, when it adds new series, If one of those series already exists in the collection, it tags them with series2 as well. When you are configured to generate series.nfo files, those tags can be imported into Jellyfin and I believe Plex to be used in those applications.

This would be immensely helpful because currently, there is no way to pull up a list of series that exist in a list.  With this feature, I could update the list to retro tag the series in that list with something, then simply filter based on the tag and bam. I have all my series that I can do all my series maintenance things with.

I think this is very useful,

#### Database Migration
YES - 231

* Closes #8551


src/NzbDrone.Core/Localization/Core/en.json
```diff
@@ -1870,6 +1870,9 @@
   "Result": "Result",
   "Retention": "Retention",
   "RetentionHelpText": "Usenet only: Set to zero to set for unlimited retention",
+  "RadarrRetroactiveApply": "Retro-Apply Tags",
```
markus101: Radarr?

Why does the key spell out the whole word `retroactively`, but the translated value shortens it to `retro`?

src/NzbDrone.Core/Localization/Core/en.json
```diff
@@ -1870,6 +1870,9 @@
...
+  "RetroApplyListTagsHelpText": "Retro-Actively apply tags to series already in Sonarr",
```
markus101: It should be `Retroactively`, not `Retro-Actively`. Also use `{appName}` so it's automatically replaced by the name of the app.

src/NzbDrone.Core/ImportLists/ImportListSyncService.cs
```diff
@@ -263,6 +264,26 @@ private void ProcessListItems(List<ImportListItemInfo> items)
...
+                var preCount = series.Tags.Count;
+                foreach (var tag in importList.Tags)
+                {
+                    series.Tags.Add(tag);
+                }
+
+                if (preCount != series.Tags.Count)
+                {
+                    _seriesService.UpdateSeries(series);
+                    _logger.Debug("{0} [{1}] Retro-Actively added tags to movie", report.TmdbId, report.Title);
```
markus101: Movie?

markus101 (commented):
Please fill out the template, correctly and completely

`<!-- Remove any of the following sections if they are not used -->`

(remove issues)

For the database migration you need to include the migration number, replacing the `###`

```
YES - ###
```

You need to stop merging branches into you PR instead you should be rebasing your changes, of the 4 commits in this branch only 2 of them are relevant code changes the the other two are merge noise.

There seems to be a lot of colliding naming, `Retro`, Retro-Avtively` and even some places where `Retroactively` is used, everything should follow the correct naming of `Retroactively`.

I've already called out a few things, but there are a lot of places where things are referring to the wrong app name or media type, they all need to be addressed.

I'm curious why this never landed in Radarr in ~18ish months, I didn't see anything other than unnecessary changes (and merge commits), but perhaps I overlooked something?

scphantm:
> I'm curious why this never landed in Radarr in ~18ish months, I didn't see anything other than unnecessary changes (and merge commits), but perhaps I overlooked something?

I really don't know.  You can tell from the timeline on that PR how things went, I never got a good answer.  I always planned on porting the feature to Sonarr because I have use of it here as well, so figured try getting it into Sonarr, then maybe have something to push Radarr to implement it.

src/NzbDrone.Core/ImportLists/ImportListSyncService.cs
```diff
@@ -225,6 +225,7 @@ private void ProcessListItems(List<ImportListItemInfo> items)
                 // Break if Series Exists in DB
                 if (existingTvdbIds.Any(x => x == item.TvdbId))
                 {
+                    RetroApplyTags(importList, item);
```
markus101: This will apply tags to series already in the list, but won't apply tags to series that are going to be added in the same update window, which I think should be accounted for.

src/NzbDrone.Core/ImportLists/ImportListSyncService.cs
```diff
@@ -263,6 +264,26 @@ private void ProcessListItems(List<ImportListItemInfo> items)
...
+                var preCount = series.Tags.Count;
+                foreach (var tag in importList.Tags)
+                {
+                    series.Tags.Add(tag);
+                }
+
+                if (preCount != series.Tags.Count)
+                {
+                    _seriesService.UpdateSeries(series);
+                    _logger.Debug("{0} [{1}] Retroactively added tags to series", report.TmdbId, report.Title);
```
markus101: Wrong ID

markus101 (changes requested):
I don't think the name "Retroactively" is the best option here, maybe something like `Add Tags to Existing Series` would be better, but open to discussing further.

If retroactively is naming we go with then we should convert use the full name everywhere in the code base.

src/NzbDrone.Core/ImportLists/ImportListSyncService.cs
scphantm: The assumption here was that by the time it gets here we have already established it already exists.  If the series doesn't exist, it would be handled by the system that adds series and the tags would be applied normally.  Am I missing another use case?

scphantm:
> I don't think the name "Retroactively" is the best option here, maybe something like `Add Tags to Existing Series` would be better, but open to discussing further.
> 
> If retroactively is naming we go with then we should convert use the full name everywhere in the code base.

Fair. 'TagExisting' maybe?  It will be easier to translate as well.  

What do you think?

src/NzbDrone.Core/ImportLists/ImportListSyncService.cs
markus101: Series aren't added until after all lists are processed, for example if you have two lists A and B that both contain the same show, `My Series`.

If both A and B are processed back to back then A will add the series with it's tags and B will see that the series is going to be added and move on. The next time B updates it'll see that `My Series` was added and it'll add it's own tags, but we should add them when the series is first added.

markus101:
Yeah I think `TagExisting` makes sense.

src/NzbDrone.Core/ImportLists/ImportListSyncService.cs
scphantm: I think i see what you are talking about, basically comes down to a decision between always consistent or eventually consistent.  I will see what I come up with

src/NzbDrone.Core/ImportLists/ImportListSyncService.cs
scphantm: What id should it be?

src/NzbDrone.Core/ImportLists/ImportListSyncService.cs
scphantm: Ok, After reading the code, I think I see where you are coming from.  If a `SyncList` is called, there is no problem because its only one list report.  But if `SyncAll` is called, the resulting data structure has reports from multiple lists.  A series could exist in multiple reports.  

So I think what I have is sound for the primary use case of series already sitting in the database.

This is my first stab at fixing the edge case.  It goes to the existing list and looks to see if the tvdbid is in there first, before the add.  If it finds one, it appends the current lists tags to it.

Does this do what you are looking for?  I'm going to have to dig pretty deep into the unit tests to validate this as this edge case is really rough to simulate.  But if this is what you want, I can work on the unit tests for it tomorrow.

src/NzbDrone.Core/ImportLists/ImportListSyncService.cs
scphantm: i see it

src/NzbDrone.Core/ImportLists/ImportListSyncService.cs
```diff
@@ -263,6 +266,39 @@ private void ProcessListItems(List<ImportListItemInfo> items)
...
+            {
+                var series = _seriesService.FindByTvdbId(report.TvdbId);
+
+                var preCount = series.Tags.Count;
+                foreach (var tag in importList.Tags)
+                {
+                    series.Tags.Add(tag);
+                }
+
+                if (preCount != series.Tags.Count)
```
mynameisbogdan: So if the tags change, but the count stays the same this would ignore it?

scphantm: the only thing i'm doing is a `Tags.Add` so how would any of the other tags mutate?  Its a Set so there's no risk of duplicates.

mynameisbogdan: Still not a fan of updating the whole series here. Maybe something with `SeriesRepository.SetFields(series, s => s.Tags)` to update only the tags would be in order.

frontend/src/Settings/ImportLists/ImportLists/EditImportListModalContent.tsx
```diff
@@ -80,6 +80,7 @@ function EditImportListModalContent({
     qualityProfileId,
     seriesType,
     seasonFolder,
+    tagExisting,
     tags,
```
mynameisbogdan:
```suggestion
    tags,
    tagExisting,
```

src/NzbDrone.Core/ImportLists/ImportListSyncService.cs
scphantm: I didn't know that was in the API.  Sure, that is a much better idea.

mynameisbogdan: Take inspiration from https://github.com/Sonarr/Sonarr/blob/bf5d48c76a6a793b7d2adc540bb11d9c51bcc3cb/src/NzbDrone.Core/Tv/SeriesService.cs#L270-L273

I would've went with `UpdateTags` but it's already taken by the autotagging functionality.

scphantm: Does that look better?

mynameisbogdan: It would've been better if SeriesService was used, and not directly the repository.

scphantm: How would I use the SeriesService?  I didn't see the methods I needed in the interface.

scphantm: There is an UpdateTags method on the SeriesService but the code looked like it was for something else.

mynameisbogdan: https://github.com/Sonarr/Sonarr/pull/8550/files#r3178502586

Create a new method, you could rename existing `UpdateTags` to `UpdateAutotaggingTags`. But make sure you properly rename the existing calls to the new name too.

scphantm: Hows that.  I made the return a void because unlike the old method, there's no logic to it other than the call, like the method you showed me.

mynameisbogdan: `UpdateAutotaggingTags` even if it returns a bool, it still unused. So unless you plan to use the return value, void is fine.

But over all is much better to just `UPDATE Series SET Tags = tags WHERE Id = 1` instead of the all fields as Dapper doesn't track changes.

src/NzbDrone.Core/ImportLists/ImportListSyncService.cs
```diff
@@ -225,10 +225,13 @@ private void ProcessListItems(List<ImportListItemInfo> items)
...
                 {
+                    TagExisting(importList, item);
                     _logger.Debug("{0} [{1}] Rejected, series exists in database", item.TvdbId, item.Title);
                     continue;
                 }
 
+                UpdateExistingTagSeriesRecord(seriesToAdd, item, importList);
+
                 // Append Series if not already in DB or already on add list
                 if (seriesToAdd.All(s => s.TvdbId != item.TvdbId))
```
markus101: This iterates over every series that is going to be added twice, once to see if something needs to be tagged, then again to see if it needs to be added, this can be reduced to once though.

Something like:

```suggestion
                UpdateExistingTagSeriesRecord(seriesToAdd, item, importList);
                var pendingSeries = existingList.FirstOrDefault(s => s.TvdbId == item.TvdbId);

                // Append Series if not already in DB or already on add list
                if (pendingSeries == null)
                ...
                }
                else {
                	UpdateTagsOnPendingSeries(...);
                }
```

src/NzbDrone.Core/ImportLists/ImportListSyncService.cs
```diff
@@ -263,6 +266,39 @@ private void ProcessListItems(List<ImportListItemInfo> items)
             _logger.ProgressInfo("Import List Sync Completed. Items found: {0}, Series added: {1}", items.Count, seriesToAdd.Count);
         }
 
+        private void UpdateExistingTagSeriesRecord(List<Series> existingList, ImportListItemInfo item, ImportListDefinition importList)
```
markus101: This function is a mouthful, but also seems like some words aren't in the correct order.

I think just `UpdateTagsOnPendingSeries` (with my above suggestion) would be a lot better.

src/NzbDrone.Core/ImportLists/ImportListSyncService.cs
```diff
@@ -263,6 +266,39 @@ private void ProcessListItems(List<ImportListItemInfo> items)
...
+            if (existing != null)
+            {
+                foreach (var tag in importList.Tags)
+                {
+                    existing.Tags.Add(tag);
+                }
+            }
+        }
+
+        private void TagExisting(ImportListDefinition importList, ImportListItemInfo report)
```
markus101: Meanwhile this function is maybe too short on the naming and uses different naming for the same types from the function above (and different orders).

`UpdateTagsOnExistingSeries` would be better, but this is going to hit the DB for every existing series in every list, instead we should store the series that _might_ need to be updated (we don't know yet), then get all those series, then update the series in the DB, similar to what we do for AutoTagging.

Something like:

```
private void UpdateTagsOnExistingSeries(Dictionary<int, HashSet<int>> existingSeriesTags, ImportListDefinition importList, ImportListItemInfo info)
{
    if (importList.Tags.Any() && existingSeriesTags.TryGetValue(info.TvdbId, out var seriesTags))
    {
        foreach (var tag in importList.Tags)
        {
            seriesTags.Add(tag);
        }
    }
    else
    {
        existingSeriesTags.Add(info.TvdbId, importList.Tags);
    }
}
```

`existingSeriesTags` will be passed in from the calling function, at the end if there are items in the dictionary (there are tags that might need to be added), then loop through them, apply the tags and update the series' tags.

markus101 (changes requested):
I think the biggest thing is how many times this could end up hitting the DB (both reads and writes) and can be cleaned up greatly.

markus101:
Marking this as draft for now, let us know when it's ready for re-review.

src/NzbDrone.Core/ImportLists/ImportListSyncService.cs
scphantm: Sorry, life happened, I'm getting back to this.  I have read this over and over and I still don't see what you are seeing.  The requirement for this (which is above) is if something is being added by 2 different lists at the same time, each list has a tag, the series would be added with both tags instead of beginning inserted and then updated.  It looks like what you are talking about would roll it back to one list inserts the series, Then the tag from the second list would be picked up next time the sync ran.

This is what we talked about before that prompted this whole system to begin with

"Ok, After reading the code, I think I see where you are coming from. If a SyncList is called, there is no problem because its only one list report. But if SyncAll is called, the resulting data structure has reports from multiple lists. A series could exist in multiple reports.

So I think what I have is sound for the primary use case of series already sitting in the database.

This is my first stab at fixing the edge case. It goes to the existing list and looks to see if the tvdbid is in there first, before the add. If it finds one, it appends the current lists tags to it."

I think the scenario this whole subsystem is trying to solve for is so remote that this whole thing should be eliminated to begin with, not worth the cycles if you want to count them this closely.  Especially for a condition that will autocorrect itself the next type a sync runs.  

Can you give me a little more idea what you want to do here?

scphantm:
sorry, i asked a question a few weeks ago and didn't realize it was draft.  Please address my question above and ill get this knocked out.

src/NzbDrone.Core/ImportLists/ImportListSyncService.cs
markus101: My bad, saw the notification and never came back to reply, though I think I mangled the suggestion a little bit (left in `UpdateExistingTagSeriesRecord`), and it's not a little pseudo code:

```suggestion
                var pendingSeries = existingList.FirstOrDefault(s => s.TvdbId == item.TvdbId);

                // Append Series if not already in DB or already on add list
                if (pendingSeries == null)
                    // Add the series to `existingList`
                    ...
                }
                else {
                    // Add the tags (what `UpdateExistingTagSeriesRecord` does without the loop)
                	UpdateTagsOnPendingSeries(...);
                }
```

`UpdateExistingTagSeriesRecord(seriesToAdd, item, importList);` loops through every series we're going to add, but also `if (seriesToAdd.All(s => s.TvdbId != item.TvdbId))` loops through series we're going to add, potentially all of them, depending where it is in the list or not.

By checking the pending series then deciding what to do whether it's already pending or not we only have to loop through it once, if it's already in the list we add the additional tags, if it's not add it to the list.

It might not be as impactful as the fixing `TagExisting`, but it's a good improvement that will save refactoring later, same as doing it in general, we could do it the next pass, but that could be anywhere from 5 minutes to 12 hours later and other things may be relying on tags to make a decision.

scphantm: Ok, I just pushed a new block of code, is this more in line with what you are thinking?  If it is, ill finish testing it and do all the other stuff ready to submit it again.

src/NzbDrone.Core/ImportLists/ImportListSyncService.cs
scphantm: But at this point in the logic, I have nothing to give me `existingSeriesTags` for a series in the DB.  At this point in the logic, the only reason I know the series exists in the database is because the tvdbid's are listed in `existingTvdbIds`, I simply check that list to see if the ID exists.  If it does, then this method queries the database to get the specific details to see if we need to update it.  What am I gaining by moving that query outside this method?  I has to happen either way. 

And it doesn't query for every series, it only queries for series where the tag existing is enabled in the list.

markus101: We don't really need to know which tags exist on the series, just what needs to be added, later after we've accumulated that for all the series (including duplicates) we can apply the tags to the series and mass update the DB, that way we're doing less DB calls overall.

We also don't update the same series multiple times if it's on multiple lists.

scphantm: You are going to have to send me examples of this.  I read the autotag logic in core and i see nothing like this.  Where is this bulk update?

mynameisbogdan:
Something went wrong with your git merge. Personally I consider git rebase better.

scphantm:
That is a rebaseOn Jul 1, 2026, at 5:03 AM, Bogdan ***@***.***> wrote:﻿mynameisbogdan left a comment (Sonarr/Sonarr#8550)
Something went wrong with your git merge. Personally I consider git rebase better.

mynameisbogdan:
Screenshot 2026-07-01 at 16 33 22

This indicates it was a merge, not rebase.

scphantm:
That wasn’t what screwed me.  That just pulled in like 3 changes I had staged.  No, the true fubar was when I rebased the dev branch into this one.  That’s when things went sideways.  I HATE HATE HATE HATE HATE rebasing for this exact reason.  I have no idea how to fix it.  Botched merges, no problem.  Botched rebases, I typically delete the repo and start over.

src/NzbDrone.Core/ImportLists/ImportListSyncService.cs
scphantm: Do you have an example of the bulk update?  I saw in the auto tag where you are doing the logic to do the bulk update, but I couldn't find the actual bulk update.

And the rebase below, I hate rebase.  for that reason.  not sure it can be fixed, or even what fixed looks like honestly.  I always squash merge larger branches like this anyways so I'm not sure it matters.

markus101: I think we want to use the same as this does, https://github.com/Sonarr/Sonarr/blob/v5-develop/src/NzbDrone.Core/ImportLists/ImportListSyncService.cs#L334

I'll see if I can get a clean rebase in

markus101:
I just squashed it all down to a single commit, but rebasing should pretty much be this (assuming `Sonarr/Sonarr` is `upstream` and the default branch is `v5-develop`):

```
git checkout your-feature-branch
git fetch upstream
git rebase upstream/v5-develop
```

frontend/src/Settings/ImportLists/ImportLists/EditImportListModalContent.tsx
```diff
@@ -259,7 +260,6 @@ function EditImportListModalContent({
               {...seasonFolder}
             />
           </FormGroup>
-
```
mynameisbogdan, line 262: Newline between FormGroups.

src/NzbDrone.Core/ImportLists/ImportListSyncService.cs
scphantm: Ok, I want to make sure I have the logic you want down because this technique makes no sense to me if I have it wrong.  You want me to take `UpdateTagsOnExistingSeries(importList, item);` on line 228 and instead of processing that with a possible update, add the series to a `might_need_updated` queue.  So when I get to the point of line 274, `_addSeriesService.AddSeries(seriesToAdd, true);`, you want me to call `_addSeriesService.UpdateMany(series);` on my new queue of items i might need to update and then the AddSeries as normal?  The only upside I see to this technique is the BaseRepository wraps the UpdateMany in a transaction.  Am I getting it?

markus101: It ended up being a little more involved that I originally thought because I ignored the whole `Id` vs `TvdbId` bit when I originally wrote the suggestion. I've pushed a commit (and also removed the new merge commit) that does it, [628ff3a](https://github.com/Sonarr/Sonarr/pull/8550/changes/628ff3a8cbdef5e890a8e6cc0963357648b09873) but didn't give much thought to the naming or extracting to other methods, so could probably be cleaned up some (and should probably skip if there are no tags for that list, even if they've enabled `TagExisting`).

Effectively:

- Check if the series already exists
- If it does
	- Check if we have already added potential tags for it
		- If we have, add the tags from this list
		- If we haven't, tag the ID + tags

At the end, loop through the the series we might need to update, do our checks and update any series that have actually updated.

Hopefully that + the code explains the direction I was thinking.

mynameisbogdan:
This is becoming really messy...

scphantm:
It is much bigger than I normally like.  I just got the unit tests to run again.  What else do you want me to do?

src/NzbDrone.Core/ImportLists/ImportListSyncService.cs
scphantm: yea, I was getting where you were going.  I just wasn't seeing the infrastructure to really do it.  Thanks for that.  I pulled out some methods and cleaned things up a bit

src/NzbDrone.Core/ImportLists/ImportListSyncService.cs
```diff
@@ -256,13 +265,68 @@ private void ProcessListItems(List<ImportListItemInfo> items)
...
+                    }
+
+                    if (currentTagsCount != series.Tags.Count)
+                    {
+                        _logger.Debug("{0} [{1}] tagged existing series", series.TvdbId, series.Title);
+                        seriesWithUpdatedTags.Add(series);
+                    }
+                }
+
+                _seriesService.UpdateSeries(seriesWithUpdatedTags, true);
```
mynameisbogdan: Did you tested this with a bigger library and some big import lists? 

I feel like adding all series in memory and then update all of once won't be resources friendly.

scphantm: My test list has about 400 in it.  Higher memory is the trade off for less rounds to the DB.  I'm not aware of any other way to do it.  I just noticed your rename suggestions, I'll have that up in a few minutes

markus101: Yeah, it occurred to me, but wanted to show the example.

We can use `public void SetFields(IList<TModel> models, params Expression<Func<TModel, object>>[] properties)` from `BasicRepository`, we'd just need to expose a method in `SeriesServices` to do call it as well as a way to get just the series ID + tags.

mynameisbogdan: Why not run update tags per series? 

If existing tags needs to checked, `Enumerable.Chunk` can make life easier splitting the ids into multiple chunks.

markus101: Doing it in a single transaction is going to cut down on DB writes vs a storm of writes for each series, less of an issue for SSDs, but we know users still love their DB on spinning rust.

mynameisbogdan: > We can use `public void SetFields(IList<TModel> models, params Expression<Func<TModel, object>>[] properties)` from `BasicRepository`, we'd just need to expose a method in `SeriesServices` to do call it as well as a way to get just the series ID + tags.

Sounds better. This at least updates only the tags, not the whole row.

```diff
diff --git a/src/NzbDrone.Core/ImportLists/ImportListSyncService.cs b/src/NzbDrone.Core/ImportLists/ImportListSyncService.cs
index 9bceccc380..a0e49ef33c 100644
--- a/src/NzbDrone.Core/ImportLists/ImportListSyncService.cs
+++ b/src/NzbDrone.Core/ImportLists/ImportListSyncService.cs
@@ -323,7 +323,7 @@ private void UpdateTagsOnPendingSeries(Dictionary<int, HashSet<int>> existingSer
                     }
                 }
 
-                _seriesService.UpdateSeries(seriesWithUpdatedTags, true);
+                _seriesService.UpdateTags(seriesWithUpdatedTags);
             }
         }
 
diff --git a/src/NzbDrone.Core/Tv/SeriesService.cs b/src/NzbDrone.Core/Tv/SeriesService.cs
index 04e4c26d00..c952834cab 100644
--- a/src/NzbDrone.Core/Tv/SeriesService.cs
+++ b/src/NzbDrone.Core/Tv/SeriesService.cs
@@ -34,7 +34,7 @@ public interface ISeriesService
         bool SeriesPathExists(string folder);
         void RemoveAddOptions(Series series);
         bool UpdateAutotaggingTags(Series series);
-        void UpdateTags(Series series);
+        void UpdateTags(List<Series> series);
     }
 
     public class SeriesService : ISeriesService
@@ -311,7 +311,7 @@ public bool UpdateAutotaggingTags(Series series)
             return false;
         }
 
-        public void UpdateTags(Series series)
+        public void UpdateTags(List<Series> series)
         {
             _seriesRepository.SetFields(series, s => s.Tags);
         }

```

src/NzbDrone.Core/ImportLists/ImportListSyncService.cs
```diff
@@ -132,7 +132,11 @@ private void ProcessListItems(List<ImportListItemInfo> items)
 
             var listExclusions = _importListExclusionService.All();
             var importLists = _importListFactory.All();
-            var existingTvdbIds = _seriesService.AllSeriesTvdbIds();
+
+            var relevantTvdbIds = items.Select(x => x.TvdbId).Distinct().ToList();
```
mynameisbogdan: what's this for?

src/NzbDrone.Core/ImportLists/ImportListSyncService.cs
```diff
@@ -256,13 +268,68 @@ private void ProcessListItems(List<ImportListItemInfo> items)
...
+                    }
+
+                    if (currentTagsCount != series.Tags.Count)
+                    {
+                        _logger.Debug("{0} [{1}] tagged existing series", series.TvdbId, series.Title);
+                        seriesWithUpdatedTags.Add(series);
+                    }
+                }
+
+                _seriesService.UpdateTags(seriesWithUpdatedTags);
```
markus101: We can reduce the nesting a bit with an early return.

```suggestion
            if (!importList.TagExisting || importList.Tags.Count > 0)
            {
                return;
          	}
                
            if (existingSeriesToUpdate.TryGetValue(existingSeriesId, out var tagsToAdd))
            {
                foreach (var importListTag in importList.Tags)
                {
                    tagsToAdd.Add(importListTag);
                }
            }
            else
            {
                existingSeriesToUpdate.Add(existingSeriesId, new HashSet<int>(importList.Tags));
            }
```

src/NzbDrone.Core/ImportLists/ImportListSyncService.cs
```diff
@@ -256,13 +268,68 @@ private void ProcessListItems(List<ImportListItemInfo> items)
...
+            }
```
markus101, line 331: Same thing here:

```suggestion
            if (existingSeriesToUpdate.Count == 0)
            {
                return
            }
                
            var possibleSeriesToUpdate = _seriesService.GetSeries(existingSeriesToUpdate.Keys);
            var seriesWithUpdatedTags = new List<Series>();

            foreach (var series in possibleSeriesToUpdate)
            {
                var tags = existingSeriesToUpdate[series.Id];
                var currentTagsCount = series.Tags.Count;

                foreach (var tag in tags)
                {
                    series.Tags.Add(tag);
                }

                if (currentTagsCount != series.Tags.Count)
                {
                    _logger.Debug("{0} [{1}] tagged existing series", series.TvdbId, series.Title);
                    seriesWithUpdatedTags.Add(series);
                }
            }

            _seriesService.UpdateTags(seriesWithUpdatedTags);
```

src/NzbDrone.Core/Tv/SeriesRepository.cs
```diff
@@ -86,14 +87,21 @@ public Series FindByPath(string path)
...
             {
-                return conn.Query<int>("SELECT \"TvdbId\" FROM \"Series\"").ToList();
+                var strSql = "SELECT \"Id\" AS Key, \"TvdbId\" AS Value FROM \"Series\"";
+                return conn.Query<KeyValuePair<int, int>>(strSql).ToDictionary(x => x.Key, x => x.Value);
             }
         }
 
+        public Dictionary<int, int> SeriesTvdbIds(List<int> tvdbIds)
+        {
+            var result = Query(s => tvdbIds.Contains(s.TvdbId));
```
markus101: This is optimizes the number of rows we get from the DB, but in turn it now fetches the entire series from the DB for each row, instead of just the two values we care about.

I think we're fine with the original implementation instead of trying to further optimize it.

markus101 (changes requested):
Just a couple suggestions to reduce nesting and one thing to revert (`SeriesTvdbIds`) since it's not optimal right now and I don't think it's worth trying to fix it at the moment.

src/NzbDrone.Core/Localization/Core/en.json
```diff
@@ -1891,6 +1891,8 @@
   "Result": "Result",
   "Retention": "Retention",
   "RetentionHelpText": "Usenet only: Set to zero to set for unlimited retention",
+  "TagExistingHelpText": "Apply tags to series already in {appName}",
+  "TagExisting": "Tag Existing",
```
markus101: These also need to be sorted alphabetically in the file now that they've been renamed.

scphantm:
```  Standard Output Messages:
 23:59:57.2 Info: --- Start: NzbDrone.Common.Test.Http.HttpClientFixture<ManagedHttpDispatcher>.should_parse_malformed_cloudflare_cookie("es-ES") --- 
 23:59:57.2 Trace: Req: [GET] https://httpbin.servarr.com/response-headers?Set-Cookie=__cfduid%3Dd29e686a9d65800021c66faca0a29b4261436890790%3B%20expires%3DMon%2C%2013-Jul-26%2016%3A19%3A50%20GMT%3B%20path%3D%2F%3B%20HttpOnly 
```

Any idea why integration tests are failing with cloudflare errors?

mynameisbogdan:
The test issue is fixed in a another PR, waiting to be merged. 😺 

https://github.com/Sonarr/Sonarr/pull/8788/commits/1980773b70645371c6332d1e5c7baccf7d3e7021

scphantm:
ok.  Anything else you want me to change while we are waiting?

src/NzbDrone.Core/ImportLists/ImportListSyncService.cs
```diff
@@ -256,13 +268,72 @@ private void ProcessListItems(List<ImportListItemInfo> items)
...
 
             _addSeriesService.AddSeries(seriesToAdd, true);
+            UpdateTagsOnPendingSeries(existingSeriesToUpdate);
 
             _logger.ProgressInfo("Import List Sync Completed. Items found: {0}, Series added: {1}", items.Count, seriesToAdd.Count);
         }
 
+        private void QueueTagsOnPendingSeries(ImportListDefinition importList, Dictionary<int, HashSet<int>> existingSeriesToUpdate, int existingSeriesId)
+        {
+            if (!importList.TagExisting || importList.Tags.Count > 0)
```
mynameisbogdan: Is this correct, to skip when tags.Count is not 0?

mynameisbogdan: Btw I was only asking, because I didn't test these changes locally. I see you already pushed a change, didn't you tested them either?

markus101 (approved):
I pushed up a small change, but looks good, thanks for your work on this!

scphantm:
Sure.  I have a similar PR open for Radarr. It will take me a few days to get to it but if you want, take a look at it and start tearing it apart now and when i can later in the week I can start getting it ready.  I'm sure there are goingto be a LOT of merge conflicts to take care of but the basics should still be in place.  https://github.com/Radarr/Radarr/pull/10537

markus101:
Not sure why autopilot triggered, but y really didn't need to rebase and push it up when there weren't any conflicts and the tests were 99% done.

scphantm:
oh, sorry, i thought i needed to hit the button.  Not familiar with your automation yet.

PR #8559, opened 2026-04-22
Add OnDownloadComplete notification event

### LINKED ISSUE #8560: Feature Request: Add OnDownloadComplete notification event (fires before import)

There is currently no notification event that fires when a download client reports a download as complete but *before* Sonarr begins importing files. The existing `OnGrab` fires when a release is sent to the download client, and `OnDownload`/`OnImportComplete` fire after files are moved into the library.

This leaves a gap for users who need to trigger actions between download completion and import — for example, syncing files from a remote seedbox before Sonarr tries to import them.

A new `OnDownloadComplete` notification event that fires in `CompletedDownloadService.Check()` when a tracked download transitions to `ImportPending` state. This would slot into the notification timeline as:

`OnGrab` → **`OnDownloadComplete`** (new) → `OnDownload` (per-file import) → `OnImportComplete` (batch)

- Polling the download client API directly from an external script, but this duplicates logic Sonarr already has and is fragile.
- Using `OnGrab` combined with a timer/delay, but download times are unpredictable so this is unreliable.
- Using `OnDownload` (import) and triggering the sync reactively, but by then Sonarr may have already failed the import because the files were not yet synced.

### Anything else?

I have an implementation ready that adds this across all notification providers, the API, frontend UI, and database migration. Happy to open a PR.

JasonEtco:
#### Description
Adds a new `OnDownloadComplete` notification event that fires when the download client reports a download as complete, before Sonarr begins importing. This fills the gap between `OnGrab` and `OnDownload` (import) in the notification timeline:

`OnGrab` → **`OnDownloadComplete`** (new) → `OnDownload` (per-file import) → `OnImportComplete` (batch)

The event is published in `CompletedDownloadService.Check()` when a tracked download transitions to `ImportPending` state. All 22 applicable notification providers implement the event (Plex, Trakt, SynologyIndexer excluded — they only handle library-level events). Test fixture updated.

Apologies for the large PR — to support this new event across all notification providers, I had to touch a lot of files. Most of the provider changes are small (5-line overrides following existing patterns). Happy to split this up into smaller PRs if preferred, but I didn't want to add this in a piecemeal way that would leave providers in an inconsistent state.

#### Database Migration
YES - 231

* Closes #8560

### DIFF
--- a/frontend/src/Settings/Notifications/Notifications/NotificationEventItems.tsx
+++ b/frontend/src/Settings/Notifications/Notifications/NotificationEventItems.tsx
@@ -33,6 +33,7 @@ function NotificationEventItems({
     onHealthRestored,
     onApplicationUpdate,
     onManualInteractionRequired,
+    onDownloadComplete,
     supportsOnGrab,
     supportsOnDownload,
     supportsOnUpgrade,
@@ -44,6 +45,7 @@ function NotificationEventItems({
     supportsOnEpisodeFileDeleteForUpgrade,
     supportsOnApplicationUpdate,
     supportsOnManualInteractionRequired,
+    supportsOnDownloadComplete,
     supportsOnHealthIssue,
     supportsOnHealthRestored,
     includeHealthWarnings,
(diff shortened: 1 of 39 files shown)


markus101:
Maybe this is a feature we'd want, but there was no prior discussion and based on the spam generated by this user and presumably their LLM of choice this is not something we're going to review until it's properly discussed.

PR #8564, opened 2026-04-24
Prevent duplicated items in state on create

mynameisbogdan:
#### Description
Open to suggestions on the function name if `addOrUpdateQueryClientItem` is nobody's liking.


markus101 (approved):
I think the name is good enough, if we come up with something later we can always change it.

PR #8592, opened 2026-05-04
Backport qBittorrent fixes to v4

mynameisbogdan:
#### Description
Backport qBittorrent fixes to v4 needed for qBit 5.2 to work.

Bonus API key support.

### DIFF
--- a/src/NzbDrone.Core/Download/Clients/QBittorrent/QBittorrentProxyV2.cs
+++ b/src/NzbDrone.Core/Download/Clients/QBittorrent/QBittorrentProxyV2.cs
@@ -338,13 +338,19 @@ namespace NzbDrone.Core.Download.Clients.QBittorrent
             ProcessRequest(request, settings);
         }
 
-        private HttpRequestBuilder BuildRequest(QBittorrentSettings settings)
+        private static HttpRequestBuilder BuildRequest(QBittorrentSettings settings)
         {
             var requestBuilder = new HttpRequestBuilder(settings.UseSsl, settings.Host, settings.Port, settings.UrlBase)
             {
                 LogResponseContent = true,
                 StoreRequestCookie = false
             };
+
+            if (settings.ApiKey.IsNotNullOrWhiteSpace())
+            {
+                requestBuilder.Headers["Authorization"] = $"Bearer {settings.ApiKey}";
+            }
+
             return requestBuilder;
         }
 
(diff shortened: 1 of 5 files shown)


src/NzbDrone.Core/Download/Clients/QBittorrent/QBittorrentProxyV2.cs
```diff
@@ -338,13 +338,19 @@ public void SetForceStart(string hash, bool enabled, QBittorrentSettings setting
             ProcessRequest(request, settings);
         }
 
-        private HttpRequestBuilder BuildRequest(QBittorrentSettings settings)
+        private static HttpRequestBuilder BuildRequest(QBittorrentSettings settings)
         {
             var requestBuilder = new HttpRequestBuilder(settings.UseSsl, settings.Host, settings.Port, settings.UrlBase)
             {
                 LogResponseContent = true,
-                NetworkCredential = new BasicNetworkCredential(settings.Username, settings.Password)
```
lamdav, line 346: Not super familiar with the codebase so I'm making some guesses here. I was debugging an issue with my Sonarr instance no longer able to invoke QBittorrent APIs sitting on a seedbox despite credentials unchanged.

I think this breaks some seedbox setup that utilizes basic auth forwarding via reverse proxy to reach qbittorrent.

mynameisbogdan: You could always add the basic auth in the link anyway, but 5.2 added Basic Auth so a conflict between seedbox provider auth and qbit auth there still might be if both enabled with different credentials.

lamdav: Does Sonarr follow semver? Wouldn't this be a breaking change to backport for users on versions <5.2 qbit?

Can you share how I might add basic auth in the link? I presume we'd do `username:password@https://example.com`; however, if done in the host field, field validation would not accept this format. Were you thinking of a specific field?

mynameisbogdan: It does follow semver, but basic auth was not supported until qbit 5.2.0, and the change was needed to support API keys. 

`https://username:password@example.com`

lamdav: I tried the basic auth within the link itself. The host validation field would show an error using `https://username:password@example.com`, `username:password@example.com`
> must be valid Host without http://

But per basic auth not being supported by qbit until 5.2, it seems certain seedbox providers were depending on the sonarr basic auth behavior to implement reverse proxy authorization before routing to the qbit service which would otherwise leverage the form data on login. 

As you mentioned, this falls apart as soon as the credentials vary between the nginx auth and the qbit auth. For the time being, this would require certain seedbox providers and reverse proxy set ups to update their auth mechanism for a seemingly non-major update.

lamdav: quick update: my seedbox provider allows for custom port mapping which allows the instance to bypass their reverse proxy it seems. So, there may be certain workarounds.

PR #8598, opened 2026-05-06
Add TMDb Import Lists

### LINKED ISSUE #4959: Support for TMDB (or similar) import lists like radarr has

I'd like to be able to add import lists for people off of say TMDB like I can in Radarr. Not sure if there's a better source for this but I find the current solution of trakt lists severely lacking compared to Radarr's ease of just collecting all the actors filmographies I want.

Again not sure what the best source is but I would be pretty content with just all the TMDB options for import lists that Radarr has.

I've considered making my own trakt lists but that rather defeats the purpose of automation and it's very easy to do it in Radarr.

### Anything else?

https://wiki.servarr.com/radarr/supported#lists

ArachisH:
#### Description
Add support for importing TMDb lists using [Sonarr's](https://github.com/Sonarr/Sonarr/issues/4959#issuecomment-2677422701) application token.

#### Submission Checklist

- [X] I have read [CONTRIBUTING.md](../CONTRIBUTING.md) and this PR follows the guidelines

* Closes #4959


src/NzbDrone.Core/ImportLists/TMDb/Discover/TMDbDiscoverRequestGenerator.cs
```diff
@@ -0,0 +1,91 @@
...
+
+        var maxImportListRequests = (int)Math.Ceiling(Math.Min(Settings.Limit, resource.TotalResults) / MAX_RESULTS_PER_PAGE);
+        var importListRequests = new ImportListRequest[maxImportListRequests];
+        importListRequests[0] = new ImportListRequest(initalRequest);
+
+        for (var i = 1; i < maxImportListRequests; i++)
+        {
+            requestBuilder.AddQueryParam("page", i + 1, true);
+            importListRequests[i] = new ImportListRequest(requestBuilder.Build());
+        }
```
mynameisbogdan: Fetching all pages is not necessary, Sonarr already has a hard limit on 1000 shows per list.

https://github.com/Sonarr/Sonarr/blob/bf5d48c76a6a793b7d2adc540bb11d9c51bcc3cb/src/NzbDrone.Core/ImportLists/HttpImportListBase.cs#L23

Just yield 10 or so ImportListRequests with page defined, and leave the fetching to sonarr as it will stop when it doesn't find any more results on the page.

See https://github.com/Radarr/Radarr/blob/065414211d275a86c7b580d000065f548aafbcb5/src/NzbDrone.Core/ImportLists/TMDb/Popular/TMDbPopularRequestGenerator.cs#L116-L126

mynameisbogdan (changes requested):

src/NzbDrone.Core/ImportLists/TMDb/Discover/TMDbDiscoverRequestGenerator.cs
ArachisH: Coolio, I missed the PageSize property on my initial read through, but I see now that Plex has it as well to do just this. I've mostly been trying to use Sonarr's current list importing implementations to remain consistent with the current code base, as I didn't want to introduce too many deviations from them if I saw something differently on Radarr. I'll be looking more closely at Radarr's logic for the other list types, but here on out it should be much more straightforward I believe.

Initially I wanted it to allow the user to 'limit' the amount of total results to fetch from the combined pages, and took inspiration from the following property, and assumed Trakt did the same thing as above. https://github.com/Sonarr/Sonarr/blob/bf5d48c76a6a793b7d2adc540bb11d9c51bcc3cb/src/NzbDrone.Core/ImportLists/Trakt/TraktSettingsBase.cs#L63

Then realized it's only there because the Trakt API natively supports the query parameter.
https://github.com/Sonarr/Sonarr/blob/bf5d48c76a6a793b7d2adc540bb11d9c51bcc3cb/src/NzbDrone.Core/ImportLists/Trakt/User/TraktUserRequestGenerator.cs#L64

mynameisbogdan: Indeed, Trakt uses a single page but TMDb is to be used with pagination with 20 items per page.

ArachisH: Initially I had placed the override of the **PageSize**(20) property on **TMDbImportBase** after the suggestion, and so then all derived types would just have it set at 20, implying that they all support paging.

I ended up adding the **MaxAllowedPages** property to **TMDbImportBase** which just returns the total amount of pages allowed to query based on the **MaxNumResultsPerQuery**, and **PageSize**. The import implementations explicitly set the PageSize property if they support it, and so that allows us to use the same signature for every request generator implementation. `RequestGenerator(Settings, MaxAllowedPages)`

Let me know if I should keep it as is, otherwise we can go back to capping it at 10 pages per query. Just wanted to ensure any import job would attempt to retrieve all results possible that is allowed by **MaxNumResultsPerQuery**.

ArachisH:
> ### Worth a look
> **Unchecked items in the PR checklist**
> 
> Each checkbox is an attestation — please fill them in honestly. If one doesn't apply to your change, note why in the description.

~~Whoops~~ Adjusted PR Description

---
Implementations are finished, mostly, we can still add a couple of more configurable settings for **TMDb Discover**. I found myself micro-optimizing things then reverting a lot of it, then reverting back again so at this point I think I should just mark it ready for review to get some feedback.

### TMDbAccount
**TMDb User** was renamed to **TMDb Account** as described in the TMDb API docs:
https://developer.themoviedb.org/v4/reference/account-favorite-tv

### TMDb Discover
As stated in the original issue/feature request, companies and keyword importing was consolidated into **TMDb Discover**, since that is the API route it natively uses.

**TMDb Popular** was converted into a definition preset, for **TMDb Discover**. As this import option uses the Discover API route with some different default settings. There are [other](https://developer.themoviedb.org/reference/tv-series-airing-today-list) presets we can add for **TMDb Discover** still, but unsure how the user would expect for a 'snapshot' type list based on date/time to behave. Radarr has these **Popular** list imports implemented, but unsure how you guys would want to handle it.

The namespace **TMDb Discover** folder is also looking a little chunky, unsure if it would be preferable to keep it all in one .cs file. (Sort.cs, SortOrder.cs, Converter.cs, etc..)

### Questions

* Is **TMDbLanguageOptionsConverter** legally allowed?
* Should it be **TMDB** or **TMDb**, saw that Radarr was using **TMDb** and went with it.

src/NzbDrone.Core/ImportLists/TMDb/Account/TMDbAccountImport.cs
```diff
@@ -0,0 +1,35 @@
+﻿using NLog;
+using NzbDrone.Common.Cloud;
+using NzbDrone.Common.Http;
+using NzbDrone.Core.Configuration;
+using NzbDrone.Core.Localization;
+using NzbDrone.Core.Parser;
+
+namespace NzbDrone.Core.ImportLists.TMDb.Account;
+
+public sealed class TMDbAccountImport : TMDbImportBase<TMDbAccountSettings>
```
markus101: Drop the `sealed` to be consistent with what we're already doing.

Files should use `Tmdb`, not `TMDb`.

src/NzbDrone.Core/Localization/Core/en.json
```diff
@@ -954,6 +954,14 @@
...
   "ImportListsSonarrSettingsTagsHelpText": "Tags from the source instance to import from",
   "ImportListsSonarrValidationInvalidUrl": "{appName} URL is invalid, are you missing a URL base?",
+  "ImportListsTMDbSettingsAuthenticateWithTMDb": "Authenticate with TMDb",
+  "ImportListsTMDbSettingsIncludedCrewDepartmentCredits": "Include Crew Department Credits",
+  "ImportListsTMDbSettingsIncludedCrewDepartmentCreditsHelpText": "Departments: Art, Camera, Costume & Makeup, Crew, Directing, Editing, Lighting, Production, Sound, Visual Effects, and Writing",
+  "ImportListsTMDbSettingsIsIncludingCastCredit": "Include Cast Credit",
+  "ImportListsTMDbSettingsIsIncludingCastCreditHelpText": "Include series where the person is given cast credit.",
+  "ImportListsTMDbSettingsParametersHelpText": "List of comma separated parameters to add/replace in the query (param=value,...)",
+  "ImportListsTMDbSettingsPersonId": "Person Id",
+  "ImportListsTMDbSettingsPersonIdHelpText": "The id of the person who is a part of the cast or crew of a series.",
```
markus101: `TMDb` -> `Tmdb` to be consistent.

src/NzbDrone.Core/ImportLists/ImportListType.cs
```diff
@@ -7,6 +7,7 @@ public enum ImportListType
         Trakt,
         Simkl,
         Other,
-        Advanced
+        Advanced,
+        TMDb
```
markus101: This should follow other enums and just be `Tmdb`

```suggestion
        Tmdb
```

src/NzbDrone.Core/ImportLists/TMDb/TMDbImportBase.cs
```diff
@@ -0,0 +1,146 @@
...
+            }
+
+            var response = _httpClient.Execute(request);
+            var resource = JsonSerializer.Deserialize<RequestTokenResponse>(response.Content);
+
+            _ = TMDbToken.TryParse(resource.RequestToken, out var requestToken);
+            if (requestToken.RedirectTo.IsNullOrWhiteSpace())
+            {
+                _logger.Warn("Request token does not contain an embedded 'redirect_to' payload object.");
+            }
```
markus101: If this fails we're going to get null reference on the `requestToken` won't we?

Something like this is probably better:


```suggestion
            if (!TMDbToken.TryParse(resource.RequestToken, out var requestToken) || requestToken.RedirectTo.IsNullOrWhiteSpace())
            {
                _logger.Warn("Request token does not contain an embedded 'redirect_to' payload object.");
            }
```

src/NzbDrone.Core/ImportLists/TMDb/TMDbSettingsBase.cs
```diff
@@ -0,0 +1,73 @@
...
+        _validator = validator;
+
+        SignIn = "startOAuth";
+    }
+
+    public virtual int MaxPages { get; set; } = 10;
+
+    public override string BaseUrl { get; set; } = "https://api.themoviedb.org";
+
+    [FieldDefinition(96, Label = "Account Id", Type = FieldType.Textbox, Hidden = HiddenType.Hidden, Advanced = true)]
```
markus101: These need to be translated strings.

Also why start numbering at 96?

src/NzbDrone.Core/ImportLists/TMDb/TMDbImportBase.cs
```diff
@@ -0,0 +1,146 @@
...
+
+            request = new HttpRequestBuilder(TMDbAuthUserApproval)
+                .AddQueryParam("request_token", requestToken.Raw)
+                .Build();
+
+            return new
+            {
+                OauthUrl = request.Url.ToString(),
+                request_token = requestToken.Raw,
```
markus101: Shouldn't mix in snake case here.

markus101 (commented):
`AddImportListModalContent` Needs to be updated so we can show the proper brand name, TMDb, instead of `Tmdb`

The auth flow is a mess right now because it's trying to handle two different cases, but we shouldn't support anything other than the path through services.

Your AI looking description mentions that `redirect_to` isn't provided, but does it actually need to be? Services is already baking in the redirect URL when getting the access token and I don't think it needs to be returned as we're using [using API v4 from TMDb](https://developer.themoviedb.org/v4/reference/auth-create-request-token).

Please rip out the alternate auth mess and if we need to make changes to services to support this then we can make them.

Code should refer to things as `Tmdb`, but places where it's displaying the name should use their correct branding, `TMDb`

ArachisH:
> `AddImportListModalContent` Needs to be updated so we can show the proper brand name, TMDb, instead of `Tmdb`
> 
Frontend is like voodoo to me man, unsure how to change things there tbh, but I've gone ahead and applied the rename to all types/files.

> The auth flow is a mess right now because it's trying to handle two different cases, but we shouldn't support anything other than the path through services.
> 
> Your AI looking description mentions that `redirect_to` isn't provided, but does it actually need to be? Services is already baking in the redirect URL when getting the access token and I don't think it needs to be returned as we're using [using API v4 from TMDb](https://developer.themoviedb.org/v4/reference/auth-create-request-token).
> 
> Please rip out the alternate auth mess and if we need to make changes to services to support this then we can make them.
> 
> Code should refer to things as `Tmdb`, but places where it's displaying the name should use their correct branding, `TMDb`

Custom token support has been ripped out, but the redirect issue remains with Sonarr after checking right now.
Although the 'redirect_to' payload isn't necessarily required in other scenarios, it seems to be the only way for Sonarr to complete/detect user approval.

Once the user approves, it doesn't notify Sonarr(redirect to **callbackUrl**) to proceed with the 'getOAuthToken' action step, and so it just hangs. I've tried doing it using the Poll'ing steps that Plex does, but it doesn't want to open the TMDb auth/access page using XHR because of CORS. Was hesitant to touch any further front-end code because I wanted to contain all of the code in this namespace/folder.

The only way to test out the implementations as I wrote them, was to incorporate the custom token support from the start. Which meant using [TMDb API v4](https://developer.themoviedb.org/v4/reference/auth-create-request-token) directly with the optional 'redirect_to' payload when generating the request token.

Get Request Token: https://services.sonarr.tv/v1/auth/tmdb/request?redirectUrl=http://localhost:8989/oauth.html
image

src/NzbDrone.Core/ImportLists/TMDb/TMDbSettingsBase.cs
ArachisH: Wanted to keep it at the bottom with the rest of the default properties that are shared across other import types, so if/when we add more stuff to the derived implementations we can just append +1 to what we see visually.

I saw Trakt doing something similar, but now I see other 'SettingsBase' types are just setting their properties to 0, does it all just get rendered in order based on line order by default?

Edit: Just realized I removed the capability to allow users to edit it, and marked it hidden, will set to 0.

src/NzbDrone.Core/ImportLists/Tmdb/TmdbImportBase.cs
```diff
@@ -0,0 +1,110 @@
...
+                .Accept(HttpAccept.Json)
+                .Resource(SonarrAuthRequest)
+                .SetHeader("Content-Type", "application/json")
+                .AddQueryParam("redirectUrl", query["callbackUrl"])
+                .Build();
+
+            var response = _httpClient.Execute(request);
+            var resource = JsonSerializer.Deserialize<RequestTokenResource>(response.Content);
+
+            if (!TmdbToken.TryParse(resource.RequestToken, out var requestToken) || requestToken.RedirectTo.IsNullOrWhiteSpace())
```
markus101: The redirect to is never going to be returned, this breaks auth.

src/NzbDrone.Core/ImportLists/Tmdb/TmdbImportBase.cs
```diff
@@ -0,0 +1,110 @@
...
+            {
+                _logger.Warn("Request token does not contain an embedded 'redirect_to' payload object.");
+            }
+
+            request = new HttpRequestBuilder(TmdbAuthUserApproval)
+                .AddQueryParam("request_token", requestToken.Raw)
```
markus101: `requestToken`

src/NzbDrone.Core/ImportLists/Tmdb/TmdbImportBase.cs
markus101: Why do we need to try and parse the token?

src/NzbDrone.Core/ImportLists/Tmdb/Discover/TmdbLanguageOptionsConverter.cs
```diff
@@ -0,0 +1,63 @@
...
+
+        var neutralCultures = CultureInfo.GetCultures(CultureTypes.NeutralCultures);
+        languageOptions.EnsureCapacity(neutralCultures.Length);
+
+        var filteredLanguageOptions = neutralCultures
+            .Where(c =>
+                !c.Equals(CultureInfo.InvariantCulture) &&          // Where culture is not the default IV instance
+                c.Parent.Equals(CultureInfo.InvariantCulture) &&    // Where culture is not a sub-culture of another
+                c.TwoLetterISOLanguageName.Length == 2)             // Could return 3-letter names...
+            .OrderBy(DetermineLanguageBias)                         // Push VIP(C) languages to the top of the drop-down
```
mynameisbogdan: Pretty sure would be better to just use the list from them, as all this logic is too complicated and not even certain it's supported by them. 

Any list change might won't break the selected option?

src/NzbDrone.Core/ImportLists/Tmdb/TmdbImportBase.cs
```diff
@@ -0,0 +1,110 @@
...
+                .Accept(HttpAccept.Json)
+                .Resource(SonarrAuthAccess)
+                .SetHeader("Content-Type", "application/json")
+                .AddQueryParam("requestToken", query["requestToken"])
+                .Build();
+
+            var response = _httpClient.Execute(request);
+            var resource = JsonSerializer.Deserialize<AccessTokenResource>(response.Content);
+
+            if (!TmdbToken.TryParse(resource.AccessToken, out var accessToken) || !accessToken.CanRead)
```
markus101: Why do we need to parse the token? This token came from TMDb, if it doesn't work something bigger is broken.

src/NzbDrone.Core/ImportLists/Tmdb/Account/TmdbAccountSettings.cs
```diff
@@ -0,0 +1,21 @@
...
+{
+    private static readonly TmdbAccountSettingsValidator Validator = new();
+
+    public TmdbAccountSettings()
+        : base(Validator)
+    {
+        AccountList = (int)TmdbAccountList.Watchlist;
+    }
+
+    [FieldDefinition(1, Label = "Account List Type", HelpText = "Select the list type of the account to import.", Type = FieldType.Select, SelectOptions = typeof(TmdbAccountList))]
```
markus101: Translate properly

src/NzbDrone.Core/ImportLists/Tmdb/List/TmdbListImport.cs
```diff
@@ -0,0 +1,95 @@
...
+
+    public override IImportListRequestGenerator GetRequestGenerator()
+    {
+        return new TmdbListRequestGenerator(Settings);
+    }
+
+    private List<FieldSelectStringOption> GetAccountListOptions()
+    {
+        const int maxPages = 3;
+        var options = new List<FieldSelectStringOption>((PageSize * maxPages) + 1)
```
mynameisbogdan: No need to break it if it might return more results for some performance.  

```suggestion
        var options = new List<FieldSelectStringOption>()
```

src/NzbDrone.Core/ImportLists/Tmdb/TmdbSettingsBase.cs
```diff
@@ -0,0 +1,49 @@
...
+    protected TmdbSettingsBase(TmdbSettingsBaseValidator<TSettings> validator)
+    {
+        _validator = validator;
+
+        SignIn = "startOAuth";
+    }
+
+    public virtual int MaxPages { get; set; } = 10;
+
+    public override string BaseUrl { get; set; } = "https://api.themoviedb.org";
```
markus101, line 32: Why is this in settings?

src/NzbDrone.Core/ImportLists/Tmdb/TmdbRequestGeneratorBase.cs
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    protected abstract void SetupSeriesRequestsBuilder(HttpRequestBuilder builder);
+
+    private IEnumerable<ImportListRequest> GetSeriesRequests()
+    {
+        var builder = new HttpRequestBuilder(Settings.BaseUrl)
+            .Accept(HttpAccept.Json)
+            .SetHeader("Authorization", $"Bearer {Settings.AuthToken}")
+            .AddQueryParam("language", "en-US");
```
mynameisbogdan: Is not en-US the default?

```suggestion
            .SetHeader("Authorization", $"Bearer {Settings.AuthToken}");
```

src/NzbDrone.Core/ImportLists/Tmdb/TmdbRequestGeneratorBase.cs
```diff
@@ -0,0 +1,46 @@
...
+
+        SetupSeriesRequestsBuilder(builder);
```
markus101: How this is used is clunky, I think it makes more sense to do it when creating the builder.

src/NzbDrone.Core/ImportLists/Tmdb/Discover/TmdbDiscoverSettings.cs
```diff
@@ -0,0 +1,84 @@
...
+
+    public TmdbDiscoverSettings()
+        : base(Validator)
+    {
+        Sort = (int)TmdbDiscoverSort.Popularity;
+        SortOrder = (int)TmdbDiscoverSortOrder.Descending;
+    }
+
+    [FieldDefinition(1, Label = "Include Adult", Type = FieldType.Checkbox, Advanced = true)]
+    public bool IncludeAdult { get; set; }
```
markus101: Is this actually valid for TV shows?

src/NzbDrone.Core/ImportLists/Tmdb/Discover/TmdbDiscoverSettings.cs
```diff
@@ -0,0 +1,84 @@
...
+
+        RuleFor(c => c.WithKeywords).Matches(WithQueryRegex)
+            .When(c => c.WithKeywords.IsNotNullOrWhiteSpace());
+
+        RuleFor(c => c.WithCompanies).Matches(WithQueryRegex)
+            .When(c => c.WithCompanies.IsNotNullOrWhiteSpace());
+    }
+
+    [GeneratedRegex(@"^\d+(?:[,|]\d+)*$")]
+    private static partial Regex WithQueryRegex { get; }
```
markus101: I don't think this complication is necessary, a compiled regex should be just fine, also the naming is overly generic for what it actually does.

src/NzbDrone.Core/ImportLists/Tmdb/Discover/TmdbDiscoverSort.cs
```diff
@@ -0,0 +1,11 @@
+﻿namespace NzbDrone.Core.ImportLists.Tmdb.Discover;
+
+public enum TmdbDiscoverSort
+{
+    First_Air_Date = 0,
```
markus101:
```suggestion
    FirstAirDate = 0,
```

markus101 (changes requested):
The redirect wasn't working due to an issue with services (wasn't passing the correct parameter to TMDb), I've pushed a fix for that. Could have saved a bunch of time not implementing a workaround we'd never have shipped (which actually didn't work without a manually added token).

> Frontend is like voodoo to me man, unsure how to change things there tbh, but I've gone ahead and applied the rename to all types/files.

In this case it's just one file, `typeOfList: titleCase(key),` just needs to be changed, but I've pushed a fix for that.

Be worth a self review to make sure it's following existing patterns and not trying to be too smart (which just ends up being complicated).

Lots of things not being translated properly, enums not following our standard naming (should be PascalCase, not Snake_Case)

src/NzbDrone.Core/ImportLists/Tmdb/Discover/TmdbDiscoverSettings.cs
mynameisbogdan: It's a valid query parameter, but should not be used...

src/NzbDrone.Core/ImportLists/Tmdb/TmdbRequestGeneratorBase.cs
ArachisH: You're correct just double checked, no need to explicitly set the default language here.

src/NzbDrone.Core/ImportLists/Tmdb/TmdbImportBase.cs
ArachisH: Parsing of the JWT object no longer required, was using it to check if the request token was generated with the redirection claim. The changes you made on the backend alleviated this issue(thanks), so it makes sense to remove it now.

Updated In: https://github.com/Sonarr/Sonarr/pull/8598/changes/eed469b4fc0f239307ebaac8d0e8934201683582

src/NzbDrone.Core/ImportLists/Tmdb/TmdbAPI.cs
markus101: If it's just a difference of casing then these aren't needed, but these are probably better in a separate file as they aren't coming from TMDb.

frontend/src/Settings/ImportLists/ImportLists/AddImportListModalContent.tsx
```diff
@@ -20,6 +20,17 @@ export interface AddImportListModalContentProps {
   onModalClose: () => void;
 }
 
+function getListGroupTitle(typeOfList: string) {
+  // Does not need to be translated as it is a proper name.
+  if (typeOfList === 'tmdb') {
+    return 'TMDb';
```
ArachisH, line 26: Should I change this to return 'TMDb List' instead?

The other list types show as '{Provider} List', and since it returns early here it only spits out '{Provider}' in the web UI.

src/NzbDrone.Core/ImportLists/Tmdb/Discover/TmdbLanguageOptionsConverter.cs
ArachisH: Brought over the supported TMDb languages from their API into **TmdbLanguage.cs**, and had to add 'Any' as a custom option to use as the default/null value.

src/NzbDrone.Core/ImportLists/Tmdb/List/TmdbListSettings.cs
```diff
@@ -0,0 +1,45 @@
...
+
+namespace NzbDrone.Core.ImportLists.Tmdb.List;
+
+public class TmdbListSettingsValidator : TmdbSettingsBaseValidator<TmdbListSettings>
+{
+    public TmdbListSettingsValidator()
+    {
+        RuleFor(c => c.ListId).Must(id => int.TryParse(id, out var idInt) && idInt > 0)
+            .WithMessage($"Must be a valid 32-bit integer greater than zero, and less than or equal to {int.MaxValue}.")
+            .When(c => c.ListId.IsNotNullOrWhiteSpace() && c.AccountListId.IsNullOrWhiteSpace());
```
mynameisbogdan: Why not just use nullable ints?

```suggestion
        RuleFor(c => c.ListId)
            .GreaterThan(0)
            .When(c => c.ListId.HasValue && !c.AccountListId.HasValue);
```

src/NzbDrone.Core/ImportLists/Tmdb/Person/TmdbPersonParser.cs
```diff
@@ -0,0 +1,47 @@
...
+
+public class TmdbPersonParser : TmdbParserBase<TmdbCreditsResource>
+{
+    private readonly bool _isIncludingCastCredits;
+    private readonly HashSet<string> _departments;
+
+    public TmdbPersonParser(TmdbPersonSettings settings)
+    {
+        _isIncludingCastCredits = settings.IncludingCastCredits;
+        _departments = GetDepartmentNames(settings.IncludeDepartmentTypes);
```
mynameisbogdan: I would just use the settings object, and allow GetDepartmentNames to be run on execution time, not on constructor. I don't think it's going to be such a performance boost.

src/NzbDrone.Core/ImportLists/Tmdb/TmdbAPI.cs
```diff
@@ -0,0 +1,151 @@
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Text.Json.Serialization;
+
+namespace NzbDrone.Core.ImportLists.Tmdb;
+
+[DebuggerDisplay("Page: {Page}/{TotalPages}, Page Results: {Results.Count,n0}")]
```
mynameisbogdan: Many fields are not being used, loading them into memory would not add much value IMO.

frontend/src/Settings/ImportLists/ImportLists/AddImportListModalContent.tsx
markus101: Actually my oversight from before, I think this should return `TMDb`, but we should just return `titleCase(typeOfList)` for other lists, then translate in the JSX like before, otherwise we end up having to duplicate the translation.

src/NzbDrone.Core/ImportLists/Tmdb/Discover/TmdbDiscoverSettings.cs
```diff
@@ -0,0 +1,117 @@
...
+        RuleFor(c => c.AirDateMaximum).Must(ValidateAirDate)
+            .WithMessage("Must be in the format: yyyy-MM-dd");
+
+        RuleFor(c => c.WithKeywords).Matches(AndOrDelimitedIdsRegex)
+            .When(c => c.WithKeywords.IsNotNullOrWhiteSpace());
+
+        RuleFor(c => c.WithCompanies).Matches(AndOrDelimitedIdsRegex)
+            .When(c => c.WithCompanies.IsNotNullOrWhiteSpace());
+
+        RuleFor(c => c.WithNetworks).Matches(AndDelimitedIdsRegex)
```
markus101, line 33: Should this just be a list then? Since there isn't `,` and `|` to take into account we can just store as a list and format the query correctly.

src/NzbDrone.Core/ImportLists/Tmdb/Discover/TmdbDiscoverSettings.cs
```diff
@@ -0,0 +1,117 @@
...
+        }
+
+        return DateTime.TryParseExact(airDate,
+            "yyyy-MM-dd",
+            CultureInfo.InvariantCulture,
+            DateTimeStyles.None,
+            out _);
+    }
+
+    private static void ValidateVoteAverage(string voteAverage, CustomContext context)
```
markus101, line 51: Does this really need to be a custom validator? There are built-in validators for comparing numbers. Could also have some cultures based issues, `7,5` would be parse correctly in some cultures, but if TMDb expects `7.5` it'd fail.

src/NzbDrone.Core/ImportLists/Tmdb/TmdbRequestGeneratorBase.cs
```diff
@@ -0,0 +1,44 @@
...
+        var pageableRequests = new ImportListPageableRequestChain();
+        pageableRequests.Add(GetSeriesRequests());
+        return pageableRequests;
+    }
+
+    protected abstract HttpRequestBuilder CreateSeriesRequestsBuilder();
+
+    private IEnumerable<ImportListRequest> GetSeriesRequests()
+    {
+        var builder = CreateSeriesRequestsBuilder();
```
markus101, line 30: Nit: Blank line before if:


```suggestion
        var builder = CreateSeriesRequestsBuilder();

```

src/NzbDrone.Core/ImportLists/Tmdb/TmdbRequestGeneratorBase.cs
```diff
@@ -0,0 +1,44 @@
...
+        if (MaxPages > 0)
```
markus101, line 31: This feels like a weird level of abstraction to apply here, why do we need another `GeneratorBase` to apply the very limited logic from here?

src/NzbDrone.Core/ImportLists/Tmdb/Discover/TmdbDiscoverSettings.cs
```diff
@@ -0,0 +1,117 @@
...
+    public int OriginalLanguage { get; set; }
+
+    [FieldDefinition(2, Label = "ImportListsTmdbSettingsSortByType", HelpText = "ImportListsTmdbSettingsSortByTypeHelpText", Type = FieldType.Select, SelectOptions = typeof(TmdbDiscoverSortByType))]
+    public int SortByType { get; set; }
+
+    [FieldDefinition(3, Label = "ImportListsTmdbSettingsVoteAverageMinimum", HelpText = "ImportListsTmdbSettingsVoteAverageMinimumHelpText", Type = FieldType.Textbox)]
+    public string VoteAverageMinimum { get; set; }
+
+    [FieldDefinition(4, Label = "ImportListsTmdbSettingsVoteCountMinimum", HelpText = "ImportListsTmdbSettingsVoteCountMinimumHelpText", Type = FieldType.Textbox, Advanced = true)]
+    public string VoteCountMinimum { get; set; }
```
markus101, line 95: Should validate that this is positive.

src/NzbDrone.Core/ImportLists/Tmdb/Person/TmdbPersonSettings.cs
```diff
@@ -0,0 +1,44 @@
...
+using FluentValidation;
+using NzbDrone.Core.Annotations;
+
+namespace NzbDrone.Core.ImportLists.Tmdb.Person;
+
+public class TmdbPersonSettingsValidator : TmdbSettingsBaseValidator<TmdbPersonSettings>
+{
+    public TmdbPersonSettingsValidator()
+    {
+        RuleFor(c => c.PersonId).Must(id => int.TryParse(id, out var idInt) && idInt > 0)
```
markus101, line 12: Don't use custom validation here, also `32-bit integer` is not going to be very useful for users

src/NzbDrone.Core/ImportLists/Tmdb/Person/TmdbPersonSettings.cs
```diff
@@ -0,0 +1,44 @@
...
+
+    public TmdbPersonSettings()
+        : base(Validator)
+    {
+        IncludingCastCredits = true;
+        IncludeDepartmentTypes = [];
+    }
+
+    [FieldDefinition(1, Label = "ImportListsTmdbSettingsPersonId", HelpText = "ImportListsTmdbSettingsPersonIdHelpText", Type = FieldType.Textbox)]
+    public string PersonId { get; set; }
```
markus101, line 37: Make it an int and we can do proper validation.

src/NzbDrone.Core/ImportLists/Tmdb/TmdbParserBase.cs
```diff
@@ -0,0 +1,63 @@
...
+    private static bool PreProcess(ImportListResponse importListResponse)
+    {
+        if (importListResponse.HttpResponse.StatusCode != HttpStatusCode.OK)
+        {
+            throw new ImportListException(importListResponse,
+                $"TMDb API(${importListResponse.Request.Url.Path}) call resulted in an unexpected StatusCode [{importListResponse.HttpResponse.StatusCode}]");
+        }
+
+        if (importListResponse.HttpResponse.Headers.ContentType != null && !importListResponse.HttpResponse.Headers.ContentType.Contains("text/json", StringComparison.OrdinalIgnoreCase) &&
+            importListResponse.HttpRequest.Headers.Accept != null && importListResponse.HttpRequest.Headers.Accept.Contains("text/json", StringComparison.OrdinalIgnoreCase))
```
markus101, line 55: I think we want `application/json` instead of `text/json`, since we're setting the content type to that.

markus101 (commented):
I think a lot of the custom validation isn't necessary.

src/NzbDrone.Core/ImportLists/Tmdb/TmdbParserBase.cs
mynameisbogdan: Indeed, there's no `text/json` AFAIK.

`curl --header 'accept: text/json' -I` still returns `application/json`.

```
HTTP/2 200
content-type: application/json;charset=utf-8
date: Sat, 20 Jun 2026 19:35:54 GMT
content-encoding: gzip
```

PR #8614, opened 2026-05-12
Upgrade React Router from v5 to v7 (data router)

bertPB:
#### Description

- Migrates all route definitions from `component={X}` / `render={fn}` to the v7 `element={<X />}` API
- Replaces `<BrowserRouter>` + legacy `<Switch>` with `createBrowserRouter` + `RouterProvider` (React Router "data router" pattern), required for `useBlocker` support in `SettingsToolbar`
- Converts nested child routes in `Logs.tsx` and `ImportSeriesPage.tsx` from absolute paths to relative/`index` paths as required by the data router
- Minor lint fixes (`index={true}`, `replace={true}` explicit boolean props, import ordering)

This is a pure infrastructure change, no visual differences, no new features.

#
Part 1 of 12 in the Sonarr v5 UI redesign split


markus101:
Probably best if Claude doesn't open the PR and stomp all over the template. Please update the description to follow the template.

bertPB:
@markus101 What further details do you want me to provide (or remove)?

Want me to write out a specific reasoning for this change?

Edited the header to match the template, apologies for that oversight.

package.json
```diff
@@ -68,8 +66,8 @@
     "react-lazyload": "3.2.1",
     "react-measure": "1.4.7",
     "react-redux": "7.2.4",
-    "react-router": "5.2.0",
-    "react-router-dom": "5.2.0",
+    "react-router": "^7.6.0",
+    "react-router-dom": "^7.6.0",
```
mynameisbogdan: We prefer them pinned to a specific version.

bertPB: Fair enough, but all deps are pinned with the .lock file , right?
But if this is convention for the repo, I don't mind changing it there as well 👍

mynameisbogdan: It's not the same when they could be bumped on yarn upgrade or a lock rebuild (which I just did on #8612 to bump transitive dependencies with vulnerabilities).

frontend/src/App/App.tsx
```diff
@@ -1,30 +1,48 @@
...
+      </Page>
+    </>
+  );
+}
+
+const router = createBrowserRouter(
+  createRoutesFromElements(
+    <Route element={<PageLayout />}>{appRouteElements()}</Route>
+  ),
+  { basename: window.Sonarr.urlBase || undefined }
```
mynameisbogdan, line 37: Wouldn't be undefined anyway in case of not being set? 

```suggestion
  { basename: window.Sonarr.urlBase }
```

frontend/src/AddSeries/ImportSeries/Import/ImportSeries.tsx
```diff
@@ -21,7 +21,7 @@ function ImportSeries() {
   const { rootFolderId: rootFolderIdString } = useParams<{
     rootFolderId: string;
   }>();
-  const rootFolderId = parseInt(rootFolderIdString);
+  const rootFolderId = parseInt(rootFolderIdString ?? '');
```
mynameisbogdan: How about? 

```js
  const { rootFolderId: rootFolderIdString = '' } = useParams<{
    rootFolderId: string;
  }>();
  const rootFolderId = parseInt(rootFolderIdString);
```

frontend/src/App/App.tsx
bertPB: This is to handle cases where urlBase could be an empty string, this makes it fallback to undefined instead

frontend/src/AddSeries/ImportSeries/Import/ImportSeries.tsx
bertPB: That's a bit cleaner indeed. Will adjust

frontend/src/Components/Page/Sidebar/PageSidebar.tsx
```diff
@@ -231,10 +231,7 @@ function PageSidebar() {
     transform: isSidebarVisible ? 0 : SIDEBAR_WIDTH * -1,
   });
 
-  const urlBase = window.Sonarr.urlBase;
-  const pathname = urlBase
-    ? location.pathname.substr(urlBase.length) || '/'
-    : location.pathname;
+  const pathname = location.pathname;
```
mynameisbogdan:
```suggestion
  const { pathname } = location;
```

frontend/src/App/App.tsx
mynameisbogdan: Ah, right. 👍

package.json
```diff
@@ -68,8 +66,8 @@
     "react-lazyload": "3.2.1",
     "react-measure": "1.4.7",
     "react-redux": "7.2.4",
-    "react-router": "5.2.0",
-    "react-router-dom": "5.2.0",
+    "react-router": "7.6.0",
+    "react-router-dom": "7.6.0",
```
mynameisbogdan: https://www.npmjs.com/advisories/1112051
https://www.npmjs.com/advisories/1112055
https://www.npmjs.com/advisories/1112062
https://www.npmjs.com/advisories/1112051
https://www.npmjs.com/advisories/1112053
https://www.npmjs.com/advisories/1112056
https://www.npmjs.com/advisories/1112062


```suggestion
    "react-router": "7.15.1",
    "react-router-dom": "7.15.1",
```

mynameisbogdan (changes requested):

package.json
mynameisbogdan: Btw you'll need to update the yarn.lock with a yarn install as I can't code suggest this chance. 😅

frontend/src/AddSeries/ImportSeries/Import/ImportSeries.tsx
```diff
@@ -18,7 +18,7 @@ import { clearImportSeries } from './importSeriesStore';
 import ImportSeriesTable from './ImportSeriesTable';
 
 function ImportSeries() {
-  const { rootFolderId: rootFolderIdString } = useParams<{
+  const { rootFolderId: rootFolderIdString = '' } = useParams<{
     rootFolderId: string;
```
markus101, line 22: If we need to set a default the type we're passing to `useParams` is wrong

```suggestion
    rootFolderId?: string;
```

frontend/src/Components/Router/Switch.tsx
```diff
@@ -1,38 +1,12 @@
-import React, { Children, ReactElement, ReactNode } from 'react';
-import { Switch as RouterSwitch } from 'react-router-dom';
-import getPathWithUrlBase from 'Utilities/getPathWithUrlBase';
```
mynameisbogdan, line 3: `getPathWithUrlBase` is not longer in use and could be just easily removed, right?

bertPB: Good call, removed it.

markus101 (approved):
Just one small suggestion to keep the type consistent, otherwise looks good.

frontend/src/AddSeries/ImportSeries/Import/ImportSeries.tsx
mynameisbogdan: I don't think it's needed, it's just useParams not having mandatory returns.

Screenshot 2026-05-16 at 21 32 01

markus101: The type is saying the value is always a string, but it's not it can sometimes be undefined.

bertPB: I have to agree with @mynameisbogdan here, rootFolderId is always a string, but useParams makes it inherently "optional" . So I would keep it as is.

markus101: My mistake I was comparing an old version of the typings where it wasn't a partial. Agreed it shouldn't be optional in the type.

bertPB:
Never contributed to a public repo like this.
I assume one of the contributors needs to merge it, even if it has been approved?

markus101:
Correct, contributors can't force things to be merged whether it's approved or not.

markus101:
Thanks for this, as discussed on Discord previously I've updated the `v5-redesign` branch for the followup work.

PR #8625, opened 2026-05-15
Update server to send full certificate chain when provided

tamwahba:
#### Description
This PR updates the server to support sending a full certificate chain when it's provided in the pfx or pem file.

This PR is similar to one I had opened in the Radarr: https://github.com/Radarr/Radarr/pull/11448.

Fixes an issue similar to Radarr's https://github.com/Radarr/Radarr/issues/11330


src/NzbDrone.Common/Http/SslCertificateLoader.cs
```diff
@@ -0,0 +1,60 @@
...
+using System.Linq;
+using System.Net.Security;
+using System.Security.Cryptography.X509Certificates;
+using NzbDrone.Common.Extensions;
+
+namespace NzbDrone.Common.Http
+{
+    public class SslCertificateLoadException : Exception
+    {
+        public SslCertificateLoadException(string message)
```
mynameisbogdan: This should be to a separate file, and while it's a new file this could use file-scoped namespace.

src/NzbDrone.Host/Bootstrap.cs
```diff
@@ -279,26 +286,13 @@ private static string BuildUrl(string scheme, string bindAddress, int port)
...
-                }
-                else if (type == X509ContentType.Pkcs12)
-                {
-                    certificate = X509CertificateLoader.LoadPkcs12FromFile(cert, password, X509KeyStorageFlags.DefaultKeySet);
-                }
-                else
-                {
-                    throw new SonarrStartupException($"Invalid certificate type: {type}");
-                }
+                certificateContext = SslCertificateLoader.LoadCertificateContext(cert, key, password);
```
mynameisbogdan: Considering we're catching the global exception, we should be able to return directly from the try. Unless it's an anti-pattern of sorts.

```suggestion
                return SslCertificateLoader.LoadCertificateContext(cert, key, password);
```

src/NzbDrone.Common/Http/SslCertificateLoader.cs
```diff
@@ -0,0 +1,60 @@
+using System;
+using System.Linq;
+using System.Net.Security;
+using System.Security.Cryptography.X509Certificates;
+using NzbDrone.Common.Extensions;
+
+namespace NzbDrone.Common.Http
+{
+    public class SslCertificateLoadException : Exception
```
mynameisbogdan: How about extending CryptographicException instead, which would allow us to maintain `catch (CryptographicException)` calls?

```suggestion
    public class SslCertificateLoadException : CryptographicException
```

mynameisbogdan (changes requested):

markus101:
Please let us know when this is ready for another look.

tamwahba:
Thank you for the review, I've made the requested changes

src/NzbDrone.Common/Http/SslCertificateLoadException.cs
```diff
@@ -0,0 +1,10 @@
+using System.Security.Cryptography;
+
+namespace NzbDrone.Common.Http;
+public class SslCertificateLoadException : CryptographicException
```
mynameisbogdan, line 5:
```suggestion

public class SslCertificateLoadException : CryptographicException
```

New line before class definition.

src/NzbDrone.Common/Http/SslCertificateLoader.cs
```diff
@@ -0,0 +1,49 @@
+using System.Linq;
+using System.Net.Security;
+using System.Security.Cryptography.X509Certificates;
+using NzbDrone.Common.Extensions;
+
+namespace NzbDrone.Common.Http;
```
mynameisbogdan, line 6:
```suggestion
namespace NzbDrone.Common.Http;

```

mynameisbogdan:
Out of curiosity, you think it's possible to have 1-2 tests to avoid breaking this functionality in the future? Self-signed certs generated on every run should be good too.

tamwahba:
> Out of curiosity, you think it's possible to have 1-2 tests to avoid breaking this functionality in the future? Self-signed certs generated on every run should be good too.

Yeah, I think that's doable, will update this PR with some tests. Moving back to draft for now

tamwahba:
I've added some tests, please let me know if there's any other changes needed.

src/NzbDrone.Api.Test/v3/Config/CertificateValidatorTests.cs
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    [OneTimeTearDown]
+    public void DeleteTestCertificates()
+    {
+        _certs?.Dispose();
+    }
+
+    [Test]
+    public void Validate_fails_when_cert_path_is_null()
```
markus101: Test names should be lower case.

```suggestion
    public void validate_fails_when_cert_path_is_null()
```

src/Sonarr.Api.V5/Settings/CertificateValidator.cs
```diff
@@ -3,7 +3,7 @@
 using FluentValidation;
```
markus101: An oversight from before, but we should extract this to `Sonarr.Http.Validation` so we don't have multiple copies and have to test it multiple times.

src/NzbDrone.Test.Common/SslTestCertificates.cs
```diff
@@ -0,0 +1,76 @@
+using System;
+using System.IO;
+using System.Security.Cryptography;
+using System.Security.Cryptography.X509Certificates;
+
+namespace NzbDrone.Test.Common;
+
+public sealed class SslTestCertificates : IDisposable
```
markus101:
```suggestion
public class SslTestCertificates : IDisposable
```

markus101: Also worth moving to `Sonarr.Http.Test` once the validator is de-duped?

markus101 (commented):
It'd be good to clean up the duplication being done with the validator so we don't omit one API version by mistake in the future with a related change.

src/NzbDrone.Test.Common/SslTestCertificates.cs
tamwahba: I kept it as-is so that it could be used for the API smoke tests. Please let me know if that looks good to you!

src/NzbDrone.Host/Bootstrap.cs
```diff
@@ -199,7 +200,13 @@ public static IHostBuilder CreateConsoleHostBuilder(string[] args, StartupContex
                         {
                             options.ConfigureHttpsDefaults(configureOptions =>
                             {
-                                configureOptions.ServerCertificate = ValidateSslCertificate(sslCertPath, sslKeyPath, sslCertPassword);
+                                var sslContext = ValidateSslCertificate(sslCertPath, sslKeyPath, sslCertPassword);
+
+                                configureOptions.ServerCertificate = sslContext.TargetCertificate;
+                                configureOptions.OnAuthenticate = (context, authOptions) =>
```
mynameisbogdan:
```
Parameter 'context' is never used
Parameter 'context' hides outer parameter with the same name
```

```suggestion
                                configureOptions.OnAuthenticate = (_, authOptions) =>
```

src/Sonarr.Http/Validation/CertificateValidator.cs
```diff
@@ -0,0 +1,55 @@
...
+using NzbDrone.Common.Http;
+using NzbDrone.Common.Instrumentation;
+
+namespace Sonarr.Http.Validation;
+
+public class CertificateValidator : PropertyValidator
+{
+    protected override string GetDefaultMessageTemplate() => "Invalid SSL certificate file or {passwordOrKey}. {message}";
+
+    private static readonly Logger Logger = NzbDroneLogger.GetLogger(typeof(CertificateValidator));
```
mynameisbogdan:
```
SA1201: A field should not follow a method
```

```suggestion
    private static readonly Logger Logger = NzbDroneLogger.GetLogger(typeof(CertificateValidator));

    protected override string GetDefaultMessageTemplate() => "Invalid SSL certificate file or {passwordOrKey}. {message}";
```

mynameisbogdan (changes requested):

tamwahba:
Would love a re-review when you have time!

src/NzbDrone.Api.Test/v3/Config/CertificateValidatorTests.cs
```diff
@@ -0,0 +1,46 @@
...
+using FluentValidation.TestHelper;
+using NUnit.Framework;
+using NzbDrone.Test.Common;
+using Sonarr.Api.V3.Config;
+using Sonarr.Http.Validation;
+
+namespace NzbDrone.Api.Test.v3.Config;
+
+[TestFixture]
+public class CertificateValidatorTests : TestBase
```
markus101: I don't see much value in the v3/v5 level tests, if we want tests at the API level they should be full integration tests instead of just testing a validator that's created just for testing (doesn't actually test that the validator is implemented in the controller properly).

markus101 (commented):
I think just dropping the v3/v5 specific tests is fine as the integration tests for this is much more complicated.

src/NzbDrone.Api.Test/v3/Config/CertificateValidatorTests.cs
tamwahba: ok, that's fair. I was using them as a smoke test that the different api resources conform to `ISsslCertificateResource` but that's not necessary either. I've removed those tests.

markus101 (approved):
Thanks for implementing and all the changes.

markus101:
@tamwahba looks like a failing test needs to get fixed before merging.

tamwahba:
I've fixed the test for windows. Thank you for the reviews!

PR #8627, opened 2026-05-16
Quality profile size limits improvements

markus101:
#### Description
There was an issue with the 207 migration that affected existing profiles so fixing that up, but also added feedback in the UI. Added a little badge if you're not in the `Edit Sizes` view already.

If you're running a DB that has migrated, but still has issues you should be able to run the follow to fix your profiles.

```
UPDATE QualityProfiles SET Items = REPLACE(Items, '"maxSize": 0', '"maxSize": null');
```

image
image


src/NzbDrone.Core/Datastore/Migration/207_add_size_to_quality_profiles.cs
```diff
@@ -154,29 +154,35 @@ private Dictionary<int, Definition207> GetSizes()
             using (var getDefinitionsCmd = _connection.CreateCommand())
             {
                 getDefinitionsCmd.Transaction = _transaction;
-                getDefinitionsCmd.CommandText = "SELECT \"Id\", \"MinSize\", \"MaxSize\", \"PreferredSize\" FROM \"QualityDefinitions\"";
+                getDefinitionsCmd.CommandText = "SELECT \"Quality\", \"MinSize\", \"MaxSize\", \"PreferredSize\" FROM \"QualityDefinitions\"";
```
mynameisbogdan, line 157: The initial migration didn't worked? Do we still have a way to reset these values to default without recreation?

markus101: If your max size was null (infinite), it didn't properly convert that during the migration and ended up being stored as 0 in the DB. You can fix it with the SQL in the description or just manually adjust the max size in while editing.

mynameisbogdan: I mostly meant the difference between `QualityDefinitions.Id` and `QualityDefinitions.Quality`, as theoretically is didn't migrated anything, correct?

Looking at the db the fields are populated, I might be wrong but the preferred sizes seems way to high for what I was under the impression I was using.

markus101: Ahh, in that case migrating a back up and comparing is probably the best way to see what went wrong.

mynameisbogdan: Screenshot 2026-05-17 at 17 57 45

This is what I mean, `Id` != `Quality`. 

Should be okay after this PR is merged, but we won't have a way to reset them like we did in v4, right?

markus101: Yeah, no reset right now, I've been thinking of that + copying though.

PR #8632, opened 2026-05-17
Foundation for the redesign.

bertPB:
#### Description
  Foundation for the actual redesign.
  - Design tokens
  - Fonts
  - General updated styling on shared form/modal/menu/table components.
   Adds new `FormRow`/`FormInput` primitives to replace the old `FormGroup`/`FormInputGroup` (these stay for now and will be migrated out later
   
   ---
   **Things like this are to be expected and will be fixed in later PRs**

localhost_8989_settings_mediamanagement


frontend/src/Components/Label.css
```diff
@@ -1,122 +1,93 @@
...
-  &.outline {
-    color: var(--infoColor);
-  }
-}
+/** Kinds — dot color via custom property, bg stays paper-warm. **/
 
+.danger::before { background-color: var(--dangerColor); opacity: 1; }
+.default::before { background-color: var(--inkMuted); opacity: 0.6; }
+.disabled::before { background-color: var(--inkMuted); opacity: 0.4; }
+.info::before { background-color: var(--infoColor); opacity: 1; }
```
markus101: What's the benefit/reason for using the pseudo element?

frontend/src/Helpers/Props/icons.ts
```diff
@@ -86,140 +86,140 @@ import {
...
-export const TEST                   = FlaskConical;
-export const UNGROUP                = Ungroup;
-export const UNKNOWN                = CircleHelp;
-export const UNMONITORED            = BookmarkX;
-export const UPDATE                 = RefreshCcw;
-export const UNSAVED_SETTING        = CircleDot;
-export const VIEW                   = Eye;
-export const WARNING                = TriangleAlert;
-export const WIKI                   = BookOpen;
+export const ACTIONS = Zap;
```
markus101, line 103: The linter pass should be applied to the appropriate commits instead of a separate commit.

1ad088c80b7054f0f1f8b22172ffdcbc7652ddbc adds a bunch of extra spacing, only for it to be removed a couple commits later in 634222f0656ebd7dad2dc014b2471bd62f7ce07e

frontend/src/Components/Page/Toolbar/PageToolbarButton.css
markus101: This will probably get covered later, but since we're already editing the file probably worth changing now. `toobarButtonHoverColor` has been removed from the dark and light theme files, what do we need to replace this with?

frontend/src/Components/Page/Toolbar/PageToolbarButton.css
```diff
@@ -2,7 +2,7 @@
   composes: link from '~Components/Link/Link.css';
 
   padding-top: 4px;
-  width: $toolbarButtonWidth;
+  width: 100px;
```
markus101: `dimensions.js` updates `toolbarButtonWidth` to `100px`, but then it's also hardcoded here. Seems like this should still use the variable, unless there is another reason for hard coding?

markus101 (commented):
Just a few things I've noticed, should be good to land once they're addressed.

frontend/src/Helpers/Props/icons.ts
bertPB: Yeah, the resquashing of all my commits for this clearly didn't work 100%. 
I guess it doesn't matter too much since we Squash PR's on merge right?

frontend/src/Components/Page/Toolbar/PageToolbarButton.css
bertPB: Hardcoded to keep some legacy components in check. ( I think )
Working on an addition to this PR that will likely remove the need for this

frontend/src/Components/Page/Toolbar/PageToolbarButton.css
bertPB: Themes will be handled completely differently, so best not to pay too much attention to these for now.
Good remark though!

frontend/src/Components/Label.css
bertPB: It adds a small dot to the label components without the need to add an extra div to the component itself.

frontend/src/Components/Form/FormInputGroup.css
```diff
@@ -9,6 +9,37 @@
...
+    border-color 160ms cubic-bezier(0.25, 1, 0.5, 1),
+    box-shadow 160ms cubic-bezier(0.25, 1, 0.5, 1);
+}
+
+.hasButtons:focus-within {
+  border-color: var(--accent);
+  box-shadow: 0 0 0 4px var(--inputFocusBoxShadowColor);
+}
+
+.hasButtons.hasError {
```
mynameisbogdan:
```suggestion
&.hasError {
```

Plus other places.

bertPB: Will do full pass to change it to the &.xyz notation

frontend/src/Components/Page/Toolbar/PageToolbarButton.css
markus101: Not sure I follow the reasoning, both are the same value of `100px`, this just also hardcodes it.

bertPB: Looked at my pre-squash commit history, this was changed on an intermediary (that's a word right?) commit where I also removed all post-css variables.
I'll put it back in this pr for now, because as it stands it indeed doesn't make any sense.

bertPB:
Addressed your comments + included JS => CSS theme migration and proper color-impaired color set :)

frontend/src/Helpers/Hooks/useTheme.ts
```diff
@@ -47,10 +46,19 @@ const useTheme = (): 'dark' | 'light' => {
...
-  return themeVariables[color];
+  useEffect(() => {
+    setValue(
+      getComputedStyle(document.documentElement)
+        .getPropertyValue(`--${color}`)
+        .trim()
+    );
+  }, [color, theme]);
+
+  return value;
```
markus101: We can't do this with `useMemo`? With the current implementation every time we call `useThemeColor` we're going to get back an empty string on the first render, then it'll re-render afterwards with the correct value, so we're forcing 2x the renders even if the DOM is already applied to the dom.

```suggestion
  return useMemo(() => {
      return getComputedStyle(document.documentElement)
        .getPropertyValue(`--${color}`)
        .trim()
  }, [color, theme]);
```

frontend/src/Components/Link/Button.css
```diff
@@ -1,125 +1,116 @@
+
```
markus101: Extra new line

frontend/src/Components/Form/FormInputButton.css
```diff
@@ -1,12 +1,27 @@
...
+  &:hover {
+    background-color: var(--paperCard);
+    color: var(--accent);
+  }
 }
 
 .middleButton {
-  border-top-right-radius: 0;
-  border-bottom-right-radius: 0;
+  /* No-op — container handles outer radius via overflow. */
```
markus101, line 26: Can we just remove this class then?

frontend/src/Helpers/Props/icons.ts
```diff
@@ -1,242 +1,225 @@
...
+export const FLAG = Flag;
+export const FOOTNOTE = Asterisk;
+export const FOLDER = Folder;
+export const FOLDER_OPEN = FolderOpen;
+export const GENRE = Drama;
+export const GLOBE = Globe;
+export const GROUP = Group;
+export const HEALTH = HeartPulse;
+export const HEART = Heart;
+export const HEART_OUTLINE = Heart;
```
markus101: What's the plan for outlined icons line this and `CIRCLE_OUTLINE` they're pointing to the filled in icon now.

package.json
markus101: Can we remove these if we're no longer using Font Awesome?

markus101 (commented):
`AdvancedSettingsButton` is still applying Font-Awesome classes, but we're no longer using the actual fonts.

frontend/src/Components/Page/Page.tsx
```diff
@@ -64,6 +64,10 @@ function Page({ children }: PageProps) {
     }
   }, [isUpdated]);
 
+  useEffect(() => {
+    document.body.classList.toggle('colorImpaired', enableColorImpairedMode);
```
mynameisbogdan: Are we switching from ColorImpairedContext to a global body class?

bertPB: Indeed, any objections to that?

package.json
bertPB: Yes! Will remove

frontend/src/Components/Form/FormInputButton.css
bertPB: Will be modified later.
Would need to check, but I'm quite sure i left it there because it may still be assigned in some places.

frontend/src/Helpers/Props/icons.ts
bertPB: Added a "filled" prop to <Icon/> that should handle this in most cases.

frontend/src/Components/Link/Button.css
bertPB: Forgot to address this, but I feel like things like this aren't that relevant on these intermediate PR's.
Still a valid point ofc

frontend/src/Helpers/Hooks/useTheme.ts
bertPB: Replaced the implementation with something that works for sure now.
useMemo would've had the same re-render issue since it fires after initial render.

frontend/src/Components/Link/Button.css
markus101: That's the point of the intermediate PRs though, each one builds the foundation for the next, but they shouldn't introduce unnecessary changes that get cleaned up later. Iterative, not break then fix later.

Some of these things will probably be caught by the linters, which isn't currently functional, but I'll try to get the fix for that landed today.

frontend/src/Helpers/Hooks/useTheme.ts
markus101: It doesn't though, it runs synchronously during render and memoizes the result so it doesn't re-run unless the dependencies change.

We're not currently enforcing this lint rule, but something we will eventually, so makes sense not to introduce new things that break the rule.
https://react.dev/reference/eslint-plugin-react-hooks/lints/set-state-in-effect

frontend/src/Components/Page/Page.tsx
mynameisbogdan: I'm not a react developer by any means, but I know altering the `document.body.classList` it's considered a side effect in react. In same thought we could add `theme-light` or `theme-dark`, or `language-en` to it for state.

frontend/src/Helpers/Hooks/useTheme.ts
bertPB: I looked into it a bit more, and while useMemo would indeed solve the "no side effects" problem, it would simply break the initial theme application.

I'm working on a solution with useLayoutEffect which looks like the best of both worlds. Change coming soon

frontend/src/Components/Page/Page.tsx
bertPB: We apply themes in the exact same way, it's (as far as I know) a quite common pattern in other frameworks as well to do it like this.

While side-effects in React are to be avoided, they are not that big of a problem in the case of a simple toggle like this, which in reality will barely be triggered.

I will however push a change so that instead of a .colorImpaired class, we use a data attr just like the themes, to bring those in line with each other 👍

frontend/src/Helpers/Hooks/useTheme.ts
bertPB: Resolving this convo right now, feel free to re-open if you want to discuss this more :)

bertPB:
@markus101 Would be nice to see this one merged so I can open the next PR :D 
If there's no more issues here ofc.

mynameisbogdan:
Can you squash the commits? I'll like to test it locally.

bertPB:
@mynameisbogdan the commits just for this PR? Or? 
Because not sure I understand what squashing the commits has to do with running it locally 😓 


EDIT: Squashed them

markus101 (approved):
Looking good.

PR #8642, opened 2026-05-19
Allow reordering of non-modifiable table columns

### LINKED ISSUE #7897: Put Actions column next to Status column

I think this one's really obvious - put the things that matter and are relevant to each other - next to eachother.
Or at least make them configurable, so I can opt-in to a design that doesn't make my wrist hurt.

I'm going through a queue of 60 items, a lot of which need to be manually imported, which is already an insufferable detail.

I'm slowly moving my cursor back and forth, over and over, from one side to the other of my 34 inch monitor. I can't move the status column or the action column, so I'm stuck doing this.

Image

Yes, I can open using the manual import option and THEN observe it, but some of the status columns don't mean I want to import, e.g. file is of lower quality, so delete, not import. I'd rather see the status or reason for manual import not being done WITHOUT having to slowly open up the dialogue.

Option A: Place the columns next to eachother.
Option B: Allow me to move them myself. I don't get why this isn't already a thing, though?

A keyboard shortcut could _possibly_ be used in place of this, but I find pretty much every web app sucks at making decent shortcuts, so, please don't even try until you fix this.

### Anything else?

Yes this is very ugly, but I'm not exactly interested in messing with the alignment using inspect element myself.

Image

jagadam97:
#### Description
Adds an isReorderable column flag separate from isModifiable so columns like Status, Actions, Details, and EventType can be repositioned via Table Options while remaining un-hideable. Applied to Queue, History, Blocklist, Missing, and Cutoff Unmet tables.

Also fixes a copy-paste bug where the Blocklist page's table option handler was calling setQueueOptions instead of setBlocklistOptions, which prevented any column reordering on Blocklist from persisting.

Screenshot_20260520_033532
Screenshot_20260520_033543
Screenshot_20260520_033609
Screenshot_20260520_033620
Screenshot_20260520_033646
Screenshot_20260520_033656
Screenshot_20260520_033709
Screenshot_20260520_033717

#### Database Migration
No - ###

* Closes #7897


mynameisbogdan:
How about we just change from `isModifiable` as a boolean to a multiple values, something like `disabled`, `enabled`, `onlyPosition`. As I find `isReorderable = isModifiable` very confusing.

WDYT @markus101?

markus101:
I hadn't considered that, but since the move to zustand stores is already breaking l think it makes a lot of sense.

jagadam97:
> How about we just change from isModifiable as a boolean to a multiple values, something like disabled, enabled, onlyPosition. As I find isReorderable = isModifiable very confusing.

Will make the change by weekend

jagadam97:
@mynameisbogdan 

have completed iomplementing the suggested changes. let me know if any improvements are needed.

markus101 (approved):
Thanks for picking this up!

frontend/src/Components/Table/Column.ts
```diff
@@ -2,6 +2,7 @@ import React from 'react';
 import { SortDirection } from 'Helpers/Props/sortDirections';
 
 type PropertyFunction<T> = () => T;
+export type IsModifiable = 'disabled' | 'enabled' | 'onlyPosition';
```
mynameisbogdan, line 5:
```suggestion
type PropertyFunction<T> = () => T;

export type IsModifiable = 'disabled' | 'enabled' | 'onlyPosition';
```
@jagadam97 Please add a newline before this type.


So the values are fine, @markus101?

markus101: Yeah, I think they make sense.

mynameisbogdan: Okay then, they were just quick suggestions I made and wanted to make sure you're okay with them too.

PR #8652, opened 2026-05-24
New: show more season titles

stevietv:
#### Description
shows season titles from skyhook when no sceneOrigin was found and thus no mappedSeasonNumber determined.

image

image

### DIFF
--- a/frontend/src/Utilities/Series/filterAlternateTitles.ts
+++ b/frontend/src/Utilities/Series/filterAlternateTitles.ts
@@ -63,6 +63,15 @@ function filterAlternateTitles(
           ? seasonNumber
           : sceneSeasonNumber;
 
+      // Add season titles with no mappedSeasonNumber
+      if (
+        mappedSeasonNumber === undefined &&
+        alternateTitle.seasonNumber === seasonNumber
+      ) {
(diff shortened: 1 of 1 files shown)


frontend/src/Utilities/Series/filterAlternateTitles.ts
```diff
@@ -63,6 +63,15 @@ function filterAlternateTitles(
...
           : sceneSeasonNumber;
 
+      // Add season titles with no mappedSeasonNumber
+      if (
+        mappedSeasonNumber === undefined &&
+        alternateTitle.seasonNumber === seasonNumber
+      ) {
+        seasonTitles.push(alternateTitle);
+        return;
+      }
```
mynameisbogdan:
```js
      const mappedSeasonNumber =
        alternateTitle.sceneOrigin !== 'tvdb' && sceneSeasonNumber !== undefined
          ? sceneSeasonNumber
          : seasonNumber;
```

WDYT about this instead? 🤔

stevietv: yeah, works and is cleaner :D thanks

PR #8654, opened 2026-05-25
Additional help text for queue removal with no show association

stevietv:
#### Description
When a download is not associated with a show its category cannot be changed and it cannot be ignored.
To avoid further support questions why, add additional help text when removal options are disabled due to no association with a show.

image

### DIFF
--- a/src/NzbDrone.Core/Localization/Core/en.json
+++ b/src/NzbDrone.Core/Localization/Core/en.json
@@ -1820,7 +1820,7 @@
   "RemoveQueueItem": "Remove - {sourceTitle}",
   "RemoveQueueItemConfirmation": "Are you sure you want to remove '{sourceTitle}' from the queue?",
   "RemoveQueueItemRemovalMethod": "Removal Method",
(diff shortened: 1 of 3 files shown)


src/NzbDrone.Core/Localization/Core/en.json
```diff
@@ -1820,6 +1820,7 @@
   "RemoveQueueItem": "Remove - {sourceTitle}",
   "RemoveQueueItemConfirmation": "Are you sure you want to remove '{sourceTitle}' from the queue?",
   "RemoveQueueItemRemovalMethod": "Removal Method",
+  "RemoveQueueItemRemovalMethodDisabledHelpText": "No other removal options can be selected as the download is not associated with a show in {appName}. Either select 'Remove from Download Client' or change its category manually.",
```
markus101: `show` -> `series` for consistency and tell them which download client to change the name in, I think we should be able to pass the download client's name in here, otherwise we could change to `in your download client.`

```suggestion
  "RemoveQueueItemRemovalMethodDisabledHelpText": "No other removal options can be selected as the download is not associated with a series in {appName}. Either select 'Remove from Download Client' or change its category manually in {downloadClientName}.",
```

stevietv: thanks for the suggestion, done.

PR #8655, opened 2026-05-25
V5 Redesign - UI Containers, Toolbar , Sidebar and Headings

bertPB:
#### Description
Redesign sidebar, page header, and toolbar to v5. Toolbar rebuilt on @fluentui/priority-overflow for automatic overflow collapsing. All toolbar callsites migrated.

localhost_8989_series

#### Database Migration
No

None


bertPB:
Going to see if I can find a better way to handle the moreMenuItems duplication, but should be RFR

frontend/src/Activity/Blocklist/Blocklist.tsx
```diff
@@ -146,6 +149,52 @@ function BlocklistContent() {
     [goToPage]
   );
 
+  const [isTableOptionsModalOpen, setIsTableOptionsModalOpen] = useState(false);
+
+  const onTableOptionsPress = useCallback(() => {
```
mynameisbogdan: I think we're started to use `handleAbc` over of `onAbc` in recent times.

mynameisbogdan:
On series index, if you click on `Select Series`, then `Stop Selecting` and then `Select Series` again, `Select All` won't work.

frontend/src/Components/Page/Header/PageHeader.tsx
```diff
@@ -42,12 +42,13 @@ function PageHeader() {
   return (
     <div className={styles.header}>
       <div className={styles.logoContainer}>
-        <Link className={styles.logoLink} to="/">
+        <Link className={styles.logoLink} to="/" aria-label={translate('Home')}>
           <img
             className={styles.logo}
             src={`${window.Sonarr.urlBase}/Content/Images/logo.svg`}
-            alt="Sonarr Logo"
+            alt=""
```
markus101, line 49: Why no alt?

frontend/src/Activity/Blocklist/Blocklist.tsx
```diff
@@ -160,37 +209,46 @@ function BlocklistContent() {
 
   return (
     <PageContent title={translate('Blocklist')}>
-      <PageToolbar>
-        <PageToolbarSection>
+      <PageToolbar moreMenuItems={moreMenuItems}>
+        <ToolbarItem id="remove-selected" priority={1} groupId="left-a">
```
markus101: Why `left-a` when there is only left/right?

frontend/src/Components/Page/Toolbar/PageToolbar.tsx
```diff
@@ -1,16 +1,108 @@
...
+import { PageToolbarButtonProps } from './PageToolbarButton';
+import PageToolbarOverflowMenuItem from './PageToolbarOverflowMenuItem';
+import PageToolbarSpacer from './PageToolbarSpacer';
 import styles from './PageToolbar.css';
 
+export type MoreMenuItem = PageToolbarButtonProps & {
+  id: string;
+  // Extras passed to `overflowComponent` when the item collapses into ⋮ — for callsite props the standard button doesn't carry.
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  overflowProps?: Record<string, any>;
```
markus101: Can we be more specific than `any`?

frontend/src/Components/Page/Toolbar/PageToolbar.tsx
```diff
@@ -1,16 +1,108 @@
...
 
 function PageToolbar({
   className = styles.toolbar,
   children,
+  moreMenuItems = [],
 }: PageToolbarProps) {
-  return <div className={className}>{children}</div>;
+  const childrenArray = Children.toArray(children).filter(
+    isValidElement
+  ) as ReactElement[];
```
markus101: `useMemo` so this isn't recomputed every render? (actually most of this should be memoized).

frontend/src/Components/Page/Toolbar/PageToolbar.tsx
```diff
@@ -1,16 +1,108 @@
...
+  const spacerIndex = childrenArray.findIndex(
+    (c) => c.type === PageToolbarSpacer
+  );
+
+  const moreInsertBefore =
+    spacerIndex >= 0 ? spacerIndex : childrenArray.length;
+
+  const rendered = childrenArray.flatMap((child, i) => {
+    if (i === moreInsertBefore) {
+      return [<MoreButton key="__more" items={moreMenuItems} />, child];
```
markus101: Why `__` in the key? Can't we just use `more`?

frontend/src/Components/Page/Toolbar/PageToolbar.tsx
```diff
@@ -1,16 +1,108 @@
...
+    }
+    return [child];
+  });
+  if (spacerIndex < 0) {
+    rendered.push(<MoreButton key="__more-end" items={moreMenuItems} />);
```
markus101: Same here.

frontend/src/Components/Page/Toolbar/PageToolbar.tsx
```diff
@@ -1,16 +1,108 @@
...
+      </Menu>
+    </span>
+  );
+}
+
+function OverflowMenuItemEntry({ item }: { item: MoreMenuItem }) {
+  const isVisible = useIsOverflowItemVisible(item.id);
+  if (isVisible) {
+    return null;
+  }
```
markus101, line 182: New lines before/after if block
```suggestion

  if (isVisible) {
    return null;
  }

```

frontend/src/Components/Page/Toolbar/PageToolbar.tsx
```diff
@@ -1,16 +1,108 @@
...
+  const { id: _id, overflowComponent, overflowProps, ...rest } = item;
+  const OverflowComponent = overflowComponent ?? PageToolbarOverflowMenuItem;
+  return <OverflowComponent {...rest} {...overflowProps} />;
```
markus101: Separate the return so it's not a giant block of text.

```suggestion

  return <OverflowComponent {...rest} {...overflowProps} />;
```

frontend/src/Activity/Blocklist/Blocklist.tsx
```diff
@@ -146,6 +149,52 @@ function BlocklistContent() {
...
+
+  const handleTableOptionsPress = useCallback(() => {
+    setIsTableOptionsModalOpen(true);
+  }, []);
+
+  const handleTableOptionsModalClose = useCallback(() => {
+    setIsTableOptionsModalOpen(false);
+  }, []);
+
+  const moreMenuItems = useMemo<MoreMenuItem[]>(
```
markus101: For every component we need to add a list of buttons that can be added to the overflow?

Since we're duplicating a bunch of code would it make sense to define all the buttons and whether they can be part of overflow or not instead?

frontend/src/Components/Page/Header/PageHeaderActionsMenu.css
```diff
@@ -1,20 +1,30 @@
...
-  margin-right: 15px;
-  width: 30px;
-  height: 60px;
-  text-align: center;
+  display: inline-flex;
+  align-items: center;
+  justify-content: center;
+  margin-left: 4px;
+  width: var(--chromeChipSize);
+  height: var(--chromeChipSize);
```
markus101: I know there was a discussion on "chrome" on Discord and could simplify this to just `chipSize`

frontend/src/Components/Page/Sidebar/PageSidebar.tsx
```diff
@@ -468,51 +450,73 @@ function PageSidebar() {
...
-                        isParentItem={false}
-                        isChildItem={true}
-                        statusComponent={child.statusComponent}
-                        onPress={handleItemPress}
-                      />
-                    );
-                  })}
-              </PageSidebarItem>
+              <Fragment key={`group-${groupIndex}`}>
+                {showSeparatorBefore && (
```
markus101: Since Use ternary instead of conditional rendering with `&&`

frontend/src/Components/Page/Sidebar/PageSidebar.tsx
```diff
@@ -468,51 +450,73 @@ function PageSidebar() {
...
+                  <div
+                    className={styles.divider}
+                    role="separator"
+                    aria-hidden="true"
+                  />
+                )}
+
+                {group.label && (
```
markus101: Same here

frontend/src/Components/Page/Sidebar/PageSidebar.tsx
```diff
@@ -468,51 +450,73 @@ function PageSidebar() {
...
+                          ? link.statusComponent
+                          : childStatusComponent
+                      }
+                      isActive={pathname === link.to && !hasActiveChild}
+                      isActiveParent={isActiveParent}
+                      isParentItem={!!link.children}
+                      onPress={handleItemPress}
+                    >
+                      {link.children &&
+                        link.to === activeParent &&
```
markus101: Existing, but since we're changing, this should be a ternary too.

frontend/src/Components/Page/Toolbar/Overflow.tsx
```diff
@@ -0,0 +1,157 @@
...
+          priority: priority ?? 0,
+          pinned,
+          groupId,
+        });
+      } else {
+        manager.removeItem(id);
+      }
+    },
+    [manager, id, priority, pinned, groupId]
+  );
```
markus101, line 123: Separate large `useCallback` from surrounding code for readability.
```suggestion

  const setNode = useCallback(
    (node: HTMLElement | null) => {
      if (node) {
        manager.addItem({
          element: node,
          id,
          priority: priority ?? 0,
          pinned,
          groupId,
        });
      } else {
        manager.removeItem(id);
      }
    },
    [manager, id, priority, pinned, groupId]
  );

```

frontend/src/Parse/useParseModal.tsx
```diff
@@ -0,0 +1,17 @@
...
+}
+
+export default function useParseModal(): ParseModalControls {
+  const [isOpen, setIsOpen] = useState(false);
+  const open = useCallback(() => setIsOpen(true), []);
+  const close = useCallback(() => setIsOpen(false), []);
+  return {
+    open,
+    modal: <ParseModal isOpen={isOpen} onModalClose={close} />,
+  };
```
markus101: Separate return

```suggestion

  return {
    open,
    modal: <ParseModal isOpen={isOpen} onModalClose={close} />,
  };
```

frontend/src/Series/Details/SeriesDetails.tsx
```diff
@@ -381,6 +384,102 @@ function SeriesDetails({ seriesId }: SeriesDetailsProps) {
...
 
+  let expandIcon = icons.EXPAND_INDETERMINATE;
+
+  if (expandedState.allExpanded) {
+    expandIcon = icons.COLLAPSE;
+  } else if (expandedState.allCollapsed) {
+    expandIcon = icons.EXPAND;
+  }
+
+  // useMemo must be before the early return to satisfy hooks rules
```
markus101: Unnecessary AI-ish style comment

frontend/src/Series/Index/SeriesIndex.tsx
```diff
@@ -58,14 +62,45 @@ function getViewComponent(view: string) {
...
+  );
+}
+
+interface SeriesIndexBodyProps {
+  seriesIndex: ReturnType<typeof useSeriesIndex>;
+}
+
+// Inner body — runs inside the SelectProvider so it can read selection state
+// directly. Renders toolbar buttons as plain PageToolbarButton elements so
+// the responsive-overflow algorithm can collapse them as needed.
```
markus101: Not sure this adds a ton of value

frontend/src/Settings/CustomFormats/CustomFormats/Manage/ManageCustomFormatsToolbarButton.tsx
```diff
@@ -5,21 +5,38 @@ import { icons } from 'Helpers/Props';
...
+  onPress?: () => void;
+  onModalClose?: () => void;
+}
+
+const noop = () => undefined;
+
+function ManageCustomFormatsToolbarButton({
+  isOpen,
+  onPress,
+  onModalClose,
```
markus101:
```suggestion
  onModalClose = noop,
```

frontend/src/Settings/CustomFormats/CustomFormats/Manage/ManageCustomFormatsToolbarButton.tsx
```diff
@@ -5,21 +5,38 @@ import { icons } from 'Helpers/Props';
...
+}: ManageCustomFormatsToolbarButtonProps) {
+  const [internalOpen, openInternal, closeInternal] = useModalOpenState(false);
+
+  const isControlled = isOpen !== undefined;
+  const effectiveOpen = isControlled ? isOpen : internalOpen;
+  const handlePress = isControlled ? onPress ?? noop : openInternal;
+  // In controlled mode, falling back to closeInternal would desync from the parent's isOpen.
+  const handleClose = isControlled ? onModalClose ?? noop : closeInternal;
```
markus101: Can we avoid `?? noop` by assigning a default?

frontend/src/Settings/CustomFormats/CustomFormats/Manage/ManageCustomFormatsToolbarButton.tsx
```diff
@@ -5,21 +5,38 @@ import { icons } from 'Helpers/Props';
...
+  isOpen,
+  onPress,
+  onModalClose,
+}: ManageCustomFormatsToolbarButtonProps) {
+  const [internalOpen, openInternal, closeInternal] = useModalOpenState(false);
+
+  const isControlled = isOpen !== undefined;
+  const effectiveOpen = isControlled ? isOpen : internalOpen;
+  const handlePress = isControlled ? onPress ?? noop : openInternal;
+  // In controlled mode, falling back to closeInternal would desync from the parent's isOpen.
```
markus101: When would this not be controlled?

markus101 (commented):
The overflow feels like a lot of duplication because we're defining buttons twice, once in the memo for the overflow and again in the JSX, I think cutting that down would be very beneficial.

frontend/src/Activity/Blocklist/Blocklist.tsx
bertPB: There's places which have / will have multiple left (or right) toolbar groups, but yeah doesn't make much sense to put "-a" in a place where only one group is present

frontend/src/Components/Page/Header/PageHeader.tsx
bertPB: The sonarr logo is not a functional UI element, and because alt gets read by screen-readers, it actually harms the experience more than it helps it.

frontend/src/Components/Page/Header/PageHeaderActionsMenu.css
bertPB: Good spot, will change!

frontend/src/Components/Page/Sidebar/PageSidebar.tsx
bertPB: Why? I feel like it's way messier (visually).
Any advantages to using a ternary operator, or just a personal pref?

frontend/src/Components/Page/Toolbar/PageToolbar.tsx
bertPB: Should probably be a rule in the formatter?
Will manually adjust for now

frontend/src/Components/Page/Toolbar/PageToolbar.tsx
bertPB: Children is recomputed every render anyway, so useMemo wouldn't do anything here.

frontend/src/Components/Page/Toolbar/PageToolbar.tsx
bertPB: Reworked

frontend/src/Series/Index/SeriesIndex.tsx
bertPB: Removed a bunch of pointless comments

bertPB:
@markus101 As for the duplicate declarations in the toolbar, I looked at making a small helper, but one thing led to another and now I just reworked it completely. It's quite a big chunk , but it's way more (imho) intuitive to use and no longer requires registering toolbar items twice.

Lmk what you think

frontend/src/Activity/Blocklist/Blocklist.tsx
bertPB: Changed toolbar setup completely in latest commit, feedback welcome

frontend/src/Components/Page/Header/PageHeader.tsx
markus101: Should the alt be excluded instead of an empty string?

frontend/src/Components/Page/Toolbar/PageToolbar.tsx
markus101: Yeah, last I looked it was a pain, but it's been at least a few years.

markus101: Opened a PR with a new rule. https://github.com/Sonarr/Sonarr/pull/8659

frontend/src/Components/Page/Header/PageHeader.tsx
bertPB: linters/formatters tend to complain of an image doesn't have an alt tag 😅
I'll see if that's the case here as well

frontend/src/Components/Page/Sidebar/PageSidebar.tsx
markus101: It's a great way to have things fall through accidentally rendering a value because something is falsey. We've been moving everything to ternaries as things are updated to avoid it.

https://kentcdodds.com/blog/use-ternaries-rather-than-and-and-in-jsx

frontend/src/Components/Page/Toolbar/Overflow.tsx
```diff
@@ -0,0 +1,161 @@
...
+  }
+  return ctx;
+}
+
+interface OverflowProps {
+  children: ReactElement;
+  padding?: number;
+}
+
+// `padding` must match container horizontal padding and stay stable per mount — changes re-observe and lose registered items.
```
markus101: Instead of a comment here can we add an effect that logs an error (in dev) to report that the padding changed after mount?

frontend/src/Components/Page/Toolbar/PageToolbar.tsx
```diff
@@ -1,16 +1,193 @@
...
+}
+
+const ToolbarRegistryContext =
+  createContext<ToolbarRegistryContextValue | null>(null);
+
+export function useToolbarRegistry() {
+  return useContext(ToolbarRegistryContext);
+}
+
 interface PageToolbarProps {
```
markus101: Existing, but can instead extend `PropsWithChildren`

frontend/src/Components/Page/Toolbar/ToolbarItem.tsx
```diff
@@ -0,0 +1,145 @@
...
+    onPress,
+    renderButton,
+    renderOverflow,
+    ...rest
+  } = props;
+  const registry = useToolbarRegistry();
+
+  useEffect(() => {
+    if (!registry) {
+      return;
```
markus101: Might be good to throw up an error in dev if used outside of the context.

frontend/.stylelintrc
```diff
@@ -308,7 +308,6 @@
     ],
     "selector-attribute-quotes": "never",
     "selector-class-pattern": "^[A-Za-z0-9]+$",
-    "selector-max-attribute": 0,
```
markus101: Instead of removing is there a maximum we should enforce?

frontend/src/Series/Index/SeriesIndex.tsx
```diff
@@ -185,163 +297,189 @@ function SeriesIndex() {
...
+              ? translate('StopSelecting')
+              : translate('SelectSeries')
+          }
+          iconName={isSelectMode ? icons.SERIES_ENDED : icons.CHECK}
+          renderButton={renderSelectButton}
+          renderOverflow={renderSelectOverflow}
+          onPress={onSelectModePress}
+        />
+
+        {isSelectMode && (
```
markus101: Ternary here too

frontend/src/Components/Table/TableOptions/TableOptionsModalWrapper.tsx
```diff
@@ -7,22 +7,40 @@ interface TableOptionsModalWrapperProps
   extends Omit<TableOptionsModalProps, 'isOpen' | 'onModalClose'> {
   columns: Column[];
   children: ReactElement<LinkProps>;
+  // Optional controlled mode — parent owns state so the overflow popup entry can trigger the same modal.
```
markus101: Do we still need the controlled logic?

frontend/src/Components/Page/Sidebar/PageSidebar.tsx
```diff
@@ -468,51 +450,73 @@ function PageSidebar() {
...
-                      />
-                    );
-                  })}
-              </PageSidebarItem>
+              <Fragment key={`group-${groupIndex}`}>
+                {showSeparatorBefore && (
+                  <div
+                    className={styles.divider}
+                    role="separator"
+                    aria-hidden="true"
```
markus101: We're simultaneously telling screen readers that we're in a different section, but also that we don't want them to announce it, which seems like a net negative, any particular reason this is being done?

frontend/src/Components/Page/Toolbar/PageToolbar.tsx
```diff
@@ -20,25 +24,94 @@ import PageToolbarOverflowMenuItem from './PageToolbarOverflowMenuItem';
...
+      }
+      if (a.id < b.id) {
+        return -1;
+      }
+      if (a.id > b.id) {
+        return 1;
+      }
+      return 0;
+    });
+  }, [items]);
```
markus101: Right now the order of buttons is changing, instead of comparing IDs we could maintain a list of IDs in order and use that to determine the order, but might be something better to do that.

frontend/src/Components/Page/Toolbar/PageToolbar.tsx
```diff
@@ -1,16 +1,193 @@
...
+  register: (item: MoreMenuItem) => void;
+  unregister: (id: string) => void;
+}
+
+function shallowEqualMenuItem(a: MoreMenuItem, b: MoreMenuItem): boolean {
+  const keys = Object.keys(a) as (keyof MoreMenuItem)[];
+  if (keys.length !== Object.keys(b).length) {
+    return false;
+  }
+  return keys.every((key) => a[key] === b[key]);
```
markus101: We should check both sides, something like this:

```suggestion
  for (const key in a) {
    if (a[key] !== b[key]) {
  	  return false;
  	}
  }

  for (const key in b) {
    if (!(key in a)) {
      return false;
    }
  }
  return true;
```

frontend/src/Components/Page/Toolbar/Overflow.tsx
```diff
@@ -0,0 +1,161 @@
...
+        manager.disconnect();
+      }
+    },
+    [manager, padding]
+  );
+
+  return (
+    <OverflowContext.Provider value={{ manager, visibility, overflowCount }}>
+      {/* `as never`: cloneElement ref injection — child must not already declare a ref. */}
+      {cloneElement(children, { ref: setContainer } as never)}
```
markus101: Do we need to cast to `never`?

markus101 (commented):
I like the changes to the toolbar, going to be a lot easier to add/update things.

frontend/.stylelintrc
bertPB: Feel free to pushback, but I feel a rule like this is a bit of a weird one. IF we were to restrict to a number, which would you pick, and why?
Why would X number of attribute selectors be good, instead of just not having the rule?

frontend/src/Components/Table/TableOptions/TableOptionsModalWrapper.tsx
bertPB: Nope, removing!

frontend/src/Components/Page/Sidebar/PageSidebar.tsx
bertPB: Good call, will remove the role

frontend/src/Components/Page/Toolbar/PageToolbar.tsx
bertPB: Both sides are being checked with the key length check.

frontend/src/Components/Page/Toolbar/PageToolbar.tsx
bertPB: There's not really a clean or "free" way to do this sadly. 
When you say "Right now the order of buttons is changing", you mean comparing the overflow menu compared to the order they're in when not overflowed, correct?

frontend/src/Components/Page/Toolbar/PageToolbar.tsx
markus101: Yes, as long as they always send the exact same objects, which they should as long as `ToolbarItem` is the only place we call `register` from different lengths will detect differences.

If there are multiple entries then this would pass, but be different:

```
a: {
  id: 'test',
  propA: undefined
}

b: {
  id: 'test',
  propB: 1
}
```

`id` matches, but so does `propA` because `propA` being undefined on `a` is the same as `b` not having a `propA` at all.

frontend/src/Components/Page/Header/PageHeader.tsx
bertPB: Ok, I looked into it a bit further, and having an empty alt tag is correct, an empty alt tag is the correct signal to screen readers that it's a decorative element, no alt tag is not explicit in that case.

frontend/.stylelintrc
markus101: I think it mostly comes does to readability and potentially efficiency (they're less efficient than classes, or at least were) if there is a better way to do it then chaining attributes or alternatively if it's better to allow some attributes vs all attributes.

I re-added this rule to my local checkout of this branch and it doesn't seem to fail on linting, did I overlook something?

I guess a follow up question is, why do we need to do selection based on attributes? And what are we using attribute selection for here?

frontend/src/Components/Page/Toolbar/PageToolbar.tsx
markus101: Yes, effectively desktop vs mobile, the overflow on the details screen is maybe the best example of how impacted it is. I think keeping them in order (same as they were before) is pretty important.

frontend/src/Components/Page/Toolbar/PageToolbar.tsx
bertPB: I see what you mean, but seems rather unlikely that we would ever have this issue right?
If you'd like to see this change, I don't mind changing it though, just felt the current shorter check was sufficient and cleaner.

frontend/.stylelintrc
bertPB: themes.css:184   [data-theme=dark]
themes.css:365   [data-color-impaired]
themes.css:391   [data-theme=dark][data-color-impaired]
PageToolbar.css:17  .toolbar [data-overflowing]

These are all the cases we use data attributes.
- First of all, this is standard practice in theming and is how it's done in most frameworks as well
- As for the data-overflowing, this is how it's handled internally in fluent-ui as well

frontend/src/Components/Page/Toolbar/PageToolbar.tsx
bertPB: Found a simple fix that's equally short/clean as my original, but should cover the same logic as yours

frontend/src/Components/Page/Toolbar/PageToolbar.tsx
bertPB: Ordering should now be consistent

frontend/src/Components/Page/Toolbar/PageToolbarButton.css
```diff
@@ -1,36 +1,54 @@
...
+    transform var(--transitionSmooth);
+  gap: 7px;
 
   &:hover {
-    color: var(--toolbarButtonHoverColor);
+    background-color: var(--paperCard);
+    box-shadow: var(--shadowChip);
+    color: var(--ink);
+    text-decoration: none;
+    transform: translateY(-1px);
```
markus101, line 23: I think this is the only thing that feels weird when using the app now, feels weird that it shifts up one pixel in addition to the other hover effects. As a comparison I tried scaling to 1.05 and that felt better, but a little too big, 1.025 might be looking a bit better.

Not sure if this fits into a larger effect that isn't in this PR, but I'd prefer not to have it, either removing it completely or changing to scale (or something else potentially).

markus101 (approved):
Just one thing on `transform: translateY(-1px);` for toolbar buttons, but otherwise looking good.

frontend/src/Components/Page/Toolbar/PageToolbarButton.css
bertPB: I'll experiment a bit with the hover animations.
If it's okay I'll do that in the next PR, gives me some space to adjust other animations as well, and this one can be merged then?

markus101: Yeah, just wanted to bring it up, looks like card does the same thing, but didn't notice before.

PR #8661, opened 2026-05-27
Add suffix handling for extra file imports and merge updated files on manual import

### LINKED ISSUE #8656: Improve "Import extra files" to import all files related to episodes

So the problem is that I am downloading anime shows from a torrent tracker where they use their own naming strategy and folder structure. The main thing is that Sonarr correctly finds the related files, but copies only one sound track (.mka). The torrent structure looks like this:

Anime name (anime root folder)
- sounds
- - x sound team
- - - EP 1 name.mka
- - y sound team
- - - EP 1 name.mka
- - z sound team
- - - EP 1 name.mka
- EP 1 name.mkv

But only one sound track gets copied.

I was planning to add my own PR to fix this issue, but I wanted to hear your opinions on this first.

There are also cases where the sound file is named with the sound team in the episode name, like this:

EP 1 name.[x sound team].mka

This naming is totally acceptable for Jellyfin, but Sonarr does not detect these files at all.

I'd like to see that all sound tracks are copied. If they have same names, their respected team name could be added in this format: EP 1 name.[x sound team].mka. or, just number them, like 1,2,3 etc. That would be OK for v1

Only copy them manually and rename by hand to match new names

### Anything else?

_No response_

embermann:
#### Description

1. Suffix handling for non-subtitle extras in OtherExtraService.cs — when multiple "other" extra files (e.g. .nfo, .jpg, .txt) are imported alongside one episode, they were all being written without disambiguation. The change introduces a copy counter and a new GetSuffix(copy, multipleCopies, title) helper that appends .1, .2, … to subsequent copies so they no longer overwrite each other. The title parameter is plumbed through for future use (currently always null at the call site).
2. Merge instead of replace in useInteractiveImport.ts — useReprocessInteractiveImportItems was replacing the whole row with the reprocessed payload . It now spreads ({ ...oldItem, ...reprocessedItem }), preserving any fields on the original item that the reprocess response doesn't return — so user-set values aren't blanked out on re-process

#### Submission Checklist

- [x] I have read [CONTRIBUTING.md](../CONTRIBUTING.md) and this PR follows the guidelines

- Closes #8656


src/NzbDrone.Core/Extras/Others/OtherExtraService.cs
```diff
@@ -155,5 +157,29 @@ public override IEnumerable<ExtraFile> ImportFiles(LocalEpisode localEpisode, Ep
 
             return importedFiles;
         }
+
+        private string GetSuffix(int copy, bool multipleCopies = false, string title = null)
+        {
+            var suffixBuilder = new StringBuilder();
+
+            if (title is not null)
```
mynameisbogdan:
```suggestion
            if (title.IsNotNullOrWhiteSpace())
```

Maybe check for empty strings too?

src/NzbDrone.Core/Extras/Others/OtherExtraService.cs
```diff
@@ -155,5 +157,29 @@ public override IEnumerable<ExtraFile> ImportFiles(LocalEpisode localEpisode, Ep
 
             return importedFiles;
         }
+
+        private string GetSuffix(int copy, bool multipleCopies = false, string title = null)
```
mynameisbogdan:
```suggestion
        private static string GetSuffix(int copy, bool multipleCopies = false, string title = null)
```

src/NzbDrone.Core/Extras/Others/OtherExtraService.cs
embermann: Maybe i should just remove this part and param "title" because it isn't used at all?

src/NzbDrone.Core/Extras/Others/OtherExtraService.cs
embermann: Is there a need to make it static? I am new in c#, maybe it gives any performance improvement or only to call it without creating instance of class?

mynameisbogdan: https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/static-classes-and-static-class-members#static-members
> Only one copy of a static member exists, regardless of how many instances of the class are created.

TLDR: As long you don't use any non-static fields in a static member, you're safe to make it static.

src/NzbDrone.Core/Extras/Others/OtherExtraService.cs
mynameisbogdan: You know better since you're the author. Was this created with the help of a LLM?

embermann: No, it was not, i made it myself. I am just not experienced enough, and want a bit of guidance.

src/NzbDrone.Core/Extras/Others/OtherExtraService.cs
embermann: Sounds reasonable, thanks.

src/NzbDrone.Core/Extras/Others/OtherExtraService.cs
```diff
@@ -138,11 +139,12 @@ public override IEnumerable<ExtraFile> ImportFiles(LocalEpisode localEpisode, Ep
                 }
             }
 
+            var copy = 1;
             foreach (var file in matchingFiles)
             {
                 try
                 {
-                    var extraFile = ImportFile(localEpisode.Series, episodeFile, file, isReadOnly, Path.GetExtension(file), null);
+                    var extraFile = ImportFile(localEpisode.Series, episodeFile, file, isReadOnly, Path.GetExtension(file), GetSuffix(copy++, matchingFiles.Count > 1));
```
mynameisbogdan: I just noticed this was copied from `SubtitleService`, and we don't really increment anywhere else in the call.

```suggestion
                    var suffix = GetSuffix(copy, matchingFiles.Count > 1);
                    var extraFile = ImportFile(localEpisode.Series, episodeFile, file, isReadOnly, Path.GetExtension(file), suffix);
```

src/NzbDrone.Core/Extras/Others/OtherExtraService.cs
```diff
@@ -138,11 +139,12 @@ public override IEnumerable<ExtraFile> ImportFiles(LocalEpisode localEpisode, Ep
...
                     _mediaFileAttributeService.SetFilePermissions(file);
                     _otherExtraFileService.Upsert(extraFile);
                     importedFiles.Add(extraFile);
```
mynameisbogdan, line 152:
```suggestion
                    importedFiles.Add(extraFile);

                    copy++;
```

mynameisbogdan (changes requested):

mynameisbogdan:
Would be nice to have some tests added in `OtherExtraServiceFixture` as well, in an attempt to avoid any breaking changes in the future.

embermann:
@mynameisbogdan tests added

frontend/src/InteractiveImport/useInteractiveImport.ts
```diff
@@ -151,7 +151,7 @@ export const useReprocessInteractiveImportItems = () => {
                 (updatedItem) => updatedItem.id === oldItem.id
               );
 
-              return reprocessedItem ? reprocessedItem : oldItem;
+              return reprocessedItem ? { ...oldItem, ...reprocessedItem } : oldItem;
```
mynameisbogdan: @markus101 What do you think about this one? Seems sane, or this should be left as is, and fix any issues on the backend side? 

My only concern with this would be masking issues when you use the UI, but not the API directly.

embermann: In my opinion, problem may be a bit deeper, because this line basically assigned ReprocessInteractiveImportItem instead of InteractiveImport to items array. But compiler haven't any objections against it. Maybe in project isn't enabled strict typing rules to catch these issues? Sorry if it is out of scope of PR

mynameisbogdan: That's okay, in API v3 the manualimport endpoint would return initially ManualImportResource (UI's InteractiveImport) and ManualImportReprocessResource (UI's ReprocessInteractiveImportItem) on reprocess, but API v5 returns ManualImportResource for reprocessed as well, but in most part they're equivalent.

embermann: Is there a point then to have so much names? Isn't it better to simplify and use same names for classes on server and UI?

markus101: Yeah, I was thinking the same, previously I had some issues with things being overwritten and fixed them up, but may have missed something. At the very least it'd be better to discuss this separately and not have it part of this PR and then decide where exactly we want to fix it.

embermann: Ok, i thought so too. I will take out this fix, and remain only related to extra file import!

embermann:
I guess PR is ready, waiting approval

src/NzbDrone.Core.Test/Extras/Others/OtherExtraServiceFixture.cs
```diff
@@ -89,5 +89,84 @@ public void should_not_import_multiple_nfo_files()
 
             results.Count.Should().Be(1);
         }
+
+        [Test]
+        [TestCase(@"audio_folder_1\Series Title S01E01.mka", @"audio_folder_2\Series Title S01E01.mka", "Series Title - S01E01.1.mka", "Series Title - S01E01.2.mka")]
+        public void should_import_all_files_with_same_name(string firstExtraFilePath, string secondExtraFilePath, string firstOutputPath, string secondOutputPath)
+        {
+            var files = new List<string> { Path.Combine(_episodeFolder, firstExtraFilePath).AsOsAgnostic(), Path.Combine(_episodeFolder, secondExtraFilePath).AsOsAgnostic() };
```
markus101:
```suggestion
            var files = new List<string>
            {
                Path.Combine(_episodeFolder, firstExtraFilePath).AsOsAgnostic(),
                Path.Combine(_episodeFolder, secondExtraFilePath).AsOsAgnostic()
            };
```

src/NzbDrone.Core/Extras/Others/OtherExtraService.cs
```diff
@@ -138,14 +139,18 @@ public override IEnumerable<ExtraFile> ImportFiles(LocalEpisode localEpisode, Ep
                 }
             }
 
+            var copy = 1;
```
markus101, line 142: Separate variable from the foreach block.
```suggestion
            var copy = 1;

```

markus101 (commented):
Looks pretty good, just minor things. Are you planning to add support for file suffix and subfolder as suffix too?

embermann:
You think in this PR? After some tests, even if audio track had this suffix, it was just for user, to distinguish between audio tracks in file system. Jellyfin reads proper data from file itself, so i guess we can safe our time on useless suffix saving. At least for me it is meningless if Jellyfin properly reads all it's data.

markus101:
Totally fine without it if it's not really used by Jellyfin.

PR #8668, opened 2026-05-29
New: Experimental multi-season release support (opt-in)

VerleneDodds:
#### Description

Adds support for parsing and importing multi-season release packs (e.g. `Show.S01-S08.BluRay`) behind an opt-in experimental toggle in Settings → Media Management → Experimental Features. The feature is disabled by default so existing behaviour is completely unchanged for users who do not enable it.

Changes include:
- Parser recognises `S01-S08`, `S01S02S03`, `Seasons 1-3`, and similar multi-season patterns
- Decision engine rejects multi-season releases when the toggle is off, with a message directing users to the setting
- Parsing service maps all seasons and episodes from a multi-season release
- Import service routes each file to its correct season folder
- Release browser displays season ranges (e.g. `S1–S3`) in the scene indicator column
- Queue progress badges show correct per-season episode counts while a multi-season pack is downloading

Experimental Features toggle in Settings → Media Management:
- Toggle visible at the bottom of the Media Management settings page under an "Experimental Features" section

Season range in release browser:
- Scene indicator column shows `S1–S3` for a consecutive range or `S1, S3` for non-consecutive seasons

#### Database Migration

No database migration required. The feature flag is stored via the existing key-value config system.

#### Submission Checklist

- [x] I have read [CONTRIBUTING.md](../CONTRIBUTING.md) and this PR follows the guidelines

- Closes #


src/NzbDrone.Core.Test/MediaFiles/DownloadedEpisodesImportServiceFixture.cs
```diff
@@ -500,30 +505,141 @@ public void should_return_rejection_if_nothing_imported_and_contains_executable_
         }
 
         [Test]
-        public void should_reject_if_download_is_multi_season()
+        public void should_reject_if_download_is_multi_season_with_no_content()
```
markus101, line 508: Wouldn't we reject any download without content?

src/NzbDrone.Core.Test/MediaFiles/DownloadedEpisodesImportServiceFixture.cs
```diff
@@ -500,30 +505,141 @@ public void should_return_rejection_if_nothing_imported_and_contains_executable_
...
         {
             GivenValidSeries();
 
             _trackedDownload.DownloadItem.Title = "Series Title S01-S11";
 
             var folderName = @"C:\media\ba09030e-1234-1234-1234-123456789abc\[HorribleSubs] Maria the Virgin Witch - 09 [720p]".AsOsAgnostic();
 
-            Mocker.GetMock<IDiskProvider>().Setup(c => c.FolderExists(folderName))
-                .Returns(true);
+            // Override global mocks so ProcessMultiSeasonFolder finds no subfolders and no root files → rejection
```
markus101, line 516: This sounds very 🤖 y

frontend/src/InteractiveSearch/ReleaseSceneIndicator.tsx
```diff
@@ -9,11 +9,31 @@ import { icons, tooltipPositions } from 'Helpers/Props';
 import translate from 'Utilities/String/translate';
 import styles from './ReleaseSceneIndicator.css';
 
+function formatSeasonRange(seasonNumbers: number[]): string {
+  const sorted = [...seasonNumbers].sort((a, b) => a - b);
```
markus101, line 13: Why can't these be sorted by the API?

frontend/src/InteractiveSearch/ReleaseSceneIndicator.tsx
```diff
@@ -9,11 +9,31 @@ import { icons, tooltipPositions } from 'Helpers/Props';
...
+
+  const isConsecutive = sorted.every(
```
markus101, line 15: In what cases would the seasons not be consecutive?

src/NzbDrone.Core/MediaFiles/DownloadedEpisodesImportService.cs
```diff
@@ -244,6 +241,95 @@ private List<ImportResult> ProcessFolder(DirectoryInfo directoryInfo, ImportMode
             return importResults;
         }
 
+        private List<ImportResult> ProcessMultiSeasonFolder(DirectoryInfo directoryInfo, ImportMode importMode, Series series, DownloadClientItem downloadClientItem, ParsedEpisodeInfo downloadClientItemInfo)
+        {
+            _logger.Debug("Processing multi-season folder: {0}", directoryInfo.FullName);
+
+            var allResults = new List<ImportResult>();
+
+            // Get season subfolders (immediate children only, filtered for extras)
```
markus101, line 250: Why do we _need_ season subfolders?

src/NzbDrone.Core/Parser/Model/ParsedEpisodeInfo.cs
```diff
@@ -22,6 +22,14 @@ public class ParsedEpisodeInfo
         public bool FullSeason { get; set; }
         public bool IsPartialSeason { get; set; }
         public bool IsMultiSeason { get; set; }
+
+        private int[] _seasonNumbers = Array.Empty<int>();
+        public int[] SeasonNumbers
+        {
+            get => _seasonNumbers ?? Array.Empty<int>();
+            set => _seasonNumbers = value;
+        }
```
markus101, line 31: Why private and public?

src/NzbDrone.Core/Parser/Model/ParsedEpisodeInfo.cs
```diff
@@ -122,6 +136,10 @@ public override string ToString()
             {
                 episodeString = string.Format("{0}", AirDate);
             }
+            else if (FullSeason && IsMultiSeason && SeasonNumbers.Length > 1)
```
markus101, line 139: How would it be multi-season and only have one season?

src/NzbDrone.Core/Parser/Model/RemoteEpisode.cs
```diff
@@ -15,6 +15,14 @@ public class RemoteEpisode
         public ParsedEpisodeInfo ParsedEpisodeInfo { get; set; }
         public SceneMapping SceneMapping { get; set; }
         public int MappedSeasonNumber { get; set; }
+
+        private int[] _mappedSeasonNumbers = Array.Empty<int>();
+        public int[] MappedSeasonNumbers
+        {
+            get => _mappedSeasonNumbers ?? Array.Empty<int>();
+            set => _mappedSeasonNumbers = value;
+        }
```
markus101, line 24: Again, why private and public?

src/NzbDrone.Core/MediaFiles/DownloadedEpisodesImportService.cs
```diff
@@ -244,6 +241,95 @@ private List<ImportResult> ProcessFolder(DirectoryInfo directoryInfo, ImportMode
...
+            if (!filteredSubfolders.Any() && !rootVideoFiles.Any())
+            {
+                _logger.Debug("No season subfolders or root video files found in multi-season folder: {0}", directoryInfo.FullName);
+                return new List<ImportResult>
+                {
+                    RejectionResult(ImportRejectionReason.MultiSeason, "Multi-season download could not be imported automatically. No season subfolders or video files were found in the expected locations. Manual import may be required.")
+                };
+            }
+
+            // Process each season subfolder
```
markus101, line 268: Seems like there are a lot of low quality 🤖 comments.

src/NzbDrone.Core/MediaFiles/DownloadedEpisodesImportService.cs
```diff
@@ -203,14 +203,11 @@ private List<ImportResult> ProcessFolder(DirectoryInfo directoryInfo, ImportMode
                 }
             }
 
-            if (downloadClientItemInfo is { IsMultiSeason: true })
+            if (downloadClientItemInfo is { IsMultiSeason: true } &&
+                _configService.EnableExperimentalMultiSeasonSupport)
```
gavincrawford, line 207: The old guard here rejects multi-season downloads unconditionally. With this new condition, this will only fire when the new flag is enabled. In this condition, even if the flag is off, multi-season downloads will fall through to `GetImportDecisions`. The renamed test `should_reject_if_download_is_multi_season_with_no_content` also no longer covers this path.

gavincrawford: To fix, when `IsMultiSeason` and the flag is off, keep returning the `MultiSeason` rejection

src/NzbDrone.Core/Parser/Parser.cs
```diff
@@ -1058,16 +1066,31 @@ private static ParsedEpisodeInfo ParseMatchCollection(MatchCollection matchColle
                     }
                 }
 
-                // If more than 1 season was parsed set IsMultiSeason to true so it can be rejected later
```
gavincrawford: What happens if there is a season bundle that, for example, contains seasons 1 & 3? This code would theoretically fabricate season two between these two points, which isn't tested in this PR yet.

src/Sonarr.Api.V5/Queue/QueueResource.cs
```diff
@@ -79,7 +85,15 @@ public static QueueResource ToResource(this NzbDrone.Core.Queue.Queue model, boo
...
                 OutputPath = model.OutputPath,
                 EpisodesWithFilesCount = model.Episodes?.Count(e => e.HasFile) ?? 0,
-                IsFullSeason = model.RemoteEpisode?.ParsedEpisodeInfo?.FullSeason ?? false
+                EpisodeCountBySeason = model.Episodes?
+                    .GroupBy(e => e.SeasonNumber)
+                    .ToDictionary(g => g.Key, g => g.Count()) ?? new(),
+                EpisodesWithFilesCountBySeason = model.Episodes?
+                    .Where(e => e.HasFile)
+                    .GroupBy(e => e.SeasonNumber)
+                    .ToDictionary(g => g.Key, g => g.Count()) ?? new(),
```
gavincrawford, line 94: This hunk dispatches two separate `model.Episodes?.GroupBy(...)`, which could possibly be consolidated to reduce query count. Not sure if this is really performance-intensive enough to be worth fixing, however.

src/NzbDrone.Core/MediaFiles/DownloadedEpisodesImportService.cs
```diff
@@ -244,6 +241,95 @@ private List<ImportResult> ProcessFolder(DirectoryInfo directoryInfo, ImportMode
...
+                }
+                catch (IOException e)
+                {
+                    _logger.Debug(e, "Unable to delete multi-season folder: {0}", e.Message);
+                }
+            }
+            else if (allResults.Empty())
+            {
+                allResults.AddIfNotNull(CheckEmptyResultForIssue(directoryInfo.FullName));
+            }
```
gavincrawford, line 328: Is this cleanup a duplicate of the implementation for `ProcessFolder`? Could we consolidate this duplicated logic? As with some other nits on this review, I'm not really even sure if this is worth the time to fix.

frontend/src/Settings/MediaManagement/MediaManagement.tsx
```diff
@@ -663,6 +663,27 @@ function MediaManagement() {
...
+                    'EnableExperimentalMultiSeasonSupportHelpText'
+                  )}
+                  helpTextWarning={translate(
+                    'EnableExperimentalMultiSeasonSupportHelpTextWarning'
+                  )}
+                  onChange={handleInputChange}
+                  {...settings.enableExperimentalMultiSeasonSupport}
+                />
+              </FormGroup>
+            </FieldSet>
```
markus101, line 686: Why introduce the concept of experimental features and put this under that umbrella?
What would need to be done to promote it to a fully supported feature?

src/Sonarr.Api.V5/Queue/QueueResource.cs
```diff
@@ -54,7 +57,10 @@ public static QueueResource ToResource(this NzbDrone.Core.Queue.Queue model, boo
                 Id = model.Id,
                 SeriesId = model.Series?.Id,
                 EpisodeIds = model.Episodes?.Select(e => e.Id).ToList() ?? [],
-                SeasonNumbers = model.SeasonNumber.HasValue ? [model.SeasonNumber.Value] : [],
+                SeasonNumbers = (model.RemoteEpisode?.ParsedEpisodeInfo?.IsMultiSeason == true &&
+                                 model.RemoteEpisode.ParsedEpisodeInfo.SeasonNumbers?.Length > 0)
+                    ? model.RemoteEpisode.ParsedEpisodeInfo.SeasonNumbers.ToList()
+                    : model.SeasonNumber.HasValue ? [model.SeasonNumber.Value] : [],
```
markus101, line 63: Double ternary is "one line", but a bear to follow, but we also shouldn't have this logic in the resource.

src/NzbDrone.Core/Parser/ParsingService.cs
```diff
@@ -708,5 +743,38 @@ private List<Episode> GetStandardEpisodes(Series series, ParsedEpisodeInfo parse
 
             return result;
         }
+
+        private List<Episode> GetMultiSeasonEpisodes(ParsedEpisodeInfo parsedEpisodeInfo, Series series, bool sceneSource)
```
markus101, line 747: Does this actually need to be handled separately or can we re-use existing methods or refactor to handle single and multi-season the same way?

src/NzbDrone.Core/Parser/Parser.cs
markus101: Probably not even a scenario we should support, same as we don't support a single episode file with episodes that aren't contiguous.

markus101 (changes requested):
> Discuss before building large features. For major features or architectural changes, please open an issue first. Once we've agreed on a solution the work can begin

From contributing.MD, this might not be a giant feature it is a pretty substantial feature as it impacts a lot of different parts of parsing in the app, a prior discussion would have been warranted.

There is already [a PR](https://github.com/Sonarr/Sonarr/pull/8508) that covers some of this, some of the comments there seem related to this PR and worth reviewing those.

Specifically trying to keep `SeasonNumber` and then adding in `SeasonNumbers` is messy.

src/NzbDrone.Core/Parser/Parser.cs
gavincrawford: I'd tend to agree. I'm not really sure if there are any torrent sources that would provide releases in this form-- non-contiguous sets of seasons-- and is there really a reasonable way to differentiate this from what the regex in this PR currently matches? (`S01-S03`, `S01 S03`, etc.)

gavincrawford:
Should we make an issue for this to discuss this feature? Definitely would be a nice feature to have, but does bring up some fairly serious edge cases.

markus101:
Not at this point, but would have saved some time and tokens discussing some of the decisions already made here before they were made.

src/NzbDrone.Core/MediaFiles/DownloadedEpisodesImportService.cs
pumalife: > The old guard here rejects multi-season downloads unconditionally. With this new condition, this will only fire when the new flag is enabled. In this condition, even if the flag is off, multi-season downloads will fall through to `GetImportDecisions`. The renamed test `should_reject_if_download_is_multi_season_with_no_content` also no longer covers this path.

ive heard this before, why do they reject it? it seems like such an obvious feature to implement.

PR #8674, opened 2026-05-30
udpate support-request labeller to v5

stevietv:
#### Description

udpate support-request labeller to v5 to allow for longer than 100 char github tokens
ref: https://github.com/dessant/support-requests/commit/247766c362b8b696ef69353f4b78faa2948cdf97

- [X] I have read [CONTRIBUTING.md](/Sonarr/Sonarr/blob/v5-develop/CONTRIBUTING.md) and this PR follows the guidelines
- [X] I have fully reviewed all code in this PR and confirm it works as intended
- [ ] This PR was authored and submitted by an AI agent without human review

PR #8675, opened 2026-05-30
Bump github actions

mynameisbogdan:
#### Description
Updating more outdated github actions, and adding permissions while at it.

- Closes #8674

### DIFF
--- a/.github/workflows/lock.yml
+++ b/.github/workflows/lock.yml
@@ -9,8 +9,12 @@ jobs:
   lock:
     runs-on: ubuntu-latest
     if: github.repository == 'Sonarr/Sonarr'
+    permissions:
+      issues: write
+      pull-requests: write
+      discussions: write
     steps:
-      - uses: dessant/lock-threads@v5
+      - uses: dessant/lock-threads@v6
         with:
           github-token: ${{ github.token }}
           issue-inactive-days: '90'
--- a/.github/workflows/support-requests.yml
+++ b/.github/workflows/support-requests.yml
@@ -4,15 +4,14 @@ on:
   issues:
     types: [labeled, unlabeled, reopened]
 
-permissions:
-  issues: write
-
 jobs:
   action:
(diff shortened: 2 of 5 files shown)


mynameisbogdan:
Should be squashed, but I added 484e9809f537e64980473dbabfb7397b317516f5 as a separate commit to show the intention.

.github/workflows/support-requests.yml
```diff
@@ -11,8 +11,10 @@ jobs:
   action:
     runs-on: ubuntu-latest
     if: github.repository == 'Sonarr/Sonarr'
+    permissions:
+      issues: write
```
stevietv, line 12: permissions key is outside of action key in lines 7-8, is it also needed here?

.github/workflows/lock.yml
```diff
@@ -9,8 +9,12 @@ jobs:
   lock:
     runs-on: ubuntu-latest
     if: github.repository == 'Sonarr/Sonarr'
+    permissions:
```
stevietv, line 12: in [docs](https://github.com/dessant/lock-threads#available-input-parameters), permssions are outside of the jobs key?

.github/workflows/support-requests.yml
mynameisbogdan: Didn't see it TBH. I'll remove it as per job is better.

.github/workflows/lock.yml
mynameisbogdan: > If a GitHub Actions job or workflow has no explicit permissions set, then the repository permissions are used. Repositories created under organizations inherit the organization permissions. The organizations or repositories created before February 2023 have the default permissions set to read-write. Often these permissions do not adhere to the principle of least privilege and can be reduced to read-only, leaving the write permission only to a specific types as issues: write or pull-requests: write.

In this case would be the same with a single job, but it's good practice to add them per jobs to adhere to least privilege principle.

PR #8680, opened 2026-05-31
V5 Redesign - Series Index page

bertPB:
#### Description
Redesigns the Series Index page to the v5 design language across all three views (the poster grid, the overview list, and
the table view).
Also reworks the progress bar, the bulk-edit selection modals, and the view option panels.

localhost_8989_ (7)
localhost_8989_ (6)
localhost_8989_ (8)

#### Submission Checklist

- Closes / (PR 4 of the redesign)


mynameisbogdan:
Smaller series titles?

bertPB:
> Smaller series titles?


In which view?

mynameisbogdan:
Posters.

```
  // Sizes
  extraSmallFontSize: '11px',
  smallFontSize: '12px',
  defaultFontSize: '14px',
  intermediateFontSize: '15px',
  largeFontSize: '16px',

  lineHeight: '1.528571429'
```

were better.

bertPB:
They're larger than the pre-v5 design.

frontend/src/Series/Index/Posters/SeriesIndexPoster.css
```diff
@@ -1,30 +1,41 @@
...
+  gap: 6px;
+
+  padding-bottom: 4px;
+  transition: transform 240ms cubic-bezier(0.16, 1, 0.3, 1);
 
   &:hover {
     z-index: 2;
-    box-shadow: 0 0 12px var(--black);
-    transition: all 200ms ease-in;
+    transform: translateY(-3px);
```
mynameisbogdan: Either lack of acceleration or something, but this effect is giving me sickness.

mynameisbogdan:
Screenshot 1

Screenshot 2

Maybe 12px might be too small, but we're losing some letters with 14px is what I'm trying to say.

frontend/src/Series/Index/ProgressBar/SeriesIndexProgressBar.css
```diff
@@ -1,13 +1,44 @@
...
+  box-shadow: none;
+  transition: width 240ms cubic-bezier(0.16, 1, 0.3, 1);
+}
+
+.progressDetailed {
+  composes: container from '~Components/ProgressBar.css';
+
+  overflow: hidden;
+
+  height: 14px !important;
```
markus101: Does all these actually need to be important?

frontend/src/Series/Index/ProgressBar/SeriesIndexProgressBar.css
```diff
@@ -1,13 +1,44 @@
...
+  background-color: var(--borderColor);
+  box-shadow: none;
   color: var(--white);
 }
 
 .progressBar {
   composes: progressBar from '~Components/ProgressBar.css';
 
-  transition: width 200ms ease;
+  height: 2px !important;
```
markus101: 2x feels super thin.

frontend/src/Series/Index/SeriesIndex.css
```diff
@@ -1,3 +1,14 @@
...
+  margin-bottom: 20px;
+  color: var(--dangerColor);
+  font-size: 13px;
+}
+
+.pageHeadingWrapper {
+  flex: 0 0 auto;
+  padding: 24px 32px 0;
+}
+
```
markus101: Doesn't look like either of these are used.

frontend/src/Series/Index/Posters/SeriesIndexPoster.css
markus101: Not making me sick, but like the toolbar shifting I don't like it. I don't think I've really seen this used much, but actually noticed it on a site that looks like it's from the 90s or early 00s 😬

frontend/src/Series/Index/SeriesIndex.css
```diff
@@ -31,13 +43,24 @@
...
 .contentBodyContainer {
   display: flex;
   flex-direction: column;
   flex-grow: 1;
 }
 
-@media only screen and (max-width: $breakpointSmall) {
+@media only screen and (max-width: 768px) {
```
markus101, line 52: This is the same value as `$breakpointSmall`, can we not use the variable?

frontend/src/Series/Index/Overview/SeriesIndexOverview.css
```diff
@@ -1,111 +1,197 @@
...
-.overview {
-  composes: link;
+.synopsis {
+  composes: link from '~Components/Link/Link.css';
 
-  overflow: hidden;
-  min-height: 0;
-}
+  /* stylelint-disable-next-line value-no-vendor-prefix -- required for -webkit-line-clamp */
+  display: -webkit-box;
```
markus101, line 183: Doesn't look like this actually uses `line-clamp`.

What does the support for non-webkit browsers look like?

frontend/src/Series/Index/Overview/SeriesIndexOverview.css
```diff
@@ -1,111 +1,197 @@
...
 
-.tags {
-  display: flex;
-  justify-content: space-around;
   overflow: hidden;
+  max-width: 110ch;
+  color: var(--inkSoft);
+  font-style: italic;
+  font-size: 13.5px;
+  font-family: $editorialFontFamily;
```
markus101: I'm not a fan of this font as it's used here, the contrast is pretty big compared to the other fonts. I also find it a lot harder to read.

frontend/src/Series/Index/Overview/SeriesIndexOverview.tsx
```diff
@@ -95,12 +86,16 @@ function SeriesIndexOverview(props: SeriesIndexOverviewProps) {
...
-  }, [rowHeight, isSmallScreen]);
-
-  const overviewHeight = contentHeight - TITLE_HEIGHT;
+    return rowHeight - (ROW_VERTICAL_PADDING * 2 + ROW_BORDER);
+  }, [rowHeight]);
+
+  const synopsisLines = Math.max(
+    1,
+    Math.floor(
+      (contentHeight - TITLE_HEIGHT - CHIP_STRIP_HEIGHT) /
```
markus101: Doesn't take into account that the chips can wrap.

frontend/src/Series/Index/ProgressBar/SeriesIndexProgressBar.tsx
```diff
@@ -36,6 +36,15 @@ function SeriesIndexProgressBar(props: SeriesIndexProgressBarProps) {
 
   const queueDetails = useQueueDetailsForSeries(seriesId, seasonNumber);
 
+  const containerClassName = useMemo(() => {
+    if (isStandalone) {
+      return undefined;
+    }
+
+    if (detailedProgressBar) return styles.progressDetailed;
+    return styles.progress;
```
markus101:
```suggestion
    if (detailedProgressBar) {
    	return styles.progressDetailed;
  	}

    return styles.progress;
```

frontend/src/Series/Index/Posters/SeriesIndexPoster.tsx
```diff
@@ -130,14 +130,46 @@ function SeriesIndexPoster(props: SeriesIndexPosterProps) {
     height: `${posterHeight}px`,
   };
 
+  const getStatusBadge = () => {
```
markus101: This function is rebuild ever render, extract it out of the render or memoize the result (but we have an early return above).

frontend/src/Series/Index/Overview/SeriesIndexOverview.tsx
```diff
@@ -137,6 +132,23 @@ function SeriesIndexOverview(props: SeriesIndexOverviewProps) {
...
+    continuing: translate('Continuing'),
+    upcoming: translate('Upcoming'),
+  };
+
+  const statusPillClasses: Partial<Record<SeriesStatus, string>> = {
+    ended: styles.ended,
+    deleted: styles.deleted,
+    continuing: styles.continuing,
+    upcoming: styles.upcoming,
+  };
```
markus101: Why are these being defined in the render loop?

frontend/src/Series/Index/Posters/SeriesIndexPoster.css
bertPB: It does have easing though, no ease-in that's true.
The easing used is appropriate for hover states, so not sure what to say here. Might be able to extend the animation duration by 100ms~ but that's crossing the 300ms border , which is not the greatest.

bertPB:
> Screenshot 1 Screenshot 2
> Maybe 12px might be too small, but we're losing some letters with 14px is what I'm trying to say.

Guess that's the tradeoff between readability and having more text visible, for me personally it was quite an easy decision to make.
If a user really really wants even more visible text, they can either pick a larger poster format or choose another view imho.

bertPB:
@markus101 Question, do you prefer it I mention in discord whenever a new redesign pr is RFR, or is the pr being open enough and you will see it anyway whenever you get to it?

markus101:
Just the PR is good, I have briefly looked at it, but haven't had a chance to do much of a review yet.

frontend/src/Series/Index/Posters/SeriesIndexPoster.css
mynameisbogdan: I would think a background shadow instead of moving it up and down would be better anyway.

frontend/src/Series/Index/Overview/SeriesIndexOverviews.tsx
```diff
@@ -95,14 +92,20 @@ function SeriesIndexOverviews(props: SeriesIndexOverviewsProps) {
...
-    ];
-
-    return heights.reduce((acc, height) => acc + height, 0);
-  }, [detailedProgressBar, posterHeight, isSmallScreen]);
+    const progress = detailedProgressBar
+      ? detailedProgressBarHeight
+      : progressBarHeight;
+
+    // posterHeight + gap + progress is the poster column; the row then adds
+    // .content's top+bottom padding and the 1px divider on top of that.
```
markus101: Comments should explain why, not what.

frontend/src/Styles/Themes/themes.css
```diff
@@ -163,6 +163,11 @@
   --tableRowHoverBackgroundColor: oklch(96% 0.008 80);
   --addSeriesBackgroundColor: oklch(96% 0.008 80);
   --seriesBackgroundColor: oklch(96% 0.008 80);
+
+  /* Sit on artwork, so theme-independent: no dark-block override */
+  --seriesOverlayScrim: oklch(15% 0.008 80 / 0.82);
```
markus101: What is a "scrim" in this context?

frontend/src/Styles/Themes/themes.css
```diff
@@ -163,6 +163,11 @@
...
+  --seriesOverlayInk: oklch(95% 0.006 80 / 0.9);
```
markus101, line 169: `seriesOverlayText`?

markus101 (changes requested):
For overview the chip row truncates pretty randomly.

Image

bertPB:
> For overview the chip row truncates pretty randomly.
> 
> Image

My bad there, I added a max-width so that the series description doesn't cover the full width, but applied it to the all of the content. Will adjust !

frontend/src/Series/Index/SeriesIndex.css
bertPB: The reason I am deprecating these specific variables, is that these are the only values that prevent me/us from fully removing post-css variables.

By the end of the redesign, only native css variables are used, while this is an improvement in general, the only downside is that css variables can not be used in media query decalrations.

Because media query breakpoints are unlikely to ever need to be redefined, I'd argue it's worth it.
If you disagree, I could re-add post-css variables just for this use-case.

frontend/src/Series/Index/ProgressBar/SeriesIndexProgressBar.css
bertPB: 2 of them were 100% redundant. I figured the container ones were needed because I wasn't sure how "composes:" works, but removing them there seems to have no negative effect, so all of them gone now.

frontend/src/Series/Index/ProgressBar/SeriesIndexProgressBar.css
bertPB: Bumped to 3px, lmk if you still feel it needs even more visibility.

frontend/src/Series/Index/SeriesIndex.css
bertPB: Removed them, not sure why these were still there either.

frontend/src/Series/Index/Overview/SeriesIndexOverview.css
bertPB: The line-clamp is done with in-line css because it needs to be dynamic. 
As for the -webkit-* prefixed items, these are supported in all browsers as well

frontend/src/Series/Index/Overview/SeriesIndexOverview.css
bertPB: Fair enough honestly, something I was questioning myself as well. Used it in alot more places initially but for longer texts it isn't great indeed.

Changed to just the basefont with italics, let me know if you like that better, or if even the italics is too much.

frontend/src/Series/Index/Overview/SeriesIndexOverview.tsx
bertPB: Fixed

frontend/src/Styles/Themes/themes.css
bertPB: we use Ink because variables like these aren't necessarily used just for text, keeps it consistent with the other paper/ink variable names

frontend/src/Styles/Themes/themes.css
bertPB: Changed to backdrop, makes way more sense.

bertPB:
@markus101 Left a few unresolved on purpose so you can check my reasoning first, resolve any you're good with.

All others are resolved, thanks for the thorough review

frontend/src/Series/Index/SeriesIndex.css
markus101: Getting rid of post-css makes sense to me, but it'd be nice to not to have a bunch of magic-ish numbers controlling the breakpoints. Can we use CSS variables for it instead?

bertPB: As mentioned above, sadly this is not possible :/
CSS Variables do not work in media-query declarations

markus101: Whoops, totally didn't process that as I read it. That's a bummer, but not the end of the world I suppose.

markus101 (approved):
Thanks!

PR #8703, opened 2026-06-09
Fixed: Exclude known titles from release title custom formats

### LINKED ISSUE #7456: Exclude Episode Titles from Custom Format matching

Right now, using the Release Title condition in Custom Formats can result in it matching on episode titles. For example, `Amazon` might be a common word in episode titles, e.g. the Futurama episode `Amazon Women in the Mood`. If we create a CF to match on the Amazon word for streaming service purposes, it can match on the episode title. Interestingly, it will not match on grab or on import, but it is matching on the renamed file. I suspect it might be related to the fact that Episode Title token is being included in the name

image
image
image

In the same way that the Episode Title is being excluded from the CF parsing on grab/import, it should ideally be excluded for the renamed file on disk. I believe this is also happening with the Language parsing since it no longer matches when episode titles feature common language words like `Italian`. I tested by creating a CF that only matches `Futurama` and it matched on grab, import, and on disk. So it seems Series Title is not being excluded at all. This issue exists in Radarr with Movie Title as well. 

For common words that might easily be included in Series and Episode titles, you need to create complicated regex to try and avoid matching on series and episode titles. It's doable for movie and series titles since they are usually followed by `SXXEXX` or year respectively, but episode titles are really painful to avoid, causing a lot of unintentional matching

### Anything else?

_No response_

TurboTheTurtle:
#### Description

Prevents Release Title custom format specifications from matching known series and episode titles when custom formats are recalculated from renamed files. This avoids false matches such as a streaming-service keyword appearing only in an episode title.

The release title and filename are checked after known series/episode title text is stripped, while unrelated release tokens remain available for matching.

N/A

#### Database Migration

No

#### Submission Checklist

- Closes #7456

#### Validation

- `dotnet test src/NzbDrone.Core.Test/Sonarr.Core.Test.csproj --filter "FullyQualifiedName~CustomFormatCalculationServiceFixture" --no-restore -p:RunAnalyzers=false`

### DIFF
--- a/src/NzbDrone.Core.Test/CustomFormats/CustomFormatCalculationServiceFixture.cs
+++ b/src/NzbDrone.Core.Test/CustomFormats/CustomFormatCalculationServiceFixture.cs
@@ -1,0 +1,120 @@
+using System.Collections.Generic;
+using FizzWare.NBuilder;
+using FluentAssertions;
+using Moq;
+using NUnit.Framework;
+using NzbDrone.Core.CustomFormats;
+using NzbDrone.Core.Languages;
+using NzbDrone.Core.MediaFiles;
+using NzbDrone.Core.Parser.Model;
+using NzbDrone.Core.Qualities;
+using NzbDrone.Core.Test.Framework;
+using NzbDrone.Core.Tv;
+
+namespace NzbDrone.Core.Test.CustomFormats
+{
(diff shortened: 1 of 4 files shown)


markus101:
Take 2 and still left the AI slop, no thanks.

Issue #4551, opened 2021-06-23
Storing multiple versions of the same file alongside each other


brandonscript:
## Storing multiple versions

The Sonarr team has begin investigating how different versions of a series could be stored alongside one another in the same directory. This ticket is to track the epic of work, including exploring UI/UX implementations, and required engineering changes. 

To manage noise, we are limiting who can comment/reply in this thread, but we will track work to be done and communicate progress.

### Goals

- Support multiple "tracks" of qualities that can be upgraded along side one another in the same directory. An example might be someone looking to store a higher quality version alongside a lower quality version to reduce the need to transcode.
- Create a user-friendly interface to make it easy to manage multiple copies, ensuring the feature is either opt-in, or easy for new or novice users to understand.
- Ensure backwards compatibility of existing setups, in that single-qualities libraries will not be impacted, with the hope that no data migration will be needed.
- Design systems that will hopefully translate to other companion platforms.

### Non-Goals

- This is to support multiple qualities _within the same library_ and in the same directory, and is not intended to manage media stored in separate directories. The recommended approach for this is still to manage multiple Sonarr instances.

### TODO 
_(work in progress)_

- UI: Handle partial completion of different qualities everywhere download or completion status is shown.
- UI: Provide easy-to-understand messaging and controls for setting up cutoffs, if changes are required.
- UI: Ensure search results that display info or recommended actions accommodate multiple qualities.
- UI: Provide informational messaging when qualities contradict or overlap each other.
- Core: Update database schema to support unique sets of files per episode instead of a single episode.
- Core: Update rules for processing when searching and manually adding.

### Challenges

- Handle files/series moved to different directories
- Handle manually added qualities to an existing title, where multiple qualities may not be set up correctly (e.g., do we auto-detect the new additional quality?)
- Handle file naming conflicts where quality/media is missing from the filename (where ambiguous qualities exist)
- Explore and consider the impact to tags, particularly where they are used to obtain custom qualities, like HDR. Tags are particularly challenging because they are ranked based on priority, not mutually exclusive.

#### Working proposal

_**Concept: using labels input to select multiple qualities on a per-series basis**_
![sonarr-1](https://user-images.githubusercontent.com/1480253/122628442-da268380-d06a-11eb-808c-66fcce408e47.png)

_**Concept: warning notice when qualities overlap in a potentially destructive, non-intuitive, or undesirable way**_
[...]

PR #8710, opened 2026-06-11
Add experimental series edition support

mici1708:
#### Description

Adds experimental support for series editions, which enables two or more Sonarr series to share the same TVDB id if each of them is a different edition.

This work paves a way for a new `SeriesEdition` field where the first set of possible values are `standard`, `directors_cut`, and `custom`. Old series will be `standard` by default and non-standard editions will be able to have their own path/root folder separately added. Besides that, the PR is edition-aware as far as duplicate checking, API/resource mapping, minimum add/edit UI support, searching aliases, and rejecting Director's Cut releases when the edition markers do not match, are concerned.

Alongside in the same initial MVP, R : ZERO Director's Cut is available as the first static version profile to prove the approach. It maps season 1 to the 13-episode Director's Cut order, yet the standard 25-episode series is kept as a different entry.

image

#### Database Migration

YES 231

#### Submission Checklist

### DIFF
--- a/frontend/src/AddSeries/AddNewSeries/AddNewSeriesModalContent.tsx
+++ b/frontend/src/AddSeries/AddNewSeries/AddNewSeriesModalContent.tsx
@@ -22,7 +22,7 @@ import ModalHeader from 'Components/Modal/ModalHeader';
 import Popover from 'Components/Tooltip/Popover';
 import { getValidationFailures } from 'Helpers/Hooks/useApiMutation';
 import { icons, inputTypes, kinds, tooltipPositions } from 'Helpers/Props';
-import { SeriesType } from 'Series/Series';
+import { SeriesEdition, SeriesType } from 'Series/Series';
 import SeriesPoster from 'Series/SeriesPoster';
 import selectSettings from 'Store/Selectors/selectSettings';
 import { useIsWindows } from 'System/Status/useSystemStatus';
@@ -61,6 +61,9 @@ function AddNewSeriesModalContent({
       ? settings.seriesType.value
       : initialSeriesType
   );
+  const [seriesEdition, setSeriesEdition] = useState<SeriesEdition>(
+    settings.seriesEdition.value
+  );
 
   const {
     monitor,
@@ -69,6 +72,7 @@ function AddNewSeriesModalContent({
     searchForCutoffUnmetEpisodes,
     searchForMissingEpisodes,
     seasonFolder,
+    seriesEdition: seriesEditionSetting,
     seriesType: seriesTypeSetting,
     tags,
   } = settings;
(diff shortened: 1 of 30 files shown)


mynameisbogdan:
This falls under #4551 with different quality profiles per edition if that's something desired. Adding 3 editions per series is more limiting and would conflict with multiple quality profiles per series and be eventually removed IMO.

frontend/src/AddSeries/AddNewSeries/AddNewSeriesSearchResult.tsx
```diff
@@ -54,9 +53,7 @@ function AddNewSeriesSearchResult({ series }: AddNewSeriesSearchResultProps) {
     event.stopPropagation();
   }, []);
 
-  const linkProps = isExistingSeries
-    ? { to: `/series/${titleSlug}` }
-    : { onPress: handlePress };
+  const linkProps = { onPress: handlePress };
```
mynameisbogdan: This change is unrelated and introduces a regression. Are you aware of this?

mynameisbogdan (changes requested):

frontend/src/AddSeries/AddNewSeries/AddNewSeriesSearchResult.tsx
mici1708: yes yes I knew it but by doing some tests it works perfectly, but if you ask me as a requirement that it be there I will do it

mynameisbogdan: What would be purpose of opening a add series modal for an existing series, don't you want to be able to click on it to visit the series details page post-add?

markus101:
I appreciate your interest in this feature, but you're approach here is off base for a many reasons.

1. You said you read Contributing.md, but this is a large feature, without any prior discussion or planning
2. You're hardcoding a ton of information, that is not something we're ever going to actually implement
3. Another existing PR is already using migration 231
4. This goes back to the first point, there is other major efforts already in progress that this will conflict with

As-is this is something we're never going to be able to accept, perhaps some time after alternate ordering is available it could be viable, but would need to be done in a way that could support it for multiple series without any of the hardcoding done here.

mici1708:
Thanks for the clarification.

That makes sense. I understand now that this is too large and too specific for a PR without prior design discussion, especially with the hardcoded episode/order data and the migration conflict.

I'll close this PR for now. If alternate ordering becomes available or there is a broader design direction for supporting multiple versions/editions, I may revisit this later with a more general approach that avoids hardcoded series-specific data and aligns with the existing work.

Thanks for taking the time to review it.

PR #8711, opened 2026-06-11
Fixed: Grab matching full season pack during anime episode search

hartmanfrost:
#### Description

For anime series the episode-search path in `SingleEpisodeSearchMatchSpecification` rejects every full-season pack with "Full season pack". As a result an anime series whose only available releases are standard-format season packs — e.g. dubbed `<Title> S01` releases, common on non-English trackers — can never be grabbed by the automatic missing-episode search or by an interactive episode search, even when the pack maps cleanly to the season and its files are correctly named (`Title.S01E01...`).

This accepts a full-season pack during an anime episode search **only when** it maps to the searched season and actually covers the requested episode:

```csharp
if (remoteEpisode.ParsedEpisodeInfo.FullSeason)
{
    if (animeEpisodeSpec.IsSeasonSearch)
    {
        return DownloadSpecDecision.Accept();
    }

    if (remoteEpisode.ParsedEpisodeInfo.SeasonNumber == animeEpisodeSpec.SeasonNumber &&
        remoteEpisode.Episodes.Any(e => e.SeasonNumber == animeEpisodeSpec.SeasonNumber &&
                                        e.EpisodeNumber == animeEpisodeSpec.EpisodeNumber))
    {
        return DownloadSpecDecision.Accept();
    }

    // skip
}
```

Why this is safe / narrow:

- It is **strictly narrower** than the existing season-search behaviour, which already accepts any full-season pack unconditionally (`IsSeasonSearch`). The new path adds the same acceptance for `IsSeasonSearch == false` but gated by a season **and** episode-coverage check.
- It does **not** touch absolute-numbered grabbing: those releases carry episode numbers (`FullSeason == false`) and never enter this branch. Zero change for absolute-only anime.
- Wrong-season and different-season packs are rejected by the season check; packs that don't contain the requested episode are rejected by the episode-coverage check (verified by unit tests).

This implements one concrete subcase of #6495 (Improve Searching For Anime Seasons) and addresses the "breaks known behaviour" concern that closed #7556: episode searching is not disabled and absolute matching is untouched.

Related to #6495

#### Did you read the contributing guidelines before submitting this pull request?

Yes

#### Tests

`AnimeSearchFixture` extended: full-season pack covering the searched episode → accepted; different season → rejected; pack missing the searched episode → rejected; existing season-search and non-full-season cases unchanged.

### DIFF
--- a/src/NzbDrone.Core.Test/DecisionEngineTests/Search/SingleEpisodeSearchMatchSpecificationTests/AnimeSearchFixture.cs
+++ b/src/NzbDrone.Core.Test/DecisionEngineTests/Search/SingleEpisodeSearchMatchSpecificationTests/AnimeSearchFixture.cs
@@ -1,9 +1,11 @@
-﻿using FluentAssertions;
+using System.Collections.Generic;
+using FluentAssertions;
 using NUnit.Framework;
 using NzbDrone.Core.DecisionEngine;
 using NzbDrone.Core.DecisionEngine.Specifications.Search;
 using NzbDrone.Core.IndexerSearch.Definitions;
 using NzbDrone.Core.Parser.Model;
+using NzbDrone.Core.Tv;
 using NzbDrone.Test.Common;
 
 namespace NzbDrone.Core.Test.DecisionEngineTests.Search.SingleEpisodeSearchMatchSpecificationTests
@@ -25,6 +27,7 @@ namespace NzbDrone.Core.Test.DecisionEngineTests.Search.SingleEpisodeSearchMatch
         [Test]
         public void should_return_false_if_full_season_result_for_single_episode_search()
         {
+            // No mapped episodes cover the requested episode, so the full-season pack is skipped
             _remoteEpisode.ParsedEpisodeInfo.FullSeason = true;
 
             Subject.IsSatisfiedBy(_remoteEpisode, _information).Accepted.Should().BeFalse();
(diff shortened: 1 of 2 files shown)


markus101:
If you search for a single episode regardless of series type season packs are rejected, this is expected behaviour, you're not searching for a season, you're searching for an episode.

In what search would you get a full season pack from a single episode search that wasn't found by the season search that was performed?

hartmanfrost:
@markus101, Thanks for pushing back on this — you're right to, and I'd rather answer (2) head-on than hand-wave.

### Short answer

Your framing in (1) is correct **in general**: a single-episode search isn't a season search, so rejecting season packs there is the right default. I'm not changing that default, and the patch doesn't touch standard (non-anime) series at all. The gap is specific to `SeriesType == Anime`, where there's a single-episode path that has **no accompanying season search to fall back on** — so the pack is the only release that can satisfy that episode, and today it's rejected.

### Answering "what search?" — the two anime paths diverge

The anime **season** search dispatches both criteria, so a pack is already caught:

```
SearchAnimeSeason(series, episodes)        // Anime season search
   ├─ AnimeSeasonSearchCriteria            (accepts full-season packs)
   └─ per episode: AnimeEpisodeSearchCriteria
```

The anime **episode** search dispatches only one:

```
EpisodeSearch(episode)  // SeriesType == Anime
   └─ SearchAnime(series, episode)
         └─ ONE AnimeEpisodeSearchCriteria  ← no AnimeSeasonSearchCriteria
```

So to your question directly — *"in what search would you get a full season pack from a single-episode search that wasn't found by the season search that was performed?"* — it's the search where **no season search is performed alongside it**. On the `SearchAnime` path nothing else is dispatched, so there's no season pass that could have grabbed the pack. This path is reached by:

- interactive **single-episode** search on an anime series, and
- automatic searches for an **individually-missing** episode (failed-import re-grab, a late-added or just-aired single episode, a cutoff-unmet single episode).

### Why this is the common case for anime, not an edge case

This is the part I undersold in the description. For **dubbed** anime (e.g. Russian/Ukrainian dubs on trackers like Anidub / Toloka), almost every title is only ever released as a **single full-season torrent** — e.g. `<Title> S01` — never per-episode. The single-season-pack format is the norm for the overwhelming majority of dubbed anime.

The absolute-numbered / ongoing format that anime mode originally assumed really only applies to the very longest, continuously-airing long-runners — Naruto, One Piece, and a handful like them. Those parse with `FullSeason == false`, so they never enter this branch and the patch can't affect them either way. The net effect today is backwards for this catalogue: the minority long-runner case works, and the common case (one season = one pack) gets dropped.

### What the patch actually accepts

It only touches the private `SingleEpisodeSearchMatchSpecification.IsSatisfiedBy(RemoteEpisode, AnimeEpisodeSearchCriteria)` overload, and it's strictly narrower than the existing unconditional `IsSeasonSearch` accept. A `FullSeason` release is accepted **only when**:

- `ParsedEpisodeInfo.SeasonNumber == animeEpisodeSpec.SeasonNumber`, **and**
- `remoteEpisode.Episodes` actually contains the searched `(SeasonNumber, EpisodeNumber)`.

Anything else still hits the same "Full season pack" reject as before. The season-search path and standard (non-anime) series behaviour are untouched.

### Scope / happy to narrow

This is the anime subcase of #6495 — deliberately not the broader #7556 that was deferred as too risky for v4 (maybe v5). It now targets **v5-develop**, rebased onto the v5 `ReleaseDecisionInformation` signature.

If the automatic individually-missing path feels too loose, I'm glad to gate this to interactive search only, or put it behind a flag, so the automatic path keeps today's behaviour and only user-initiated single-episode searches benefit. Let me know which direction you'd prefer and I'll adjust.

markus101:
Hello Claude,

> If the automatic individually-missing path feels too loose

This isn't too broad or loose, it's just wrong. If you search for a single episode a season pack should always be rejected because you're looking for a single episode, not a season pack.

If you want a season pack, search for the season, not an individual episode.

At best this sounds like "anime season search is slow", I want to "fix it" by working around it, which is not a solution. Right now, you can either change the series type to standard and search for a season or use interactive search and accept the season pack manually.

> I'm glad to gate this to interactive search only, or put it behind a flag, so the automatic path keeps today's behaviour and only user-initiated single-episode searches benefit. Let me know which direction you'd prefer and I'll adjust.

For interactive search it doesn't matter if it's rejected you can always accept it yourself so this is just extra code for next to no benefit. Putting it behind a flag isn't a solution we're going to implement.

Closing this because discussing this particular fix with AI output isn't productive.

PR #8733, opened 2026-06-20
Fixed: Kodi Metadata Files Issues

darksol736:
#### Description

Port of Radarr https://github.com/Radarr/Radarr/pull/11520 (more information and discussion there)
Original issue brought up in https://github.com/Radarr/Radarr/issues/11516

Fix a few issues in the `<streamdetails>` section of the generated Kodi .nfo files:
- `<codec>` is supposed to be the ffmpeg codec name or Kodi specific codes when profiles are involved
- `<framerate>` had an excessive number of decimals. Reduced to 3.

Differences with the Radarr PR:
- ffprobe 8.1.2 recognizes natively the hd formats with objects (Atmos and DTS:X) so standard profile names of ffmpeg are used instead of the custom workaround patches on top of ffprobe 5.1.4
- no fix needed to address audio language duplication since a previous change lists multiple audio streams correctly.

#### Database Migration

NO - ###

#### Submission Checklist

### DIFF
--- a/src/NzbDrone.Core.Test/Extras/Metadata/Consumers/Xbmc/XbmcMetadataFormatterTests/FormatAudioCodecFixture.cs
+++ b/src/NzbDrone.Core.Test/Extras/Metadata/Consumers/Xbmc/XbmcMetadataFormatterTests/FormatAudioCodecFixture.cs
@@ -1,0 +1,48 @@
+using FluentAssertions;
+using NUnit.Framework;
+using NzbDrone.Core.Extras.Metadata.Consumers.Xbmc;
+using NzbDrone.Core.MediaFiles.MediaInfo;
+using NzbDrone.Test.Common;
+
+namespace NzbDrone.Core.Test.Extras.Metadata.Consumers.Xbmc.XbmcMetadataFormatterTests
+{
+    [TestFixture]
+    public class FormatAudioCodecFixture : TestBase
+    {
+        [TestCase(new[] { "dts", "DTS-HD MA" }, "dtshd_ma")]
+        [TestCase(new[] { "dts", "DTS" }, "dts")]
+        [TestCase(new[] { "truehd", "Dolby TrueHD + Dolby Atmos" }, "truehd_atmos")]
+        [TestCase(new[] { "truehd", "" }, "truehd")]
+        [TestCase(new[] { "eac3", "Dolby Digital Plus + Dolby Atmos" }, "eac3_ddp_atmos")]
+        [TestCase(new[] { "eac3", "" }, "eac3")]
+        [TestCase(new[] { "aac", "" }, "aac")]
+        [TestCase(new[] { "aac", "HE-AAC" }, "he_aac")]
+        public void should_format_audio_format(string[] audioFormat, string expectedFormat)
+        {
+            var mediaInfo = new MediaInfoAudioStreamModel
+            {
+                Format = audioFormat[0],
+                Profile = audioFormat[1]
+            };
+
+            XbmcMetadataFormatter.FormatAudioCodec(mediaInfo).Should().Be(expectedFormat);
+        }
+
+        [Test]
+        public void should_return_audio_format_by_default()
+        {
+            var mediaInfo = new MediaInfoAudioStreamModel
+            {
+                Format = "Other Audio Format",
+            };
+
+            XbmcMetadataFormatter.FormatAudioCodec(mediaInfo).Should().Be(mediaInfo.Format);
+        }
+
+        [Test]
+        public void should_return_empty_if_audio_stream_is_null()
+        {
+            XbmcMetadataFormatter.FormatAudioCodec(null).Should().Be(string.Empty);
+        }
+    }
+}
--- a/src/NzbDrone.Core/Extras/Metadata/Consumers/Xbmc/XbmcMetadata.cs
+++ b/src/NzbDrone.Core/Extras/Metadata/Consumers/Xbmc/XbmcMetadata.cs
@@ -354,8 +354,8 @@ namespace NzbDrone.Core.Extras.Metadata.Consumers.Xbmc
                     var video = new XElement("video");
                     video.Add(new XElement("aspect", (float)episodeFile.MediaInfo.Width / (float)episodeFile.MediaInfo.Height));
                     video.Add(new XElement("bitrate", episodeFile.MediaInfo.VideoBitrate));
-                    video.Add(new XElement("codec", MediaInfoFormatter.FormatVideoCodec(episodeFile.MediaInfo, sceneName)));
-                    video.Add(new XElement("framerate", episodeFile.MediaInfo.VideoFps));
+                    video.Add(new XElement("codec", episodeFile.MediaInfo.VideoFormat));
+                    video.Add(new XElement("framerate", episodeFile.MediaInfo.VideoFps.ToString("0.###")));
                     video.Add(new XElement("height", episodeFile.MediaInfo.Height));
                     video.Add(new XElement("scantype", episodeFile.MediaInfo.ScanType));
                     video.Add(new XElement("width", episodeFile.MediaInfo.Width));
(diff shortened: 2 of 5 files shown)


src/NzbDrone.Core/Extras/Metadata/Consumers/Xbmc/XbmcMetadata.cs
```diff
@@ -529,5 +529,47 @@ private bool GetExistingWatchedStatus(Series series, string episodeFilePath)
 
             return Regex.IsMatch(fileContent, "<watched>true</watched>");
         }
+
+        public static string FormatAudioCodec(MediaInfoAudioStreamModel mediaInfo)
```
markus101: Since this is only used internally it should be private, but then you can't test it. Extracting out to a separate file and class is the best solution here, which can live alongside this file, maybe something like `KodiMetadataFormatter`?

src/NzbDrone.Core/Extras/Metadata/Consumers/Xbmc/XbmcMetadata.cs
```diff
@@ -529,5 +529,47 @@ private bool GetExistingWatchedStatus(Series series, string episodeFilePath)
...
+        {
+            var format = mediaInfo.Format;
+            var profile = mediaInfo.Profile ?? string.Empty;
```
mynameisbogdan: Since the split, it's not mediaInfo anymore. 

```suggestion
        public static string FormatAudioCodec(MediaInfoAudioStreamModel audioStream)
        {
            if (audioStream?.Format == null)
            {
                return string.Empty;
            }

            var format = audioStream.Format;
            var profile = audioStream.Profile ?? string.Empty;
```

src/NzbDrone.Core.Test/Extras/Metadata/Consumers/Xbmc/XbmcMetadataTests/FormatAudioCodecFixture.cs
```diff
@@ -0,0 +1,42 @@
+using FluentAssertions;
+using NUnit.Framework;
+using NzbDrone.Core.Extras.Metadata.Consumers.Xbmc;
+using NzbDrone.Core.MediaFiles.MediaInfo;
+using NzbDrone.Test.Common;
+
+namespace NzbDrone.Core.Test.Extras.Metadata.Consumers.Xbmc.XbmcMetadataTests
```
mynameisbogdan: The folder name doesn't match the namespace.

`XbmcMetadataTests` != `XbmcMedatataTests`.

src/NzbDrone.Core.Test/Extras/Metadata/Consumers/Xbmc/XbmcMedatataTests/FormatAudioCodecFixture.cs
```diff
@@ -0,0 +1,42 @@
...
+            {
+                Format = audioFormat[0],
+                Profile = audioFormat[1]
+            };
+
+            XbmcMetadata.FormatAudioCodec(mediaInfo).Should().Be(expectedFormat);
+        }
+
+        [Test]
+        public void should_return_AudioFormat_by_default()
```
mynameisbogdan:
```suggestion
        public void should_return_audio_format_by_default()
```

src/NzbDrone.Core.Test/Extras/Metadata/Consumers/Xbmc/XbmcMetadataFormatterTests/FormatAudioCodecFixture.cs
```diff
@@ -0,0 +1,42 @@
...
+        [Test]
+        public void should_return_AudioFormat_by_default()
+        {
+            var mediaInfo = new MediaInfoAudioStreamModel
+            {
+                Format = "Other Audio Format",
+            };
+
+            XbmcMetadata.FormatAudioCodec(mediaInfo).Should().Be(mediaInfo.Format);
+        }
```
mynameisbogdan, line 46:
```suggestion
        }

        [Test]
        public void should_return_null_if_audio_stream_is_null()
        {
            XbmcMetadata.FormatAudioCodec(null).Should().Be(string.Empty);
        }
```

mynameisbogdan (changes requested):

darksol736:
Applied the suggestions and requests.
Was it OK to submit for v5-develop instead of develop? There were no guidelines in the contributing doc and I don't know the workflow between branches.

mynameisbogdan:
It's okay, v5-develop is the default branch now.

darksol736:
By the way I believe there is a minor glitch in v5-develop and MediaInfoFormatter.FormatAudioCodec
`
"DTS:X" or "DTS-HD MA + DTS:X IMAX" => "DTS-X",
`
should be 
`
"DTS:X" or "DTS-HD MA + DTS:X" or "DTS-HD MA + DTS:X IMAX" => "DTS-X",
`

The 5.1.4 ffmpeg workaround "DTS:X" covered both and they each have a profile in 8.1.2

mynameisbogdan:
> By the way I believe there is a minor glitch in v5-develop and MediaInfoFormatter.FormatAudioCodec
> `"DTS:X" or "DTS-HD MA + DTS:X IMAX" => "DTS-X",`
> should be
> `"DTS:X" or "DTS-HD MA + DTS:X" or "DTS-HD MA + DTS:X IMAX" => "DTS-X",`

Very good catch! I added the IMAX in 227db9ef39fd151cf036ab2b11ed4e2eaf00c495 but didn't catch this one with ffmpeg 8.0.1. 

Do you mind adding a separate commit with the fix? 

Or a new PR if you want to, just keep in mind the repo is set to allow only 2 opened PRs per user at the moment to deal with LLMs going wild.

darksol736:
> > By the way I believe there is a minor glitch in v5-develop and MediaInfoFormatter.FormatAudioCodec
> Do you mind adding a separate commit with the fix?

Sure. Squashed the previous fixes and added commit for  "DTS-HD MA + DTS:X"

I leave up to you whether using AudioCodecId still makes sense - a vanilla 8.1.2 will not populate that field like the patched 5.1.4 used to. And whether you want to keep the two DTS:X profiles as one or distinguish them.

mynameisbogdan:
I'll have to check what you mean exactly, but I remember we left the old ones intentionally just in case of some weird cases.

darksol736:
Test directory changed to match the class name XbmcMetadataFormatter.

mynameisbogdan:
> I leave up to you whether using AudioCodecId still makes sense - a vanilla 8.1.2 will not populate that field like the patched 5.1.4 used to. And whether you want to keep the two DTS:X profiles as one or distinguish them.

I looked into this, and indeed, audioCodecId is a thing of the past ffmpeg 5.x. 

Let's leave them as is, I don't want to run into any more special cases, as I don't see any conflicts at the moment. 

Okay with you @markus101?

src/NzbDrone.Core/MediaFiles/MediaInfo/MediaInfoFormatter.cs
```diff
@@ -69,7 +69,7 @@ public static string FormatAudioCodec(MediaInfoAudioStreamModel audioStream, str
             {
                 return audioProfile switch
                 {
-                    "DTS:X" or "DTS-HD MA + DTS:X IMAX" => "DTS-X",
+                    "DTS:X" or "DTS-HD MA + DTS:X" or "DTS-HD MA + DTS:X IMAX" => "DTS-X",
```
mynameisbogdan, line 72: Do you mind renaming this commit to `Fixed: Parsing DTS-HD MA + DTS:X as DTS-X in media info` or similar that's less verbose?

darksol736: sure, done.

mynameisbogdan (approved):

darksol736:
Thanks! Will you accept a backport to v4 (with the appropriate ffmpeg-related changes of course)?

markus101:
Not at this time, we're only backporting high priority fixes and patching security issues in v4.

PR #8737, opened 2026-06-21
Fixed: Slow series monitoring changes on series with many episodes

### LINKED ISSUE #8738: Changing series monitoring is very slow / UI hangs on series with many episodes

### Current Behavior

Opening "Series Monitoring" (or selecting a series and using the bulk Season Pass) and changing the monitoring option on a series with a large episode count makes the Save button spin for several seconds before the dialog closes. It's most noticeable on long anime/daily series — for example Pokémon with ~1,300 episodes. The change does apply correctly; it's just slow, and the delay scales with the episode count (small series apply almost instantly).

### Expected Behavior

Changing a series' monitoring should complete quickly and consistently, regardless of how many episodes the series has.

### Steps To Reproduce

1. Add/open a series with a large number of episodes (e.g. Pokémon, ~1,300 episodes).
2. Click "Series Monitoring" (or select it in the series index and use Season Pass).
3. Pick any monitoring option and click Save.
4. The spinner hangs for several seconds before the dialog closes.

### Environment

```
- OS: Docker on Synology DS1821 (DSM)
- Sonarr: 4.0.17.2969
- Package: linuxserver.io 4.0.17.2969-ls174
- Docker Install: Yes
- Using Reverse Proxy: No
- Browser: Helium
- Database: SQLite 3.51.2
- .NET: 6.0.13
```

v5-develop

### Trace Logs?

[sonarr.trace.txt](https://github.com/user-attachments/files/29184454/sonarr.trace.txt)

### Anything else?

_No response_

primetime43:
#### Description

Changing a series' monitoring (the Series Monitoring dialog or the bulk Season Pass) was slow on series with a large episode count — the save would hang for several seconds on something like Pokémon (~1,300 episodes).

The cause was per-row work. `SetEpisodeMonitoredStatus` loaded every episode for the series, flipped the `Monitored` flag in memory, and wrote them all back. That write happens as one `UPDATE ... WHERE Id = @Id` per episode, so the cost scaled with the episode count and ran synchronously inside the Season Pass request.

This replaces it with a set-based update in `EpisodeRepository`: a scoped `UPDATE` applies the monitoring rule directly in SQL (special seasons handled separately). `SetMonitored` returns the season numbers that still have monitored episodes, so `EpisodeMonitoredService` recomputes the season monitored flags without a second query.

On the frontend, the Season Pass success handler now invalidates the cached episodes only for the series that changed, instead of every `/episode` query.

Existing `EpisodeMonitoredService` tests were updated to the new service API, and a `SetEpisodeMonitoredBySeriesFixture` was added to cover the repository behaviour against a real database for each monitor type.

#### Submission Checklist

- Closes #8738

### DIFF
--- a/frontend/src/Series/Details/EpisodeRow.tsx
+++ b/frontend/src/Series/Details/EpisodeRow.tsx
@@ -322,6 +322,4 @@ function EpisodeRow({
   );
 }
 
-// Memoized so that updating one episode's monitored state (which preserves the
-// object identity of unchanged episodes in the cache) doesn't re-render every row.
-export default React.memo(EpisodeRow);
+export default EpisodeRow;
--- a/frontend/src/Series/Details/EpisodeRow.tsx
+++ b/frontend/src/Series/Details/EpisodeRow.tsx
@@ -322,4 +322,6 @@ function EpisodeRow({
   );
 }
 
-export default EpisodeRow;
+// Memoized so that updating one episode's monitored state (which preserves the
+// object identity of unchanged episodes in the cache) doesn't re-render every row.
+export default React.memo(EpisodeRow);
--- a/frontend/src/Series/useSeries.ts
+++ b/frontend/src/Series/useSeries.ts
@@ -918,7 +918,13 @@ export const useUpdateSeriesMonitor = (
     mutationOptions: {
       onSuccess: (_, variables) => {
         if (shouldFetchEpisodesAfterUpdate) {
-          queryClient.invalidateQueries({ queryKey: ['/episode'] });
+          // Only refetch episodes for the series that actually changed, instead of
+          // invalidating every cached `/episode` query in the app.
+          variables.series.forEach((s) => {
+            queryClient.invalidateQueries({
+              queryKey: ['/episode', { seriesId: s.id }],
+            });
+          });
         }
 
         queryClient.setQueryData<Series[]>(['/series'], (oldSeries) => {
--- a/frontend/src/Series/useSeries.ts
+++ b/frontend/src/Series/useSeries.ts
@@ -918,8 +918,6 @@ export const useUpdateSeriesMonitor = (
     mutationOptions: {
       onSuccess: (_, variables) => {
         if (shouldFetchEpisodesAfterUpdate) {
-          // Only refetch episodes for the series that actually changed, instead of
-          // invalidating every cached `/episode` query in the app.
           variables.series.forEach((s) => {
             queryClient.invalidateQueries({
               queryKey: ['/episode', { seriesId: s.id }],
--- a/src/NzbDrone.Core.Test/TvTests/EpisodeMonitoredServiceTests/SetEpisodeMontitoredFixture.cs
+++ b/src/NzbDrone.Core.Test/TvTests/EpisodeMonitoredServiceTests/SetEpisodeMontitoredFixture.cs
@@ -110,7 +110,6 @@ namespace NzbDrone.Core.Test.TvTests.EpisodeMonitoredServiceTests
         [Test]
         public void should_not_monitor_first_season_when_monitoring_pilot()
         {
-            // The pilot episode in the first season is monitored, but the season itself must not be.
             GivenMonitoredSeasons(1);
 
             Subject.SetEpisodeMonitoredStatus(_series, new MonitoringOptions { Monitor = MonitorTypes.Pilot });
--- a/src/NzbDrone.Core.Test/TvTests/EpisodeMonitoredServiceTests/SetEpisodeMontitoredFixture.cs
+++ b/src/NzbDrone.Core.Test/TvTests/EpisodeMonitoredServiceTests/SetEpisodeMontitoredFixture.cs
@@ -26,14 +26,13 @@ namespace NzbDrone.Core.Test.TvTests.EpisodeMonitoredServiceTests
                                                             })
                                      .Build();
 
-            // By default no seasons have monitored episodes after the update; individual tests override.
             GivenMonitoredSeasons();
         }
 
         private void GivenMonitoredSeasons(params int[] seasonNumbers)
         {
             Mocker.GetMock<IEpisodeService>()
-                  .Setup(s => s.GetMonitoredSeasonNumbers(It.IsAny<int>()))
+                  .Setup(s => s.SetEpisodeMonitoredBySeries(It.IsAny<int>(), It.IsAny<MonitorTypes>(), It.IsAny<int>(), It.IsAny<int>()))
                   .Returns(seasonNumbers.ToList());
         }
 
--- a/src/NzbDrone.Core.Test/TvTests/EpisodeMonitoredServiceTests/SetEpisodeMontitoredFixture.cs
+++ b/src/NzbDrone.Core.Test/TvTests/EpisodeMonitoredServiceTests/SetEpisodeMontitoredFixture.cs
@@ -1,10 +1,8 @@
-using System;
 using System.Collections.Generic;
 using System.Linq;
 using FizzWare.NBuilder;
 using Moq;
 using NUnit.Framework;
-using NzbDrone.Common.Extensions;
 using NzbDrone.Core.Test.Framework;
 using NzbDrone.Core.Tv;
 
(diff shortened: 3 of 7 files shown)


frontend/src/Series/Details/EpisodeRow.tsx
```diff
@@ -322,4 +322,6 @@ function EpisodeRow({
   );
 }
 
-export default EpisodeRow;
+// Memoized so that updating one episode's monitored state (which preserves the
+// object identity of unchanged episodes in the cache) doesn't re-render every row.
+export default React.memo(EpisodeRow);
```
markus101: This isn't actually memoizing for other reasons, other future changes will fix it, so just revert for now.

frontend/src/Series/useSeries.ts
```diff
@@ -918,7 +918,13 @@ export const useUpdateSeriesMonitor = (
     mutationOptions: {
       onSuccess: (_, variables) => {
         if (shouldFetchEpisodesAfterUpdate) {
-          queryClient.invalidateQueries({ queryKey: ['/episode'] });
+          // Only refetch episodes for the series that actually changed, instead of
+          // invalidating every cached `/episode` query in the app.
```
markus101: 🤖 comment should be on the PR, not actually in the code

src/NzbDrone.Core.Test/TvTests/EpisodeMonitoredServiceTests/SetEpisodeMontitoredFixture.cs
```diff
@@ -14,415 +12,156 @@ namespace NzbDrone.Core.Test.TvTests.EpisodeMonitoredServiceTests
...
-                                        .With(e => e.EpisodeFileId = 0)
-                                        .With(e => e.AirDateUtc = null)
-                                        .Build()
-                                        .ToList();
-
-            Mocker.GetMock<IEpisodeService>()
-                  .Setup(s => s.GetEpisodeBySeries(It.IsAny<int>()))
-                  .Returns(_episodes);
+            // By default no seasons have monitored episodes after the update; individual tests override.
+            GivenMonitoredSeasons();
```
markus101: This doesn't match the seasons that are built for the series, episodes should typically follow the season monitoring by default.

src/NzbDrone.Core.Test/TvTests/EpisodeMonitoredServiceTests/SetEpisodeMontitoredFixture.cs
```diff
@@ -14,415 +12,156 @@ namespace NzbDrone.Core.Test.TvTests.EpisodeMonitoredServiceTests
...
-            VerifyNotMonitored(n => n.SeasonNumber == 1);
+            VerifySeasonNotMonitored(2);
         }
 
         [Test]
-        public void should_not_monitor_any_recent_episodes_if_all_episodes_aired_more_than_90_days_ago()
+        public void should_not_monitor_first_season_when_monitoring_pilot()
         {
-            _episodes.ForEach(e => e.AirDateUtc = DateTime.UtcNow.AddDays(-100));
+            // The pilot episode in the first season is monitored, but the season itself must not be.
```
markus101: This comment doesn't align with what the test is setting up.

src/NzbDrone.Core/Tv/EpisodeRepository.cs
```diff
@@ -192,6 +195,96 @@ public void SetMonitored(IEnumerable<int> ids, bool monitored)
...
+            using (var conn = _database.OpenConnection())
+            using (var tran = conn.BeginTransaction(IsolationLevel.ReadCommitted))
+            {
+                SetMonitoredWhere(conn, tran, $"\"SeriesId\" = @seriesId AND ({predicate})", true, parameters);
+                SetMonitoredWhere(conn, tran, $"\"SeriesId\" = @seriesId AND NOT ({predicate})", false, parameters);
+                tran.Commit();
+            }
+        }
+
+        public List<int> GetMonitoredSeasonNumbers(int seriesId)
```
markus101: I think something more verbose would be valuable, maybe`GetSeasonNumbersWithMonitoredEpisodes`, though not needed at all if `SetMonitored` returns the seasons that have monitored episodes.

src/NzbDrone.Core/Tv/EpisodeRepository.cs
```diff
@@ -192,6 +195,96 @@ public void SetMonitored(IEnumerable<int> ids, bool monitored)
...
+                    new { seriesId, monitored = true }).ToList();
+            }
+        }
+
+        private void SetMonitoredWhere(IDbConnection conn, IDbTransaction tran, string whereClause, bool monitored, DynamicParameters parameters)
+        {
+            var p = new DynamicParameters(parameters);
+            p.Add("monitored", monitored);
+
+            // The "Monitored" <> @monitored guard means only rows that actually change are written.
```
markus101: Again, a comment that could valuable in the PR, but not really in the code.

src/NzbDrone.Core/Tv/EpisodeRepository.cs
```diff
@@ -192,6 +195,96 @@ public void SetMonitored(IEnumerable<int> ids, bool monitored)
...
+                case MonitorTypes.Recent:
+                    predicate = "\"SeasonNumber\" > 0 AND (\"AirDateUtc\" IS NULL OR \"AirDateUtc\" >= @cutoff)";
+                    parameters.Add("cutoff", DateTime.UtcNow.AddDays(-90));
+                    break;
+                case MonitorTypes.None:
+                    // Matches nothing, so the "unmonitor the rest" pass clears the whole series.
+                    predicate = "1 = 0";
+                    break;
+                default:
+                    // Unknown monitor type: leave episodes untouched (matches previous no-op switch behavior).
```
markus101: Again another great 🤖 comment that should be on the PR, not in code.

src/NzbDrone.Core/Tv/EpisodeRepository.cs
```diff
@@ -192,6 +195,96 @@ public void SetMonitored(IEnumerable<int> ids, bool monitored)
...
+                    predicate = "\"SeasonNumber\" > 0 AND \"SeasonNumber\" = @lastSeason";
+                    parameters.Add("lastSeason", lastSeason);
+                    break;
+                case MonitorTypes.Recent:
+                    predicate = "\"SeasonNumber\" > 0 AND (\"AirDateUtc\" IS NULL OR \"AirDateUtc\" >= @cutoff)";
+                    parameters.Add("cutoff", DateTime.UtcNow.AddDays(-90));
+                    break;
+                case MonitorTypes.None:
+                    // Matches nothing, so the "unmonitor the rest" pass clears the whole series.
+                    predicate = "1 = 0";
```
markus101: This feels hacky, just to squeeze it into a predicate, also the inverse is never going to be true so we're making a query for zero action.

Can just do the same as specials with an early return.

src/NzbDrone.Core/Tv/EpisodeRepository.cs
```diff
@@ -192,6 +195,96 @@ public void SetMonitored(IEnumerable<int> ids, bool monitored)
...
+        public List<int> GetMonitoredSeasonNumbers(int seriesId)
+        {
+            using (var conn = _database.OpenConnection())
+            {
+                return conn.Query<int>("SELECT DISTINCT \"SeasonNumber\" FROM \"Episodes\" WHERE \"SeriesId\" = @seriesId AND \"Monitored\" = @monitored",
+                    new { seriesId, monitored = true }).ToList();
+            }
+        }
+
+        private void SetMonitoredWhere(IDbConnection conn, IDbTransaction tran, string whereClause, bool monitored, DynamicParameters parameters)
```
markus101: Don't throw the private function in the middle of the public ones.

src/NzbDrone.Core/Tv/EpisodeRepository.cs
```diff
@@ -192,6 +195,96 @@ public void SetMonitored(IEnumerable<int> ids, bool monitored)
             SetFields(episodes, p => p.Monitored);
         }
 
+        public void SetMonitored(int seriesId, MonitorTypes monitor, int firstSeason, int lastSeason)
+        {
+            var parameters = new DynamicParameters();
+            parameters.Add("seriesId", seriesId);
+
+            // Specials-only adjustments must not touch regular seasons, so they're a single scoped update.
```
markus101: 🤖 comment

src/NzbDrone.Core/Tv/EpisodeRepository.cs
```diff
@@ -192,6 +195,96 @@ public void SetMonitored(IEnumerable<int> ids, bool monitored)
...
+                using (var conn = _database.OpenConnection())
+                {
+                    SetMonitoredWhere(conn, null, "\"SeriesId\" = @seriesId AND \"SeasonNumber\" = 0", monitor == MonitorTypes.MonitorSpecials, parameters);
+                }
+
+                return;
+            }
+
+            // The predicate describes which episodes should be monitored. Everything else in the
+            // series is unmonitored, mirroring the previous in-memory ToggleEpisodesMonitoredState.
```
markus101: 🤖 comment, again PR comment != code comment

markus101 (changes requested):
It is faster especially on large series, but the PR is full of 🤖 comments and the template was completed poorly at best and either copied right from your LLM of choice without any human review or poorly manually copied and pasted.

src/NzbDrone.Core/Tv/EpisodeRepository.cs
```diff
@@ -192,6 +195,96 @@ public void SetMonitored(IEnumerable<int> ids, bool monitored)
...
 
+        public void SetMonitored(int seriesId, MonitorTypes monitor, int firstSeason, int lastSeason)
+        {
+            var parameters = new DynamicParameters();
+            parameters.Add("seriesId", seriesId);
+
+            // Specials-only adjustments must not touch regular seasons, so they're a single scoped update.
+            if (monitor == MonitorTypes.MonitorSpecials || monitor == MonitorTypes.UnmonitorSpecials)
+            {
+                using (var conn = _database.OpenConnection())
```
markus101: Can be converted to using declaration instead.

src/NzbDrone.Core/Tv/EpisodeRepository.cs
```diff
@@ -192,6 +195,96 @@ public void SetMonitored(IEnumerable<int> ids, bool monitored)
...
+                }
+
+                return;
+            }
+
+            // The predicate describes which episodes should be monitored. Everything else in the
+            // series is unmonitored, mirroring the previous in-memory ToggleEpisodesMonitoredState.
+            string predicate;
+
+            switch (monitor)
```
markus101: I think switching to if statements instead of an incomplete switch might be better (and not drastically longer)

primetime43:
So we don't want comments in code? Why is that? I always thought we'd want to have code comments to help other devs with explanations. (I removed the comments like you said)

markus101:
Unless it's complex code (then it's a question of reducing complexity to increase readability/understanding), comments are valuable when they explain _why_ something is being done, but describing _what_ the code is doing can be understood by reading the code. None of the comments really explained why the code what written that way, but they also didn't really need to explain why because the code was relatively simple.

Claude an other LLMs love to add comments to explain to the user what it's doing, but they're not valuable to maintaining the code most of the time.

mynameisbogdan:
Since we're skipping the whole model events logic, while I can see in useSeries hook, does this handle correctly the state for calendar, wanted pages among others I might've missed?

`_episodeService.SetMonitored(episodeIds, monitored)` would've another alternative, but still one update per query.

primetime43:
> Since we're skipping the whole model events logic, while I can see in useSeries hook, does this handle correctly the state for calendar, wanted pages among others I might've missed?
> 
> `_episodeService.SetMonitored(episodeIds, monitored)` would've another alternative, but still one update per query.

No this change doesn't touch that. The calendar does not update on a monitoring change. It seems it's already been broken on this branch. I've noticed that issue, I have updated series monitoring of items to none and it still shows on the calendar.

I guess the fix for that would be to do this?
image

markus101:
Your suggestion fixes it for Sonarr (on the calendar), but breaks any other client listening for episode events and not listening for series events in those same scenarios.

primetime43:
> Your suggestion fixes it for Sonarr (on the calendar), but breaks any other client listening for episode events and not listening for series events in those same scenarios.

What do you think of that to fix it?

markus101:
I was having some issues before, but after dropping your latest commit everything seems to be working. We also don't currently send episode model events for this bulk update, so we're not losing anything without that commit. If you want to drop that I can take another pass.

primetime43:
> I was having some issues before, but after dropping your latest commit everything seems to be working. We also don't currently send episode model events for this bulk update, so we're not losing anything without that commit. If you want to drop that I can take another pass.

What do you want me to do? Undo this latest commit? 8d926839a84eab9126d6c97b5b97b71bda08c7ee

mynameisbogdan:
https://github.com/Sonarr/Sonarr/commit/9c27855512f650fe1c5912cf63b7d885de302d45 is latest.

primetime43:
What's the next step? Anything need fixed or corrected? Or is it good now?

markus101 (approved):
Thanks for the fixes.

PR #8751, opened 2026-06-27
Fixed: Label actionable settings cards for screen readers

coopa11y:
#### Description

Adds accessible labels to settings cards that open another view or modal, so screen readers announce a clear action instead of only the visible card text.

The shared card component now passes the label through to the underlying action. Existing actionable settings cards have been updated with labels, including provider cards such as Indexers and Download Clients. Labels that include a custom name use translation variables so they can be localized in the right order.

This also keeps the add provider cards keyboard-accessible without adding a separate card wrapper.

Checked with eslint for the touched frontend files, and verified there are no remaining actionable card usages missing a label.

#### Submission Checklist


frontend/src/Components/Link/ActionCard.tsx
```diff
@@ -0,0 +1,54 @@
+import React, { ReactNode, SyntheticEvent } from 'react';
+import Link from './Link';
+
+interface ActionCardProps {
+  className: string;
+  underlayClassName: string;
+  overlayClassName: string;
+  nameClassName: string;
+  actionsClassName: string;
```
markus101: Are the implementation of these classes actually different between indexers and download clients? If not do we need the extra customization if they're just going to have the same CSS?

frontend/src/Settings/DownloadClients/DownloadClients/AddDownloadClientItem.tsx
```diff
@@ -32,47 +32,50 @@ function AddDownloadClientItem({
...
-      <div className={styles.overlay}>
-        <div className={styles.name}>{implementationName}</div>
-
-        <div className={styles.actions}>
-          {hasPresets ? (
-            <span>
-              <Button size={sizes.SMALL} onPress={handleDownloadClientSelect}>
-                {translate('Custom')}
-              </Button>
+    <ActionCard
```
markus101: Why only use it for add items and not the cards in general?

frontend/src/Components/Link/ActionCard.tsx
```diff
@@ -0,0 +1,54 @@
+import React, { ReactNode, SyntheticEvent } from 'react';
```
markus101: I don't think this should necessarily be in `Link`, I think Settings would be more appropriate.

coopa11y:
Thanks, that makes sense.

I reworked this to use the existing card component instead of adding a separate one just for this case. The card now accepts the accessible label and passes it through to the underlying action.

I also checked the existing card usages and added labels to the actionable settings cards, instead of only handling the Indexer and Download Client add cards. That covers the existing provider cards as well as the other settings cards using the same pattern.

Checked with eslint for the touched frontend files, and verified there are no remaining card usages with an action missing a label.

frontend/src/Settings/Metadata/Metadata/Metadata.tsx
```diff
@@ -44,6 +44,7 @@ function Metadata({ id, name, enable, fields }: MetadataProps) {
     <Card
       className={styles.metadata}
       overlayContent={true}
+      aria-label={`${translate('Metadata')}: ${name}`}
```
markus101: The name should be a variable on the translation because not all languages will read it the same way. Since it's not for display in the app, should it it `Metadata, {name}`?

frontend/src/Settings/DownloadClients/DownloadClients/DownloadClient.tsx
```diff
@@ -69,6 +69,9 @@ function DownloadClient({
     <Card
       className={styles.downloadClient}
       overlayContent={true}
+      aria-label={translate('EditDownloadClientImplementation', {
```
markus101: This is crossing between `name` and `implementationName`, which we might not want treated the same way (one being for display in the UI and the other for aria-label)

markus101 (changes requested):
Text that needs to be presented in a continuous way, either in the UI or for aria-labels needs to use variables, not concatenating them into a single string, otherwise they will not be translated properly in all languages.

coopa11y:
Good point.

I checked the existing keys again. The close matches are for provider names or different UI contexts, so I kept these labels separate instead of reusing those.

I updated the labels that include configured names to use their own translation keys. The new keys use the `Name` suffix to make that distinction clearer.

No existing localization keys were changed.

src/NzbDrone.Core/Localization/Core/en.json
```diff
@@ -613,19 +613,27 @@
   "Duration": "Duration",
   "Edit": "Edit",
   "EditAutoTag": "Edit Auto Tag",
+  "EditAutoTagName": "Edit Auto Tag - {name}",
```
markus101: Is the dash the best separator here?

frontend/src/Components/Link/ActionCard.tsx
coopa11y: Updated the new name-based accessible labels to use a comma instead of a dash, so they read more naturally as a continuous label.

I left the existing implementation-name labels unchanged since those were already there and seem to be a separate pattern.

markus101 (approved):
Thanks!

PR #8754, opened 2026-06-28
Fixed: Label table selection checkboxes

coopa11y:
#### Description

Labels the table row selection checkboxes so screen readers announce a useful name instead of leaving the checkbox unnamed.

The label uses a new localized row selection string, and the shared checkbox component now passes an optional accessible label through to the native checkbox input.

#### Submission Checklist

### DIFF
--- a/src/NzbDrone.Core/Localization/Core/en.json
+++ b/src/NzbDrone.Core/Localization/Core/en.json
@@ -1970,6 +1970,7 @@
   "SelectReleaseType": "Select Release Type",
   "SelectSeason": "Select Season",
   "SelectSeasonModalTitle": "{modalTitle} - Select Season",
+  "SelectRow": "Select row",
   "SelectSeries": "Select Series",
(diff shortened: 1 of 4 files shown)


src/NzbDrone.Core/Localization/Core/en.json
```diff
@@ -1970,6 +1970,7 @@
   "SelectReleaseType": "Select Release Type",
   "SelectSeason": "Select Season",
   "SelectSeasonModalTitle": "{modalTitle} - Select Season",
+  "SelectRow": "Select Row",
```
markus101: Since it's not a title, I think this is more correct:

```suggestion
  "SelectRow": "Select row",
```

markus101 (approved):
Just one suggestion.

src/NzbDrone.Core/Localization/Core/en.json
coopa11y: Updated to sentence case since this is an accessible label rather than a title.

PR #8760, opened 2026-07-05
Fixed: Label icon-only settings actions

coopa11y:
#### Description

Uses the shared `IconButton` component for the icon-only add and edit actions in delay profiles and remote path mappings.

These controls already have matching translated text, so the existing translation keys are used for both the accessible label and tooltip text.

#### Submission Checklist

### DIFF
--- a/frontend/src/Settings/Profiles/Delay/DelayProfile.tsx
+++ b/frontend/src/Settings/Profiles/Delay/DelayProfile.tsx
@@ -198,8 +198,8 @@ function DelayProfile({
         <div className={styles.actions}>
           <IconButton
             name={icons.EDIT}
-            aria-label={translate('EditDelayProfile')}
             className={id === 1 ? styles.editButton : undefined}
+            aria-label={translate('EditDelayProfile')}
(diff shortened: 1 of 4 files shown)


frontend/src/Settings/Profiles/Delay/DelayProfile.tsx
```diff
@@ -197,7 +197,9 @@ function DelayProfile({
 
         <div className={styles.actions}>
           <Link
+            aria-label={translate('EditDelayProfile')}
             className={id === 1 ? styles.editButton : undefined}
```
mynameisbogdan: For consistency sake. 

```suggestion
            className={id === 1 ? styles.editButton : undefined}
            aria-label={translate('EditDelayProfile')}
```

mynameisbogdan (changes requested):

frontend/src/Settings/Profiles/Delay/DelayProfile.tsx
coopa11y: Done, matched that ordering.

PR #8769, opened 2026-07-08
Keep search input visible with internal scrolling

lucasaugustodeveloper:
## Description

The **SearchResults** component was updated to use internal scrolling, ensuring that the search input remains visible at all times. This eliminates the need to scroll back to the top of the page when navigating through results in the middle or at the end of the list. Additionally, a class was added to the **PageContentBody** component to allow page scrolling to be disabled when internal scrolling is required.

#### Submission Checklist


frontend/src/AddSeries/AddNewSeries/AddNewSeriesSearchResult.css
```diff
@@ -5,6 +5,9 @@
   width: 100%;
   color: inherit;
 }
```
mynameisbogdan, line 7: Can't be nested to .searchResult?

```css
&:nth-child(odd) {
  margin: 0;
}
```

Also why only the odd ones? 😸

lucasaugustodeveloper: removal of duplicate margin

frontend/src/AddSeries/AddNewSeries/AddNewSeries.tsx
```diff
@@ -52,7 +52,7 @@ function AddNewSeries() {
 
   return (
     <PageContent title={translate('AddNewSeries')}>
-      <PageContentBody>
+      <PageContentBody className="noScroll">
```
markus101: Styles need to be defined in the `AddNewSeries.css`, not in `PageContentBody.css` (should never add a class via a magic string like this).

frontend/src/AddSeries/AddNewSeries/AddNewSeriesSearchResult.css
lucasaugustodeveloper: I removed that option; upon further testing, it doesn't make sense.

markus101: Instead of overriding it we can just cut the margin in half.

frontend/src/AddSeries/AddNewSeries/AddNewSeries.css
```diff
@@ -57,4 +57,10 @@
 
 .searchResults {
   margin-top: 30px;
+
+  height: 100dvh;
+  width: 100%;
+  overflow-y: scroll;
+  padding-bottom: 150px;
+
```
markus101, line 65: Random extra new lines here

markus101 (changes requested):

frontend/src/AddSeries/AddNewSeries/AddNewSeries.css
```diff
@@ -57,4 +57,10 @@
 
 .searchResults {
   margin-top: 30px;
+
+  height: 100dvh;
+  width: 100%;
+  overflow-y: scroll;
+  padding-bottom: 150px;
```
mynameisbogdan: This padding-bottom is a nasty hack that messes up the scrollbar by going below the window contents.

Screenshot 2026-07-08 at 18 43 33
Screenshot 2026-07-08 at 18 43 41

frontend/src/AddSeries/AddNewSeries/AddNewSeries.css
lucasaugustodeveloper: removed

frontend/src/AddSeries/AddNewSeries/AddNewSeries.css
lucasaugustodeveloper: removed

frontend/src/AddSeries/AddNewSeries/AddNewSeries.tsx
lucasaugustodeveloper: change for style in AddNewSeries.css

mise.toml
```diff
@@ -0,0 +1,2 @@
+[tools]
+node = "20"
```
mynameisbogdan: This should not be part of this PR.

frontend/src/AddSeries/AddNewSeries/AddNewSeries.css
```diff
@@ -57,4 +57,12 @@
 
 .searchResults {
   margin-top: 30px;
+
+  height: 80dvh;
+  width: 100%;
+  overflow-y: scroll;
+}
+
+.contentBody {
```
markus101, line 66: This should build on top of `contentBody` from `PageContentBody`, not just replace it:

```suggestion
.contentBody {
    composes: contentBody from '~Components/Page/PageContentBody.css';

```

lucasaugustodeveloper (commented):
Add css of component PageContentBody

frontend/src/AddSeries/AddNewSeries/AddNewSeries.css
```diff
@@ -57,4 +57,14 @@
 
 .searchResults {
   margin-top: 30px;
+
+  height: 80dvh;
```
mynameisbogdan, line 61: Screenshot 2026-07-10 at 00 18 19

This fixed value trims a few useful pixel from the bottom on a 5k monitor.

lucasaugustodeveloper (commented):
Add media query for 4K monitors and larger

mynameisbogdan:
There isn't really a way to use the full height of the search results without going 80-90%? Because going by percentage isn't going to look pretty for some.

lucasaugustodeveloper:
@mynameisbogdan 

Based on my knowledge and the work I've done so far, considering how the page height works, I don't think there's another way to keep the search field always visible.

Any other positioning approach I tested before opening the PR caused the search field to scroll up along with the page.

Another detail I took into consideration was the screen resolution of the users who will be using the application. The most common standard is Full HD. If I based this change on my own monitors, it would look even more unusual, since I use one 2K monitor, one 4K monitor, and two 2.5K portable monitors.

markus101:
A redesign that will be part of v5 is currently in progress, with that this should be easier to achieve without issues on different screen sizes and the author of that work has something in mind already. In order to not conflict with those efforts and not land something that is inconsistent on different sizes we're not going to land this at this time, but thank you for you contribution as it sparked another discussion about it.

PR #8775, opened 2026-07-10
New: Option to reject and blocklist torrents containing rejected extensions at grab time

nairdaleo:
#### Description

Public trackers increasingly serve fake releases containing only `.exe`/`.scr`/`.lnk` payloads. Since v4.0.11 the per-indexer **Fail Downloads** setting handles these at *import* time — but the poisoned download still has to be fully downloaded by the client first, sitting in the queue until completion before it can be failed, blocklisted and re-searched, this means the malware has a chance to get shared and multiplied. 

This PR moves the same check to **grab time**: the file list of the fetched `.torrent` is validated against the indexer's existing Fail Downloads setting (Executables / Potentially Dangerous / User Defined Extensions) before the release is sent to the download client. Offending releases are blocklisted and rejected immediately, so the decision engine falls through to the next-best release in the same pass and the client never sees the malicious torrent.

A new method `TorrentClientBase.EnsureTorrentContainsNoRejectedFiles` is added that finds extensions with `TorrentFileInfoReader.GetFileNamesFromTorrentFile()` when called from `DownloadFromWebUrl`, this throws a `ReleaseBlockedException` that gets caught by the generic catch and thus this release is skipped. The `foreach` continues and the loop moves to the next candidate. 

Yes, I've written this myself, the code was written by Claude under my supervision, including the unit tests. 

And yes, I also completely missed the mark on the contributing template, my apologies. 

#### Submission Checklist


src/NzbDrone.Core/Download/TorrentClientBase.cs
```diff
@@ -246,6 +251,83 @@ private string DownloadFromMagnetUrl(RemoteEpisode remoteEpisode, IIndexer index
             return actualHash;
         }
 
+        private void EnsureTorrentContainsNoRejectedFiles(RemoteEpisode remoteEpisode, IIndexer indexer, byte[] torrentFile)
```
markus101: 1. Name could be a little easier to read.
2. I have some concerns about customized extensions if someone adds a common extension, like `.nfo` and we just start tossing everything out, but not sure it's worth addressing.

```suggestion
        private void EnsureTorrentDoesNotContainRejectedFiles(RemoteEpisode remoteEpisode, IIndexer indexer, byte[] torrentFile)
```

src/NzbDrone.Core/Download/TorrentClientBase.cs
```diff
@@ -246,6 +251,83 @@ private string DownloadFromMagnetUrl(RemoteEpisode remoteEpisode, IIndexer index
...
+                return;
+            }
+
+            _logger.Debug("Torrent for '{0}' rejected: {1}. Blocklisting release.", remoteEpisode.Release.Title, rejection);
+            _blocklistService.Block(remoteEpisode, rejection, "TorrentFileValidation");
+
+            throw new ReleaseBlockedException(remoteEpisode.Release, rejection);
+        }
+
+        private string GetFileRejectionReason(List<string> fileNames, HashSet<FailDownloads> failDownloads)
```
markus101: What if there are multiple files that need to be rejected?
Feels a little clunky to say it's getting a rejection reason if it's not actually going to be rejected, I think something like `ValidateFileNames` would be more appropriate and just throw directly from it if necessary.

src/NzbDrone.Core/Download/TorrentClientBase.cs
```diff
@@ -246,6 +251,83 @@ private string DownloadFromMagnetUrl(RemoteEpisode remoteEpisode, IIndexer index
...
+                    var userRejectedExtensions = _configService.UserRejectedExtensions.Split(new[] { ',' }, StringSplitOptions.RemoveEmptyEntries)
+                        .Select(e => e.Trim(' ', '.')
+                            .Insert(0, "."))
+                        .ToList();
+
+                    if (userRejectedExtensions.Contains(extension, StringComparer.OrdinalIgnoreCase))
+                    {
+                        return $"Caution: Found file with user defined rejected extension: '{extension}'";
+                    }
+                }
```
markus101, line 310: I think there is value in pulling this all out to a shared function that can handle it for this + import time rejections.

 At the very least, the `userRejectedExtensions` bit needs to be shared.

markus101 (commented):
> Merge branch 'v5-develop' into grab-time-dangerous-file-rejection

You need to rebase your branch on top of `v5-develop` and drop this merge commit.

How often are you actually grabbing torrent files as opposed to magnet links?

nairdaleo:
> > Merge branch 'v5-develop' into grab-time-dangerous-file-rejection
> 
> You need to rebase your branch on top of `v5-develop` and drop this merge commit.
> 

alright, dropped. Had to cherry pick my changes as a single commit on top of `v5-develop`

> How often are you actually grabbing torrent files as opposed to magnet links?

Well, that depends on my sources. Some offer magnets, some offer torrent files. I... don't keep that close an eye on which one does what. 

But it really depends on what the clients offer right? Once we hand the grab to the download client we rely on whatever we can query from the client. I could only find qBittorrent and transmission file query options once some metadata exists (and transmission for example, doesn't expose any "metadata is ready!" option, this will have to be inferred somehow), this seems a bit out of scope for this PR, once the method to validate the file lists is available we can tackle each client as their own PRs. But you were right, this has more value as a shared function.

nairdaleo:
about this comment:

> I have some concerns about customized extensions if someone adds a common extension, like `.nfo` and we just start tossing everything out, but not sure it's worth addressing.

I agree, though I believe this is one of those UI things where it's possibly better to tell the user that rejecting from common extensions like `.nfo` and `.txt` will result in *very* selective acceptance. Technically the current release has the same problem doesn't it? Rejection box nor UI perform any sort of input validation, the only difference is we won't reject on grab, we'll reject after the fact. 

Still, probably worth flagging as a UI issue for that box

markus101:
> Technically the current release has the same problem doesn't it?

No, because they only apply if the torrent doesn't have a media file.

nairdaleo:
> > Technically the current release has the same problem doesn't it?
> 
> No, because they only apply if the torrent doesn't have a media file.

oh, you're right, had to go re-read that code again, failing on import just triggers re-search. Ok, still, probably more of a UI issue. 

Do you think that should be addressed in this PR or separately?

src/NzbDrone.Core/MediaFiles/FileExtensions.cs
```diff
@@ -48,6 +49,12 @@ public static class FileExtensions
             ".sh"
         };
 
+        public static IReadOnlyList<string> ParseUserRejectedExtensions(string raw) =>
+            raw?.Split(new[] { ',' }, StringSplitOptions.RemoveEmptyEntries)
+               .Select(e => e.Trim(' ', '.').Insert(0, "."))
+               .ToList()
+            ?? Array.Empty<string>();
+
```
markus101: This is too much for a simple arrow function, also I think the name can be more generic.

```suggestion
        public static List<string> ParseExtensions(string input)
        {
            if (input.IsNullOrWhiteSpace()
            {
                return Array.Empty<string>();
            }
            
            return input.Split(new[] { ',' }, StringSplitOptions.RemoveEmptyEntries)
               .Select(e => e.Trim(' ', '.').Insert(0, "."))
               .ToList();
        }

```

src/NzbDrone.Core/Download/TorrentClientBase.cs
```diff
@@ -246,6 +255,78 @@ private string DownloadFromMagnetUrl(RemoteEpisode remoteEpisode, IIndexer index
...
+            }
+
+            if (rejections.Count == 0)
+            {
+                return;
+            }
+
+            var rejection = string.Join("; ", rejections);
+
+            _logger.Debug("Torrent for '{0}' rejected: {1}. Blocklisting release.", remoteEpisode.Release.Title, rejection);
```
markus101: Should probably be a warning.

src/NzbDrone.Core/Download/TorrentClientBase.cs
```diff
@@ -246,6 +255,78 @@ private string DownloadFromMagnetUrl(RemoteEpisode remoteEpisode, IIndexer index
...
+
+                if (extension.IsNullOrWhiteSpace())
+                {
+                    continue;
+                }
+
+                if (failDownloads.Contains(FailDownloads.PotentiallyDangerous) &&
+                    FileExtensions.DangerousExtensions.Contains(extension))
+                {
+                    rejections.Add($"Caution: Found potentially dangerous file with extension: {extension}");
```
markus101: I think the caution on every line is unnecessary, instead that can be prepended when building the rejection message. Could also improve things by grouping by type and then compiling a single message, `found executables with extensions: exe, bat` or similar.

src/NzbDrone.Core/Download/TorrentClientBase.cs
```diff
@@ -246,6 +255,78 @@ private string DownloadFromMagnetUrl(RemoteEpisode remoteEpisode, IIndexer index
...
+                    rejections.Add($"Caution: Found file with user defined rejected extension: '{extension}'");
+                }
+            }
+
+            if (rejections.Count == 0)
+            {
+                return;
+            }
+
+            var rejection = string.Join("; ", rejections);
```
markus101: Probably better to do it one per line (or one type per line when combined with the suggestion above).

markus101 (commented):
I think it's fine to leave the rejection as-is, but will test it and see if that makes sense.

src/NzbDrone.Core/MediaFiles/FileExtensions.cs
mynameisbogdan: `[]` instead of `Array.Empty<string>()`.

src/NzbDrone.Core/Download/TorrentClientBase.cs
nairdaleo: addressed in 0569b6926 by collecting all reasons and using `ValidateFileNames`

src/NzbDrone.Core/MediaFiles/FileExtensions.cs
nairdaleo: Addressed in b94aee05c — renamed to `ParseExtensions`, expanded to a block body with early return, and using `[]` per @mynameisbogdan's note.

src/NzbDrone.Core/Download/TorrentClientBase.cs
nairdaleo: Addressed in 4f5151377 — extensions are now grouped by type and compiled into a single message per type (e.g. `Found executables with extensions: exe, bat`), with the `Caution:` prefix prepended once.

src/NzbDrone.Core/Download/TorrentClientBase.cs
nairdaleo: Addressed in 4f5151377 — one type per line now.

src/NzbDrone.Core/Download/TorrentClientBase.cs
nairdaleo: Addressed in 4f5151377 — now logged as a warning.

src/NzbDrone.Core/Download/TorrentClientBase.cs
nairdaleo: Renamed to `EnsureTorrentDoesNotContainRejectedFiles` in 95e25a4bb. Left the common-extension concern (e.g. someone adding `.nfo`) as-is per your note — it's an explicit user opt-in via UserDefinedExtensions, so it only rejects what the user configured.

src/NzbDrone.Core/Download/TorrentClientBase.cs
nairdaleo: The `userRejectedExtensions` parsing is now shared via `FileExtensions.ParseExtensions` (b94aee05c). I held off merging the full rejection paths: import-time rejects executables/user extensions unconditionally, while grab-time is gated by the `FailDownloads` flags, so a single shared method would change import behavior.

nairdaleo:
Great suggestions btw, I'm having fun and learning a lot

src/NzbDrone.Core.Test/Download/DownloadClientTests/QBittorrentTests/QBittorrentFixture.cs
```diff
@@ -454,6 +458,74 @@ public void api_261_should_use_content_path(string state)
             result.OutputPath.FullPath.Should().Be(torrent.ContentPath);
         }
 
+        [Test]
+        public void Download_should_blocklist_and_throw_when_torrent_contains_dangerous_file()
```
markus101: These shouldn't be implementation specific tests.

src/NzbDrone.Core.Test/Download/DownloadClientTests/Blackhole/TorrentBlackholeFixture.cs
```diff
@@ -321,5 +324,29 @@ public async Task should_return_null_hash()
 
             result.Should().BeNull();
         }
+
+        [Test]
+        public void Download_should_blocklist_and_throw_when_torrent_contains_dangerous_file_with_magnet_fallback()
```
markus101: Same here, there should be a test client created (within the test project) to test this behaviour instead of testing the various client implementations.

src/NzbDrone.Core/Download/TorrentClientBase.cs
```diff
@@ -246,6 +255,97 @@ private string DownloadFromMagnetUrl(RemoteEpisode remoteEpisode, IIndexer index
...
+
+        private void ValidateFileNames(RemoteEpisode remoteEpisode, List<string> fileNames, HashSet<FailDownloads> failDownloads)
+        {
+            var userRejectedExtensions = failDownloads.Contains(FailDownloads.UserDefinedExtensions)
+                ? FileExtensions.ParseExtensions(_configService.UserRejectedExtensions)
+                : [];
+
+            var dangerousExtensions = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+            var executableExtensions = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+            var userRejected = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
```
markus101, line 299: To match the other variables:
```suggestion
            var userRejectedExtensions = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
```

markus101 (commented):
Just a few things. I also brought up the rejecting at grab time aspect and whether a separate option makes sense or we just need to change the wording for the indexer-specific setting.

I think a separate setting similar to `RejectBlocklistedTorrentHashesWhileGrabbing` on Torrent indexers only would make it super clear, and we can further explain that it will reject torrent files with _any_ rejected extension, maybe help text would be something like: `Also apply 'Fail Downloads' setting when grabbing torrent files. This will reject torrents with any blocked extension regardless of a valid media file being present`.

src/NzbDrone.Core/Download/TorrentClientBase.cs
nairdaleo: easy peasy, done in 4740a7dcc — `userRejectedExtensions` is now a case-insensitive `HashSet<string>` matching the sibling collections, and the explicit `StringComparer` on the `Contains` check is dropped.

src/NzbDrone.Core.Test/Download/DownloadClientTests/QBittorrentTests/QBittorrentFixture.cs
nairdaleo: Done in 1b1eef9cb — pulled these out of the qBittorrent fixture onto a test-only `TestTorrentClient : TorrentClientBase` with a new `TorrentClientBaseFixture`, so the rejection behaviour is tested on the base class instead of a concrete client.

src/NzbDrone.Core.Test/Download/DownloadClientTests/Blackhole/TorrentBlackholeFixture.cs
nairdaleo: Done in 1b1eef9cb — same `TestTorrentClient/TorrentClientBaseFixture`. The magnet-fallback case this covered is preserved: `TestTorrentClient` exposes a toggleable `PreferTorrentFile`, so `should_blocklist_and_throw_when_torrent_contains_dangerous_file_with_magnet_fallback` still exercises the `PreferTorrentFile` path.

src/NzbDrone.Core/Download/TorrentClientBase.cs
nairdaleo: made a few more changes: added `FileExtensions.ParseUserRejectedExtensions` in cb684e223 — returns a case-insensitive `HashSet<string>` like the other ones. Both `TorrentClientBase.ValidateFileNames` and `DownloadedEpisodesImportService.ProcessFile` now call it, eliminating the duplicate parsing logic. The import-time path also drops the redundant `is not null` guard since the shared method always returns a valid set.

src/NzbDrone.Core/Download/TorrentClientBase.cs
nairdaleo: Found another one in the parse-failure case (`catch` when reading the torrent file list), promoted to `Warn` in cb684e223. A skipped security check shouldn't require debug logging to notice.

src/NzbDrone.Core.Test/Download/DownloadClientTests/QBittorrentTests/QBittorrentFixture.cs
nairdaleo: Follow-up: expanded the test fixture in 6bc03c39a and 9dcd19422:

- Renamed the mislabeled "magnet fallback" test as it tests `PreferTorrentFile=true`, not a magnet fallback
- Added a real magnet bypass test (`should_not_check_torrent_files_when_downloading_magnet_link`)
- Fixed the empty-FailDownloads gate test to actually exercise the gate (was hitting the flag-off return instead)
- Added case-insensitivity test for user-defined extensions
- Added mixed-type grouping test that exercises all three FailDownloads groups in one torrent, verifies the combined blocklist message

src/NzbDrone.Core.Test/Download/DownloadClientTests/Blackhole/TorrentBlackholeFixture.cs
nairdaleo: Expanded the test fixture in 6bc03c39a and 9dcd19422

nairdaleo:
Fixed the field numbering in a8b757c41 : the new `RejectTorrentFilesWithBlockedExtensionsWhileGrabbing` checkbox was rendering after `FailDownloads` in the UI, detached from its `RejectBlocklistedTorrentHashes` sibling. Renumbered all 8 torrent indexer settings so the two reject checkboxes sit together, with `MultiLanguages` and `FailDownloads` following. Torznab was already correct.

src/NzbDrone.Core/Localization/Core/en.json
```diff
@@ -1084,6 +1084,8 @@
   "IndexerSettingsPasskey": "Passkey",
   "IndexerSettingsRejectBlocklistedTorrentHashes": "Reject Blocklisted Torrent Hashes While Grabbing",
   "IndexerSettingsRejectBlocklistedTorrentHashesHelpText": "If a torrent is blocked by hash it may not properly be rejected during RSS/Search for some indexers, enabling this will allow it to be rejected after the torrent is grabbed, but before it is sent to the client.",
+  "IndexerSettingsRejectTorrentFilesWithBlockedExtensions": "Reject Torrent Files With Blocked Extensions",
```
markus101: Just a small nit with casing.
```suggestion
  "IndexerSettingsRejectTorrentFilesWithBlockedExtensions": "Reject Torrent Files with Blocked Extensions",
```

src/NzbDrone.Core/Localization/Core/en.json
```diff
@@ -1084,6 +1084,8 @@
...
+  "IndexerSettingsRejectTorrentFilesWithBlockedExtensionsHelpText": "Also apply 'Fail Downloads' setting when grabbing torrent files. This will reject torrents with any blocked extension regardless of a valid media file being present.",
```
markus101: Maybe `with` to `that contain`?

src/NzbDrone.Core.Test/Download/DownloadClientTests/Blackhole/TorrentBlackholeFixture.cs
```diff
@@ -10,12 +10,15 @@
 using NUnit.Framework;
 using NzbDrone.Common.Disk;
 using NzbDrone.Common.Http;
+using NzbDrone.Core.Blocklisting;
```
markus101: Looks like these extra name spaces aren't needed

src/NzbDrone.Core.Test/Download/DownloadClientTests/QBittorrentTests/QBittorrentFixture.cs
```diff
@@ -8,11 +8,15 @@
 using NUnit.Framework;
 using NzbDrone.Common.Disk;
 using NzbDrone.Common.Http;
+using NzbDrone.Core.Blocklisting;
```
markus101: Same here

src/NzbDrone.Core.Test/Download/DownloadClientTests/TorrentClientBaseTests/TorrentClientBaseFixture.cs
```diff
@@ -0,0 +1,203 @@
...
+                "payload.lnk",
+                "info.nfo");
+
+            Assert.ThrowsAsync<ReleaseBlockedException>(async () => await Subject.Download(remoteEpisode, indexer));
+
+            Mocker.GetMock<IBlocklistService>()
+                  .Verify(s => s.Block(remoteEpisode, It.Is<string>(msg =>
+                      msg.Contains("potentially dangerous") &&
+                      msg.Contains("executables") &&
+                      msg.Contains("user defined")), It.IsAny<string>()), Times.Once());
```
markus101: This gets caught by style-cop rule SA1117

markus101 (commented):
Looks like this is working well, just a little clean up and should be good to land.

src/NzbDrone.Core/Localization/Core/en.json
adrian-leon-fvt: Fixed in c2ecb5c4b

src/NzbDrone.Core.Test/Download/DownloadClientTests/Blackhole/TorrentBlackholeFixture.cs
adrian-leon-fvt: Fixed in c2ecb5c4b

src/NzbDrone.Core/Localization/Core/en.json
nairdaleo: Fixed in c2ecb5c4b

src/NzbDrone.Core.Test/Download/DownloadClientTests/Blackhole/TorrentBlackholeFixture.cs
nairdaleo: Fixed in c2ecb5c4b

src/NzbDrone.Core/Localization/Core/en.json
nairdaleo: Fixed in c2ecb5c4b

src/NzbDrone.Core.Test/Download/DownloadClientTests/QBittorrentTests/QBittorrentFixture.cs
nairdaleo: Fixed in c2ecb5c4b

src/NzbDrone.Core.Test/Download/DownloadClientTests/TorrentClientBaseTests/TorrentClientBaseFixture.cs
nairdaleo: Fixed in c2ecb5c4b

markus101:
Thanks for implementing this!

PR #8779, opened 2026-07-11
chore: keep operator instruction files out of the tracked tree

Starosdev:
Operator instruction files are maintained locally per machine; this removes them from the tracked tree. No functional change.


Starosdev:
Opened against the wrong repository - this belongs on a fork. Apologies for the noise.

PR #8795, opened 2026-07-18
V5 Redesign - Better drag & drop

bertPB:
#### Description

Replaces the old react-dnd with the way more modern dnd-kit library, making all drag and drop actions way smoother.

Recording 2026-07-18 200852

#### Submission Checklist


frontend/src/Settings/Profiles/Quality/useQualityProfileDnd.ts
```diff
@@ -0,0 +1,187 @@
...
+
+export const ROOT_CONTAINER = 'root';
+
+export function qualityKey(id: number) {
+  return `q-${id}`;
+}
+
+export function groupContainerKey(id: number) {
+  return `g-${id}`;
+}
```
mynameisbogdan:
```suggestion
export const qualityKey = (id: number) => `q-${id}`;
export const groupContainerKey = (id: number) => `g-${id}`;
```

markus101: If it's going to be a single line I'd probably drop the empty line on L13 too.

frontend/src/Settings/Profiles/Quality/QualityProfileItemGroup.tsx
```diff
@@ -9,56 +10,52 @@ import IconButton from 'Components/Link/IconButton';
...
   onSizeChange,
 }: QualityProfileItemGroupProps) {
-  const groupBaseIndex = parseInt(qualityIndex) - 1;
+  const { ref, handleRef, isDragging } = useSortable({
+    id: groupContainerKey(groupId),
+    index,
+    group: ROOT_CONTAINER,
+    type: 'group',
+    accept: ['quality', 'group'],
+    collisionPriority: CollisionPriority.Low,
```
markus101: This fixes an issue where dragging a quality below a group then trying to drag it back above doesn't work. For example given these two items:

Image

If you drag `Bluray-2160p` below the `WEB 2160p` group you can't drag it back up below `Bluray-2160p Remux` without dragging it above then back down.

```suggestion
    collisionPriority:
      mode === 'editGroups' ? CollisionPriority.Low : CollisionPriority.Normal,
```

frontend/src/Settings/Profiles/Quality/QualityProfileItemGroup.css
markus101: It's pretty hard to drag an item into a group from the bottom, I played around with it a bit and found adding `padding-bottom: 7px` helps, but looks a little off with the current styling because the last item looks taller. Not sure if there is something else to make that better.

markus101 (commented):
Working well, just a couple things.

frontend/src/Settings/Profiles/Quality/QualityProfileItemGroup.css
bertPB: > It's pretty hard to drag an item into a group from the bottom, I played around with it a bit and found adding `padding-bottom: 7px` helps, but looks a little off with the current styling because the last item looks taller. Not sure if there is something else to make that better.


Hmm, not sure I can replicate what you are talking about, I've tried a bunch of different group / item permutations, but no issues there.
Can you maybe post a clip of it somewhere?

markus101: Gave it a test while recording and it seems fine, not sure what I was running into before, but disregard.

markus101 (commented):
A couple more things I've noticed when playing with this a bit more:

1. The color difference between `paper` and `paperWarm` is fine when it's static, but a lot harder to differentiate while dragging
2. The border shown for separation is cutoff on the left side when it's in a group.

frontend/src/Settings/Profiles/Quality/useQualityProfileDnd.ts
bertPB: While I initially agreed, I'd argue we should only do this if we change the other functions to arrow functions as well. 
Or do you not mind having 2 different ways of declaring functions in a single file?

mynameisbogdan: I think we should be fine with using arrow functions for simple one-liner helpers like in this case.

bertPB: Alright, will just change these 2 then , thanks :)

bertPB:
> A couple more things I've noticed when playing with this a bit more:
> 
> 1. The color difference between `paper` and `paperWarm` is fine when it's static, but a lot harder to differentiate while dragging
> 2. The border shown for separation is cutoff on the left side when it's in a group.

Fixed both, let me know what you think

markus101 (approved):
Much better!

PR #8799, opened 2026-07-19
Fix: Defer series path update until file move completes

kushiemoon-dev:
#### Description

`SeriesEditorController.SaveAll` writes the new `Series.Path` to the DB synchronously while the actual folder move runs async through `BulkMoveSeriesCommand`. If the process crashes (or disk fills up) in between, the DB ends up pointing at a path the files never reached.

Same bug already fixed on the Radarr side (Sonarr/Radarr#11562). `SeriesService.UpdateSeries` now defers the path write for series pending a move (tracked via `Series.PendingPath`), and `MoveSeriesService` persists the confirmed path once the move actually succeeds, instead of the controller writing it eagerly.

Covers the single-series edit path too, no Radarr changes needed.

#### Database Migration

YES - 234

#### Submission Checklist


src/NzbDrone.Core/Tv/SeriesService.cs
```diff
@@ -31,6 +31,7 @@ public interface ISeriesService
         Dictionary<int, int> GetAllSeriesQualityProfiles();
         Series UpdateSeries(Series series, bool updateEpisodesToMatchSeason = true, bool publishUpdatedEvent = true);
         List<Series> UpdateSeries(List<Series> series, bool useExistingRelativeFolder);
+        List<Series> UpdateSeries(List<Series> series, bool useExistingRelativeFolder, bool deferPathUpdate);
```
mynameisbogdan:
```suggestion
        List<Series> UpdateSeries(List<Series> series, bool useExistingRelativeFolder, bool deferPathUpdate = false);
```

src/NzbDrone.Core/Tv/SeriesService.cs
```diff
@@ -235,14 +236,27 @@ public Series UpdateSeries(Series series, bool updateEpisodesToMatchSeason = tru
         }
 
         public List<Series> UpdateSeries(List<Series> series, bool useExistingRelativeFolder)
+        {
+            return UpdateSeries(series, useExistingRelativeFolder, false);
+        }
+
+        public List<Series> UpdateSeries(List<Series> series, bool useExistingRelativeFolder, bool deferPathUpdate)
```
mynameisbogdan:
```suggestion
        public List<Series> UpdateSeries(List<Series> series, bool useExistingRelativeFolder, bool deferPathUpdate = false)
```

mynameisbogdan (changes requested):

src/NzbDrone.Core/Tv/SeriesService.cs
```diff
@@ -235,14 +236,27 @@ public Series UpdateSeries(Series series, bool updateEpisodesToMatchSeason = tru
...
             {
                 _logger.Trace("Updating: {0}", s.Title);
 
-                if (!s.RootFolderPath.IsNullOrWhiteSpace())
+                if (deferPathUpdate)
+                {
+                    // The file move is queued as a separate, asynchronous command. Path is
+                    // updated by MoveSeriesService once the move has actually completed, so an
+                    // interrupted move (crash, full disk) doesn't leave the DB pointing at files
+                    // that were never actually relocated.
```
markus101: I don't think the comment is necessary here.

```suggestion
```

src/NzbDrone.Core/Tv/SeriesService.cs
```diff
@@ -235,14 +236,27 @@ public Series UpdateSeries(Series series, bool updateEpisodesToMatchSeason = tru
...
+                    _logger.Trace("Path update for {0} deferred until file move completes", s.Title);
```
markus101: I think the logging can be improved a bit and should move this into the original `if (!s.RootFolderPath.IsNullOrWhiteSpace())` so the current and future path can be logged.

Something like: 
```
_logger.Trace("Path for '{0}' will be updated from {1} to {2} after files are moved successfully", s.Title, s.Path, updatedPath);           
```

src/NzbDrone.Core/Tv/SeriesService.cs
```diff
@@ -244,9 +250,18 @@ public List<Series> UpdateSeries(List<Series> series, bool useExistingRelativeFo
...
 
-                    _logger.Trace("Changing path for {0} to {1}", s.Title, s.Path);
+                        _logger.Trace("Path for '{0}' will be updated from {1} to {2} after files are moved successfully", s.Title, s.Path, updatedPath);
+                    }
+                    else
+                    {
+                        s.Path = _seriesPathBuilder.BuildPath(s, useExistingRelativeFolder);
+
+                        _logger.Trace("Changing path for {0} to {1}", s.Title, s.Path);
+                    }
```
markus101: The path can just be built once.

```suggestion
                    var updatedPath = _seriesPathBuilder.BuildPath(s, useExistingRelativeFolder);

                    if (deferPathUpdate)
                    {
                        _logger.Trace("Path for '{0}' will be updated from {1} to {2} after files are moved successfully", s.Title, s.Path, updatedPath);
                    }
                    else
                    {
                        s.Path = updatedPath;

                        _logger.Trace("Changing path for {0} to {1}", s.Title, s.Path);
                    }
```

src/NzbDrone.Core/Tv/SeriesService.cs
```diff
@@ -31,6 +31,7 @@ public interface ISeriesService
         Dictionary<int, int> GetAllSeriesQualityProfiles();
         Series UpdateSeries(Series series, bool updateEpisodesToMatchSeason = true, bool publishUpdatedEvent = true);
         List<Series> UpdateSeries(List<Series> series, bool useExistingRelativeFolder);
+        List<Series> UpdateSeries(List<Series> series, bool useExistingRelativeFolder, bool deferPathUpdate = false);
```
markus101: Since we're setting a default I don't think we need to have an overload here.

src/Sonarr.Api.V5/Series/SeriesEditorController.cs
```diff
@@ -103,7 +103,7 @@ public Results<Ok<List<SeriesResource>>, BadRequest> SaveAll([FromBody] SeriesEd
             });
         }
 
-        return TypedResults.Ok(_seriesService.UpdateSeries(seriesToUpdate, !resource.MoveFiles).ToResource());
+        return TypedResults.Ok(_seriesService.UpdateSeries(seriesToUpdate, !resource.MoveFiles, resource.MoveFiles).ToResource());
```
markus101: This feels a little fragile, one param is the inverse of a boolean, but the next is just the same as the boolean.

Can we just just the `deferPathUpdate` param and always defer the path update when we're moving the files (`useExistingRelativeFolder` is false)

markus101 (changes requested):
Took another pass and have a few more thoughts.

1. The only time we really defer the path update is from the v5 API when are moving files, do we even need the extra boolean (which is just the inverse of `useExistingRelativeFolder`)?

2. The v3 API should be kept in sync with v5 in this case.

3. One case that isn't covered is editing a single series and moving the files, since it's updated directly we don't defer it, but it could suffer the same issue.

4. The other issue with deferring is the user has changed the path, but nowhere do we really show that they've changed it if the path updating is deferred, I think we actually need to store both paths in the DB and then clean up the "NextPath" once the files have been moved. While it's in progress we need to have something in the UI to indicate it's being moved.

src/Sonarr.Api.V5/Series/SeriesEditorController.cs
mynameisbogdan: I'm a little confused by this myself, what are actually asking for? 

This or something else?
```suggestion
        return TypedResults.Ok(_seriesService.UpdateSeries(seriesToUpdate, useExistingRelativeFolder: !resource.MoveFiles, deferPathUpdate: resource.MoveFiles).ToResource());
```
or ```var updatedPath = _seriesPathBuilder.BuildPath(s, !deferPathUpdate);```?

markus101: Just passing one boolean argument:

```
return TypedResults.Ok(_seriesService.UpdateSeries(seriesToUpdate, !resource.MoveFiles).ToResource());
```

kushiemoon-dev: Done in 829d814e6: dropped `deferPathUpdate` from the signature entirely, it's derived as `!useExistingRelativeFolder` inside `UpdateSeries`. Both editor controllers (v3 and v5) now just pass `!resource.MoveFiles`.

markus101:
Moving to draft until we can indicate which series are being moved (and prevent the same series from being moved multiple times).

src/NzbDrone.Core/Datastore/Migration/234_add_next_path_to_series.cs
```diff
@@ -0,0 +1,13 @@
...
+using NzbDrone.Core.Datastore.Migration.Framework;
+
+namespace NzbDrone.Core.Datastore.Migration;
+
+[Migration(233)]
+public class add_next_path_to_series : NzbDroneMigrationBase
+{
+    protected override void MainDbUpgrade()
+    {
+        Alter.Table("Series").AddColumn("NextPath").AsString().Nullable();
```
mynameisbogdan, line 11: `QueueMoveToPath`, `QueuedMovePath`, `PendingMovePath`, `PendingPath`?

kushiemoon-dev:
Pushed `NextPath` + the UI indicator + the concurrency guard: was going to leave those as follow-up issues (points 5/6 above), but they're really what closes the loop on the original race, so folded them in here instead.

`Series.NextPath` (migration 234, had to bump it after rebasing past the quality-rank migration) holds the pending destination while a deferred move is in flight. It gets cleared on all 5 exit paths of `MoveSingleSeries`, not just success/failure, but also the three early-return branches, otherwise an abandoned move leaves it stuck forever. `SeriesEditorController` (V3+V5) now skips queuing a second `BulkMoveSeriesCommand` for a series that already has one pending, without blocking the rest of the request. Series index shows a pending icon for it, riding the same SignalR broadcast the single-series edit already fires, so no new plumbing there. Exposed on both API resources for parity.

Caught one bug in self-review worth mentioning: had the move command getting queued before the `NextPath` write actually committed, which reintroduces the same stuck-state problem this PR exists to fix, since a same-volume move is just a rename, so it can finish and clear `NextPath` before the request even persisted it. Reordered so the write happens first.

Also added an integration test for the queue-skip guard, marked `[Explicit]`, since it relies on source/destination being on separate filesystems to force a slow enough copy to observe the race, which doesn't hold on most CI runners, so left it as a local check instead of something that'll just flake.

Branch had drifted behind `v5-develop` too, so rebased. That picked up a real conflict in `SeriesStatusCell.tsx` (the StatusIndicator accessibility change landed there) and the migration collision mentioned above. Full build + test suite green on top.

src/NzbDrone.Core/Datastore/Migration/234_add_next_path_to_series.cs
kushiemoon-dev: Went with `NextPath` since that's what markus101 called it in his review, but can swap if he prefers one of these.

mynameisbogdan: Personally I'm not a fan of NextPath, but if Mark is okay with it. 🤷‍♂️

kushiemoon-dev: i can easly change it if needed ? both are okay for me just need to know wich one will be the better.

kushiemoon-dev:
Keeping `NextPath`: no objections raised in the last few days, and it was Mark's original suggestion. Marking ready for review; UI indicator + concurrency guard from the draft note are in and tested.

mynameisbogdan:
I object! 

/jk

src/NzbDrone.Core/Datastore/Migration/234_add_next_path_to_series.cs
markus101: I think all the options aren't super clear or accurate (`NextPath` included), but the only other one that comes to mind is `NewPath` and I'm not sure it's really any better (possibly worse).

src/NzbDrone.Integration.Test/ApiTests/SeriesEditorFixture.cs
```diff
@@ -42,5 +46,77 @@ public void should_be_able_to_update_multiple_series()
...
+            EnsureNoSeries(266189, "The Blacklist");
+
+            // The source folder lives on a separate filesystem (tmpfs) from the root folders
+            // below (which live under the btrfs-backed test output dir), and is seeded with real
+            // files. This forces the first move to be a genuine cross-filesystem file copy rather
+            // than an instant same-volume rename, so it is still in flight when the second request
+            // arrives - without this, the move (and the NextPath clear that follows it) completes
+            // before the second HTTP round-trip, and the guard can never be observed as pending.
+            var sourcePath = Path.Combine(Path.GetTempPath(), "sonarr_test_src_" + Guid.NewGuid());
+            Directory.CreateDirectory(sourcePath);
```
markus101: Lots of 🤖 looking comments here

src/NzbDrone.Integration.Test/ApiTests/SeriesEditorFixture.cs
```diff
@@ -42,5 +46,77 @@ public void should_be_able_to_update_multiple_series()
...
+            // non-confounded proof the guard worked. This count check is known-confoundable by a
+            // separate bug in CommandEqualityComparer, which compares string properties (e.g.
+            // DestinationRootFolder) as character sets via IEnumerable<char> rather than by value,
+            // so it can under-report the true command count independent of whether the guard ran.
+            Commands.All().Count(c => c.Name == "BulkMoveSeries").Should().Be(queuedAfterFirst);
+
+            // Cleanup only: let the in-flight move finish before TearDown deletes the temp dirs it
+            // is writing into. Must stay after all assertions above - the test's signal depends on
+            // the second request racing the still-running first move.
+            Commands.WaitAll();
```
markus101: Same here. 🤖

markus101: Same here. 🤖

frontend/src/Series/Index/Table/SeriesStatusCell.tsx
```diff
@@ -70,6 +72,16 @@ function SeriesStatusCell({
...
+
+      {nextPath ? (
+        <StatusIndicator
+          className={styles.statusIcon}
+          label={translate('SeriesMovePending', { path: nextPath })}
+          title={translate('SeriesMovePending', { path: nextPath })}
+        >
+          <Icon name={icons.PENDING} />
+        </StatusIndicator>
+      ) : null}
```
markus101: I think this should be next to the path instead.

Also needs to be shown in other places where the path is shown (series details), edit modal.

markus101 (commented):
Thanks for the changes, the direction looks good. We also need to handle moving a single series by editing the path (goes through `UpdateSeries` in the `SeriesController`

kushiemoon-dev:
Single-series edit had the same race: UpdateSeries persisted the requested Path immediately while MoveSeriesCommand ran async. Applied the same defer-behind-NextPath fix there for both V3 and V5, plus fixed the broadcast to reflect persisted state instead of the raw requested path (it was firing the client's Path regardless of moveFiles before).

No V5 integration test client existed in the repo at all, so added a minimal one (just Put with moveFiles, nothing else) and 4 tests per version mirroring the bulk ones already here.

Also caught myself repeating the same command-ordering bug this PR already fixed once, pushing MoveSeriesCommand before the NextPath write landed. Same fix, persist first then push.

Ran the suite locally (SDK 10.0.302): 13/15 pass, 2 skipped (SignalR, same as RootFolderFixture). Both Explicit pairs pass consistently across repeated runs.

kushiemoon-dev:
Pushed that yesterday actually, see the comment above - single-series UpdateSeries (V3+V5) defers behind NextPath the same way now, broadcast fixed too. Was already in flight when you posted this.

kushiemoon-dev:
Fixed a few things from a closer pass: Test2 in both fixtures was only checking NextPath, not Path: the actual thing the bug clobbered. Also SignalR test connection was broken (wrong port assumption plus missing access_token for the hub's auth), fixed both and un-ignored those two tests since they actually pass now.

Bigger one: only 1 of 4 tests per API version runs in CI, the rest are Explicit/Ignore, so the guard logic had no real coverage. Added mocked controller tests for it (first ones in this repo), then checked they'd actually catch the bug by reverting the ordering fix locally and confirming they fail.

src/NzbDrone.Integration.Test/ApiTests/SeriesEditorFixture.cs
```diff
@@ -42,5 +46,65 @@ public void should_be_able_to_update_multiple_series()
             result.Should().HaveCount(2);
             result.TrueForAll(s => s.QualityProfileId == 2).Should().BeTrue();
         }
+
+        [Test]
+        [Explicit("Requires source and destination temp directories on separate filesystems to reliably observe the in-flight move window; not reliable on CI runners where they share a volume.")]
+        public void should_not_queue_a_second_move_for_a_series_with_a_move_already_pending()
```
markus101: I don't see value in a test virtually no one will run.

src/NzbDrone.Integration.Test/ApiTests/SeriesFixture.cs
```diff
@@ -178,5 +180,141 @@ public void delete_series()
 
             Series.All().Should().NotContain(v => v.TvdbId == 266189);
         }
+
+        [Test]
+        [Explicit("Requires source and destination temp directories on separate filesystems to reliably observe the in-flight move window; not reliable on CI runners where they share a volume.")]
+        public void update_series_with_movefiles_should_defer_path_update_until_move_completes()
```
markus101: Same here

src/NzbDrone.Integration.Test/ApiTests/SeriesFixture.cs
```diff
@@ -178,5 +180,141 @@ public void delete_series()
...
+
+            var afterMove = Series.Get(series.Id);
+
+            afterMove.Path.Should().Be(destinationPath);
+            afterMove.NextPath.Should().BeNullOrEmpty();
+        }
+
+        [Test]
+        [Explicit("Requires source and destination temp directories on separate filesystems to reliably observe the in-flight move window; not reliable on CI runners where they share a volume.")]
+        public void update_series_with_movefiles_should_not_queue_second_move_when_one_already_pending()
```
markus101: And here.

src/NzbDrone.Integration.Test/ApiTests/SeriesV5Fixture.cs
```diff
@@ -0,0 +1,156 @@
+using System;
```
markus101: Let's drop this for now, we'll move API tests over to v5 later on.

src/NzbDrone.Integration.Test/Client/SeriesClientV5.cs
```diff
@@ -0,0 +1,21 @@
+using RestSharp;
```
markus101: here as well, not necessary.

src/Sonarr.Api.V5/Series/SeriesEditorController.cs
```diff
@@ -60,11 +60,15 @@ public Results<Ok<List<SeriesResource>>, BadRequest> SaveAll([FromBody] SeriesEd
             if (resource.RootFolderPath.IsNotNullOrWhiteSpace())
             {
                 series.RootFolderPath = resource.RootFolderPath;
-                seriesToMove.Add(new BulkMoveSeries
+
+                if (series.NextPath.IsNullOrWhiteSpace())
```
markus101: Shouldn't this already be handled in the service? If not it should so the controller doesn't need to understand how `NextPath` is used.

If this is meant to prevent it from being triggered twice, then we shouldn't set the root folder path either.

src/Sonarr.Api.V3/Series/SeriesController.cs
```diff
@@ -182,12 +182,26 @@ public ActionResult<SeriesResource> AddSeries([FromBody] SeriesResource seriesRe
         public ActionResult<SeriesResource> UpdateSeries([FromBody] SeriesResource seriesResource, [FromQuery] bool moveFiles = false)
         {
             var series = _seriesService.GetSeries(seriesResource.Id);
+            var sourcePath = series.Path;
+            var destinationPath = seriesResource.Path;
+            var moveAlreadyPending = series.NextPath.IsNotNullOrWhiteSpace();
```
markus101: Every use of this becomes a double negative because this is checking if it's NotNull then they all check if it's not `moveAlreadyPending`. I think it'd be better to flip this boolean to something else to avoid the negative here + on both usages.

src/NzbDrone.Integration.Test/IntegrationTestBase.cs
```diff
@@ -103,6 +104,10 @@ protected virtual void InitRestClients()
             RestClient.AddDefaultHeader("Authentication", ApiKey);
             RestClient.AddDefaultHeader("X-Api-Key", ApiKey);
 
+            var restClientV5 = new RestClient(RootUrl + "api/v5/");
```
markus101: Related to dropping v5 tests for now.

src/NzbDrone.Integration.Test/ApiTests/SeriesFixture.cs
```diff
@@ -178,5 +180,141 @@ public void delete_series()
...
             Series.All().Should().NotContain(v => v.TvdbId == 266189);
         }
+
+        [Test]
+        [Explicit("Requires source and destination temp directories on separate filesystems to reliably observe the in-flight move window; not reliable on CI runners where they share a volume.")]
+        public void update_series_with_movefiles_should_defer_path_update_until_move_completes()
+        {
+            EnsureNoSeries(266189, "The Blacklist");
+
+            // Source on tmpfs, dest on btrfs - forces a real cross-filesystem copy so the
```
markus101: 🤖 comment?

src/NzbDrone.Integration.Test/ApiTests/SeriesFixture.cs
```diff
@@ -178,5 +180,141 @@ public void delete_series()
...
+            afterMove.NextPath.Should().BeNullOrEmpty();
+        }
+
+        [Test]
+        [Explicit("Requires source and destination temp directories on separate filesystems to reliably observe the in-flight move window; not reliable on CI runners where they share a volume.")]
+        public void update_series_with_movefiles_should_not_queue_second_move_when_one_already_pending()
+        {
+            EnsureNoSeries(266189, "The Blacklist");
+
+            // Source on tmpfs, dest on btrfs - forces a real cross-filesystem copy so the
```
markus101: Also 🤖 comment?

src/NzbDrone.Core/Tv/SeriesService.cs
```diff
@@ -238,15 +238,36 @@ public List<Series> UpdateSeries(List<Series> series, bool useExistingRelativeFo
...
             foreach (var s in series)
             {
                 _logger.Trace("Updating: {0}", s.Title);
 
                 if (!s.RootFolderPath.IsNullOrWhiteSpace())
                 {
-                    s.Path = _seriesPathBuilder.BuildPath(s, useExistingRelativeFolder);
+                    if (deferPathUpdate && s.NextPath.IsNotNullOrWhiteSpace())
+                    {
+                        _logger.Trace("Not queuing another path update for '{0}', a move to {1} is already pending", s.Title, s.NextPath);
```
markus101:
```suggestion
                        _logger.Trace("Series '{0}' is already being moved from {1} to {2}, skipping path update.", s.Title, s.Path, s.NextPath);
```

src/NzbDrone.Core/Localization/Core/en.json
```diff
@@ -2019,6 +2017,7 @@
   "SeriesLoadError": "Unable to load Series",
   "SeriesMatchType": "Series Match Type",
   "SeriesMonitoring": "Series Monitoring",
+  "SeriesMovePending": "A move to '{path}' is pending",
```
markus101:
```suggestion
  "SeriesMovePending": "Waiting for {appName} to move files to '{path}'",
```

frontend/src/Series/Details/SeriesDetails.tsx
```diff
@@ -632,6 +634,13 @@ function SeriesDetails({ seriesId }: SeriesDetailsProps) {
                     <div>
                       <Icon name={icons.FOLDER} size={17} />
                       <span className={styles.path}>{path}</span>
+
+                      {nextPath ? (
```
markus101: Looks like the alignment is off

src/NzbDrone.Api.Test/Series/SeriesControllerUpdateSeriesFixture.cs
```diff
@@ -0,0 +1,111 @@
+using System.Collections.Generic;
```
markus101: > Bigger one: only 1 of 4 tests per API version runs in CI, the rest are Explicit/Ignore, so the guard logic had no real coverage. Added mocked controller tests for it (first ones in this repo) - checked they'd actually catch the bug by reverting the ordering fix locally and confirming they fail.

This is a sign to me that there is too much business logic int he controllers and needs to be moved to a service instead of layering on new types of tests.

frontend/src/Series/SeriesMovePendingIcon.tsx
```diff
@@ -0,0 +1,24 @@
...
+import Icon from 'Components/Icon';
+import { icons } from 'Helpers/Props';
+import translate from 'Utilities/String/translate';
+
+interface SeriesMovePendingIconProps {
+  className?: string;
+  nextPath: string;
+}
+
+function SeriesMovePendingIcon({
```
markus101: `SeriesMovePending` is a better name

markus101 (commented):
Mostly concerned with the test coverage that was added that introduces new testing patterns or doesn't actually run the tests, but that's a symptom of too much business logic living in the controller so addressing that will clean up a good chunk of verbose tests and allow us to test the services instead.

kushiemoon-dev:
Thanks for all the review i sleep and take a look on all of that i thought the test coverage was not enought on what i done but visibly it have only introduce bad one 🥲 i get my hand on that !

src/NzbDrone.Core/Datastore/Migration/234_add_next_path_to_series.cs
kushiemoon-dev: Renamed to `PendingPath`: nobody landed on a name everyone liked, `NextPath` included, so went with the clearest one on the table.

src/NzbDrone.Integration.Test/ApiTests/SeriesEditorFixture.cs
kushiemoon-dev: Removed.

src/NzbDrone.Integration.Test/ApiTests/SeriesFixture.cs
kushiemoon-dev: Removed.

src/NzbDrone.Integration.Test/ApiTests/SeriesFixture.cs
kushiemoon-dev: Same, removed.

src/NzbDrone.Integration.Test/ApiTests/SeriesV5Fixture.cs
kushiemoon-dev: Deleted.

src/NzbDrone.Integration.Test/Client/SeriesClientV5.cs
kushiemoon-dev: Deleted this too.

src/Sonarr.Api.V5/Series/SeriesEditorController.cs
kushiemoon-dev: Moved into `SeriesService`, since the controller doesn't need to understand `PendingPath` anymore.

src/Sonarr.Api.V3/Series/SeriesController.cs
kushiemoon-dev: Gone with the refactor: moved into the service and flipped it positive (`HasPendingMove`).

src/NzbDrone.Integration.Test/IntegrationTestBase.cs
kushiemoon-dev: Reverted along with dropping the V5 client.

src/NzbDrone.Core/Tv/SeriesService.cs
kushiemoon-dev: Used your wording.

src/NzbDrone.Core/Localization/Core/en.json
kushiemoon-dev: Used your wording.

frontend/src/Series/Details/SeriesDetails.tsx
kushiemoon-dev: Missing an align rule on the icon: added `vertical-align: middle`, lines up with the path text now.

src/NzbDrone.Api.Test/Series/SeriesControllerUpdateSeriesFixture.cs
kushiemoon-dev: Should've moved this into the service instead of layering on more test types. Moved the defer/guard/broadcast logic into `SeriesService` for single-series and bulk both, same path both API versions hit now. Deleted `SeriesControllerUpdateSeriesFixture.cs` and the V5 one, no controller-level Moq tests left in the repo, covered by service-level tests instead alongside `SeriesServiceTests`. Dropped the Explicit tests and V5 client too since the service tests actually run in CI.

frontend/src/Series/SeriesMovePendingIcon.tsx
kushiemoon-dev: Renamed to `SeriesMovePending`.

src/NzbDrone.Core/Tv/SeriesService.cs
kushiemoon-dev: Gone: dropped `deferPathUpdate` from the signature in 829d814e6/d63dec4, back to just `useExistingRelativeFolder` plus the `out` overload.

src/NzbDrone.Core/Tv/SeriesService.cs
kushiemoon-dev: Same as above, gone in 829d814e6.

src/NzbDrone.Core/Tv/SeriesService.cs
kushiemoon-dev: Gone too, cleaned out in d63dec4.

src/NzbDrone.Core/Tv/SeriesService.cs
kushiemoon-dev: Moved in d63dec4, log's inside the `HasPendingMove` branch now.

src/NzbDrone.Core/Tv/SeriesService.cs
kushiemoon-dev: That overload's gone, `deferPathUpdate` dropped in 829d814e6.

src/NzbDrone.Integration.Test/ApiTests/SeriesEditorFixture.cs
kushiemoon-dev: Fixture's gone, dropped in the same cleanup as the thread below.

src/NzbDrone.Integration.Test/ApiTests/SeriesEditorFixture.cs
kushiemoon-dev: Same, gone with the rest of that fixture.

frontend/src/Series/Index/Table/SeriesStatusCell.tsx
kushiemoon-dev: Done, and moved into a shared `SeriesMovePending` component (b5bfa603b), which now shows in Details + Edit modal too.

src/NzbDrone.Core/Datastore/Migration/234_add_next_path_to_series.cs
```diff
@@ -0,0 +1,13 @@
+using FluentMigrator;
+using NzbDrone.Core.Datastore.Migration.Framework;
+
+namespace NzbDrone.Core.Datastore.Migration;
+
+[Migration(234)]
+public class add_next_path_to_series : NzbDroneMigrationBase
```
markus101: File and class name should be updated to `PendingPath`

src/NzbDrone.Core/Tv/MoveSeriesService.cs
```diff
@@ -40,12 +40,14 @@ private void MoveSingleSeries(Series series, string sourcePath, string destinati
...
             {
                 _logger.Warn("Folder '{0}' for '{1}' is invalid, unable to move series. Try moving files manually", sourcePath, series.Title);
+                UpdatePath(series.Id, sourcePath);
                 return;
             }
 
             if (!_diskProvider.FolderExists(sourcePath))
             {
                 _logger.Debug("Folder '{0}' for '{1}' does not exist, not moving.", sourcePath, series.Title);
+                UpdatePath(series.Id, sourcePath);
```
markus101: If the folder doesn't exist we should just update the path, this could happen if a user hasn't yet imported any files, but they want to bulk move all series that do have files.

src/NzbDrone.Core/Tv/MoveSeriesService.cs
```diff
@@ -40,12 +40,14 @@ private void MoveSingleSeries(Series series, string sourcePath, string destinati
             if (!sourcePath.IsPathValid(PathValidationType.CurrentOs))
             {
                 _logger.Warn("Folder '{0}' for '{1}' is invalid, unable to move series. Try moving files manually", sourcePath, series.Title);
+                UpdatePath(series.Id, sourcePath);
```
markus101: Same here, if the source path isn't valid we should just update the path and tell them to move the files manually, otherwise we mostly sliently throw out their path change.

src/NzbDrone.Core/Tv/SeriesService.cs
```diff
@@ -235,18 +238,46 @@ public Series UpdateSeries(Series series, bool updateEpisodesToMatchSeason = tru
         }
 
         public List<Series> UpdateSeries(List<Series> series, bool useExistingRelativeFolder)
+        {
+            return UpdateSeries(series, useExistingRelativeFolder, out _);
+        }
+
+        public List<Series> UpdateSeries(List<Series> series, bool useExistingRelativeFolder, out List<BulkMoveSeries> seriesToMove)
```
markus101: We don't need the out, we can just trigger to move from the service, instead of having that logic live in the controller.

src/NzbDrone.Core/Tv/SeriesService.cs
```diff
@@ -235,18 +238,46 @@ public Series UpdateSeries(Series series, bool updateEpisodesToMatchSeason = tru
...
                 _logger.Trace("Updating: {0}", s.Title);
 
                 if (!s.RootFolderPath.IsNullOrWhiteSpace())
                 {
-                    s.Path = _seriesPathBuilder.BuildPath(s, useExistingRelativeFolder);
-
-                    _logger.Trace("Changing path for {0} to {1}", s.Title, s.Path);
+                    if (deferPathUpdate && HasPendingMove(s))
+                    {
+                        _logger.Trace("Series '{0}' is already being moved from {1} to {2}, skipping path update.", s.Title, s.Path, s.PendingPath);
```
markus101: This should be updated to a warning so it's not as quite.

src/NzbDrone.Core/Tv/SeriesService.cs
```diff
@@ -235,18 +238,46 @@ public Series UpdateSeries(Series series, bool updateEpisodesToMatchSeason = tru
         }
 
         public List<Series> UpdateSeries(List<Series> series, bool useExistingRelativeFolder)
```
markus101: I think it might make sense to just call `useExistingRelativeFolder`, `moveFiles`, the only other caller is from `ImportListSyncService` (and related tests) so an easy fix.

It'd eliminate the need for `deferPathUpdate`

src/NzbDrone.Core/Tv/SeriesService.cs
```diff
@@ -260,9 +291,45 @@ public List<Series> UpdateSeries(List<Series> series, bool useExistingRelativeFo
...
+
+                _logger.Trace("Changing path for {0} to {1}", series.Title, series.Path);
+            }
+
+            return UpdateSeries(series, true, true);
+        }
+
+        private static bool HasPendingMove(Series series)
+        {
+            return series.PendingPath.IsNotNullOrWhiteSpace();
```
markus101: If the move stops part-way through because Sonarr was restarted the user would never be able to change the path again. Probably need something to check whether a move is pending, but there isn't a corresponding command queued.

src/NzbDrone.Core/Tv/SeriesService.cs
```diff
@@ -235,18 +238,46 @@ public Series UpdateSeries(Series series, bool updateEpisodesToMatchSeason = tru
...
+                        {
+                            s.PendingPath = updatedPath;
+                            moveQueue.Add(new BulkMoveSeries { SeriesId = s.Id, SourcePath = s.Path });
+
+                            _logger.Trace("Path for '{0}' will be updated from {1} to {2} after files are moved successfully", s.Title, s.Path, updatedPath);
+                        }
+                        else
+                        {
+                            s.Path = updatedPath;
+                            s.PendingPath = null;
```
markus101, line 278: If the user updates paths and has Sonarr move files, then updates something unrelated this would stomp on their pending path and cause the move to fail if it hasn't started or lose their new path if it was in progress.

src/Sonarr.Api.V3/Series/SeriesResource.cs
```diff
@@ -178,6 +180,7 @@ public static NzbDrone.Core.Tv.Series ToModel(this SeriesResource resource)
                        OriginalLanguage = resource.OriginalLanguage,
 
                        Path = resource.Path,
+                       PendingPath = resource.PendingPath,
```
markus101: Similar to the bulk update setting `PendingPath` to null, this could lead to it being lost (also applies to v5), we probably don't want to have the API allow overwriting, otherwise an API caller could trigger a move then overwrite it.

kushiemoon-dev:
The test for the last push is in test on a real instance at the moment.

src/NzbDrone.Core/Tv/SeriesService.cs
```diff
@@ -260,9 +290,74 @@ public List<Series> UpdateSeries(List<Series> series, bool useExistingRelativeFo
...
+                {
+                    series.PendingPath = destinationPath;
+                    queueMove = true;
+
+                    _logger.Trace("Path for '{0}' will be updated from {1} to {2} after files are moved successfully", series.Title, series.Path, destinationPath);
+                }
+            }
+            else
+            {
+                series.Path = destinationPath;
```
markus101, line 328: Do we also want to clear the `PendingPath` here?

What's the expected result if the user as queued up a series to move then decides to not move the files and updates the path again?

src/NzbDrone.Core/Tv/SeriesService.cs
```diff
@@ -260,9 +290,74 @@ public List<Series> UpdateSeries(List<Series> series, bool useExistingRelativeFo
...
+        {
+            var queueMove = false;
+
+            if (moveFiles)
+            {
+                series.Path = sourcePath;
+
+                if (HasPendingMove(series))
+                {
+                    _logger.Trace("Series '{0}' is already being moved from {1} to {2}, skipping path update.", series.Title, series.Path, series.PendingPath);
```
markus101: Shouldn't this also be a warning?

src/NzbDrone.Integration.Test/Sonarr.Integration.Test.csproj
```diff
@@ -4,11 +4,12 @@
     <OutputType>Library</OutputType>
   </PropertyGroup>
   <ItemGroup>
-    <PackageReference Include="Microsoft.AspNetCore.SignalR.Client" Version="10.0.11" />
+    <PackageReference Include="Microsoft.AspNetCore.SignalR.Client" Version="10.0.10" />
   </ItemGroup>
   <ItemGroup>
     <ProjectReference Include="..\NzbDrone.Test.Common\Sonarr.Test.Common.csproj" />
     <ProjectReference Include="..\Sonarr.Api.V3\Sonarr.Api.V3.csproj" />
+    <ProjectReference Include="..\Sonarr.Api.V5\Sonarr.Api.V5.csproj" />
```
markus101: Not needed any more.

frontend/src/Series/Series.ts
```diff
@@ -82,6 +82,7 @@ interface Series extends ModelBase {
   monitored: boolean;
   monitorNewItems: MonitorNewItems;
   network: string;
+  pendingPath?: string;
```
markus101: Needs to be re-ordered to keep alphabetical ordering.

src/NzbDrone.Integration.Test/ApiTests/SeriesFixture.cs
```diff
@@ -178,5 +178,43 @@ public void delete_series()
...
+            series.Path = destinationPath;
+
+            ConnectSignalR().Wait();
+
+            Series.Put(series, moveFiles: true);
+
+            WaitForCompletion(() => SignalRMessages.Any(m =>
+                m.Name == "series" &&
+                ((System.Text.Json.JsonElement)m.Body).GetProperty("resource").GetProperty("path").GetString() == oldPath &&
+                ((System.Text.Json.JsonElement)m.Body).GetProperty("resource").GetProperty("pendingPath").GetString() == destinationPath));
```
markus101: Can we actually import `System.Text.Json`? Can we avoid casting this twice?

src/NzbDrone.Core/Tv/SeriesService.cs
```diff
@@ -234,19 +240,43 @@ public Series UpdateSeries(Series series, bool updateEpisodesToMatchSeason = tru
             return updatedSeries;
         }
 
-        public List<Series> UpdateSeries(List<Series> series, bool useExistingRelativeFolder)
+        public List<Series> UpdateSeries(List<Series> series, bool moveFiles)
         {
             _logger.Debug("Updating {0} series", series.Count);
 
+            var moveQueue = new List<BulkMoveSeries>();
```
markus101, line 250: To avoid calling `HasPendingMove` and fetching the command queue `n` times:

```suggestion
            var moveQueue = new List<BulkMoveSeries>();
            var seriesIdsWithPendingMove = GetSeriesIdsWithPendingMove();
```

Along with a change to `HasPendingMove` and a new method to get the series IDs.

```
private HashSet<int> GetSeriesIdsWithPendingMove()
{
    return _commandQueueManager.All()
        .Where(c => c.Status is CommandStatus.Queued or CommandStatus.Started)
        .SelectMany(c => c.Body switch
        {
            MoveSeriesCommand move => new[] { move.SeriesId },
            BulkMoveSeriesCommand bulk => bulk.Series.Select(m => m.SeriesId),
            _ => Enumerable.Empty<int>()
        })
        .ToHashSet();
}

private static bool HasPendingMove(Series series, HashSet<int> seriesIdsWithPendingMove)
{
    return series.PendingPath.IsNotNullOrWhiteSpace() && seriesIdsWithPendingMove.Contains(series.Id);
}
```

src/NzbDrone.Core.Test/TvTests/SeriesServiceTests/UpdateMultipleSeriesFixture.cs
```diff
@@ -50,22 +74,53 @@ public void should_update_path_when_rootFolderPath_is_supplied()
...
+            var result = Subject.UpdateSeries(_series, true);
+
+            result.Should().OnlyContain(s => s.Path == originalPaths[s.Id]);
+        }
+
         [Test]
         public void should_not_update_path_when_rootFolderPath_is_empty()
         {
-            Subject.UpdateSeries(_series, false).ForEach(s =>
+            Subject.UpdateSeries(_series, true).ForEach(s =>
```
markus101: Looks like flipping this is to `true` makes it identical to `should_not_update_path_when_deferred_and_rootFolderPath_is_empty` below, should one of these tests be updated?

src/NzbDrone.Core.Test/TvTests/SeriesServiceTests/UpdateMultipleSeriesFixture.cs
```diff
@@ -50,22 +74,53 @@ public void should_update_path_when_rootFolderPath_is_supplied()
...
             Mocker.GetMock<IBuildSeriesPaths>()
-                  .Setup(s => s.BuildPath(It.IsAny<Series>(), false))
+                  .Setup(s => s.BuildPath(It.IsAny<Series>(), true))
                   .Returns<Series, bool>((s, u) => Path.Combine(s.RootFolderPath, s.Title));
 
             Subject.UpdateSeries(_series, false).ForEach(s => s.Path.Should().StartWith(newRoot));
         }
 
+        [Test]
+        public void should_not_update_path_for_series_pending_a_file_move()
```
markus101: Doesn't look like this is actually testing what it says.

src/NzbDrone.Core.Test/TvTests/SeriesServiceTests/UpdateMultipleSeriesFixture.cs
```diff
@@ -82,7 +137,7 @@ public void should_be_able_to_update_many_series()
                   .Setup(s => s.GetSeriesFolder(It.IsAny<Series>(), (NamingConfig)null))
                   .Returns<Series, NamingConfig>((s, n) => s.Title);
 
-            Subject.UpdateSeries(series, false);
+            Subject.UpdateSeries(series, true);
```
markus101: Does this need `true`? The test is about updating many series, and `true` now it also adds a bulk move that isn't checked for.

src/NzbDrone.Core.Test/TvTests/SeriesServiceTests/UpdateMultipleSeriesFixture.cs
```diff
@@ -98,9 +153,118 @@ public void should_add_and_remove_tags()
                     TagsToRemove = new HashSet<int> { 1 }
                 });
 
-            var result = Subject.UpdateSeries(_series, false);
+            var result = Subject.UpdateSeries(_series, true);
```
markus101: Same here, we don't need to move files.

src/NzbDrone.Core.Test/TvTests/MoveSeriesServiceFixture.cs
```diff
@@ -128,6 +140,67 @@ public void should_skip_series_folder_if_it_does_not_exist()
...
+            Subject.Execute(_command);
+
+            ExceptionVerification.ExpectedErrors(1);
+
+            Mocker.GetMock<ISeriesService>()
+                  .Verify(v => v.UpdateSeries(It.Is<Series>(s => s.PendingPath == null), It.IsAny<bool>(), It.IsAny<bool>()), Times.Once());
+        }
+
+        [Test]
+        public void should_clear_next_path_when_already_at_destination()
```
markus101: These three still say `next_path` after the rename to `PendingPath`.

src/NzbDrone.Core.Test/TvTests/MoveSeriesServiceFixture.cs
```diff
@@ -86,6 +86,18 @@ public void should_revert_series_path_on_error()
                   .Verify(v => v.UpdateSeries(It.IsAny<Series>(), It.IsAny<bool>(), It.IsAny<bool>()), Times.Once());
         }
 
+        [Test]
+        public void should_update_series_path_only_after_successful_move()
```
markus101: Since we're not asserting ordering, a better name would be:

```suggestion
        public void should_update_series_path_after_successful_move()
```

markus101 (changes requested):
If Sonarr goes down after the move has started, but before it completes then `PendingPath` is going to be stuck in limbo.

src/NzbDrone.Core/Tv/SeriesService.cs
```diff
@@ -260,9 +290,74 @@ public List<Series> UpdateSeries(List<Series> series, bool useExistingRelativeFo
...
+
+        private bool HasPendingMove(Series series)
+        {
+            if (series.PendingPath.IsNullOrWhiteSpace())
+            {
+                return false;
+            }
+
+            return _commandQueueManager.All().Any(c =>
+                (c.Status == CommandStatus.Queued || c.Status == CommandStatus.Started) &&
```
mynameisbogdan:
```suggestion
                c.Status is CommandStatus.Queued or CommandStatus.Started &&
```

mynameisbogdan (changes requested):

src/NzbDrone.Integration.Test/ApiTests/SeriesFixture.cs
mynameisbogdan:
```suggestion
                m.Name == "series" &&
                m.Body is JsonElement body &&
                body.GetProperty("resource").GetProperty("path").GetString() == oldPath &&
                body.GetProperty("resource").GetProperty("pendingPath").GetString() == destinationPath));
```

mynameisbogdan (changes requested):

src/NzbDrone.Core/Tv/SeriesService.cs
```diff
@@ -260,9 +292,133 @@ public List<Series> UpdateSeries(List<Series> series, bool useExistingRelativeFo
...
+                    {
+                        seriesIds.Add(m.SeriesId);
+                    }
+                }
+            }
+
+            return seriesIds;
+        }
+
+        public void Handle(ApplicationStartedEvent message)
```
markus101, line 391: This should be handled last so we don't try to handle this before the queue is actually ready. Alternatively we could have a queue initialized event, but I think this is good enough.

```suggestion
        [EventHandleOrder(EventHandleOrder.Last)]
        public void Handle(ApplicationStartedEvent message)
```

src/NzbDrone.Core/Tv/SeriesService.cs
```diff
@@ -260,9 +292,133 @@ public List<Series> UpdateSeries(List<Series> series, bool useExistingRelativeFo
...
+                series.PendingPath = null;
+
+                _logger.Trace("Changing path for {0} to {1}", series.Title, series.Path);
+            }
+
+            var updatedSeries = UpdateSeries(series, true, true);
+
+            if (queueMove)
+            {
+                _commandQueueManager.Push(new MoveSeriesCommand
```
markus101: `HasPendingMove` is a TOCTOU check: it reads the in-memory queue here, but the `PendingPath` write happens in `UpdateSeries(series, true, true)` below and the `MoveSeriesCommand` is only pushed after that. Nothing reserves the series in between.

Two near-simultaneous PUTs with `moveFiles=true` for the same series (a double-submitted Save is enough) both see an empty `PendingPath` and no queued command, so both persist a pending path and both push a move. The first moves `A -> B`; the second then finds `A` missing, takes the `!FolderExists` branch in `MoveSingleSeries`, and sets `Path = C` while the files are actually at `B`.

Two ways to close it:

- Cheap, and consistent with the rest of the codebase: a `private static readonly object` guarding check-write-push here and in the `moveFiles` branch of the bulk `UpdateSeries`. The critical section is one row update plus a queue push, so contention is nil.
- Correct: make the reservation a conditional write — a repository method that sets `PendingPath` only `WHERE Id = @id AND PendingPath IS NULL`, and branch on rows-affected (zero means someone else won; log the same warning and skip). This also makes the DB, rather than the in-memory queue, the source of truth for the reservation.

src/NzbDrone.Core/Tv/MoveSeriesService.cs
```diff
@@ -73,21 +76,24 @@ private void MoveSingleSeries(Series series, string sourcePath, string destinati
...
+                UpdatePath(series.Id, destinationPath);
+
                 _eventAggregator.PublishEvent(new SeriesMovedEvent(series, sourcePath, destinationPath));
             }
             catch (IOException ex)
             {
                 _logger.Error(ex, "Unable to move series from '{0}' to '{1}'. Try moving files manually", sourcePath, destinationPath);
 
-                RevertPath(series.Id, sourcePath);
+                UpdatePath(series.Id, sourcePath);
```
markus101: `PendingPath` is now only cleared by `UpdatePath` or the startup sweep, so a move that neither completes nor throws `IOException` leaves it set indefinitely: the UI keeps showing "Waiting for Sonarr to move files to X" and `Path` never changes until the next restart. Two gaps:

**Non-`IOException` failure** (e.g. `UnauthorizedAccessException` out of `TransferFolder`) — add a rethrowing catch below this one so the path reverts and the command still reports Failed:

```csharp
catch (Exception)
{
    UpdatePath(series.Id, sourcePath);
    throw;
}
```

Note that for `BulkMoveSeriesCommand` this aborts the remaining series, whose pending paths then wait for the startup sweep — if that matters, catch per-iteration in `Execute(BulkMoveSeriesCommand)` so the loop continues.

**Cancelled mid-run** — `CommandExecutor` publishes `CommandExecutedEvent(commandModel)` on every exit path, including cancellation and failure. Handling that in `SeriesService` and clearing `PendingPath` for any series id in a `MoveSeriesCommand`/`BulkMoveSeriesCommand` body is a single backstop covering both gaps, and can reuse the id extraction already in `GetSeriesIdsWithPendingMove`. Worth confirming whether cancelling a *queued* (never-started) command reaches that publish — if not, that case still falls to the startup sweep.

src/NzbDrone.Core/Tv/MoveSeriesService.cs
```diff
@@ -73,21 +76,24 @@ private void MoveSingleSeries(Series series, string sourcePath, string destinati
 
                 _logger.ProgressInfo("{0} moved successfully to {1}", series.Title, destinationPath);
 
+                UpdatePath(series.Id, destinationPath);
+
                 _eventAggregator.PublishEvent(new SeriesMovedEvent(series, sourcePath, destinationPath));
             }
             catch (IOException ex)
```
markus101, line 84: We should also catch `Exception` so it doesn't leave `PendingPath` abandoned.

```
catch (Exception)
{
    UpdatePath(series.Id, sourcePath);
    throw;
}
```

markus101 (changes requested):
Looks like there are still some outstanding changes, also have a couple additional things to look at.

kushiemoon-dev:
Hi @markus101,

Checking in on this PR, it's been about a week and a half without follow-up since your review on September 5th. Wanted to confirm if anything specific is still blocking a fresh review pass.

Since then, I've replied to each comment inline and pushed changes addressing:
- The NextPath approach
- V3/V5 parity
- The single-series edit path
- Controller vs. service test coverage

Let me know if any of these need more work, or if there's anything else you'd like me to address.

Thanks!

mynameisbogdan:
Please give it a few more days, I want to test it but I don't have a way to test without hardlinks without doing a lot of work. 

I really want to check if the UI is updated properly with the new UI. I know for a fact that the old v3 ignores null values from being changed in the react state since null are ignored from being json serialized.

markus101:
> Checking in on this PR since it's been about three weeks without follow-up since your review in July

I reviewed on August 29th, then again on September 5th (and mentioned my previous review hadn't been addressed), August 29th is almost 3 weeks, but neither of those is July.

Given the time frame this has been open and the back and forth this needs additional time for testing and is lower priority than other open PRs and work to get v5 to a beta release.

PR #8812, opened 2026-07-24
Fixed: Pagination for Trakt import lists

mynameisbogdan:
#### Description

Adding pagination support to Trakt lists, as now there's a hard limit of 250 for the query parameter `limit`. 

To maintain support for fetching exactly the number of items as per limit set in settings, an items.Take(limit) is needed in `CleanupListItems`. Needed as Trakt's pagination does not handle offset, just `page` + `limit`.

- Related https://github.com/Radarr/Radarr/issues/11538

### DIFF
--- a/src/NzbDrone.Core/ImportLists/Trakt/User/TraktUserRequestGenerator.cs
+++ b/src/NzbDrone.Core/ImportLists/Trakt/User/TraktUserRequestGenerator.cs
@@ -1,4 +1,6 @@
+using System;
 using System.Collections.Generic;
+using System.Globalization;
 using NzbDrone.Common.Extensions;
 using NzbDrone.Common.Http;
 
@@ -8,11 +10,15 @@ namespace NzbDrone.Core.ImportLists.Trakt.User
     {
         private readonly TraktUserSettings _settings;
         private readonly string _clientId;
+        private readonly int _pageSize;
+        private readonly int _maxNumResults;
 
-        public TraktUserRequestGenerator(TraktUserSettings settings, string clientId)
+        public TraktUserRequestGenerator(TraktUserSettings settings, string clientId, int pageSize, int maxNumResults)
         {
             _settings = settings;
             _clientId = clientId;
+            _pageSize = pageSize;
+            _maxNumResults = maxNumResults;
         }
 
         public virtual ImportListPageableRequestChain GetListItems()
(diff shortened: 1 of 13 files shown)


kushiemoon-dev:
This is the same bug we hit on Radarr (Radarr/Radarr#11563). I closed that one in favor of doing it here first since `TraktQueryHelper` is shared across all three generators and a page-aware fix touches all of them at once. Good to see it landed with the pageSize threaded through the constructors instead of just patching the shared helper in place.

Two small things, neither blocking:

- Each page always requests `limit=250` even when `Settings.Limit` is smaller (e.g. Limit=10 still pulls a full page before `Take()` trims it down). Not wrong, just extra Trakt API traffic for low-limit configs.
- The page cap of 4 is hardcoded rather than derived from `MaxNumResultsPerQuery / PageSize` (1000/250). Works today, but it'll silently drift if either constant changes later.

CI's green and this has been sitting for three weeks with no review. Happy to test it against a real watchlist/list/popular setup if that'd help get it over the line.

mynameisbogdan:
It's still a draft, trakt made changes who paywalled their API so this is low priority for now.

kushiemoon-dev:
Oh okay all is good then ^^ i didnt saw the stuff about the paywall what that mean ??

mynameisbogdan:
There's no announcement on the regard, they just did it. 

https://redirect.github.com/trakt/trakt-api/discussions/898
https://redirect.github.com/trakt/trakt-api/issues/906

You can have only one connecting app to your account and you can't create API apps with a standard account. You need VIP to have more than one connecting app and to create API apps. 

And they removed existing API apps for those without a VIP, but even some users with VIP reported they had missing apps too.

kushiemoon-dev:
Oh shit... thats fucking sucks why ???

kushiemoon-dev:
Why they block it either without annoucement ??

mynameisbogdan:
Because $$, cheapest would be at least $74/year I think. And because they don't care about the users.

mynameisbogdan:
> Two small things, neither blocking:
> 
>     * Each page always requests `limit=250` even when `Settings.Limit` is smaller (e.g. Limit=10 still pulls a full page before `Take()` trims it down). Not wrong, just extra Trakt API traffic for low-limit configs.
> 
>     * The page cap of 4 is hardcoded rather than derived from `MaxNumResultsPerQuery / PageSize` (1000/250). Works today, but it'll silently drift if either constant changes later.
> 
> 
> CI's green and this has been sitting for three weeks with no review. Happy to test it against a real watchlist/list/popular setup if that'd help get it over the line.

2 was indeed a must have, but 1 eh considering the API situation and how it applies to only if limit < 250, I wasted a few minutes on calling the remainder if only one pages is being fetched.

src/NzbDrone.Core/ImportLists/Trakt/User/TraktUserRequestGenerator.cs
```diff
@@ -41,36 +57,50 @@ private IEnumerable<ImportListRequest> GetSeriesRequest()
...
+            }
 
-            request.HttpRequest.Headers.Add("trakt-api-version", "2");
-            request.HttpRequest.Headers.Add("trakt-api-key", _clientId);
+            var limit = Math.Clamp(_settings.Limit, 0, _maxNumResults);
+            var maxPages = (int)Math.Ceiling(decimal.Divide(limit, _pageSize));
+            var remainderLimit = limit % _pageSize;
 
-            if (_settings.AccessToken.IsNotNullOrWhiteSpace())
+            for (var page = 1; page <= maxPages; page++)
```
markus101: Since we're doing a lot of the same things for paging against different URLs I think extracting things out to a base class would be beneficial.

PR #8819, opened 2026-07-26
Fixed: NullReferenceException in decision specs when episode file is missing

fatbob01:
### Description

Four decision engine specifications dereference `EpisodeFile.Value` without a null check, while three others guard the identical expression.

Unguarded:
- `DelaySpecification.cs:54`
- `ProperSpecification.cs:42`
- `RepackSpecification.cs:43`
- `DeletedEpisodeFileSpecification.cs:44`

Already guarded:
- `AnimeVersionUpgradeSpecification.cs:47`
- `UpgradeAllowedSpecification.cs:33`
- `UpgradeDiskSpecification.cs:45`

All seven iterate the same `subject.Episodes.Where(c => c.EpisodeFileId != 0).Select(c => c.EpisodeFile.Value)` expression, so the three that already guard establish that a null is reachable there.

An episode can transiently have a non-zero `EpisodeFileId` while its `EpisodeFile` no longer resolves: `EpisodeService.Handle(EpisodeFileDeletedEvent)` clears `EpisodeFileId` via `EpisodeRepository.ClearFileId()` from an event handler rather than in the same unit of work as the delete. Anything that deletes a file and then triggers a search through the API can observe that window.

When it happens the spec throws, `DownloadDecisionMaker.EvaluateSpec` catches per-spec and rejects the release, so **every** candidate release is rejected and the search reports `0 reports downloaded` with no indication of the cause:

```
|Error|DownloadDecisionMaker|Couldn't evaluate decision on <release title redacted>
[v4.1.1.806] System.NullReferenceException: Object reference not set to an instance of an object.
   at NzbDrone.Core.DecisionEngine.Specifications.RssSync.DelaySpecification.IsSatisfiedBy(RemoteEpisode subject, SearchCriteriaBase searchCriteria) in ./Sonarr.Core/DecisionEngine/Specifications/RssSync/DelaySpecification.cs:line 54
   at NzbDrone.Core.DecisionEngine.DownloadDecisionMaker.EvaluateSpec(...) in ./Sonarr.Core/DecisionEngine/DownloadDecisionMaker.cs:line 236
|Info|EpisodeSearchService|Episode search completed. 0 reports downloaded.
```

That trace is from a fork build, but `DelaySpecification.cs` there is byte-identical to `develop`, `main` and `v4.0.19.2979`, and the fork's changes are unrelated to this code path.

Two behaviours follow from the early returns and make this easy to misread as a delay profile problem rather than an NRE: manual searches are unaffected, because `DelaySpecification` returns early on `UserInvokedSearch`; and setting the delay profile to 0 also avoids it, via the `delay == 0` early return.

This applies the existing guard to the four remaining specs, using `EpisodeFile?.Value` so a null `LazyLoaded` wrapper is handled as well as a null value.

For reference, Radarr's equivalent `DelaySpecification` already guards this with `subject.Movie.MovieFileId != 0 && file != null`.


mynameisbogdan:
Can you please fill a bug report, instead of pushing your whole fork, so we can take a look at the issue? 

And please don't use a LLM to fill the bug report, and make sure the issue isn't related to your many changes by using a clean build and provide trace logs as found in the logs. 

I noticed a pattern in LLMs in using just 2 lines from the stacktrace and the lines + line numbers don't even match, which makes me believe they resort to hallucinating them.

Thank you.

src/NzbDrone.Core/DecisionEngine/Specifications/RepackSpecification.cs
```diff
@@ -40,8 +40,13 @@ public DownloadSpecDecision IsSatisfiedBy(RemoteEpisode subject, SearchCriteriaB
                 return DownloadSpecDecision.Accept();
             }
 
-            foreach (var file in subject.Episodes.Where(c => c.EpisodeFileId != 0).Select(c => c.EpisodeFile.Value))
+            foreach (var file in subject.Episodes.Where(c => c.EpisodeFileId != 0).Select(c => c.EpisodeFile?.Value))
```
mynameisbogdan, line 43: > An episode can transiently have a non-zero EpisodeFileId while its
EpisodeFile no longer resolves, since EpisodeService clears EpisodeFileId
from an EpisodeFileDeletedEvent handler rather than in the same unit of
work as the delete. Any automation that deletes a file and immediately
triggers a search can observe that window.

How fast are you even triggering this search anyway? I think you're abusing that search for the database not being able to keep with the changes if you're already triggering searches, which I may add, they have a delay between triggering the search, getting the responses and parsing them. 

Never mind there's at least 2 seconds delay between requests to the same indexer, it's very unlikely you searched a single episode with imdbid and found a result from just one request.

There's a delay between episode removal, search being triggered and a decision specifications being applied, basically I'm inclined to believe this description was generated with claude just to "sell" it.

PR #8823, opened 2026-07-26
Fixed: Make manual import cells keyboard accessible

coopa11y:
#### Description

Editable cells in Manual Import could only be activated with a pointer because the action was attached directly to the table cell.

This keeps the existing table layout and mouse behavior while placing a native button inside each editable cell. Keyboard users can now open the Series, Season, Episode, Release Group, Quality, Language, Release Type, and Indexer Flags controls with Enter or Space. The existing translated action text provides the accessible name, and disabled cells remain unavailable.

#### Submission Checklist


stevietv:
hey @coopa11y - have you also had a look at the v5-redesign branch? I'd hate for all your hard work to be reversed when the redesign lands!

mynameisbogdan:
> have you also had a look at the v5-redesign branch? I'd hate for all your hard work to be reversed when the redesign lands!

I think it's better v5-develop and rebase them into v5-design till it's finished since we don't know when it might be ready IMO.

coopa11y:
@stevietv @mynameisbogdan Thanks for flagging this. Would it be helpful if I started auditing `v5-redesign` now and porting over any accessibility fixes that don’t carry across cleanly? I’d be happy to begin checking the redesigned shared components and screens, but I don’t want to duplicate work or create conflicts while the branch is still evolving.

mynameisbogdan:
@markus101 WDYT? Since you'll the one dealing with the rebases...

markus101:
It's something I've been looking at as I'm reviewing both, so far the impact has been pretty minimal. I don't think it's necessary to take a look at the `v5-redesign` yet for specific issues.

If anything comes up that'll be a major conflict I'll mention it and may not land the related PR. Right now I think the only piece that's changing enough that it wouldn't make sense to look at is the Import series flow, since it's format is changing pretty heavily.

frontend/src/Components/Table/Cells/TableRowCellButton.tsx
```diff
@@ -9,10 +10,19 @@ interface TableRowCellButtonProps extends LinkProps {
 }
 
 function TableRowCellButton(props: TableRowCellButtonProps) {
-  const { className = styles.cell, ...otherProps } = props;
+  const { className, children, title, ...otherProps } = props;
```
markus101: The changes to `className` here make the composition a little weird, instead of using `classNames` consumers should correctly compose `cell` from `~Components/Table/Cells/TableRowCellButton.css` to match the pattern already established.

Looks like this would just be this patch + reverting the `classNames` addition here:

```
diff --git forkSrcPrefix/frontend/src/InteractiveImport/Interactive/InteractiveImportRow.css forkDstPrefix/frontend/src/InteractiveImport/Interactive/InteractiveImportRow.css
index d3cfb118ada2d8b3fd86018d8b80226a0cbc00bf..f39eee7fe0cb0d5fc2f84387098be0bb133f96c4 100644
--- forkSrcPrefix/frontend/src/InteractiveImport/Interactive/InteractiveImportRow.css
+++ forkDstPrefix/frontend/src/InteractiveImport/Interactive/InteractiveImportRow.css
@@ -6,7 +6,7 @@
 
 .quality,
 .languages {
-  composes: cell from '~Components/Table/Cells/TableRowCell.css';
+  composes: cell from '~Components/Table/Cells/TableRowCellButton.css';
 
   text-align: center;
 }
```

markus101 (commented):
Just one thing, otherwise looks good.

frontend/src/Components/Table/Cells/TableRowCellButton.tsx
coopa11y: Thanks, fixed. I restored the existing default class behavior in `TableRowCellButton` and updated the Manual Import quality and language cells to compose from `TableRowCellButton.css` directly. ESLint, Stylelint, and the full frontend build pass.

PR #8825, opened 2026-07-26
Redesign Calendar views for screen-reader navigation

coopa11y:
#### Description

This PR makes the Calendar usable as a structured, navigable view with screen readers while preserving the existing visual event presentation and Calendar functionality.

The Calendar was primarily rendered as nested visual containers. Dates, weekdays, event groups, statuses, and icon-only actions did not consistently expose their relationships or purpose to assistive technology. Several interactive event underlays had no accessible name, status information depended on color or pointer tooltips, and the grid could not be navigated using standard screen-reader table commands.

This is intentionally a complete Calendar accessibility pass rather than a series of partially functional intermediate states. I am opening the PR early to get feedback on the overall model, its relationship to `v5-redesign`, and whether any portions should be adjusted or separated before review.

##### Calendar structure and navigation

- Replaces the visual-only Calendar grid structure with a native HTML table.
- Exposes weekday column headers and full dates through `th`, `time`, and date headings.
- Preserves Month, Week, Forecast, and Day layouts without introducing a custom ARIA grid or application-specific keyboard model.
- Allows screen readers to use their standard table, row, column, cell, and heading navigation commands.
- Marks the current date and provides useful empty-day and empty-range text.
- Removes the obsolete separate weekday-header component after moving headers into the table.

##### View and range controls

- Exposes Month, Week, Forecast, Day, and Agenda as named toggle buttons using `aria-pressed`.
- Orders the Calendar controls consistently as:
  1. view selection;
  2. Previous, Next, and Today range navigation;
  3. the current range heading;
  4. the resulting Calendar table or Agenda.
- Gives Previous and Next names that include the active view.
- Uses one level-two range heading at every viewport size.
- Announces loading and range-heading changes without moving focus.
- Corrects Calendar update timers and refreshes data without unexpectedly moving the selected range back to today.

##### Events, groups, and Agenda

- Gives every episode action a complete accessible name containing the series, episode number, title, air date and time, and relevant status information.
- Includes downloaded, downloading, missing, unaired, unmonitored, premiere, finale, special, numbering, and quality-cutoff information without relying on color alone.
- Makes collapsed episode groups keyboard operable and exposes their expanded state, controlled content, episode count, season, and status.
- Ensures `aria-controls` references an existing element in both collapsed and expanded states.
- Rebuilds Agenda as date-labelled sections containing semantic event lists.
- Adds meaningful empty Agenda feedback.

##### Legend, options, and iCal

- Exposes the Calendar legend as a labelled list.
- Gives every legend entry an accessible name containing both the status and its explanation.
- Hides decorative legend icons from assistive technology.
- Associates Calendar option labels with their form controls.
- Renames the toolbar action from ???iCal Link??? to ???iCal Feed??? because it opens a modal rather than navigating as a link.
- Associates every iCal modal label with its input.
- Gives the Copy and Subscribe actions explicit accessible names.
- Announces clipboard success and failure.

##### Search for Missing

- Implements the previously empty Search for Missing action for eligible episodes in the displayed range.
- Searches only episodes that have aired, have no episode file, and are not already queued.
- Includes the complete first and final displayed days when determining eligibility.
- Keeps the action disabled when the displayed range has no eligible episodes.

##### Shared accessibility support

- Adds a reusable visually hidden text component for accessible descriptions and live statuses.
- Connects shared checkbox and select inputs to external form labels through stable IDs.
- Adds accessible clipboard feedback for other existing uses of the shared Clipboard button.
- Allows queue progress popovers to expose a useful accessible label.

#### User impact

Screen-reader users can now choose a Calendar view, move between ranges, confirm the resulting date range, navigate the Calendar as a table, move by date headings, inspect named episode actions, expand episode groups, understand event statuses, and use the Agenda, legend, options, and iCal controls without depending on visual position, color, or pointer hover.

The native table approach intentionally relies on established browser and assistive-technology behavior instead of adding a separate keyboard interaction model that would need to be learned and maintained.

#### Testing performed

- ESLint passes for the changed TypeScript and TSX files.
- Stylelint passes for the changed Calendar styles.
- `git diff --check` passes.
- The production Webpack frontend build completes successfully.
- The production UI was layered onto the current Sonarr `nvda-a11y` Docker image and run against an existing Sonarr database on Unraid.
- Container startup, HTTP response, Tailscale proxying, and deployed UI checksums were verified after each test build.
- Iterative manual testing was performed with NVDA against the live Calendar, including control order, range headings, table navigation, event names, and legend output.

The latest iCal announcements, inclusive Search for Missing boundaries, and persistent episode-group control target are included in the production-tested build and should receive another focused NVDA pass before this draft is marked ready.

To be added before marking the PR ready. Most changes are semantic, but the view and range controls now have a visible stacked order that should be shown at desktop and narrow widths.

#### Questions for review

I would particularly appreciate early feedback on:

- whether this should continue to target `v5-develop` or be ported onto `v5-redesign`, which already contains overlapping Calendar work;
- whether maintainers prefer the native table model or a different semantic model for the visual Calendar;
- whether the view, range-navigation, heading, and table order matches the intended redesign direction;
- whether any shared-component or Search for Missing changes should be separated despite being exercised directly by the Calendar;
- any additional screen-reader or keyboard workflows maintainers would like tested before the draft is marked ready.

I am happy to revise the structure, split narrowly scoped portions, or adapt the implementation to the redesign branch based on maintainer guidance.

#### Submission Checklist


markus101:
- This will definitely conflict with the redesign work, though this should not be applied to that branch, it should be tackled after that work is merged
- Does this need to be implemented as a table? Even though this is somewhat represented in a tabular format I'm not sure it's best implemented as a table.
- I don't think the changes to the header align with the current design or the redesign, it now takes up more space (3 separate lines instead of one single line) and visually doesn't look as good
- I think it'd be best to keep the changes unrelated to the accessibility of the calendar separate

coopa11y:
@markus101 Thanks for the feedback. Waiting until the redesign work is merged makes sense, rather than trying to adapt this PR to a moving target.

I used a native table because it gives screen-reader users established row, column, and cell navigation without requiring a custom ARIA grid and keyboard interaction model. That said, I?m not tied to the table implementation. Is there another semantic model you think would better fit the Calendar while still allowing screen-reader users to navigate dates and understand the relationship between weekdays, dates, and events?

I can also restore the header controls to a single-line visual layout while arranging their DOM order so the controls and current date range are presented logically to assistive technology.

Finally, I can narrow this PR to Calendar-specific accessibility changes and separate the shared-component and Search for Missing changes where they are not strictly required by the Calendar.

Would you prefer that I leave this PR as a draft until the redesign is merged and then rework it against the resulting Calendar implementation?

markus101:
Thanks for the explanation! I appreciate both the consideration for implementing it that way as well as the information. I don't think it's crucial to go a bespoke solution just to stick with the existing structure.

Everything else sounds good as well, let's leave this in draft and hopefully it won't be much longer before we can get that work landed to rebase this rework on.

PR #8835, opened 2026-07-29
fix: Properly resolve file renaming conflicts

### LINKED ISSUE #5610: If you reassign episodes within a season of a show, then rename, the files will delete

### Current Behavior

If you reassign episodes within the same season (due to the source import being incorrectly ordered), then rename, sonarr will end up deleting the files.

### Expected Behavior

Sonarr doesn't delete the files.

### Steps To Reproduce

1. Open a show in sonarr with 3 or more episodes in one season.
2. On that season, click Manage Files.
3. Reassign the episodes, e.g. E01 -> E02, E02 -> E03, E03 -> E01
4. Click Import
5. Click Rename files on that season
6. Refresh the show. at least 2 of your 3 files are now gone

### Environment

```markdown
- OS: ubuntu 18.04
- Sonarr: 4.0.0.470
- Docker Install: linuxserver.io 
- Using Reverse Proxy: yes nginx + ssl
- Browser: Chrome 64 bit
```

Develop

### Trace Logs?

[sonarr.trace (2).txt](https://github.com/Sonarr/Sonarr/files/11366293/sonarr.trace.2.txt)

### Anything else?

In the attached trace logs, I performed the above steps with the show "Run On." I suspect whats happening is when it renames the first episode, it overwrites the second before the second can be renamed, causing a cascading effect of having every other file overwritten before rename.

FizzyGalacticus:
~I apologize if new PRs are supposed to target the v5-develop branch, I didn't see anywhere that made it clear which branch to target, however I do believe this fix is worthy of making it into v4/stable. Please let me know if I should instead target the v5 branch.~

#### Description

When trying to move episode files that result in a loop or conflict, the moves currently fails.

Example: moving files `A.mkv` -> `B.mkv`, `B.mkv` -> `A.mkv` results in the move not taking place.

This is the same for longer chains or cycles: `A.mkv` -> `B.mkv`, `B.mkv` -> `C.mkv`, `C.mkv` -> `D.mkv`, ..., `Z.mkv` -> `A.mkv`

These fixes pre-build a list of moves that are immediately available without conflicts as well as a list of "cycles", where moves are dependent on moving a file to a temporary location in order to properly resolve all moves.

Renames get marked as "blocked" if:

* There exists a file at the target location already that is not part of this rename process, or
* Two or more files in this rename process map to the same target

#### Submission Checklist

- Closes #5610

### DIFF
--- a/src/NzbDrone.Core.Test/MediaFiles/RenameEpisodeFileServiceFixture.cs
+++ b/src/NzbDrone.Core.Test/MediaFiles/RenameEpisodeFileServiceFixture.cs
@@ -1,14 +1,19 @@
-﻿using System.Collections.Generic;
+using System.Collections.Generic;
+using System.IO;
 using System.Linq;
 using FizzWare.NBuilder;
 using Moq;
 using NUnit.Framework;
+using NzbDrone.Common.Disk;
+using NzbDrone.Core.CustomFormats;
 using NzbDrone.Core.MediaFiles;
 using NzbDrone.Core.MediaFiles.Commands;
 using NzbDrone.Core.MediaFiles.Events;
 using NzbDrone.Core.Messaging.Events;
+using NzbDrone.Core.Organizer;
 using NzbDrone.Core.Test.Framework;
 using NzbDrone.Core.Tv;
+using NzbDrone.Test.Common;
 
 namespace NzbDrone.Core.Test.MediaFiles
 {
@@ -21,6 +26,7 @@ namespace NzbDrone.Core.Test.MediaFiles
         public void Setup()
         {
             _series = Builder<Series>.CreateNew()
+                                     .With(s => s.Path = @"C:\Test\TV\Series Title".AsOsAgnostic())
                                      .Build();
 
             _episodeFiles = Builder<EpisodeFile>.CreateListOfSize(2)
@@ -33,6 +39,16 @@ namespace NzbDrone.Core.Test.MediaFiles
             Mocker.GetMock<ISeriesService>()
                   .Setup(s => s.GetSeries(_series.Id))
                   .Returns(_series);
+
+            Mocker.GetMock<IEpisodeService>()
+                  .Setup(s => s.GetEpisodesByFileId(It.IsAny<int>()))
+                  .Returns(new List<Episode>());
+
+            // By default every file renames to a unique, uncontested path so tests that don't care about
+            // naming conflicts don't need to stub the builder themselves.
+            Mocker.GetMock<IBuildFileNames>()
+                  .Setup(f => f.BuildFilePath(It.IsAny<List<Episode>>(), It.IsAny<Series>(), It.IsAny<EpisodeFile>(), It.IsAny<string>(), It.IsAny<NamingConfig>(), It.IsAny<List<CustomFormat>>()))
+                  .Returns<List<Episode>, Series, EpisodeFile, string, NamingConfig, List<CustomFormat>>((episodes, series, file, ext, namingConfig, customFormats) => Path.Combine(series.Path, "Renamed-" + file.RelativePath));
         }
 
         private void GivenNoEpisodeFiles()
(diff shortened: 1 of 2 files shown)


mynameisbogdan:
I have a feeling this conflicts with #8799.

markus101:
It might, but they shouldn't have major conflicts, more that they'd compliment each other and have some potential minor overlap.

src/NzbDrone.Core/MediaFiles/RenameEpisodeFileService.cs
```diff
@@ -130,52 +153,278 @@ private IEnumerable<RenameEpisodeFilePreview> GetPreviews(Series series, List<Ep
...
+        {
+            var pending = new List<EpisodeFileRename>();
 
             foreach (var episodeFile in episodeFiles)
             {
-                var previousRelativePath = episodeFile.RelativePath;
-                var previousPath = Path.Combine(series.Path, episodeFile.RelativePath);
+                var currentPath = Path.Combine(series.Path, episodeFile.RelativePath);
+                var episodes = _episodeService.GetEpisodesByFileId(episodeFile.Id);
+                var targetPath = _filenameBuilder.BuildFilePath(episodes, series, episodeFile, Path.GetExtension(episodeFile.RelativePath));
```
markus101, line 194: `BuildFilePath` calls `episodes.First()` (`FileNameBuilder.cs:234`), which throws `InvalidOperationException` when a file has no linked episodes. This will now result in the whole operation failing instead of just the affected file.

This could happen when a file is removed between fetching previews and starting the rename.

markus101, line 194: Every destination is computed twice: `GetEpisodesByFileId` + `BuildFilePath` here, then the identical pair again inside `MoveEpisodeFile`. On a full-library rename that doubles the episode queries and the naming/custom-format work.

It also means the planned target and the actual move target are computed independently and could drift. Can we pass the path through to the mover instead?

src/NzbDrone.Core/MediaFiles/RenameEpisodeFileService.cs
```diff
@@ -130,52 +153,278 @@ private IEnumerable<RenameEpisodeFilePreview> GetPreviews(Series series, List<Ep
...
+                    current = next;
                 }
-                catch (Exception ex)
+
+                if (!settled)
                 {
-                    _logger.Error(ex, "Failed to rename file {0}", previousPath);
+                    // Unreachable unless the traversal above no longer satisfies the bound described
+                    // above - fail loudly rather than silently dropping files from the plan.
+                    throw new InvalidOperationException("Unable to resolve episode rename dependency chain");
```
markus101, line 312: Is it unreachable or something we actually need to throw?

src/NzbDrone.Core/MediaFiles/RenameEpisodeFileService.cs
```diff
@@ -130,52 +153,278 @@ private IEnumerable<RenameEpisodeFilePreview> GetPreviews(Series series, List<Ep
...
+            try
+            {
+                _diskProvider.MoveFile(displaced.CurrentPath, tempPath);
+            }
+            catch (Exception ex)
+            {
+                _logger.Error(ex, "Failed to rename file {0}", displaced.CurrentPath);
+                return;
+            }
+
```
markus101, line 405: This can permanently orphan a file. The temporary state needs to survive a crash, or the original name needs to be restored on any failure before returning.

src/NzbDrone.Core/MediaFiles/RenameEpisodeFileService.cs
```diff
@@ -130,52 +153,278 @@ private IEnumerable<RenameEpisodeFilePreview> GetPreviews(Series series, List<Ep
...
 
-                    _mediaFileService.Update(episodeFile);
+        // Follows each file's target to whoever currently occupies it (if anyone) to determine, before any
+        // file is touched, whether the whole chain of dependent moves can ultimately complete. A file whose
+        // target is occupied by another file in this batch can only move once that file has moved; a chain
+        // that bottoms out at a free path can all be moved (last-to-first); a chain that bottoms out at a
+        // path occupied by a file outside this batch can never complete and is left untouched entirely; a
+        // chain that loops back on itself (e.g. episodes swapping numbers) is a closed rotation that can
+        // always be completed using one temporary displacement.
+        private RenamePlan BuildRenamePlan(List<EpisodeFileRename> pending)
```
markus101, line 221: This seems overly complex for what we're doing. Makes it hard to follow what it's supposed to do and very that it's actually doing that correctly.

src/NzbDrone.Core.Test/MediaFiles/RenameEpisodeFileServiceFixture.cs
```diff
@@ -71,11 +96,14 @@ public void should_not_publish_event_if_no_files_are_renamed()
         {
             GivenEpisodeFiles();
 
-            Mocker.GetMock<IMoveEpisodeFiles>()
-                  .Setup(s => s.MoveEpisodeFile(It.IsAny<EpisodeFile>(), It.IsAny<Series>()))
-                  .Throws(new SameFilenameException("Same file name", "Filename"));
+            // Both files already sit at their correctly-computed path - nothing to do.
```
markus101, line 99: We're now missing a test that handles the `SameFilenameException` flow it looks like.

markus101 (changes requested):
This is really hard to follow and is full of low quality AI-like comments that explain a lot of what is happening, which the code is self already covers. Moving to draft for the time being.

PR #8836, opened 2026-07-30
New: Match anime season packs named only by their season title

JonzaUK:
#### Description

Some anime releases for season packs don't include any season data such as S02 etc so they are discarded, sonarr already knows the scene mappings which include season aliases. This adds a separate method in parser outside of the regex that recovers the series title and then is able to match the season alias to the season using parsingservicing

#### Submission Checklist

### DIFF
--- a/src/NzbDrone.Core.Test/DecisionEngineTests/DownloadDecisionMakerFixture.cs
+++ b/src/NzbDrone.Core.Test/DecisionEngineTests/DownloadDecisionMakerFixture.cs
@@ -150,7 +150,10 @@ namespace NzbDrone.Core.Test.DecisionEngineTests
         public void should_not_attempt_to_map_episode_if_not_parsable()
         {
             GivenSpecifications(_pass1, _pass2, _pass3);
-            _reports[0].Title = "Not parsable";
+
+            // Stripping the metadata leaves brackets rather than a title, so neither parser
+            // recovers anything to map.
+            _reports[0].Title = "[Not][parsable][1080p][x265(10bit)]";
 
             Subject.GetRssDecision(_reports).ToList();
 
@@ -159,6 +162,19 @@ namespace NzbDrone.Core.Test.DecisionEngineTests
             _pass1.Verify(c => c.IsSatisfiedBy(It.IsAny<RemoteEpisode>(), It.IsAny<ReleaseDecisionInformation>()), Times.Never());
             _pass2.Verify(c => c.IsSatisfiedBy(It.IsAny<RemoteEpisode>(), It.IsAny<ReleaseDecisionInformation>()), Times.Never());
             _pass3.Verify(c => c.IsSatisfiedBy(It.IsAny<RemoteEpisode>(), It.IsAny<ReleaseDecisionInformation>()), Times.Never());
+        }
+
+        [Test]
+        public void should_map_a_title_the_regexes_cannot_parse()
+        {
+            GivenSpecifications(_pass1, _pass2, _pass3);
+            _reports[0].Title = "Not parsable";
+
+            Subject.GetRssDecision(_reports).ToList();
+
+            // A release carrying nothing but a title is still a title, so it is looked up
+            // rather than dropped. Without a matching series it is rejected as unknown.
+            Mocker.GetMock<IParsingService>().Verify(c => c.Map(It.IsAny<ParsedEpisodeInfo>(), It.IsAny<int>(), It.IsAny<int>(), It.IsAny<string>(), It.IsAny<SearchCriteriaBase>()), Times.Once());
         }
 
         [Test]
--- a/src/NzbDrone.Core.Test/DecisionEngineTests/DownloadDecisionMakerFixture.cs
+++ b/src/NzbDrone.Core.Test/DecisionEngineTests/DownloadDecisionMakerFixture.cs
@@ -172,8 +172,6 @@ namespace NzbDrone.Core.Test.DecisionEngineTests
 
             Subject.GetRssDecision(_reports).ToList();
 
-            // A release carrying nothing but a title is still a title, so it is looked up
-            // rather than dropped. Without a matching series it is rejected as unknown.
             Mocker.GetMock<IParsingService>().Verify(c => c.Map(It.IsAny<ParsedEpisodeInfo>(), It.IsAny<int>(), It.IsAny<int>(), It.IsAny<string>(), It.IsAny<SearchCriteriaBase>()), Times.Once());
         }
 
--- a/src/NzbDrone.Core.Test/ParserTests/SeasonTitleParserFixture.cs
+++ b/src/NzbDrone.Core.Test/ParserTests/SeasonTitleParserFixture.cs
@@ -20,17 +20,11 @@ namespace NzbDrone.Core.Test.ParserTests
             result.SeriesTitle.Should().Be(expectedTitle);
             result.FullSeason.Should().BeTrue();
             result.IsSeasonTitleOnly.Should().BeTrue();
-
-            // Which season the title refers to is held in the scene mappings, so it is left
-            // unset here and resolved by ParsingService. Not 0, that is the special season.
             result.SeasonNumber.Should().Be(-1);
             result.EpisodeNumbers.Should().BeEmpty();
             result.AbsoluteEpisodeNumbers.Should().BeEmpty();
         }
 
-        // DownloadDecisionComparer reads ParsedEpisodeInfo.Quality while sorting, so leaving it
-        // unset fails the whole search with "failed to compare two elements in the array"
-        // rather than skipping one release.
         [TestCase("[Group] Monkey Island Curse (BD 1280x720 x264 AAC)")]
         [TestCase("[Group] Monkey Island: Curse [Dual Audio 10bit 1080p][HEVC-x265]")]
         public void should_populate_the_fields_the_decision_engine_reads(string releaseTitle)
@@ -42,9 +36,6 @@ namespace NzbDrone.Core.Test.ParserTests
             result.Languages.Should().NotBeNullOrEmpty();
         }
 
-        // Some groups bracket every part of the name including the title, and a bracketed group
-        // can contain another. Both leave metadata behind instead of a title, which is worse
-        // than returning nothing because it gets used as a lookup key.
         [TestCase("[Group][Monkey Island: Curse][BDRip][1080P_x265(10bit)-FLAC][ALL]")]
         [TestCase("[Group][Monkey Island Curse][1080p][x264]")]
         public void should_not_return_a_title_when_stripping_leaves_metadata(string releaseTitle)
@@ -52,8 +43,6 @@ namespace NzbDrone.Core.Test.ParserTests
             Parser.Parser.ParseSeasonTitle(releaseTitle).Should().BeNull();
         }
 
-        // Groups mark extras inside brackets, which is what stripping metadata removes, so an
-        // opening or an OVA would otherwise look exactly like a season pack.
         [TestCase("[Group] Monkey Island Curse (OVA)")]
         [TestCase("[Group] Monkey Island Curse [OVA]")]
         [TestCase("[Group] Monkey Island Curse (Special)")]
@@ -64,7 +53,6 @@ namespace NzbDrone.Core.Test.ParserTests
             Parser.Parser.ParseSeasonTitle(releaseTitle).Should().BeNull();
         }
 
-        // ParsingService resolves the season off the flag, so only this parser may set it.
         [TestCase("Monkey Island S02E05 1080p WEB-DL")]
         [TestCase("[Group] Monkey Island - 05 (1080p)")]
         [TestCase("Monkey Island S02 1080p WEB-DL")]
@@ -73,9 +61,6 @@ namespace NzbDrone.Core.Test.ParserTests
             Parser.Parser.ParseTitle(releaseTitle).IsSeasonTitleOnly.Should().BeFalse();
         }
 
-        // The helper is not wired into ParseTitle, so anything ParseTitle rejects today it
-        // still rejects. Otherwise nothing would ever be unparsable and crap detection, daily
-        // date validation and the import path all lose their only signal.
         [TestCase("THIS SHOULD NEVER PARSE")]
         [TestCase("Vrq6e1Aba3U amCjuEgV5R2QvdsLEGYF3YQAQkw8")]
         [TestCase("[Group] Monkey Island Curse (BS11 1280x720 x264 AAC)")]
--- a/src/NzbDrone.Core.Test/ParserTests/SeasonTitleParserFixture.cs
+++ b/src/NzbDrone.Core.Test/ParserTests/SeasonTitleParserFixture.cs
@@ -19,7 +19,7 @@ namespace NzbDrone.Core.Test.ParserTests
             result.Should().NotBeNull();
             result.SeriesTitle.Should().Be(expectedTitle);
             result.FullSeason.Should().BeTrue();
-            result.IsSeasonTitleOnly.Should().BeTrue();
+            result.IsSeasonTitle.Should().BeTrue();
             result.SeasonNumber.Should().Be(-1);
             result.EpisodeNumbers.Should().BeEmpty();
             result.AbsoluteEpisodeNumbers.Should().BeEmpty();
@@ -38,6 +38,7 @@ namespace NzbDrone.Core.Test.ParserTests
 
         [TestCase("[Group][Monkey Island: Curse][BDRip][1080P_x265(10bit)-FLAC][ALL]")]
         [TestCase("[Group][Monkey Island Curse][1080p][x264]")]
+        [TestCase("[Group] Monkey Island Curse () (1080p)")]
         public void should_not_return_a_title_when_stripping_leaves_metadata(string releaseTitle)
         {
             Parser.Parser.ParseSeasonTitle(releaseTitle).Should().BeNull();
@@ -58,7 +59,7 @@ namespace NzbDrone.Core.Test.ParserTests
         [TestCase("Monkey Island S02 1080p WEB-DL")]
         public void should_not_flag_a_release_the_regexes_can_parse(string releaseTitle)
         {
-            Parser.Parser.ParseTitle(releaseTitle).IsSeasonTitleOnly.Should().BeFalse();
+            Parser.Parser.ParseTitle(releaseTitle).IsSeasonTitle.Should().BeFalse();
         }
 
         [TestCase("THIS SHOULD NEVER PARSE")]
--- a/src/NzbDrone.Core.Test/ParserTests/SeasonTitleParserFixture.cs
+++ b/src/NzbDrone.Core.Test/ParserTests/SeasonTitleParserFixture.cs
@@ -1,0 +1,77 @@
+using FluentAssertions;
+using NUnit.Framework;
+using NzbDrone.Core.Test.Framework;
+
+namespace NzbDrone.Core.Test.ParserTests
+{
+    [TestFixture]
+    public class SeasonTitleParserFixture : CoreTest
+    {
+        [TestCase("[Ohys-Raws] Dr. Stone Stone Wars (BS11 1280x720 x264 AAC)", "Dr. Stone Stone Wars")]
+        [TestCase("[S3rNx] Dr. Stone Stone Wars (BDRip 1920x1080 x264 FLAC)", "Dr. Stone Stone Wars")]
+        [TestCase("[DmonHiro] Dr.Stone - Stone Wars (BD, 720p)", "Dr.Stone - Stone Wars")]
+        [TestCase("Dr. Stone Stone Wars [Bluray.1080p.HEVC.AAC.ITA.FLAC.JAP.Sub.Ita]", "Dr. Stone Stone Wars")]
+        [TestCase("[Cleo] Dr. Stone: Stone Wars [Dual Audio 10bit 1080p][HEVC-x265]", "Dr. Stone: Stone Wars")]
+        public void should_return_the_title_when_there_is_no_season_or_episode(string releaseTitle, string expectedTitle)
+        {
+            var result = Parser.Parser.ParseSeasonTitle(releaseTitle);
+
+            result.Should().NotBeNull();
+            result.SeriesTitle.Should().Be(expectedTitle);
+            result.FullSeason.Should().BeTrue();
+
+            // Which season the title refers to is held in the scene mappings, so it is left
+            // unset here and resolved by ParsingService.
+            result.SeasonNumber.Should().Be(0);
+            result.EpisodeNumbers.Should().BeEmpty();
+            result.AbsoluteEpisodeNumbers.Should().BeEmpty();
+        }
+
+        // DownloadDecisionComparer reads ParsedEpisodeInfo.Quality while sorting, so leaving it
+        // unset fails the whole search with "failed to compare two elements in the array"
+        // rather than skipping one release.
+        [TestCase("[Ohys-Raws] Dr. Stone Stone Wars (BD 1280x720 x264 AAC)")]
+        [TestCase("[Cleo] Dr. Stone: Stone Wars [Dual Audio 10bit 1080p][HEVC-x265]")]
+        public void should_populate_the_fields_the_decision_engine_reads(string releaseTitle)
+        {
+            var result = Parser.Parser.ParseSeasonTitle(releaseTitle);
+
+            result.Quality.Should().NotBeNull();
+            result.Quality.Quality.Should().NotBeNull();
+            result.Languages.Should().NotBeNullOrEmpty();
+        }
+
+        // Some groups bracket every part of the name including the title, and a bracketed group
+        // can contain another. Both leave metadata behind instead of a title, which is worse
+        // than returning nothing because it gets used as a lookup key.
+        [TestCase("[FY-Raws][Dr.STONE: Stone Wars][BDRip][1080P_x265(10bit)-FLAC][ALL]")]
+        [TestCase("[Group][Some Title][1080p][x265(10bit)]")]
+        public void should_not_return_a_title_when_stripping_leaves_metadata(string releaseTitle)
+        {
+            Parser.Parser.ParseSeasonTitle(releaseTitle).Should().BeNull();
+        }
+
+        // Groups mark extras inside brackets, which is what stripping metadata removes, so an
+        // opening or an OVA would otherwise look exactly like a season pack.
+        [TestCase("[Ohys-Raws] Dr. Stone Stone Wars (OVA)")]
+        [TestCase("[Ohys-Raws] Dr. Stone Stone Wars [OVA]")]
+        [TestCase("[Ohys-Raws] Dr. Stone Stone Wars (Special)")]
+        [TestCase("[Ohys-Raws] Dr. Stone Stone Wars (NCOP)")]
+        [TestCase("[Ohys-Raws] Dr. Stone Stone Wars (NCED)")]
+        public void should_not_return_a_title_for_an_extra(string releaseTitle)
+        {
+            Parser.Parser.ParseSeasonTitle(releaseTitle).Should().BeNull();
+        }
+
+        // The helper is not wired into ParseTitle, so anything ParseTitle rejects today it
+        // still rejects. Otherwise nothing would ever be unparsable and crap detection, daily
+        // date validation and the import path all lose their only signal.
+        [TestCase("THIS SHOULD NEVER PARSE")]
+        [TestCase("Vrq6e1Aba3U amCjuEgV5R2QvdsLEGYF3YQAQkw8")]
+        [TestCase("[Ohys-Raws] Dr. Stone Stone Wars (BS11 1280x720 x264 AAC)")]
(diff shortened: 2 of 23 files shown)


src/NzbDrone.Core/Parser/Parser.cs
```diff
@@ -801,6 +811,54 @@ public static ParsedEpisodeInfo ParseTitle(string title)
...
+                ReleaseTitle = releaseTitle,
+                SeriesTitle = seriesTitle,
+                SeriesTitleInfo = new SeriesTitleInfo
+                {
+                    Title = seriesTitle,
+                    TitleWithoutYear = seriesTitle
+                },
+                SeasonNumber = 0,
+                EpisodeNumbers = Array.Empty<int>(),
+                AbsoluteEpisodeNumbers = Array.Empty<int>(),
```
mynameisbogdan:
```suggestion
                EpisodeNumbers = [],
                AbsoluteEpisodeNumbers = [],
```

src/NzbDrone.Core/Parser/Parser.cs
```diff
@@ -801,6 +811,54 @@ public static ParsedEpisodeInfo ParseTitle(string title)
...
+            {
+                return null;
+            }
+
+            var seriesTitle = DuplicateSpacesRegex.Replace(ReleaseMetadataRegex.Replace(releaseTitle, " "), " ").Trim();
+
+            // Anything left in brackets means the title was not recovered. Either it was itself
+            // bracketed, as in [Group][Title][Source][Codec], or a bracketed group contained
+            // another and could not be removed. Both leave metadata behind rather than a title.
+            if (seriesTitle.IsNullOrWhiteSpace() || seriesTitle.IndexOfAny(new[] { '[', ']', '(', ')', '{', '}' }) >= 0)
```
mynameisbogdan: We can use the new collections, I think they got added in .NET 8.

```suggestion
            if (seriesTitle.IsNullOrWhiteSpace() || seriesTitle.IndexOfAny(['[', ']', '(', ')', '{', '}']) >= 0)
```

src/NzbDrone.Core/Parser/Parser.cs
```diff
@@ -801,6 +811,54 @@ public static ParsedEpisodeInfo ParseTitle(string title)
...
+            var result = new ParsedEpisodeInfo
+            {
+                ReleaseTitle = releaseTitle,
+                SeriesTitle = seriesTitle,
+                SeriesTitleInfo = new SeriesTitleInfo
+                {
+                    Title = seriesTitle,
+                    TitleWithoutYear = seriesTitle
+                },
+                SeasonNumber = 0,
```
mynameisbogdan: I was thinking here of using something outside of our current usage, since 0 is used as special season. WDYT?

```suggestion
                SeasonNumber = -1,
```

JonzaUK: good spot didn't think of this, I'll update the ParsingService and test

src/NzbDrone.Core/DecisionEngine/DownloadDecisionMaker.cs
```diff
@@ -88,6 +88,14 @@ private IEnumerable<DownloadDecision> GetDecisions(List<ReleaseInfo> reports, bo
                         }
                     }
 
+                    // Only for searches, where the season being looked for is already known and
+                    // a pack named by its season title is what was being asked for. RSS keeps
+                    // ignoring releases it cannot parse.
+                    if (searchCriteria != null)
+                    {
+                        parsedEpisodeInfo ??= Parser.Parser.ParseSeasonTitle(report.Title);
```
markus101: I don't like that this tries to parse it conditionally, either we can parse it or we can't, but being a search shouldn't influence whether we can parse it a different way.

Instead we should try parsing similar to how specials are parsed above.

src/NzbDrone.Core.Test/ParserTests/SeasonTitleParserFixture.cs
```diff
@@ -0,0 +1,77 @@
...
+namespace NzbDrone.Core.Test.ParserTests
+{
+    [TestFixture]
+    public class SeasonTitleParserFixture : CoreTest
+    {
+        [TestCase("[Ohys-Raws] Dr. Stone Stone Wars (BS11 1280x720 x264 AAC)", "Dr. Stone Stone Wars")]
+        [TestCase("[S3rNx] Dr. Stone Stone Wars (BDRip 1920x1080 x264 FLAC)", "Dr. Stone Stone Wars")]
+        [TestCase("[DmonHiro] Dr.Stone - Stone Wars (BD, 720p)", "Dr.Stone - Stone Wars")]
+        [TestCase("Dr. Stone Stone Wars [Bluray.1080p.HEVC.AAC.ITA.FLAC.JAP.Sub.Ita]", "Dr. Stone Stone Wars")]
+        [TestCase("[Cleo] Dr. Stone: Stone Wars [Dual Audio 10bit 1080p][HEVC-x265]", "Dr. Stone: Stone Wars")]
```
markus101: Replace the series title with an example title instead of an actual one.

src/NzbDrone.Core/Parser/ParsingService.cs
```diff
@@ -194,6 +194,18 @@ private RemoteEpisode Map(ParsedEpisodeInfo parsedEpisodeInfo, int tvdbId, int t
                     remoteEpisode.MappedSeasonNumber += sceneMapping.SeasonNumber.Value - sceneMapping.SceneSeasonNumber.Value;
                 }
 
+                // A release named only by a season title carries no season number for the regex
+                // parser to find, so the season it refers to comes from the mapping instead.
+                if (parsedEpisodeInfo.FullSeason && parsedEpisodeInfo.SeasonNumber == -1)
```
markus101: I think we should add another indicator for these, like we do for specials.

src/NzbDrone.Core/Parser/Parser.cs
```diff
@@ -516,6 +516,16 @@ public static class Parser
         // Regex to detect whether the title was reversed.
         private static readonly Regex ReversedTitleRegex = new Regex(@"(?:^|[-._ ])(p027|p0801|\d{2,3}E-?\d{2}S)[-._ ]", RegexOptions.Compiled);
 
+        // Used to recover the series title from a release that carries no season, episode or
+        // date, where the title is all there is to go on.
+        private static readonly Regex ReleaseMetadataRegex = new Regex(@"[\[({][^\[\]{}()]*[\])}]", RegexOptions.Compiled);
```
markus101: I think rename to indicate it's for anime and drop the comment.

src/NzbDrone.Core/Parser/Parser.cs
```diff
@@ -516,6 +516,16 @@ public static class Parser
...
+
+        // Groups mark extras inside brackets, which is exactly what stripping metadata removes,
+        // so these have to be checked before the brackets are taken off. Same terms the anime
+        // regexes above treat as specials.
+        private static readonly Regex SpecialMarkerRegex = new Regex(@"\b(special|ova|ovd|ncop|nced)\b",
```
markus101: Similar here as well.

src/NzbDrone.Core/Parser/Parser.cs
```diff
@@ -801,6 +811,54 @@ public static ParsedEpisodeInfo ParseTitle(string title)
...
+                {
+                    Title = seriesTitle,
+                    TitleWithoutYear = seriesTitle
+                },
+                SeasonNumber = -1,
+                EpisodeNumbers = [],
+                AbsoluteEpisodeNumbers = [],
+                FullSeason = true,
+
+                // The decision engine reads these while sorting, so they cannot be left unset.
```
markus101: I don't think the comment is necessary.

src/NzbDrone.Core/Parser/Parser.cs
```diff
@@ -801,6 +811,54 @@ public static ParsedEpisodeInfo ParseTitle(string title)
...
+        public static ParsedEpisodeInfo ParseSeasonTitle(string title)
+        {
+            var releaseTitle = FileExtensions.RemoveFileExtension(title);
+
+            if (SpecialMarkerRegex.IsMatch(releaseTitle))
+            {
+                return null;
+            }
+
+            var seriesTitle = DuplicateSpacesRegex.Replace(ReleaseMetadataRegex.Replace(releaseTitle, " "), " ").Trim();
```
markus101: Will be easier to follow if it's split to multiple lines (also makes debugging easier), 
```suggestion
            var seriesTitle = ReleaseMetadataRegex.Replace(releaseTitle, " ");
            seriesTiltle = DuplicateSpacesRegex.Replace(seriesTiltle, " ");
            seriesTitle = seriesTitle.Trim();
```

markus101 (changes requested):
The overall direction looks fine, but shouldn't be limited to searches.

mynameisbogdan:
> but shouldn't be limited to searches.

There's a high chance of mismatching, I wonder if there's a way to match by year as well in case if it's a movie. Just thinking out loud.

src/NzbDrone.Core.Test/ParserTests/SeasonTitleParserFixture.cs
```diff
@@ -0,0 +1,87 @@
...
+        {
+            var result = Parser.Parser.ParseSeasonTitle(releaseTitle);
+
+            result.Should().NotBeNull();
+            result.SeriesTitle.Should().Be(expectedTitle);
+            result.FullSeason.Should().BeTrue();
+            result.IsSeasonTitleOnly.Should().BeTrue();
+
+            // Which season the title refers to is held in the scene mappings, so it is left
+            // unset here and resolved by ParsingService. Not 0, that is the special season.
```
markus101: Don't think this is necessary, can drop the comments from all the tests.

src/NzbDrone.Core/Parser/Parser.cs
```diff
@@ -516,6 +516,11 @@ public static class Parser
         // Regex to detect whether the title was reversed.
         private static readonly Regex ReversedTitleRegex = new Regex(@"(?:^|[-._ ])(p027|p0801|\d{2,3}E-?\d{2}S)[-._ ]", RegexOptions.Compiled);
 
+        private static readonly Regex AnimeReleaseMetadataRegex = new Regex(@"[\[({][^\[\]{}()]*[\])}]", RegexOptions.Compiled);
```
markus101: I think we probably want to match on non-empty brackets? Also a small name change

```suggestion
        private static readonly Regex AnimeSeasonReleaseMetadataRegex = new Regex(@"[\[({][^\[\]{}()]+[\])}]", RegexOptions.Compiled);
```

src/NzbDrone.Core/DecisionEngine/DownloadDecisionMaker.cs
```diff
@@ -88,6 +88,8 @@ private IEnumerable<DownloadDecision> GetDecisions(List<ReleaseInfo> reports, bo
                         }
                     }
 
+                    parsedEpisodeInfo ??= Parser.Parser.ParseSeasonTitle(report.Title);
```
markus101: I know this is functionally the same, but we can add a bit more information here

```suggestion
                    if (parsedEpisodeInfo == null)
                    {
                    	// Attempt to parse as a release that includes the season title without as season number
                    	parsedEpisodeInfo = Parser.Parser.ParseSeasonTitle(report.Title);
                  	}
```

src/NzbDrone.Core/Parser/Model/ParsedEpisodeInfo.cs
```diff
@@ -23,6 +23,7 @@ public class ParsedEpisodeInfo
         public bool IsPartialSeason { get; set; }
         public bool IsMultiSeason { get; set; }
         public bool IsSeasonExtra { get; set; }
+        public bool IsSeasonTitleOnly { get; set; }
```
markus101: Maybe just `IsSeasonTitle`

```suggestion
        public bool IsSeasonTitle { get; set; }
```

src/NzbDrone.Core/Parser/ParsingService.cs
```diff
@@ -194,6 +194,18 @@ private RemoteEpisode Map(ParsedEpisodeInfo parsedEpisodeInfo, int tvdbId, int t
...
+                // A release named only by a season title carries no season number for the regex
+                // parser to find, so the season it refers to comes from the mapping instead.
+                if (parsedEpisodeInfo.IsSeasonTitleOnly)
+                {
+                    // An alias with no season of its own leaves it unresolved, which matches no
+                    // episodes. Falling back to 0 would match the specials instead.
+                    var mappedSeason = sceneMapping.SceneSeasonNumber ?? sceneMapping.SeasonNumber ?? -1;
+
+                    remoteEpisode.MappedSeasonNumber = mappedSeason;
+                    parsedEpisodeInfo.SeasonNumber = mappedSeason;
```
markus101: I think we should keep `ParsedEpisodeInfo` immutable and not change it here and just rely on the mapped season number.

src/NzbDrone.Core/Parser/ParsingService.cs
```diff
@@ -194,6 +194,18 @@ private RemoteEpisode Map(ParsedEpisodeInfo parsedEpisodeInfo, int tvdbId, int t
                     remoteEpisode.MappedSeasonNumber += sceneMapping.SeasonNumber.Value - sceneMapping.SceneSeasonNumber.Value;
                 }
 
+                // A release named only by a season title carries no season number for the regex
+                // parser to find, so the season it refers to comes from the mapping instead.
+                if (parsedEpisodeInfo.IsSeasonTitleOnly)
+                {
+                    // An alias with no season of its own leaves it unresolved, which matches no
+                    // episodes. Falling back to 0 would match the specials instead.
```
markus101:
```suggestion
                    // Assign the season number from the mapping so we have a valid season number
```

src/NzbDrone.Core/Parser/ParsingService.cs
```diff
@@ -194,6 +194,18 @@ private RemoteEpisode Map(ParsedEpisodeInfo parsedEpisodeInfo, int tvdbId, int t
                     remoteEpisode.MappedSeasonNumber += sceneMapping.SeasonNumber.Value - sceneMapping.SceneSeasonNumber.Value;
                 }
 
+                // A release named only by a season title carries no season number for the regex
+                // parser to find, so the season it refers to comes from the mapping instead.
```
markus101: I think we can drop this separate comment.
```suggestion
```

src/NzbDrone.Core/Parser/Parser.cs
```diff
@@ -801,6 +806,55 @@ public static ParsedEpisodeInfo ParseTitle(string title)
...
+                return null;
+            }
+
+            var seriesTitle = AnimeReleaseMetadataRegex.Replace(releaseTitle, " ");
+            seriesTitle = DuplicateSpacesRegex.Replace(seriesTitle, " ");
+            seriesTitle = seriesTitle.Trim();
+
+            // Anything left in brackets means the title was not recovered. Either it was itself
+            // bracketed, as in [Group][Title][Source][Codec], or a bracketed group contained
+            // another and could not be removed. Both leave metadata behind rather than a title.
```
markus101: I think this can be simplified.

```suggestion
            // If any brackets remain we were unable to extract the season title correctly.
```

markus101 (commented):
Mostly small things and comment clean up.

src/Sonarr.Api.V3/Parse/ParseController.cs
```diff
@@ -35,7 +35,11 @@ public ParseResource Parse(string title, string path)
                 return null;
             }
 
-            var parsedEpisodeInfo = path.IsNotNullOrWhiteSpace() ? Parser.ParsePath(path) : Parser.ParseTitle(title);
+            // Only for a title, so this matches what a release goes through. Importing a file
+            // does not use the season title parser, so parsing a path should not either.
+            var parsedEpisodeInfo = path.IsNotNullOrWhiteSpace()
+                ? Parser.ParsePath(path)
+                : Parser.ParseTitle(title) ?? Parser.ParseSeasonTitle(title);
```
mynameisbogdan: This is confusing, it's `path.IsNotNullOrWhiteSpace() ? Parser.ParsePath(path) : (Parser.ParseTitle(title) ?? Parser.ParseSeasonTitle(title))` or `(path.IsNotNullOrWhiteSpace() ? Parser.ParsePath(path) : Parser.ParseTitle(title)) ?? Parser.ParseSeasonTitle(title)`? I would split this in 2 lines IMO.

src/NzbDrone.Core/Parser/Parser.cs
```diff
@@ -801,6 +806,53 @@ public static ParsedEpisodeInfo ParseTitle(string title)
             return null;
         }
 
+        // Anime season packs are commonly named after the season's own title with no season or
+        // episode number anywhere, so none of the regexes above have anything to match on. The
+        // season title is the only identifier the release carries. Which season it refers to is
+        // held in the scene mappings, so that is resolved later by ParsingService.
```
markus101: Let's move this inside the method, right above `var seriesTitle = AnimeSeasonReleaseMetadataRegex.Replace(releaseTitle, " ");` would be appropriate I think.

src/NzbDrone.Core/Parser/ParsingService.cs
```diff
@@ -194,6 +194,12 @@ private RemoteEpisode Map(ParsedEpisodeInfo parsedEpisodeInfo, int tvdbId, int t
                     remoteEpisode.MappedSeasonNumber += sceneMapping.SeasonNumber.Value - sceneMapping.SceneSeasonNumber.Value;
                 }
 
+                if (parsedEpisodeInfo.IsSeasonTitle)
+                {
+                    // Assign the season number from the mapping so we have a valid season number
+                    remoteEpisode.MappedSeasonNumber = sceneMapping.SceneSeasonNumber ?? sceneMapping.SeasonNumber ?? -1;
```
markus101: Should this fall back to `-1` or should we just let that fail lower down?

src/Sonarr.Api.V3/Parse/ParseController.cs
```diff
@@ -37,6 +37,12 @@ public ParseResource Parse(string title, string path)
 
             var parsedEpisodeInfo = path.IsNotNullOrWhiteSpace() ? Parser.ParsePath(path) : Parser.ParseTitle(title);
 
+            // Only a title falls back to the season title parser, importing a file does not use it.
```
markus101:
```suggestion
            // Try to parse a season title if previous parsing failed and we're not dealing with a path.
```

src/Sonarr.Api.V5/Parse/ParseController.cs
```diff
@@ -42,6 +42,12 @@ public Ok<ParseResource> Parse(string? title, string? path)
 
         var parsedEpisodeInfo = path.IsNotNullOrWhiteSpace() ? Parser.ParsePath(path) : Parser.ParseTitle(title);
 
+        // Only a title falls back to the season title parser, importing a file does not use it.
```
markus101:
```suggestion
        // Try to parse a season title if previous parsing failed and we're not dealing with a path.
```

markus101 (commented):
Looks good, I need to test it out a bit, but just some smaller things.

src/NzbDrone.Core/Parser/ParsingService.cs
mynameisbogdan: When we didn't had `IsSeasonTitle` I said to use -1 same we use 0 for specials, but now I think I would say to default to null so we can skip a lot of custom logic for -1. WDYT?

JonzaUK: dropped it and let it fail lower down `MappedSeasonNumber` already starts as the parsed season which is -1 here so the fallback was just writing the same value back, it only assigns now if the mapping actually has a season number.

mynameisbogdan: But doesn't fallback to 0 when empty/null?

https://github.com/Sonarr/Sonarr/blob/bc2942c28d3c48e9eea61a68baf032ba181aef1e/src/NzbDrone.Core/Parser/Model/RemoteEpisode.cs#L17


`MappedSeasonNumber` is not nullable here.

JonzaUK: it doesnt `MappedSeasonNumber` gets set to `parsedEpisodeInfo.SeasonNumber` when
the RemoteEpisode is built a few lines up so its -1 there not 0?

mynameisbogdan: https://github.com/Sonarr/Sonarr/pull/8836/commits/6370fa1fd247fa26b9c4b678ccff69da5aa87d85

but you did this change, so it's going to fallback to 0, not null.

JonzaUK: that commit only touches the IsSeasonTitle block, `MappedSeasonNumber` is already set when the RemoteEpisode is built so it never hits the int default

https://github.com/Sonarr/Sonarr/blob/6370fa1fd247fa26b9c4b678ccff69da5aa87d85/src/NzbDrone.Core/Parser/ParsingService.cs#L184

thats -1 for a season title parse added a test for it

happy to do the null change, its a bit of work though, `SeasonNumber` is an int and v3 `ReleaseResource` would change too

src/NzbDrone.Core.Test/DecisionEngineTests/DownloadDecisionMakerFixture.cs
```diff
@@ -161,6 +164,19 @@ public void should_not_attempt_to_map_episode_if_not_parsable()
...
+        [Test]
+        public void should_map_a_title_the_regexes_cannot_parse()
+        {
+            GivenSpecifications(_pass1, _pass2, _pass3);
+            _reports[0].Title = "Not parsable";
+
+            Subject.GetRssDecision(_reports).ToList();
+
+            // A release carrying nothing but a title is still a title, so it is looked up
+            // rather than dropped. Without a matching series it is rejected as unknown.
```
markus101: I think we can drop this, it's inferred by the test name

src/NzbDrone.Core.Test/ParserTests/SeasonTitleParserFixture.cs
```diff
@@ -0,0 +1,73 @@
+using FluentAssertions;
+using NUnit.Framework;
+using NzbDrone.Core.Test.Framework;
+
+namespace NzbDrone.Core.Test.ParserTests
+{
+    [TestFixture]
+    public class SeasonTitleParserFixture : CoreTest
+    {
+        [TestCase("[Group] Monkey Island Curse (BS11 1280x720 x264 AAC)", "Monkey Island Curse")]
```
markus101: Tests all need to be replaced with fake series titles. (something like `Series Title` or `Series Curse`)

src/NzbDrone.Core.Test/ParserTests/SeasonTitleParserFixture.cs
```diff
@@ -0,0 +1,73 @@
+using FluentAssertions;
+using NUnit.Framework;
+using NzbDrone.Core.Test.Framework;
+
+namespace NzbDrone.Core.Test.ParserTests
```
markus101: Convert to file scoped namespace
```suggestion
namespace NzbDrone.Core.Test.ParserTests;
```

(and drop the braces/extra indentation)

src/NzbDrone.Core/Parser/Parser.cs
```diff
@@ -801,6 +806,53 @@ public static ParsedEpisodeInfo ParseTitle(string title)
...
+
+            if (AnimeSpecialMarkerRegex.IsMatch(releaseTitle))
+            {
+                return null;
+            }
+
+            // Anime season packs are commonly named after the season's own title with no season or
+            // episode number anywhere, so none of the regexes above have anything to match on. The
+            // season title is the only identifier the release carries. Which season it refers to is
+            // held in the scene mappings, so that is resolved later by ParsingService.
```
markus101:
```suggestion
            // Anime season packs are commonly named after the season's own title with no season or
            // episode numbers. ParsingService will determine the correct series and season number
            // from scene mappings.
```

markus101 (commented):
Working great, just a little clean up and then this should be good to land.

mynameisbogdan:
> then this should be good to land.

I'm still seeing a mismatch between SeasonNumber -1 and 0 for season titles.

JonzaUK:
> > then this should be good to land.
> 
> I'm still seeing a mismatch between SeasonNumber -1 and 0 for season titles.

Can you confirm where you are seeing the mismatch?

src/NzbDrone.Core/Parser/Parser.cs
```diff
@@ -801,6 +806,53 @@ public static ParsedEpisodeInfo ParseTitle(string title)
...
+            var result = new ParsedEpisodeInfo
+            {
+                ReleaseTitle = releaseTitle,
+                SeriesTitle = seriesTitle,
+                SeriesTitleInfo = new SeriesTitleInfo
+                {
+                    Title = seriesTitle,
+                    TitleWithoutYear = seriesTitle
+                },
+                SeasonNumber = -1,
```
mynameisbogdan: Here

src/NzbDrone.Core/Parser/ParsingService.cs
```diff
@@ -194,6 +194,17 @@ private RemoteEpisode Map(ParsedEpisodeInfo parsedEpisodeInfo, int tvdbId, int t
...
                 }
 
+                if (parsedEpisodeInfo.IsSeasonTitle)
+                {
+                    var mappedSeasonNumber = sceneMapping.SceneSeasonNumber ?? sceneMapping.SeasonNumber;
+
+                    if (mappedSeasonNumber.HasValue)
+                    {
+                        // Assign the season number from the mapping so we have a valid season number
+                        remoteEpisode.MappedSeasonNumber = mappedSeasonNumber.Value;
```
mynameisbogdan, line 204: And here as the default is `0` for non-nullable `int` as previously discussed.

mynameisbogdan (changes requested):

mynameisbogdan:
> Can you confirm where you are seeing the mismatch?

Left feedback where.

JonzaUK:
its not 0 there `MappedSeasonNumber` gets set from the parsed season when the `RemoteEpisode` is built at 184 so its already -1 before 204 runs. added a test for that exact case a while back, https://github.com/Sonarr/Sonarr/blob/51f21b2a3c9652d0b68ffee4526807aa2e6c14b6/src/NzbDrone.Core.Test/ParserTests/ParsingServiceTests/MapFixture.cs#L391-L407
I said before I could do the null changes and v3 can fallback to -1 since v3 is an int and then nothing breaks? @markus101 thoughts?

markus101:
Null and v3 API falls back to `-1` seems reasonable to me.

src/NzbDrone.Core/Parser/Parser.cs
```diff
@@ -516,6 +516,11 @@ public static class Parser
         // Regex to detect whether the title was reversed.
         private static readonly Regex ReversedTitleRegex = new Regex(@"(?:^|[-._ ])(p027|p0801|\d{2,3}E-?\d{2}S)[-._ ]", RegexOptions.Compiled);
 
+        private static readonly Regex AnimeSeasonReleaseMetadataRegex = new Regex(@"[\[({][^\[\]{}()]+[\])}]", RegexOptions.Compiled);
```
markus101: This will match across bracket types, for example `[foo)` or `(foo]`. Small improvement:


```suggestion
        private static readonly Regex AnimeSeasonReleaseMetadataRegex = new Regex(@"\[[^\[\]]+\]|\([^()]+\)|\{[^{}]+\}", RegexOptions.Compiled);
```

src/Sonarr.Api.V3/Indexers/ReleaseResource.cs
```diff
@@ -120,7 +120,7 @@ public static ReleaseResource ToResource(this DownloadDecision model)
                 ReleaseHash = parsedEpisodeInfo.ReleaseHash,
                 Title = releaseInfo.Title,
                 FullSeason = parsedEpisodeInfo.FullSeason,
-                SeasonNumber = parsedEpisodeInfo.SeasonNumber,
+                SeasonNumber = parsedEpisodeInfo.SeasonNumber ?? -1,
```
markus101, line 123: We need to do something similar for `ParseResource` where we return a `ParsedEpisodeInfo` directly.
Something like this diff: 

[parse.txt](https://github.com/user-attachments/files/31602200/parse.txt)

markus101 (commented):
This is outside the changeset, but needs to be updated so results from a search don't get rejected:

https://github.com/Sonarr/Sonarr/blob/b84a621e991747360b63e074d06fc7d8b534ef63/src/NzbDrone.Core/DecisionEngine/Specifications/Search/SeasonMatchSpecification.cs#L36

markus101 (approved):
Applied the final round of suggestions and will get this landed soon. Thanks for your work on this!

JonzaUK:
> Applied the final round of suggestions and will get this landed soon. Thanks for your work on this!

Thanks for applying the suggestions, I was on Holiday last week hope to see this land soon.

PR #8837, opened 2026-07-30
New: Also search anime season titles without the season parameter

JonzaUK:
#### Description

When searching for anime, sonarr will add the season parameter causing some results to be blank.
This will search the season aliases without that parameter, meaning more results when using some trackers that use string only searching and helps avoid no results in some senarios such as searching for Dr Stone S2 with the season parameter will search for Dr Stone S2 S02 which returns no results.

#### Submission Checklist

### DIFF
--- a/src/NzbDrone.Core.Test/IndexerSearchTests/ReleaseSearchServiceFixture.cs
+++ b/src/NzbDrone.Core.Test/IndexerSearchTests/ReleaseSearchServiceFixture.cs
@@ -587,6 +587,40 @@ namespace NzbDrone.Core.Test.IndexerSearchTests
         }
 
         [Test]
+        public async Task season_search_for_anime_should_pass_search_mode_from_scene_mapping()
+        {
+            WithEpisodes();
+            _xemSeries.SeriesType = SeriesTypes.Anime;
+            _xemEpisodes.ForEach(e => e.EpisodeFileId = 0);
+
+            // Season 1 episodes map to scene season 2 (see WithEpisodes), so the scene
+            // mapping must target scene season 2 to survive the season filter.
+            Mocker.GetMock<ISceneMappingService>()
+                  .Setup(s => s.FindByTvdbId(It.IsAny<int>()))
+                  .Returns(new List<SceneMapping>
+                  {
+                      new SceneMapping
+                      {
+                          TvdbId = _xemSeries.TvdbId,
+                          SearchTerm = "Dr Stone Stone Wars",
+                          ParseTerm = "drstonestonewars",
+                          SeasonNumber = 2,
+                          SceneSeasonNumber = 2,
+                          SearchMode = SearchMode.SearchTitle
+                      }
+                  });
+
+            var allCriteria = WatchForSearchCriteria();
+
+            await Subject.SeasonSearch(_xemSeries.Id, 1, true, false, true, false);
+
+            var criteria = allCriteria.OfType<AnimeSeasonSearchCriteria>().ToList();
+
+            criteria.Should().Contain(c => c.SearchMode == SearchMode.SearchTitle);
+            criteria.Should().Contain(c => c.SceneTitles.Contains("Dr Stone Stone Wars"));
+        }
+
+        [Test]
         public async Task season_search_for_daily_should_search_multiple_years()
         {
             WithEpisode(1, 1, null, null, "2005-12-30");
--- a/src/NzbDrone.Core.Test/IndexerSearchTests/ReleaseSearchServiceFixture.cs
+++ b/src/NzbDrone.Core.Test/IndexerSearchTests/ReleaseSearchServiceFixture.cs
@@ -602,8 +602,8 @@ namespace NzbDrone.Core.Test.IndexerSearchTests
                       new SceneMapping
                       {
                           TvdbId = _xemSeries.TvdbId,
-                          SearchTerm = "Dr Stone Stone Wars",
-                          ParseTerm = "drstonestonewars",
+                          SearchTerm = "Sonarrs Season Title",
+                          ParseTerm = "sonarrsseasontitle",
                           SeasonNumber = 2,
                           SceneSeasonNumber = 2,
                           SearchMode = SearchMode.SearchTitle
@@ -617,7 +617,7 @@ namespace NzbDrone.Core.Test.IndexerSearchTests
             var criteria = allCriteria.OfType<AnimeSeasonSearchCriteria>().ToList();
 
             criteria.Should().Contain(c => c.SearchMode == SearchMode.SearchTitle);
-            criteria.Should().Contain(c => c.SceneTitles.Contains("Dr Stone Stone Wars"));
+            criteria.Should().Contain(c => c.SceneTitles.Contains("Sonarrs Season Title"));
         }
 
         [Test]
--- a/src/NzbDrone.Core.Test/IndexerSearchTests/ReleaseSearchServiceFixture.cs
+++ b/src/NzbDrone.Core.Test/IndexerSearchTests/ReleaseSearchServiceFixture.cs
@@ -621,7 +621,7 @@ namespace NzbDrone.Core.Test.IndexerSearchTests
         }
 
         [Test]
-        public async Task season_search_for_anime_should_not_group_a_series_wide_alias_with_a_season_title()
+        public async Task season_search_for_anime_should_search_a_series_wide_alias_and_a_season_title_together()
         {
             WithEpisodes();
             _xemSeries.SeriesType = SeriesTypes.Anime;
@@ -654,11 +654,10 @@ namespace NzbDrone.Core.Test.IndexerSearchTests
 
             var criteria = allCriteria.OfType<AnimeSeasonSearchCriteria>().ToList();
 
-            criteria.Should().Contain(c => c.SceneTitles.Contains("Sonarrs Season Title") &&
-                                           !c.SceneTitles.Contains("Sonarrs Series Alias"));
-
-            criteria.Should().Contain(c => c.SceneTitles.Contains("Sonarrs Series Alias") &&
-                                           !c.SearchMode.HasFlag(SearchMode.SearchSeasonTitle));
+            criteria.Should().ContainSingle();
+            criteria.First().SceneTitles.Should().Contain("Sonarrs Season Title");
+            criteria.First().SceneTitles.Should().Contain("Sonarrs Series Alias");
+            criteria.First().SeasonSceneTitles.Should().BeEquivalentTo(new[] { "Sonarrs Season Title" });
         }
 
         [Test]
--- a/src/NzbDrone.Core.Test/IndexerSearchTests/ReleaseSearchServiceFixture.cs
+++ b/src/NzbDrone.Core.Test/IndexerSearchTests/ReleaseSearchServiceFixture.cs
@@ -621,6 +621,47 @@ namespace NzbDrone.Core.Test.IndexerSearchTests
         }
 
         [Test]
+        public async Task season_search_for_anime_should_not_group_a_series_wide_alias_with_a_season_title()
+        {
+            WithEpisodes();
+            _xemSeries.SeriesType = SeriesTypes.Anime;
+            _xemEpisodes.ForEach(e => e.EpisodeFileId = 0);
+
+            Mocker.GetMock<ISceneMappingService>()
+                  .Setup(s => s.FindByTvdbId(It.IsAny<int>()))
+                  .Returns(new List<SceneMapping>
+                  {
+                      new SceneMapping
+                      {
+                          TvdbId = _xemSeries.TvdbId,
+                          SearchTerm = "Sonarrs Season Title",
+                          ParseTerm = "sonarrsseasontitle",
+                          SeasonNumber = 2,
+                          SceneSeasonNumber = 2
+                      },
+                      new SceneMapping
+                      {
+                          TvdbId = _xemSeries.TvdbId,
+                          SearchTerm = "Sonarrs Series Alias",
+                          ParseTerm = "sonarrsseriesalias",
+                          SearchMode = SearchMode.SearchTitle
+                      }
+                  });
+
+            var allCriteria = WatchForSearchCriteria();
+
+            await Subject.SeasonSearch(_xemSeries.Id, 1, true, false, true, false);
+
+            var criteria = allCriteria.OfType<AnimeSeasonSearchCriteria>().ToList();
+
+            criteria.Should().Contain(c => c.SceneTitles.Contains("Sonarrs Season Title") &&
+                                           !c.SceneTitles.Contains("Sonarrs Series Alias"));
+
+            criteria.Should().Contain(c => c.SceneTitles.Contains("Sonarrs Series Alias") &&
+                                           !c.SearchMode.HasFlag(SearchMode.SeasonSearchTitle));
+        }
+
+        [Test]
         public async Task season_search_for_daily_should_search_multiple_years()
         {
             WithEpisode(1, 1, null, null, "2005-12-30");
--- a/src/NzbDrone.Core.Test/IndexerSearchTests/ReleaseSearchServiceFixture.cs
+++ b/src/NzbDrone.Core.Test/IndexerSearchTests/ReleaseSearchServiceFixture.cs
@@ -658,7 +658,7 @@ namespace NzbDrone.Core.Test.IndexerSearchTests
                                            !c.SceneTitles.Contains("Sonarrs Series Alias"));
 
             criteria.Should().Contain(c => c.SceneTitles.Contains("Sonarrs Series Alias") &&
-                                           !c.SearchMode.HasFlag(SearchMode.SeasonSearchTitle));
+                                           !c.SearchMode.HasFlag(SearchMode.SearchSeasonTitle));
         }
 
         [Test]
--- a/src/NzbDrone.Core.Test/IndexerTests/NewznabTests/NewznabRequestGeneratorFixture.cs
+++ b/src/NzbDrone.Core.Test/IndexerTests/NewznabTests/NewznabRequestGeneratorFixture.cs
@@ -200,7 +200,7 @@ namespace NzbDrone.Core.Test.IndexerTests.NewznabTests
             var pages = results.GetAllTiers().Select(t => t.First().Url.FullUri).ToList();
 
             pages.Should().ContainSingle(p => p.Contains("q="));
-            pages.Should().Contain(p => p.Contains("q=Monkey%20Island") && !p.Contains("season="));
+            pages.Should().Contain(p => p.Contains("?t=search&") && p.Contains("q=Monkey%20Island") && !p.Contains("season="));
         }
 
         [Test]
--- a/src/NzbDrone.Core.Test/IndexerTests/NewznabTests/NewznabRequestGeneratorFixture.cs
+++ b/src/NzbDrone.Core.Test/IndexerTests/NewznabTests/NewznabRequestGeneratorFixture.cs
@@ -194,7 +194,7 @@ namespace NzbDrone.Core.Test.IndexerTests.NewznabTests
         public void should_search_season_title_alias_without_the_season_parameter()
         {
             Subject.Settings.AnimeStandardFormatSearch = true;
-            _animeSeasonSearchCriteria.SearchMode = SearchMode.SeasonSearchTitle;
+            _animeSeasonSearchCriteria.SearchMode = SearchMode.SearchSeasonTitle;
 
             var results = Subject.GetSearchRequests(_animeSeasonSearchCriteria);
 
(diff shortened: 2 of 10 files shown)


src/NzbDrone.Core.Test/IndexerSearchTests/ReleaseSearchServiceFixture.cs
```diff
@@ -505,6 +505,40 @@ public async Task season_search_for_anime_should_not_search_for_season_with_file
...
+            // Season 1 episodes map to scene season 2 (see WithEpisodes), so the scene
+            // mapping must target scene season 2 to survive the season filter.
+            Mocker.GetMock<ISceneMappingService>()
+                  .Setup(s => s.FindByTvdbId(It.IsAny<int>()))
+                  .Returns(new List<SceneMapping>
+                  {
+                      new SceneMapping
+                      {
+                          TvdbId = _xemSeries.TvdbId,
+                          SearchTerm = "Dr Stone Stone Wars",
```
markus101: Don't use actual names, replace with placeholders.

src/NzbDrone.Core/Indexers/Newznab/NewznabRequestGenerator.cs
```diff
@@ -466,6 +466,24 @@ public virtual IndexerPageableRequestChain GetSearchRequests(AnimeSeasonSearchCr
...
                         $"&q={NewsnabifyTitle(queryTitle)}&season={NewznabifySeasonNumber(searchCriteria.SeasonNumber)}"));
                 }
+
+                // A season title alias already identifies the season, so sending the season
+                // parameter as well excludes the pack on trackers that match it as text.
+                // Dr. Stone: Stone Wars is season 2, but the pack is not named S02.
+                //
+                // Searched in the same tier rather than as a fallback, because the season
+                // qualified search usually does return something, just not the pack. A
+                // fallback tier would only run when it returned nothing at all.
```
markus101: The comment is useful, but can be shortened up / less verbose.

src/NzbDrone.Core.Test/IndexerTests/NewznabTests/NewznabRequestGeneratorFixture.cs
```diff
@@ -189,6 +190,39 @@ public void should_search_by_standard_season_number()
...
 
+        [Test]
+        public void should_search_season_title_alias_with_and_without_season_parameter()
+        {
+            Subject.Settings.AnimeStandardFormatSearch = true;
+            _animeSeasonSearchCriteria.SearchMode = SearchMode.SearchTitle;
+
+            var results = Subject.GetSearchRequests(_animeSeasonSearchCriteria);
+
+            // Same tier, so both run. A fallback tier would only run when the season
```
markus101: This re-explains what the underlying code does, can be dropped.

markus101 (changes requested):
Seems like a good direction, just small things to address.

src/NzbDrone.Core/Indexers/Newznab/NewznabRequestGenerator.cs
```diff
@@ -466,6 +466,19 @@ public virtual IndexerPageableRequestChain GetSearchRequests(AnimeSeasonSearchCr
                         "tvsearch",
                         $"&q={NewsnabifyTitle(queryTitle)}&season={NewznabifySeasonNumber(searchCriteria.SeasonNumber)}"));
                 }
+
+                // A season title alias already identifies the season, so trackers that match
+                // the season parameter as text will exclude the pack it names.
+                if (searchCriteria.SearchMode.HasFlag(SearchMode.SearchTitle))
+                {
+                    foreach (var queryTitle in queryTitles)
```
markus101: If we just search for all possible titles do we risk searching for a series-level alias that will yield results for multiple series? If so we could end up with a lot more search results than expected and fill up the list of results with things from the wrong season.

Should we separate season query titles from other season titles to avoid that?

src/NzbDrone.Core/Indexers/Newznab/NewznabRequestGenerator.cs
```diff
@@ -466,6 +466,19 @@ public virtual IndexerPageableRequestChain GetSearchRequests(AnimeSeasonSearchCr
...
+                    {
+                        pageableRequests.Add(GetPagedRequests(MaxPages,
```
mynameisbogdan: We'll need to support this in NyaaRequestGenerator too, no?

src/NzbDrone.Core/Indexers/Newznab/NewznabRequestGenerator.cs
mynameisbogdan: Also I noticed this falls under `AnimeStandardFormatSearch` being enabled. Is that right, should not work if standard search is disabled too?

mynameisbogdan: > Should we separate season query titles from other season titles to avoid that?

Screenshot 2026-08-06 at 15 26 34

We should!

src/NzbDrone.Core/Indexers/Nyaa/NyaaRequestGenerator.cs
```diff
@@ -89,7 +90,14 @@ public virtual IndexerPageableRequestChain GetSearchRequests(AnimeSeasonSearchCr
 
             foreach (var searchTitle in searchCriteria.SceneTitles.Select(PrepareQuery))
             {
-                if (Settings.AnimeStandardFormatSearch && searchCriteria.SeasonNumber > 0)
+                // A season title alias already identifies the season, so the season number is
+                // not added to it. There is no season parameter here, so it would become part
+                // of the query text and exclude the pack the alias names.
+                if (searchCriteria.SearchMode.HasFlag(SearchMode.SearchTitle))
```
markus101: I haven't tested this, but I think we still have some potential issues here that I wasn't super clear about in https://github.com/Sonarr/Sonarr/pull/8837#discussion_r3722796100

If we have a series wide title, like the title for a series or an alias that applies to the whole series won't we incorrectly search for those without any modifier because `SceneTitles` doesn't separate those two things?

Using B's example of `Dr. Stone` for Season 2 we'd search for the following:

- `Dr . Stone` 
- `Dr. Stone S2`
- `Dr. Stone: Stone Wars`

It's the first one that's the concern, if we just search for that we're going to get an enormous number of results from all seasons since it's the common part of all aliases.

src/NzbDrone.Core/Indexers/Newznab/NewznabRequestGenerator.cs
```diff
@@ -450,15 +450,33 @@ public virtual IndexerPageableRequestChain GetSearchRequests(AnimeSeasonSearchCr
...
+            {
+                return pageableRequests;
+            }
+
+            var queryTitles = TextSearchEngine == "raw" ? searchCriteria.AllSceneTitles : searchCriteria.CleanSceneTitles;
+
+            // A season title alias already identifies the season, so the season number is not
+            // added to it. Doing so excludes the pack the alias names on trackers that match
+            // the season parameter as text.
+            if (searchCriteria.SearchMode.HasFlag(SearchMode.SearchTitle))
```
markus101: Same issue here as I called out for Nyaa.

src/NzbDrone.Core/Indexers/Nyaa/NyaaRequestGenerator.cs
JonzaUK: could add a flag for whether the title actually names a season? put it in the equality so the two dont merge, and branch on that instead of the search mode. series wide titles keep the season number, only the season alias goes bare. same number of requests. same for newznab

mynameisbogdan: Like a `SearchMode.SeasonSearchTitle`?

JonzaUK: went with your suggestion its SearchMode.SeasonSearchTitle now. the series title stays Default and keeps the season number, the season aliases get SeasonSearchTitle and goes bare. its either/or per criteria rather than both on the same title, so the query count doesn’t change.

markus101: I don't think adding a flag solves it though, how does the indexer know that a specific Query Title is a season title and it's safe to search for it without a season number?

QueryTitles needs to be expanded to include that information so the indexer can make a decision on how to handle it, doesn't it?

JonzaUK: I agree but this was mainly aimed at nyaa which uses strings on better indexers just using the season query would be enough on say nzb indexers but for torrents there is no good indexer with the same level of releases, could scope this down to be specific for nyaa?

markus101: Right, but you still have the problem I outlined, we'll search for `Dr. Stone` without any season identifier

mynameisbogdan: You need to test the changes too, as removing the grouping in GetSceneSeasonMappings sends the search type plus their specific QueryTitles, at least it did last time I tested the changes last time.

JonzaUK: with the current changes Dr Stone is only searched with the season param, season alias are searched bare.

src/NzbDrone.Core/Indexers/Nyaa/NyaaRequestGenerator.cs
```diff
@@ -89,7 +90,14 @@ public virtual IndexerPageableRequestChain GetSearchRequests(AnimeSeasonSearchCr
 
             foreach (var searchTitle in searchCriteria.SceneTitles.Select(PrepareQuery))
             {
-                if (Settings.AnimeStandardFormatSearch && searchCriteria.SeasonNumber > 0)
+                // A season title alias already identifies the season, so the season number is
+                // not added to it. There is no season parameter here, so it would become part
+                // of the query text and exclude the pack the alias names.
```
markus101: Move inside the if block and reword a bit:


```suggestion
                // A season title alias already identifies the season, no need to add a season number.
```

src/NzbDrone.Core/Indexers/Newznab/NewznabRequestGenerator.cs
```diff
@@ -450,15 +450,33 @@ public virtual IndexerPageableRequestChain GetSearchRequests(AnimeSeasonSearchCr
...
+            if (!SupportsSearch || searchCriteria.SeasonNumber <= 0)
+            {
+                return pageableRequests;
+            }
+
+            var queryTitles = TextSearchEngine == "raw" ? searchCriteria.AllSceneTitles : searchCriteria.CleanSceneTitles;
+
+            // A season title alias already identifies the season, so the season number is not
+            // added to it. Doing so excludes the pack the alias names on trackers that match
+            // the season parameter as text.
```
markus101: Same thing here

src/NzbDrone.Core/DataAugmentation/Scene/SearchMode.cs
```diff
@@ -7,6 +7,11 @@ public enum SearchMode
     {
         Default = 0,
         SearchID = 1,
-        SearchTitle = 2
+        SearchTitle = 2,
+        SeasonTitle = 4,
```
markus101: I think this fits a little better.
```suggestion
        SearchSeasonTitle = 4,
```

src/NzbDrone.Core/DataAugmentation/Scene/SearchMode.cs
```diff
@@ -7,6 +7,11 @@ public enum SearchMode
...
+
+        // Carries the SearchTitle bit so the generators that only care about a title search
+        // are unaffected, the season searches check for this one to drop the season number.
+        SeasonSearchTitle = SearchTitle | SeasonTitle
```
markus101: Do we need this? I feel like the naming here conflicts with the existing/suggested name and muddies up how we're using.

markus101 (commented):
Working well, but I think some clean up around `SearchMode` is needed, it's a little hard to follow what we're enabling and what we're looking for in the various places.

src/NzbDrone.Core/DataAugmentation/Scene/SearchMode.cs
JonzaUK: Done

src/NzbDrone.Core/DataAugmentation/Scene/SearchMode.cs
JonzaUK: Dropped the combined value and updated the call

src/NzbDrone.Core/IndexerSearch/ReleaseSearchService.cs
```diff
@@ -414,13 +412,11 @@ private async Task<List<DownloadDecision>> SearchAnimeSeason(Series series, List
 
             var allEpisodesAiredOrAiringSoon = seasonEpisodes.All(ep => ep.AirDateUtc.HasValue && !ep.AirDateUtc.Value.After(DateTime.UtcNow.AddHours(24)));
 
-            var seasonsToSearch = GetSceneSeasonMappings(series, episodesToSearch)
-                .GroupBy(ep => ep.SeasonNumber)
-                .Select(epList => epList.First())
-                .ToList();
+            var seasonsToSearch = GetSceneSeasonMappings(series, episodesToSearch);
```
markus101: B and I did a little more testing and we've noticed something that I think should be addressed:

```
2026-08-24 19:24:10.3|Info|ReleaseSearchService|Searching indexers for [Dr. STONE : S04]. 1 active indexers
2026-08-24 19:24:10.3|Debug|ReleaseSearchService|Total of 0 reports were found for [Dr. STONE : S04] from 1 indexers
...
2026-08-24 19:24:06.8|Info|ReleaseSearchService|Searching indexers for [Dr. STONE : S04]. 1 active indexers
2026-08-24 19:24:06.8|Debug|Nyaa|Downloading Feed https://nyaa.si/?page=rss&cats=1_0&filter=1&term=Dr.+Stone+S4
2026-08-24 19:24:06.8|Trace|HttpClient|Req: [GET] https://nyaa.si/?page=rss&cats=1_0&filter=1&term=Dr.+Stone+S4
```

Because of the changes here this now triggers two searches, one for the season titles, one for the default search, which leads to the two different searches being triggered.

I think this is also where I got caught up last time, thinking we'd emit a single search which would in turn decide whether the season needs to be appended or not.

src/NzbDrone.Core/DataAugmentation/Scene/SearchMode.cs
```diff
@@ -7,6 +7,7 @@ public enum SearchMode
     {
         Default = 0,
         SearchID = 1,
-        SearchTitle = 2
+        SearchTitle = 2,
+        SearchSeasonTitle = 4
```
markus101: This is really only set in one place, I don't think it's really needed.

src/NzbDrone.Core/IndexerSearch/ReleaseSearchService.cs
markus101: B did a little more testing and noticed something that we think should be addressed:

```
2026-08-24 19:24:10.3|Info|ReleaseSearchService|Searching indexers for [Dr. STONE : S04]. 1 active indexers
2026-08-24 19:24:10.3|Debug|ReleaseSearchService|Total of 0 reports were found for [Dr. STONE : S04] from 1 indexers
...
2026-08-24 19:24:06.8|Info|ReleaseSearchService|Searching indexers for [Dr. STONE : S04]. 1 active indexers
2026-08-24 19:24:06.8|Debug|Nyaa|Downloading Feed https://nyaa.si/?page=rss&cats=1_0&filter=1&term=Dr.+Stone+S4
2026-08-24 19:24:06.8|Trace|HttpClient|Req: [GET] https://nyaa.si/?page=rss&cats=1_0&filter=1&term=Dr.+Stone+S4
```

Because of the changes here this now triggers two searches, one for the season titles, one for the default search, which leads to the two different searches being triggered.

I think this is also where I got caught up last time, thinking we'd emit a single search which would in turn decide whether the season needs to be appended or not.

I'm attaching a diff which cleans some things up with minimal extra changes, but reduces the searches back down to one.

```
2026-08-24 19:55:09.9|Info|ReleaseSearchService|Searching indexers for [Dr. STONE : S04]. 1 active indexers
2026-08-24 19:55:09.9|Debug|Nyaa|Downloading Feed https://nyaa.si/?page=rss&cats=1_0&filter=1&term=Dr.+Stone+S4
2026-08-24 19:55:09.9|Trace|HttpClient|Req: [GET] https://nyaa.si/?page=rss&cats=1_0&filter=1&term=Dr.+Stone+S4
```

Diff: [anime-search.txt](https://github.com/user-attachments/files/31401440/anime-search.txt)

markus101 (commented):
I've added a diff with a suggested fix, let me know if you have any questions, but reducing the number of searches cleans up the logging, but also avoid writing last search time multiple times.

src/NzbDrone.Core/IndexerSearch/Definitions/AnimeSeasonSearchCriteria.cs
```diff
@@ -1,8 +1,15 @@
+using System.Collections.Generic;
+using System.Linq;
+
 namespace NzbDrone.Core.IndexerSearch.Definitions
 {
     public class AnimeSeasonSearchCriteria : SearchCriteriaBase
     {
         public int SeasonNumber { get; set; }
+        public List<string> SeasonSceneTitles { get; set; } = new List<string>();
```
mynameisbogdan: You can use collection initializer since C# 12.

```suggestion
        public List<string> SeasonSceneTitles { get; set; } = [];
```

src/NzbDrone.Core/IndexerSearch/Definitions/AnimeSeasonSearchCriteria.cs
```diff
@@ -1,8 +1,15 @@
...
+
+        public List<string> AllSeasonSceneTitles => SeasonSceneTitles.Concat(CleanSeasonSceneTitles).Distinct().ToList();
+        public List<string> CleanSeasonSceneTitles => SeasonSceneTitles.Select(GetCleanSceneTitle).Distinct().ToList();
```
mynameisbogdan: Should use InvariantCultureIgnoreCase here too?

```suggestion
        public List<string> AllSeasonSceneTitles => SeasonSceneTitles.Concat(CleanSeasonSceneTitles).Distinct(StringComparer.InvariantCultureIgnoreCase).ToList();
        public List<string> CleanSeasonSceneTitles => SeasonSceneTitles.Select(GetCleanSceneTitle).Distinct(StringComparer.InvariantCultureIgnoreCase).ToList();
```

src/NzbDrone.Core/IndexerSearch/ReleaseSearchService.cs
```diff
@@ -414,15 +424,30 @@ private async Task<List<DownloadDecision>> SearchAnimeSeason(Series series, List
...
             var allEpisodesAiredOrAiringSoon = seasonEpisodes.All(ep => ep.AirDateUtc.HasValue && !ep.AirDateUtc.Value.After(DateTime.UtcNow.AddHours(24)));
 
-            var seasonsToSearch = GetSceneSeasonMappings(series, episodesToSearch)
-                .GroupBy(ep => ep.SeasonNumber)
-                .Select(epList => epList.First())
+            var seasonMappings = GetSceneSeasonMappings(series, episodesToSearch);
+
+            var seasonsToSearch = seasonMappings
+                .GroupBy(m => m.SeasonNumber)
+                .Select(g => g.First())
```
markus101: We can keep the same variables.

```suggestion
                .GroupBy(ep => ep.SeasonNumber)
                .Select(epList => epList.First())
```

src/NzbDrone.Core/Indexers/Newznab/NewznabRequestGenerator.cs
```diff
@@ -450,16 +450,33 @@ public virtual IndexerPageableRequestChain GetSearchRequests(AnimeSeasonSearchCr
...
-
-                foreach (var queryTitle in queryTitles)
+            foreach (var queryTitle in queryTitles)
+            {
+                if (seasonQueryTitles.Contains(queryTitle, StringComparer.InvariantCultureIgnoreCase))
+                {
+                    // A season title alias already identifies the season, no need to add a season number.
+                    pageableRequests.Add(GetPagedRequests(MaxPages,
+                        Settings.AnimeCategories,
+                        "tvsearch",
```
markus101: Since we're not adding any other parameters I think this should just be `search`(matches the absolute episode searching we do for single episodes):


```suggestion
                        "search",
```

markus101 (approved):
Working great, just two minor suggestions. Going to hold off merging until #8836 lands though.

Thanks for all the work on this!

PR #8844, opened 2026-08-02
Fixed: Announce clipboard copy results to screen readers

coopa11y:
#### Description

Clipboard buttons currently show a visual icon after a copy attempt, but screen reader users do not receive confirmation that the copy succeeded or failed. Some icon-only clipboard buttons also do not have a consistent accessible name.

This updates the shared clipboard button so every instance has an accessible name, successful copies are announced as status updates, and failures are announced as alerts. The clipboard and result icons are hidden from assistive technology so they do not add unnecessary information.

Because this is handled by the shared component, it improves clipboard buttons in Security Settings, Custom Format exports, Series Details, and the iCal dialog. It reuses Sonarr's existing status indicator component and does not add any new screen-reader-only CSS.

The full frontend ESLint check, frontend Webpack build, and `git diff --check` pass.

#### Submission Checklist


frontend/src/Components/Link/ClipboardButton.tsx
```diff
@@ -17,10 +19,15 @@ export default function ClipboardButton({
...
   label,
+  title = translate('CopyToClipboard'),
   className = styles.button,
   ...otherProps
 }: ClipboardButtonProps) {
   const [state, setState] = useState<ClipboardState>(null);
+  const statusLabel =
+    state === 'error'
+      ? translate('CopyToClipboardError')
+      : translate('CopiedToClipboard');
```
mynameisbogdan: If used only once, why not set it this directly in `<StatusIndicator />`?

coopa11y: @mynameisbogdan Good point, I moved the translation directly into StatusIndicator and removed the one-use variable.

PR #8859, opened 2026-08-05
New: Treat stalled torrents as failed after a configurable timeout

### LINKED ISSUE #958: Torrent, remove and blacklist torrent without seeds
Hi all
I understand that there is a rule that the torrent have to have atleast 1 seed to fetch the torrent.

But often i get that is no seeds at all when i look it up in Transmission.

So i suggest that we can have a setting where we choose how many seeds we want.

And since that is not possible to see on all trackers we should also have a rule that if nothings happens in like 15-30 minutes the torrent are deleted and blacklisted and a new one will be downloaded.

Thanks in advance and thanks for a great application.

Darlio:
IMPORTANT -->

This text was generated by an AI so it is better written than by me, I could rewrite things, but it would just make it worse... but all facts and tests and real life usage are mine.

Also it would take me a lot more time than an AI to write all this, and in the end... I use it and it's helping me a lot to get rid of the 'dead' torrents... So why not share ?

Please consider reviewing this, and throw it away if this seems to you not 'enterprise-grade' enough (which your software is)... Or help me get it to your standard ! Any comment and suggestions are welcome and will be processed !

Here are some facts why I needed this, and now this is solved -->
- 36 of 37 incomplete torrents with availability < 1.0 — no complete copy ever = they could never complete !
- Because of them, and my config of qBit etc... they were holding all active slots... so no room for real torrents !
- one was spotted at 94.1% for 38 days !

--------

#### Database Migration

NO

#### Description

This is not 'mine' it is a port from Radarr/Radarr#11555 by @Hzoid. The commit is authored to them.

WHAT IT ADDS
Optional "Stalled Torrent Timeout" (advanced, under Failed Download Handling, default 0 = off).
When a torrent the client is actively downloading has made no progress for longer than the timeout, it is treated as failed through the existing pipeline: blocklisted, with removal and redownload following the current Failed Download Handling settings. Progress is tracked from a change in remaining size or client status, on the existing once-a-minute tracked-download refresh.

SCOPE
- torrent protocol only
- only downloads grabbed by Sonarr
- paused and queued torrents are not counted
- disabled by default

DIFFERENCES FROM THE RADARR PR
- warn message says Sonarr
- `return;` added after the existing encrypted/failed branch in Check()
- help text notes the timer is held in memory and restarts with Sonarr
- settings control ported to the v5 TypeScript UI (DownloadClientOptions.tsx, useDownloadClientSettings.ts); the .js file the Radarr PR edits does not exist on v5-develop

VALIDATION on v5-develop (.NET 10)
- dotnet build src/Sonarr.sln -- 0 warnings, 0 errors
- Sonarr.Core.Test, TrackedDownloadService fixtures -- 9/9
- eslint -- clean on both frontend files
- mutation test: reverting ONLY TrackedDownloadService.cs still builds 0 warnings / 0 errors, and the new test then fails with "Expected nullable date and time to have a value, but found <null>"; restored -> 9/9 again. The test is therefore not tautological.

UPSTREAM
Radarr/Radarr#11555 has open review feedback (setting validation; gating the stall check on download progress) that is not reflected here. Draft until that settles.

* Closes #958

### DIFF
--- a/frontend/src/Settings/DownloadClients/Options/DownloadClientOptions.tsx
+++ b/frontend/src/Settings/DownloadClients/Options/DownloadClientOptions.tsx
@@ -124,6 +124,22 @@ function DownloadClientOptions({
                 />
               </FormGroup>
             ) : null}
+
+            <FormGroup
+              advancedSettings={showAdvancedSettings}
+              isAdvanced={true}
+              size={sizes.MEDIUM}
+            >
+              <FormLabel>{translate('StalledTorrentTimeout')}</FormLabel>
+
+              <FormInputGroup
+                type={inputTypes.NUMBER}
+                name="stalledTorrentTimeout"
+                helpText={translate('StalledTorrentTimeoutHelpText')}
+                onChange={handleInputChange}
+                {...settings.stalledTorrentTimeout}
+              />
+            </FormGroup>
           </Form>
         </FieldSet>
       ) : null}
--- a/frontend/src/Settings/DownloadClients/Options/useDownloadClientSettings.ts
+++ b/frontend/src/Settings/DownloadClients/Options/useDownloadClientSettings.ts
@@ -5,6 +5,7 @@ export interface DownloadClientSettingsModel {
   enableCompletedDownloadHandling: boolean;
   autoRedownloadFailed: boolean;
   autoRedownloadFailedFromInteractiveSearch: boolean;
+  stalledTorrentTimeout: number;
 }
 
 const PATH = '/settings/downloadclient';
--- a/src/NzbDrone.Core.Test/Download/TrackedDownloads/TrackedDownloadServiceFixture.cs
+++ b/src/NzbDrone.Core.Test/Download/TrackedDownloads/TrackedDownloadServiceFixture.cs
@@ -1,8 +1,10 @@
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using FluentAssertions;
 using Moq;
 using NUnit.Framework;
+using NzbDrone.Common.Cache;
 using NzbDrone.Core.Download;
 using NzbDrone.Core.Download.History;
 using NzbDrone.Core.Download.TrackedDownloads;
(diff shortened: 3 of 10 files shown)


Darlio:
Fair flag, and I'd rather answer it straight than argue with it.
I'm an infrastructure engineer, not a C# developer. I use an AI assistant for drafting, including the description above, and also this comment, by the way :-)... and that's what tripped this.

The engineering decisions, the verification and the responsibility are mine.

The feature isn't mine, and isn't AI-generated. It's @Hzoid's, from Radarr/Radarr#11555. The commit is authored to them on purpose. My part was the port.

The port was real work rather than a cherry-pick. DownloadClientOptions.js no longer exists on v5-develop, so the setting had to be rebuilt against the TypeScript UI. Two conflicts needed resolving by hand — v5 added a source parameter to PublishDownloadFailedEvent, and a test adjacent to the one this brings.

Verified rather than assumed, on v5-develop with .NET 10: build 0/0, TrackedDownloadService fixtures 9/9, eslint clean. My first attempt failed on SA1508 and unquoted translate() keys — I mention them because "it builds" only means something if you know what it took.

Running in production since the start of August, timeout 120 minutes.

Kept as a draft: #11555 still has open review feedback, and I'd rather this tracks whatever is agreed there.

Hzoid:
Can confirm this was work done by my original PR. Thanks @Darlio for porting, but my original PR hasn't been approved yet.

Darlio:
> Can confirm this was work done by my original PR. Thanks @Darlio for porting, but my original PR hasn't been approved yet.

You are welcome... I should say THANK YOU for what you did !!!! If this could be taken it would be great for both Radarr and Sonarr !!!!

mynameisbogdan:
> but my original PR hasn't been approved yet.

AFAIK it's not fully finished since it still lacks the validation for the timeout. And same to ignore if it's download progress is above 5-10%.

Looking at Radarr's PR, none of these 2 concerns were took into consideration.

PR #8865, opened 2026-08-08
Fix TrackedDownloadAlreadyImported matching an unrelated download's import history by episode ID alone

### LINKED ISSUE #8864: TrackedDownloadAlreadyImported can mark an un-imported download "already imported" by episode ID alone, causing removeCompletedDownloads to delete files before they're ever imported

### Current Behavior

`TrackedDownloadAlreadyImported.IsImported()` can mark a download as fully imported based solely on matching `EpisodeId` against Sonarr's history - without checking whether the matched `DownloadFolderImported` history event actually came from *this specific* download's current output directory.

In production, a download client's own duplicate detection (NZBGet, in this case - logs show `Skipping duplicate ..., found in history with exactly same content`) appears to reuse/match the same `DownloadId` across genuinely distinct grab attempts. When that happens, `HistoryService.FindByDownloadId()` surfaces an **older, already-imported attempt's** history for the episode, even though the **current** download's files were never imported. `IsImported()` treats this as proof the current download is complete, fires `DownloadCompletedEvent`, and - because `removeCompletedDownloads` is enabled on the download client - `DownloadEventHub.RemoveFromDownloadClient` calls `RemoveItem(item, deleteData: true)`, permanently deleting the never-imported files.

Observed twice in the same environment within a few hours:
- An entire Rabbids Invasion season pack (8 episodes) - directory vanished ~10 minutes after the download completed, before any import ever ran.
- 11 separate Random Rings episodes deleted in a single ~5 second sweep, each followed 1-2 seconds later by a failed import attempt on the same (now-deleted) path.

### Expected Behavior

Sonarr should only treat a download as "already imported" - and therefore eligible for `removeCompletedDownloads` to delete its files - if the matched history event actually corresponds to *this* download's own files, not merely to some earlier history entry with a matching episode ID. A distinct, never-imported download should never be treated as complete just because a different, unrelated download of the same episode(s) was imported previously.

### Steps To Reproduce

I was not able to reliably reproduce this on demand interactively (it depends on a download client re-matching/reusing a `DownloadId` across grab attempts, which I could not trigger deterministically). Instead I:

1. Captured the exact sequence live in Sonarr's own Debug log (excerpt below) for a genuine, unprompted production occurrence.
2. Traced the responsible code path from that log evidence back to source (`CompletedDownloadService.VerifyImport()` → `TrackedDownloadAlreadyImported.IsImported()` → `DownloadEventHub.RemoveFromDownloadClient` → `DownloadClientBase.DeleteItemData()`), confirming the exact log strings match (`"[{0}] Deleting folder '{1}'."` etc.).
3. Wrote a unit test (`TrackedDownloadAlreadyImportedFixture.should_return_false_if_download_folder_imported_history_came_from_a_different_download`) that constructs the same scenario directly against `IsImported()` - a `TrackedDownload` whose output path differs from the `DroppedPath` recorded in its matched history item - and confirmed it **fails** against the current code (`IsImported()` incorrectly returns `true`) and **passes** once the output-path check is added.

PR with the fix and both regression tests: #8865

### Environment

- OS: Ubuntu (Linux)
- Sonarr: 4.0.19.2997
- Docker Install: No
- Using Reverse Proxy: No
- Browser: N/A (backend-only issue)
- Database: SQLite

Develop

### Trace Logs

I have **Debug**-level logs (not Trace) capturing the exact incident live - I raised the log level specifically to catch this after noticing the symptom, and it fired again within the log's capture window before I was able to also enable Trace. The Debug output already contains the two log lines that pinpoint the exact code path (see below); Trace would add more surrounding detail but the causal sequence is already unambiguous from Debug plus the source trace.

```
2026-08-07 17:40:03.6|Debug|Nzbget|[Random.Rings.S02E07.Tilly.Calls.the.Future.1080p.DSNP.WEB-DL.AAC2.0.x264-AndreMor] Deleting folder '/media/TransferPool/DL-OnPrem-TV/Random.Rings.S02E07.Tilly.Calls.the.Future.1080p.DSNP.WEB-DL.AAC2.0.x264-AndreMor'.
2026-08-07 17:40:03.9|Debug|DownloadEventHub|[Random.Rings.S02E07.Tilly.Calls.the.Future.1080p.DSNP.WEB-DL.AAC2.0.x264-AndreMor] Removing download from NZBGet-Local history
2026-08-07 17:40:05.4|Debug|DownloadedEpisodesImportService|Processing path: /media/TransferPool/DL-OnPrem-TV/Random.Rings.S02E07.Tilly.Calls.the.Future.1080p.DSNP.WEB-DL.AAC2.0.x264-AndreMor
2026-08-07 17:40:05.4|Error|DownloadedEpisodesImportService|Import failed, path does not exist or is not accessible by Sonarr: /media/TransferPool/DL-OnPrem-TV/Random.Rings.S02E07.Tilly.Calls.the.Future.1080p.DSNP.WEB-DL.AAC2.0.x264-AndreMor. Ensure the path exists and the user running Sonarr has the correct permissions to access this file/folder
```

The same pattern (`Deleting folder` → `Removing download from ... history` → `Import failed, path does not exist`, 1-2 seconds apart) repeats for 10 other Random Rings episodes in the same log window, and separately for an entire Rabbids Invasion season pack roughly ten minutes after that download completed.

### Anything else

**Root cause (source trace, current `v5-develop`):**

`src/NzbDrone.Core/Download/TrackedDownloads/TrackedDownloadAlreadyImported.cs`:

```csharp
public bool IsImported(TrackedDownload trackedDownload, List<EpisodeHistory> historyItems)
{
    ...
    var allEpisodesImportedInHistory = trackedDownload.RemoteEpisode.Episodes.All(e =>
    {
        var lastHistoryItem = historyItems.FirstOrDefault(h => h.EpisodeId == e.Id);

        if (lastHistoryItem == null)
        {
            return false;
        }

        return lastHistoryItem.EventType == EpisodeHistoryEventType.DownloadFolderImported;
    });

    return allEpisodesImportedInHistory;
}
```

Called from `CompletedDownloadService.VerifyImport()`, which fires `DownloadCompletedEvent` when this returns `true`. `DownloadEventHub.RemoveFromDownloadClient` handles that event and calls `RemoveItem(item, deleteData: true)` when `removeCompletedDownloads` is enabled, and `DownloadClientBase.DeleteItemData()` then deletes `item.OutputPath` outright - matching the `"[{0}] Deleting folder '{1}'."` log line above exactly.

**Fix + regression tests:** #8865 - requires the matched history event's own recorded `DroppedPath` to fall under the download's current `ImportItem.OutputPath`, plus two new unit tests reproducing the false positive (confirmed failing on current code, passing with the fix). Happy to adjust the approach based on maintainer feedback.

**Workaround in place on our end:** disabled `removeCompletedDownloads`/`removeFailedDownloads` on the affected download client and replaced it with an external script that independently verifies, per file, that Sonarr's history shows a `downloadFolderImported` event whose `droppedPath` matches that exact file (not just episode ID) before removing anything from the download staging directory.

wharris623:
Fixes #8864

## Problem

`TrackedDownloadAlreadyImported.IsImported()` treats a download as fully imported whenever the most recent history event for each episode ID is `DownloadFolderImported` - without checking whether that import actually came from *this* download's current output path. When a download client's own duplicate detection causes `DownloadId` to be reused/matched across genuinely distinct grab attempts (observed with NZBGet's "found in history with exactly same content" handling), this surfaces an older, unrelated attempt's import history for a new, not-yet-imported download of the same episode. Combined with `removeCompletedDownloads`, Sonarr then deletes the new download's files via `RemoveItem(deleteData: true)` before ever attempting to import them.

Full root cause writeup and production log evidence (11 episodes deleted in one sweep, folder deletion followed 1-2 seconds later by "Import failed, path does not exist" for the same path) in #8864.

## Fix

Require the matched history item's own recorded `DroppedPath` to fall under the download's current `ImportItem.OutputPath` before trusting it as evidence the download is complete. A stale import from a different download's directory can no longer satisfy the check, regardless of episode ID or `DownloadId` matching.

## Test plan

- Updated the existing `TrackedDownloadAlreadyImportedFixture` tests to supply a matching `DroppedPath` (they previously didn't set `Data` at all, which would have broken under the fix).
- Added two new regression tests:
  - `should_return_false_if_download_folder_imported_history_came_from_a_different_download` - single-episode case, history import path doesn't match current download.
  - `should_return_false_if_only_one_episode_in_multi_episode_download_has_matching_import_path` - multi-episode case, one episode's history matches, the other's doesn't.
- Confirmed both new tests fail against the pre-fix code (reproducing the bug) and pass with the fix.
- Full `dotnet test src/Sonarr.sln --filter "FullyQualifiedName~TrackedDownloadAlreadyImportedFixture"` run: 8/8 passing, no build warnings under `TreatWarningsAsErrors`.

### DIFF
--- a/src/NzbDrone.Core.Test/Download/TrackedDownloads/TrackedDownloadAlreadyImportedFixture.cs
+++ b/src/NzbDrone.Core.Test/Download/TrackedDownloads/TrackedDownloadAlreadyImportedFixture.cs
@@ -1,19 +1,24 @@
 ﻿using System.Collections.Generic;
+using System.IO;
 using FizzWare.NBuilder;
 using FluentAssertions;
 using NUnit.Framework;
+using NzbDrone.Common.Disk;
 using NzbDrone.Core.Download;
 using NzbDrone.Core.Download.TrackedDownloads;
 using NzbDrone.Core.History;
 using NzbDrone.Core.Parser.Model;
 using NzbDrone.Core.Test.Framework;
 using NzbDrone.Core.Tv;
+using NzbDrone.Test.Common;
 
 namespace NzbDrone.Core.Test.Download.TrackedDownloads
 {
     [TestFixture]
     public class TrackedDownloadAlreadyImportedFixture : CoreTest<TrackedDownloadAlreadyImported>
     {
+        private readonly OsPath _outputPath = new OsPath(@"C:\DropFolder\MyDownload".AsOsAgnostic());
+        private readonly OsPath _otherDownloadOutputPath = new OsPath(@"C:\DropFolder\SomeOtherDownload".AsOsAgnostic());
         private List<Episode> _episodes;
         private TrackedDownload _trackedDownload;
         private List<EpisodeHistory> _historyItems;
@@ -28,11 +33,13 @@ namespace NzbDrone.Core.Test.Download.TrackedDownloads
                                                       .Build();
 
             var downloadItem = Builder<DownloadClientItem>.CreateNew()
+                                                         .With(d => d.OutputPath = _outputPath)
                                                          .Build();
 
             _trackedDownload = Builder<TrackedDownload>.CreateNew()
                                                        .With(t => t.RemoteEpisode = remoteEpisode)
                                                        .With(t => t.DownloadItem = downloadItem)
+                                                       .With(t => t.ImportItem = downloadItem)
                                                        .Build();
 
             _historyItems = new List<EpisodeHistory>();
@@ -46,13 +53,24 @@ namespace NzbDrone.Core.Test.Download.TrackedDownloads
 
         public void GivenHistoryForEpisode(Episode episode, params EpisodeHistoryEventType[] eventTypes)
         {
+            GivenHistoryForEpisode(episode, _outputPath, eventTypes);
+        }
+
+        public void GivenHistoryForEpisode(Episode episode, OsPath droppedFromPath, params EpisodeHistoryEventType[] eventTypes)
+        {
             foreach (var eventType in eventTypes)
             {
-                _historyItems.Add(
-                    Builder<EpisodeHistory>.CreateNew()
+                var history = Builder<EpisodeHistory>.CreateNew()
                                             .With(h => h.EpisodeId = episode.Id)
                                             .With(h => h.EventType = eventType)
-                                            .Build());
+                                            .Build();
+
+                if (eventType == EpisodeHistoryEventType.DownloadFolderImported)
+                {
+                    history.Data["DroppedPath"] = Path.Combine(droppedFromPath.FullPath, "episode.mkv");
+                }
+
+                _historyItems.Add(history);
             }
         }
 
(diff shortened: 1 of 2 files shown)


wharris623:
Closing - see explanation on #8864. The production evidence I cited was actually a manual queue cleanup, not this code path. Apologies for the noise.

PR #8868, opened 2026-08-10
Caching for the same series path in free space specification

mynameisbogdan:
#### Description

Caching the disk space for the same series path for a small amount of time to avoid IO to the same remote mount, which might be slow when >1000 releases in search results.

### DIFF
--- a/src/NzbDrone.Core/DecisionEngine/Specifications/FreeSpaceSpecification.cs
+++ b/src/NzbDrone.Core/DecisionEngine/Specifications/FreeSpaceSpecification.cs
@@ -1,5 +1,7 @@
+using System;
 using System.IO;
 using NLog;
+using NzbDrone.Common.Cache;
 using NzbDrone.Common.Disk;
 using NzbDrone.Common.Extensions;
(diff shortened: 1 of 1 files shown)


src/NzbDrone.Core/DecisionEngine/Specifications/FreeSpaceSpecification.cs
```diff
@@ -33,16 +39,8 @@ public DownloadSpecDecision IsSatisfiedBy(RemoteEpisode subject, ReleaseDecision
...
 
-            try
-            {
-                freeSpace = _diskProvider.GetAvailableSpace(path);
-            }
-            catch (DirectoryNotFoundException)
-            {
-                // Ignore so it'll be skipped in the following checks
-            }
+            var freeSpace = _cache.Get(path, () => GetFreeSpaceForPath(path), TimeSpan.FromSeconds(5));
```
markus101: A little nit: can drop the extra empty line now that there isn't a try block.

markus101 (approved):

PR #8883, opened 2026-08-15
Fixed: Identify indexer failures after testing all indexers

coopa11y:
#### Description

Test All Indexers currently stops spinning when testing finishes, but it does not show which indexers passed or failed.

This keeps the individual results already returned by the API, including results returned with a failed response. It announces a summary, lists indexer names with their warning or failure messages, and marks each indexer card as Test passed, Test warning, Test failed, or Not tested.

Disabled indexers and indexers with invalid settings are identified as not tested.

#### Submission Checklist


frontend/src/Settings/ProviderTestAllResult.ts
```diff
@@ -0,0 +1,36 @@
...
+  id: number;
+  isValid: boolean;
+  validationFailures: ValidationFailure[];
+}
+
+export type ProviderTestStatus = 'passed' | 'warning' | 'failed' | 'notTested';
+
+export function getProviderTestAllResults(error: ApiError | null) {
+  if (error?.statusCode !== 400 || !Array.isArray(error.statusBody)) {
+    return undefined;
```
markus101, line 14: Results should still include valid responses, unless we only want errors.

We should also be consistent in what is returned, keeping it an array even if everything is successful.

src/NzbDrone.Core/Localization/Core/en.json
```diff
@@ -1058,6 +1059,9 @@
   "IndexerSearchNoAvailableIndexersHealthCheckMessage": "All search-capable indexers are temporarily unavailable due to recent indexer errors",
   "IndexerSearchNoInteractiveHealthCheckMessage": "No indexers available with Interactive Search enabled, {appName} will not provide any interactive search results",
   "IndexerSettings": "Indexer Settings",
+  "IndexerTestAllFailed": "Unable to test indexers, try again later",
+  "IndexerTestAllNotTestedHelpText": "Disabled indexers and indexers with invalid settings are not tested",
```
markus101, line 1063: This is present tense, but it doesn't show up until after it was searched.

src/NzbDrone.Core/Localization/Core/en.json
```diff
@@ -1058,6 +1059,9 @@
...
+  "IndexerTestAllSummary": "Indexer tests complete: {passed} passed, {warnings} warnings, {failed} failed, {notTested} not tested",
```
markus101, line 1064: `warnings` is going to have issues with singular results, `1 warnings` won't work.

markus101 (changes requested):
Popping up an information box after testing feels very clunky. Either we need to have a persistent box where status of each indexer is shown or we need to show things on individual indexers instead of trying to conditionally show this based on a button being pressed. If we just generally have the status on the indexer I think that'd be a good place to show it, instead of trying to convert settings into an informational area as well.

This will conflict with the redesign work though.

coopa11y:
@markus101 Thanks, that makes sense. I agree the conditional summary box is not the right fit for the settings page.

The redesign already has SettingsCardStatus, which looks like the better place for persistent indexer status, but these files differ quite a bit between v5-develop and v5-redesign. Would you prefer that I rework this against v5-redesign and close this PR, or keep it on v5-develop with the results shown only on the individual indexers?

I’ll hold off on the smaller result and wording changes until we agree on that direction.

markus101:
I think against the v5-redesign would be best, but worth holding off on doing that work right now as it's not completely done. I'll move this to draft for now.

PR #8895, opened 2026-08-21
Fixed: BroadcasTheNet searches for three digit episode numbers

DivMode:
#### Description

BTN records long running series with three digit episode numbers in the group name, for example `S01E099`. The `Name` filter is matched against that group name, and `BroadcastheNetRequestGenerator` built it with the `:00` format specifier, which pads to a minimum of two digits, so an episode search for episode 99 sent `S01%E99%`.

That never matches. The literal text `E99` does not occur anywhere inside `S01E099` — the characters after the `E` are `0`, `9`, `9`. The same applies to every episode from 1 to 99: a search for episode 9 sends `S01%E09%`, and `S01E009` contains `E00` and `009` but never `E09`.

Episode 100 and above have always worked, because at three digits the two forms converge on the same string. That produces a sharp boundary on an affected series: every episode below 100 returns no results, everything from 100 onwards is found.

This adds a wildcard after the `E` for episode numbers below 100, so one request matches both the two digit and the zero padded form. Numbers of 100 and above keep the existing string, where a wildcard would only broaden the match for no gain. Both the standard and the anime episode paths are affected and both are fixed. The anime path matters in practice because series long enough to pass episode 99 are frequently set to Series Type: Anime. Each keeps the `Name` format it already used — the standard path keeps its wildcards, the anime path does not gain any.

Season searches are deliberately left alone. They already send `S01E%`, which matches three digit numbering fine, which is why season packs were never affected.

Verified against a live BTN indexer. Result counts are the API's own `results` value for the given `Name` filter:

| `Name` filter | results |
| --- | --- |
| `S01%E99%` (previous behaviour, episode 99) | 0 |
| `S01%E%99%` (this change) | 1, group `S01E099` |
| `S01E99` (previous anime behaviour) | 0 |
| `S01E%99` (this change, anime) | 1, group `S01E099` |
| `S01%E01%` on a normally numbered series | 11 |
| `S01%E%01%` on the same series | 11, unchanged |

One trade-off worth naming: on a three digit series a low episode number can also match its neighbours, because the trailing wildcard admits a further digit — searching episode 9 returns `S01E009` plus `S01E090` to `S01E099`. Those are discarded when the release name is parsed, and the count stays far inside one page, so it remains a single request. The trailing wildcard cannot be dropped to avoid this: it is what matches multi episode groups such as `S01E01E02` and variant groups such as `S01E01 - Extended`.

An earlier revision of this PR used a second request tier holding the zero padded form. The wildcard replaces it on @mynameisbogdan's suggestion, which avoids the extra request and also removes a limitation that version had: in the anime path the fallback tier sat behind the season query, so a matching season pack would stop the chain before the padded query ran.

#### Tests

New `BroadcastheNetRequestGeneratorFixture` with 13 tests:

- the wildcarded query is the only request issued, for episodes 1, 9, 97 and 99
- no wildcard and no extra request for 100, 101 and 999
- a multi episode search wildcards only the members below 100
- season search is unchanged
- the anime path, wildcarded below 100 and untouched at 100
- no requests at all when the series has neither a TVDB nor a TVRage id

Note that `BroadcastheNetFixture.should_back_off_and_report_api_key_invalid` fails on `v5-develop` before this change as well when the fixture is run on its own, so it is unrelated to this PR.

#### Database Migration

None.

#### Submission Checklist

None open that I could find.

### DIFF
--- a/src/NzbDrone.Core.Test/IndexerTests/BroadcastheNetTests/BroadcastheNetRequestGeneratorFixture.cs
+++ b/src/NzbDrone.Core.Test/IndexerTests/BroadcastheNetTests/BroadcastheNetRequestGeneratorFixture.cs
@@ -1,0 +1,150 @@
+using System.Collections.Generic;
+using System.Linq;
+using FluentAssertions;
+using NUnit.Framework;
+using NzbDrone.Core.Indexers;
+using NzbDrone.Core.Indexers.BroadcastheNet;
+using NzbDrone.Core.IndexerSearch.Definitions;
+using NzbDrone.Core.Test.Framework;
+using NzbDrone.Core.Tv;
+
+namespace NzbDrone.Core.Test.IndexerTests.BroadcastheNetTests
+{
+    public class BroadcastheNetRequestGeneratorFixture : CoreTest<BroadcastheNetRequestGenerator>
+    {
+        private Series _series;
+
+        [SetUp]
+        public void SetUp()
+        {
+            Subject.Settings = new BroadcastheNetSettings
+            {
+                BaseUrl = "https://api.broadcasthe.net/",
+                ApiKey = "abc"
+            };
+
+            Subject.PageSize = 100;
+
+            _series = new Series { TvdbId = 433335 };
+        }
+
+        private SingleEpisodeSearchCriteria SingleEpisode(int seasonNumber, params int[] episodeNumbers)
+        {
+            return new SingleEpisodeSearchCriteria
+            {
+                Series = _series,
+                SceneTitles = new List<string> { "Series Title" },
+                SeasonNumber = seasonNumber,
+                EpisodeNumber = episodeNumbers.First(),
+                Episodes = episodeNumbers
+                    .Select(e => new Episode { SeasonNumber = seasonNumber, EpisodeNumber = e })
+                    .ToList()
+            };
+        }
+
+        private static List<string> QueriedNames(IndexerPageableRequestChain chain, int tier)
+        {
+            return chain.GetTier(tier)
+                .Select(pageable => pageable.First().HttpRequest.ContentSummary)
+                .ToList();
+        }
+
+        [Test]
+        public void should_search_for_three_digit_episode_number_in_the_same_request()
+        {
+            var results = Subject.GetSearchRequests(SingleEpisode(1, 99));
+
+            results.Tiers.Should().Be(1);
+            QueriedNames(results, 0).Should().HaveCount(1);
+            QueriedNames(results, 0).First().Should().Contain(@"""Name"":""S01%E%99%""");
+        }
+
+        [TestCase(1, "S01%E%01%")]
+        [TestCase(9, "S01%E%09%")]
+        [TestCase(97, "S01%E%97%")]
+        [TestCase(99, "S01%E%99%")]
+        public void should_wildcard_episode_numbers_below_one_hundred(int episodeNumber, string expected)
+        {
+            var results = Subject.GetSearchRequests(SingleEpisode(1, episodeNumber));
+
+            results.Tiers.Should().Be(1);
+            QueriedNames(results, 0).Should().HaveCount(1);
+            QueriedNames(results, 0).First().Should().Contain($@"""Name"":""{expected}""");
+        }
+
+        [TestCase(100, "S01%E100%")]
+        [TestCase(101, "S01%E101%")]
+        [TestCase(999, "S01%E999%")]
+        public void should_not_wildcard_episode_numbers_over_ninety_nine(int episodeNumber, string expected)
+        {
+            var results = Subject.GetSearchRequests(SingleEpisode(1, episodeNumber));
+
+            results.Tiers.Should().Be(1);
+            QueriedNames(results, 0).Should().HaveCount(1);
+            QueriedNames(results, 0).First().Should().Contain($@"""Name"":""{expected}""");
+        }
+
+        [Test]
+        public void should_only_wildcard_the_episodes_below_one_hundred_of_a_multi_episode_search()
+        {
+            var results = Subject.GetSearchRequests(SingleEpisode(1, 99, 100));
+
+            results.Tiers.Should().Be(1);
+            QueriedNames(results, 0).Should().HaveCount(2);
+            QueriedNames(results, 0)[0].Should().Contain(@"""Name"":""S01%E%99%""");
+            QueriedNames(results, 0)[1].Should().Contain(@"""Name"":""S01%E100%""");
+        }
+
+        [Test]
+        public void should_not_search_without_a_tvdb_or_tvrage_id()
+        {
+            _series.TvdbId = 0;
+
+            var results = Subject.GetSearchRequests(SingleEpisode(1, 99));
+
+            results.GetAllTiers().Should().BeEmpty();
+        }
+
+        [Test]
+        public void should_not_change_season_search()
+        {
+            var criteria = new SeasonSearchCriteria
+            {
+                Series = _series,
+                SceneTitles = new List<string> { "Series Title" },
+                SeasonNumber = 1,
+                Episodes = new List<Episode> { new Episode { SeasonNumber = 1, EpisodeNumber = 99 } }
+            };
+
+            var results = Subject.GetSearchRequests(criteria);
+
+            results.Tiers.Should().Be(1);
+            var names = QueriedNames(results, 0);
+            names.Should().HaveCount(2);
+            names[0].Should().Contain(@"""Name"":""Season 1%""");
+
+            // the season wildcard already matches three digit episode numbers
+            names[1].Should().Contain(@"""Name"":""S01E%""");
+        }
+
+        [TestCase(99, "S01E%99")]
+        [TestCase(100, "S01E100")]
+        public void should_search_for_three_digit_episode_number_for_anime(int episodeNumber, string expected)
+        {
+            var criteria = new AnimeEpisodeSearchCriteria
+            {
+                Series = _series,
+                SceneTitles = new List<string> { "Series Title" },
+                SeasonNumber = 1,
+                EpisodeNumber = episodeNumber,
+                AbsoluteEpisodeNumber = episodeNumber,
+                Episodes = new List<Episode> { new Episode { SeasonNumber = 1, EpisodeNumber = episodeNumber } }
+            };
+
+            var results = Subject.GetSearchRequests(criteria);
+
+            results.Tiers.Should().Be(1);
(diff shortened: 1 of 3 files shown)


mynameisbogdan:
Why not add wildcard to prevent an extra request making the search too slow since it's already one request every 5 second?

```diff
diff --git a/src/NzbDrone.Core/Indexers/BroadcastheNet/BroadcastheNetRequestGenerator.cs b/src/NzbDrone.Core/Indexers/BroadcastheNet/BroadcastheNetRequestGenerator.cs
index b68ea1376f..c0acf1520e 100644
--- a/src/NzbDrone.Core/Indexers/BroadcastheNet/BroadcastheNetRequestGenerator.cs
+++ b/src/NzbDrone.Core/Indexers/BroadcastheNet/BroadcastheNetRequestGenerator.cs
@@ -51,7 +51,9 @@ public virtual IndexerPageableRequestChain GetSearchRequests(SingleEpisodeSearch
                     parameters = parameters.Clone();
 
                     parameters.Category = "Episode";
-                    parameters.Name = $"S{episode.SeasonNumber:00}%E{episode.EpisodeNumber:00}%";
+                    parameters.Name = episode.EpisodeNumber is > 0 and < 100
+                        ? $"S{episode.SeasonNumber:00}%E%{episode.EpisodeNumber:00}%"
+                        : $"S{episode.SeasonNumber:00}%E{episode.EpisodeNumber:00}%";
 
                     pageableRequests.Add(GetPagedRequests(MaxPages, parameters));
                 }
@@ -154,7 +156,9 @@ public virtual IndexerPageableRequestChain GetSearchRequests(AnimeEpisodeSearchC
                     parameters = parameters.Clone();
 
                     parameters.Category = "Episode";
-                    parameters.Name = $"S{episode.SeasonNumber:00}E{episode.EpisodeNumber:00}";
+                    parameters.Name = episode.EpisodeNumber is > 0 and < 100
+                        ? $"S{episode.SeasonNumber:00}E%{episode.EpisodeNumber:00}"
+                        : $"S{episode.SeasonNumber:00}E{episode.EpisodeNumber:00}";
 
                     pageableRequests.Add(GetPagedRequests(MaxPages, parameters));
                 }

```

src/NzbDrone.Core.Test/ParserTests/SingleEpisodeParserFixture.cs
```diff
@@ -249,5 +249,17 @@ public void should_parse_split_episode(string postTitle, string title, int seaso
             result.FullSeason.Should().BeFalse();
             result.IsSplitEpisode.Should().BeTrue();
         }
+
+        [TestCase("Renegade.Immortal.S01E099.1080p.WEB-DL.AAC2.0.H.264-VARYG", "Renegade Immortal", 1, 99)]
+        [TestCase("Renegade.Immortal.S01E100.1080p.WEB-DL.AAC2.0.H.264-VARYG", "Renegade Immortal", 1, 100)]
+        [TestCase("Renegade.Immortal.S01E009.1080p.WEB-DL.AAC2.0.H.264-VARYG", "Renegade Immortal", 1, 9)]
```
mynameisbogdan:
```suggestion
        [TestCase("Series.Title.S01E099.1080p.WEB-DL.AAC2.0.H.264-VARYG", "Series Title", 1, 99)]
        [TestCase("Series.Title.S01E100.1080p.WEB-DL.AAC2.0.H.264-GROUP", "Series Title", 1, 100)]
        [TestCase("Series.Title.S01E009.1080p.WEB-DL.AAC2.0.H.264-GROUP", "Series Title", 1, 9)]
```

src/NzbDrone.Core/Indexers/BroadcastheNet/BroadcastheNetRequestGenerator.cs
```diff
@@ -51,7 +51,14 @@ public virtual IndexerPageableRequestChain GetSearchRequests(SingleEpisodeSearch
                     parameters = parameters.Clone();
 
                     parameters.Category = "Episode";
-                    parameters.Name = $"S{episode.SeasonNumber:00}%E{episode.EpisodeNumber:00}%";
+
+                    // Long running series are often uploaded with three digit episode numbers
+                    // (S01E099). 'E99' doesn't occur anywhere in 'E099', so a wildcard after the
+                    // 'E' is needed to match both forms with a single request. Episode numbers of
+                    // 100 and up already render as three digits and need no wildcard.
```
mynameisbogdan:
```suggestion
```

src/NzbDrone.Core/Indexers/BroadcastheNet/BroadcastheNetRequestGenerator.cs
```diff
@@ -154,7 +161,13 @@ public virtual IndexerPageableRequestChain GetSearchRequests(AnimeEpisodeSearchC
                     parameters = parameters.Clone();
 
                     parameters.Category = "Episode";
-                    parameters.Name = $"S{episode.SeasonNumber:00}E{episode.EpisodeNumber:00}";
+
+                    // See GetSearchRequests(SingleEpisodeSearchCriteria), three digit episode
+                    // numbers are especially common on the long running series that use absolute
+                    // numbering.
```
mynameisbogdan:
```suggestion
```

mynameisbogdan (changes requested):

src/NzbDrone.Core.Test/IndexerTests/BroadcastheNetTests/BroadcastheNetRequestGeneratorFixture.cs
```diff
@@ -0,0 +1,150 @@
...
+            results.GetAllTiers().Should().BeEmpty();
+        }
+
+        [Test]
+        public void should_not_change_season_search()
+        {
+            var criteria = new SeasonSearchCriteria
+            {
+                Series = _series,
+                SceneTitles = new List<string> { "Renegade Immortal" },
```
mynameisbogdan:
```suggestion
                SceneTitles = new List<string> { "Series Title" },
```

src/NzbDrone.Core.Test/IndexerTests/BroadcastheNetTests/BroadcastheNetRequestGeneratorFixture.cs
```diff
@@ -0,0 +1,150 @@
...
+        }
+
+        [TestCase(99, "S01E%99")]
+        [TestCase(100, "S01E100")]
+        public void should_search_for_three_digit_episode_number_for_anime(int episodeNumber, string expected)
+        {
+            var criteria = new AnimeEpisodeSearchCriteria
+            {
+                Series = _series,
+                SceneTitles = new List<string> { "Renegade Immortal" },
```
mynameisbogdan:
```suggestion
                SceneTitles = new List<string> { "Series Title" },
```

src/NzbDrone.Core.Test/IndexerTests/BroadcastheNetTests/BroadcastheNetRequestGeneratorFixture.cs
```diff
@@ -0,0 +1,150 @@
...
+
+            _series = new Series { TvdbId = 433335 };
+        }
+
+        private SingleEpisodeSearchCriteria SingleEpisode(int seasonNumber, params int[] episodeNumbers)
+        {
+            return new SingleEpisodeSearchCriteria
+            {
+                Series = _series,
+                SceneTitles = new List<string> { "Renegade Immortal" },
```
mynameisbogdan:
```suggestion
                SceneTitles = new List<string> { "Series Title" },
```

mynameisbogdan (changes requested):

DivMode:
All six suggestions applied in `271b69f08` — both comment blocks dropped, generic series title in the request generator fixture and the parser cases.

On the wildcard itself, thanks — it is the better shape, and I checked against the live API whether anything tighter exists before settling on it:

- `Season`, `Episode`, `ReleaseName` and `GroupName` filter keys are silently ignored — each returns the same count as a deliberately invalid key, so they are not honoured
- `Search` is honoured, but treats `%` literally and ANDs its terms, so it cannot cover both paddings
- `Name` as an array is not OR'd

So `Name` is the only lever, and a LIKE cannot express an optional leading zero. The wildcard is the only single request shape that covers both forms. The cost is neighbouring groups on low episode numbers — searching episode 9 also returns `S01E090` to `S01E099` — which are dropped at parse time and stay far inside one page, so it is still one request. The trailing `%` cannot be dropped to narrow that: it is what matches `S01E01E02` and `S01E01 - Extended`.

It also removes a limitation the tiered revision had, where the anime fallback sat behind the season query and a matching season pack would stop the chain before the padded query ran. Description updated to match.

mynameisbogdan:
Thanks for the credit, Claude. 😉

src/NzbDrone.Core.Test/ParserTests/SingleEpisodeParserFixture.cs
```diff
@@ -249,5 +249,17 @@ public void should_parse_split_episode(string postTitle, string title, int seaso
             result.FullSeason.Should().BeFalse();
             result.IsSplitEpisode.Should().BeTrue();
         }
+
+        [TestCase("Series.Title.S01E099.1080p.WEB-DL.AAC2.0.H.264-VARYG", "Series Title", 1, 99)]
+        [TestCase("Series.Title.S01E100.1080p.WEB-DL.AAC2.0.H.264-GROUP", "Series Title", 1, 100)]
+        [TestCase("Series.Title.S01E009.1080p.WEB-DL.AAC2.0.H.264-GROUP", "Series Title", 1, 9)]
+        public void should_parse_three_digit_padded_episode_numbers(string postTitle, string title, int seasonNumber, int episodeNumber)
```
markus101: These tests aren't testing any changes to parsing, they should either be removed or be part of `should_parse_single_episode`.

We already have this one testing 3-digit numbers:
`Judge Developer 2016 02 25 S20E142` is an existing test case.

markus101 (changes requested):

src/NzbDrone.Core.Test/ParserTests/SingleEpisodeParserFixture.cs
DivMode: Removed in a92acbef4 — you are right, they asserted existing parser behaviour rather than anything this change touches, and `Judge Developer 2016 02 25 S20E142` already covers three-digit numbers.

`SingleEpisodeParserFixture.cs` is now identical to `v5-develop`, so the PR is down to the request generator and its new fixture.

markus101 (approved):

Issue #8900, opened 2026-08-23
Import throws NullReferenceException in AddQualityTokens after the existing episode file has already been deleted


dronkailp:
### Is there an existing issue for this?

- [X] I have searched the existing open and closed issues

### Current Behavior

This one cost me a bit of nerve, because the existing file gets deleted before the crash happens.

When an episode file ends up without a usable `QualityModel`, `AddQualityTokens` throws while building the file name. The problem is the order of things: `UpgradeEpisodeFile` first deletes the current episode file (into the recycle bin) and removes it from the database, and only then calls `MoveEpisodeFile`, which is where the name is built. So the episode ends up with no file at all, and the import that was supposed to replace it never happens.

Since the download stays in the queue, it just keeps trying. One release gave me 928 failed imports in 24 hours, sending the existing file to the recycle bin each time it got that far.

What triggers it for me is a release Sonarr can't parse — Russian episode numbering, like "Series. 1 серия.mkv". If I ask `GET /api/v3/manualimport` about that folder, it answers `quality: null`, `episodes: []`, "Unable to parse file" for every file in it.

I found #2880 and #3177 while searching — same method, but those are old and closed against v2/v3. This is on 4.0.19 and it goes through `UpgradeEpisodeFile`.

### Expected Behavior

A file with no quality should end up as a rejected import, not an exception — and definitely not after the old file is already gone.

Two ways to fix it, either one helps:

1. Guard in `AddQualityTokens`:

```csharp
var quality = episodeFile.Quality;

if (quality?.Quality == null || quality.Revision == null)
{
    _logger.Warn("Episode file has no usable quality: {0}", episodeFile.RelativePath ?? episodeFile.Path);
    quality = new QualityModel(quality?.Quality ?? Quality.Unknown);
}
```

2. Check the quality in `UpgradeEpisodeFile` before deleting the existing file, so a failure can't leave the episode empty.

I'm running the first one locally and the failures stopped. Happy to open a PR.

### Steps To Reproduce

1. Take an episode that already has a file.
2. Put a file Sonarr can't parse into the download folder. Russian episode numbering does it every time: "Series. 1 серия.mkv".
3. Let Sonarr process the completed download.
4. The existing file goes to the recycle bin, then the import throws and the episode is left with nothing. Next pass, same thing again.

### Environment

```
- OS: Alpine Linux
- Sonarr: 4.0.19.2979
- Docker Install: Yes
- Using Reverse Proxy: No
- Browser: n/a
[...]

PR #8903, opened 2026-08-25
Fixed: existing episode file deleted before the new file name is built

dronkailp:
#### Description

`UpgradeEpisodeFile` deletes the existing episode file - recycle bin at line 65, database at line 73 - and only then calls `MoveEpisodeFile` at line 84, where the very first statement (`EpisodeFileMovingService.cs:85`) builds the new file name. Anything that throws while building that name throws after the old file is already gone, and `ImportApprovedEpisodes.cs:193` swallows it in `catch (Exception)` with a single Warn. The episode is left with no file, the download stays in the queue, and the next pass does the same thing again. On my setup one release produced 928 failed imports in 24 hours, binning the existing file every time.

The method already guards against exactly this shape of problem a few lines earlier, with the comment spelling out why:

```
// If there are existing episode files and the root folder is missing, throw, so the old file isn't left behind during the import process.
```

That covers a missing root folder. It does not cover a name that cannot be built. This PR adds the same kind of guard: when there are existing files, build the destination path before the delete loop, so a failure there leaves the existing file where it is.

Test: `should_not_delete_the_existing_file_when_the_new_file_name_cannot_be_built` in `UpgradeMediaFileServiceFixture`. On a clean `v5-develop` it fails with `Moq.MockException: Expected invocation on the mock should never have been performed, but was 1 times: DeleteFile(...)` - the file is in the recycle bin although the move threw. With this change the fixture is 11/11. Full `Sonarr.Core.Test` run: 5155 passed, 28 skipped, 6 failures, all in the network-dependent scene mapping tests, which fail the same way on a clean checkout in the same container.

Known trade-off: the path is built twice, once for the check and once inside `MoveEpisodeFile`. The cheaper alternative is to pass the built path down, which changes the `IMoveEpisodeFiles` signature - happy to do it that way instead if you prefer, or to move the guard elsewhere.

This came out of #8900. My original report there blamed `AddQualityTokens` based on a line number from an optimised build, and that part did not hold up - I have said so in the issue. The ordering problem above is what is left, and it does not depend on which exception was thrown.

#### Submission Checklist

- Related to #8900

### DIFF
--- a/src/NzbDrone.Core.Test/MediaFiles/UpgradeMediaFileServiceFixture.cs
+++ b/src/NzbDrone.Core.Test/MediaFiles/UpgradeMediaFileServiceFixture.cs
@@ -1,3 +1,5 @@
+using System;
+using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using FizzWare.NBuilder;
@@ -5,9 +7,11 @@ using FluentAssertions;
 using Moq;
 using NUnit.Framework;
 using NzbDrone.Common.Disk;
+using NzbDrone.Core.CustomFormats;
 using NzbDrone.Core.Datastore;
 using NzbDrone.Core.MediaFiles;
 using NzbDrone.Core.MediaFiles.EpisodeImport;
+using NzbDrone.Core.Organizer;
 using NzbDrone.Core.Parser.Model;
 using NzbDrone.Core.Test.Framework;
 using NzbDrone.Core.Tv;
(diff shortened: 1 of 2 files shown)


markus101:
This is fixing a symptom of the problem, not the problem itself. Once Sonarr has made the decision to import a file it should have everything it needs to import it, if it doesn't then the _input_ is wrong.

Based on the logs you've provided on the issue and the fixes already in v5 this PR isn't necessary, but we can take a look again once the proper logs are available and we have a cause of the issue.

dronkailp:
Closing this - you were right that it's a symptom fix, and the report it came from doesn't hold up.

I built clean instances with trace logging (4.0.19.3006 and 4.0.19.2979, the build I actually run) and tried three ways to reproduce the failure this PR guards against. All three imported cleanly, trace shows the exact path with no exception. Details are in #8900, which I closed too.

The ordering observation stands on its own - the existing file is removed before the new name is built - but with no reproducible failure behind it, it's a change without a cause, and you already said that's not what you want. If it resurfaces on a stock build I'll come back with logs first.

Thanks for the time both of you spent on it.

PR #8937, opened 2026-09-04
Fix series mapping by relative base folder name

### LINKED ISSUE #8929: Manual Import: series lookup by folder name discards its result, causing false 'Unknown Series' rejections
**Describe the bug**

`ManualImportService.ProcessFile` silently discards the result of its first (folder-name-based) series lookup, so that lookup never actually contributes to matching. Manual import / `DownloadedEpisodesScan` then falls through to a second lookup that parses the *entire relative path* (subfolder + filename) instead of the clean folder name — which is a much noisier string to parse, and fails for a number of otherwise well-formed release names, rejecting the item as "Unknown Series" even though the series is present, monitored, and its episode data is intact.

**To Reproduce**

This path is only exercised when a download has lost its tracked-download linkage (e.g. after a downloader hiccup, or the download client entry expiring) and Sonarr falls back to pure folder/file-name matching for import. With a tracked download that's still linked to its original grab, the queue's own `RemoteEpisode.Series` is used instead and this bug never triggers — which is likely why it's gone unnoticed.

1. Have Sonarr lose the tracked-download link for a completed download (in our case: multiple downloader restarts during an unrelated outage).
2. Call `GET /api/v3/manualimport?folder=<path to the completed download folder>` for a release whose *file name* doesn't parse as cleanly as its *folder name* (e.g. folder `Trigger Point S04E01 1080p ITV WEB-DL AAC2 0 H 264-playWEB EZTV`, file `Trigger Point S04E01 1080p ITV WEB-DL AAC2 0 H 264-playWEB[EZTVx.to].mkv`).
3. Response comes back with `"series": null` and `"rejections": [{"reason": "Unknown Series", "type": "permanent"}]`, even though the series ("Trigger Point") exists, is monitored, and `GET /api/v3/episode?seriesId=<id>` returns the correct episode.

**Expected behavior**

The folder-name lookup's result should be used when it succeeds, the same way the immediately-following lookup is:

```csharp
if (series == null)
{
    _parsingService.GetSeries(relativeFile.Split('\\', '/')[0]);   // <-- result discarded
}

if (series == null)
{
    series = _parsingService.GetSeries(relativeFile);              // <-- only this one is used
}
```

https://github.com/Sonarr/Sonarr/blob/v4.0.19.3009/src/NzbDrone.Core/MediaFiles/EpisodeImport/Manual/ManualImportService.cs#L305-L309

The fix looks like a one-line change:

```csharp
if (series == null)
{
    series = _parsingService.GetSeries(relativeFile.Split('\\', '/')[0]);
}
```

**Environment**

- Sonarr version: 4.0.16.2944 (confirmed the same code is still present, unchanged, in the latest tag as of this report: 4.0.19.3009)
- Docker: `lscr.io/linuxserver/sonarr`
- Reproduced via direct API calls (`/api/v3/manualimport`, `/api/v3/parse`, `/api/v3/episode`) to rule out a UI-side issue — confirmed the series/episode data itself is valid and the parser correctly extracts series/season/episode from the file name via `/api/v3/parse`; the bug is specifically in the lookup-and-discard in `ManualImportService`.

Traced this while debugging a batch of stuck "Unknown Series" imports after an unrelated infrastructure incident dropped several downloads' tracked-download links. Happy to provide more repro detail if useful.

mynameisbogdan:
#### Description
Attempt finding the series by the base folder name from the relative path.

- Closes #8929

### DIFF
--- a/src/NzbDrone.Core/MediaFiles/EpisodeImport/Manual/ManualImportService.cs
+++ b/src/NzbDrone.Core/MediaFiles/EpisodeImport/Manual/ManualImportService.cs
@@ -307,7 +307,7 @@ namespace NzbDrone.Core.MediaFiles.EpisodeImport.Manual
 
                 if (series == null)
                 {
-                    _parsingService.GetSeries(relativeFile.Split('\\', '/')[0]);
+                    series = _parsingService.GetSeries(relativeFile.Split('\\', '/')[0]);
                 }
 
                 if (series == null)


src/NzbDrone.Core/MediaFiles/EpisodeImport/Manual/ManualImportService.cs
```diff
@@ -307,7 +307,7 @@ private ManualImportItem ProcessFile(string rootFolder, string baseFolder, strin
 
                 if (series == null)
                 {
-                    _parsingService.GetSeries(relativeFile.Split('\\', '/')[0]);
```
markus101, line 310: [Nailed it](https://media.tenor.com/ILs8Yn2x4LsAAAAM/andy-bernard.gif)

mynameisbogdan: [60% of the time works every time.](https://c.tenor.com/sH_KNNF07EoAAAAd/tenor.gif)

PR #8940, opened 2026-09-06
Restore auto width behavior from previous FontAwesome versions

mynameisbogdan:
#### Description

`fixedWidth` was deprecated in favor of `widthAuto` as enabled by default since 7.0. We're switching to the old behavior as it impacts some parts of the UI to look weird due to differences in width.

https://docs.fontawesome.com/web/use-with/react/style#automatic-widths

Before
Screenshot 2026-09-06 at 12 44 36

After
Screenshot 2026-09-06 at 12 44 48


markus101:
The redesign drops Font Awesome in favour of Lucid React, so I don't think we need this (also working on unblocking the final bits of that).

mynameisbogdan:
We don't, even v4 is on FA 6.

https://github.com/Sonarr/Sonarr/blob/cab419ade8ac7fcab5bf80394ee492abd35d5f5a/package.json#L24-L28

Issue #8938, opened 2026-09-05
Multi-season support


IonBazan:
### Is there an existing issue for this?

- [x] I have searched the existing open and closed issues

### Is your feature request related to a problem? Please describe

This is yet another proposal for multi-season download support, for those trying to archive or watch completed shows.

### Describe the solution you'd like

While there are already similar tickets closed, like #8133 and attempts to solve my issue in #8508 and #8668, I would like to propose a limited and scoped solution for specific (and IMHO most common) scenario:
- Downloading an entire show or multi-season packs using:
  - Interactive search
  - Manually adding category in existing DL client job

I would like to hear from maintainers and other users their opinions on following proposed approach.
Keep ignoring MS packs in:
- automatic search
- RSS
- auto-upgrades

Show a warning next to interactive search results, but allow to auto-import it from a DL triggered from such search or externally added to the DL client. 
This way, we would follow the same experience as manual import would, but need to consider edge cases like mixed qualities, auto upgrades and incomplete seasons.

I am not fully familiar with auto-upgrades feature so I believe some additional restrictions might need to be applied - like forcing users to unmonitor the show first or disable auto-upgrade if season was picked up this way. 

### Describe alternatives you've considered

Current v4 suffers from "data loss" issue reported in #8133 and v5 now refuses to handle multi-season packs if they were added externally which leaves manual import as the only way of handling such packs. It's much safer now but it means users must manually monitor DL progress and trigger the import once it's done.

### Anything else?

I have tried drafting an implementation around comments from #8668 but have not tested it yet. Will submit PR once the discussion is finalised, as per `CONTRIBUTING.md`.

I am aware that historically multi-season packs support was discouraged but I believe many users would still benefit from that functionality, even by sacrificing auto-upgrades, RSS monitoring or other functionalities for that particular TV Show.

### Implementation plan

- [x] Parser: track full season range on multi-season packs - #8941 
- [ ] ⌛  Resolve episodes across every season in a pack - #8958
- [ ] Allow multi-season packs to import completely
- [ ] Queue: report every season a download covers

markus101:
I'm not opposed to having better support for them and I think avoiding them for automatic grabs is a reasonable first step, though I'm not sure that the upgrade path is necessarily an issue, but `Allow Season Pack Upgrades` would need to consider them.

High level I see a few things that need to happen:

- Parsing, making sure we're handling all the different naming formats
- Adding multiple seasons to `ParsedEpisodeInfo`
[...]

PR #8941, opened 2026-09-06
Track full season range for multi-season packs

IonBazan:
#### Description

First stage to prepare for multi-season packs support. `ParsedEpisodeInfo` now stores the complete list of seasons in a pack (`Show.S01-S05...`) instead of only the first one. `SeasonNumber` becomes a thin accessor over the new `SeasonNumbers` array, and `IsMultiSeason` is now computed (`SeasonNumbers.Length > 1`) rather than a separately set flag.

Also fixed the parsing logic so it handles more than two season markers and non-contiguous ones:
- Exactly 2 seasons mentioned with a gap (e.g. `S01 S04`) is treated as a range: `1,2,3,4`.
- More than 2 seasons mentioned (e.g. `S01 S02 S03 S04` or `S01 S03 S04`) is treated as the literal list, gaps included.

Today a pack like `S01-S05` collapses to season 1 only, which is the root cause of multi-season packs only importing their first season.

### Notes

- No new regexes added, just the existing multi-season regex extended to capture more than two markers.
- Kept the legacy `SeasonNumber` setter working so old code paths and stored JSON (`PendingReleases`) keep working without a migration.
- Added a round-trip test against the real JSON converter used for `PendingReleases` to make sure old and new season data work.

#### Submission Checklist

Not closing but related to #8938


src/NzbDrone.Core/Parser/Model/ParsedEpisodeInfo.cs
```diff
@@ -13,15 +14,23 @@ public class ParsedEpisodeInfo
         public string SeriesTitle { get; set; }
         public SeriesTitleInfo SeriesTitleInfo { get; set; }
         public QualityModel Quality { get; set; }
-        public int SeasonNumber { get; set; }
+
+        [JsonPropertyOrder(-1)]
```
mynameisbogdan, line 20: Does it need to be first?

```suggestion
```

src/NzbDrone.Core/Parser/Parser.cs
```diff
@@ -1069,21 +1069,24 @@ private static ParsedEpisodeInfo ParseMatchCollection(MatchCollection matchColle
...
-                if (seasons.Any())
+                else if (distinctSeasons.Count == 2)
+                {
+                    result.SeasonNumbers = Enumerable.Range(distinctSeasons.First(), distinctSeasons.Last() - distinctSeasons.First() + 1).ToArray();
+                }
+                else if (distinctSeasons.Count == 1)
                 {
-                    // If at least one season was parsed use the first season as the season
-                    result.SeasonNumber = seasons.First();
+                    result.SeasonNumbers = new[] { distinctSeasons[0] };
```
mynameisbogdan: We can use collections in v5. 

```suggestion
                    result.SeasonNumbers = [distinctSeasons.First()];
```

src/NzbDrone.Core/Parser/Parser.cs
```diff
@@ -1069,21 +1069,24 @@ private static ParsedEpisodeInfo ParseMatchCollection(MatchCollection matchColle
...
                 }
                 else if (!result.AbsoluteEpisodeNumbers.Any() && result.EpisodeNumbers.Any())
                 {
                     // If no season was found and it's not an absolute only release it should be treated as a mini series and season 1
-                    result.SeasonNumber = 1;
+                    result.SeasonNumbers = new[] { 1 };
```
mynameisbogdan: We can use collections in v5.

```suggestion
                    result.SeasonNumbers = [1];
```

src/NzbDrone.Core/Parser/Model/ParsedEpisodeInfo.cs
IonBazan: This might look tricky - it controls the serialization and deserialization order of this property for backward compatibility - see `EmbeddedDocumentConverterFixture.cs` test.

mynameisbogdan: Order shouldn't matter IMO. 

I'll take a look later.

IonBazan: Just double-checked: The test actually passes when we remove the `JsonPropertyOrder`, but it's only because `SeasonNumbers` property is declared after `SeasonNumber` - unintentional reordering of properties would break the actual behaviour. I believe that's because reflection returns properties in the order of their declaration. This is why I believe it's worth keeping it there for explicit contract. 

Happy to hear about any alternative approaches though - I'm still new to C# world.

mynameisbogdan: > I'm still new to C# world.

Please try using a modern IDE like Rider, it's free for FOSS too. It will tell you when to use collections, instead of `Array.Empty<int>()`, etc.

mynameisbogdan: > This might look tricky - it controls the serialization and deserialization order of this property for backward compatibility - see `EmbeddedDocumentConverterFixture.cs` test.

I'm missing the issue in `EmbeddedDocumentConverterFixture`, where should I look?

I removed `JsonPropertyOrder` and deserialize still work fine as expected as order isn't important in JSON. I would only matter if you're doing a serialize and comparing the output.

src/NzbDrone.Core.Test/Datastore/Converters/EmbeddedDocumentConverterFixture.cs
```diff
@@ -0,0 +1,59 @@
...
+            return new EmbeddedDocumentConverter<ParsedEpisodeInfo>().Parse(json);
+        }
+
+        [Test]
+        public void should_round_trip_season_numbers_for_a_multi_season_pack()
+        {
+            var parsedEpisodeInfo = new ParsedEpisodeInfo
+            {
+                SeriesTitle = "Series Title",
+                SeasonNumbers = new[] { 1, 2, 3 },
```
mynameisbogdan:
```suggestion
                SeasonNumbers = [1, 2, 3],
```

src/NzbDrone.Core/Parser/Model/ParsedEpisodeInfo.cs
```diff
@@ -13,15 +14,23 @@ public class ParsedEpisodeInfo
...
+        public int SeasonNumber
+        {
+            get => SeasonNumbers.Length > 0 ? SeasonNumbers[0] : 0;
+            set => SeasonNumbers = new[] { value };
```
mynameisbogdan:
```suggestion
            set => SeasonNumbers = [value];
```

src/NzbDrone.Core/Parser/Model/ParsedEpisodeInfo.cs
```diff
@@ -13,15 +14,23 @@ public class ParsedEpisodeInfo
...
+        }
+
+        public int[] SeasonNumbers { get; set; } = Array.Empty<int>();
```
mynameisbogdan:
```suggestion
        public int[] SeasonNumbers { get; set; } = [];
```

src/NzbDrone.Core/Parser/Parser.cs
```diff
@@ -1069,21 +1069,20 @@ private static ParsedEpisodeInfo ParseMatchCollection(MatchCollection matchColle
...
-                    result.SeasonNumber = seasons.First();
+                    result.SeasonNumbers = Enumerable.Range(distinctSeasons.First(), distinctSeasons.Last() - distinctSeasons.First() + 1).ToArray();
                 }
                 else if (!result.AbsoluteEpisodeNumbers.Any() && result.EpisodeNumbers.Any())
                 {
                     // If no season was found and it's not an absolute only release it should be treated as a mini series and season 1
-                    result.SeasonNumber = 1;
+                    result.SeasonNumbers = [1];
                     result.IsMiniSeries = true;
                 }
```
mynameisbogdan: Maybe use could use array from the start, not list + array.

```csharp
                var distinctSeasons = seasons.Distinct().OrderBy(s => s).ToArray();

                if (distinctSeasons.Length == 2)
                {
                    result.SeasonNumbers = Enumerable.Range(distinctSeasons.First(), distinctSeasons.Last() - distinctSeasons.First() + 1).ToArray();
                }
                else if (distinctSeasons.Length is 1 or > 2)
                {
                    result.SeasonNumbers = distinctSeasons;
                }
                else if (!result.AbsoluteEpisodeNumbers.Any() && result.EpisodeNumbers.Any())
                {
                    // If no season was found, and it's not an absolute only release it should be treated as a mini series and season 1
                    result.SeasonNumbers = [1];
                    result.IsMiniSeries = true;
                }
```

src/NzbDrone.Core/Parser/Model/ParsedEpisodeInfo.cs
IonBazan: I changed the property order to demonstrate the issue - removing the annotation now will cause test to fail. An alternative would be to add `[JsonIgnore(Condition = JsonIgnoreCondition.WhenWriting)]` to exclude that field from writing completely. WDYT?

src/NzbDrone.Core/Parser/Parser.cs
IonBazan: This is really neat, thanks!

src/NzbDrone.Core/Parser/Model/ParsedEpisodeInfo.cs
mynameisbogdan: How are running the tests? They didn't failed for me in Rider. 

Can you remove the order and push it here so I can see in the CI?

src/NzbDrone.Core/Parser/Model/ParsedEpisodeInfo.cs
IonBazan: There are plenty of existing fields using old syntax - I can migrate them separately or just before merging this PR if you want.

markus101: No need to clean up existing things at this time.

mynameisbogdan: Just new additions is enough for now.

src/NzbDrone.Core/Parser/Model/ParsedEpisodeInfo.cs
IonBazan: Make sure you pull latest changes:
image

This is after I changed the `SeasonNumbers` declaration to be before `SeasonNumber` property in 7577e0f. Removing the `JsonPropertyOrder` makes the tests red.

mynameisbogdan: Ah, I understand now. It's due to `Serialize(parsedEpisodeInfo)` outputting `{"seasonNumbers": [1,2, 3],  "seasonNumber": 1}` and `Deserialize(json)` is setting the whole list to a single number.

TBH, we could just use a HastSet like 
```csharp
        public HashSet<int> SeasonNumbers { get; set; } = [];
        public int SeasonNumber
        {
            get => SeasonNumbers.FirstOrDefault(0);
            set => SeasonNumbers.Add(value);
        }
```

But at this point we should just refactor to use `SeasonNumbers` as a list everywhere, right?

markus101: > But at this point we should just refactor to use SeasonNumbers as a list everywhere, right?

Short answer, yes, a little longer answer, is there might be some v3 things that expect it, but I think that should be only on things the API outputs, not things it accepts. So it would be better to have a `ParsedEpisodeInfoResource` instead of returning `ParsedEpisodeInfo` directly.

IonBazan: @markus101 should I do both (separate API resource, remove old field completely) now or in next PR? I was considering replacing all usages of SeasonNumber with the new field but I'm afraid this could also require adding all episodes to seasons in some places (stage 2 we discussed).

We can get this merged as starting point, to pass CI approvals in subsequent PRs, if you don't mind.

markus101: I think it's fine to do it in a follow up PR, will try to give this a further review tomorrow.

src/NzbDrone.Core/Parser/Model/ParsedEpisodeInfo.cs
```diff
@@ -13,15 +14,23 @@ public class ParsedEpisodeInfo
         public string SeriesTitle { get; set; }
         public SeriesTitleInfo SeriesTitleInfo { get; set; }
         public QualityModel Quality { get; set; }
-        public int SeasonNumber { get; set; }
+        public int[] SeasonNumbers { get; set; } = [];
+
+        [JsonPropertyOrder(-1)]
```
markus101, line 20:
```suggestion
        // TODO: Remove this once `SeasonNumbers` replaces `SeasonNumber`
        [JsonPropertyOrder(-1)]
```

markus101 (approved):
Going to add that TODO so hopefully it doesn't get forgotten about.

PR #8956, opened 2026-09-11
Fixed: Don't cache media covers that aren't downloaded

markus101:
#### Description

All media cover images are cached with a partial hash of the remote URL, the problem is when the file doesn't yet exist in disk we provide the same cache, when it's eventually downloaded the URL doesn't change so the UI doesn't see an updated URL and keeps showing the error image instead.

A small cache on the backend allows us to cache whether the file exists on disk and when it does we'll add the `h` query param, otherwise we won't. When the `h` query param is missing we won't tell the browser to cache the image either. When the file becomes available the cache is updated and SignalR or new API requests will serve up the image with the correct query param allowing the UI to get the updated image.

#### Submission Checklist

- https://github.com/Radarr/Radarr/issues/11700


src/NzbDrone.Core/MediaCover/MediaCoverService.cs
```diff
@@ -70,7 +76,7 @@ public string GetCoverPath(int seriesId, MediaCoverTypes coverType, int? height
             return Path.Combine(GetSeriesCoverPath(seriesId), coverType.ToString().ToLowerInvariant() + heightSuffix + GetExtension(coverType));
         }
 
-        public void ConvertToLocalUrls(int seriesId, IEnumerable<MediaCover> covers)
+        public void ConvertToLocalUrls(int seriesId, DateTime added, IEnumerable<MediaCover> covers)
```
mynameisbogdan, line 79: What do you think if you put it last and make it nullable? Radarr makes use of it for other resources, like cast/crew images that won't be ever fetched. 

Maybe something can be done eventually in the future for sonarr for artists, season or episode images!?! 

```suggestion
        public void ConvertToLocalUrls(int seriesId, IEnumerable<MediaCover> covers, DateTime added = null)
```

markus101: Yeah, that makes sense, I'll flip it.

PR #8958, opened 2026-09-11
Resolve episodes across every season in multi-season packs

IonBazan:
#### Description

Stage 2 of the multi-season pack work (see #8938, follow-up to #8941). Episode resolution now spans every season in a pack instead of only the first one, and search results no longer show a false "Wrong season" rejection for a season that's actually covered by the pack.

- `ParsingService.GetEpisodes` resolves episodes across the full `SeasonNumbers` range for a multi-season pack, applying any scene-mapping offset to every season instead of just the first.
- `SeasonMatchSpecification` and `SingleEpisodeSearchMatchSpecification` now check whether the searched season falls within `SeasonNumbers` instead of comparing against only the first season. The pack is still rejected everywhere by `MultiSeasonSpecification`, this only fixes the accompanying rejection reason so it doesn't also claim "wrong season" when the season searched for is actually in the pack.
- Replaced remaining internal and test usages of the legacy `SeasonNumber` with `SeasonNumbers`. `SeasonNumber` stays on `ParsedEpisodeInfo` only as the accessor for old JSON on queued pending releases; nothing else reads or writes it anymore.
- Exposed `SeasonNumbers` on the release API resources (V3 `ReleaseResource`, V5 `ParsedEpisodeInfoResource`) alongside the existing `SeasonNumber` field.
- Fixed `.With(p => p.SeasonNumber == 1)` predicate in `MatchesFolderSpecificationFixture` that used `==` instead of `=` and never actually set the value.

#### Submission Checklist

Towards #8938


src/Sonarr.Api.V5/Release/ParsedEpisodeInfoResource.cs
```diff
@@ -30,7 +31,8 @@ public static ParsedEpisodeInfoResource ToResource(this ParsedEpisodeInfo parsed
             ReleaseGroup = parsedEpisodeInfo.ReleaseGroup,
             ReleaseHash = parsedEpisodeInfo.ReleaseHash,
             FullSeason = parsedEpisodeInfo.FullSeason,
-            SeasonNumber = parsedEpisodeInfo.SeasonNumber,
+            SeasonNumber = parsedEpisodeInfo.SeasonNumbers.FirstOrDefault(),
```
IonBazan, line 34: I guess this can be removed here?

src/NzbDrone.Core.Test/DecisionEngineTests/Search/SingleEpisodeSearchMatchSpecificationTests/StandardEpisodeSearch.cs
```diff
@@ -79,5 +81,28 @@ public void should_return_true_if_full_season_result_for_full_season_search()
         {
             Subject.IsSatisfiedBy(_remoteEpisode, _information).Accepted.Should().BeTrue();
         }
+
+        [Test]
+        public void should_return_true_if_searched_season_is_within_multi_season_pack_range()
+        {
+            // e.g. "Show.S05E24.S06E01" - a season finale bundled with the next season's premiere.
```
markus101: Is this a real example because I think there is only one example with a Star Trek episode that spans seasons, but it's incredibly rare.

src/Sonarr.Api.V5/Release/ParsedEpisodeInfoResource.cs
markus101: I think so, will probably need some UI changes too.

src/NzbDrone.Core/Parser/ParsingService.cs
```diff
@@ -253,13 +253,24 @@ public List<Episode> GetEpisodes(ParsedEpisodeInfo parsedEpisodeInfo, Series ser
...
 
         private List<Episode> GetEpisodes(ParsedEpisodeInfo parsedEpisodeInfo, Series series, int mappedSeasonNumber, bool sceneSource, SearchCriteriaBase searchCriteria)
         {
             if (parsedEpisodeInfo.FullSeason)
             {
+                if (parsedEpisodeInfo.IsMultiSeason)
+                {
+                    // Scene numbering is per-season, so it isn't meaningful for a pack spanning
+                    // multiple seasons; apply the scene-mapping offset (if any) to every season instead.
+                    var offset = mappedSeasonNumber - parsedEpisodeInfo.SeasonNumbers.FirstOrDefault();
```
markus101: I think this is going to be interesting, really depends if the release follows TheTVDB or something else entirely.

PR #8966, opened 2026-09-15
Fixed: Plex only updates one matching library section

kushiemoon-dev:
same bug as Radarr/Radarr#11698, ported the fix over.

when 2+ plex sections point at the same folder only the first one got
refreshed after an import/delete, loop stopped as soon as it found one
match. now it keeps going and updates every section that matches, no
match still falls back to updating everything like before.

added tests for the multi match case, single match, and the fallback.


mynameisbogdan:
Can you rebase with a single commit instead?

kushiemoon-dev:
rebased it down to a single commit

kushiemoon-dev:
Will correct but was have check and re test everything in a sandbox working sonarr.

PR #8976, opened 2026-09-19
Fixed: Manual Interaction - Required notifications failing for unknown series

rubegartor:
#### Description

Fixes a `NullReferenceException` in `NotificationService.Handle(ManualInteractionRequiredEvent)` that stopped "Manual Interaction Required" notifications (Custom Script, etc.) from being sent when the download couldn't be matched to a series.

When Sonarr can't match the download to a series, `Episode` (the `RemoteEpisode`) is null. The rest of the handler already allows for that, but the tag check dereferenced `message.Episode.Series` directly.

Changes:
- `ShouldHandleSeries` now handles a null series: connections without tags still receive the notification, while connections with tags skip it, consistent with how tag filtering works for every other event.
- `Handle(ManualInteractionRequiredEvent)` uses null-conditional access for `Episode?.Series` and `ParsedEpisodeInfo?.Quality`.
- Custom Script `OnManualInteractionRequired` no longer throws when the series is unknown (series variables are skipped) or when `DownloadItem` is null (empty `Sonarr_Download_Size` / `Sonarr_Download_Title`).

#### Submission Checklist

Same fix applied in: https://github.com/Radarr/Radarr/pull/11716


IonBazan:
Can we cover this with a test?

src/NzbDrone.Core/Notifications/CustomScript/CustomScript.cs
```diff
@@ -310,13 +310,17 @@ public override void OnManualInteractionRequired(ManualInteractionRequiredMessag
...
+                AddSeriesVariables(environmentVariables, series);
+            }
 
             environmentVariables.Add("Sonarr_Download_Client", message.DownloadClientInfo?.Name ?? string.Empty);
             environmentVariables.Add("Sonarr_Download_Client_Type", message.DownloadClientInfo?.Type ?? string.Empty);
             environmentVariables.Add("Sonarr_Download_Id", message.DownloadId ?? string.Empty);
-            environmentVariables.Add("Sonarr_Download_Size", message.TrackedDownload.DownloadItem.TotalSize.ToString());
-            environmentVariables.Add("Sonarr_Download_Title", message.TrackedDownload.DownloadItem.Title);
+            environmentVariables.Add("Sonarr_Download_Size", message.TrackedDownload.DownloadItem?.TotalSize.ToString() ?? string.Empty);
+            environmentVariables.Add("Sonarr_Download_Title", message.TrackedDownload.DownloadItem?.Title ?? string.Empty);
```
markus101, line 320: How does Sonarr get here without a tracked download + download item?

src/NzbDrone.Core/Notifications/CustomScript/CustomScript.cs
```diff
@@ -310,13 +310,17 @@ public override void OnManualInteractionRequired(ManualInteractionRequiredMessag
             var environmentVariables = new StringDictionary();
 
             AddInstanceVariables(environmentVariables, "ManualInteractionRequired");
-            AddSeriesVariables(environmentVariables, series);
+
+            if (series != null)
```
markus101: I think an early return in `AddSeriesVariables` is nicer.

markus101 (changes requested):

src/NzbDrone.Core/Notifications/CustomScript/CustomScript.cs
mynameisbogdan: I guess to be on the safer side, even if technically it shouldn't happen.

rubegartor: It really shouldn't happen, but to be on the safe side and make it as similar as possible to the fix implemented in Radarr

rubegartor:
> Can we cover this with a test?

Done

PR #8977, opened 2026-09-20
Internal restart fixes

TypNull:
#### Description

With the internal restart system in place, the `DatabaseTarget` has to be unregistered before restarting. Currently that check runs against `null` and therefore never unregisters. This fix looks up the registered `DatabaseTarget` and unregisters it. Otherwise the old target survives the restart without its filter rules and spams the event log.

Also, as Bogdan noticed, after a restart the old container is still around and timers from it (e.g. the health check triggered by saving a host setting) can still fire and publish events, which leads to a `Container is disposed` error. Events published through a disposed container are now dropped, so the old host stays quiet while clients reconnect to the new one.

#### Submission Checklist

- [X] I have fully reviewed all code in this PR, tested it and hope it works as intended

https://github.com/Radarr/Radarr/issues/11712
(Sonarr and Lidarr too)


mynameisbogdan (approved):
Works as intended on Radarr.

Thanks for looking into it.

